PR #21, opened 2024-09-28
feat(nuxt-example): add sign-up and password reset pages, update auth…

Dyrean:
Hey,

Added some basic missing features for Nuxt, similar features that svelte-kit-example has. Could be improve for more features such as organization, twoFactor in the feature like next.js demo has.


Bekacru:
Thanks, this is great.

Issue #26, opened 2024-09-29
BLOB/TEXT column 'id' used in key specification without a key length


joesoeph:
I am getting an error when running the command `npx better-auth migrate` on MySQL Database. This is an example of the `auth.ts` I am using.

```typescript
export const auth = betterAuth({
  database: {
    provider: 'mysql',
    url: 'mysql://root:@127.0.0.1:3306/superOceanDB',
  },
  emailAndPassword: {
    enabled: true,
  },
  user: {
    modelName: 'users',
  },
  account: {
    modelName: 'accounts',
  },
  session: {
    modelName: 'sessions',
  },
});
```

And this is the error that appears in the terminal window:
```bash
...at async CreateTableBuilder.execute (file:///Users/yusuf/Documents/sandbox/next-js/super-ocean/node_modules/kysely/dist/esm/schema/create-table-builder.js:350:9) {
  code: 'ER_BLOB_KEY_WITHOUT_LENGTH',
  errno: 1170,
  sqlState: '42000',
  sqlMessage: "BLOB/TEXT column 'id' used in key specification without a key length",
  sql: 'create table `users` (`id` text primary key, `name` text not null, `email` text not null, `emailVerified` boolean not null, `image` text, `createdAt` date not null, `updatedAt` date not null)'
}

Node.js v20.14.0
```

The error does not occur when running with an SQLite database.

When I debug the generated query,
```sql
CREATE TABLE `users` (
	`id` text PRIMARY KEY,
	`name` text NOT NULL,
	`email` text NOT NULL,
	`emailVerified` boolean NOT NULL,
	`image` text,
	`createdAt` date NOT NULL,
	`updatedAt` date NOT NULL)
```
I get the same error

`Query 1 ERROR at Line 1: : BLOB/TEXT column 'id' used in key specification without a key length`

I suspect this is due to the table schema generator.

baraich:
I believe the query should be 

```sql
CREATE TABLE `users` (
	`id` text PRIMARY KEY NOT NULL,
	`name` text NOT NULL,
	`email` text NOT NULL,
	`emailVerified` boolean NOT NULL,
	`image` text,
	`createdAt` date NOT NULL,
	`updatedAt` date NOT NULL
)
```

baraich:
@Bekacru Bro, what have you done, my exam was tomorrow and for contribution I neglected it and you merged another pull request. And you can also see that I made a commit earlier to you.

jericopulvera:
[...]

PR #28, opened 2024-09-29
closes: #26

baraich:
## Changes
According to the requirements the following error occurred because MySQL can index only the first _N_ chars of a `BLOB` or `TEXT` column. So The error mainly happens when there is a field/column type of `TEXT` or `BLOB` or those belong to `TEXT` or `BLOB` types such as `TINYBLOB`, `MEDIUMBLOB`, `LONGBLOB`, `TINYTEXT`, `MEDIUMTEXT`, and `LONGTEXT` that you try to make a primary key or index. With full `BLOB` or `TEXT` without the length value, MySQL is unable to guarantee the uniqueness of the column as it’s of variable and dynamic size. So, when using `BLOB` or `TEXT` types as an index, the value of N must be supplied so that MySQL can determine the key length. However, MySQL doesn’t support a key length limit on `TEXT` or `BLOB`. `TEXT(88)` simply won’t work.

A workaround, was conditionally rendering type in the schema generated. When the `dbType` is mysql then we set the type to `VARCHAR(255)` otherwise `TEXT`.

> `VARCHAR` can accepts upto 256 characters.

### DIFF
--- a/packages/better-auth/src/cli/utils/get-migration.ts
+++ b/packages/better-auth/src/cli/utils/get-migration.ts
@@ -1,7 +1,4 @@
-import type {
-	AlterTableColumnAlteringBuilder,
-	CreateTableBuilder,
-} from "kysely";
+import type { AlterTableColumnAlteringBuilder, CreateTableBuilder } from "kysely";
 import type { FieldAttribute, FieldType } from "../../db";
 import { logger } from "../../utils/logger";
 import type { BetterAuthOptions } from "../../types";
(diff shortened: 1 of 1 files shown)


Bekacru:
oh It just got addressed https://github.com/better-auth/better-auth/pull/27
thanks! anyway

baraich:
@Bekacru Welcome, brother. Though it was going to be my first opensource contribution.

baraich:
@Bekacru Please can you go through the workaround that I gave and atleast give feedback on it, it's really devastating for me that I spent an 45-50 minutes neglecting tomorrows board examination. Looking forward for feedback from you.

Thanks in advance.

PR #47, opened 2024-09-30
docs: Better const name for auth client usage

paaauldev:


Kinfe123:
since all of the tabs of code snippets do named as client .. it creates inconsistency  tho

Bekacru:
> since all of the tabs of code snippets do named as client .. it creates inconsistency tho

we should standardize authClient instead through out the docs

PR #70, opened 2024-10-02
ft: delete user plugin

SooditK:
### Adds a Delete User Plugin

it takes a parameter, whether the organization plugin is being used

if there are organizations associated to the user, the user will not be marked as deleted

P.S. Please ignore `pnpm-lock.yaml`

### DIFF
--- a/examples/remix-example/package.json
+++ b/examples/remix-example/package.json
@@ -45,7 +45,7 @@
     "@remix-run/react": "^2.12.1",
     "@remix-run/serve": "^2.12.1",
     "@types/better-sqlite3": "^7.6.11",
-    "better-auth": "workspace:0.2.9-beta.7",
+    "better-auth": "workspace:*",
     "better-sqlite3": "^11.3.0",
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.1.1",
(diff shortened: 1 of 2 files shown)


Bekacru:
Hey, sorry but could you implement soft delete directly in the core instead of as a plugin? Add an option to choose between hard and soft delete. For soft deletes, we’ll need an extra field in the core with a `deletedAt` column instead of `isDeleted`.

Bekacru:
And there is already hard delete in the core

SooditK:
Hii @Bekacru no worries, I'd be happy to implement it in the core
I'm still not sure that why do we need a separate table for this, we can easily update either the `user` or `account` table with the `isDeleted` boolean flag

one more thing - if we do it in a separate `deleted` table, then every time we fetch/sign-in as a user, then we need to do an extra join to check if the user is deleted or not

let me know your thoughts

Bekacru:
> Hii @Bekacru no worries, I'd be happy to implement it in the core
> 
> I'm still not sure that why do we need a separate table for this, we can easily update either the `user` or `account` table with the `isDeleted` boolean flag
> 
> 
> 
> one more thing - if we do it in a separate `deleted` table, then every time we fetch/sign-in as a user, then we need to do an extra join to check if the user is deleted or not
> 
> 
> 
> let me know your thoughts

Sorry I meant extra field

PR #83, opened 2024-10-04
ft: soft delete in core

SooditK:
Adds a `deletedAt` field at the core
prevents the user from signing in

also updating `BetterAuthOptions` interface with a `softDelete` field, if true it will populate the `deletedAt` field otherwise it will delete the user

---

Quick Question: in soft delete, we don't touch the `account` table at all (which is fine), but in hard delete, shouldn't we also delete the account & the sessions related to that user

and if organization plugin is used and the user is part of an organization, what do you think should be the behavior of soft delete & hard delete?


Bekacru:
and please add a proper test for the functionality

SooditK:
Hii @Bekacru please take a look at the updated code, I tried everything but the snapshots were failing due to some weird spacing issue I assume

everything else should be working fine, please let me know if I'm missing anything here
👍🏼

Bekacru:
Thanks for the PR. We decided not to implement soft deletes, as they complicate things at the moment. :))

platform-kit:
@Bekacru is there any way to implement soft-deletes via some kind of workaround? A hook of some kind?

PR #98, opened 2024-10-05
feat: use `localhost` as email domain in anonymous auth

typed-sigterm:
Although `temporary-{timestamp}-better-auth@email.com` is almost impossible to cause conflict, `email.com` may cause confusion 😕 Maybe the approach from Git is better: `localhost`.

PS: As `localhost` (actually not a domain) is different from other domains, it does not contain `.`, which may cause some RegExps to fail. May need to be explained in the docs.
PS 2: Considering support for clustering and concurrency, perhaps UUID or `generateId()` would be better than timestamp.


Bekacru:
Yeah I agree, we should switch to using `uuid` and instead of `localhost` we should use the actual domain since it's already scoped. Also, let's remove the `better-auth` prefix. It should be `temp-${uuid}@${domainName}`. We should also allow for `domainName` overrides in the plugin options.

typed-sigterm:
> Yeah I agree, we should switch to using `uuid` and instead of `localhost` we should use the actual domain since it's already scoped. Also, let's remove the `better-auth` prefix. It should be `temp-${uuid}@${domainName}`. We should also allow for `domainName` overrides in the plugin options.

I'm sorry that I can't identify what *scope* and *actual domain* refers to. (not a native English speaker) I came up with `localhost` just because it's identifiable and understandable.
Others (except tests) were added in e8ab665 :heart:

Bekacru:
If you can please fix the conflict and this will be merged.

Bekacru:
Thanks! I've merged it into main, so I'll go ahead and close this.

PR #106, opened 2024-10-06
fix: handle unsupported returning in Drizzle MySQL adapter

### LINKED ISSUE #105: Using Drizzle fails on creation due to unsupported use of `returning()` with MySQL
MySQL itself doesn’t have native support for RETURNING after using INSERT:

https://orm.drizzle.team/docs/insert#insert-returningid

The adapter is using the code for PostgreSQL and SQLite here:

https://github.com/better-auth/better-auth/blob/41d81565f5eeefbca85e356bc793639d4d4c5381/packages/better-auth/src/adapters/drizzle-adapter/index.ts#L93

inducingchaos:
### Related Issue
Closes #105

### Changes
- Updated the DTS dev script memory allocation to avoid overflow issues.
- Modified the Drizzle adapter to bypass the `RETURNING` clause for MySQL databases, and instead use the ID from `data` to fetch the row.

### Testing
- Manually tested on a MySQL instance to ensure correct behaviour for insert queries. Update remains untested.


Bekacru:
thanks, this looks good. Could you resolve the conflicts and override the lock file from the main branch?

Bekacru:
merged on main

PR #108, opened 2024-10-07
only set session cookie if `sendEmailVerificationOnSignUp` is falsy

saaskitten:
Allowing unverified email addresses to login can lead to various vulnerability issues/risks. This implements mandatory email verification before they can sign in when `sendEmailVerificationOnSignUp` is `true`.


Bekacru:
I don't think making email verification a strict default is the expected experience. It's often better to allow users to access a session even if their email isn't verified, leaving it up to the app to enforce restrictions on certain features until verification is complete. And If you block session creation during signup, you'd also have to handle preventing sessions at login for unverified emails, which adds unnecessary complexity.

Issue #124, opened 2024-10-08
[REFACTOR] Improve logging


tklein1801:
As a developer and operator of my auth-service, I would like to see cleaner logging to get more information about what is happening. The current logging is very vague and only outputs logs when errors occur (as it seems).

Therefore, it would be desirable to improve the logging and to offer a possibility to transport the logs to an external service. Winston, for example, would be suitable for this, giving you the option of only outputting logs, saving them in a file or transferring them to external services or our own database.

PR #125, opened 2024-10-08
refactor: Use winston for logging

tklein1801:
Created for #124 

I have installed [Winston](https://www.npmjs.com/package/winston) and replaced the logging. Before the branch is merged, the current log format should be removed as well as the question of whether there should be a ‘transport’ for the logs by default.

- [ ] Approve or rework new log format
- [ ] Improve logging (add more logging)
- [ ] Update [documentation](https://www.better-auth.com/docs/introduction)

### DIFF
--- a/docs/content/docs/authentication/email-password.mdx
+++ b/docs/content/docs/authentication/email-password.mdx
@@ -42,7 +42,6 @@ To signup a user, you can use the `signUp.email` function provided by the client
 - `name`: The name of the user.
 - `image`: The image of the user. (optional)
 - `callbackURL`: The url to redirect to after the user has signed up. (optional)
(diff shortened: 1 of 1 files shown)


Bekacru:
My only objection is that winston feels huge to include with an auth library. It’d be great if we could find a more lightweight alternative.

tklein1801:
I'll take a look around some logging libraries

tklein1801:
@Bekacru could I light your interest in [pino](https://github.com/pinojs/pino/tree/main)? 
Way smaller than winston, actively maintained and does support custom transporting your logs

Chris-Kode:
i have been using the pino library, and its working well for logging, is a good option

frasza:
Pino does sound great tbh.

PR #155, opened 2024-10-12
Docs: Elysia integration

yukikwi:
This is an integration document for better auth & [elysia](https://elysiajs.com/)


yukikwi:
Example code: https://github.com/yukikwi/elysia-boiler-plate

PR #189, opened 2024-10-16
fix: OAuth flow errors

### LINKED ISSUE #186: Facebook oauth_code_verification_failed due to missing code_challenge
**Describe the bug:**

When using the library to authenticate via Facebook OAuth with PKCE, the token exchange step fails with the error: oauth_code_verification_failed. The error message indicates that the code_verifier is provided without a corresponding code_challenge in the authorization request, causing the OAuth process to fail.

**To Reproduce:**
- Set up Facebook OAuth in the project.
- Initiate the OAuth flow by redirecting to the Facebook login page.
- After the user logs in and the authorization code is returned, the library attempts to exchange the code for access tokens.
- The process fails with the error: oauth_code_verification_failed.

**Expected behavior:**

It should successfully handle the PKCE flow, ensuring that both the code_verifier and code_challenge are correctly sent and validated. The access token should be retrieved after the authorization code is exchanged.

**Additional context:**

Error message received during the token exchange:
```
{ 
  error: { 
    message: 'code_verifier param is provided without a code_challenge in authorization request.',
    type: 'OAuthException',
    code: 1,
    fbtrace_id: 'AiVg2G-HnpatNZDjjXMWGXo'
  },
  status: 400,
  statusText: 'Bad Request'
}
```

Bekacru:
closes #186

### DIFF
--- a/packages/better-auth/src/oauth2/utils.ts
+++ b/packages/better-auth/src/oauth2/utils.ts
@@ -24,7 +24,11 @@ export function getOAuth2Tokens(data: Record<string, any>): OAuth2Tokens {
 		accessTokenExpiresAt: data.expires_at
 			? new Date((Date.now() + data.expires_in) * 1000)
 			: undefined,
-		scopes: data.scope?.split(" ") || [],
+		scopes: data?.scope
+			? typeof data.scope === "string"
+				? data.scope.split(" ")
+				: data.scope
+			: [],
 		idToken: data.id_token,
 	};
 }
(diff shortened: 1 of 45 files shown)


packages/better-auth/src/oauth2/utils.ts
```diff
@@ -24,7 +24,11 @@ export function getOAuth2Tokens(data: Record<string, any>): OAuth2Tokens {
 		accessTokenExpiresAt: data.expires_at
 			? new Date((Date.now() + data.expires_in) * 1000)
 			: undefined,
-		scopes: data.scope?.split(" ") || [],
+		scopes: data?.scope
+			? typeof data.scope === "string"
+				? data.scope.split(" ")
```
ahmedrowaihi, line 25: ain't better to split with `,` instead of white-space ?
so scopes strings be like: `scope1, scope2` instead of `scope1 scope2`

PR #210, opened 2024-10-17
docs:fix: Fix error on type Session

Loa212:
since `createAuthClient()` and `betterAuth()`  do not have the required props, $Infer is breaking in the example. 

PS. I am not super-familiar with twoslash or fumadocs so forgive me if this is not the best approach to solve the issue.


Bekacru:
It should work. If this is cause the docs are failing to load on dev, make sure to build the package (`pnpm build`). Since the docs use the workspace package, you need to have it built for the type to work correctly.

PR #214, opened 2024-10-17
feat: add 'where' clause support to add filters in the list users, admin plugin

Chris-Kode:
This PR is to add the option to filter the users in the function listUsers for the admin plugin.


Bekacru:
let's rename where to filter

Chris-Kode:
@Bekacru probably we can extend to use some additional operators for example: "like", "ilike", "not like", "not ilike", more focused to do searchs. Added some info in the docs by the way

Bekacru:
Yeah we should add search functionality.  Instead of "like" it should just be search, so that it works across different databases. Thanks for this, I'll merge it now and feel free to give search a try if you can :))

PR #223, opened 2024-10-18
feat: custom email token generation

hehehai:
- Support custom email token


Bekacru:
Hey, could you add the option in the docs, and please run `pnpm lint:fix`, and let’s get this merged

opswiz:
Should the custom token generated have a minimum length constrain for security reasons 🤔. Can't allow like 1 or two chars, totally not safe.

Bekacru:
merged in #1362

PR #232, opened 2024-10-18
fix: baseUrl in vue client

NitrinCloud:
When an explicit baseUrl is defined in the client like shown in the docs. The website will throw following error since it currently tries to call //session.

Error thrown:
`[nuxt] [useFetch] the request URL must not start with "//".`

This pr uses the already implemented `getBaseURL` function which is used in other api calls as well, though it will still fallback to /api/auth in edge cases.


BayBreezy:
Just for ppl that might end up here.
I had a config where I was passing the `baseURL` prop to the client in Nuxt when creating it.
```ts
import { inferAdditionalFields } from "better-auth/client/plugins";

import { auth } from "../../lib/auth";

});
```

Remove that `baseURL` prop and the error goes away.

PR #234, opened 2024-10-19
fix: nuxt internal fetch url

NitrinCloud:
nuxt doesn't allow to fetch internal api using the complete url when rendering on the server.

This pr allows to still have a decoupled backend or a fullstack nuxt app.


matserix:
This fixes a bug caused by #232 which will always load a complete url out of the environment

BayBreezy:
Just for ppl that might end up here.
I had a config where I was passing the `baseURL` prop to the client in Nuxt when creating it.
```ts
import { inferAdditionalFields } from "better-auth/client/plugins";

import { auth } from "../../lib/auth";

});
```

Remove that `baseURL` prop and the error goes away.

PR #235, opened 2024-10-19
fix: use autosignin option from configuration

silverxp:


silverxp:
This pull request is wrong. Closing it.

PR #245, opened 2024-10-19
Feat: Added support to pre-build "like" operator, and added search endpoint for admin purposes.

Chris-Kode:
### Some features to make more searchable our users:

- Added some pre-build "like" features that its gonna help to make more complex queries and make more searchable.
- Added a endpoint /search-users to admin purpose, to be more complete about admin actions like, search a user by his name, etc.
- Added some tests, for the adapters, testing each new operator.

### DIFF
--- a/docs/content/docs/plugins/admin.mdx
+++ b/docs/content/docs/plugins/admin.mdx
@@ -119,6 +119,31 @@ const users = await authClient.admin.listUsers({
 });
 ```
 
+### Search Users
+
+Allows you to search for users by their email or name adding the possibility to make more complex queries.
+
+```ts title="admin.ts"
+const users = await authClient.admin.searchUsers({
+    search: {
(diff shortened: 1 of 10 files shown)


Bekacru:
This is really good! Thanks. 

One comment: we shouldn’t separate the search and list functions. Instead, we can add a search parameter to the list function. The same goes for the internal adapter—there's no need for another function. We can just use listUser with a search query.

Chris-Kode:
> This is really good! Thanks.
> 
> One comment: we shouldn’t separate the search and list functions. Instead, we can add a search parameter to the list function. The same goes for the internal adapter—there's no need for another function. We can just use listUser with a search query.

You are right will be simplier, and listUser be more powerful function

Chris-Kode:
@Bekacru done!

PR #282, opened 2024-10-21
fix: Anonymous linking

### LINKED ISSUE #283: Anonymous account linking not setting "isAnonymous" to false and documentation shows wrong function
**Describe the bug**
Anonymous account linking is not setting "isAnonymous" to false after an account is linked and documentation shows wrong function to call in the plugin description

bunducdragos:
Closes #283


Bekacru:
Thanks!

One comment, please update the api to `anonymous.linkAccount` instead. You'll just need to modify the path to `/anonymous/link-account`. And return the docs as they were.

PR #344, opened 2024-10-26
enhance alias path resolution

### LINKED ISSUE #343: path aliases not resolving correctly for cli
Hi

First off thank you for the effort everyone is putting into this library. There is definitely a need for alternative local auth libraries.

While playing around and trying to setup better-auth in a t3app I could not run the generate command when using the path aliases setup in tsconfig.json. This was addressed in the #80 but this doesn't seem to have resolved the issue in full.

**Describe the bug**
When using custom path aliases, the cli cannot correctly resolve the imports. 

**To Reproduce**
a minimal reproduction with nextjs and prisma was created on [CodeSandbox](https://codesandbox.io/p/devbox/jvgshq)

the following steps was taken to create the reproduction:

1. create new nextjs app
2. install prisma using sqlite as datasource.
3. install better-auth
4. create new folder "server" in the root of the project
5. add the following alias to the tsconfig: "@server/*": ["./server/*"]
6. create 2 folders in the server folder. "auth" and "db"
7. create prismaClient.ts file in db folder. Create and export prisma client in this file
8. create auth.ts file in the auth folder and setup auth config in the auth.ts using imported prisma client from "@server/db/prismaClient"
9. run "npx @better-auth/cli@latest generate --config server/auth/auth.ts"

**Expected behavior**
schema should be generated and added to the schema.prisma file in "./prisma" folder as this is the default output path

**Actual behavior**
cli reports an error: "Couldn't read your config".

**Additional context**
I've looked through get-config.ts and the getPathAliases function doesn't seem to correctly address the problem:
1) it hardcodes the path to "../" for all aliases
2) it assumes the alias is only one character long for example "@"
3) it does not cater for the different aliases that can be used according the typescript [documentation](https://www.typescriptlang.org/tsconfig/#paths).
4) it does not take into account the base url in tsconfig.json.
5) I have forked the repo and made alterations and will post a PR shortly

nlynzaad:
this pull request resolves #343 through:
 
1) calculating the resolved base url by extracting it from tsconfig and joining to the cwd that is at the root of the tsconfig file. the baseUrl is relative to the tsconfig file.
2) checking if the alias and/or alias path is of a resolver type or a direct alias to a file
3) since the alias paths is relative to the base url the resulting path is the path join of the resolvedBaseUrl and the alias path
4) ensure the full alias is used.

### DIFF
--- a/packages/cli/src/utils/get-config.ts
+++ b/packages/cli/src/utils/get-config.ts
@@ -9,6 +9,7 @@ import babelPresetReact from "@babel/preset-react";
 import fs from "fs";
 import { BetterAuthError } from "better-auth";
 import { addSvelteKitEnvModules } from "./add-svelte-kit-env-modules";
+import D from "path";
 
 let possiblePaths = ["auth.ts", "auth.tsx"];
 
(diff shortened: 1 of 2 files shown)


Bekacru:
this is great! thanks ❤️

PR #347, opened 2024-10-26
[docs] Update `basic-usage.mdx`

Snazzy72:
By default, `email.signUp` will automatically log the user in upon a successful signup. Make this known in the docs and show them that this behaviour can be changed


Bekacru:
could you please also add it in the reference?

ghost:
> could you please also add it in the reference?
Oh yea for sure. My bad.

PR #378, opened 2024-10-31
Add API option to add member without invite

Elliot-Alexander:
Changes to the organization plugin that adds a new method to add members to a specified organization. Realistically I see this being primarily server side, so keen to get some guidance around the implementation. 
The use case I had in mind was potentially allow organization admins to add others into org directly, without invite; to speed up onboarding and allow for better org management. 

```typescript
await client.organization.addMember({
	organizationId: orgId,
	userIdOrEmail: "test2@test.com",
	role: "member",
});
```


Bekacru:
Thanks for the PR! Yeah I think this would be useful. But as you mentioned, this needs to be a server-only. Allowing en endpoint exposed tot he client where someone can add a member to their organization without an invitation would be risky, as they could add anyone they want. All you need to make an endpoint server only is add `metadata` in the options. 


```ts
createAuthEndpoint("/path", {
   //...
   metadata: {
      SERVER_ONLY: true
   }
}, async(ctx)=> {})
```

Elliot-Alexander:
Oh fantastic that's so easy! Will update that asap. Credit where credit is due, you've done a fantastic job with this codebase, from directory structure to code quality it was super easy to jump into and work.

Elliot-Alexander:
How have you tested server only functions currently? I couldn't find any examples of ones that are implemented at the moment. (I've update the PR with the server only work and modified the logic to reflect this, as well as remove the existing test)

Bekacru:
You can use the `auth.api` object instead of the client.

Elliot-Alexander:
Apologies for the delays on this, am overseas without my machine

PR #396, opened 2024-11-02
fix: onRequest callback not interrupting requests

### LINKED ISSUE #395: Plugin onRequest: interrupting the request does not work
**Describe the bug**
The documentation states that "To stop the request and send a response, return an object with a response property that contains a Response object.". **However no matter what I tried the request does not get interrupted**.

**To Reproduce**

1. Create a basic plugin:
```ts
export const myPlugin = {
    id: "my-plugin",
    onRequest: async (request) => {
        return {
          response: new Response("Request interrupted", { status: 500 }),
        };
    },
}
```
2. Add the plugin to your better-auth instance
```ts
  // ...
  plugins: [
    myPlugin(),
  ],
  // ...
})
```
3. Run a better-auth instance
5. See that the requests are not interrupted

**Expected behavior**
All requests should be interrupted (with the code snippet above)

0scrm:
Fixes #395 

The response type coming back from `plugin.onRequest(req, ctx)` is 
```ts
const response: void | {
    response: Response;
} | {
    request: Request;
}
```

I suspect that because of that union type, we were returning `response` although it is not an actual Fetch Response. For this reason, the request was actually not interrupted.

This fix is a pretty simple solution, I also considered using a type guard, let me know what you think :) 
```ts
function hasResponseInterruptor(response: void | { response: Response } | { request: Request }): response is { response: Response } {
	return response !== undefined && typeof response === 'object' && 'response' in response;
}
```

### DIFF
--- a/packages/better-auth/src/api/index.ts
+++ b/packages/better-auth/src/api/index.ts
@@ -245,8 +245,8 @@ export const router = <C extends AuthContext, Option extends BetterAuthOptions>(
 			for (const plugin of ctx.options.plugins || []) {
 				if (plugin.onRequest) {
 					const response = await plugin.onRequest(req, ctx);
-					if (response) {
-						return response;
+					if (response && 'response' in response) {
+						return response.response
 					}
 				}
 			}


Bekacru:
Yeah, I think this should work. it looks good to me.

PR #355, opened 2024-10-27
feat: add `SurrealDB` adapter

oskar-gmerek:
What and why:

As a user I would like to keep my users data in my favorite database - [SurrealDB](https://surrealdb.com).
There is at least two ways to do it: 
- **implement dialect for Kysely** - I have no motivation to do that, I'm not using Kysely and current SurrealDB dialect is outdated
- **implement adapter for Better-Auth** - This is my choice and my attempt to achieve it

Plan:

- [x] Implement adapter
- [x] Make sure that all adapter tests passed
- [x] Create adapter documentation

PR #399, opened 2024-11-03
feat: add `SurrealDB` adapter

oskar-gmerek:
Continuation of #355 

@Bekacru This is currently up to date with main branch. Blocked CI should be resolved.


Bekacru:
hey! thanks for fixing the issue. I'm just reconsidering it, since maintaining the adapter going forward (before v1) isn’t something I’m really looking forward to :D And now there is a chance we may end up adding some features to the adapter, so let’s give it a week or two, and I’ll ping you here if we end up adding a feature so you could help. Then we’ll merge it

oskar-gmerek:
I see. Just ping me when you will be ready

Bekacru:
Hey @oskar-gmerek, sorry for the late response! could you make this its own package and open-source it? I’m thinking of creating a "Community-Maintained Adapters" section in the docs for those who need it. If we have to maintain it ourselves, it could easily become unmanageable but if you're up for it, it'd be really great.

oskar-gmerek:
@Bekacru That’s totally understandable. It’s actually a good idea to make it a separate package, and I will give it a try.

oskar-gmerek:
I have make public https://github.com/oskar-gmerek/surreal-better-auth.git

Would be great @Bekacru if you could review it

Bekacru:
Hey @oskar-gmerek, this looks great! thank you. Just one thing I noticed we should look into how we can expose the test instance and adapter tests so you won't have to copy that into your code. This way, it can stay up to date with future changes more easily. I'll open an issue on your repo once that's set up.

oskar-gmerek:
Yes, exposing that would be great. That will make a things much easier to maintain.

PR #403, opened 2024-11-03
Update sveltekit integration docs 

nalonix:
Update to sveltekit handler.
It's intended to attach the user session to **events.locals** so that it's available through out the app. 

This will give access to the session (and user data) in Page / Layout load functions [ _event.locals.session_ | _event.locals.user_ ] and Form actions [ _event.locals.session_ | _event.locals.user_ ]. 

For Svelte files, we can grab the session in a load function and pass it down to the svelte page / layout file.

For TS: requires updating Locals interface in `app.d.ts` by adding session and user properties with types Session and User respectively [ from _"better-auth/types"_ ].

**IF THIS IS NOT A VIABLE OPTION. WE CAN JUST UPDATE THE DOCS. SEE BELOW.** 

The session and user can be attached to **event.locals** inside `hooks.server.ts`.

```
import { auth } from '$lib/auth'; // path to your auth file
import { svelteKitHandler } from 'better-auth/svelte-kit';

export async function handle({ event, resolve }) {
		headers: event.request.headers
	});

	event.locals.session = session?.session;
	event.locals.user = session?.user;

	return svelteKitHandler({ event, resolve, auth });
}
```

and update `app.d.ts`

```
import type { User, Session } from 'better-auth/types';

declare global {
	namespace App {
		interface Locals {
			session: Session | undefined;
			user: User | undefined;
		}
	}
}

export {};
```

Acessing sessions / user and passing down as page data:

```
import { redirect } from '@sveltejs/kit';

export async function load({ locals }) {
	const session = locals.session;

	if (!session) {
		throw redirect(302, '/auth/signin');
	}

	return {
		session
	};
}
```
Now session is accessible inside the page like every other page data: 

```
// Svelte 5 
	const { data } = $props();
	console.log(data.session);

// Previous verions 
    export let data;
    console.log(data.session)
    
```

### DIFF
--- a/docs/content/docs/integrations/svelte-kit.mdx
+++ b/docs/content/docs/integrations/svelte-kit.mdx
@@ -1,24 +1,13 @@
 ---
+
 title: Svelte Kit Integration
 description: Learn how to integrate Better Auth with Svelte Kit
+
 ---
 
-Before you start, make sure you have a Better Auth instance configured. If you haven't done that yet, check out the [installation](/docs/installation).
+Before you start, make sure you have a Better Auth instance configured. If you haven't done that yet, check out the [installation](https://www.notion.so/docs/installation).
 
-### Mount the handler
-
-We need to mount the handler to svelte kit server hook.
-
-```ts title="hooks.server.ts"
-import { auth } from "$lib/auth";
-import { svelteKitHandler } from "better-auth/svelte-kit";
-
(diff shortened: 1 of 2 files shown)


Jumaron:
+1 to that, it is similar to what lucia was doing right?
Wonder if there are any drawbacks tho.

PR #408, opened 2024-11-03
fix: magic links sent to unregistered emails when disabled sign up

JSLNO:
As of right right now, we have to do something like this:
```ts
	// ...database
	plugins: [
		magicLink({
			disableSignUp: true,
			sendMagicLink: async (data: {
				email: string
				token: string
				url: string
			}) => {
				const isSignedUp =
					(await db.user.count({
						where: {
							email: data.email
						}
					})) > 0

				if (!isSignedUp) {
					throw new Error('User not found')
				}

				// ...send magic link to email
			}
		})
	]
})
```
but it brings some flaws with it. We are not able to provide an accurate error message to the client, because an `INTERNAL_SERVER_ERROR` is thrown.

With the commited adjustments an `USER_NOT_FOUND` error is thrown when a user isn't registered and disableSignUp is set to true. This allows us to provide a more detailed error message to the client.

### DIFF
--- a/packages/better-auth/src/plugins/magic-link/index.ts
+++ b/packages/better-auth/src/plugins/magic-link/index.ts
@@ -55,6 +55,17 @@ export const magicLink = (options: MagicLinkOptions) => {
 				},
 				async (ctx) => {
 					const { email } = ctx.body;
+
+					if (options.disableSignUp) {
+						const user = await ctx.context.internalAdapter.findUserByEmail(email)
+
+						if (!user) {
+							throw new APIError("USER_NOT_FOUND", {
+								message: "User not found"
+							});
+						}
+					}
+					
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/magic-link/index.ts
```diff
@@ -55,6 +55,17 @@ export const magicLink = (options: MagicLinkOptions) => {
 				},
 				async (ctx) => {
 					const { email } = ctx.body;
+
+					if (options.disableSignUp) {
+						const user = await ctx.context.internalAdapter.findUserByEmail(email);
+
+						if (!user) {
+							throw new APIError("USER_NOT_FOUND", {
```
Bekacru: change `USER_NOT_FOUND` to `BAD_REQUEST`

Issue #415, opened 2024-11-04
Email verification `redirectTo` is improperly composed 


svaraborut:
**Describe the bug**
The `redirectTo` field of `authClient.forgetPassword({ ... redirectTo: '/auth/login' })` has an inconsistent behavior
with `callbackURL` within other methods. It is a perfectly resolvable relative URL, but it is not resolved properly by the
API. To the email this URL is sent
`http://localhost:4444/api/auth/reset-password/...?callbackURL=/auth/login` which upon visit redirects
me to `http://localhost:4444/api/auth/auth/login?token=J7pJqT6tmM6vWzg6vq427` which is wrong.

**Expected behavior**
The URL shall be resolved properly

svaraborut:
I can solve this

Bekacru:
> I can solve this

go for it

Issue #414, opened 2024-11-04
OAuth errors redirect to the internal page


svaraborut:
**Describe the bug**
When OAuth for integrated providers fails, the application redirects to the internal error page `/api/auth/error?error` despite providing `callbackURL`.

**Expected behavior**
If `callbackURL` is provided redirect to the callback page.

**Additional context**
Despite not being explicitly mentioned it is expected that the authentication via integrated providers would have a consistent behavior with [Generic OAuth](https://www.better-auth.com/docs/plugins/generic-oauth#error-handling).

Bekacru:
The reason we're not using the callback URL for error is because it's primarily intended for a success URL where auth is required (e.g., `/dashboard`). But, I'm open to the ide to add `errorCallbackURL` or something similar, which would default to the `currentURL` if not provided. If neither is specified, it would fall back to `/api/auth/error`.

svaraborut:
Yes, this would be the ideal behavior, as the default error page is quite drastic and not customizable, breaking the app flow even when the user does something innocent like "Cancel" during the OAuth flow.

I originally reported this issue as in `v0.6.2` some errors where redirected to the fallback page but some (like "Cancel") were redirected to the `callbackURL`. Now in `v0.7.2.beta-5` I see all of them go straight to the fallback.

PR #417, opened 2024-11-04
Fix Issue 415: Correct forget password redirect

svaraborut:
[Fixed Issue 415](https://github.com/better-auth/better-auth/issues/415)

I've added some `redirectError` and `redirectCallback` which are candidates to be injected directly to the `better-call` context and used through all the routes to reduce issues. `redirectError` is also a candidate function for the  implementation of `errorCallbackURL` proposed in [414](https://github.com/better-auth/better-auth/issues/414).

Fixed also the erroneous behaviour of redirection to the embedded error page if `callbackURL` is present, but `token` is missing

### DIFF
--- a/packages/better-auth/src/api/routes/forget-password.ts
+++ b/packages/better-auth/src/api/routes/forget-password.ts
@@ -1,6 +1,31 @@
 import { z } from "zod";
 import { createAuthEndpoint } from "../call";
 import { APIError } from "better-call";
+import type { AuthContext } from "../../init";
+
+function redirectError(
+	ctx: AuthContext,
+	callbackURL: string | undefined,
+	query?: Record<string, string>,
+): string {
+	const url = callbackURL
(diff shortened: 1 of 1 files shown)


firstaxel:
Good evening I think this should apply to all routes so we can provide a custom error page instead of the default better Auth error page which is not going for people wanting to use the package in production, it will be perfect if we make a option to add a custom error pages to the better Auth instance on the server.

firstaxel:
There any errors provided can be handled currently with, cause if it is me, I need customization for all my pages and if we can get a custom error page it is will be perfect not just the forget password route to have a custom error page

svaraborut:
Currently all endpoints will redirect you to `callbackURL?error=... ` this is how you can distinguish the error state. We have a feature request that is under development to add `errorCallbackUrl` to provide a different redirect url in case of error.

PR #439, opened 2024-11-06
frmt and ux

Kinfe123:
- **fix: count fmt and ux**
- **fix: display start if success**
- **fix: lint**


Bekacru:
Screenshot 2024-11-07 at 1 52 47 AM

PR #440, opened 2024-11-06
refactor: Replace `consola` with `pino`

### LINKED ISSUE #124: [REFACTOR] Improve logging

tklein1801:
So far, the options for customizing in logging are very limited or only log-level and activation can be set. It should be discussed which aspects of the logger should be configurable by the user

closes #124


tklein1801:
#124

tklein1801:
@Bekacru just resolved the merge conflict. Do you mind running the CI again?

Bekacru:
Sure. Also please remove pkg-lock.json and the pnpm lock file too. And remove any new deps since we're not using them.

tklein1801:
Sure. Done.

tklein1801:
Sure. Should be it.

Bekacru:
My current concern is that this will remove the nice greenish success message in the CLI for migrate and generate. Maybe we could have the CLI use the consola logger directly instead, since it can't be configured. If you could make this change, this will be good to go!

tklein1801:
Sure, let me look at that. Currently, however, Consola should already be used because, as you wrote, an unconfigured logger is imported from better-auth and therefore the standard handler (Consola) should be used.

I will check the behavior again and make changes if necessary.

packages/cli/src/commands/generate.ts
```diff
@@ -124,7 +124,7 @@ export async function generateAction(opts: any) {
 			options.output || path.join(cwd, schema.fileName),
 			schema.code,
 		);
-		logger.success(`🚀 Schema was generated successfully!`);
+		logger.info(`🚀 Schema was generated successfully!`);
```
Bekacru: Here it should be `success` instead

packages/cli/src/commands/generate.ts
```diff
@@ -97,7 +97,7 @@ export async function generateAction(opts: any) {
 			} else {
 				await fs.appendFile(path.join(cwd, schema.fileName), schema.code);
 			}
-			logger.success(
```
Bekacru: same here

packages/cli/src/commands/generate.ts
```diff
@@ -152,7 +152,7 @@ export async function generateAction(opts: any) {
 		options.output || path.join(cwd, schema.fileName),
 		schema.code,
 	);
-	logger.success(`🚀 Schema was generated successfully!`);
+	logger.info(`🚀 Schema was generated successfully!`);
```
Bekacru: here

packages/cli/src/commands/migrate.ts
```diff
@@ -102,7 +102,7 @@ export async function migrateAction(opts: any) {
 	spinner?.start("migrating...");
 	await runMigrations();
 	spinner.stop();
-	logger.success("🚀 migration was completed successfully!");
+	logger.info("🚀 migration was completed successfully!");
```
Bekacru, line 105: here as well

packages/cli/src/utils/get-config.ts
```diff
@@ -140,7 +140,6 @@ export async function getConfig({
 							config.auth?.options || config.default?.options || null;
 						if (!configFile) {
 							logger.error("[#better-auth]: Couldn't read your auth config.");
-							logger.break();
```
Bekacru, line 144: you can change this to just empty `console.log("")` instead

Bekacru (changes requested):

tklein1801:
Sorry, I misunderstood you. How about the following solution?
We reintroduce the log level `success` for Consolas to make it possible to output successful messages, but if a custom logger is implemented, success is replaced by info as the log level, as success does not technically exist as a log level.

tklein1801:
I made a mistake with the log-Level the get-config.ts file inside the cli Package. 

Will fix later

tklein1801:
Done. Everything should be set now

Bekacru:
hey @tklein1801 please update the snapshot

tklein1801:
Hi @Bekacru, I've already done this. I am still looking at the other two failed tests

tklein1801:
**Test `oauth.test.ts` `oauth2 - should handle server error during OAuth flow`**

Runs into `cookie-utils.ts` and stopps right here

I think if the process cannot continue due to an error, an error should be thrown.
```ts
if (!name || value === undefined) {
  console.warn(`Malformed cookie: ${cookieString}`);
  return;
}
```

tklein1801:
Well, I'll be honest... It took me a while to work out where the error was. 
Build and test are succeeding in local...

Issue #451, opened 2024-11-07
`auth.api.signOut()` missing types


taylorfsteele:
**Describe the bug**
The `signOut` function via the api is missing types.

```
import { auth } from "~/lib/auth";

const someFunctionOnTheServer = async () => {
  const headers = getHeaders();

  await auth.api.signOut({ headers }) // No types
} 
```

**To Reproduce**

1. Create a [better-auth instance](https://www.better-auth.com/docs/installation#create-a-better-auth-instance)
2. Call `auth.api.signOut()`

**Expected behavior**
`api.signOut()` throws a type error without headers and provides types for the function

**Screenshots**
![image](https://github.com/user-attachments/assets/d36cbb4c-b283-49c1-aeaf-2dd7d221c491)

Other api functions types seem fine.

![image](https://github.com/user-attachments/assets/09a2a6c6-c399-4130-a801-7ee255668c68)

better-auth: 0.7.5-beta.5

Bekacru:
merged and released on `0.7.5-beta.6`

PR #453, opened 2024-11-07
fix: add `requireHeaders` to sign out api

taylorfsteele:
Fix for #451

From what I can tell, this fixes the types on `signOut`

Before:
image

After:
image

Let me know if there's anything else I can add 🙂


Bekacru:
looks good

Issue #292, opened 2024-10-22
module not found: Can't resolve 'crypto'


goodchengwang:
**Describe the bug**
use bearer plugin causes an error module not found: Can't resolve 'crypto'

**To Reproduce**
Using the bearer plugin causes an error
```ts
import { betterAuth } from 'better-auth'
import { drizzleAdapter } from 'better-auth/adapters/drizzle'
import { bearer } from 'better-auth/plugins'

import { db } from '@acme/db/client' // your drizzle instance

import { env } from '../env'

export const auth = betterAuth({
  database: drizzleAdapter(db, {
    provider: 'sqlite',
  }),
  socialProviders: {
    github: {
      clientId: env.GITHUB_CLIENT_ID,
      clientSecret: env.GITHUB_CLIENT_SECRET,
    },
  },
  plugins: [bearer()],
})
```

**Expected behavior**
no error

**Screenshots**
![image](https://github.com/user-attachments/assets/007d4c88-8267-4980-94e5-65807b8e2c23)
```js
Did you mean './crypto'?
Requests that should resolve in the current directory need to start with './'.
Requests that start with a name are treated as module requests and resolve within module directories (node_modules, /home/goodcheng/.volta/tools/shared, /home/goodcheng/Development/SelfCode/collectionbox/apps/web).
If changing the source code is not an option there is also a resolve options called 'preferRelative' which tries to resolve these kind of requests in the current directory too.

https://nextjs.org/docs/messages/module-not-found

Import trace for requested module:
../../packages/auth/src/auth.ts
../../packages/auth/src/index.ts
./src/app/api/auth/[...all]/route.ts
[...]

PR #454, opened 2024-11-07
Use explicit 'node:' prefix for crypto import

dpuscher:
To make better-auth more compatible with serverless environments like e.g. Cloudflare Pages, the import of the native crypto module needs to be prefixed with `node:`.


dpuscher:
This might be a fix for https://github.com/better-auth/better-auth/issues/292

Bekacru:
Hey thanks for the PR! but closing this in favor of #456 (node:) prefixed imports tends to break next js middleware

PR #498, opened 2024-11-11
refactor: minify

productdevbook:
This pull request includes a small change to the `packages/better-auth/package.json` file. The change involves modifying the build script to remove the `--minify` option.

The reason for removing the --minify option is to obtain a more readable and analyzable output during development and debugging. Minification optimizes and reduces the code size but decreases readability, which can make debugging more challenging. 

* [`packages/better-auth/package.json`](diffhunk://#diff-07bd5b61fca590f15772445019f72de76c25e488e3ed941e9860c8d2a51fdef5L12-R12): Removed the `--minify` option from the build script.

### DIFF
--- a/packages/better-auth/package.json
+++ b/packages/better-auth/package.json
@@ -9,7 +9,7 @@
     "directory": "packages/better-auth"
   },
   "scripts": {
-    "build": "cross-env NODE_OPTIONS=--max-old-space-size=8000 tsup --clean --dts --minify",
+    "build": "cross-env NODE_OPTIONS=--max-old-space-size=8000 tsup --clean --dts",
     "dev": "cross-env NODE_OPTIONS='--max-old-space-size=4000' tsup --watch --sourcemap",
     "dev:dts": "cross-env NODE_OPTIONS='--max-old-space-size=8192' tsup --watch --dts",
     "test": "pnpm prisma:push && vitest",


packages/better-auth/package.json
```diff
@@ -9,7 +9,7 @@
     "directory": "packages/better-auth"
   },
   "scripts": {
-    "build": "cross-env NODE_OPTIONS=--max-old-space-size=8000 tsup --clean --dts --minify",
```
emroot, line 12: How about keeping the original build command, and introducing a new comamnd "dev:build" instead?

productdevbook: It’s a significant issue when you encounter an error in a project, check the source code, and realize it’s fully minified. While compressing code is suitable for CDN usage, it’s unnecessary for an npm package. Modern frameworks already handle compression during the build process.

(I should be able to put console.log in the source code immediately.)

productdevbook:
@Bekacru 

every time I have to build and copy the source code in the project.

reslear:
@Bekacru I totally agree with @productdevbook. Ideally I would do away with tsup altogether. 

CleanShot 2024-12-31 at 03 20 19@2x

so we could share and test `.patch` using `pnpm patch` files directly in the project !

like it's done in 
https://github.com/wevm/wagmi/blob/54a91f186260b6062199e8255fc79ae002c77c93/packages/core/package.json#L13

It's just genius and simply generates the correct directories, source data is generated, not minified js. 


CleanShot 2024-12-31 at 03 22 24@2x

reslear:
we also have many same code: 

CleanShot 2024-12-31 at 03 52 04@2x

kzlar:
+1 for this. Most packages do not come minified, never quite understood why better-auth is different in that manner

reslear:
hi @Bekacru already merged https://github.com/better-auth/better-auth/pull/1195 thx!
@decentparadox close please.

PR #504, opened 2024-11-12
wip: Role Based Access Control

Bekacru:
This plugin provides users with a more flexible, agnostic approach to access control by introducing scoped roles and permissions. With this, roles can be defined based on a particular scope—whether it’s the entire system, specific organizations, teams, or any other grouping.


packages/better-auth/src/plugins/access-control/index.ts
```diff
@@ -0,0 +1,489 @@
...
+								value: user.id,
+							},
+						],
+					});
+					if (!userRoles.length) {
+						return {
+							permissions: [],
+						};
+					}
+					const rolePromises = userRoles.map(async (userRole) => {
```
emroot, line 141: Can prob run a single findMany query here and use `in: roleIds`

packages/better-auth/src/plugins/access-control/index.ts
```diff
@@ -0,0 +1,489 @@
...
+						const role = await ctx.context.adapter.findOne<StoredRole>({
+							model: "role",
+							where: [
+								{
+									field: "id",
+									value: userRole.roleId,
+								},
+								{
+									field: "scope",
+									value: opts.globalScope,
```
emroot, line 151: Prob wanna check for c.query.scope here too

PR #509, opened 2024-11-12
Update email-password.mdx

fres-sudo:
Since the `sendVerificationEmail` do not have any type `token` param in the actual code, I just quickly fixed the small mistake in the documentation that I noticed


Bekacru:
The API has been updated since then, so it should now be correct.

PR #491, opened 2024-11-11
fix: apple provider

Bekacru:
closes #490

PR #521, opened 2024-11-13
Allow POST request for apple on callback endpoint

emroot:
Looks like Apple needs `form_post`, here's a PR that updates the callback endpoint to support POST request in case we decide to go that route.


Bekacru:
merged to the initial pr (#491)

PR #535, opened 2024-11-14
feat: enhance custom id generation with type support

jstri:
Adds support for custom Stripe-style IDs in `generateId` function and moves it to `AuthContext`.

The function is relocated from adapters to `AuthContext` to:
- Provide a single, adapter-agnostic interface that works across all adapters (including memory)
- Enable the possibility for custom ID generation beyond models (e.g., verification tokens)


packages/better-auth/src/api/routes/forget-password.ts
```diff
@@ -178,7 +178,7 @@ export const resetPassword = createAuthEndpoint(
 				userId,
 				providerId: "credential",
 				password: hashedPassword,
-				accountId: ctx.context.uuid(),
+				accountId: userId,
```
jstri, line 184: Used `userId` instead because the following is stated in docs: `The id of the account as provided by the SSO or equal to userId for credential accounts`

docs/content/docs/concepts/database.mdx
```diff
@@ -523,44 +523,47 @@ res.user.lang // > "fr"
...
 export const auth = betterAuth({
-   database: prismaAdapter(prisma, {
-      provider: "sqlite",
-      generateId: () => {
-         // custom id generation logic
-      }
-   })
+   database: {
+      db: db,
+      useAutoGeneratedId: true
```
emroot: why add a new option when you can simply use `generateId => false`, which does the same. or can also be achieved with `generateId() => undefined`. This introduce 2 ways of achieving the same thing.

Bekacru:
merged with #595

PR #548, opened 2024-11-15
feat: update passkey

### LINKED ISSUE #526: Feature: Update Passkey
An update passkey endpoint that allows the user to update passkey data (in the current schema i guess only the name field). Allowing the user to better manage his passkeys.

diegopluna:
Closes #526.

### DIFF
--- a/packages/better-auth/src/plugins/passkey/client.ts
+++ b/packages/better-auth/src/plugins/passkey/client.ts
@@ -153,6 +153,32 @@ export const getPasskeyActions = (
 		}
 	};
 
+	const updatePasskey = async (
+		opts: {
+			fetchOptions?: BetterFetchOption;
+			id: string;
+			name: string;
+		},
+		fetchOpts?: BetterFetchOption,
+	) => {
+		const updated = await $fetch<{
+			passkey: Passkey;
+		}>("/passkey/update-passkey", {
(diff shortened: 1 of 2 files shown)


diegopluna:
Originally I was gonna try to name it renamePasskey but I don't know if there are any plans to expand the current schema so updatePasskey seems more appropriate.

Tried writing test  for it but apparently the current test for the passkeys plugin does not create a passkey so I will probably give a look into that.

packages/better-auth/src/plugins/passkey/client.ts
```diff
@@ -153,6 +153,32 @@ export const getPasskeyActions = (
 		}
 	};
 
+	const updatePasskey = async (
+		opts: {
+			fetchOptions?: BetterFetchOption;
+			id: string;
+			name: string;
+		},
+		fetchOpts?: BetterFetchOption,
```
emroot, line 162: I don't understand why fetchOpts is passed twice.

diegopluna: Honestly, neither do I :sweat_smile: . I was just following along the structure of the previous methods in the code. And they were passing it twice just like this. Maybe that is something that should be looked at on all the current methods.

Issue #171, opened 2024-10-14
feature-request: Add Nest js support


danyalutsevich:
This is feature request to add nest js support

NoHaxito:
@Bekacru i have an idea on how to "implement" without modifying the current code (only docs)
Is it okay if i do a PR with a 
 "temporary guide"? (If you plan to do a "custom" integration as is the case with Next.js, Svelte, etc.)

The only thing is to create a controller and module and get the `req` and `res` parameters from the decorators.

Bekacru:
> is it okay if i do a PR with a "temporary guide"? (If you plan to do a "custom" integration as is the case with Next.js, Svelte, etc.)

yes feel free to open a pr. and we're not going to have integration we only need the docs.

NoHaxito:
> > is it okay if i do a PR with a "temporary guide"? (If you plan to do a "custom" integration as is the case with Next.js, Svelte, etc.)
> 
> yes feel free to open a pr. and we're not going to have integration we only need the docs.
> 

Ok! I will try to have the nestjs docs soon 🙌🏼🚀

ismoiliy98:
Any progress on this?

danyalutsevich:
@ismoiliy98  join us in discord
we have couple of problems with module system in nest and better-auth
feels like we are about to finish)

BayBreezy:
I tried adding better auth to nest but the ESM nature of better auth is nor playing nicey with NestJS. @danyalutsevich which discord do i need to join to see the discusiion?

danyalutsevich:
@BayBreezy you can check the [repo](https://github.com/Snazzy72/better-auth-nestjs) from this #359 issue 
you can use dynamic import as a workaround

discord: https://discord.com/invite/GYC3W7tZzb 
if it tells you that invite is expired copy the link and paste it to join server window
<img width="440" alt="Screenshot 2024-11-02 at 15 09 36" src="https://github.com/user-attachments/assets/394ebf35-782a-4733-9f53-aacef27fdad4">

danyalutsevich:
I dont know if it's possible to merge issues on github
we have duplicate issues

#406 #359

subenksaha:
I think the problem is from nanoid module. I am getting the following error:
![Screenshot 2024-11-13 at 9 57 55 PM](https://github.com/user-attachments/assets/b829abe4-f664-442a-b19e-34637e8b319d)

[...]

PR #550, opened 2024-11-15
#171 replaced nanoid with @paralleldrive/cuid2 for better commonjs and esm…

subenksaha:
Replaced nanoid with @paralleldrive/cuid2 based on the following error message:

Screenshot 2024-11-13 at 9 57 55 PM

nano id don't have support for both esm and commonjs as per thier website: 

https://github.com/ai/nanoid#:~:text=Nano%20ID%205%20works%20only%20with%20ESM%20projects%2C%20in%20tests%20or%20Node.js%20scripts.%20For%20CommonJS%20you%20need%20to%20use%20latest%20Node.js%2020%20or%2022%20with%20%2D%2Dexperimental%2Drequire%2Dmodule%3A


marcomuser:
Why not crypto.randomUUID() instead?

subenksaha:
> Why not crypto.randomUUID() instead?

crypto.randomUUID() is not url friendly, also not lexicographically sortable.

nanoid was url friendly, that's why I thought cuid2 can be better replacement of nanoid.

Also, as the keys are stored as string in database, uuid will take more space than nanoid/cuid2, cuid2 is also secure and collision resistant.

PR #551, opened 2024-11-15
fix logout

emroot:
make sure to pass all the options when deleting cookies, especially domain for crossDomainCookies, otherwise the cookies don't get deleted


Bekacru:
Hey, could you please run `pnpm lint:fix`

emroot:
done.

PR #558, opened 2024-11-16
Error when trying to do a migration with a custom user modelName and the 2fa plugin

Limerio:


Limerio:
Link to this issue #568

PR #562, opened 2024-11-17
Reset password expiration failed and date converted to timestamp to add precision

Limerio:
Initially, I was investigating why the format of expiresAt is a date (format: `YYYY-MM-DD`). What seemed unusual was that while the expiration time is set to one hour later, the expiresAt field doesn't include any time information. To address this, I attempted to convert it to a timestamp (which I think, it's not the best approach, as it converts all dates to timestamps – feel free to contribute and improve this solution). However, I encountered another issue: when you call `new Date()`, it returns a date in your machine's local timezone rather than UTC. Therefore, I needed to convert it to UTC to ensure that both the timestamp and UTC time were in the same timezone.

### DIFF
--- a/packages/better-auth/src/api/routes/forget-password.ts
+++ b/packages/better-auth/src/api/routes/forget-password.ts
@@ -1,8 +1,8 @@
-import { z } from "zod";
-import { createAuthEndpoint } from "../call";
 import { APIError } from "better-call";
+import { z } from "zod";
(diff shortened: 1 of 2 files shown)


Limerio:
Link to this issue #567

Bekacru:
Hey thanks for the PR and opening the issue. Closing this in favor of #572

PR #565, opened 2024-11-17
feat: custom roles

### LINKED ISSUE #40: Organization support for multiple roles and policies
Hi,

Having three sets of roles for the organization plugins (owner, admin, member) is helpful in many use cases, but integrating better-auth with complex APIs could require having different roles and sets of permissions for each role.

In my projects, I have at least a couple of examples where the domain I worked on required at least 4-5 different "sets of permissions" grouped in roles.

There are many Authentication libraries out there, either SaaS or open source. This one has me hyped the most because it offers tons of ready-made and simple-to-use options.
The only issue that I've found mostly ignored by all the authentication libraries is authorization. Some of them make compromises, and others simply don't give it to you.

Better-auth seems to be on the right path of being the "all in one" Auth $\^2$  library, so I suggest trying to implement the best authorization library I believe there could be.

Having roles in code is pretty straightforward and an excellent option, and as far as I can understand from the docs, the permissions are a CRUD "resource-based" system. It would also be nice to have permissions at the attribute (read column) level, other than the resource (read row) level, but that can be relatively more complex than the library's scope. And will probably fall into not just "authentication," but more like the "authentication."

So, as of now, I would like, if possible, to suggest two edits that are very interconnected with each other:

1. Change the way the roles are defined to be customizable by the user
2. Add some examples of using the custom permissions system after the setup.

For "2." Maybe I've lost some concepts reading the docs, but I don't understand how, after adding a new "project" resource permission, I can link it to an actual API endpoint or backend endpoint, etc...

Also, for the future, getting back to the authentication part, it would be cool to look at [the Directus policy-based access control](https://docs.directus.io/blog/understanding-policy-based-access-control.html) system and how it is integrated with roles.

Bekacru:
closes #40

This is the first attempt to introduce better access control to Better Auth. It will allow within the org plugin to define custom roles and permissions. More advanced access control plugins with different methods will be introduced after v1.


packages/better-auth/src/plugins/organization/access/access.ts
```diff
@@ -0,0 +1,73 @@
...
+						success: false,
+						error: `You are not allowed to access resource: ${requestedResource}`,
+					};
+				}
+				const success =
+					connector === "OR"
+						? (requestedActions as string[]).some((requestedAction) =>
+								allowedActions.includes(requestedAction),
+							)
+						: (requestedActions as string[]).every((requestedAction) =>
```
Limerio, line 48:
```suggestion
			for (const [requestedResource, requestedActions] of Object.entries<string[]>(
				request,
			)) {
				const allowedActions = statements[requestedResource];
				if (!allowedActions) {
					return {
						success: false,
						error: `You are not allowed to access resource: ${requestedResource}`,
					};
				}
				const success =
					connector === "OR"
						? requestedActions.some((requestedAction) =>
								allowedActions.includes(requestedAction),
							)
						: requestedActions.every((requestedAction) =>
```

packages/better-auth/src/plugins/organization/access/access.ts
```diff
@@ -0,0 +1,73 @@
...
+
+type AuthorizeResponse =
+	| { success: false; error: string }
+	| { success: true; error?: undefined };
+
+function parsingError(message: string, path: string): Error {
+	const error = new Error(message);
+	(error as any).path = path;
+	return error;
+}
```
Limerio, line 13: This function isn't use

packages/better-auth/src/plugins/organization/client.ts
```diff
@@ -31,9 +37,28 @@ export const organizationClient = <O extends OrganizationClientOptions>(
...
+				ac: O["ac"] extends AccessControl
+					? O["ac"]
+					: AccessControl<DefaultStatements>;
+				roles: O["roles"] extends Record<string, Role<any>>
+					? O["roles"]
+					: {
+							admin: Role<any>;
+							member: Role<any>;
+							owner: Role<any>;
+						};
```
Limerio:
```suggestion
				roles: O["roles"] extends Record<string, Role>
					? O["roles"]
					: {
							admin: Role;
							member: Role;
							owner: Role;
						};
```

packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -67,27 +68,28 @@ export interface OrganizationOptions {
...
+	 * Configure the roles and permissions for the
+	 * organization plugin.
 	 */
 	ac?: AccessControl;
 	/**
 	 * Custom permissions for roles.
 	 */
 	roles?: {
-		[key in "admin" | "member" | "owner"]?: Role<any>;
+		[key in string]?: Role<any>;
```
Limerio, line 92:
```suggestion
		[key in string]?: Role;
```

Limerio: And I think it's better to replace by a `Record<string, Role>` because a developer can do
```
roles: {
  aRole: undefined
}
```
Which I don't think you want that to be the case (otherwise ignore my comment 😅)

packages/better-auth/src/plugins/organization/routes/crud-members.ts
```diff
@@ -95,87 +96,88 @@ export const removeMember = createAuthEndpoint(
...
 );
 
-export const updateMemberRole = createAuthEndpoint(
-	"/organization/update-member-role",
-	{
-		method: "POST",
-		body: z.object({
-			role: z.enum(["admin", "member", "owner"]),
-			memberId: z.string(),
+export const updateMemberRole = <O extends OrganizationOptions>(option: O) =>
```
Limerio, line 99:
```suggestion
export const updateMemberRole = <O extends OrganizationOptions | undefined>(option: O) =>
```

packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -201,13 +210,13 @@ export const organization = <O extends OrganizationOptions>(options?: O) => {
 		setActiveOrganization,
 		getFullOrganization,
 		listOrganization,
-		createInvitation,
+		createInvitation: createInvitation(options as O),
```
Limerio, line 213:
```suggestion
		createInvitation: createInvitation(options),
```

packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -201,13 +210,13 @@ export const organization = <O extends OrganizationOptions>(options?: O) => {
...
 		cancelInvitation,
 		acceptInvitation,
 		getInvitation,
 		rejectInvitation,
 		removeMember,
-		updateMemberRole,
+		updateMemberRole: updateMemberRole(options as O),
```
Limerio, line 219:
```suggestion
		updateMemberRole: updateMemberRole(options),
```
Link to : https://github.com/better-auth/better-auth/pull/565/files#r1845484912

packages/better-auth/src/plugins/organization/routes/crud-invites.ts
```diff
@@ -4,115 +4,119 @@ import { getSessionFromCtx } from "../../../api/routes";
...
-				role: invitation.role,
-				email: invitation.email,
-				organization: organization,
-				inviter: {
-					...member,
-					user: session.user,
+			await ctx.context.orgOptions.sendInvitationEmail?.(
+				{
+					id: invitation.id,
+					role: invitation.role as string,
```
Limerio, line 106:
```suggestion
					role: invitation.role,
```
`invitation.role` is already define as a string

packages/better-auth/src/plugins/organization/routes/crud-invites.ts
```diff
@@ -4,115 +4,119 @@ import { getSessionFromCtx } from "../../../api/routes";
...
+					email: invitation.email,
+					organization: organization,
```
Limerio, line 108:
```suggestion
					organization,
```

packages/better-auth/src/plugins/organization/routes/crud-invites.ts
```diff
@@ -4,115 +4,119 @@ import { getSessionFromCtx } from "../../../api/routes";
...
-			invitation: {
-				role: ctx.body.role,
+			if (alreadyMember) {
+				throw new APIError("BAD_REQUEST", {
+					message: "User is already a member of this organization",
+				});
+			}
+			const alreadyInvited = await adapter.findPendingInvitation({
 				email: ctx.body.email,
 				organizationId: organizationId,
```
Limerio, line 79:
```suggestion
				organizationId,
```

packages/better-auth/src/plugins/organization/routes/crud-invites.ts
```diff
@@ -4,115 +4,119 @@ import { getSessionFromCtx } from "../../../api/routes";
...
-			throw new APIError("BAD_REQUEST", {
-				message: "User is already invited to this organization",
+			if (canInvite.error) {
+				throw new APIError("FORBIDDEN", {
+					message: "You are not allowed to invite members",
+				});
+			}
+			const alreadyMember = await adapter.findMemberByEmail({
+				email: ctx.body.email,
+				organizationId: organizationId,
```
Limerio, line 70:
```suggestion
				organizationId,
```

packages/better-auth/src/plugins/organization/routes/crud-invites.ts
```diff
@@ -4,115 +4,119 @@ import { getSessionFromCtx } from "../../../api/routes";
...
+				ctx.body.organizationId || session.session.activeOrganizationId;
+			if (!organizationId) {
+				throw new APIError("BAD_REQUEST", {
+					message: "Organization not found",
+				});
+			}
+			const adapter = getOrgAdapter(ctx.context, ctx.context.orgOptions);
+			const member = await adapter.findMemberByOrgId({
+				userId: session.user.id,
+				organizationId: organizationId,
```
Limerio, line 47:
```suggestion
				organizationId,
```

packages/better-auth/src/plugins/organization/routes/crud-members.ts
```diff
@@ -95,87 +96,88 @@ export const removeMember = createAuthEndpoint(
...
+					status: 400,
+					body: {
+						message: "No active organization found!",
+					},
+				});
+			}
+			const adapter = getOrgAdapter(ctx.context, ctx.context.orgOptions);
+			const member = await adapter.findMemberByOrgId({
+				userId: session.user.id,
+				organizationId: organizationId,
```
Limerio, line 129:
```suggestion
				organizationId,
```

PR #594, opened 2024-11-19
docs: fix example use of `useSession` to use synchronous instead

lamualfa:
`useSession` hook must be used as a sync function.

https://www.better-auth.com/docs/concepts/session-management#use-session


ping-maxwell:
Maybe it's clearer to show it like this instead too:
```ts
const { data: session, error } = authClient.useSession();
```

lamualfa:
> Maybe it's clearer to show it like this instead too:
> 
> ```ts
> const { data: session, error } = authClient.useSession();
> ```

Done.

PR #626, opened 2024-11-23
[expo] make the way cookies are handled more resilent

AlfayezDev:
I had a bug where if i use idToken on expo it either crashes the app altogether or makes the redirection very awkward with the browser behavior, i wanted to disable the redirect flag, but this ends up not storing the cookie, i noticed that when the request is successful, `set-cookie` header does return in the response, which is the same header returned with the `Browser`, felt like a good and simple solution to this problem.

### DIFF
--- a/packages/expo/src/client.ts
+++ b/packages/expo/src/client.ts
@@ -159,20 +159,30 @@ export const expoClient = (opts?: ExpoClientOptions) => {
 							storage.setItem(localCacheName, JSON.stringify(data));
 						}
 
-						if (
-							context.data.redirect &&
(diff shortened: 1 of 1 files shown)


pakenfit:
Hey @AlfayezDev I made this PR  #636  I had the same issue I update the server side plugin to return the cookie to the client with no redirect

AlfayezDev:
> Hey @AlfayezDev I made this PR #636 I had the same issue I update the server side plugin to return the cookie to the client with no redirect

sounds great, i'll close this one then

PR #627, opened 2024-11-23
Update svetekit integration docs 

nalonix:
here's the update to the docs.

### DIFF
--- a/docs/content/docs/integrations/svelte-kit.mdx
+++ b/docs/content/docs/integrations/svelte-kit.mdx
@@ -1,24 +1,13 @@
 ---
-title: SvelteKit Integration
-description: Integrate Better Auth with SvelteKit.
+
+title: Svelte Kit Integration
+description: Learn how to integrate Better Auth with Svelte Kit
+
 ---
 
-Before you start, make sure you have a Better Auth instance configured. If you haven't done that yet, check out the [installation](/docs/installation).
+Before you start, make sure you have a Better Auth instance configured. If you haven't done that yet, check out the [installation](https://www.notion.so/docs/installation).
 
-### Mount the handler
-
-We need to mount the handler to SvelteKit server hook.
-
-```ts title="hooks.server.ts"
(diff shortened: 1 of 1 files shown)


docs/content/docs/integrations/svelte-kit.mdx
```diff
@@ -33,61 +22,159 @@ export const authClient = createAuthClient({
...
-            await authClient.signIn.social({
-              provider: "github",
-            });
-          }}
-        >
-          Continue with github
-        </button>
-      {/if}
+      <p>{$session?.data?.user.name}</p>
+      <button on:click={async () => await authClient.signOut()}>
```
Devr-pro, line 37: should be onclick={} instead of on:click={}

docs/content/docs/integrations/svelte-kit.mdx
```diff
@@ -33,61 +22,159 @@ export const authClient = createAuthClient({
...
+        Sign Out
+      </button>
     </div>
+  {:else}
+    <button
+      on:click={async () => {
```
Devr-pro, line 43: on:click -> onclick

Devr-pro (changes requested):

docs/content/docs/integrations/svelte-kit.mdx
```diff
@@ -33,61 +22,159 @@ export const authClient = createAuthClient({
...
+
+If you're using TypeScript, update `app.d.ts` to define the `session` and `user` types:
+
+```ts title="app.d.ts"
+import type { User, Session } from "better-auth/types";
+
+declare global {
+    namespace App {
+        interface Locals {
+            session: Session | undefined;
```
kaysef, line 87: why not 
```ts
session: import("$lib/auth").$Infer.Session['session'] | undefined | null;
```
Based on [this section of the docs](https://www.better-auth.com/docs/concepts/typescript#inferring-types). It also ensures every configurations done to the `auth` instance is captured in the type

PR #636, opened 2024-11-23
feat: improve social auth with idToken on expo

pakenfit:
This PR enhances the `google` and `apple` social auth on `expo` using for example `expo-apple-authentication` and `@react-native-google-signin/google-signin`

It consists on:

- After gettting the `idToken` from `expo-apple-authentication` or `@react-native-google-signin/google-signin`
- Use social auth with idToken in body :

 ````
    provider: "google",
    callbackURL: "/home-screen",
    idToken: {
        token: response.data.idToken,
     },
});
````
- The server returns the session data, user and a boolean `idToken: true` indicates the auth with `idToken`
- The expo plugin server side intercepts the response and add cookie to the response body.
- The expo plugin client side store the cookie in the secure storage.


Bekacru:
I'm not sure what this is meant to solve. By default, the client stores the cookie if there's a `set-cookie` header in the response, which should be the case for `idToken`.

pakenfit:
@Bekacru When using idToken for auth the server return : 

````
return c.json({
        session: data.data!.session,
        user: data.data!.user,
        url: `${
          c.body.callbackURL || c.query?.currentURL || c.context.options.baseURL
        }`,
        redirect: true,
      });
````
The expo client plugin try to use expo web browser to auth : 

````
if (
              context.data.redirect &&
              context.request.url.toString().includes("/sign-in")
            ) {
              const callbackURL = JSON.parse(context.request.body)?.callbackURL;
              const to = callbackURL;
              const signInURL = context.data?.url;
              const result = await Browser.openAuthSessionAsync(signInURL, to);
              if (result.type !== "success") return;
              const url = new URL(result.url);
              const cookie = String(url.searchParams.get("cookie"));
              if (!cookie) return;
              storage.setItem(cookieName, getSetCookie(cookie));
              store?.notify("$sessionSignal");
            }
````
it crashes the app as mentionned in #626 

My PR avoid the redirection and just store the `cookie` returned by the server in client side

luksch42:
@Bekacru Thats my problem we're discussing on discord

Bekacru:
> @Bekacru Thats my problem we're discussing on discord

Then, we could simply avoid it by checking if the original request contains `idToken` and not redirecting.

luksch42:
> > @Bekacru Thats my problem we're discussing on discord
> 
> Then, we could simply avoid it by checking if the original request contains `idToken` and not redirecting.

Yeah its working if the browser session is cancelled which just returns @pakenfit

Bekacru:
It won't redirect on idtoken auth starting from `v1.0.9-beta.5`. feel free to open an issue or a pr if the issue is still unresolved.

PR #638, opened 2024-11-24
Add isUnique to table component

sajadmh:


sajadmh (commented):
Improved component structure to make it easier to add more field types like isUnique later on

PR #650, opened 2024-11-24
fix: id is not a property in idToken payload when accountLinked is enabled

fabianmerino:
It is possible that the idToken does not contain the required id or data, so it is better to unify and destructure the information; this resolves the error when accountLinking is enabled


Bekacru:
yeah this make sense. you could still pass `getUserInfo` instead but this is a better default.

PR #666, opened 2024-11-26
Fix "email not found" Error in Twitter OAuth

### LINKED ISSUE #591: twitter (now X) api v2 doesn't return user's email
**Describe the bug**
we can't use twitter social login, because of email_not_found error.

https://github.com/better-auth/better-auth/blob/f7287b02a1e739ec3739a6d82053070e430c4245/packages/better-auth/src/social-providers/twitter.ts

here is their api document
https://developer.x.com/en/docs/x-api/users/lookup/api-reference/get-users-me#tab1

**To Reproduce**

0. setup twitter oauth2 client_id and client_secret (see screenshot)
1. Go to 'https://www.coaching-sospeso.org/auth/login'
2. Click on '트위터로 계속하기' button (sky blue one with twitter bird icon)
3. authorize our application
4. redirect to https://www.coaching-sospeso.org/auth/login?error=email_not_found

and also i can't login with x(twitter) in [better-auth demo app](https://demo.better-auth.com/sign-in)

**Expected behavior**
logged in well

**Screenshots**
here is our configuration. we 'request email from users'
image

**Desktop (please complete the following information):**

- OS: macos 15.1, window10, ubuntu linux
- Browser chrome
- Version 130.0

**Smartphone (please complete the following information):**

- Device: Android 
- OS: android 14
- Browser chrome
- Version 130.0

**Additional context**

I think we can't login with twitter v2 api when we need email as required.
https://devcommunity.x.com/t/how-to-get-email-in-twitter-v2-api/224225

AkeboshiHimari:
## Why is this change needed?
- The Twitter v2 API does not return the `email` field.
- However, the `getUserInfo` function in `/social-providers/twitter` expects an `email` value.
- This causes a sign-in failure with the error: `email_not_found` when using Twitter OAuth.
## Changes Made
- Replaced the usage of the `email` field with the `username` value returned by the Twitter API.
- Adjusted the `getUserInfo` function to handle this change.
## Change Impact
- Twitter OAuth now functions correctly without triggering the `email_not_found` error.
## Related Issues
- Fixes #591

### DIFF
--- a/docs/content/docs/authentication/twitter.mdx
+++ b/docs/content/docs/authentication/twitter.mdx
@@ -9,6 +9,12 @@ description: Twitter provider setup and usage.
         Get your Twitter credentials from the [Twitter Developer Portal](https://developer.twitter.com/en/portal/dashboard).
 
         Make sure to set the redirect URL to `http://localhost:3000/api/auth/callback/twitter` for local development. For production, you should set it to the URL of your application. If you change the base path of the auth routes, you should update the redirect URL accordingly.
+        
+        {/*
+        <Callout type="warn">
+            If twitter doesn't return the email address, the authentication won't be successful. Make sure to ask for the email address when you create the Twitter app.
+        </Callout>
+        */}
         <Callout type="info">
             Twitter API v2 does not provide email addresses. As a workaround, the user's `email` field uses the `username` value instead.
         </Callout>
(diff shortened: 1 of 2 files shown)


Bekacru:
Hey thanks for the PR! 

username isn't really an email, which is my hesitation. But, until we find a better solution, this is ideal enough. Just mention it in the docs somewhere.

lagrange-lab:
@Bekacru Thanks for the feedback!

I’ve updated the docs:
- ddaf86dd1aeb2a8bad9840fa40a6f17af44d106d Added a callout explaining that the `username` is used in place of the `email` field, since Twitter API v2 doesn't provide the email address. 
- eea7da3e0f92210a25acfe9974bc5cc64dfbc52c Commented out the previous note about email being required for authentication and the need to request an email when creating the Twitter app.

Bekacru:
thanks!

PR #669, opened 2024-11-26
fix: ensure persistence to secondary storage upon storage eviction

jstri:
In some cases (e.g., using Cloudflare Tiered Cache) secondary storage can be evicted during a session's lifetime, especially for long-lived sessions. After eviction, reads either return `null` or fallback to the database if the `storeSessionInDatabase` setting is enabled. This PR ensures that sessions are re-persisted to secondary storage under the following conditions:
- `secondaryStorage` is configured
- session is not found in secondary storage
- session exists in database (`storeSessionInDatabase` setting is enabled)

This ensures that sessions can be retrieved from secondary storage on subsequent reads or if the session needs to be updated (due to extended expiration time), it can be saved in secondary storage as well.

### DIFF
--- a/packages/better-auth/src/db/internal-adapter.ts
+++ b/packages/better-auth/src/db/internal-adapter.ts
@@ -27,6 +27,24 @@ export const createInternalAdapter = (
 	const sessionExpiration = options.session?.expiresIn || 60 * 60 * 24 * 7; // 7 days
 	const { createWithHooks, updateWithHooks, updateManyWithHooks } =
 		getWithHooks(adapter, ctx);
+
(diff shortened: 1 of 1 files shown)


Bekacru:
Hey, thanks for the PR! yeah make sense

Roronoa-Zoroo:
This breaks when using secondary storage
Particularly this
`data.session.expiresAt
				? Math.floor((data.session.expiresAt.getTime() - Date.now()) / 1000)
				: sessionExpiration,`
Error: session.expiresAt.getTime is not a function.

PR #672, opened 2024-11-27
fix: mysql migrations with drizzle

### LINKED ISSUE #26: BLOB/TEXT column 'id' used in key specification without a key length

gltjk:
fix #26 bug in drizzle migrations using similar solution to #27


Bekacru:
I think we only need to apply it for `id` and `reference` fields. Other fields can remain `text`.

gltjk:
> I think we only need to apply it for `id` and `reference` fields. Other fields can remain `text`.

Thanks for the reply! I see what you mean but I believe there’s a bit more to it. In MySQL, text fields with unique indexes (like `email` / `username` / `phoneNumber` in `user` table and `token` in `session` table) also throw errors saying "BLOB/TEXT column 'XXX' used in key specification without a key length".

And for those fields without such constraints (like `name` / `ipAddress` / `userAgent` / `accessToken` / `refreshToken`), I think `VARCHAR` is still a better fit. It’s faster to read and sort because it’s stored with the row, while `TEXT` is stored separately, which slows things down. Honestly, I’d only use `TEXT` for something like storing full articles or big chunks of text. For most other fields, `VARCHAR` is simpler, faster, and avoids running into weird MySQL quirks later.

As for the default length for `VARCHAR`, I used 255 in the PR, which is also your choice in #27 . In case that's not long enough, we may also increase it to up to 65535 (already equal to what `TEXT` can hold at most). If that's still not enough, we can then turn to `MEDIUMTEXT` (16MB) or `LONGTEXT` (4GB), but I doubt if any predefined fields in `better-auth` plugins would require that much.

Bekacru:
One field that needs more than 255 characters is the `value` field in the `verification` table, as it can store a stringified JSON value, like an OAuth state, which often exceeds 255 characters. Currently, we don't utilize JSON fields; instead, we use strings to maintain consistency across all dbs. This will limit us not just to use`varchar`. I know we could increase the limit, but at that point, using `text` would be a better option. You're also correct that `unique` values require `varchar` since they need to be indexed. If you look at the Kysely implementation, you'll notice we selectively apply `varchar`.

https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/db/get-migration.ts#L171

gltjk:
> One field that needs more than 255 characters is the `value` field in the `verification` table, as it can store a stringified JSON value, like an OAuth state, which often exceeds 255 characters. Currently, we don't utilize JSON fields; instead, we use strings to maintain consistency across all dbs. This will limit us not just to use`varchar`. I know we could increase the limit, but at that point, using `text` would be a better option. You're also correct that `unique` values require `varchar` since they need to be indexed. If you look at the Kysely implementation, you'll notice we selectively apply `varchar`.
> 
> https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/db/get-migration.ts#L171

Thanks for the feedback! I just found the original issue (#26 ) and the solution to it (#27 ), which is why I submitted my PR. I didn’t check the current code thoroughly—that’s on me. I’ve now updated my PR to follow the current Kysely implementation.

Bekacru:
Hey thanks for taking the time and fixing the PR :)) appreciate it!

PR #686, opened 2024-11-28
feat(builder): enhance UI and accessibility for auth options

ylyra:
Added htmlFor in builder options label to be able to change switch state by clicking on the label


Bekacru:
thanks for the PR! yeah why not ;)

PR #693, opened 2024-11-29
fix: save refreshToken in generic-oauth plugin

ramenha0141:
Fixed an issue in the Generic OAuth plugin where refreshToken, idToken, and accessTokenExpiresAt were not being saved.


ramenha0141:
Imports may have been sorted by mistake.

Bekacru:
thanks!

PR #706, opened 2024-12-01
Update import for generated drizzle schema file

pureliumy:
There's a missing import of `varchar` for postgres when generating drizzle schema.


Bekacru:
Hey thanks for this PR! Closing it in favor of #719 as we don't need to use `VARCHAR` in pg in the first place.

PR #710, opened 2024-12-01
feat: add support for configurable digits length

### LINKED ISSUE #653: Configure digits length on OTP
TOTP has options for configuring the number of digits expected.  It looks like OTP is built on top of the TOTP implementation, but does not currently handle that configuration option.  Can this be exposed?  Do you want a PR?

Kaseax:
This pull request closes #653 . It adds the option, to make the digit length of otp configurable.


Bekacru:
moved otp to use random string instead of relying on totp (since it's probably not expected behavior)

Issue #722, opened 2024-12-02
Session Token cookie maxAge not being updated when session updateAge


Kabeer10:
**Describe the bug**
The `maxAge` of the session token cookie is not being updated when `updateAge` is set in the betterAuth configuration.

**To Reproduce**

1. Set up betterAuth with the following configuration:
```
import { betterAuth } from "better-auth";

export const auth = betterAuth({
    session: {
        expiresIn: 60 * 2, // 2 Min
        updateAge: 60 // 1min
    }
});
```
3. Log in to create a session.
```
const { data, error } = await authClient.signIn.email({
  email: "test@example.com",
  password: "password1234",
});
```
4. Check the session token cookie's maxAge after 1 Min of activity.
5. Observe that the maxAge is not updated as expected. **But session in DB is getting updated**

**Expected behavior**
The `maxAge` of the session token cookie should be updated every 1 Min (as per updateAge) to extend the session expiration.

**Desktop (please complete the following information):**

- OS: MacOS 15
- Browser: Brave Browser
- Better-Auth Version : 1.0.7

**Additional context**
The cookie gets deleted after 2Min and user gets logged out. The Expiry in DB is however updated after 1 Min.

Bekacru:
My guess is this is happening because you're trying to call `getSession` on a server component (if you're using next) or somewhere it can't update the cookie. Are you, by any chance, calling `getSession` on a server?

Kabeer10:
im Using Astro. im calling getSession on client

```
//@lib/auth-client.ts

import { createAuthClient } from "better-auth/client"

export const authClient = createAuthClient({
  baseURL: env.API_URL + "/auth",
  plugins: [usernameClient()],
});
```

im calling getSession with help of useQuery

```
export default function Navbar() {
  const session = useQuery(
    {
      queryKey: ["user-session"],
      queryFn: async () => await authClient.getSession(),
    },
    queryClient
  );
  console.log(session.data);
  return <nav className="bg-background">Navbar</nav>;
}
```

When i inspect in network tab. The response session expiry is updated. but i dont see any set-cookie header.

Is there any other way cookie is being updated?

Bekacru:
Okay, this should update it. But, the update works in your case only if `getSession` is called after 1 minute has passed and before 2 minutes have elapsed.

Kabeer10:
**UPDATE**

i get set-cookie header when 
` updateAge: 0`

i tried setting
` updateAge: 30` // 30 sec

[...]

PR #736, opened 2024-12-03
fixes cookie MaxAge not being updated when calling getSession

Kabeer10:
Not fetching session inside `setSessionCookie` seems to work. Please verify it doesn't break other things. 

Closes issue #722

### DIFF
--- a/packages/better-auth/src/cookies/index.ts
+++ b/packages/better-auth/src/cookies/index.ts
@@ -107,21 +107,21 @@ export async function setSessionCookie(
 		? undefined
 		: ctx.context.sessionConfig.expiresIn;
 
-	const hasAlreadySession = await getSessionFromCtx(ctx).catch(() => {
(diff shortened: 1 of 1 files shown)


Bekacru:
hey could you please confirm me if this still happens on `1.0.9-beta.6`?

Kabeer10:
It still seems to happen in `"better-auth": "1.0.9-beta.6"`

Bekacru:
Alright, I went with your solution. The existing session token check isn’t required. We’ll find a better way later. This is more pressing. It should be working on `1.0.9-beta.7`

Kabeer10:
The fix you made using `disableRefresh` for `getSession` seems to be a nice way to handle this. As this will prevent session update when checking for session in server (if there is cookie involved). The server will have to pass on the cookie to browser then.

Will need to implement in astro middleware also. i think

PR #802, opened 2024-12-07
Docs/concept/plugin refactor 

Multinite:
- Made sure the reader knows the ID can be the same for both server & client plugins.
- Added demo code to show the creation of both the server & the client plugins.
- Made the `myPlugin` and `myClientPlugin` have an `export`. An opinionated default.
- The heading "When to use Middleware, Hooks, On Request, On Response" was removed:
  - That heading was written after the doucmentation of middleware, hooks, on request, etc. Most people skim the docs to find what they want, to read all of that code not knowing what each function does, makes it harder for them to get to the end and finally read that heading which provides more context.
  - I added a new heading, which is above middleware, hooks, etc and that explains each primary function and when you would use them.
- Fixed the client's endpoint interface code block which incorrectly imported {myPlugin} from ".//plugin", rather than "./plugin".
- Added documentation for `getSessionFromCtx`.

### DIFF
--- a/docs/content/docs/concepts/plugins.mdx
+++ b/docs/content/docs/concepts/plugins.mdx
@@ -325,6 +325,37 @@ const myPlugin = {
 }
 ```
 
+### Server-plugin helper functions
+
+Some additional helper functions for creating server plugins.
+
+#### `getSessionFromCtx`
+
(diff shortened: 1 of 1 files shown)


Bekacru:
thanks for the PR!

Issue #813, opened 2024-12-08
Move `demo/nextjs` to `examples/nextjs-example`


Multinite:
**Is your feature request related to a problem? Please describe.**
Shouldn't the nextjs demo project be apart of the other example projects too?

**Describe the solution you'd like**
Move `demo/nextjs` to `examples/nextjs-example`

ping-maxwell:
I can do this, just let me know if you want this.

wrsrsh:
Yeah sure

PR #816, opened 2024-12-08
fix: moved `nextjs` demo to the `example` folder.

Multinite:
closes https://github.com/better-auth/better-auth/issues/813


ping-maxwell:
We still need `demo/nextjs` as it turns out, since the demo is live, unlike other example frameworks.
 https://demo.better-auth.com/
Please check out https://github.com/better-auth/better-auth/pull/817, as this correctly solves this issue.

PR #820, opened 2024-12-08
Disable select for "git:(main) x" in hero

tiagorangel1:
Very small change to disable select for "git:(main) x" in hero. This is useful in order to avoid people copying the "git:(main) x" part and getting an error

image


Bekacru:
thanks 🖤

PR #831, opened 2024-12-09
add(docs): demo code to set password on the server.

ping-maxwell:
image


Bekacru:
hey, thanks again for this PR! really appreciate it. But, we're planning to document the ctx API in v1.1. It'll have its own section, and we’ll likely make some changes to the API. So, let's hold off until then

ping-maxwell:
Sounds good.

PR #837, opened 2024-12-09
added: pathMethods to username client plugin

Multinite:


Bekacru:
thanks for the PR. But first I don't why we're including username related pathnames in email-otp. It should be in the username plugin but even then if the endpoint accepts a body the fetch client defaults to body by default. It's only necessary if it's ambiguous because It doesn't accept body while requiring a post request.

ping-maxwell:
Oh my goodness, I meant to add this to the username client plugin.... not email otp...

ping-maxwell:
Since it isn't necessary, I'll just close this.

PR #845, opened 2024-12-10
fix: email OTP error codes and docs

jasongerbes:
This PR includes improvements for the Email OTP plugin:

- Consistently return errors with values from `ERROR_CODES`.
- Consistently return `OTP_EXPIRED` when the OTP has expired.
- Add details about password reset via OTP to docs.

It also includes a fix for the 'your first plugin' guide (`throw APIError` -> `throw new APIError`).


Bekacru:
thanks for the PR!

PR #852, opened 2024-12-11
feat: remove oslo and migrate to better-auth-utils for all crypto and encoding utils

### LINKED ISSUE #674: Move from `oslo` to `@oslojs/*` packages
[Oslo](https://github.com/pilcrowonpaper/oslo) is getting deprecated at the end of this year, and the recommendation is to use the packages under the [Oslo project](https://oslojs.dev/) which are (almost) a 1:1 export of the original Oslo.

This would also get rid of the `Package oslo can't be external` warnings that pops up on the console when using this package in Next.js.

I'm happy to tackle this if needed, I took a quick glance at the code and moving to the new packages should be pretty straightforward.

(Awesome work by the way! I'm really liking what you've done with this library)

Bekacru:
Since the version of Oslo we use in Better Auth is about to reach end-of-life, we needed to migrate all utilized utilities elsewhere. We were considering using OsloJS (the new Oslo version), but it introduces separate packages for each implementation. which increases the amount of deps we rely on, which is fine but not ideal. And, most implementations in oslojs are written from scratch (we prefer sticking to web crypto impls.). For this we made our in house crypto and auth uility [Better Auth Utils](https://github.com/better-auth/utils). 

This PR aims to migrate all crypto and encoding implementations to **better-auth-utils**.

closes #674


adelrodriguez:
thanks for taking care of this, I've been busy the past week and hadn't had a chance to tackle it

PR #863, opened 2024-12-12
docs: add Nuxt server API route getSession example

TimvdEijnden:
Added an example to get the session on a server route. When implementing this I looked at the other integrations to see how it worked.

### DIFF
--- a/docs/content/docs/integrations/nuxt.mdx
+++ b/docs/content/docs/integrations/nuxt.mdx
@@ -68,6 +68,27 @@ const session = authClient.useSession()
 </template>
 ```
 
+### Server Usage
+
+The `api` object exported from the auth instance contains all the actions that you can perform on the server. Every endpoint made inside Better Auth is a invocable as a function. Including plugins endpoints.
+
+**Example: Getting Session on a server API route**
+
+```tsx title="server/api/example.ts"
+import { auth } from "~/lib/auth";
+
+export default defineEventHandler((event) => {
+    const session = await auth.api.getSession({
+      headers: event.headers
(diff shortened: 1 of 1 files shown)


docs/content/docs/integrations/nuxt.mdx
```diff
@@ -68,6 +68,27 @@ const session = authClient.useSession()
 </template>
 ```
 
+### Server Usage
+
+The `api` object exported from the auth instance contains all the actions that you can perform on the server. Every endpoint made inside Better Auth is a invocable as a function. Including plugins endpoints.
+
+**Example: Getting Session on a server API route**
+
+```tsx title="server/api/example.ts"
```
TimvdEijnden, line 77: Thanks for the quick merge but `tsx` should actually be `ts`

TimvdEijnden, line 77: Thanks for the quick merge but `tsx` should actually be `ts`, sorry 

@Bekacru

poor-john:
Little fix here:
this code needs the "async" for being able to await for the promise.
`import { auth } from "~/lib/auth";
 
export default defineEventHandler( async (event) => {
      headers: event.headers
    });
 
   if(session) {
     // access the session.session && session.user
   }
});`

poor-john:
Little fix here:
this code needs the "async" for being able to await for the promise.
```typescript
import { auth } from "~/lib/auth";

export default defineEventHandler( async (event) => {
    headers: event.headers
  });

  if(session) {
    // access the session.session && session.user
  }
});
```

PR #869, opened 2024-12-12
feat(adapter): Implement Mikro ORM Adapter

octet-stream:
This PR adds adapter for [Mikro ORM](https://mikro-orm.io/)

### DIFF
--- a/packages/better-auth/tsconfig.json
+++ b/packages/better-auth/tsconfig.json
@@ -14,7 +14,8 @@
 		"verbatimModuleSyntax": true,
 		"strict": true,
 		"noImplicitOverride": true,
-		"noFallthroughCasesInSwitch": true
+		"noFallthroughCasesInSwitch": true,
+		"experimentalDecorators": true
 	},
 	"exclude": ["node_modules"],
 	"include": ["src"]
--- a/packages/better-auth/package.json
+++ b/packages/better-auth/package.json
@@ -374,6 +374,9 @@
     }
   },
   "devDependencies": {
+    "@mikro-orm/better-sqlite": "^6.4.1",
+    "@mikro-orm/core": "^6.4.1",
+    "@mikro-orm/postgresql": "^6.4.1",
     "@prisma/client": "^5.22.0",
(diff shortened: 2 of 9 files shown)


packages/better-auth/tsconfig.json
```diff
@@ -14,7 +14,8 @@
 		"verbatimModuleSyntax": true,
 		"strict": true,
 		"noImplicitOverride": true,
-		"noFallthroughCasesInSwitch": true
+		"noFallthroughCasesInSwitch": true,
+		"experimentalDecorators": true
```
octet-stream, line 18: This option enabled so I can define Mikro ORM entities with legacy decorators. If you want to keep it off, I can rewrite schema using [`EntitySchema`](https://mikro-orm.io/docs/entity-schema) and disable this option.

octet-stream:
Also, test can't run on my local machine. I suppose this is because they can't connect to database. I have MariaDB and PostgreSQL installed, so default ports for them already taken, but they are not configured for tests. And I can see you have `docker-compose.yaml` at the project's root, but can't find it being used anywhere for tests. Because of this I configured my adapter to run against SQLite and the tests are passing.

Bekacru:
Hey thanks for the PR! But, our current plan is to stick with the adapters we already have. Since adapters require maintenance and we still need to add features for them, it becomes easily unmanageable. That said, feel free to make this open source! We’d love to include it in the documentation as a community-maintained adapter.

octet-stream:
Yeah, sure I totally understand this is going to be a huge problem really soon. I'll make it to a separate library then.

Also, please add support for joins, it will really help to get all the benefits from RDBMS. I was going to ask about it in a separate issue, but I didn't really come up with a good proposal.

Shared test suite for adapters will also help much, so developers can just take and use it for tests. I'll probably return with better proposal for improvements on adapters tests once I finish porting my adapter to it's own library.

octet-stream:
Published Mikro ORM adapter to npm: https://www.npmjs.com/package/better-auth-mikro-orm

PR #894, opened 2024-12-15
docs: fix redis example for secondaryStorage

matt-potter:
Redis client needs to connect before use, and the stored stringified JSON needs to be parsed before return


matt-potter:
Fixes https://github.com/better-auth/better-auth/issues/895

Issue #858, opened 2024-12-11
useSession hook does not update after passkey authentication


eyenalxai:
workaround: You can call `authClient.$store.notify("$sessionSignal")` on the client where needed

**Describe the bug**
The `useSession` hook does not update when a user successfully authenticates using a passkey. The session state remains unchanged

**To Reproduce**

1. Sign in using passkey authentication
2. Observe that the `useSession` hook does not update to reflect the authenticated status

**Expected behavior**
The `useSession` hook should update to reflect the authenticated session state after a user successfully authenticates with a passkey.

**Additional context**
When I authenticate with a passkey, `"/passkey/verify-authentication"` path is not passing through any atom listeners at all, unlike, for example, `"/two-factor/verify-totp"`, so I can't even bind it to the `"$sessionSignal"` (if I even understood correctly how internals work -_-) and I can't figure out why is what. When comparing both client plugins nothing seems that different in how they're implemented.

Bekacru:
not stale

Mokhtar25:
facing the same issue

SkepticMystic:
Also

snibbo71:
As am I. I have, following the dosubot thingy above, implemented the workaround mentioned in my code - though I really not sure of the ramifications of it... I have the following code;

await authClient.signIn.passkey({
                fetchOptions: {
                  onSuccess: () => {
                    authClient.$store.notify("$sessionSignal");
                    toast.success("You are now logged in, welcome");
                  },
                  onError: () => {
                    toast.error("An error occurred with your passkey");
                  },
                },
              });

There's some more logic in my code but that's the bit I've changed to get the authClient.useSession() to notice when a passkey user logs in.

issam-seghir:
#4731

PR #911, opened 2024-12-16
fix: session signal on passkey authentication

LovelessCodes:
closes [#858](https://github.com/better-auth/better-auth/issues/858)


Bekacru:
The problem is that the signals are currently only called in a proxy context. `signIn.passkey` is actually a defined action. We need to refactor a bit to pass the signals to the actions so they can be called from there as well. But unfortunately, the path matcher won't work.

PR #912, opened 2024-12-16
docs(2fa): has wrong information about client properties

LovelessCodes:
`onTwoFactorRedirect` doesn't seem to be available on the client anymore, so I've switched up to use `redirect` & `twoFactorPage` properties on the documentation about 2fa implementation.


Bekacru:
https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/two-factor/client.ts#L9

it's still `onTwoFactorRedirect`

LovelessCodes:
You are totally correct, I must've been on an old build.

PR #922, opened 2024-12-17
fix: remove nanoid dependency and generate IDs using better auth utils

### LINKED ISSUE #469: Error with nanoid module
**Describe the bug**
Error [ERR_REQUIRE_ESM]: require() of ES Module \node_modules\nanoid\index.js from \node_modules\better-auth\dist\index.cjs not supported. 
Instead change the require of index.js in \node_modules\better-auth\dist\index.cjs to a dynamic import() which is available in all CommonJS modules.

**To Reproduce**
This is my betterAuth initialization code. Upon further use, the server crashes and returns an error

```ts
import { prismaAdapter } from 'better-auth/adapters/prisma'

import { prisma } from '@/config/database'

    provider: 'postgresql',
  }),
})
```

**Additional context**
I use NodeJS with Express.

Here is my tsconfig.json:

```json
{
  "compilerOptions": {
    "target": "es2021",
    "module": "commonjs",
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "skipLibCheck": true,
    "rootDir": "src",
    "outDir": "dist",
    "paths": {
      "@/*": ["./src/*"]
    }
  },
  "ts-node": {
    "require": ["tsconfig-paths/register"]
  }
}
```

Bekacru:
closes #469


daveycodez:
Interesting I was curious about the choice of nanoid's over uuid's and now seeing they are getting pulled. Or is Better Auth utils just still generating nanoid's without the dependency

PR #940, opened 2024-12-18
fix(demo): correct name on .env.example

danimarin24:
A very small and silly change, but one that my OCD could not cope with.


Bekacru:
thanks :D

PR #957, opened 2024-12-19
feat: create Cloudflare Turnstile middleware plugin

### LINKED ISSUE #956: Support Cloudflare Turnstile

It is often a good idea to put some endpoints behind a CAPTCHA challenge. The sign-up, sign-in and forgot password endpoints exposed by better-auth could benefit from such security.

This would be a perfect candidate for a plugin. Using the middleware to get a captcha response token, we could then submit it to Cloudflare and only let the request through if the challenge is passed.
The plugin would allow for custom "protected endpoints" array, so fit all use-cases.

Without creating a custom plugin, I cannot think of a way to protect the better-auth default endpoints behind a CAPTCHA challenge. Unless we don't use the better-auth client and build custom endpoints.

### Additional context

I have built such a plugin already for my personal project. I will happily submit a PR!
I would be also happy to build a similar plugin for Google ReCAPTCHA

0scrm:
Fixes #956 

**This is the initial implementation of a Cloudflare Turstile plugin.** 

It's pretty straightforward: the server plugin is an `onRequest` middleware that submits the CAPTCHA token to Cloudflare's `/siteverify` endpoint and intercepts  any invalid/missing CAPTCHA challenge.


osrim:
Error in the pipeline:

```bash
 ❯ src/plugins/turnstile/turnstile.test.ts > turnstile > Should return 400 if no captcha token is found in the request headers
 → API Error: INTERNAL_SERVER_ERROR Something went wrong
 ✓ src/plugins/open-api/open-api.test.ts  (1 test) 4ms
 ✓ src/api/routes/sign-out.test.ts  (1 test) 199ms
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯


 FAIL  src/plugins/turnstile/turnstile.test.ts > turnstile > Should return 400 if no captcha token is found in the request headers
BetterCallAPIError: API Error: INTERNAL_SERVER_ERROR Something went wrong
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Serialized Error: { status: 'INTERNAL_SERVER_ERROR', headers: { constructor: 'Function<Headers>', append: 'Function<append>', delete: 'Function<delete>', get: 'Function<get>', has: 'Function<has>', set: 'Function<set>', getSetCookie: 'Function<getSetCookie>', keys: 'Function<keys>', values: 'Function<values>', entries: 'Function<entries>', forEach: 'Function<forEach>' }, body: { message: 'Something went wrong', code: 'SOMETHING_WENT_WRONG' } }
```

Bekacru:
Hey, great work on this! I think the middleware error is happening because you're throwing `APIError`. We should document this more clearly, but in `onRequest`, throwing `APIError` isn't handled by the router. Instead, the server assumes an unexpected error occurred. You should return a `Response` object instead, like this:

```ts
return new Response();
```

osrim:
That was it, thanks! 

Should the documentation be part of this PR? Not sure what process you follow for releasing the documentation :)

Bekacru:
would be nice if the doc change is another PR

Bekacru:
Hey this is great! could you add the doc for the plugin as well?

Bekacru:
> would be nice if the doc change is another PR

I think I misunderstood your question here. I was suggesting the doc for mentioning `APIError` shouldn't be thrown in a middleware should be another PR. The plugin docs would be great if they are included within this PR

osrim:
Done! Added the plugin to the **Utility** section of the docs.

jasongerbes:
Nice work! Perhaps it would be good to add some of the Email OTP endpoints to the defaults

osrim:
> Nice work! Perhaps it would be good to add some of the Email OTP endpoints to the defaults

Email OTP is another plugin though, I would argue it makes sense to keep these endpoints out of the defaults. Wdyt?

jasongerbes:
> Email OTP is another plugin though, I would argue it makes sense to keep these endpoints out of the defaults. Wdyt?

That's a fair point. I'm not sure how the integration between plugins is expected to be handled.

Perhaps @Bekacru could help with this question?

dergreg:
What is missing for a merge? I could also use this functionality

osrim:
Aside from the open question on default endpoints, I think this PR is ready to be merged. We just need approval / feedback from @Bekacru :)

osrim:
@Bekacru as discussed in #1332, I updated this PR to support multiple captcha providers out of the box (google-recaptcha and cloudflare-turnstile for now). 

The logo used in the docs: https://iconduck.com/icons/146285/captcha (Apache 2.0 licence, from what I can see there's no need to credit the author).

Bekacru:
thanks! 🖤

Issue #421, opened 2024-11-05
Feature Request: Sign-in with Identity Token


oktaysenkan:
**Is your feature request related to a problem? Please describe.**
I need to implement authentication with an OIDC ID token, but the current library doesn’t provide a direct method for this.

**Describe the solution you'd like**
I would like a method similar to `supabase.auth.signInWithIdToken` that allows signing in with an OIDC ID token.

**Additional context**
This example works with native login prompt.

```tsx
import * as AppleAuthentication from 'expo-apple-authentication';

const credential = await AppleAuthentication.signInAsync({
    requestedScopes: [
        AppleAuthentication.AppleAuthenticationScope.FULL_NAME,
        AppleAuthentication.AppleAuthenticationScope.EMAIL,
    ],
});

console.log(credential.identityToken)
```

![image](https://github.com/user-attachments/assets/e8626b85-cd04-4e03-a80a-4c04db7d0e9a)

[Supabase Reference](https://supabase.com/docs/reference/javascript/auth-signinwithidtoken)

abegehr:
This is great! Using the Apple identityToken, do we still have to set clientId and clientSecret on the apple provider? If so, when generating the apple clientSecret it needs to have an expiration date. Is it fine to use a super-long-lived token for clientSecret? https://developer.apple.com/documentation/accountorganizationaldatasharing/creating-a-client-secret

abegehr:
Max. token lifetime that apple excepts seems to be 6 months: https://github.com/Faruqt/generate-apple-client-secret-key?tab=readme-ov-file#generate-apple-client-secret-key - therefore best-case would be to generate a clientSecret in the application itself from the p8-key.

abegehr:
In case, we want to generate the apple client secret on-the-go in the future, here is a sample: https://github.com/WcaleNieWolny/capgo-social-login-backend-demo/blob/main/index.ts#L25

abegehr:
I finally got "Sign-in with Apple" working on my iOS Capacitor app (should be the same for all hybrid platforms):

The main thing that tripped me up, is that better-auth requires `clientId` and `clientSecret` on `socialProviders.apple` in config -> this is where you're place the service id and jwt secret generated from p8-file for "Sign-in with Apple" on web.
However on native iOS, it doesn't use the service id but the app id (bundle id) as client id, so if using the service id as clientId in `signIn.social()` with `idToken`, it throws an error:  `JWTClaimValidationFailed: unexpected "aud" claim value`. 

[...]

PR #966, opened 2024-12-20
fix: sign In with Apple With ID Token not work

hyoban:
> The main thing that tripped me up, is that better-auth requires clientId and clientSecret on socialProviders.apple in config -> this is where you're place the service id and jwt secret generated from p8-file for "Sign-in with Apple" on web.
However on native iOS, it doesn't use the service id but the app id (bundle id) as client id, so if using the service id as clientId in signIn.social() with idToken, it throws an error: JWTClaimValidationFailed: unexpected "aud" claim value.

> For iOS native "SignIn with Apple" with idToken, it expects the app/bundle id as clientId and doesn't require the clientSecret.

https://github.com/better-auth/better-auth/issues/421#issuecomment-2466911181

The return value of the `decodeJwt` function is already the payload

### DIFF
--- a/docs/content/docs/authentication/apple.mdx
+++ b/docs/content/docs/authentication/apple.mdx
@@ -25,10 +25,14 @@ description: Apple provider setup and usage.
                 apple: { // [!code highlight]
                     clientId: process.env.APPLE_CLIENT_ID as string, // [!code highlight]
                     clientSecret: process.env.APPLE_CLIENT_SECRET as string, // [!code highlight]
+                    // Optional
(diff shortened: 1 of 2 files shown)


Bekacru:
thanks! although I think with clientId it should work on latest. We moved to jose decodeJWT just now. It's not released yet. Before the change the payload was under the payload object.

hyoban:
In our usage, service id and bundle id cannot be the same value, which caused me to encounter this problem

Bekacru:
yeah that was a problem. tagged on `1.0.23-beta.6`.

PR #985, opened 2024-12-21
feat: Add account parameter to custom password verifier

kzlar:
Hi,

Following some discussion on discord (https://discord.com/channels/1288403910284935179/1288403910284935182/1320099161659674725) I figured I'd take a stab at adding an `account` parameter to the custom password verifier.
In short, this helps in implementing external authentication, similar to what is provided by authjs (https://next-auth.js.org/configuration/providers/credentials).
I'd say this is probably not something most people should use, but given that the API already exists, I think this PR makes it more useful without adding more risks.

Opened this as Draft to collect feedback and also because for some reason the build fails on my setup (but in my defense the main branch also fails so perhaps its a local issue)

### DIFF
--- a/packages/better-auth/src/init.ts
+++ b/packages/better-auth/src/init.ts
@@ -4,6 +4,7 @@ import { createInternalAdapter } from "./db";
 import { getAuthTables } from "./db/get-tables";
 import { getAdapter } from "./db/utils";
 import type {
+	Account,
 	Adapter,
 	BetterAuthOptions,
 	BetterAuthPlugin,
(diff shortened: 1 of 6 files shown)


packages/better-auth/src/init.ts
```diff
@@ -195,7 +196,7 @@ export type AuthContext = {
 	secondaryStorage: SecondaryStorage | undefined;
 	password: {
 		hash: (password: string) => Promise<string>;
-		verify: (data: { password: string; hash: string }) => Promise<boolean>;
+		verify: (data: { password: string; hash: string, account: Account }) => Promise<boolean>;
```
Bekacru, line 199: let's add a second argument here for `request?: Request`

Bekacru: you can pass `ctx.request` on the calls

Bekacru (changes requested):

PR #1004, opened 2024-12-23
fix: create a new session on email verification if current session email doesn't match

### LINKED ISSUE #996: Email Verification isn't logging the user in

### To Reproduce

Click an email verification link, and you have to enter the email and password again.

Email Verification link should act like a magic link and log the user in

1.1.2

```bash
Nextjs
```

Client

_No response_

### Additional context

_No response_

Bekacru:
closes #996


daveycodez:
@Bekacru Does this also fix reset password flow? I'm having the same issue after clicking the Reset Password link from the email. It's taking me to /reset-password but I'm still anonymous

PR #1037, opened 2024-12-27
feat(organization): support multiple roles

### LINKED ISSUE #1028: Multiple Roles on Organization Member

For my use-case, I need organization support, which I have in Auth0 and now also in Better Auth, very happy about that. But for members of these organizations, I need to have roles. This is possible, but unfortunately limited in an important sense. In Auth0 I can have multiple roles per member, say this person is the "project manager but also the accountant" (just an example). This is not possible right now with the organization plugin, where one member can only have a singular role. I'm using this extensively, and would require it in order to fully move to Better Auth.  

I'd like the option of having multiple roles per member in organizations. 

Customizing the member in the betterAuth() call, unfortunately it's not supported in the same way as user is ("additionalFields").

Customizing the member schema directly, as recommended in discord. But this is unfortunately not a perfect solution since it seems like this would become quite problematic since that custom property won't be included in the BetterAuth built-in calls. Doing this seems like a ticking timebomb since I'm not a solo-developer on this project. 

Considered a kind of merge in the role name, like having it more like "owner:member:etc" but this won't scale and I've got quite a lot of roles, where I'd need all possible combinations... Not feasible. 

### Additional context

_No response_

Bekacru:
closes #1028

### DIFF
--- a/packages/better-auth/src/plugins/organization/has-permission.ts
+++ b/packages/better-auth/src/plugins/organization/has-permission.ts
@@ -1,0 +1,21 @@
+import { defaultRoles } from "./access";
+import type { OrganizationOptions } from "./organization";
+
+export const hasPermission = (input: {
+	role: string;
+	options: OrganizationOptions;
+	permission: {
+		[key: string]: string[];
+	};
+}) => {
+	const roles = input.role.split(",");
+	const acRoles = input.options.roles || defaultRoles;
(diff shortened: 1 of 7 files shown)


albinjon:
Hi! 
Can I assist in some way with this one? Just let me know, I'd be happy to help.

PR #1042, opened 2024-12-27
Add expiresIn for email verification

keytrap-x86:
This adds the possibility to set an expiration time (in seconds) for the verification mail.
I left it a 3600s as default, but you can add any time.

Example:
```typescript
emailVerification: {
        expiresIn: 7200, // 2 hours
            // Send the email verification email
        },
    }
```


Bekacru:
thanks!

PR #1051, opened 2024-12-28
fea(jwt)t: sign JWT and get private key

Bekacru:
expose sign jwt and a method to get a private key

- [x] implementation 
- [ ] test
- [ ] docs

### DIFF
--- a/packages/better-auth/src/plugins/jwt/index.ts
+++ b/packages/better-auth/src/plugins/jwt/index.ts
@@ -442,9 +442,7 @@ export const jwt = (options?: JwtOptions) => {
 								data: JSON.parse(key.privateKey),
 							})
 						: key.privateKey;
-
-					const privateKey = await importJWK(JSON.parse(privateWebKey));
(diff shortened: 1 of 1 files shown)


daveycodez:
We need optimized versions of these functions that don't hit the API, especially for Edge middleware or proxies where JWT is likely to be signed on the fly.

There should be a function signJWT(jwtSecret?, jwks, payload) so we can use cached versions to sign instantly without needing any HTTP requests or database queries. And maybe even default to use process.env.JWT_SECRET if not passed.

Although shouldn't the JWKS secret be the same as BETTER_AUTH_SECRET ?

This is important for Neon RLS where we need to cache the JWKS anyways, I have it set up as my domain.com/jwks.json for example which is on the edge cache. I can just import it into my SSR routes & middleware so I don't need to go fetch it from the DB

daveycodez:
Another thing we will need is an optimistic signing based on cookie

better-auth.session_data

That just immediately verifies and signs the existing cookie

PR #1059, opened 2024-12-29
feat: unlink account

### LINKED ISSUE #1038: Option to Unlink Social Providers

Currently there doesn't seem to be a way to unlink a Social Provider out of the box, would be great if the authClient provided the option to do this. E.g. authClient.unlinkSocial

Only other way I can think of doing this is a custom API route, getting the current session, then deleting that account ID from the DB manually. Would be great to have this feature built in, most sites allow you to remove OAuth connections

Bekacru:
closes #1038


hyoban:
Could you merge this? It is an important feature for managing linked social accounts.

Issue #1036, opened 2024-12-27
auth.api.signUpEmail return value problem


wh5938316:
### Is this suited for github?

- [ ] Yes, this is suited for github

### To Reproduce

In the latest version, I noticed that the return value of `auth.api.signUpEmail` is `{ token: null }`, instead of returning the user object as it used to (if I remember correctly). This is causing issues for me, as I need the created user object for subsequent processing in my application.

1. What is the purpose of this change?
2. If signUpEmail no longer returns the user object, how should I retrieve the created user after calling signUpEmail?

```javascript
const data = await betterAuth().api.signUpEmail({
    body: {
      email: _.get(data, 'email'),
      password: _.get(data, 'password'),
      name: _.get(data, 'name')!,
    },
});
```

In version 1.0.0:
https://github.com/better-auth/better-auth/blob/8e81ceb0e2727e220a9119a6bccca2c01fc878c3/packages/better-auth/src/api/routes/sign-up.ts#L215-L218

Now:
https://github.com/better-auth/better-auth/blob/2c81db93be61b7154fe37ec9fd709806366ca93b/packages/better-auth/src/api/routes/sign-up.ts#L233-L235

### Current vs. Expected behavior

I would like clarification on the following:
1. Was this change intentional as part of a design decision?
2. If so, how should I obtain the user information after calling signUpEmail?

### What version of Better Auth are you using?

1.1.3

### Provide environment information

```bash
Nest.js
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});
```

wh5938316:
@Bekacru hi

This update originates from PR #968, which states:

> Endpoints that previously returned the entire user object now only return a token. This change allows for storing sensitive fields in the user table and eliminates unnecessary returned data. As a result, all endpoints, except /get-session, now enforce this.

What I don’t understand is that the purpose of this PR is to allow storing sensitive fields in the user table and eliminate unnecessary returned data. However, I believe returning the created user in `api.signUpEmail` is necessary, and specifying the returned fields already avoids the issue of exposing sensitive fields.

Interestingly, in the current version, it seems that `api.signInEmail` is an exception, as it returns the user information.

[...]

PR #1067, opened 2024-12-30
feat: restore the user data in the response

wh5938316:
I propose restoring the user data in the response of api.signUpEmail to complete the registration flow in my Nest.js application. Without this change, I am restricted to using versions of better-auth prior to v1.0.23.

The issue I am facing is described in #1036. Importantly, this modification does not break any existing functionality. Notably, the api.signInEmail endpoint already includes the user data in its response (which seems to be an oversight).

I kindly request that this PR be merged as a temporary measure.

### DIFF
--- a/packages/better-auth/src/api/routes/email-verification.ts
+++ b/packages/better-auth/src/api/routes/email-verification.ts
@@ -267,6 +267,15 @@ export const verifyEmail = createAuthEndpoint(
 			}
 			return ctx.json({
 				status: true,
+				user: {
+					id: updatedUser.id,
+					email: updatedUser.email,
+					name: updatedUser.name,
+					image: updatedUser.image,
+					emailVerified: updatedUser.emailVerified,
+					createdAt: updatedUser.createdAt,
+					updatedAt: updatedUser.updatedAt,
+				},
 			});
 		}
 		await ctx.context.internalAdapter.updateUserByEmail(parsed.email, {
@@ -293,6 +302,7 @@ export const verifyEmail = createAuthEndpoint(
 		}
 		return ctx.json({
 			status: true,
+			user: null,
 		});
 	},
 );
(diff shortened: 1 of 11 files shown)


Bekacru:
Hey, thanks for the PR! We can go with this, but could you please add the user data to other endpoints and plugins that generate a new session and only return the `token` as well?

wh5938316:
Sure, happy to help!

By the way, I’d like to discuss something. Based on my understanding of REST API conventions, a POST /sources endpoint typically returns the created resource. Following this logic, signUp creates a user resource, so it makes sense to return the user. Similarly, signIn creates a session resource, so returning the session would be reasonable as well.

Do you agree with this reasoning? If so, I can include the basic session data in the response as part of this PR. If not, I’ll limit this PR to restoring the user data only.

Looking forward to your feedback!

wh5938316:
This approach is also reflected in Supabase’s API design. For example, the auth.signInWithPassword API returns both the session and user data: [Supabase Auth API Reference](https://supabase.com/docs/reference/javascript/auth-signinwithpassword).

Bekacru:
We used to return the entire object. But, we want to avoid this because users can easily extend the tables and might introduce sensitive data that shouldn't be exposed to the user. Returning only the base session fields we provide may not be a bad idea. But for now let's keep it with the user data.

wh5938316:
Hi, @Bekacru 

I’ve added the user data to all endpoints and plugins that generate a new session and previously only returned the token, as requested.

Please let me know if there’s anything else to adjust. If everything looks good, I’d appreciate it if you could merge this PR.

Thank you!

wh5938316:
In addition to the requested changes, I noticed that in the one-tap plugin, the following section of code:

https://github.com/better-auth/better-auth/blob/cc94e6a7c07c5742938a15e6e340d3dde3de17b3/packages/better-auth/src/plugins/one-tap/index.ts#L110-L113

currently returns the complete session in the main branch. I believe this might be an oversight. I have updated it accordingly in this PR, but I recommend giving this specific change an extra review.

Let me know if this aligns with your expectations!

Bekacru:
yeah this looks great. thanks!

PR #1083, opened 2024-12-31
feat: support overwriting auth-url params in the generic-oauth plugin

### LINKED ISSUE #1064: In Generic OAuth plugin, add option to send additional custom OAuth parameters

Some OAuth providers support the use of additional querystring parameters when calling their authorization endpoints. (For example, Asana supports sending the parameter `display_ui=always` in which case the user will be shown the Asana authorization dialog even if they have previously authorized the app.) Currently there is no support in the Generic OAuth plugin for passing additional custom parameters such as this.

I would like to see support added to the Generic OAuth plugin for passing additional custom parameters to the OAuth authorization endpoint. This could be defined by adding an `additionalParameters` option to the Generic OAuth plugin configuration.

I'm unable to think of any alternative ways to pass an additional custom parameter.

### Additional context

_No response_

jcqvisser:
Various oauth providers support custom parameters as part of their `authorization_endpoint` url. This allows users of the `genericOAuth` plugin to provide those custom parameters.

closes #1064

### DIFF
--- a/packages/better-auth/src/plugins/generic-oauth/index.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/index.ts
@@ -101,6 +101,11 @@ interface GenericOAuthConfig {
 				emailVerified?: boolean;
 				[key: string]: any;
 		  }>;
+	/**
+	 * Additional search-params to add to the authorizationUrl.
+	 * Warning: Search-params added here overwrite any default params.
+	 */
+	authorizationUrlParams?: Record<string, string>
 }
 
 interface GenericOAuthOptions {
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/generic-oauth/index.ts
```diff
@@ -101,6 +101,11 @@ interface GenericOAuthConfig {
 				emailVerified?: boolean;
 				[key: string]: any;
 		  }>;
+	/**
+	 * Additional search-params to add to the authorizationUrl.
+	 * Warning: Search-params added here overwrite any default params.
+	 */
+	authorizationUrlParams?: Record<string, string>
```
jcqvisser, line 108: I'm not sure what the best name for this would be...

PR #1107, opened 2025-01-03
chore: move nextjs demo to examples

TinsFox:
Move nextjs demo to the examples directory, unified management can make it easier for users to find


ping-maxwell:
@TinsFox  Please check out https://github.com/better-auth/better-auth/pull/816 for why we don't move nextjs demo to the examples directory.

TinsFox:
> @TinsFox Please check out #816 for why we don't move nextjs demo to the examples directory.

Of course, we need this demo project, but this does not prevent us from modifying its location. After modifying the position, it can be more intuitive.

But maybe we have other things that need to be addressed at the same time

- [x] Modify the address on the document to point to the new directory
- [ ] Modify the deployed listening directory. If it is Vercel, you need to modify its Root Directory pointer. This may require @Bekacru to operate.
	image
- [x] Maybe also need to solve Stackblitz's read position

TinsFox:
Hi @Bekacru

The examples of nextjs have been completely migrated to examples, including documentation. If you find this useful, you can merge the PR and then modify the deployed listening directory and redeploy the demo, so that all migrations can be completed.

PR #1110, opened 2025-01-03
feat: cleanup expired verification data on fetch

### LINKED ISSUE #1046: Clean Out Expired Verifications from Database

_No response_

It looks like the verifications table doesn't get cleaned up automatically, would be great to have a DELETE where expired query applied to it at some point during verification process. I imagine this table can get pretty overflowing with a large user base.

custom cronjob

### Additional context

_No response_

### LINKED ISSUE #1085: Issue: Accumulation of Zombie Rows and Misplaced User Deletion Values

### To Reproduce

I have simple email OTP sign in created using email OTP plugin, as described [here](https://www.better-auth.com/docs/plugins/email-otp).

User can click button for resending OTP, which does this:

```ts
const result = await authClient.emailOtp.sendVerificationOtp({  
   email,  
   type: "sign-in",  
});  
```

If this button is pressed multiple times, unnecessary rows pile up in database:

image

These previous OTP's are useless, user can't sign in using them, so it does not make sense to keep them. They even stay in database after successful verification for that user.

When new row is added to verification table, all previous rows with same identifier should be deleted. This is also a security improvement because only one verification token or OTP will be available.

Additionally, after successful verification all rows associated with it should be cleared (both valid and invalid ones), just in case. Currently, only last row is deleted. 

This does not apply only to email OTP, but to all things which use verification table (like user deletion).

***

Just little side note here. It seems that identifier and token are misplaced in rows associated with user deletion. User identifier is in _value_ column, while deletion token is in _identifier_. Maybe this is intentional, but it is little bit strange. For email OTP situation is opposite:

image

Probably token and identifier should switch places in rows for user deletion.

1.1.7

```bash
Safari, MacOS
```

Backend

_No response_

### Additional context

_No response_

Bekacru:
closes #1085 
closes #1046

### DIFF
--- a/packages/better-auth/src/api/routes/update-user.ts
+++ b/packages/better-auth/src/api/routes/update-user.ts
@@ -480,9 +480,6 @@ export const deleteUserCallback = createAuthEndpoint(
 			`delete-account-${ctx.query.token}`,
 		);
 		if (!token || token.expiresAt < new Date()) {
-			if (token) {
-				await ctx.context.internalAdapter.deleteVerificationValue(token.id);
-			}
 			throw new APIError("NOT_FOUND", {
 				message: BASE_ERROR_CODES.INVALID_TOKEN,
 			});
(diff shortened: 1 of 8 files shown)


reslear:
By the way, what if this data is stored in redis (secondaryStorage) and after phoneNumber verified add to db ?

Bekacru:
for secondary storage you should use ttl. and if the ttl value isn't provided from better-auth you can use some default value that's reasonably long.

PR #1128, opened 2025-01-04
fix(docs): extending plugin schemas demo code

Multinite:


ping-maxwell:
# Before
image

# After
image

PR #1137, opened 2025-01-04
refactor(DrizzleAdapter): minor code improvements

ping-maxwell:
* made `getSchema` have a return type for a more strict type safety.
* removed `getModelName` from within the `getSchema` call in line 189 since the `getSchema` fn already uses `getModleName`, it’s redundant.


Bekacru:
this is fixed on upstream already

PR #1152, opened 2025-01-07
Adding possible config path /auth

Ymirke:
I'm currently working on a project that will need to extend better-auth a bit, so I would like to have everything related to auth under `src/auth`, and having the instance exported from `src/auth/auth.ts`.

I'm currently doing that, but exporting better-auth instance from /src/auth.ts. 

Unsure if there are any implications to consider with this.


Bekacru:
this should be fine. just mention this is a possible path on the docs.

Ymirke:
Added it to the docs

PR #1166, opened 2025-01-09
feat: Add Membership Validation to getFullOrganization Endpoint closes #1165

### LINKED ISSUE #1165: Enforce Membership Validation for getFullOrganization Endpoint

Currently, the `getFullOrganization` endpoint allows any user to retrieve an organization's details by providing a valid `organizationId` or `organizationSlug`, regardless of their membership in that organization. This can lead to unauthorized data access and security concerns.

The `getFullOrganization` endpoint should validate that the requesting user is a member of the organization before returning its details. This can be implemented by:

- Fetching the list of organizations the user belongs to (using `adapter.listOrganizations` or a similar method).
- Checking whether the provided `organizationId` or `organizationSlug` matches any of the organizations the user is a member of.
- Returning an appropriate error (e.g., `ORGANIZATION_ERROR_CODES.USER_IS_NOT_A_MEMBER_OF_THE_ORGANIZATION`) if the user is not a member.

1. Adding middleware to enforce this check globally for endpoints accessing organization details.
2. Relying on a session-bound `activeOrganizationId` but this is insufficient as it doesn't handle `organizationSlug`.

### Additional context

- This change would align the behavior of getFullOrganization with other endpoints like listOrganizations, which inherently filters results based on the user's membership.
- Suggested Error Code: `ORGANIZATION_ERROR_CODES.USER_IS_NOT_A_MEMBER_OF_THE_ORGANIZATION`.
- Example Use Case: Preventing unauthorized users from accessing sensitive organization metadata by simply knowing a valid organizationId or organizationSlug.

msywulak:
This PR adds membership validation to the `getFullOrganization` endpoint by checking if the requesting user is a member of the specified organization. This change enhances security by ensuring that only authorized users can access the organization's details.

Reused isMember check from `setActiveOrganization`

closes #1165

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-org.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-org.ts
@@ -396,7 +396,6 @@ export const getFullOrganization = createAuthEndpoint(
 			(member) => member.userId === session.user.id,
 		);
 		if (!isMember) {
-			await adapter.setActiveOrganization(session.session.token, null);
 			throw new APIError("FORBIDDEN", {
 				message:
 					ORGANIZATION_ERROR_CODES.USER_IS_NOT_A_MEMBER_OF_THE_ORGANIZATION,
--- a/packages/better-auth/src/plugins/organization/routes/crud-org.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-org.ts
@@ -392,6 +392,16 @@ export const getFullOrganization = createAuthEndpoint(
 			organizationId,
 			isSlug: !!ctx.query?.organizationSlug,
 		});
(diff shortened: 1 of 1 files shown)


hakai2k:
Isn't this directly conflicting with the `membersLimit` query parameter on the client side? If the person making the request wants to retrieve only 5 records and his own member id isn't available in the `organization.members` list - as per this commit we just set `adapter.setActiveOrganization(session.session.token, null);` logging him out of the organization and `membersLimit` never working.

I am no expert but don't you think it would be better if we made a request to the database to check if the member making the request belongs to that organization and go off of that?

hakai2k:
@msywulak

PR #1169, opened 2025-01-10
Fix (bearer): 'set-auth-token' header not exposed with CORS (#917)

### LINKED ISSUE #917: ctx.response.headers.get("set-auth-token") is null

### To Reproduce

image

ctx.response.headers.get("set-auth-token") is null.

1.0.22

```bash
-OS mac
-chrome
```

Client

```typescript

  trustedOrigins: ["http://", "https://", "file://"],
    schema: {
      user: user,
      account: account,
      session: session,
      verification: verification,
    },
    provider: "pg", // or "mysql", "sqlite"
  }),
  emailAndPassword: {
    enabled: true
  },
  plugins: [
    bearer()
  ]
});
```

### Additional context

_No response_

jacoborus:
closes #917

### DIFF
--- a/packages/better-auth/src/plugins/bearer/index.ts
+++ b/packages/better-auth/src/plugins/bearer/index.ts
@@ -105,6 +105,7 @@ export const bearer = (options?: BearerOptions) => {
 						}
 						const token = sessionCookie.value;
 						ctx.responseHeader.set("set-auth-token", token);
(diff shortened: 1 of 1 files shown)


movoid12:
@jacoborus great! that fixed the issue for me by adding this to my cors middleware im my hono server 👍

Issue #1023, opened 2024-12-26
feat: Add separate social (OAuth) sign-up and sign-in flows


aryanprince:
### Is this suited for github?

- [X] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

Currently, there's no way to distinguish between OAuth sign-up vs sign-in flows. Better Auth's client API only provides `authClient.signIn.social()` which handles both new and existing users. This makes it impossible to:

1. Tell new users they need to create an account first
2. Have separate UI/UX flows for new vs returning users
3. Handle sign-up and sign-in errors differently

<div align="center">
    <img src=https://github.com/user-attachments/assets/fc1bd1ea-ed5d-4369-aae1-fa24c6bcc653 width=400/>
</div>

### Describe the solution you'd like

Add a new `authClient.signUp.social()` method (on the client API) specifically for new user registration via OAuth. This would:

1. Only work for new users (error if account exists)
2. Match how it's already used for email/password auth where `authClient.signUp.email()` and `authClient.signIn.email()` are separate functions

Example usage:

```ts
// For new users
await authClient.signUp.social()

// For existing users
await authClient.signIn.social()
```

### Describe alternatives you've considered

I couldn't do this successfully on my own, but maybe if I manually queried the DB (using Drizzle ORM for example) to check if the returned OAuth token exists to know if it's an existing user may have worked. But again, I believe this should be part of an API from Better Auth itself.

### Additional context

Having separate OAuth sign ins and sign ups would improve the user experience a lot, by making it clear whether a user is creating a new account or accessing an existing one.

But I much prefer letting the user know that they have an account or not, instead of them having to "accidentally" creating new accounts when trying to sign in with different OAuth accounts.

### Backwards Compatibility Note

I've noticed that a lot of websites DO allow new users to "sign in" to create a new account when using OAuth. So maybe the API can still allow sign ups and sign ins using `signIn.social()` by default, and we could have an extra option like `{ preventSignUps: true }` on the `signIn.social()` function so new users can't register with the "sign in" button.

daveycodez:
This is interesting. On one hand, there is something nice about keeping it simple and handling both sign ups and sign ins, but I do understand the point about accidentally creating a new account, although this is a user error. I think the solution in this scenario would be for the user to log out, log in as their original account, and then be able to Link that social account even though it already belongs to another "user" object. I don't know if this works, but I think it should. So I would say the best of both worlds is

1. Option to preventSignUps on signIn.social
[...]

PR #1179, opened 2025-01-10
Feature: Disable Implicit Signup for oauth implementations

kzlar:
- Added disableImplicitSignup option to social providers, generic oauth privders and sso plugin
- With disableImplicitSignup as true, signing in with a new email will fail
- Client may use requestSignUp as true to indicate they're explicitly asking for a sign up
- Added unit tests for social providers, generic oauth and sso plugin


hiaaryan:
@kzlar This is great! I've been looking to implement this through hooks but yours would ease the process up a bit. #1023

kzlar:
> @kzlar This is great! I've been looking to implement this through hooks but yours would ease the process up a bit. #1023

Hey @hiaaryan , let me know if you're able to do this with hooks, I was only able to do it in the case of sign-in with id token, but not if I use the regular oauth flow.

@Bekacru fyi seems like there's interest in this, would appreciate your review :)

IFTTD:
Exactly what is needed as a lot of organization uses cases do require a manual sign up or by invite only. I thought this would have already been implemented like the disableSignUp for magic Links is. I spent a lot of time with hooks and other methods to get this done but none work or work good. Hopefully this gets merged soon.

arthberman:
+1

bigx333:
Also been looking forward to this feature

CahilFTC:
+1 this feature would be very useful

austinimperial:
+1, this would be a great add

PR #1201, opened 2025-01-13
feat: check organization slug

### LINKED ISSUE #589: Support for unique organization slugs
One use case of organizations' slugs, it's to be used to generate unique links for each organization and for this to work, we need each slug to be unique and we also need to have a way to check whether a slug is already being used from the client.

The ideal solution for me, would be to make the organization `slug` column unique, and also adding a new function to the client to check if a slug is already taken.

The current alternative is to manually make the column unique and create an API to check whether a slug is taken.

**Additional context**

Bekacru:
closes #589


D3visionNL:
@Bekacru and I believe this one was accidentally closed too lol

PR #1206, opened 2025-01-13
feat: add refetch function for all client hooks

### LINKED ISSUE #753: Mutate the session state when using the useSession() hook for JWT
Add an update function to the return of useSession.

```js
const { update } = useSession()
await update({... arbitrary data})
```

This will update the session anywhere, where you are referencing it with useSession(). The update() method will also return the updated session.

Bekacru:
closes #753


shadiramadan:
Super excited for this one @Bekacru !  I've migrated from Next-auth and have had to comment out some functionality (including my usage of cloudflare-turnstile) until this milestone gets released.  I'm stoked!

PR #1208, opened 2025-01-14
feat: oauth2 link account

### LINKED ISSUE #851: Ability to link Generic OAuth provider to account
I'd like the ability to link a generic oauth provider to an already existing account.

Using the generic oauth plugin, i'd like the ability to do:
```ts
genericOAuth({
      config: [
        {
          providerId: "roblox",
          clientId: env.ROBLOX_CLIENT_ID,
          clientSecret:
            env.ROBLOX_CLIENT_SECRET,
          discoveryUrl:
            "https://apis.roblox.com/oauth/.well-known/openid-configuration",
        },
      ],
    }),
```

to then do after, when the user already has an account (they'll already be signed in at this point, and I just want the ability to link the roblox account to their account)

```ts
await authClient.linkSocial({
	provider: "roblox",
	callbackURL: "/dashboard?redirect=true",
});
```

I've tried adding it to the trusted providers
```ts
account: {
    accountLinking: {
      enabled: true,
      trustedProviders: ["roblox"],
    },
  },
```
however this still produces a type error when trying to pass "roblox" as a provider param to linkSocial, and a 400 error on the server.

maybe an officially supported roblox social provider too?

Bekacru:
closes #851

### DIFF
--- a/packages/better-auth/src/plugins/generic-oauth/index.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/index.ts
@@ -1,7 +1,7 @@
 import { betterFetch } from "@better-fetch/fetch";
 import { APIError } from "better-call";
 import { z } from "zod";
-import { createAuthEndpoint } from "../../api";
+import { createAuthEndpoint, sessionMiddleware } from "../../api";
 import { setSessionCookie } from "../../cookies";
 import {
 	createAuthorizationURL,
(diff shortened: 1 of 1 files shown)


arxk:
Since this pull is linked to the thread that was talking about Roblox, could we see support for providers that don't give an email (such as Roblox). This would be great.

PR #1213, opened 2025-01-14
refactor: supports facebook `access_token` and add docs

reslear:


reslear:
hi @Bekacru pls merge because, because I missed it last time including the documentation thanks )

Issue #1192, opened 2025-01-11
list-accounts phoneNumber userId mongodb Invalid id value


reslear:
### Is this suited for github?

- [X] Yes, this is suited for github

### To Reproduce

mongodb and phoneNumber plugin, 
run `listAccounts`

### Current vs. Expected behavior

I don't know exactly why it's happening, but I can guess:

if userId is ObjectId in mongodb: 

![CleanShot 2025-01-11 at 23 45 12@2x](https://github.com/user-attachments/assets/4041490f-e2aa-499b-80bc-5e8067638a24)

![CleanShot 2025-01-11 at 23 50 45@2x](https://github.com/user-attachments/assets/bbec98cc-0677-47e4-ba18-82ffbeb5219e)

`findAccounts` use `userId `:

https://github.com/better-auth/better-auth/blob/00b3d5a80ef588b35ca910f570fb2de0cc9ad45d/packages/better-auth/src/db/internal-adapter.ts#L701-L712

but `userId` not ObjectId only `id/_id`: 

https://github.com/better-auth/better-auth/blob/00b3d5a80ef588b35ca910f570fb2de0cc9ad45d/packages/better-auth/src/adapters/mongodb-adapter/mongodb-adapter.ts#L10-L12

### What version of Better Auth are you using?

1.1.9

### Provide environment information

```bash
- nodejs
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  database: mongodbAdapter(db),
  emailAndPassword: {  
    enabled: true
  },
  advanced: {
    generateId: false,
  },
  plugins: [
    phoneNumber()
  ]
});
```

reslear:
@Bekacru i see result `user` is clean, no userId

```ts

listUserAccounts {
  session: {
    ipAddress: '',
    userAgent: 'Mozilla/5.0 ',
    expiresAt: 2025-01-28T17:59:13.014Z,
    token: '7AioLt1g5RKTcGcStDS50NhsPNZyCKL6',
    createdAt: 2025-01-21T17:59:13.014Z,
    updatedAt: 2025-01-21T17:59:13.014Z
  },
  user: { createdAt: Invalid Date, updatedAt: Invalid Date }
}

```

![Image](https://github.com/user-attachments/assets/d6f75f2d-e0d3-432d-898f-634b1c0938f6)

in phoneNumber plugin i add log:

[...]

PR #1218, opened 2025-01-15
refactor: update the use of phoneNumber

smhmayboudi:
The use of opts.phoneNumber instead of "phoneNumber".


reslear:
~~maybe this fix https://github.com/better-auth/better-auth/issues/1192~~ don't fix

PR #1229, opened 2025-01-16
Update Svelte Example to Svelte 5

nevthereal:
The first of probably a few PRs. Thought, now that Svelte 5 is out we could update the syntax.


garikAsplund:
Obviously this is just one file and more on the docs side of things, but is there a Svelte 5 upgrade coming? I don't think I'd be interested in using for any of my projects until then unfortunately.

PR #1237, opened 2025-01-17
fix: pass token in reset-password

hyoban:
Because we no longer read the token from `currentURL`.

I've also made it a required parameter, feel free to remove it if you don't want it.


Bekacru:
Let's make it optional so we can release it as a patch. We'll consider making it required in the next minor release.

hyoban:
Updated to a simple doc change

Bekacru:
thanks!

PR #1238, opened 2025-01-18
feat: allow manually linking account with different email address than user

### LINKED ISSUE #918: `linkSocial` cannot link to social accounts with different email

### To Reproduce

1. Create a user
2. Link a social account with a different email
3. `error=email_doesn%27t_match` shown in URL

Logged-in users should be able to link to social accounts in different emails

1.0.20

```bash
System:
OS: macOS 15.2
CPU: (8) arm64 Apple M1 Pro
Memory: 120.31 MB / 16.00 GB
Shell: 5.9 - /bin/zsh
Binaries:
Node: 22.11.0 - ~/.n/bin/node
Yarn: 1.22.19 - ~/.n/bin/yarn
npm: 10.9.0 - ~/.n/bin/npm
pnpm: 9.15.0 - ~/.n/bin/pnpm
Watchman: 2024.12.02.00 - /opt/homebrew/bin/watchman
Browsers:
Safari: 18.2
```

Backend

```typescript
emailAndPassword: {
enabled: true
},
});
```

### Additional context

Afterward, users should be able to log in to accounts with different email addresses by binding social accounts without creating a new account.

D3visionNL:
closes #918 

This PR includes both the fix, a settings option to disable the email check, as well as an edit to the users-accounts documentation to display the settings option.

This is my first time ever creating a PR so please feel free to let me know if I need to improve or change something.

### DIFF
--- a/docs/content/docs/concepts/users-accounts.mdx
+++ b/docs/content/docs/concepts/users-accounts.mdx
@@ -324,6 +324,17 @@ Users already signed in can manually link their account to additional social pro
   });
   ```
 
+  If you want your users to be able to link a social account with a different email address than the user, or if you want to use a provider that does not return email addresses, you will need to disable email matching in the account linking settings. 
+  ```ts title="auth.ts"
+  const auth = new BetterAuth({
+      account: {
+          accountLinking: {
+              requireEmailMatch: false
+          }
+      },
+  });
+  ```
+
 - **Linking Credential-Based Accounts:** To link a credential-based account (e.g., email and password), users can initiate a "forgot password" flow, or you can call the `setPassword` method on the server. 
 
   ```ts
--- a/packages/better-auth/src/types/options.ts
+++ b/packages/better-auth/src/types/options.ts
@@ -412,6 +412,12 @@ export type BetterAuthOptions = {
 			trustedProviders?: Array<
 				LiteralUnion<SocialProviderList[number] | "email-password", string>
 			>;
+			/**
+			 * If disabled (false), this will allow users to manually linking accounts with different email addresses than the main user.
+			 * 
+			 * @default true
+			 */
+			requireEmailMatch?: boolean;
 		};
 	};
 	/**
--- a/packages/better-auth/src/api/routes/callback.ts
+++ b/packages/better-auth/src/api/routes/callback.ts
@@ -115,7 +115,7 @@ export const callbackOAuth = createAuthEndpoint(
 			);
 		}
 		if (link) {
-			if (link.email !== userInfo.email.toLowerCase()) {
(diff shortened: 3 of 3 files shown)


Bekacru:
The problem with linking different email accounts is that it can easily lead to account takeovers. If there’s an unattended session, someone could link it to a random account. Opting in might not be a bad idea, especially if it includes checking whether the session is fresh. So let's do opt in instead of opt out.

D3visionNL:
> The problem with linking different email accounts is that it can easily lead to account takeovers. If there’s an unattended session, someone could link it to a random account. Opting in might not be a bad idea, especially if it includes checking whether the session is fresh. So let's do opt in instead of opt out.

@Bekacru What exactly do you mean by opting in instead of opting out? The way it works in my PR is that they need to explicitly disable the email check and it still has the regular email check behaviour by default so wouldn't this already be an opt in feature?

Bekacru:
@D3visionNL sorry I think I missed `requireEmailMatch` being true by default. But let's rename it to `allowDifferentEmails` instead which should be false by default.

D3visionNL:
@Bekacru done!

docs/content/docs/concepts/users-accounts.mdx
```diff
@@ -324,6 +324,17 @@ Users already signed in can manually link their account to additional social pro
   });
   ```
 
+  If you want your users to be able to link a social account with a different email address than the user, or if you want to use a provider that does not return email addresses, you will need to enable this in the account linking settings. 
+  ```ts title="auth.ts"
+  const auth = new BetterAuth({
```
Bekacru: this should be `betterAuth({})` not `new BetterAuth`

docs/content/docs/concepts/users-accounts.mdx
```diff
@@ -324,6 +324,17 @@ Users already signed in can manually link their account to additional social pro
   });
   ```
 
+  If you want your users to be able to link a social account with a different email address than the user, or if you want to use a provider that does not return email addresses, you will need to enable this in the account linking settings.
```
Bekacru, line 327: If you want your users to link a social account with a different email address than their existing one, you’ll need to enable this in the account linking settings.

packages/better-auth/src/types/options.ts
```diff
@@ -412,6 +412,12 @@ export type BetterAuthOptions = {
 			trustedProviders?: Array<
 				LiteralUnion<SocialProviderList[number] | "email-password", string>
 			>;
+			/**
+			 * If enabled (true), this will allow users to manually linking accounts with different email addresses than the main user.
```
Bekacru, line 416: add here

⚠️ Warning: enabling this might lead to account takeovers, so proceed with caution.

Bekacru (changes requested):

D3visionNL:
@Bekacru done again lol

PR #1240, opened 2025-01-18
feat: disable default scopes and allow scopes to be passed on request

### LINKED ISSUE #1079: feat: request additional scopes from user

Hey! I'm currently working on an app that requires users to grant access to their GitHub repositories to function properly. I know additional scopes can be added to better-auth's configuration during user sign-up. However, users might not want to grant repository access on their first visit.

```ts

  appName: "Auth Test",
  socialProviders: {
    github: {
      clientId: process.env.GITHUB_CLIENT_ID!,
      clientSecret: process.env.GITHUB_CLIENT_SECRET!,
      scope: ["repo"], // additional scopes apart from which are used by default for retrieving common user information
    },
  },
});
```

I would like a flow where even if additional scopes are not written in the config, I should be able to request those post sign-up when needed. An example of this can be -

```ts
async function getGithubAccess() {
  await authClient.signIn.social(
    { provider: "github" },
    { scopes: ["repo"] }
  );
}

<button onClick={getGithubAccess}>Give access to your github repositories</button>
```

Also update the access token and refresh token in database.

Using Clerk, this can be achieved using the <UserButton /> component by passing `additionalOAuthScopes` prop, but it doesn’t offer much control. 

Source - https://clerk.com/docs/authentication/social-connections/oauth#request-additional-o-auth-scopes-after-sign-up

### Additional context

_No response_

Bekacru:
closes #1079

### DIFF
--- a/docs/content/docs/basic-usage.mdx
+++ b/docs/content/docs/basic-usage.mdx
@@ -169,6 +169,7 @@ To sign in using a social provider you need to call `signIn.social`. It takes an
 - **errorCallbackURL**: the URL to redirect to if an error occurs during the sign in process. (optional)
 - **newUserCallbackURL**: the URL to redirect to if the user is newly registered. (optional)
 - **disableRedirect**: a boolean to disable the automatic redirect to the provider. This is useful when you want to show for example a popup. (optional)
+- **scopes**: custom list of scopes to replace or additionally request.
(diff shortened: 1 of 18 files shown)


nahtnam:
Anything blocking this from merging?

PR #1255, opened 2025-01-21
feat: support mssql directly

Bekacru:
fix mssql related errors to fully support mssql databases directly through kysley


andreasbhansen:
**EDIT:**
This issue has been solved in https://github.com/better-auth/better-auth/commit/a8bfe2b2e51855b1bd5ad2158ff2c18ecd23633c and released to beta in https://github.com/better-auth/better-auth/commit/05d586ea58bac45abaa04cd99e4bd755acacb642.
____

@Bekacru I wonder if there's an issue with the `withReturning` statement:
```
async withReturning(values, builder, model, where) {
      let res;
      if (config?.type !== "mysql" && config?.type !== "mssql") {
        res = await builder.returningAll().executeTakeFirst();
      } else {
        await builder.execute();
        const field = values.id ? "id" : where[0].field ? where[0].field : "id";
        const value = values[field] || where[0].value;
        res = await db.selectFrom(getModelName(model)).selectAll().where(getField(model, field), "=", value).executeTakeFirst();
      }
      return res;
}
```

When I attempt a login using `authClient`, I get:
`TypeError TypeError: Cannot read properties of undefined (reading 'field')`.

Looking at the stack trace, it seems the KyselyAdapter's `create` method calls this:
```
return transformOutput(
    await withReturning(transformed, builder, model, []), // <--- Issue with [] ?
    model,
    select
);
```

... with an empty array, but `withReturning` attempts to read from the first element in the empty array (`where[0].field`).

Am I misunderstanding something here, or is this a bug?
I *think* that `.outputAll('inserted')` (see [docs](https://github.com/kysely-org/kysely/blob/660dfb43d5c48dd74d35f5fedae28f395df68fac/src/query-builder/insert-query-builder.ts#L821)) is the Kysely MSSQL equivalent of `.returningAll()`, in case it's more appropriate than the current implementation.

PR #1257, opened 2025-01-21
Added example for server side OAuth provider authentication in SvelteKit 

zicho:
Added example for server side OAuth provider authentication in SvelteKit using github and microsoft providers.


Kinfe123:
can you please nextjs intergation example for consistency sake and then i will merge it

zicho:
> can you please nextjs intergation example for consistency sake and then i will merge it

I do not have any experience working with next unfortunately. I think that would be best left to someone else.

PR #1267, opened 2025-01-22
feat: add databaseHooks for organizaton create & update

tomhhealy:
Hi - first contribution. Love the work you've put into Better Auth - what a fantastic resource!

In this PR, I've added before & after database hooks for the `Organization` plugin and updated the docs with a small snippet for a simple "log to console" when an Organization is created or updated. 
@Bekacru I can't seem to find the message from Discord now, but I believe you had plans to work on the organization plugin so I hope this fits in with those! 😃 

The reason behind adding this was because of custom logic I wanted to execute at the time of the user creating the organization and I've been using the before hooks elsewhere already.


shadiramadan:
Yes!  I so need this too!  @Bekacru can this be prioritized for the next milestone?

The only thing is @tomhhealy I think this change sort of breaks out of just being a plugin.  I think the plugin interface might have to first properly support database hooks.  This should also extend to the invitation system too.

Bekacru:
Hey @tomhhealy thanks for the PR! We'll review it and it should land on 1.2

shadiramadan:
To follow up on this database hooks for:
- organizations
- members
- invitations
etc. should all be possible.

gillesmag:
Hi @Bekacru, I would also be interested in having the database hooks for organization members land soon. Is there anything I can do to help?

Kinfe123:
may be creating a member specific hook on separate pr would help since org hook is already out there which you can take a look here - https://www.better-auth.com/docs/plugins/organization#organization-creation-hooks

PR #1275, opened 2025-01-24
feat: add cookie helper for middlewares

### LINKED ISSUE #1261: Update middleware docs to use cookies

_No response_

According to Next.js head of developer relations it is against best practices to perform any UI blocking or HTTP requests in middleware, and middleware should be used for optimistic routing whereas the actual route protection should happen on the route itself.

Use cookies

```ts
const protectedRoutes: string[] = ["/books"]

export async function middleware(request: NextRequest) {
    if (protectedRoutes.includes(request.nextUrl.pathname)) {
        if (!request.cookies.has("better-auth.session_token")) {
            return NextResponse.redirect(new URL("/", request.url))
        }
    }
```

It would be great to have better-auth auth.ts export a hasCookie function that checks for cookies so we don't have to hardcode the cookie name here.

### Additional context

_No response_

Bekacru:
closes #1261

### DIFF
--- a/demo/nextjs/middleware.ts
+++ b/demo/nextjs/middleware.ts
@@ -1,20 +1,9 @@
-import { betterFetch } from "@better-fetch/fetch";
 import { NextRequest, NextResponse } from "next/server";
-import type { Session } from "./lib/auth-types";
+import { getSessionCookie } from "better-auth";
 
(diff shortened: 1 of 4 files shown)


reslear:
cool this will also be useful for https://capacitorjs.com/

PR #1289, opened 2025-01-27
Use redirect URI from options rather than manually overriding

arlyon:
In some places in the codebase we allow the user to override the redirect URI. Sometimes this is respected and sometimes it is not. This change ensures that we handle it consistently, and that it is respected in the generic oauth plugin.

My concrete use case was setting up oauth proxy (the plugin didn't quite fit my use-case) and my generic oauth provider was not behaving.

### DIFF
--- a/packages/better-auth/src/social-providers/twitter.ts
+++ b/packages/better-auth/src/social-providers/twitter.ts
@@ -119,7 +119,7 @@ export const twitter = (options: TwitterOption) => {
 				code,
 				codeVerifier,
 				authentication: "basic",
-				redirectURI: options.redirectURI || redirectURI,
+				redirectURI,
 				options,
 				tokenEndpoint: "https://api.x.com/2/oauth2/token",
 			});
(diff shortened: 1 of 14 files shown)


packages/better-auth/src/social-providers/twitter.ts
```diff
@@ -119,7 +119,7 @@ export const twitter = (options: TwitterOption) => {
 				code,
 				codeVerifier,
 				authentication: "basic",
-				redirectURI: options.redirectURI || redirectURI,
```
Bekacru, line 122: I'm not sure why we're removing `options.redirectURI`. If the user manually provides the redirect URI, it should overwrite the default redirect URI passed when the function is called.

arlyon: `options.redirectURI` is not being removed. It is passed in in the `options` setting already, and that is handled now uniformly by `validateAuthorizationCode` by checking the options you pass in, rather than having to remember to do that at the call site in 10 different places.

PR #1290, opened 2025-01-27
Allow creating new orgs without changing the current active org

arlyon:
Does what it says on the tin. Adds a new flag—defaulted to `false`—which when set true will forgo changing the currently active org to the newly created org.

For an example of a use-case, I have a particular API that should be used when an organization creates another. This needs to be run from the parent org and switching the session back immediately after feels clunky.

EDIT: types all working. decided to not use `default` from zod

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-org.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-org.ts
@@ -35,6 +35,7 @@ export const createOrganization = createAuthEndpoint(
 					description: "The metadata of the organization",
 				})
 				.optional(),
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/organization/routes/crud-org.ts
```diff
@@ -35,6 +35,7 @@ export const createOrganization = createAuthEndpoint(
 					description: "The metadata of the organization",
 				})
 				.optional(),
+			keepCurrentActiveOrg: z.boolean().optional(),
```
Bekacru, line 38: can we add a description for it in the schema for open api

arlyon:
This option now exists.

PR #1291, opened 2025-01-27
feat: allow generating ids asynchronously

arlyon:
Many crypto APIs allow async access (for the purposes of gathering
entropy etc). This change allows optionally returning a Promise
instead meaning these async APIs are accessible.

For the particular use-case that motivated this change, nextjs + ppr + dynamicIO requires async crypto APIs to work correctly and since I am using KSUID for generating IDs, the sync API is no longer viable.


arlyon:
Rebased

PR #1301, opened 2025-01-29
docs(next): make server actions example more intuitive

ping-maxwell:


Kinfe123:
those server action calls could also be inlined and so i dont think types props would solve the intuitiveness

PR #1308, opened 2025-01-30
fix: return passkey data if valid

benbehringer:
This pull request modifies two conditional checks in the passkey client actions (signInPasskey and registerPasskey) so that the functions return the verified data when it exists instead of returning it only when there is no verified data. By switching from if (!verified.data) to if (verified.data), we ensure that a successful server response is propagated back to the caller.


Kinfe123:
we are actually indeep returning an error on such case

PR #1351, opened 2025-02-04
🐛 Fix Google IdToken Auth

abegehr:
"The value of iss in the ID token is equal to accounts.google.com or https://accounts.google.com."

Source: https://developers.google.com/identity/gsi/web/guides/verify-google-id-token

I ran into this when trying to auth via idToken obtained from auth within a Chrome Extension. According to the Google doc above the iss can also be "accounts.google.com", which was the case for the idToken I got.


abegehr:
I tested this change by overriding verifyIdToken on my better-auth google config object and it works!

PR #1359, opened 2025-02-05
fix(docs) session seconday storage value redis example

zohaibakber:
When using secondary storage, in this case redis the expected session value should be should be proper json but this example doesnt cover that hence getting this error whenever session is used.

>ERROR [Better Auth]: INTERNAL_SERVER_ERROR SyntaxError: "[object Object]" is not valid JSON

### DIFF
--- a/docs/content/docs/concepts/database.mdx
+++ b/docs/content/docs/concepts/database.mdx
@@ -234,7 +234,7 @@ export const auth = betterAuth({
 	secondaryStorage: {
 		get: async (key) => {
 			const value = await redis.get(key);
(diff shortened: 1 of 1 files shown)


izakfilmalter:
Thanks for this fix!

SpatzlHD:
where exactly does this error occure because adding the JSON.stringify methode causes the error:
`# SERVER_ERROR:  [TypeError: list.filter is not a function]`
also take a look at #1641 so it would be interesting if the implementation has to be somehow updated to account for both errors

PR #1366, opened 2025-02-06
fix(cli): wrong kysely schema filename

### LINKED ISSUE #1268: CLI `generate` command with `--output` has incorrect prompt.

### To Reproduce

When you use `--output` to specify an output location it works as expected. However the prompt for confirmation is not correct. It asks if you want to use the default location, even though a custom location was specified and is respected.

1. Run `npx @better-auth/cli generate --output ./schema/auth.sql`
  * Incorrect prompt shown: "Do you want to generate the schema to ./better-auth_migrations/2025-01-23T01-01-56.548Z.sql"
2. Accept the prompt
  * File is correctly created at `./schema/auth.sql` not `./better-auth_migrations/2025-01-23T01-01-56.548Z.sql`

with `--output ./schema/auth.sql`
**Current**: The prompt says "Do you want to generate the schema to ./better-auth_migrations/2025-01-23T01-01-56.548Z.sql"
**Expected**: The prompt says "Do you want to generate the schema to ./schema/auth.sql"

1.1.14

```bash
- OS: Mac OS
```

Other

```typescript

```

### Additional context

_No response_

alfanjauhari:
Change kysely generate schema file name confirmation based on output option

Previously if we fill the `--output` option, we still got a confirmation message with the better auth default file name instead of the `--output` that we input before.

Closes #1268

### DIFF
--- a/packages/cli/src/generators/kysely.ts
+++ b/packages/cli/src/generators/kysely.ts
@@ -1,13 +1,18 @@
 import { getMigrations } from "better-auth/db";
 import type { SchemaGenerator } from "./types";
 
-export const generateMigrations: SchemaGenerator = async ({ options }) => {
+export const generateMigrations: SchemaGenerator = async ({
+	options,
+	file,
+}) => {
 	const { compileMigrations } = await getMigrations(options);
 	const migrations = await compileMigrations();
 	return {
 		code: migrations,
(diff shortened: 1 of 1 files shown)


zekhoi:
tidak asing

PR #1379, opened 2025-02-07
docs: Update Elysia Integration Section

hansaskov:
I was in the process of adding Better-stack to my Elysia project and would like to improve the better-stack documentation using some of the more powerful features for creating a middleware with Elysia.

Key improvements include
- Simplified way to mount the handler
- More powerful middleware for handling requests
- User and Session are returned without null
- Three examples for how to use it

### DIFF
--- a/docs/content/docs/installation.mdx
+++ b/docs/content/docs/installation.mdx
@@ -337,7 +337,7 @@ Better Auth supports any backend framework with standard Request and Response ob
     ```
     </Tab>
     <Tab value="elysia">
-    ```ts
+    ```ts title="src/index.ts"
     import { Elysia } from "elysia";
     import { auth } from "./auth";
 
--- a/docs/content/docs/installation.mdx
+++ b/docs/content/docs/installation.mdx
@@ -338,20 +338,14 @@ Better Auth supports any backend framework with standard Request and Response ob
     </Tab>
     <Tab value="elysia">
     ```ts
-    import { Elysia, Context } from "elysia";
+    import { Elysia } from "elysia";
     import { auth } from "./auth";
 
-    const betterAuthView = (context: Context) => {
-        const BETTER_AUTH_ACCEPT_METHODS = ["POST", "GET"]
-        // validate request method
-        if(BETTER_AUTH_ACCEPT_METHODS.includes(context.request.method)) {
-            return auth.handler(context.request);
-        } else {
-            context.error(405)
-        }
-    }
+    export const betterAuthPlugin = new Elysia({ prefix: "/api/auth/*" })
+        .post("/", ({ request }) => auth.handler(request))
+        .get("/", ({ request }) => auth.handler(request));
 
-    const app = new Elysia().all("/api/auth/*", betterAuthView).listen(3000);
+    const app = new Elysia().use(betterAuthPlugin).listen(3000);
 
     console.log(
     `🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}`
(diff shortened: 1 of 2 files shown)


hansaskov:
Here is the full code fore reference. 

```ts 
import { Elysia, error } from "elysia"
import { auth } from "./auth/better-auth"

export const betterAuthPlugin = new Elysia({ prefix: "/api/auth/*" })
	.post("/", ({ request }) => auth.handler(request))
	.get("/", ({ request }) => auth.handler(request))

// Creates an isUser flag for subsequent routes.
// When enabled, the middleware is run and User and Session are available in the subsequent context
export const isUserMiddleware = new Elysia()
	.macro({
		isUser: {
			async resolve({ request }) {
				const session = await auth.api.getSession({ headers: request.headers })

				if (!session) {
					return error("Unauthorized") // Switch out with your own error handling.
				}

				return {
					user: session.user,
					session: session.session,
				}
			},
		},
	})
	.as("plugin")

const app = new Elysia()
	.use(betterAuthPlugin)
	// Enable the middleware
	.use(isUserMiddleware)
	// isUserMiddleware is false by default and won't run unless specified.
	.get("/", "Front page")
	// To enable it pass a third argument and set isUser to true. User and Session will now be present in the context fully typed
	.get("/user", ({ user }) => user, { isUser: true })
	// If you do not enable it, then you will get the type error: Property 'user' does not exist
	// .get("/user", ({ user }) => user) // Expect this to error!!
	// To change the default behavior you can set isUser to true for the subsequent routes
	.guard({ isUser: true })
	// User and Session will now be available in the context.
	.get("/user", ({ user }) => user)
	.listen(3000)

console.log(
	`🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}`
)


```

hansaskov:
Here is the full code for reference: 

````ts
import { Elysia, error } from "elysia"
import { auth } from "./auth/better-auth"

export const betterAuthPlugin = new Elysia({ prefix: "/api/auth/*" })
	.post("/", ({ request }) => auth.handler(request))
	.get("/", ({ request }) => auth.handler(request))

// Creates an isUser flag for subsequent routes.
// When enabled, the middleware is run and User and Session are available in the subsequent context
export const isUserMiddleware = new Elysia()
	.macro({
		isUser: {
			async resolve({ request }) {
				const session = await auth.api.getSession({ headers: request.headers })

				if (!session) {
					return error("Unauthorized") // Switch out with your own error handling.
				}

				return {
					user: session.user,
					session: session.session,
				}
			},
		},
	})
	.as("plugin")

const app = new Elysia()
	.use(betterAuthPlugin)
	// Enable the middleware
	.use(isUserMiddleware)
	// isUserMiddleware is false by default and won't run unless specified.
	.get("/", "Front page")
	// To enable it pass a third argument and set isUser to true. User and Session will now be present in the context fully typed
	.get("/user", ({ user }) => user, { isUser: true })
	// If you do not enable it, then you will get the type error: Property 'user' does not exist
	// .get("/user", ({ user }) => user) // Expect this to error!!
	// To change the default behavior you can set isUser to true for the subsequent routes
	.guard({ isUser: true })
	// User and Session will now be available in the context.
	.get("/user", ({ user }) => user)
	.listen(3000)

console.log(
	`🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}`
)


```

hansaskov:
@Bekacru Please say if there is something i need to do in order to help contribute 😊

Issue #1112, opened 2025-01-03
feature request: api-key plugin auth


reslear:
### Is this suited for github?

- [X] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

![image](https://github.com/user-attachments/assets/85a1ab58-e816-418b-b9e7-8b90730bd461)

![image](https://github.com/user-attachments/assets/fd0af524-2273-49fd-a940-a8e983142c09)

Api key functionality - (middleware, store, list ...), use database collection like `tokens` for storing, use admin plugin for management, add routes for CRUD operations. 

### Describe the solution you'd like

Example: https://github.com/hydra-newmedia/passport-headerapikey
or https://docs.strapi.io/dev-docs/configurations/api-tokens

it would also be cool to be able to have each user create a key for themselves 

### Describe alternatives you've considered

maybe use jwts for this solution

epulla:
I'd love to have this features ❤ having a one-time token/key for public APIs. Is there any plan for this feature request soon?

ping-maxwell:
https://github.com/better-auth/better-auth/pull/1380

reslear:
https://github.com/better-auth/better-auth/pull/1515

PR #1380, opened 2025-02-07
feat: API Key plugin

ping-maxwell:
The API Key plugin allows you to create and manage API keys for your application. It provides a way to authenticate and authorize API requests by verifying API keys.

# Docs
image

# Tests
image


reslear:
probaly closes https://github.com/better-auth/better-auth/issues/1112 ?
@ping-maxwell you can add to message

trojanowski:
@ping-maxwell I can see one potential security concern with the solution from this PR. API keys are sensitive credentials that can grant access to multiple actions, potentially including destructive ones, and should be treated with the same security considerations as passwords. But in the current implementation, these keys are stored in plain text in the database.

I suggest storing them in a hashed form instead. We could add an optional `start` field containing the first few characters in plain text to help users identify their keys in the UI. This approach is similar to what [Unkey implements](https://github.com/unkeyed/unkey/blob/28164ef0966ca1d8b0def96b11156924f1ecde78/internal/db/src/schema/keys.ts#L28-L29).

Since the keys are randomly generated with high entropy, a SHA-256 hash should be sufficient for security, potentially with a configurable salt.

Additionally, it would be valuable to implement scoping functionality for the keys to provide more granular access control.

ping-maxwell:
Yeah, me and @Bekacru have discussed this already.
There are multiple things that will be reworked, and Bekacru will be looking into reworking this.

Thanks for your concern, and taking time and looking into it though!

ping-maxwell:
> Since the keys are randomly generated with high entropy, a SHA-256 hash should be sufficient for security, potentially with a configurable salt.

Yeah, we'll be doing just that.
Also, there will be configurations on the plugin to allow users to have their own custom key generation code if they would like.

R5dan:
Few things I would like to see:
1) Custom "cost" - not all endpoints are equal
2) Allow a list of ratelimits - 10 per second but a max of 100 per minute
3) I don't think its clear which functions trigger adding to the endpoint - I believe `updateRateLimit` is the function that is called, but it probably shouldn't be called with the `list` call. Can there be a specific endpoint just for updating it
4) Let updating API keys have undefined bodies - these will just remain the same

ping-maxwell:
Please just wait until Bekacru refactors the code.
Your thoughts and questions about ratelimits, and anything else is pretty much already thought through already.
It's just not in the codebase yet.

Wait for Bekacru to implement it, and if you spot any additional things that needs to be added, feel free to mention then.

versecafe:
> > Since the keys are randomly generated with high entropy, a SHA-256 hash should be sufficient for security, potentially with a configurable salt.
> 
> Yeah, we'll be doing just that. Also, there will be configurations on the plugin to allow users to have their own custom key generation code if they would like.

This would be great as options like TypeIDs or plain prefixed API keys tend to make keys more usable for end consumers by allowing a consistent prefix ie `ant_*` for anthropic keys

ping-maxwell:
Sorry guys, I butchered this branch. 😅
Considering all the replanning/rework needed, I'll be opening a new PR for this.
I'll send a link here once the branch is open.

HeyGarrison:
This is a sweet feature! +1

ping-maxwell:
New PR for the reworked API key: https://github.com/better-auth/better-auth/pull/1515

PR #1388, opened 2025-02-08
fix(docs): code highlight fix

ryu1013-job:
The explanation for adding `passkeyClient()` to the plugins was covered in the previous chapter, so I believe that this chapter should focus on highlighting the part where it is actually used, namely `const data = await authClient.passkey.addPasskey();`.


Kinfe123:
it is already been fixed

Issue #778, opened 2024-12-05
[OAuth] Allow sign up with extra fields


chris-verclytte:
**Is your feature request related to a problem? Please describe.**
I currently have a use case that involve creating a user with a user type. 
When using the login/password flow, everything goes well, I can create a user with any field I'd like.
However, it's not the same when dealing with OAuth flow, I can't pass the user type along the flow.
My database does not allow for userType to be populated with a NULL value so create a user without type and editing afterwards is not a solution I want to consider. 

**Describe the solution you'd like**
The ideal solution would be to provide the extra field in the OAuth signin :
```js
// Inside a React form for instance
await authClient.signIn.social({
    provider: provider.id,
    data: {
    	userType: 'client'
    }
    callbackURL: `/dashboard`,
})
```
and then get this data provided to mapProfileToUser:
```js
google: {
    mapProfileToUser: (profile: GoogleProfile, data: UserExtraData) => ({
        email: profile.email,
        name: profile.name
        image: profile.picture
        type: data.userType,
    })
},
```
I think the extra data could be temporarily stored along with callbackURL, codeVerifier and expiresAt inside the verification table.

**Describe alternatives you've considered**
I've considered the following alternatives: 
1) Pass data as query params inside the callback and "catch it" before redirecting
2) Add a temporary cookie before saving and use it when saving the data to database (some users are currently doing it this way with next-auth/auth.js)
3) Add the data inside the state of the OAuth provider URL and then catch it inside callback (this currently breaks the flow as the state is verified when receiving the callback)

chris-verclytte:
Here's an idea of how it could look like: https://github.com/chris-verclytte/better-auth/pull/1.

And an example of implementation passing a user type to be used for OAuth user creation : 
```ts
// action.ts
export async function signInWrapper(
  providerId: 'google',
  userType: UserType,
) {
  // Validation to prevent user from adding any value in userType 

  return auth.api.signInSocial({
    body: {
      newUserCallbackUrl: '/dashboard/welcome',
      callbackURL: '/dashboard',
      provider: providerId,
      data: {
        userType,
      },
    },
  })
}
```

```ts
// auth.config.ts

// ...

export const auth = betterAuth({
 // ... other config params
  databaseHooks: {
    user: {
      create: {
        before: async ({ userType, ...user }) => {
          // Merge standard user info with extra data
          return { ...user, type: userType }
        },
      },
    },
  },

  socialProviders: {
   // ... other providers
    google: {
[...]

PR #1395, opened 2025-02-09
feat: allow extra data when registering with OAuth

chris-verclytte:
Provides a solution for #778.

# Use case
When registering a new user, one might wants to initialize a specific field - a good example would be providing different sign up forms for different user types. 
When using "authClient.signUp.email(...)", you could easily provide additional fields to be populated inside the user entity doing so: 
```js
authClient.signUp.email({
    email: "test@example.com",
    password: "password1234",
    name: "test",
    image: "https://example.com/image.png",
    type: 'USER_TYPE_1', // Here goes the extra field
})
```
 
However, if the system allows users to register with OAuth providers, one would needs the kind of additional data handling but the "authClient.signIn.social(...)" API does not allow to pass extra field when registering the user : 
```js
const data = await authClient.signIn.social({
    provider: "google",
    type: 'USER_TYPE_1', // <-- This won't work 
})
```

# Proposal
This PR proposal is to create an additionalData API on signIn.social so that we could also pass additional data when registering with OAuth:
```js
const data = await authClient.signIn.social({
    provider: "google",
	additionalData:  {
	    type: 'USER_TYPE_1',
	}
})
``` 
Then the data would become available inside `getUserInfo` hooks in order to pass it to the internal adapter creation function.
In the end the end-developer would just need to update Better Auth config file by using the additional field inside of getUserInfo : 

```js
// ... Rest of the config
socialProviders: {
    google: {
      clientId: <google_client_id>,
      clientSecret: <google_client_secret>,
      mapProfileToUser: (profile: GoogleProfile, additionalData: Record<string, unknown> = {}) => ({
        email: profile.email,
        name: `${profile.given_name} ${profile.family_name}`,
        emailVerified: profile.email_verified === true,
        image: profile.picture,
        ...additionalData, // <-- Here you can use additionalData as you want
      }),
    },
   // ... Repeat for every social provider you want to handle
  },
```

### DIFF
--- a/packages/better-auth/src/api/index.ts
+++ b/packages/better-auth/src/api/index.ts
@@ -97,7 +97,7 @@ export function getEndpoints<
 			.flat() || [];
 
 	const baseEndpoints = {
-		signInSocial,
+		signInSocial: signInSocial<Option>(),
 		callbackOAuth,
 		getSession: getSession<Option>(),
 		signOut,
--- a/packages/better-auth/src/api/routes/callback.ts
+++ b/packages/better-auth/src/api/routes/callback.ts
@@ -67,7 +67,7 @@ export const callbackOAuth = createAuthEndpoint(
 				`${c.context.baseURL}/error?error=oauth_provider_not_found`,
 			);
 		}
-		const { codeVerifier, callbackURL, link, errorURL, newUserURL } =
+		const { codeVerifier, callbackURL, link, errorURL, newUserURL, additionalData } =
 			await parseState(c);
 
 		let tokens: OAuth2Tokens;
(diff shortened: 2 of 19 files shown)


yamz8:
any updates here?

chris-verclytte:
@yamz8 I am waiting for a confirmation that this PR might find a place in some release to polish the PR.

jeffrey0606:
any update? 👀

jonasmerlin:
@chris-verclytte Thanks for implementing this! Would this work with the magic link plugin as well, or would that require extra work still? (I'm asking because I'm currently waiting until this is merged before switching to better auth, and would be open to contributing if there is work needed to make it work with magic links)

Bekacru:
I think if we decide to support this feature, we'll likely have `signUp.email`

Bekacru:
Thanks for the PR!

robinth91:
@Bekacru There is useful for the oauth flows where there is no easy way to pass custom data across the flow. So is it possible to reopen this?

robin-thomas:
@Bekacru There is useful for the oauth flows where there is no easy way to pass custom data across the flow. So is it possible to reopen this?

PR #1400, opened 2025-02-10
fix(cli): include default values when generating Drizzle schema

### LINKED ISSUE #1387: Schema generation via cli does not respect defaultValue while using Drizzle adapter

### To Reproduce

1. Create a plugin extending the user schema
2. Add a field which includes a defaultValue

Example:
```js
is_verified: {
            type: "boolean",
            required: true,
            unique: false,
            defaultValue: false,
          }
```

Following the steps from the previous action, the generated schema file should include a default value in the field. The output should look like this:

Image
But instead of that, we don't get the `.default(false)` part and the output is like this:
Image

Due to this error, we encounter the following message: `Failed to create user [error: null value in column "is_verified" of relation "user" violates not-null constraint']`

I noticed this issue previously in #1343 but I was unable to reopen it due to insufficient permissions. That is why I am creating this new issue.

1.1.16

```bash
- OS: Windows 11
- Browser: Chrome, Firefox (the issue is not about the browser)
```

Package

```typescript
import { username } from "better-auth/plugins";
import { emailHarmony } from "better-auth-harmony";

import { db } from "@/lib/db";
import { account, session, user, verification } from "@/lib/db/schema";
import { userExtenderPlugin } from "./plugins/extend-user";

    provider: "pg",
    schema: {
      user,
      session,
      verification,
      account,
    },
  }),
  emailAndPassword: {
    enabled: true, // If you want to use email and password auth
  },
  socialProviders: {
    google: {
    },
  },
  plugins: [username(), emailHarmony(), userExtenderPlugin()],
});
```

### Additional context

_No response_

Kumneger0:
closes #1387

### DIFF
--- a/docs/content/docs/integrations/svelte-kit.mdx
+++ b/docs/content/docs/integrations/svelte-kit.mdx
@@ -67,6 +67,103 @@ Some of the actions are reactive. The client use [nano-store](https://github.com
     </div>
 ```
 
+### Example: Getting User session and role in a layout 
+```ts title="+layout.server.ts"
+import type { LayoutServerLoad } from './$types';
+import { error, redirect } from '@sveltejs/kit';
+import { authClient } from '$lib/auth-client';
+
+
+export const load: LayoutServerLoad = async ({ request }) => {
+    console.log('Admin Layout Server load ');
+    const session = await authClient.getSession({
+        fetchOptions: {
+            headers: request.headers
+        }
+    });
+    if (!session.data) {
+        console.log('Admin : No session');
+        throw redirect(302, '/login');
+    }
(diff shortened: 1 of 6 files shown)


ardasoyturk:
@Bekacru hey, sorry for the mention, but this PR has been awaiting review for quite some time. I encountered this bug and created the linked issue. I haven't verified if it's resolved since I manually updated the schema instead of using the CLI. However, it is indeed a bug, and I would appreciate it if you could review this when you have a chance. Thanks.

Kinfe123 (approved):

Kumneger0:
Hey! Just noticed this issue has already been fixed in PR #2506. Closing this one now. 👋

PR #1402, opened 2025-02-10
feat: move to better call version 1

Bekacru:
This PR moves the entire internal system to Better Call v1 (which is currently in beta).
It brings significantly improved ts editor performance, along with other performance enhancements, bug fixes and cleanups.


ping-maxwell:
🔥

PR #1424, opened 2025-02-12
feat: add access control to admin plugin

jslno:
Hey,
The commited changes adds role based access control to the admin plugin, just like in the organizations plugin.

## Tests

image

<details>
<summary>
<h2>Docs</h2>
</summary>

</details>


packages/better-auth/src/plugins/admin/has-permission.ts
jslno, line 1: Could be moved to plugins/access with an accessControl instead of options parameter

packages/better-auth/src/plugins/admin/has-permission.ts
```diff
@@ -0,0 +1,21 @@
...
+	permission: {
+		[key: string]: string[];
+	};
+}) => {
+	const roles = input.role.split(",");
+	const acRoles = input.options.roles || defaultRoles;
+	for (const role of roles) {
+		const _role = acRoles[role as keyof typeof acRoles];
+		const result = _role?.authorize(input.permission);
+		if (result?.success) {
```
jslno, line 16: maybe add double negation?

Bekacru:
Thanks! This is really good. We've been working on an access control plugin for v1.2. It still needs some more consideration, so we're going to go with this.

jslno:
Tried to merge, but it resolved in more conflicts. I'll try again tomorrow.

PR #1445, opened 2025-02-14
Activate OTP Verification without Insecure skipVerificationOnEnable Option

GregorWedlich:
This PR updates the OTP verification flow so that OTP-based two-factor authentication works similarly to TOTP verification. Previously, the OTP endpoint checked whether twoFactorEnabled was already active. Since OTP is used for activation, that check could not succeed unless the insecure option skipVerificationOnEnable: true was used, which bypasses the verification process and lowers the overall security level.

With these changes, upon correct OTP submission, the user’s twoFactorEnabled status is updated to true, and the session is refreshed accordingly.


Rakhsan:
yes this was the pr I was looking for. I ma wait till this is merged cuz shit is broken as hell right now. I tried 2FA OTP and it just failed

Issue #1415, opened 2025-02-11
No Error on Google One Tap


itisyb:
### Is this suited for github?

- [ ] Yes, this is suited for github

### To Reproduce

<img width="417" alt="Image" src="https://github.com/user-attachments/assets/c373fbb3-8b24-441a-9374-4b60583c646e" />

After one tap apear I click continue

I don't see any network request after clicking continue as "Yash"

### Current vs. Expected behavior

Nothing happens and I don't see any logs.
Is there a way to figure out the logs.

### What version of Better Auth are you using?

1.1.15

### Provide environment information

```bash
- Mac chrome browser
```

### Which area(s) are affected? (Select all that apply)

Other

### Auth config (if applicable)

```typescript
import { createAuthClient } from "better-auth/svelte"
import { oneTapClient } from "better-auth/client/plugins"

export const authClient = createAuthClient({
    baseURL: "http://localhost:5173/api/auth",
    plugins: [
        oneTapClient({
            clientId: "my id"
        })
    ],
})
```

### Additional context

Again I just need to see how can I see the errors

Bekacru:
Open your browser console, there might be an error there. If it's from Better Auth, you should see an error either in your server logs or in the network request tab.

itisyb:
Yes @Bekacru I opened Chrome browse, arc browser
Same for both

itisyb:
https://www.loom.com/share/19ea2ea988af49daa02ba292af895a4b?sid=be6243c0-9a15-400b-97d5-f032da147d00

itisyb:
Forgot to mention
normal google button flow is working same google id

Bekacru:
It seems like Better Auth is returning an error. When you click the button, make sure you're on the Network tab so you can see exactly what the error is.

itisyb:
Nope
@Bekacru ,
https://www.loom.com/share/b945824cd2d84a2397df61474273a2f9?sid=3e5798f1-6313-47ec-9a33-6a92254c57ff

itisyb:
Btw how does google onetap works?

when i used this few months ago in vanilla js

I had to setup up a path/endpoint in my server(similar to normal google)

but in this I didn't do any such thing cause I was using doc am I missing something?

itisyb:
Hi @Bekacru ,
if you could tell me the last part I think I will figure it out sorry to bother you

Bekacru:
[...]

PR #1452, opened 2025-02-15
feat(one-tap):  improve Google One Tap integration with JWT verification and improved prompt handling

Bekacru:
address issues raised by #1415 thread including handling jwt verification using jwks, using prompts with exponential back off and doc improvements

### DIFF
--- a/demo/nextjs/components/one-tap.tsx
+++ b/demo/nextjs/components/one-tap.tsx
@@ -1,0 +1,178 @@
+"use client";
+
+import { client, signIn } from "@/lib/auth-client";
+import { useEffect, useState } from "react";
+import {
+	Dialog,
+	DialogContent,
+	DialogDescription,
+	DialogHeader,
+	DialogTitle,
+} from "./ui/dialog";
+import { Input } from "./ui/input";
+import { useRouter } from "next/navigation";
+import Link from "next/link";
+import { PasswordInput } from "./ui/password-input";
+import { Checkbox } from "./ui/checkbox";
+import { Button } from "./ui/button";
+import { Key, Loader2 } from "lucide-react";
+import { toast } from "sonner";
+import { Label } from "./ui/label";
+
+export function OneTap() {
+	const [isOpen, setIsOpen] = useState(false);
+	useEffect(() => {
+		client.oneTap({
+			onPromptNotification(notification) {
+				setIsOpen(true);
+			},
+		});
+	}, []);
+	return (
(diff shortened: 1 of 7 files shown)


itisyb:
Thanks for the quick turnaround man @Bekacru 
but I am Still not seeing errror 😭

|─ better-auth@1.1.19-beta.2

itisyb:
Seems same
Also I noticed people are really struggling with this https://www.reddit.com/r/sveltejs/comments/1ipu32z/comment/md4hzwd/?context=3
Which is weirdly simple in vanilla 

It looks like the onTapClient only UI is initialise no other call is made as I tried to log that window.google.....
in svelte atleast

itisyb:
Hi @Bekacru ,Btw it also didn't solve 
So made it custom 

which throwed me error of origin mismatch
and someone in stackoverflow mention to also add 
(http://localhost:[port]/) and without http://localhost without port( I only had with port)


which solved my issue

and then I removed my custom code 
and it worked in Better-auth too 🥳


I would also like to share my spaghetti code which at least threw an error which I was looking for since start that there's origin mismatch which better auth didn't 😢 

Hopefully you will add that error thing. 


Here's my code through which I found the issue
```
return new Promise<void>((resolve, reject) => {
  console.log('[loadGoogleScript] Starting to load Google script');
  
  if (window.google) {
  console.log('[loadGoogleScript] Script already initialized');
  resolve();
  return;
  }
  
  const script = document.createElement('script');
  script.src = 'https://accounts.google.com/gsi/client';
  script.async = true;
  
  script.onload = () => {
  console.log('[loadGoogleScript] Script loaded successfully');
  resolve();
  };
  
  script.onerror = (error) => {
  console.error('[loadGoogleScript] Error loading script:', error);
  reject(error);
  };
  
  document.head.appendChild(script);
  });
  };
  
  function handleCredentialResponse(response: any) {
  console.log('Google response received:', response);
  // Here you can handle the response by sending it to your backend
  // For now, we'll just log it
  }
  
  async function initializeGoogle() {
  try {
  await loadGoogleScript();
  
  window.google.accounts.id.initialize({
  client_id: 'ClientID',
  callback: handleCredentialResponse,
  auto_select: false,
  cancel_on_tap_outside: true,
  context: 'signin'
  });
  
  // Initialize One Tap prompt
  window.google.accounts.id.prompt();
  } catch (error) {
  console.error('Failed to initialize Google Sign-In:', error);
  }
  }
  initializeGoogle();
```
  
  
  Also if you could add that localhost thing in doc it would be helpful for others 😄

PR #1460, opened 2025-02-16
Transform sqlite values in Drizzle adapter.

RyanNieuwoudt:
Not sure whether this is considered a fix or a new feature, depending on whether better-auth officially supports sqlite via `drizzle-adapter`. In this configuration, authentication does not work. Inserting into verifications currently fails with:

> D1_TYPE_ERROR: Type 'object' not supported for value…

The the dates are passed as objects (not supported in D1) instead of strings ([supported in D1](https://developers.cloudflare.com/d1/worker-api/#type-conversion)).

This pull request resolves the issue, and perhaps enables `sqlite` via drizzle-adapter more generally. It is based on existing transformations in [kysely-adapter](https://github.com/better-auth/better-auth/blob/03fb6b2dfaf559011863e4fb603ed175b3b0af84/packages/better-auth/src/adapters/kysely-adapter/kysely-adapter.ts#L33), and is intended to match existing code.

### DIFF
--- a/packages/better-auth/src/adapters/drizzle-adapter/drizzle-adapter.ts
+++ b/packages/better-auth/src/adapters/drizzle-adapter/drizzle-adapter.ts
@@ -24,6 +24,40 @@ const createTransform = (
 		return f.fieldName || field;
 	}
 
+	function transformValueToDB(value: any, model: string, field: string) {
+		const { provider = "sqlite" } = config || {};
+		const f = schema[model].fields[field];
+		if (
+			f.type === "boolean" &&
+			(provider === "sqlite" ) &&
+			value !== null &&
+			value !== undefined
+		) {
+			return value ? 1 : 0;
+		}
+		if (f.type === "date" && value && value instanceof Date) {
+			return provider === "sqlite" ? value.toISOString() : value;
+		}
+		return value;
+	}
+
+	function transformValueFromDB(value: any, model: string, field: string) {
+		const { provider = "sqlite" } = config || {};
+
+		const f = schema[model].fields[field];
+		if (
+			f.type === "boolean" &&
+			(provider === "sqlite") &&
+			value !== null
+		) {
+			return value === 1;
+		}
+		if (f.type === "date" && value) {
+			return new Date(value);
+		}
+		return value;
+	}
+
 	function getSchema(modelName: string) {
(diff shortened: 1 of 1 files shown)


matschik:
Without this PR, it's impossible to make Drizzle + D1 working.

I got exactly the same issue `Error: D1_TYPE_ERROR: Type 'object' not supported for value 'Sun Mar 23 2025 01:12:09 GMT+0100 (Central European Standard Time)'`.

I can confirm it's working well using Kysely but I want to use Drizzle. So I'm using Kysely until this PR is merged.

Bekacru:
this should work without this PR. Make sure date fields have mode timpestamp

```ts
createdAt: integer("created_at", { mode: "timestamp" }).notNull(),
```

RyanNieuwoudt:
> this should work without this PR. Make sure date fields have mode timpestamp

Thanks, that [aligns with Drizzle's documentation on dates in sqlite](https://orm.drizzle.team/docs/guides/timestamp-default-value#sqlite), particularly the second example using unix timestamps. In the first example, dates can be defined as [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) strings, eg:

```
timestamp: text('timestamp')
    .notNull()
    .default(sql`(current_timestamp)`),
```

This PR aims to support [ISO-8601 dates in sqlite](https://sqlite.org/lang_datefunc.html) via Drizzle. What would be the next step towards getting this PR ready to merge?

(While I had tried to remain close to existing code in the kysely-adapter, the simplifications you made are a nice improvement.)

Bekacru:
@RyanNieuwoudt I think Drizzle would fail if you try to insert ISO string dates when it's expecting a Date object. Unless we require the user to provide a flag, or try to infer the field type from the schema, we can only enforce one or the other. And we made a lot of changes to the adapter API see this PR #2036

Bekacru:
sorry but closing this. feel free to open another PR with the new API :)

RyanNieuwoudt:
Ok. I considered putting up another PR, but  have decided to use timestamp dates for the better-auth tables in my system, even though I'm using ISO8601 dates elsewhere. This has the advantage of being the default schema expected by this library, and allows me to remove `disableCleanup` from the verifications model.

For others following this PR, hopefully this is a viable option for you too.

PR #1472, opened 2025-02-17
Add option to set a function to generate OTPs

CordlessWool:
The idea is to allow more than numbers, to reduce the length and still have a high level of security. This also gives more freedom by defining the letters. Could be hex or something else.

In my concrete need I want to have a string number combination without zero and O or other letters too similar to a number and I want to have a shorter OTP with higher security.

### DIFF
--- a/packages/better-auth/src/plugins/email-otp/index.ts
+++ b/packages/better-auth/src/plugins/email-otp/index.ts
@@ -6,740 +6,740 @@ import { getDate } from "../../utils/date";
 import { setSessionCookie } from "../../cookies";
 
 export interface EmailOTPOptions {
-  /**
-   * Function to send email verification
-   */
-  sendVerificationOTP: (
-    data: {
-      email: string;
-      otp: string;
-      type: "sign-in" | "email-verification" | "forget-password";
-    },
-    request?: Request,
-  ) => Promise<void>;
-  /**
-   * Length of the OTP
-   *
-   * @default 6
-   */
-  otpLength?: number;
-  /**
-   * Expiry time of the OTP in seconds
-   *
-   * @default 300 (5 minutes)
-   */
-  expiresIn?: number;
-  /**
-   * Function to create OTP
-   */
-  createOTP?: (
-    data: {
-      email: string;
-      length: number;
-      type: "sign-in" | "email-verification" | "forget-password";
-    },
-    request?: Request,
-  ) => string;
-  /**
-   * Send email verification on sign-up
-   *
-   * @Default false
-   */
-  sendVerificationOnSignUp?: boolean;
-  /**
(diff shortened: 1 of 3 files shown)


packages/better-auth/src/plugins/email-otp/index.ts
```diff
@@ -29,6 +29,17 @@ export interface EmailOTPOptions {
 	 * @default 300 (5 minutes)
 	 */
 	expiresIn?: number;
+	/**
+	 * Function to create OTP
+	 */
+	createOTP?: (
```
Bekacru: let's rename this to `generateOTP` instead

packages/better-auth/src/plugins/email-otp/index.ts
```diff
@@ -29,6 +29,17 @@ export interface EmailOTPOptions {
...
+		data: {
+			email: string;
+			length: number;
```
Bekacru: I'm not sure why length is passed here. The user doesn't need to know the length to generate a custom OTP since they define it themselves or it falls back to the default.

Bekacru (changes requested):

packages/better-auth/src/plugins/email-otp/index.ts
CordlessWool: Because of the original function. I also thought about this point and I can also directly set the length in the default function

packages/better-auth/src/plugins/email-otp/index.ts
```diff
@@ -49,9 +59,9 @@ const types = ["email-verification", "sign-in", "forget-password"] as const;
 export const emailOTP = (options: EmailOTPOptions) => {
 	const opts = {
 		expiresIn: 5 * 60,
-		otpLength: 6,
+		generateOTP: () => generateRandomString(options.otpLength ?? 6, "0-9"),
```
CordlessWool, line 62: this here would be the alternative to remove length (change is in the last commit)

Maqed:
Why shouldn't we make this feature available for every plugin that has OTP?

CordlessWool:
> Why shouldn't we make this feature available for every plugin that has OTP?

You mean for email verification, passwort reset and sign in or what do you mean with plugins?
At least for all of them it is the same principle, I wanna have a shorter key with a higher safety. At least developer have different requirement and this gives more flexible. About my case I already wrote in the pr description. At all it is more flexibility and in my opinion it is the same for password reset, sign in or email verification, isn't it?

If you meant something other with plugins, please give me more details :)

CordlessWool:
@Bekacru are all change resquest clarified or there still something open I missed?

Bekacru (approved):

Bekacru:
thanks for the PR!

PR #1482, opened 2025-02-18
feat: add Trakt social provider

alessandrojean:
While I was working on #1481, I thought: why not add [Trakt](https://trakt.tv) as a built-in provider as well?

Not sure if it is a very niche use-case or if it's not in the library's plan to add more providers, so feel free to close if it's the case, no worries.

The API OAuth docs are available [here](https://trakt.docs.apiary.io/#reference/authentication-oauth).


Bekacru:
We try to avoid niche providers so they can be implemented with the generic oauth plugin instead. Thanks for making this PR!

PR #1485, opened 2025-02-18
feat(docs): add search documentation in mobile

Maqed:
There was no search in the mobile view, So I added it as its non-existence compromised UX
image


Maqed:
@Bekacru @Kinfe123 What is the linter that better auth works with? Because I just use prettier but it seems that my configs for it doesn't work.

Kinfe123:
Just run `pnpm lint:fix` after you have done with changes and push it. It uses biome under the hood.

Maqed:
Please just Ignore this pull request. Check [moving search documentation box to main navbar PR](https://github.com/better-auth/better-auth/pull/1536) instead.

PR #1498, opened 2025-02-19
fix: userId should be unique on 2fa plugin

t3duk:
When using the Prisma DB adaptor (haven't tested others) when trying to generate new backup codes (when 2fa is enabled) it fails. This is because when running `update` (instead of `updateMany`) a unique field is required in the `where`.

This works for me, but do let me know if any issues come of this as I would of suspected someone would of already reported this / fixed it before me as it is a pretty core functionality of the 2fa plugin.

### DIFF
--- a/packages/better-auth/src/plugins/two-factor/schema.ts
+++ b/packages/better-auth/src/plugins/two-factor/schema.ts
@@ -26,6 +26,7 @@ export const schema = {
 			userId: {
 				type: "string",
 				required: true,
+				unique: true,
 				returned: false,
 				references: {
 					model: "user",


tedbrine:
Not entirely sure how you guys do schema changes here so lmk what other files need changing aside from the `schema.ts` to get generation tests passing

GregorWedlich:
I use Prisma with Postgre and i can generate new backup codes.

Better-Auth 1.1.18

tedbrine:
@GregorWedlich What does your schema look like?

GregorWedlich:
```prisma
model User {
  id            String    @id @default(cuid())
  name          String
  email         String    @unique
  emailVerified Boolean
  image         String?
  createdAt     DateTime  @default(now())
  updatedAt     DateTime  @updatedAt
  sessions      Session[]
  accounts      Account[]
  ...
  twoFactorEnabled Boolean   @default(false)
  ...

  @@map("user")
  TwoFactor TwoFactor[]
}

model TwoFactor {
  id          String   @id @default(cuid())
  userId      String   @unique
  secret      String?  
  backupCodes String?
  
  user        User     @relation(fields: [userId], references: [id], onDelete: Cascade)

  @@map("twoFactor")
}
```

tedbrine:
See, your userId is set as unique on your schema.
I wonder why mine is not...

tedbrine:
I think this is a case by case thing. I'll close it, and if anyone has a similar problem they can fix it by adding `@unique` to their schema.

PR #1515, opened 2025-02-20
Feat: API Key plugin (rework)

ping-maxwell:
# API Key Plugin

image

## Core features:

- Allows you to create, update, get, list & delete API keys.
- Allows you to to pass an API Key in the header of an API call, and it will mock a session out of the user who created that API key. (configurable in plugin options)
- Rate-limiting built-in.
  - Custom rate-limits can be applied per API key, which only goes into effect when an API key is being verified.
  - Everything else can use the built-in Better Auth rate-limiting functionality to rate-limit endpoints.
- API Key `remaining` count for limited (or optionally unlimited) verification usage.
- API Key refilling functionality to refill the `remaining` count over a period of time. (useful for subscription use-cases)
- Optionally enable or disable a key.
- Allows you to pass additional metadata information for each key. 
- Allows you to pass a prefix for the API key generation.
- Other stuff that I probably forgot. 😁

## Tests

<details>

<summary>screenshot of it all</summary>

Almost 3x more tests than the old PR. 😅

image

</details>

<details>

<summary>Create</summary>

- ✅ create api key client without headers (should fail)
- ✅ create api key client with headers
- ✅ create api key server without headers (should fail)
- ✅ create api key server with headers
- ✅ create api key with custom name
- ✅ create api key with custom name that's smaller than allowed minimum (should fail)
- ✅ create api key with custom name that's larger than the allowed maximum (should fail)
- ✅ create api key with custom prefix
- ✅ create api key with custom prefix that's smaller than allowed minimum (should fail)
- ✅ create api key with custom prefix that's larger than the allowed maximum (should fail)
- ✅ create api key with custom expiresIn
- ✅ should fail to create key with custom expiresin when `customExpiresTime` is enabled
- ✅ create api key with custom expiresIn that's smaller than the allowed minimum (should fail)
- ✅ create api key with custom expiresIn that's larger than the allowed maximum (should fail)
- ✅ create api key with custom refillAndAmount from client auth (should fail because this is a server-only-property)
- ✅ create api key with custom refillInterval (should fail because it requires refillAmount as well)
- ✅ create api key with custom refillAmount (should fail because refillInterval is required)
- ✅ create api key with custom refillInterval & refillAmount
- ✅ create api key with custom remaining
- ✅ create api key with custom remaining that's smaller than allowed minimum (should fail)
- ✅ create api key with custom remaining that's larger than the allowed maximum (should fail)
- ✅ create api key with custom invalid metadata (should fail)
- ✅ create api key with custom valid metadata
- ✅ create api key's returned metadata should be an object
- ✅ create api key with with metadata when metadata is disabled (should fail)
- ✅ created api key's start property should be valid
- ✅ created api key's start property should be null if `shouldStore` is false
- ✅ created api key's start property should be the first `charactersLength` characters of the hashed key
- ✅ create api key with custom rate limit options from client auth (should fail because this is a server-only-property)
- ✅ create api key with custom rate limit options should apply successfully

</details>

<details>

<summary>Verify</summary>

- ✅ verify api key without headers (should fail)
- ✅ verify api key with headers
- ✅ verify api key with invalid key (should fail)
- ✅ verify api key 10 times in a row (should fail to rate-limits)
- ✅ should allow us to verify api key after rate-limit window has passed
- ✅ should check if verifying an api key's remaining count does go down
- ✅ should fail if the api key has no remaining
- ✅ Api key should fail if it's expired

</details>

<details>

<summary>Update</summary>

- ✅ update api key name without headers (should fail)
- ✅ update api key name with headers
- ✅ should fail to update api key name with a length larger than the allowed maximum
- ✅ should fail to update api key name with a length smaller than the allowed minimum
- ✅ should fail to update api key with no values to update
- ✅ update api key expiresIn value
- ✅ should fail to update expiresIn value if `disableCustomExpiresTime` is enabled
- ✅ update api key expiresIn value that's smaller than allowed minimum (should fail)
- ✅ update api key expiresIn value that's larger than the allowed maximum (should fail)
- ✅ update api key remaining value
- ✅ update api key remaining value that's smaller than allowed minimum (should fail)
- ✅ update api key remaining value that's larger than the allowed maximum (should fail)
- ✅ update api key refillInterval value (should fail since it requires refillAmount as well)
- ✅ update api key refillAmount value (should fail since it requires refillInterval as well)
- ✅ update api key refillInterval & refillAmount value
- ✅ update api key enabled value
- ✅ update api key with invalid metadata value
- ✅ update api key with valid metadata value
- ✅ update api key's returned metadata should be an object

</details>

<details>

<summary>Delete</summary>

- ✅ delete api key without headers (should fail)
- ✅ delete api key with headers
- ✅ delete api key that doesn't exist (should fail)

</details>

<details>
<summary>Get</summary>

- ✅ get api key by ID without headers (should fail)
- ✅ get api key by ID with headers
- ✅ get api key that doesn't exist (should fail)
- ✅ get api key's returned metadata should be an object
- ✅ get api key's returned key should not be defined

</details>

<details>
<summary>List</summary>

- ✅ list api keys without headers (should fail)
- ✅ list api keys with headers
- ✅ list api keys should contain metadata that are either objects or null

</details>

<details>
<summary>Sessions</summary>

- ✅ When making any request, if headers include `x-api-key`, then the the request should generate a mock session object for the user.
- ✅ Custom headers list should still work when calling getSession
- ✅ Custom key getter function should work

</details>

## Closes
https://github.com/better-auth/better-auth/issues/1112

## What's left? (todo)

- ✅ Update documentation: done
- ❌ Update nextjs/demo: Not started
- ❌ Add openAPI metadata: Not started


R5dan:
Few additional features I would like:
- Multiple rate limits
- Separate rate limit for IPs - might be implemented already with standard rate limits
- Different locations for API keys e.g. query params
- Different names for API key
- Different tiers of API key - make it easier to have different tiers, with different access and rate limits. I want to be able to apply it to a key / create a key from it
- Different values / costs for specific routes
- Able to block certain routes to different tiers

Happy to implement these here or in another PR

ping-maxwell:
> Few additional features I would like:
> 
> * Multiple rate limits
> * Separate rate limit for IPs - might be implemented already with standard rate limits
> * Different locations for API keys e.g. query params
> * Different names for API key
> * Different tiers of API key - make it easier to have different tiers, with different access and rate limits. I want to be able to apply it to a key / create a key from it
> * Different values / costs for specific routes
> * Able to block certain routes to different tiers
> 
> Happy to implement these here or in another PR

* Multiple rate limits
You can define rate-limits custom & unique to each api key you create. 

* Separate rate limit for IPs - might be implemented already with standard rate limits
Yeah.

* Different locations for API keys e.g. query params
I assume you mean for the session system, where it mocks a session based on a given API key in the headers.
You can configure this by using the plugin configs `customAPIKeyGetter`.
The type looks like this:
```ts
customAPIKeyGetter?: (ctx: GenericEndpointContext) => string | null;
```
You can create this function, and use the context.request to grab the needed info to get the API key.
So, for the use-case of query params, you would just grab the query params out of the request, and return the API key.

* Different names for API key
Yeah, you can already do this.

* Different tiers of API key - make it easier to have different tiers, with different access and rate limits. I want to be able to apply it to a key / create a key from it
I recommend building your own system on top of this plugin to achieve this.
We allow you to pass `metadata` to each key, so you can probably utilize this to gather information about what "tire" a given API key is. And of course, you can apply the rate-limits to each key uniquely.

* Different values / costs for specific routes
You can customize the rate-limits for both the plugin's specific rate-limits (which only apply to verification), but also the [better-auth instance's rate-limit configs](https://www.better-auth.com/docs/concepts/rate-limit).

* Able to block certain routes to different tiers
I recommend building your own system for this.
You can use a server endpoint to receive the needed info to either create, update, delete, or any action for any API keys, so from here, you should write your own logic for the tires system you have in mind. You can check your key "tires" by it's metadata.

If you have any more questions or requests, LMK 👍

R5dan:
Can I ask why we cant have tiers? Seems like quite a useful and simple feature to add, although I might be missing something.

florianmartens:
Here's my 2ct. to @R5dan 

> Different tiers of API key - make it easier to have different tiers, with different access and rate limits. I want to be able to apply it to a key / create a key from it

I think a library should be general purpose. Having API-Keys that are tiered is not a common pattern. Usually API keys have certain scopes/permissions/entitlements or whatever you call it in your company. A library like `better-auth` should not be opinionated about how you give access to your product.

In terms of rate limits: Here I'm also not quite on your side. Rate limits are different to the way you decide to provision your product. It seems like you want to use rate limiting to say "User on tier A gets 500 requests". This should not be part of an auth solution. It seems you are looking for a general solution that helps you provision your product. Usually a rate limiter is there to prevent abuse and single users from crashing your system. It is not part of your provisioning system.

R5dan:
> Usually API keys have certain scopes/permissions/entitlements or whatever

This is what I meant by tiers, and I think that this is quite common, especially where the API is the product. Not where it is linear but where you can have roles have complete separate permissions or they can have the same

> A library like `better-auth` should not be opinionated about how you give access to your product.

I don't think I'm asking it to be. I'm asking for an optional plugin to gain an optional feature which other users can ignore if they want. Although another option for what I'm asking for is the metadata in this case. I just thought it would be a useful option.

> Rate limits are different to the way you decide to provision your product

I would agree most of the time, but there are cases where an upgrade does come with additional rates, for example a free tier, which is easy to get is more heavily rate limited than a premium tier paying more, as there is less risk of abuse by the paying users. Additionally an admin / dev could have unlimited rate limits for testing etc.

Plus this is all possible to be done already, just applying it to each key individually where in many cases it would be easier to apply a role to a key, and then add / remove permissions, or disable it temporarily.

Although thank you @florianmartens as although I disagree, I do see some of your points.

> * Different values / costs for specific routes
>   You can customize the rate-limits for both the plugin's specific rate-limits (which only apply to verification), but also the [better-auth instance's rate-limit configs](https://www.better-auth.com/docs/concepts/rate-limit).

Sorry, I think you misunderstood me. I want to have 1 route "cost" 1 "rate limit currency" but another cost 10, as one might be more intensive to run.

ping-maxwell:
> > Usually API keys have certain scopes/permissions/entitlements or whatever
> 
> This is what I meant by tiers, and I think that this is quite common, especially where the API is the product. Not where it is linear but where you can have roles have complete separate permissions or they can have the same

I understand you now. Yeah, there will be a permission system. @Bekacru will handle this as he needs to rework access-control across all of Better-Auth.

> I would agree most of the time, but there are cases where an upgrade does come with additional rates, for example a free tier, which is easy to get is more heavily rate limited than a premium tier paying more, as there is less risk of abuse by the paying users. Additionally an admin / dev could have unlimited rate limits for testing etc.

There are endpoints/methods to update the API keys rate-limits, so if a user upgrades, you can update their API key rate-limits accordingly.

Also, I want to correct an idea which you guys may not be informed on, regarding:
>  It seems like you want to use rate limiting to say "User on tier A gets 500 requests". This should not be part of an auth solution. 

While we do have a rate-limit system, there is **also** a refill system, let me explain:
Each api key can optionally have an assigned `remaining` count. If the `remaining` hits 0, then the key is revoked/expired.
When a user verifies the key, the `remaining` count goes down.
A key can also have `refillAmount` and `refillInterval` values.
These values, as the names would suggest, refill the `remaining` count when the interval is passed.
So for a system like "User on tier A gets 500 requests", it's totally possible to implement, plus the refill system over time.

R5dan:
Thank you for clearing that up. I did think that access control should be changed, but couldn't work out exactly what bothered me, so good to hear its getting updated.

Is there going to be a way to refill above the standard amount? Without that being the new default to add. So if a user gets additional credits they can use them but it won't set it to refill that many every time?

Bekacru:
@R5dan yeah you should be able to change the remaining without affecting the refill amount

ping-maxwell:
Alright, I believe the API Key plugin is ready. 🫡

perevezencev:
Please add documentation for integration with elysia 🙏

ping-maxwell:
You shouldn't need any spcial docs for specific integrations.
Every plugin API works the same conceptually.
The only difference across framework is how you set up Better Auth.

There might be something I'm missing, could you elaborate further? @perevezencev

PR #1519, opened 2025-02-20
Don’t save email address as user name

benkingcode:
Currently the email OTP plugin saves the user's email address as their "name". This is undesirable as it leads to duplicate data, and also the email address should reasonably be assumed to be private, where as the name can be public.

### DIFF
--- a/packages/better-auth/src/plugins/email-otp/index.ts
+++ b/packages/better-auth/src/plugins/email-otp/index.ts
@@ -446,6 +446,7 @@ export const emailOTP = (options: EmailOTPOptions) => {
 						const newUser = await ctx.context.internalAdapter.createUser({
 							email,
 							emailVerified: true,
+							name: '',
 						});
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/email-otp/index.ts
```diff
@@ -446,7 +446,6 @@ export const emailOTP = (options: EmailOTPOptions) => {
 						const newUser = await ctx.context.internalAdapter.createUser({
 							email,
 							emailVerified: true,
-							name: email,
```
Bekacru, line 449: We were also planning to remove this, but make sure to assign an empty string instead so it doesn't fail.

Bekacru (changes requested):

packages/better-auth/src/plugins/email-otp/index.ts
benkingcode: No problem, updated

PR #1523, opened 2025-02-21
feat: base64 conversion without Buffer

### LINKED ISSUE #1375: Support Cloudflare Workers without nodejs_compat

When running better-auth via Hono.js on a Cloudflare Worker, NodeJS apis are not available by default, leading to runtime error `ReferenceError: Buffer is not defined`, ref: https://github.com/better-auth/better-auth/issues/593#issuecomment-2488279122 (first reported by @lassegit).

As @avanderbergh points out [here](https://github.com/better-auth/better-auth/issues/593#issuecomment-2488412971), Cloudflare can polyfill NodeJS APIs (including Buffer) using `compatibility_flags = ["nodejs_compat"]` which makes it run. 🎉

It would be nice, to have better-auth support Coudflare Workers and other non-NodeJS runtimes by default. Since HonoJS uses non-NodeJS runtimes and is primarily placed in better-auth docs under backend. 

Just enable `compatibility_flags = ["nodejs_compat"]`.
Add to better-auth's Hono docs: https://www.better-auth.com/docs/integrations/hono

### Additional context

_No response_

BlankParticle:
BetterAuth doesn't work on cloudflare workers without `nodejs_compact` flags due to usage of `Buffer` to do base64 encoding/decoding.
This PR replaces all `Buffer` usage with `base64` from `@better-auth/utils/base64` and `TextEncoder`/`TextDecoder`.

~~There will be a follow up PR to add a cloudflare worker example and test suite to make sure workerd compatibly doesn't break in future.~~
Follow up #1524

fixes #1375

### DIFF
--- a/packages/better-auth/src/crypto/hash.ts
+++ b/packages/better-auth/src/crypto/hash.ts
@@ -1,11 +1,12 @@
 import { constantTimeEqual } from "./buffer";
 import { createHash } from "@better-auth/utils/hash";
+import { base64 } from "@better-auth/utils/base64";
 
 export async function hashToBase64(
 	data: string | ArrayBuffer,
 ): Promise<string> {
 	const buffer = await createHash("SHA-256").digest(data);
-	return Buffer.from(buffer).toString("base64");
+	return base64.encode(buffer);
 }
 
 export async function compareHash(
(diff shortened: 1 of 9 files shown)


BlankParticle:
@Bekacru do you think we should this merged to `v1.2` instead of `main` so that it can be tested better with beta versions?

PR #1524, opened 2025-02-21
feat: add cloudflare worker basic tests

BlankParticle:
Follow up to #1523.

Added a demo cloudflare worker app and a basic test suite to make sure future changes to better-auth don't break cloudflare workers.


yamz8:
Nice! @Bekacru I think a lot of folks will find this useful. Plus If you combine it with a Turbo repo and a separate backend, it will definitely go viral.

PR #1528, opened 2025-02-21
chore: fix type exports and sync dep versions to improve build times

BlankParticle:
removes some export from `better-auth` root package and syncs package version across the monorepo to improve build times.

On my machine, building only the `better-auth` package

```bash
# Before
pnpm build  184.44s user 51.85s system 141% cpu 2:46.68 total

# After
pnpm build  149.19s user 29.45s system 136% cpu 2:10.87 total
```


BlankParticle:
somehow some of the code from `v1.2` branch was not formatted, so when i ran format it fixed those files

Bekacru:
thanks! this is great!

PR #1535, opened 2025-02-22
docs(rename): `node` to `express` in integration docs

ping-maxwell:
image

image


Maqed:
I think it's cool renaming it as it was deceiving when I first read the docs. I thought better-auth didn't support express.

igorteixeiradev:
how i use this with express, I see a lot of 404

PR #1562, opened 2025-02-24
demo (nextjs): Avoid page refresh on session termination

theinfinit:
- don't refresh when session from other devices is terminated. Instead, update list of active sessions on success
- refresh happens only when current session is terminated to finish log out

### Demo of new behavior

https://github.com/user-attachments/assets/206f4b9f-1b82-4605-bc5d-788aca7477d9


Kinfe123:
i have included your pr in my demo build fix and will be closing this. thanks!

PR #1566, opened 2025-02-24
fix(two-factor): should respect `remember me` value when a user logins in with 2fa

### LINKED ISSUE #1564: twoFactor (2FA) plugin doesn't respect "Remember me" login option (not mistaken with 2FA trustDevice)

### To Reproduce

1. Launch NextJS demo.
2. Log-in (open DevTools to confirm that `session` cookie is present without 2FA)
3. Enable 2FA.
4. Log-out (All cookies should be gone now).
5. Log-in again providing 2FA code. Make sure to NOT check "Remember me" (dont_remember cookie will not be present)
6. 7 day cookie will be added instead of session cookie.
7. Log-out (All cookies should be gone now, except `better-auth.two_factor`, which is not removed on sign-out, but probably should).

Following the steps from the previous section, I expected that "Remember me" option will be respected, even if user enabled 2FA for their account. I observed, instead, that after user enables 2FA for their account, better-auth is adding 7 day cookie instead of `session` cookie.

Additionally I observed that `better-auth.two_factor` cookie is not removed on sign-out. When I tried to go to `/two-factor/` page without login step, I was able to successfully authenticate in "one-step" verification and get access to dashboard. 

Both issues are considered with untrusted devices used by non-technical user. If someone is using untrusted browser, without using private window, forgets to sign out, they should rest assured that their account is secure, if they haven't selected "Remember me" option.

Image

Tested against main branch v1.1.20

```bash
OS: Windows 11
Browser: Vivaldi 7.1.3570.54 (Stable channel) (64-bit) (chromium)
```

Client, Backend

```typescript
Config as in demo/nextjs, but with local better-sqlite DB.
```

### Additional context

I tested my reproduction against main branch v1.1.20.

Bekacru:
closes #1564

### DIFF
--- a/packages/better-auth/src/cookies/cookies.test.ts
+++ b/packages/better-auth/src/cookies/cookies.test.ts
@@ -14,7 +14,6 @@ describe("cookies", async () => {
 			{
 				onResponse(context) {
 					const setCookie = context.response.headers.get("set-cookie");
-					console;
 					expect(setCookie).toBeDefined();
 					expect(setCookie).toContain("Path=/");
 					expect(setCookie).toContain("HttpOnly");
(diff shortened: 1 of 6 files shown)


theinfinit:
That was fast 🔥

Issue #1558, opened 2025-02-24
"@better-auth/cli generate" missing the end semicolon which fails Cloudflare D1 migration


raphaelmsr:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Clone `https://github.com/Bekacru/better-auth-nextjs-cf-d1-example` and
* run ` bunx @better-auth/cli generate`
*  then `bunx wrangler d1 migrations apply my-database  --remote` 
           
The outputted SQL's last line is 
`create table "verification" ("id" text not null primary key, "identifier" text not null, "value" text not null, "expiresAt" date not null, "createdAt" date, "updatedAt" date)` which misses the last semicolon

### Current vs. Expected behavior

SQL should be 
`create table "verification" ("id" text not null primary key, "identifier" text not null, "value" text not null, "expiresAt" date not null, "createdAt" date, "updatedAt" date);`

### What version of Better Auth are you using?

1.1.19

### Provide environment information

```bash
macOS Sequoia 15.3.1 (24D70)
NextJS 14
Cloudflare D1
Bun
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript

```

raphaelmsr:
Issue comes from the 
`compiled.join(";\n\n")` which doesn't add the end semi-colon

https://github.com/better-auth/better-auth/blob/52ed183bdca101d9140ed79ec4466f44c26b0e07/packages/better-auth/src/db/get-migration.ts#L272C1-L273C1

PR #1569, opened 2025-02-25
Add last semi-colon to the compiled migrations to support Cloudflare D1 migration

PitonsSkipper:
Add last semi-colon to the compiled migrations

Fixes https://github.com/better-auth/better-auth/issues/1558


PitonsSkipper:
Fixes https://github.com/better-auth/better-auth/issues/1558

PR #1589, opened 2025-02-27
fix: repair broken pnpm lock file

BlankParticle:
The `pnpm-lock.yaml` is found to be in a broken state
image

It seems the cause might have been a merge conflict where 2 versions of this files were merged,
incase of that its recommended to run `pnpm install` which should automatically fix the merge conflict, but it was not done and the file was left in a broken state.


BlankParticle:
fixed by commit already

PR #1662, opened 2025-03-03
feat: allow promo codes

Kaseax:
currently, it is not possible to allow promotion codes for a stripe checkout session. this merge request adds a option to the stripe plugin, to enable the promotion codes.


gursheyss:
You can already allow promo codes with custom checkout params https://www.better-auth.com/docs/plugins/stripe#custom-checkout-session-parameters

fynnbauer:
yes, but after writing to bekacru he said it would make sense to offer promotion codes as a direct option.

Bekacru:
will close this in favor of just using getCheckoutParams

PR #1667, opened 2025-03-04
feat: salesforce oauth

nktnet1:
Added Salesforce provider + documentations.
- https://help.salesforce.com/s/articleView?id=xcloud.remoteaccess_oauth_web_server_flow.htm

Demo:

[output.webm](https://github.com/user-attachments/assets/19c1fe91-e54a-430f-a2b8-d2d39345706e)


nktnet1:
Hi @Bekacru,

Thank you (and Kinfe123) for fixing and merging the Zoom social provider pull request (#1671).

Do you know if there is anything that can be done on my end to help push this PR forward? Our team is looking to use this for auth in one of our projects.

Thanks 🙏 ~

tehnrd:
I am also happy to assist with this PR. 

Does this add support for [My Domain](https://help.salesforce.com/s/articleView?id=xcloud.domain_name_overview.htm&type=5)? I didn't see it. 

This is a critical requirement if building a B2B SaaS app that connects to multiple and different Salesforce instances. Using the my domain URL will automatically redirect the user to their company's SSO provider as part of the OAuth flow.

Edit: reading the docs more closely, the SSO plugin with my domain might be a better solution, where as this social login would be for generic front door login

nktnet1:
> Does this add support for [My Domain](https://help.salesforce.com/s/articleView?id=xcloud.domain_name_overview.htm&type=5)? I didn't see it.

Hi @tehnrd,

Can you elaborate more on this? What do you think is missing?

I'm able to redirect to my company's email by simply specifying a custom instance url, e.g.

image

I've made a branch updating the demo/nextjs app in this branch if you want to test it out:
- https://github.com/nktnet1/better-auth/tree/salesforce-oauth-with-demo

Just need to have these env:
```
TURSO_DATABASE_URL=
TURSO_AUTH_TOKEN=
SALESFORCE_CLIENT_ID=
SALESFORCE_CLIENT_SECRET=
```

The instructions for setting up the client app is in the docs included with this MR - you can also run the docs site locally.

tehnrd:
@nktnet1 I think for an identity-only login solution, this will work okay. I _think_ where it starts to come up short is if you use the login.salesforce.com front door, the instance URL for the org is not saved anywhere, so you cannot use the provided refresh/access token to access the API. 

You could set up the instance URL on the config and then try to access this, but then you are hardcoding the Instance URL in the auth config to only one domain. This may be okay if you are building an internal app where the instance URL will never change, but this will not work if you build an app that needs to connect to many different Salesforce instances. I think the SSO plugin handles this second scenario, and I'm currently investigating.

nktnet1:
Yeah, I see.

I'm not sure if better-auth currently has a recommended way to deal with this requirement - took a quick glance at the [Google](https://www.better-auth.com/docs/authentication/google) and [Microsoft](https://www.better-auth.com/docs/authentication/microsoft) and there's no support for it either.

At the moment, the three ways (workarounds) I can think of are
- [SSO plugin](https://www.better-auth.com/docs/plugins/sso) (as you mentioned)
- [Generic Oauth Plugin](https://www.better-auth.com/docs/plugins/generic-oauth)
- Wrapper function to create auth configs (e.g. [like this](https://github.com/nktnet1/rt-stack/blob/afd8e53c1c76228554f774fdfbf1d1d7d1f22476/packages/auth/src/server.ts#L13-L36)), then call it multiple times passing different args.

drewcasebeer:
@tehnrd Were you able to determine if the SSO plugin was the way to do this?

Kinfe123 (approved):

nktnet1:
Closing - #4183 merged.

PR #1675, opened 2025-03-04
docs: installation docs to automatic install using `init` cmd

ping-maxwell:
image


himself65 (approved):
This looks good. I like it

PR #1716, opened 2025-03-07
fix: getSessionCookie function and improve docs.

### LINKED ISSUE #1487: Cookie parsing issue in production build with getSessionCookie function

### To Reproduce

1. Set up a Next.js project with better-auth.
2. Add middleware to check session:

**Code Snippet**  
```ts
import { NextRequest, NextResponse } from "next/server";
import { getSessionCookie } from "better-auth";

export function middleware(request: NextRequest) {
  const sessionCookie = getSessionCookie(request, {
    cookieName: "session_token",
    cookiePrefix: "better-auth"
  });

  if (!sessionCookie) {
    const redirectResponse = NextResponse.redirect(
      new URL("/login", request.url)
    );
    redirectResponse.headers.set("x-middleware-cache", "no-cache");
    return redirectResponse;
  }

  return NextResponse.next();
}



getSessionCookie function returns null inproduction with config object or without. works perfectly in dev mode.

The getSessionCookie function fails to parse cookies in production builds (npm run start) while working correctly in development (npm run dev). The issue appears to be caused by an undefined variable al in the minified code.

Expected behavior
The getSessionCookie function should correctly parse and return the session cookie in both development and production environments.

Actual behavior
Development: Works correctly, returns the session cookie.
Production: Returns null, even though the cookie is present in the request headers.

The cookie is present in headers: "better-auth.session_token=wQbfYY7OC6SkNaSrx63w7yrW2Unp2WCn.yw37B3fnA2WlkkFa2W5dr28Dj7buVLyTIsbKnM0LA3g%3D"

1.1.18

```bash
Next.js version: 15.1.6
"react": "^18.3.1",
Operating System: macOS/Windows/Linux
```

Backend

```typescript
import { prisma } from "./prisma-client";
import emailService from "@/services/email.service";

    provider: "postgresql",
  }),
  emailAndPassword: {
    enabled: true,
  },
  socialProviders: {
    google: {
    },
  },
  emailVerification: {
      await emailService.sendVerificationEmail(user.email, url);
    },
  },
  plugins: [admin({ impersonationSessionDuration: 60 * 60 })],
});
```

### Claude (AI) suggestrion

The issue is that al is undefined in the minified code, but it's being used in a ternary condition. This is likely a minification issue where a variable name got mangled.

```ts
(e,t)=>{
    let r=(e instanceof Headers?e:e.headers).get("cookie");
    if(!r)return null;
    let{cookieName:n="session_token",cookiePrefix:i="better-auth"}=t||{},
    a=al?`__Secure-${i}.${n}`:`${i}.${n}`;  // <-- Problem is here
    return(function(e){
        let t=e.split("; "),r=new Map;
        return t.forEach(e=>{let[t,n]=e.split("=");r.set(t,n)}),r
    })(r).get(a)||null
}

ahmed-m-abbass:
Improved the logic in getSessionCookie() to correctly handle secure prefix and added useSecureCookies config.
closes #1487.

### DIFF
--- a/docs/content/docs/integrations/next.mdx
+++ b/docs/content/docs/integrations/next.mdx
@@ -131,15 +131,26 @@ In Next.js middleware, it's recommended to only check for the existence of a ses
 
 You can use the `getSessionCookie` helper from Better Auth for this purpose:
 
+<Callout type="warn">
+The <code>getSessionCookie()</code> function does not automatically reference the auth config specified in <code>auth.ts</code>. Therefore, you need to ensure that the configuration in <code>getSessionCookie()</code> matches the config defined in your <code>auth.ts</code>.
+</Callout>
+
 ```ts
 import { NextRequest, NextResponse } from "next/server";
 import { getSessionCookie } from "better-auth/cookies";
 
 export async function middleware(request: NextRequest) {
-	const sessionCookie = getSessionCookie(request); // Optionally pass config as the second argument if cookie name or prefix is customized.
+	const sessionCookie = getSessionCookie(request, {
+        // Optionally pass config if cookie name, prefix or useSecureCookies option is customized in auth config.
+		cookieName = "session_token",
+		cookiePrefix = "better-auth",
+		useSecureCookies = true,
+    });
+
 	if (!sessionCookie) {
 		return NextResponse.redirect(new URL("/", request.url));
 	}
+
 	return NextResponse.next();
 }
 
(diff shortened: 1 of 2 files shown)


daveycodez:
useSecureCookie needs to default to false but this check shouldn’t even exist. It should check for the presence of either existing, not conditionally. This is an optimistic check for presence, not validity. This is still breaking by default for users and it’s still easier to manually check cookies in middleware atm due to unpredictability.

https://discord.com/channels/1288403910284935179/1348847982748897380/1348917566600511498

I’m on a trip right now but this is how this needs to work. I left a code snippet for how cookie should be checked

@Bekacru @ahmed-m-abbass

PR #1729, opened 2025-03-08
add(docs): libraries category

ping-maxwell:
image

image


Kinfe123:
may be mentioning the utils somewhere on the docs might be a good idea instead of having its own page since this is not always related to the auth for the most cases like more of the sec part like hashing and protection and also this could be something we can add a ref for the usage as well. what do you think ?

ping-maxwell:
Yeah.

Also since the release of the better-auth-awesome repo, this category may not be needed anymore.
I think might be a better idea to just close this instead.

PR #1737, opened 2025-03-08
feat: refactor database adapters

Netrifier:
1. **Drizzle Adapter:**
   - Refactored the `convertWhereClause` function to handle all operator conversions.
   - Changed the order of parameters `convertWhereClause(model: string, where?: Where[])` to make them consistent with other adapters.
   - Add `[# Drizzle Adapter]` prefix to error messages.
   - Fixed count method to return count `res[0].count` as drizzle returns an array with the count.

2. **Memory Adapter:**
   - Refactored the `convertWhereClause` function to handle all operator conversions.
   - Add `[# Memory Adapter]` prefix to error messages.
   - Fixed count method to use where clause.

3. **MongoDB Adapter:**
   - Updated `serializeID` and `deserializeID` to add proper typescript types.
   - Changed the order of parameters `convertWhereClause(model: string, where?: Where[])` to make them consistent with other adapters.
   - Changed the Error type to `BetterAuthError` instead of `Error`.
   - Added `escapeRegex` function to escape regex input to prevent regex injection.
   - Fixed regex in `contains`, `starts_with`, and `ends_with` operators.
   - Add `[# MongoDB Adapter]` prefix to error messages.
   - Fixed count method to use where clause.

   **MongoDB Tests:**
   - Fixed the `clearDB` function to delete `sessions` collection instead of `session` collection.
   - Added a test to verify proper regex input sanitization.

4. **Prisma Adapter:**
   - Extracted the check if model exists logic to `checkModelExistsOrThrow` function.
   - Refactored the `convertWhereClause` function.
   - Updated the `operatorToPrismaOperator` function to handle all operators.
   - Add `[# Prisma Adapter]` prefix to error messages.

5. **Tests:**
   - Added test for `adapter.count()` method with and without where clause.
   - Added tests for the operators `ne`, `eq`, `lt`, `lte`, `gt`, `gte`, `in` and undefined operator option.
image
image
image
image
image
image

PR #1747, opened 2025-03-09
fix(drizzle adapter): count returning undefined

Creaous:
During the creation of a custom plugin that I am working on, I found out that the Drizzle adapter count function would return undefined. This is due to the count function in Drizzle returning with `[{ count: 1 }]` instead of `{ count: 1 }`. This pull request changes `res.count` to `res[0].count` which fixes this issue.

Example issue repository (uses `internalAdapter.countTotalUsers()`): https://github.com/Creaous/better-auth-drizzle-count-issue


Creaous:
closed in favour of #1737

PR #1750, opened 2025-03-09
refactor(CLI): init command

ping-maxwell:
Minor changes. 

(Just joking 🤣)

Recoded the entire init command because the old version was just a pure mess.
There are a lot more "checks" in place and a much more solid foundation around each step of the init command.

Also, I've added these on the init command since the previous:
- added social providers option
- sqlite db adapter now prompts to also install the @types package as well (and saved in dev-deps)
- we now generate the API route file for each supported framework
- and probably other things I added that just forgot

The PR is much larger than the previous init code because:
1. It's now code-split
2. There is actual foundational code
And 3, most importantly, it now is agnostic (only frontend/fullstack), whereas the previous only supported Nextjs

Note: This PR does not include the "story mode" idea.


Bekacru:
close and re-open?

PR #1758, opened 2025-03-10
feat: battlenet social provider

sandervspl:


sandervspl:
Battle.net oauth does not provide the user's email, and better-auth requires an email when creating a user. I will close this PR until something changes in that regard.

PR #1776, opened 2025-03-11
docs : fixed sign-up component of builder

Exe16Kishan:
it was not recognising this as react component  image

after adding default to it , the error is fixed
image


Exe16Kishan:
I fixed a problem in the stack that should make things smoother for users—no more errors! Could you take a quick look when you get a chance? Thanks a bunch

PR #1791, opened 2025-03-12
refactor(plugin): two-factor plugin disable endpoint shouldn't validate password length

LovelessCodes:


himself65:
Already fixed, closing

PR #1803, opened 2025-03-13
feat(account): multiple account with the same provider

Kinfe123:


JesperSvensson00:
Feel like this change should have been communicated better in the change log since old implementations of the unlinkAccount function no longer works.

Issue #1077, opened 2024-12-31
linkSocial idToken supports 


reslear:
### Is this suited for github?

- [X] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

linkSocial with idToken supports, and prevent redirect 

i have a custom social login for native platforms: 

```ts
async function signInGoogle() {

  const token = await GoogleLogin.login()
  
  await authClient.signIn.social({
    provider: 'google',
    idToken: {
      token,
    },
  })
}

```

### Describe the solution you'd like

To link my account, I need to do the same thing now, but I can't because there are redirects. 

```ts
async function linkGoogle() {

  const token = await GoogleLogin.login()
  
  await authClient.signIn.linkSocial({
    provider: 'google',
    
    // not supports 
    idToken: {
      token,
    },
  })
}

```

### Describe alternatives you've considered

-

reslear:
hi @Bekacru is ok if i copy/paste this part?

from:

https://github.com/better-auth/better-auth/blob/52ed183bdca101d9140ed79ec4466f44c26b0e07/packages/better-auth/src/api/routes/sign-in.ts#L166-L243

between:

https://github.com/better-auth/better-auth/blob/52ed183bdca101d9140ed79ec4466f44c26b0e07/packages/better-auth/src/api/routes/account.ts#L138-L139

reslear:
@dosubot
 not stale [#1830](https://github.com/better-auth/better-auth/pull/1830)

Bekacru:
hi @reslear sorry for the bot :)) will review this PR and merge

PR #1830, opened 2025-03-15
feat: Link account with idToken 

reslear:
closes: #1077

- [x] implementation
- [x] docs


reslear:
Hi @Bekacru I'm not sure I did it right, you should double-check, so pls review.


And also maybe need add checking if emails match unless explicitly allowed like: 


```ts
if (
  !c.context.options.account?.accountLinking?.allowDifferentEmails &&
  session.user.email !== userInfo.user.email.toLowerCase()
) {
  throw new APIError("")
}
```

packages/better-auth/src/api/routes/account.ts
```diff
@@ -142,6 +157,86 @@ export const linkSocialAccount = createAuthEndpoint(
...
+			if (!userInfo.user.email) {
+				c.context.logger.error("User email not found", {
+					provider: c.body.provider,
+				});
+				throw new APIError("UNAUTHORIZED", {
+					message: BASE_ERROR_CODES.USER_EMAIL_NOT_FOUND,
+				});
+			}
+
+			const data = await handleOAuthUserInfo(c, {
```
kzlar: Hi @reslear,
I'm wondering if you tested this? I copied your implementation to a plugin and found that it did not work. `handleOAuthUserInfo` seems to deal with creating users instead of linking them.
Instead what worked for me is copying the `link` implementation of the `callback` route:
https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/api/routes/callback.ts#L123
See the gist here:
https://gist.github.com/kzlar/7d003ebfe9874e37417d65f0312e4162

reslear: need @Bekacru help

reslear: Hi @kzlar How's it working in production? After a couple weeks? 

I added you as a maintainer could you edit this pr? https://github.com/reslear/better-auth/tree/feat/link-social-id-token

packages/better-auth/src/api/routes/callback.ts
```diff
@@ -126,6 +126,19 @@ export const callbackOAuth = createAuthEndpoint(
 		}
 
 		if (link) {
+			const trustedProviders =
```
kzlar, line 129: Bekacru, note that I added this since I think it is crucial and missing from the current implementation

reslear:
hi @Bekacru I think we're ready, please do review.

packages/better-auth/src/api/routes/callback.ts
reslear: @kzlar could you also fix the tests?

reslear:
hi @Bekacru @Kinfe123 @ping-maxwell I think we're ready, please do review, this is a much-needed feature 😎

ping-maxwell (approved):

reslear:
hi @Bekacru did you get to see it?

reslear:
@Bekacru friendly ping ❤️

reslear:
@Bekacru friendly ping ❤️

PR #1842, opened 2025-03-16
Fix Incorrect JSON Handling in secondaryStorage.get Documentation

meisolated:
This PR updates the `database.mdx` documentation file to fix an issue caused by incorrect handling of JSON data in `secondaryStorage.get`.  

#### **Issue**  
The current documentation suggests using:  
```js
return value ? JSON.stringify(value) : null;
```  
This results in the stored JSON string being re-stringified, causing an issue when parsing it later. Specifically, when `safeJSONParse` is used, it returns a string instead of an array, leading to:  
```
TypeError: list.filter is not a function.
```

#### **Fix**  
The correct approach is to return the value directly instead of wrapping it in `JSON.stringify(value)`. The updated code should be:  
```js
return value ? value : null;
```

#### **Changes**  
- Updated `database.mdx` to reflect the correct implementation.  
- Added a note explaining that `JSON.stringify` should not be used in `get`, as Redis stores values as strings by default.  

#### **Impact**  
This fix ensures that retrieved JSON data remains correctly formatted and prevents runtime errors related to `.filter` being called on a string instead of an array.

### DIFF
--- a/docs/content/docs/concepts/database.mdx
+++ b/docs/content/docs/concepts/database.mdx
@@ -82,7 +82,7 @@ export const auth = betterAuth({
 	secondaryStorage: {
 		get: async (key) => {
 			const value = await redis.get(key);
-			return value ? JSON.stringify(value) : null;
+			return value ? JSON.parse(value) : null;
 		},
 		set: async (key, value, ttl) => {
 			if (ttl) await redis.set(key, value, { EX: ttl });
--- a/docs/content/docs/concepts/database.mdx
+++ b/docs/content/docs/concepts/database.mdx
@@ -82,7 +82,7 @@ export const auth = betterAuth({
 	secondaryStorage: {
 		get: async (key) => {
 			const value = await redis.get(key);
-			return value ? JSON.parse(value) : null;
+			return value ? value : null;
 		},
 		set: async (key, value, ttl) => {
 			if (ttl) await redis.set(key, value, { EX: ttl });


docs/content/docs/concepts/database.mdx
```diff
@@ -82,7 +82,7 @@ export const auth = betterAuth({
 	secondaryStorage: {
 		get: async (key) => {
 			const value = await redis.get(key);
-			return value ? JSON.stringify(value) : null;
+			return value ? value : null;
```
reslear, line 85: `return value ?? null;`

meisolated: You're right, I was working on a project, few different things were breaking, so I just went with whatever solved the problem at that moment. 

when i come to think of it, this is also doable actually.
```ts
return (await redis.get(key)) ?? null
```

reslear: @meisolated I fixed it, by the way. I don't understand why it had to be done again

https://github.com/better-auth/better-auth/pull/1047 

Does it look like?

meisolated: Thanks for fixing it, I have no idea, seeing your PR, it should have been correct in the documentation.

reslear: @meisolated 

I found out who messed it up: https://github.com/better-auth/better-auth/pull/1359 by @zohaibakber

meisolated: Ah, the mystery is solved! Thanks for catching that and fixing it!

zohaibakber: > @meisolated
> 
> I found out who messed it up: #1359 by @zohaibakber
Im using upstash redis package and this is whats happens without stringifying it sousing JSON.stringify(value) fixes it for me, so yeah maybe I thought this was a bug thats why I created that pr apparently its not so yeay whatever
image

PR #1854, opened 2025-03-17
fix: impersonatedBy not appearing in client (type fix)

atharvadeosthale:
This PR fixes the admin plugin, where when the user sessions are listed on the client, `impersonatedBy` field is not suggested by the code editor but the `impersonatedBy` field is clearly available when using `console.log()`.


atharvadeosthale:
- Image 1: `impersonatedBy` not appearing in `session`
- Image 2: `impersonatedBy` field does exists when `console.log()`

image

image

atharvadeosthale:
Bumping this!

PR #1865, opened 2025-03-18
fix(get-request-ip): improve header value retrieval for IP address extraction

ozgurozalp:
This pull request includes a small but important change to the `getIp` function in the `packages/better-auth/src/utils/get-request-ip.ts` file. The change ensures compatibility with different types of request objects by checking if the `headers` object has a `get` method before attempting to call it.

* [`packages/better-auth/src/utils/get-request-ip.ts`](diffhunk://#diff-b386c725fba47b675a264ee1c46a6a31cf037534f1aae2f758cb7fbf1a18300bL29-R29): Modified the `getIp` function to use `headers.get(key)` if the `headers` object has a `get` method, otherwise it falls back to `headers[key]`.

### DIFF
--- a/packages/better-auth/src/utils/get-request-ip.ts
+++ b/packages/better-auth/src/utils/get-request-ip.ts
@@ -26,7 +26,7 @@ export function getIp(
 	];
 	const headers = "headers" in req ? req.headers : req;
 	for (const key of keys) {
-		const value = headers.get(key);
+		const value = "get" in headers ? headers.get(key): headers[key];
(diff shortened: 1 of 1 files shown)


ozgurozalp:
I made this change because it's not possible to read header data this way when using Express.js.


*update: I guess it's not a problem for express lol, but it would still be an extra check :d

jamesone:
Interesting - I wonder if we need to create a parser that happens at the top level and propogates through the rest of the lib -

PR #1867, opened 2025-03-18
feat/team slug

Kinfe123:
- **feat: limit query on list org**
- **feat: team slug**

### DIFF
--- a/packages/better-auth/src/plugins/organization/organization.ts
+++ b/packages/better-auth/src/plugins/organization/organization.ts
@@ -84,6 +84,7 @@ export interface OrganizationOptions {
 	 * @default 100
 	 */
 	membershipLimit?: number;
+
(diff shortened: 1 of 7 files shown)


packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -396,6 +398,11 @@ export const organization = <O extends OrganizationOptions>(options?: O) => {
 							required: false,
 							fieldName: options?.schema?.team?.fields?.updatedAt,
 						},
+						slug: {
+							type: "string",
+							required: false,
+							fieldName: options?.schema?.team?.fields?.slug,
+						},
```
himself65, line 405: Should this be unique?

himself65 (changes requested):

PR #1876, opened 2025-03-19
feat(cli): custom cli config

### LINKED ISSUE #874: [CLI] Custom path to tsconfig
### Given

nx workspace with Angular, Elysia and `tsconfig.base.json` in the root of the project.

### Issue

Can't generate `schema.prisma` because of missing `tsconfig.json`.

### Solutions:

* Add an option like `--tsConfig` or `--ts-config` to specify the path to a custom TypeScript config.
  `bunx @better-auth/cli generate --config libs/core/api/src/lib/auth.ts --ts-config tsconfig.base.json`
* Alternatively, allow a `better-auth.config.ts` file in the project root to manage configurations.
  ```ts
  // better-auth.config.ts

  import { defineConfig } from '@better-auth/cli';

  export default defineConfig({
    config: 'libs/core/api/src/lib/auth.ts', // or even better `@my-company/core/api`. Assuming auth is exported.
    tsConfig: 'tsconfig.base.json',
  });
  ```

  `bunx @better-auth/cli generate` would use the config defined above.

Kinfe123:
closes #874


ping-maxwell:
👀 Looks very promising. Good idea.

ping-maxwell:
Is it a good idea to have `defineConfig` come from `@better-auth/cli` versus it coming straight from `better-auth`?

I know that in the end the user would have to run npx @better-auth/cli and have that cli package as something installed on their system, but if it was just `better-auth` it would be 1 less dependency in their package.json.

Could probably have a `better-auth/cli` instead?

Kinfe123:
@ping-maxwell hmmm... yeah that actually makes sense but there are def a cons on this. I initially thought of this but i was looking for something that keeps the core better-auth package lightweight and focused on runtime functionality. Allows for modular and independent versioning of the CLI too since it is not sth related to the core of better-auth. and also i have seen most of the projects keep this kind of pattern to make it more flexible and versioned as well.. what do you think? @Bekacru

ping-maxwell:
> @ping-maxwell hmmm... yeah that actually makes sense but there are def a cons on this. I initially thought of this but i was looking for something that keeps the core better-auth package lightweight and focused on runtime functionality. Allows for modular and independent versioning of the CLI too since it is not sth related to the core of better-auth. and also i have seen most of the projects keep this kind of pattern to make it more flexible and versioned as well.. what do you think? @Bekacru

Yeah that makes sense. 
Though I'm not sure how much more weight `defineConfig` would be adding onto Better-Auth, since I assume it's just to make sure the user defined the right types? (correct me if I'm wrong)
Besides that, the independent versioning is always a nice win.

Issue #1892, opened 2025-03-20
Stripe Customer Portal


joeychrys:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

Hey, I think it would be nice to have a way for a user to access the stripe portal directly.

### Describe the solution you'd like

Something like the following would work. I've already tested it. I'll create a PR for it in a bit.

```
customerPortal: createAuthEndpoint(
			"/subscription/portal",
			{
				method: "POST",
				body: z.object({
					returnUrl: z.string().default("/"),
				}),
				use: [
					sessionMiddleware,
					originCheck((ctx) => ctx.body.returnUrl),
				],
			},
			async (ctx) => {
				const { user } = ctx.context.session;
				
				if (!user.stripeCustomerId) {
					throw new APIError("BAD_REQUEST", {
						message: "No Stripe customer found for this user",
					});
				}

				try {
					const { url } = await client.billingPortal.sessions.create({
						customer: user.stripeCustomerId,
						return_url: getUrl(ctx, ctx.body.returnUrl),
					});

					return ctx.json({
						url,
						redirect: true,
					});
				} catch (error) {
					ctx.context.logger.error("Error creating portal session", error);
					throw new APIError("INTERNAL_SERVER_ERROR", {
						message: "Failed to create customer portal session",
					});
				}
			},
		),

```

### Describe alternatives you've considered

You could manually create the portal but it would be somewhat inconvenient.

mazeincoding:
+1 for this

I see you already made a PR though - gonna check it out

joacohoyos:
Any update on this getting merged any time soon?

Bekacru:
I think this should already avalible. Feel free to tag me if not

PR #1893, opened 2025-03-20
feature: stripe customer portal

joeychrys:
This is an addition to the stripe plugin that creates a stripe customer portal and redirects. 

Tests have also been added for this addition.

### DIFF
--- a/packages/stripe/src/index.ts
+++ b/packages/stripe/src/index.ts
@@ -1,24 +1,23 @@
 import {
-	type BetterAuthPlugin,
 	type GenericEndpointContext,
+	type BetterAuthPlugin,
 	logger,
 } from "better-auth";
-import {
-	APIError,
-	getSessionFromCtx,
-	originCheck,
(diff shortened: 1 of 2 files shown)


joeychrys:
Link to issue that I made: 

https://github.com/better-auth/better-auth/issues/1892

lmvdz:
@joeychrys maybe it should be moved outside of the subscription endpoints object? into a `billingEndpoints` object, this will allow the plugin users to not have subscriptions enabled to still be able to use this feature.

joeychrys:
> @joeychrys maybe it should be moved outside of the subscription endpoints object? into a `billingEndpoints` object, this will allow the plugin users to not have subscriptions enabled to still be able to use this feature.

Ah good point. I'll look into it today or later this week.

followbl:
pump this in to my veins

Bekacru:
thanks for the PR! closed in favor of #3625

PR #1904, opened 2025-03-20
docs: Add scopes and redirect warning to Google doc

hbmartin:


LZL0:
Bumping this.

hbmartin:
@LZL0 thanks 🙏  Unfortunately there's nothing more I can do at this point.
And the response (or lack ) to my other PR's doesn't leave me hopeful (e.g. https://github.com/better-auth/better-auth/pull/1906 and https://github.com/better-auth/better-auth/pull/1905 )

LZL0:
> @LZL0 thanks 🙏 Unfortunately there's nothing more I can do at this point. And the response (or lack ) to my other PR's doesn't leave me hopeful (e.g. #1906 and #1905 )

@Bekacru could you please look at these docs?

Kinfe123:
ig this docs is good thing but the default client would get you an access to the data that better auth needs so there error is just malconfiguration on their redirection url which is something that is cleared out even on the google console docs. so feel free to make the docs optimized and clean.

hbmartin:
> ig this docs is good thing but the default client would get you an access to the data that better auth needs so there error is just malconfiguration on their redirection url which is something that is cleared out even on the google console docs. so feel free to make the docs optimized and clean.

Uh... what??

LZL0:
> ig this docs is good thing but the default client would get you an access to the data that better auth needs so there error is just malconfiguration on their redirection url which is something that is cleared out even on the google console docs. so feel free to make the docs optimized and clean.

The pull request only includes changes to the docs.

Kinfe123:
We need to be somehow consistent even on drafting docs that's what I meant.

PR #1905, opened 2025-03-20
Move nextjs demo to examples directory

hbmartin:
The existence of both demo and examples folder is confusing imo
There are NO code changes just a `git mv`


Kinfe123:
We are not currently planning to do that and we would like to make everything as clear as possible but we will consider it if it is something necessary to do

PR #1922, opened 2025-03-21
chore: update polar docs

catalinpit:
Add a mention that specifies you don't need to configure Polar on the client


catalinpit:
Also, regarding this:

CleanShot 2025-03-21 at 20 16 50@2x

What's the point of installing `@polar-sh/better-auth` on the client?

Bekacru:
> What's the point of installing `@polar-sh/better-auth` on the client?

nothing. feel free to remove the callout ;)

PR #1325, opened 2025-02-01
add(docs): database adapter guide

ping-maxwell:

PR #1926, opened 2025-03-21
feat: `createAdapter` and `useNumberId`

ping-maxwell:
## TLDR

This PR was meant to only be the addition of `createAdapter`. However over time I realized a lot of other things relating to the adapter or databases that needed to be improved. Therefore this PR now includes a lot more things.

The primary things this PR does:
* Introduce `createAdapter`.
* Support `useNumberId`.
* Modify database related options in the Better auth config to be under `options.advanced.database.` to keep things together.

## What is `createAdapter`?

The `createAdapter` function will handle everything a database adapter shouldn't be.

The devs who create database adapters shouldn't have to worry about stuff like incorrect model names just because the end better-auth user had special configurations, or the fact that ID generations can be customized.

This `createAdapter` helper allows devs who create database adapters to only focus on database logic. No more need of worrying about whether the end user wanted plurals for model names, or if a given model already has defaultValues, let alone if the model name is accurate according to the schema. There are a lot of minor details which are important that a dev may not be aware of when creating a Better-Auth adapter.
This helper function will handle everything that Better-auth really should be handling before passing data to the adapter.

Another really important point:
In the future, if Better-Auth wants to make any adjustment towards how adapters work (eg add support for soft-deletes), development becomes **much** easier, faster and definitely more consistent to achieve. 

### **`createAdapter`**:
* Add: `createAdapter` helper.
* Add: tests for this new `createAdapter` helper.
* Add: Documentation on `createAdapter` in docs. 

### **`useNumberId`**:
* Add: `useNumberId` option under `advanced.database` to convert id fields referenced in where clauses and similar cases into numbers.
* Add: Documentation for `useNumberId`.
* Add: support for autoIncrement in drizzle schema generation when `useNumberId` is enabled.
* Add: New `useNumberId` adapter tests, each DB adapter now tests this too.
* Update: `getMigrations` to support `useNumberId`.

### **Misc:**
* Refactor: removed all `id` fields in any adapter.create call across Better-Auth. If there is a missed case where the `id` is there, it will throw an error in the console as well as stack traces to locate it. We should look to remove all `id` fields as the adapter will handle this.
* Add: support for `uuid` in Drizzle schema generation for postgresql DB type.
* Add: support for `string[]` and `number[]` field types in drizzle adapter schema generation.
* Fix: Drizzle schema generation doesn't set the correct foreign key `onDelete` action based on schema.
* Fix: Prisma schema generation doesn't set the correct foreign key `onDelete` action based on schema.
* Add: MemoryAdapter parameter now has options to enable or disable debug logs.
* Update(breaking): Moved `generateId` in BetterAuthOptions to be under `advanced.database`
* Add: `defaultFindManyLimit` to BetterAuthOption's `advanced.database`.

### **Adapter tests:**
* Update: existing adapter tests to also make sure that all `create` methods will return an `id`.
* Add: Option to skip any tests as needed for adapter tests.
* Update: renamed `memory.test` to `adapter.memory.test`
* Update: `package.json`'s `test:adapters` script to run `vitest adapter` instead of `vitest adapters` since adapter tests names start with `adapter.name.test`, instead of `adapters.name.test` (Both the old script and new work, just changed it since it seemed more accurate)
* Update: `package.json`'s primsa push scripts. We now automatically push the prisma schemas during each tests automatically. Also there are 2 separate schemas now, one for the normal adpater tests, the other is for the autoIncrement number id tests.
* Add: Option to pass `testPrefix` which adds that prefix in front of each test name.

## Potential Breaking Changes ⚠️

* `generateId` is no longer under `advanced`, instead it's now under `advanced.database`
  * Update: This is no longer removed in `advanced`, but I added `@deprecated` indicating to move to `advanced.database.generateId`. Also, there will be warning logs telling them to move it.

## What's Next?

Future plans (by me) that may or may not be accepted into BetterAuth regarding adapters/databases:
1. Transaction support. 
3. Soft-delete support.
4. Support custom primary keys other than `id`. (Very unlikely)
5. Database Indexing support.
6. Database Hooks to support any table.
7. createAdapter automatically routes secondaryStorage related tables *to* secondary storage.
8. Support a more dynamic secondary storage system.

## Todo:

- [x] `createAdapter`
  - [x] Tests for `createAdapter`
  - [x] Add Documentation.
  - [x] Update **DrizzleAdapter** to use it.
  - [x] Update **KysleyAdapter** to use it.
  - [x] Update **MemoryAdapter** to use it.
  - [x] Update **PrismaAdapter** to use it.
  - [x] Update **MongoDbAdapter** to use it.
- [x] Update adapter tests to allow disabling any tests as needed
  - [x] Update documentation regarding this ^
- [x] Add `defaultFindManyLimit` to BetterAuthOption's `advanced.database` obj.
  - [x] Update documentation regarding this ^
- [x] Redesign the overall `id` system.
  - [x] Remove all `id` fields in every `adapter.create` call. (Done as far as I'm aware, however it's possible that there are ones that I missed. In which case there will be warning logs to help track these down and remove them)
  - [x] Support `useNumberId`.
    - [x] `createAdapter` helper to support `useNumberId`.
    - [x] Create tests for `useNumberId`.
    - [x] Documentation for `useNumberId`. 
    - [x] **DrizzleAdapter:**
      - [x] Adapter should support `useNumberId`
      - [x] Schema generation support
      - [x] Add the new `useNumberId` tests
      - [x] ^ same for the `mysql` test file
    - [x] **KyselyAdapter:**
      - [x] Adapter should support `useNumberId`
      - [x] Schema generation support
      - [x] Migration support
      - [x] Add the new `useNumberId` tests
    - [x] **MemoryAdapter:**
      - [x] Adapter should support `useNumberId`
      - [x] Add the new `useNumberId` tests
    - [x] **MongoDbAdapter:**
      - [x] ~~Adapter should support `useNumberId`~~ (MongoDB doesn't support numeric IDs)
      - [x] ~~Add the new `useNumberId` tests~~
    - [x] **PrismaAdapter:**
      - [x] Adapter should support `useNumberId`
      - [x] Schema generation support
      - [x] Add the new `useNumberId` tests

### Prisma Adapter Tests
Since the PrismaClient can only work on 1 prisma.schema at a time, it was tricky to make 2 primsa adapter tests with different PrismaClient instances.
The solution I ended up was to run a script which pushes the schema, after which will run an `import` to get the PrismaClient, then run the adapter test. Since the import was called asynchronously, after the schema was pushed, it means the PrismaClient is up-to-date with the new schema changes.

The second challenge was getting both prisma adapter tests to run one after another. Since normally tests can run alongside each other. This would cause each prisma adapter tests to push it's individual schemas consecutively, causing errors. So the solution I ended up doing was creating a `state.txt` file. The 2nd adapter tests will watch this state.txt file to change, and if it does, checks it's `IDLE`, if so it will run it's test. And I'm sure you can already guess that the first test will set this to `RUNNING` and then set it back to `IDLE` once it's finished.

### Kysely Adapter tests
Similar situation as the Prisma adapter tests, similar solution to it.

### `createAdapter` adapter-test debug log options & functions

Added the option to pass `isRunningAdapterTests` in the `debugLogs` option in the `createAdapter` config.
This option by default has the JSDoc of `@deprecated` thus hiding it (Or placing it at the very bottom of the menu for choosing methods in `debugLogs` options obj) from the IDE in case end-users accidentally enable it.
This should only be enabled during our adapter tests, which will essentially make all debugLog transactions save into memory, and whenever a test case fails, we will print that test's debug logs.

This really helps with debugging/finding issues while developing an adapter.

Note: These specifial *adapterTestDebugLog* functions are only in the result of `createAdapter` **IF** the config has enabled `isRunningAdapterTests`. Otherwise the result is just the normal `Adapter`.

## Closes:

- https://github.com/better-auth/better-auth/pull/1325
- https://github.com/better-auth/better-auth/pull/1976
- https://github.com/better-auth/better-auth/pull/1299


ping-maxwell:
~~Self reminder: Disable debug logs on all adapters once finished~~
Done.

ping-maxwell:
Added a bunch of important tests which normal adapter tests may not take into account.
image

dennisjnnh:
@ping-maxwell Was this update supposed to break this config already?
```
  advanced: {
    cookiePrefix: 'redacted',
    generateId: false,
    ipAddress: { ipAddressHeaders: ['x-forwarded-for'] },
  },
```

with better-auth@1.2.6, I have to use this config, otherwise, better auth does not notice the generateId config at advanced.generateId.
```
  advanced: {
    cookiePrefix: 'redacted',
    database: { generateId: false },
    ipAddress: { ipAddressHeaders: ['x-forwarded-for'] },
  },
```

PR #1941, opened 2025-03-22
fix(stripe): allow plan retrieval by annual discount price ID

Lionvsx:
Fix https://github.com/better-auth/better-auth/issues/1942


Lionvsx:
Fix https://github.com/better-auth/better-auth/issues/1942

Issue #1891, opened 2025-03-20
Rate Limit doesn't work for Email OTP


dinogomez:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Use the Email OTP
2. Add rate limiting with secondary storage (I'm using upstash redis)
3. Spam the verification endpoint when submitting a wrong otp token.

### Current vs. Expected behavior

```ts
                        "/sign-in/email-otp": {
				window: 300,
				max: 5,
			},
			"/sign-in/email": {
				window: 300,
				max: 3,
			},
			"/email-otp": {
				window: 300,
				max: 3,
			},
			"/send-verification-email": {
				window: 300,
				max: 3,
			},
```

I tried several paths, including the `/api/auth/sign-in/email-otp` but still doesn't work.

But im still able to spam the api.
```
POST /api/auth/sign-in/email-otp 400 in 450ms
 POST /api/auth/sign-in/email-otp 400 in 274ms
 POST /api/auth/sign-in/email-otp 400 in 276ms
 POST /api/auth/sign-in/email-otp 400 in 264ms
 POST /api/auth/sign-in/email-otp 400 in 265ms
 POST /api/auth/sign-in/email-otp 400 in 267ms
 POST /api/auth/sign-in/email-otp 400 in 274ms
 POST /api/auth/sign-in/email-otp 400 in 276ms
 POST /api/auth/sign-in/email-otp 400 in 267ms
 POST /api/auth/sign-in/email-otp 400 in 277ms
 POST /api/auth/sign-in/email-otp 400 in 271ms
 POST /api/auth/sign-in/email-otp 400 in 279ms
 POST /api/auth/sign-in/email-otp 400 in 734ms
```

### What version of Better Auth are you using?

1.2.4

### Provide environment information

```bash
- NextJS 15.2.2
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
export const auth = betterAuth({
	database: drizzleAdapter(db, {
		provider: "pg",
		schema: {
			user: schema.user,
			account: schema.account,
			session: schema.session,
			verification: schema.verification,
		},
	}),
	trustedOrigins: [
		"http://localhost:3000",
		"REDACTED",
		"REDACTED",
		"REDACTED",
	],
	rateLimit: {
		storage: "secondary-storage",
		window: 60, 
		max: 5, 
		customRules: {
			"/sign-in/email-otp": {
				window: 300,
				max: 5,
[...]

PR #1956, opened 2025-03-23
Update rate-limit.mdx

ShahriarKh:
add `enabled: true` to docs

related: #1891


Bekacru:
`enabled:true` is only necessary to force enable on dev. It's by default enabled in prod and disabled in dev.

ShahriarKh:
I updated my changes to reflect that

Kinfe123 (approved):

PR #1959, opened 2025-03-23
fix(rate-limiter): handle missing IP address in rate limit function

### LINKED ISSUE #1958: nullish value on ratelimit key filed

### To Reproduce

Just configure a rate limit on a route (or all routes) and use database as your storage.

-

1.2.4

```bash
- OS [Windows 11]
- Browser: [Chrome]
- API Client: [Better Auth's Scala]
```

Package

```typescript
import { Pool } from "pg";
import { openAPI, phoneNumber } from "better-auth/plugins";

  emailAndPassword: {
    enabled: true,
  },
  rateLimit: {
    storage: "database",
    window: parseInt(Bun.env.RATE_LIMIT_TIME_WINDOW ?? "60"),
    max: parseInt(Bun.env.MAX_TRIES ?? "1"),
    enabled: true,
  },
 database: new Pool({
    connectionString: Bun.env.POSTGRES_STRING,
  }),
plugins: [
    openAPI(),
    phoneNumber({
      sendOTP({ phoneNumber, code }, request) {
        console.log(`Sending code: ${code} to phone number: ${phoneNumber}`);
      },

      signUpOnVerification: {
        getTempEmail: (phoneNumber) => {
          return `${phoneNumber}@example.com`;
        },
        //optionally, you can also pass `getTempName` function to generate a temporary name for the user
        getTempName: (phoneNumber) => {
          return phoneNumber; //by default, it will use the phone number as the name
        },
      },
    }),
  ]
)}
```

### Additional context

Hi, I hope you're having some fun time (like us using better-auth)!
I wanted to mention that there's some nullish value on ratelimit key field when the storage is set to database (haven't tested the custom-storage).

here's the photo:

Image

Bekacru:
closes #1958

### DIFF
--- a/packages/better-auth/src/api/rate-limiter/index.ts
+++ b/packages/better-auth/src/api/rate-limiter/index.ts
@@ -119,7 +119,11 @@ export async function onRequestRateLimit(req: Request, ctx: AuthContext) {
 	);
 	let window = ctx.rateLimit.window;
 	let max = ctx.rateLimit.max;
-	const key = getIp(req, ctx.options) + path;
+	const ip = getIp(req, ctx.options);
+	if (!ip) {
+		return;
+	}
+	const key = ip + path;
 	const specialRules = getDefaultSpecialRules();
 	const specialRule = specialRules.find((rule) => rule.pathMatcher(path));
 


packages/better-auth/src/api/rate-limiter/index.ts
```diff
@@ -119,7 +119,11 @@ export async function onRequestRateLimit(req: Request, ctx: AuthContext) {
 	);
 	let window = ctx.rateLimit.window;
 	let max = ctx.rateLimit.max;
-	const key = getIp(req, ctx.options) + path;
+	const ip = getIp(req, ctx.options);
+	if (!ip) {
+		return;
```
drakmail, line 124: shouldn’t this completely skip rate limits checks if an ip can’t be determined?

PR #1960, opened 2025-03-23
fix(rate-limiter): custom rate limiing table name breaking db query

### LINKED ISSUE #1957: rate limit modelName can't have schema name
rateLimit can only use tables created in `public` schema:
```tsx
    rateLimit: {
        // ...other options
        storage: "database",
        modelName: "rateLimit",             // works
        modelName: "someSchema.rateLimit",  // doesn't work
    },
})
```
Using a custom schema shows this in response:
```
undefined is not an object (evaluating 'schema[model].fields')
```

I think this issue is similar: #1884 

Database: Postgres 16
better-auth: 1.2.4

Bekacru:
closes #1957

### DIFF
--- a/packages/better-auth/src/api/rate-limiter/index.ts
+++ b/packages/better-auth/src/api/rate-limiter/index.ts
@@ -55,7 +55,7 @@ function createDBStorage(ctx: AuthContext, modelName?: string) {
 			try {
 				if (_update) {
 					await db.updateMany({
-						model: modelName ?? "rateLimit",
+						model: "rateLimit",
 						where: [{ field: "key", value: key }],
 						update: {
 							count: value.count,
@@ -64,7 +64,7 @@ function createDBStorage(ctx: AuthContext, modelName?: string) {
 					});
 				} else {
 					await db.create({
-						model: modelName ?? "rateLimit",
+						model: "rateLimit",
 						data: {
 							key,
 							count: value.count,


ShahriarKh:
Awesome speed! However, it seems like there's a regression...
I tried using `1.2.5-beta.8` and rateLimit doesn't work anymore (even when using memory storage). I switched back to `1.2.4` and it started working again.
I didn't change any code, and my rateLimit config is simple:
```
  rateLimit: {
    enabled: true,
    window: 10,
    max: 1,
```
I'm using Elysia 1.2.9 and I don't know if it's related. However, seems like something broke the rate limiting between 1.2.4 and 1.2.5-beta.8

ShahriarKh:
Well, I tested all beta versions (`1.2.5-beta.1` -> `1.2.5-beta.7`) and there's no problem. So, the issue is in `1.2.5-beta.8`

ShahriarKh:
I think this is causing the problem:
https://github.com/better-auth/better-auth/blob/b41febc54f4313395fa0f8644e61dc7c4a0f2f9c/packages/better-auth/src/api/rate-limiter/index.ts#L123-L125

ShahriarKh:
@Bekacru Sorry for the ping; just wanted to make sure you see this since the pr & related issue are both closed.

Bekacru:
@ShahriarKh The code you referenced might be causing the issue, but it's correct. If it can't read the IP, it can't rate-limit it. But I'll check if it's somehow failing to read the IP even when it exists in the request.

ShahriarKh:
Got you. So, in previous versions, if it failed to read the IP the rate limiting would still work; limiting requests from the "`null`" IP. But this treated all clients who doesn't have an IP in their request the same. For example, if I used postman to send requests and somebody else used httpie to do so, we would be treated as one client and our limitations would add up.

With this new update, the rate limiter simply ignores null IPs and continues to handle their requests. This makes it vulnerable and bad actors who remove these [headers](https://github.com/better-auth/better-auth/blob/b41febc54f4313395fa0f8644e61dc7c4a0f2f9c/packages/better-auth/src/utils/get-request-ip.ts#L17C3-L25C15) can simply bypass the rate limiting.

I am not sure if it's an issue in production (I only tested this in development with Scalar where IP is `null`), but if somebody can send a request without IP the rate limit will be of no use.

We have some choices:
- continue to rate limit requests with `null` IP like before.
- reject requests without IP instead of letting them continue:
  ```diff
  if (!ip) {
  - return;
  + return rateLimitResponse(retryAfter);
  }
  ```
- add some checks to allow null IPs in developemnt but reject (or rateLimit them with `null` ip) in production.

ShahriarKh:
btw, if you think we need more investigation, I can create a new issue because others probably don't see our conversation here 😁 (or if you re-open #1958, we can continue there)

PR #1976, opened 2025-03-24
feat: support use number id

Bekacru:
Add the` useNumberId` option under `advanced` to convert id fields referenced in where clauses and similar cases into numbers.

- [x] prisma support
- [ ] drizzle support
- [ ] kysley support
- [ ] mongodb support
- [ ] documentation


ping-maxwell:
👀 Looks like I need to support this in `createAdapter` too

Issue #1118, opened 2025-01-04
Please set BETTER_AUTH_SECRET in your environment variables or pass secret in your auth config.


jeffscottward:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Setup very basic Astro.js and Better-Auth.
Run Astro.js preview command `astro preview` or similar depending on your choice of `pnpm` or `npm` etc 

### Current vs. Expected behavior

To not throw an error. I tried adding the secret to the config, and that didn't help either. 

### What version of Better Auth are you using?

latest as of Jan 3 2025

### Provide environment information

```bash
Mac OS M1 Sonoma 14.6.1
```

### Which area(s) are affected? (Select all that apply)

Other

### Auth config (if applicable)

```typescript
export const auth = betterAuth({
  database: {
    dialect,
    type: "sqlite",
  },
  plugins: [
    magicLink({
      expiresIn: 5 * 60, // 5 minutes
      sendMagicLink: async ({ email, token, url }, request) => {
        await sendReactEmail({
          to: email,
          subject: "Sign in to site",
          text: `Click this link to sign in to site: ${url}`,
          template: MagicLinkEmail({
            magicLink: url,
          }),
        });
      },
    }),
  ],
});
```

### Additional context

.env

BETTER_AUTH_URL=https://sampleurl.com
BETTER_AUTH_SECRET=QD2loiXIqtE8q1ES3Y5VVERFxlkqfLt7

terminal

> astro preview

19:15:10 [@astrojs/node] Server listening on http://localhost:4321

[Better Auth]  ERROR   You are using the default secret. Please set BETTER_AUTH_SECRET in your environment variables or pass secret in your auth config.

Scooter1337:
Make sure your `.env` file contents are loaded into `process.env` using e.g. 
```typescript
dotenv.config();
```

Or something similar.

jeffscottward:
It’s automatic. See the Astro.js docs
https://docs.astro.build/en/guides/environment-variables/


------------------------------

Jeff Scott Ward
Founder / A.I. & U.I. Engineer
*F0.ai*
------------------------------

Scooter1337:
> It’s automatic. See the Astro.js docs https://docs.astro.build/en/guides/environment-variables/

Astro (Vite) loads them into import.meta.env, while better-auth uses process.env. That might be the inconsistency here?

jeffscottward:
I tried the import meta method but it still didn’t take which really
[...]

PR #1997, opened 2025-03-25
feat: add autodetect of better_auth env vars for vite/astro

moshetanzer:
fixes: https://github.com/better-auth/better-auth/issues/1118#


moshetanzer:
Note: In the past this seems to have broken non vite bundlers.

PR #2011, opened 2025-03-26
Update optimizing-for-performance.mdx

ismailsabet:
I'm guessing that's what you meant, because there isn't a `userId` field in the `invitation` table.


Bekacru:
correct. thanks!

PR #2080, opened 2025-04-02
feat(genericOAuth & sso): Optional force user info to be retrieved from endpoint

### LINKED ISSUE #2003: genericOAuth & sso Plugin - Always invoke user info endpoint.

When using the default `responseType: "code"`, both an ID Token and Access token are returned. In certain IDPs ([Okta](https://support.okta.com/help/s/article/attribute-claim-missing-from-id-token?language=en_US)), this results in only a partial profile being returned in the ID Token.

But, because the the token still contains an ID Token, the `userinfo` endpoint is not invoked.

A modification to the [`getUserInfo`](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/generic-oauth/index.ts#L138) function and [`sso`](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/sso/index.ts#L548) to always invoke the `userinfo` endpoint if a flag is set within their respective OAuth configurations.

- Change `responseType` to a value other than `code`
  - Okta requires a nonce to be supplied when using a value other than `code`

- Utilise `getUserInfo`
  - Currently what we're using but is not configurable within the sso plugin
  - Side note - providing the discovered `userInfoUrl` value as a parameter to this signature would also be great.

### Additional context

_No response_

codylittle:
Add an optional boolean config to force the user profile to be retrieved from configured or discovered `userInfo` endpoint.

Docs updated to reflect changes too.

Resolves #2003


codylittle:
@Bekacru does the use of a union type for `getUserInfo` align with the type semantics of the package? Or are there any other changes required?

PR #2085, opened 2025-04-02
docs: fix path of nextJS app router route handler in next.mdx

Yanis64:
Default path of route handler for app router is /app/api

https://nextjs.org/docs/app/building-your-application/routing/route-handlers


Kinfe123:
since it is more of a relative path considering the people would have like including a route groups and others , this is not going to be descriptive enough as well.

PR #2091, opened 2025-04-02
fix(apple): update responseType to include code and id_token

Bekacru:


eYdr1en:
This fix works as expected.  It correctly resolves the Apple OAuth issue for assigning the user's first and last name in Better Auth.

PR #2104, opened 2025-04-03
fix(organization): updateMemberRole failing if issuer has multiple roles

### LINKED ISSUE #2100: updateMemberRole only works if the updating user has a single role

### To Reproduce

To reproduce:

1. Create a better-auth instance with the organizations plugin. I don't think the database matters.
2. Create a organization
3. Update your own member role to two roles
` 
const { error } = await authClient.organization.updateMemberRole({
memberId: "YOUR_MEMBER_ID",
role: ["owner", "admin"]
});
`
4. Do step 3 again

Currently:
The first time the member role is set, the "role" column in the "member" table will be set to "owner,admin", which is what I would expect.
The second time you do it, you get an error ROLE_NOT_FOUND.

I think the code for this endpoint is checking if the role "owner,admin" is present, not either one of them.

1.2.5

```bash
- OS: MacOs 15.4
- Browser: Chrome
```

Backend

```typescript

  database: { dialect, type: "postgres" },
  emailAndPassword: {
    enabled: true,
    disableSignUp: true,
    requireEmailVerification: false,
    maxPasswordLength: 100,
    minPasswordLength: 8,
  },
  socialProviders: {
    microsoft: {
      enabled: true,
      clientId: process.env.MICROSOFT_CLIENT_ID as string,
      clientSecret: process.env.MICROSOFT_CLIENT_SECRET as string,
      tenantId: process.env.MICROSOFT_TENANT_ID as string,

      // tenantId: "common",
    },
  },
  plugins: [organization({ ac, roles }), admin()],
});
```

### Additional context

In my example, I added custom roles to my better-auth instance. This bug occurs even if you didnt. As you can see in the example, I used the standard roles, 'admin', 'owner'

dustin-we:
closes #2100 

This is my first ever contribution to OSS, so I would appreciate feedback.

This fixes the above mentioned issue and also some other stuff issues I noticed in the updateMemberRole endpoint.

### DIFF
--- a/packages/better-auth/src/plugins/organization/organization.test.ts
+++ b/packages/better-auth/src/plugins/organization/organization.test.ts
@@ -335,9 +335,7 @@ describe("organization", async (it) => {
 			},
 		});
 
-		const activeMember = org?.data?.members.find(
-			(m) => m.userId === user.id,
-		)
+		const activeMember = org?.data?.members.find((m) => m.userId === user.id);
 
 		expect(activeMember?.role).toBe("owner");
 
@@ -363,7 +361,6 @@ describe("organization", async (it) => {
 
 		expect(c2.data?.role).toBe("owner");
 	});
-
 
 	const adminUser = {
 		email: "test3@test.com",
@@ -844,6 +841,4 @@ describe("access control", async (it) => {
 		}
 		expect(error).toBeInstanceOf(BetterAuthError);
 	});
-
-	
 });
--- a/packages/better-auth/src/plugins/organization/organization.test.ts
+++ b/packages/better-auth/src/plugins/organization/organization.test.ts
@@ -324,6 +324,47 @@ describe("organization", async (it) => {
 		expect(c.data?.role).toBe("member,admin");
 	});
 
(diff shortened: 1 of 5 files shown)


packages/better-auth/src/plugins/organization/organization.test.ts
```diff
@@ -802,4 +844,6 @@ describe("access control", async (it) => {
 		}
 		expect(error).toBeInstanceOf(BetterAuthError);
 	});
+
```
dsonet: @dustin-we You should remove these extra empty lines and align the code styles here. You can run `pnpm lint` local to check.

dsonet (changes requested):

packages/better-auth/src/plugins/organization/organization.test.ts
dustin-we: Hi, I know this has been done already, but how would I have done that? Can I add commits to a PR?

dsonet: Yes, you can add more commits to a PR. @Bekacru already did that though.

PR #2107, opened 2025-04-03
Fix: Generate valid openapi schema and add `operationId`s

### LINKED ISSUE #2097: OpenAPI plugin generates an invalid schema (Structural and Semantic errors and missing model)

### To Reproduce

1. create backend with default auth config + OpenAPI plugin
2. get the schema in json `const schema = await auth.api.generateOpenAPISchema()`
3. validate the json at e.g. https://editor.swagger.io

The following issues arise:
* `type: date` is invalid and should be `type: string \n format: date-time` (there are about 10 occurrences for this for all the `created_at` etc)
* `/reset-password/{token}` the `token` param is declared but never defined
* `/delete-user` has an empty array for `required`
* When a `secondaryStorage` is defined in the auth config, the `Session` model/schema is omitted from the openapi schema, causing a semantic error
* No routes have an `operationId` set -> this does not cause an error, but would be nice to have when generating api clients based on the openapi schema

1.2.5

```bash
- MacOS 15.4
- Safari 15
```

Documentation, Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
  plugins: [openAPI()]
});
```

### Additional context

It seems someone was already working on fixing some of these things here: https://github.com/better-auth/better-auth/pull/1699 but never got reviewed/merged
Let me know if you'd want a new PR

thomasmol:
Fixes #2097

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/index.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/index.ts
@@ -145,6 +145,7 @@ export const oidcProvider = (options: OIDCOptions) => {
 				"/.well-known/openid-configuration",
 				{
 					method: "GET",
+					operationId: "getOpenIdConfig",
 					metadata: {
 						isAction: false,
 					},
@@ -158,6 +159,7 @@ export const oidcProvider = (options: OIDCOptions) => {
 				"/oauth2/authorize",
 				{
 					method: "GET",
+					operationId: "oauth2Authorize",
 					query: z.record(z.string(), z.any()),
 				},
 				async (ctx) => {
@@ -168,6 +170,7 @@ export const oidcProvider = (options: OIDCOptions) => {
 				"/oauth2/consent",
 				{
 					method: "POST",
+					operationId: "oauth2Consent",
 					body: z.object({
 						accept: z.boolean(),
 					}),
@@ -251,6 +254,7 @@ export const oidcProvider = (options: OIDCOptions) => {
 				"/oauth2/token",
 				{
 					method: "POST",
+					operationId: "oauth2Token",
 					body: z.record(z.any()),
 					metadata: {
 						isAction: false,
@@ -623,6 +627,8 @@ export const oidcProvider = (options: OIDCOptions) => {
 				"/oauth2/userinfo",
 				{
 					method: "GET",
+					operationId: "oauth2Userinfo",
+					use: [sessionMiddleware],
 					metadata: {
 						isAction: false,
 					},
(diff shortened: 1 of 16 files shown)


krzkz94:
Yes please, was thinking I was crazy when openapi-typescript started throwing errors about missing refs 

```
✨ openapi-typescript 7.6.1
 ✘  Can't resolve $ref at #/paths/~1api~1auth~1get-session/get/responses/200/content/application~1json/schema/properties/session
 ✘  Can't resolve $ref at #/paths/~1api~1auth~1list-sessions/get/responses/200/content/application~1json/schema/items
 ✘  Can't resolve $ref at #/paths/~1api~1auth~1sign-in~1anonymous/post/responses/200/content/application~1json/schema/properties/session
 ✘  Can't resolve $ref at #/paths/~1api~1auth~1admin~1list-user-sessions/post/responses/200/content/application~1json/schema/properties/sessions/items
 ✘  Can't resolve $ref at #/paths/~1api~1auth~1admin~1impersonate-user/post/responses/200/content/application~1json/schema/properties/session
```

krzkz94:
@Kinfe123 hey, any chance this could be merged? Would fix an issue with type generation when exporting OpenAPI specs into TS files

himself65:
Hi, I think this is good, could you please rebase the code?

thomasmol:
Ready for review @himself65 !

dilincoln:
Since this fixes the broken open API schema, it will be beneficial to have it on the `latest` channel as soon as possible. Given that this is a fix, should we wait for 1.4?

packages/better-auth/src/plugins/oidc-provider/index.ts
```diff
@@ -902,7 +906,8 @@ export const oidcProvider = (options: OIDCOptions) => {
 				"/oauth2/userinfo",
 				{
 					method: "GET",
-
+					operationId: "oauth2Userinfo",
+					use: [sessionMiddleware],
```
Jackman3005, line 910: Hello 👋

Why was `use: [sessionMiddleware]` added in this PR? I think it may be causing me issues with 401s on my `/oauth2/userinfo` endpoint for a trusted client when I have a valid token...

Jackman3005: @thomasmol  @Bekacru

dilincoln:
The only error that I'm facing now is:

image

A workaround for me was patching `path` objects to inject the `items` prop:

```ts
const addItemsForArrays = (obj: unknown): OpenAPIV3.PathsObject => {
	const seen = new WeakSet<object>()

	function walk(node: unknown) {
		if (node && typeof node === "object") {
			const o = node as Record<string, unknown>
			if (seen.has(o)) {
				return
			}

			seen.add(o)

			const t = (o.type ?? o.types) as unknown
			if (t === "array" && !Object.hasOwn(o, "items")) {
				o.items = {}
			}

			for (const k of Object.keys(o)) {
				walk(o[k])
			}

			if (Array.isArray(o)) {
				for (const el of o) walk(el)
			}
		}
	}

	walk(obj)

	return obj as OpenAPIV3.PathsObject
}
```

packages/better-auth/src/plugins/oidc-provider/index.ts
Jackman3005: I'm not sure if there is a situation where this endpoint wants to authenticate via cookies, but even if so, it can't reject with 401 before checking the `Authorization: Bearer xxx` header as it does currently. I've created an [issue to track/fix this](https://github.com/better-auth/better-auth/issues/6284).

PR #2140, opened 2025-04-05
docs(example): add example project for Hono and D1 integration

shinaps:


himself65:
We move it to another repo

https://github.com/better-auth/examples

PR #2158, opened 2025-04-06
refactor: `betterAuth` fn now infers models

ping-maxwell:
We now infer models based on the user options: 

image

Previously:

image

---

## How?

The `betterAuth` function now has generics for plugins, as well as any model data that could be modified by `options`, such as User additional fields.

Changes are straightforward:
* All `BetterAuthOptions` are moved straight into the `betterAuth` function parameters.
* The `betterAuth` function has generics that track any model inference related options.
* The old `BetterAuthOptions` type export is replaced with just a `Parameters<typeof betterAuth>[0]` sort of thing.
  * That `BetterAuthOptions` **also** includes all generics similar to `betterAuth`, but has default values so don't worry about having each possible `BetterAuthOptions` needing to provide generic values, nothing should break. I made it possible to pass the generics similar to `betterAuth` for the purpose of type inferencing within the `betterAuth` function code.

The result is a nice experience working with the `betterAuth` function knowing full-well that any methods within the options will be type-safe regarding model objects.

However this does come with a cost:
Any function which in-takes a `BetterAuthOptions` **will not** be equal to the `options` we'll end up getting from the `betterAuth` function parameters.
This is because our `options` from the BA function has generics, thus meaning some values could be dynamic/different.
In reality, they will 100% be the same/equal, if comparing BetterAuthOptions to the result `options` object.
The reason I say this is a cost is because anytime we have to call a function within the `betterAuth` function, and it in-takes `options` that's of type `BetterAuthOptions`, it will throw a type error saying it doesn't match.
To reemphasize, this is just because there is a difference in generics.

My solution to this is just to use a ts-ignore, since they're technically still the same BetterAuthOptions.
The alternative solution is to have each of those function to also in-take every possible generic to build the same `BetterAuthOptions` to equal `options`.
But this would mean we need to do these generics for each function we end up using. Not to mention maintaining this.

Using `ts-ignore` was just easier. It's just a type-err within the `betterAuth` function, so no end-users would deal with this.


numman-ali:
Awesome stuff thank you for making better auth even better !

Would this support naming of plugin namespace and methods ie organizations to tenants?
Example would be auth.api.tenants instead of auth.api.organisations?

ping-maxwell:
> Awesome stuff thank you for making better auth even better !
> 
> Would this support naming of plugin namespace and methods ie organizations to tenants? Example would be auth.api.tenants instead of auth.api.organisations?

No, sorry.

ping-maxwell:
@Bekacru After the merge of this, we can finally make signUp method to infer plugin schemas. Since if a plugin extended user table, and had the correct `input` option, it makes it allowed to be passed by signUpEmail, but right now there isn't any inference on that.

ping-maxwell:
I intend for this PR to be merged with my other to have any-model DB hook support.

TODO: This means I need to add proper types to support any model in DB hooks config.

AdolfoRoquero:
@ping-maxwell Thanks for fixing it! What is the status on this? I am running into the issue and was wondering if this was going to be merged at some point soon.

ping-maxwell:
Hey @adolfoRoquero, it's very unlikely this will be merged any time soon.<br>There are type issues in this PR that are not fixed, but what I mean is that it would require quite a big refactor on Better Auth to actually achieve this. Maybe for a later date at a larger release milestone this will be available.

DiegoGonzalezCruz:
Oh no I was expecting this to be solved! :'(

ping-maxwell:
PR very out-dated. Will need a new one.

PR #2161, opened 2025-04-07
fix(drizzle adapter): Add missing where operators

Suu-ly:
This pull request adds in the missing `gt`, `gte` and `ne` operators for the drizzle adapter which previously would have used the `eq` filter instead.


Kinfe123:
it is already fixed!

Issue #2171, opened 2025-04-08
Stripe Plugin: deprecated the current_period_start and current_period_end fields


Evy04:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Try create new stripe account and setup the plugin.
2. Create a plan then signup for that plan
3. Make sure that webhook is running and recieving notification.
4. Notice that it fails with
```JS 
3149 const transformed = transformInput(update, model, "update");
→ 3150 const result = await db[getModelName(model)].update({
         where: {
           id: "cus_S5ZkzLrh2oHinN"
         },
         data: {
           plan: "pro",
           stripeSubscriptionId: "sub_1RBPxKLW6jnevwpozuutGeB4",
           status: "active",
           periodStart: new Date("Invalid Date"),
                        ~~~~~~~~
           periodEnd: new Date("Invalid Date"),
           seats: 1
         }
       })
```

### Current vs. Expected behavior

It should create subscription without any issue.

### What version of Better Auth are you using?

@better-auth+stripe@1.2.5_typescript@5.8.2

### Provide environment information

```bash
- Mac OS
- Chrome
```

### Which area(s) are affected? (Select all that apply)

Package

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
  plugins: [
    stripe({
      stripeClient,
      stripeWebhookSecret: process.env.STRIPE_WEBHOOK_SECRET!,
      createCustomerOnSignUp: true,
      subscription: {
        enabled: true,
        plans: async () => {
          const products = await stripeClient.products.list();

          const plans = products.data.map((product) => ({
            name: product.name,
            priceId: product.default_price as string | undefined,
          }));

          return plans;
        },
      },
    }),
     
  ],
});
```

### Additional context

https://docs.stripe.com/changelog/basil/2025-03-31/deprecate-subscription-current-period-start-and-end

Evy04:
Note: temp fix- Downgrade to stripe "stripe": "17.7.0",

Bekacru:
this should be fixed since 1.2.7

PR #2173, opened 2025-04-08
Fix stripe webhook handler

Evy04:
Temp fix for #2171

Not sure if multiple items needs to be handled, since sub will likely not have multiple


himself65:
outdated. I think it's not the issue now

feel free to re-open it when you are ready

PR #2213, opened 2025-04-10
Corrections

DePasqualeOrg:
There are a lot of small mistakes throughout the docs, which I'm happy to correct if you're open to it. This is just a small sample to get your reaction before I spend more time on this.


DePasqualeOrg:
@Bekacru, I'm reading through the docs as I start to use Better Auth. Could you let me know if you're okay with corrections like the ones I've already committed?

Also, I'm wondering why the title of the tab on the website's landing page is `helo_`. It's not clear to me what this means.

Kinfe123 (approved):

PR #2218, opened 2025-04-11
fix(vue): baseURL not working

yooouuri:
~/utils/auth.ts

```ts
import { createAuthClient } from 'better-auth/vue'

const { useSession, signIn, signUp } = createAuthClient({
  baseURL: 'http://localhost:8000',
})

export default {
  useSession,
  signIn,
  signUp,
}
```

page.vue

```ts
const { data: session } = await auth.useSession(useFetch)
```

The baseURL is not merged into the actual url for the API call.

image

### DIFF
--- a/packages/better-auth/src/client/vue/index.ts
+++ b/packages/better-auth/src/client/vue/index.ts
@@ -17,6 +17,7 @@ import type {
 	BetterFetchResponse,
 } from "@better-fetch/fetch";
 import type { BASE_ERROR_CODES } from "../../error/codes";
+import { getBaseURL } from "../../utils/url";
 
 function getAtomKey(str: string) {
 	return `use${capitalizeFirstLetter(str)}`;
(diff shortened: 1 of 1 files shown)


Nickersoft:
I'm currently using the Svelte client and think I am noticing the same behavior.. 🤔

ammarrik:
i set up the react frontend on `localhost:3000` and hono backend on `locahost:3001`, even though i set the `baseURL` to `localhost:3001`, better-auth still send requests to `3000`.

himself65:
> i set up the react frontend on `localhost:3000` and hono backend on `locahost:3001`, even though i set the `baseURL` to `localhost:3001`, better-auth still send requests to `3000`.

Hi, what's your setup here?

himself65 (commented):
I think we should use baseURL from $fetch. I think this should be fixed in the better-fetch first

PR #2221, opened 2025-04-11
Update hero.tsx - Fix typo in hero section - "npm add better-auth" -> "npm install better-auth"

ferran9908:
I know that `npm add` is an alias for `npm install` but a lot of beginners would get confused when trying out the package.


Kinfe123:
since it copiable thing and it is also something they need to know as begineer. `add` is just fine. but if we think that it is creating a lot of confusion will go with this one

PR #2228, opened 2025-04-11
feat(types): Add a type for shared options

CrutchTheClutch:
This is a small quality of life change to provide a type that is exclusively for configuration options that are shared between better-auth and the better-auth client

### DIFF
--- a/packages/better-auth/src/types/options.ts
+++ b/packages/better-auth/src/types/options.ts
@@ -19,15 +19,7 @@ import type { Logger } from "../utils";
 import type { AuthMiddleware } from "../plugins";
 import type { LiteralUnion, OmitId } from "./helper";
 
-export type BetterAuthOptions = {
-	/**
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/types/options.ts
```diff
@@ -47,6 +39,17 @@ export type BetterAuthOptions = {
 	 * @default "/api/auth"
 	 */
 	basePath?: string;
+};
+
+export type BetterAuthOptions = BetterAuthSharedOptions & {
```
himself65: It's better to use interface and extend the SharedOptions

himself65 (changes requested):

CrutchTheClutch:
I recreated the branch from the current canary and cherry picked the commits @himself65

himself65:
Could you please rebase the code?

PR #2234, opened 2025-04-12
feat(two-factor): refactor two-factor authentication with better error handling, configurable otp limits and verification

Bekacru:


aymericfontaine:
@Bekacru

I noticed that in a recent change, the `period` value in `otpOptions` has been switched from minutes to seconds. I only realized this by testing it myself.

Are there any release notes or changelogs where such breaking changes are documented, so we can better anticipate them in the future?

Thanks!

PR #2246, opened 2025-04-13
Correct Svelte Kit to SvelteKit

gandandev:
Corrected all instances of “Svelte Kit” to “SvelteKit” in documentation for consistency with official branding.
Did not rename svelte-kit to sveltekit in code as it could introduce breaking changes.


DePasqualeOrg:
I already did this, along with many other such corrections, here: https://github.com/better-auth/better-auth/pull/2213

Issue #2250, opened 2025-04-13
Checking if User is intended recipient of Invite is casesensetive


SNRSE:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Setup Better Auth with Organisations and email and password
2.  create a user with an Email that has upercase letters
3.  invite that user to an Organisation 
4. try to accept the invitation from this User

### Current vs. Expected behavior

Right now, inviting Mail@example.com will cause an issue for the user mail@example.com when trying to accept it. For logging In the email gets converted toLowerCase so the same behaviors should happen when checking if the user is the recipient of an Invite

### What version of Better Auth are you using?

1.2.5

### Provide environment information

```bash
does not apply as it is an issue with the backend
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});
```

SNRSE:
I have also created the following PR to adress this Bug : https://github.com/better-auth/better-auth/pull/2251

SNRSE:
PR is merged so closing the issue

PR #2251, opened 2025-04-13
Fix for #2250 : Checking if User is intended recipient of Invite is casesensetive

SNRSE:
This resolves a bug where, while you can use 
Email@example.com to log into the account email@example.com. Inviting Email@example.com would mean they can not accept the invite. 

This is done by using .tolowercase on both emails before comparing them


SNRSE:
Resolved the issue th eboth highlited, and checked the file to make sure all occurrences of the check behave the same 
+ now it is also converted to lowercase on creation.

PR #2255, opened 2025-04-13
fix(haveIBeenPwned): Meaningful error code

JE4GLE:
In my humble opinion, `THE_PASSWORD_YOU_ENTERED_HAS_BEEN_COMPROMISED_PLEASE_CHOOSE_A_DIFFERENT_PASSWORD` is **not** a valid error code.

Error codes should be short and are not required the be understand by an end user.

This MR changes the error code to `PASSWORD_COMPROMISED`.

### DIFF
--- a/packages/better-auth/src/plugins/haveibeenpwned/index.ts
+++ b/packages/better-auth/src/plugins/haveibeenpwned/index.ts
@@ -5,7 +5,8 @@ import type { BetterAuthPlugin } from "../../types";
 import { createAuthMiddleware } from "../../api";
 
 const ERROR_CODES = {
-	PASSWORD_COMPROMISED: "Password compromised",
+	PASSWORD_COMPROMISED:
+		"The password you entered has been compromised. Please choose a different password.",
 } as const;
 
 async function checkPasswordCompromise(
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/plugins/haveibeenpwned/index.ts
```diff
@@ -5,8 +5,7 @@ import type { BetterAuthPlugin } from "../../types";
 import { createAuthMiddleware } from "../../api";
 
 const ERROR_CODES = {
-	PASSWORD_COMPROMISED:
-		"THE_PASSWORD_YOU_ENTERED_HAS_BEEN_COMPROMISED_PLEASE_CHOOSE_A_DIFFERENT_PASSWORD",
+	PASSWORD_COMPROMISED: "PASSWORD_COMPROMISED"
```
Bekacru: this is meant to be `code:message`

JE4GLE: Then we need to change something else, because I get the following:
```json
{
    code: "THE_PASSWORD_YOU_ENTERED_HAS_BEEN_COMPROMISED_PLEASE_CHOOSE_A_DIFFERENT_PASSWORD",
    message: "My custom message",
    status: 400,
    statusText: "Bad Request"
}
```

JE4GLE: See [the test run](https://github.com/better-auth/better-auth/actions/runs/14429918450/job/40463461079) on your commit

Bekacru: if you don't provide a code when you throw an error, it just defaults to merging the error message with `_` and all capital. Need to explicitly pass the error code.

PR #2265, opened 2025-04-13
feat(stripe): Adding priceId in db (as in the docs schema)

Konixy:
lmk if a forgot anything

### DIFF
--- a/packages/stripe/src/hooks.ts
+++ b/packages/stripe/src/hooks.ts
@@ -44,6 +44,7 @@ export async function onCheckoutSessionCompleted(
 							periodStart: new Date(subscription.current_period_start * 1000),
 							periodEnd: new Date(subscription.current_period_end * 1000),
 							stripeSubscriptionId: checkoutSession.subscription as string,
+							priceId,
 							seats,
 							...trial,
 						},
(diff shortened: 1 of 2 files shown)


Konixy:
> I'm not sure we should store the price ID in the database, since right now. it's in the option config, so basically it's the fixed value

I believe it is much more developer-friendly because having multiple plans with yearly and monthly subscriptions for each plan saves time by storing them directly in the database. Additionally, I think it was previously included in the `Schema` section of [the docs](https://www.better-auth.com/docs/plugins/stripe#subscription), which is why I made this pull request.

bytaesu:
Hello @Konixy,

I don’t think this needs to be included in the database. Isn’t `priceId` usually managed as a constant in the app? 🤔

Konixy:
> Hello @Konixy,
> 
> 
> 
> I don’t think this needs to be included in the database. Isn’t `priceId` usually managed as a constant in the app? 🤔

It is, but subscriptions can have different priceIds depending on the user configuration. That's why it's more convenient to have it on the subscription object coming from the db imo

himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #2280, opened 2025-04-14
feat(stripe): improve support for usage based pricing

Pukimaa:
DRAFT: Improve support for usage based pricing, by allowing seats to be ommited, which is required when creating a subscription with a usage-based price id.


Pukimaa:
due to the complexity, usage based pricing will require a lot more work than i anticipated. i might open up a new pr, when i have the time and energy to work on supporting usage based pricing in better auth. please contact me, if you want to work on it too!

PR #2295, opened 2025-04-15
feat: make overriding oauth map fields smarter 

Bekacru:


packages/better-auth/src/oauth2/link-account.ts
```diff
@@ -104,18 +104,37 @@ export async function handleOAuthUserInfo(
...
+			id: _,
+			name,
+			email,
+			emailVerified,
+			image,
+			createdAt,
+			...restUserInfo
+		} = userInfo;
+		const coreFields =
+			typeof overrideUserInfo === "boolean"
```
dsonet: What if `overrideUserInfo = false` is passed here? The implementation returns the same as if it’s true.

dsonet (changes requested):

PR #2311, opened 2025-04-16
Docs: Delete nonsense word `helo_` from navigation bar

DePasqualeOrg:
Currently the website uses the nonsense word `helo_` (or `_helo` on mobile) to navigate to the root page. This is redundant, since clicking on the Better Auth logo also navigates to the root page. It also doesn't make sense to use a nonsense word for this.

### DIFF
--- a/docs/components/nav-bar.tsx
+++ b/docs/components/nav-bar.tsx
@@ -108,10 +108,6 @@ export const Navbar = () => {
 
 export const navMenu = [
 	{
-		name: "helo_",
-		path: "/",
-	},
-	{
 		name: "docs",
 		path: "/docs",
 	},
--- a/docs/components/nav-mobile.tsx
+++ b/docs/components/nav-mobile.tsx
@@ -201,11 +201,6 @@ export const navMenu: {
 	}[];
 }[] = [
 	{
-		name: "_helo",
-		path: "/",
-	},
-
-	{
 		name: "docs",
 		path: "/docs",
 	},


DePasqualeOrg:
@Kinfe123 @Bekacru, your website has a nonsense word prominently displayed at the top. It looks bizarre and unprofessional. The best solution is to delete it, which I did in this PR, since users can already navigate to the root page by clicking on the logo. I'm just trying to help you look good. Why are you insisting on keeping this nonsense word? Just take the win and merge this.

DePasqualeOrg:
@Bekacru, @Kinfe123, there's no good reason to keep this nonsense word in your navigation bar. I don't understand why you can't just take the win and merge this.

Bekacru:
@DePasqualeOrg hey respectfully, we do whatever we see fit :)) If we thought it should be removed, we wouldn’t have waited for your input. Thanks and please stop tagging us and other contributors regarding this

DePasqualeOrg:
It seems to be emblematic of a careless approach to the docs in general, and it makes me wonder about other aspects of this project. I've already spent a lot of time fixing careless mistakes in the docs that should have been easily avoidable. If you just close PRs like this that are easy wins, it doesn't reflect well on this project.

DePasqualeOrg:
I see you [decided](https://github.com/better-auth/better-auth/commit/e5909128ca7ecdf0ef68558ac7aa9888e7663233#diff-30ffcdc02c84d14f17a87b5e8fcc9672968abd408324ef5fc313551033f04f8aR13) this change was correct after all, @Bekacru.

PR #2316, opened 2025-04-16
feat/stripe-update

kkMihai:
Dependency Upgrade:

* [`packages/stripe/package.json`](diffhunk://#diff-2757b5358a7885372a3e494ce4ec980483ceda7522dd1b61eb52e151772ac9e1L49-R49): Upgraded the `stripe` package from version `17.7.0` to `18.0.0`.

Code Consistency:

* [`packages/stripe/src/hooks.ts`](diffhunk://#diff-72e8dafae00019b1cb067d0577a11924409d13aacf09333fcd9406aa5389b8f4L1-R1): Reordered imports for better readability and consistency.
* [`packages/stripe/src/index.ts`](diffhunk://#diff-ccb5282d4490ceec8e8c99f978be2383dd1915db21316f3baba6066420070b08L2-R13): Reordered imports for better readability and consistency. [[1]](diffhunk://#diff-ccb5282d4490ceec8e8c99f978be2383dd1915db21316f3baba6066420070b08L2-R13) [[2]](diffhunk://#diff-ccb5282d4490ceec8e8c99f978be2383dd1915db21316f3baba6066420070b08L262-R262) [[3]](diffhunk://#diff-ccb5282d4490ceec8e8c99f978be2383dd1915db21316f3baba6066420070b08L326-R326) [[4]](diffhunk://#diff-ccb5282d4490ceec8e8c99f978be2383dd1915db21316f3baba6066420070b08L339) [[5]](diffhunk://#diff-ccb5282d4490ceec8e8c99f978be2383dd1915db21316f3baba6066420070b08L946-R947)

Subscription Period Date Calculation:

* [`packages/stripe/src/hooks.ts`](diffhunk://#diff-72e8dafae00019b1cb067d0577a11924409d13aacf09333fcd9406aa5389b8f4L44-R49): Updated the logic for calculating `periodStart` and `periodEnd` dates using `subscription.items.data[0].current_period_start` and `subscription.items.data[0].current_period_end` respectively. [[1]](diffhunk://#diff-72e8dafae00019b1cb067d0577a11924409d13aacf09333fcd9406aa5389b8f4L44-R49) [[2]](diffhunk://#diff-72e8dafae00019b1cb067d0577a11924409d13aacf09333fcd9406aa5389b8f4L141-R150)
[...]

PR #2318, opened 2025-04-16
fix(docs): make it clear that we use an outdated stripe version

Pukimaa:
This should resolve confusion with errors related to breaking changes that were introduced in the basil api version of stripe.

> [!CAUTION]
> **This PR should not be merged or reverted, if #2316 gets merged!**

If the next (beta) release will take a while to be put out it might make sense to merge this PR in order to have the docs reflect the current state until the next version. Then in #2316 the changes made via this PR needs to be removed.


Pukimaa:
Commenting #2316 to reference, because of the notice in my pr description

dsonet:
@Bekacru It seems this one should be reverted, since strip already got updated to v18 in https://github.com/better-auth/better-auth/pull/2366.

PR #2325, opened 2025-04-16
feat(two-factor): Added hook on 2fa disable

JE4GLE:
My Use Case: I want to send an email to the user when 2FA has been disabled. Relying on client-side actions to send trigger the sending of an email is too unreliable and insecure.


benoitcanton:
Alternatively, you could use the existing hooks.

```
	if (ctx.path === "/two-factor/disable") {
		await sendEmail(ctx.context.session)
	}
}),
```

PR #2326, opened 2025-04-16
fix(oidc-provider): Update advertised ID token signing alg to HS256

benoitcanton:
The current generated key used to sign the ID token, is using `HS256` algorithm while the openid-configuration endpoint is advertising `RS256` as the only signing algorithm.

Line 598:
```
let secretKey = {
        alg: "HS256",
        key: await subtle.generateKey(
	        {
		        name: "HMAC",
		        hash: "SHA-256",
	        },
	        true,
	        ["sign", "verify"],
        ),
};
```

While it doesn't address the fact that the key used is not one of the keys advertised in the JWKS endpoint (probably due to the fact that they are not from the same plugin?)-At least, now, the `alg` property of the ID token header is matching the openid-configuration.

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/index.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/index.ts
@@ -45,7 +45,7 @@ const getMetadata = (
 			"urn:mace:incommon:iap:bronze",
 		],
 		subject_types_supported: ["public"],
-		id_token_signing_alg_values_supported: ["RS256", "none"],
+		id_token_signing_alg_values_supported: ["HS256", "none"],
 		token_endpoint_auth_methods_supported: [
 			"client_secret_basic",
 			"client_secret_post",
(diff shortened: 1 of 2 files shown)


Bekacru:
The current OIDC plugin only signs the ID token using the shared client secret. We originally planned to support JWKS and RS256 shortly after launch, which is why we didn’t update the open id config. our general assumption was that most clients wouldn’t verify the signature for now. Especially since we only support the authorization code flow, the ID token can generally be trusted.

That said, the ideal solution going forward is to move ID token signing to RS256 and expose a JWKS endpoint.

benoitcanton:
> The current OIDC plugin only signs the ID token using the shared client secret.

I must be missing this part... I can't see it.

> That said, the ideal solution going forward is to move ID token signing to RS256 and expose a JWKS endpoint.

That would be great, how do you see a scenario where both JWT and OIDC provider plugin are installed, would they both  have their own JWKS endpoint / set of keys? Both `/api/auth/jwks` and `/api/auth/oauth2/jwks` would be exposed? The JWKS endpoint of the OIDC provider plugin can probably borrow a lot from the JWT plugin JWKS endpoint...

Bekacru:
Thanks for the PR! closed in favor of https://github.com/better-auth/better-auth/pull/2755

PR #2330, opened 2025-04-17
Fixed error when origin check fails

### LINKED ISSUE #2329: Origin Check Middleware Fails with Query Parameters in Relative URLs

### To Reproduce

Ex:

1. Add url parameter to redirectTo field of forgetPassword
2. Try to send forget password email
3. It works if you add in base url manually

Currently you get error that full url including url parameters is not trusted origin, should strip parameters and then check.

1.2.7

```bash
- Browser Safari
```

Package

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

Image

Image

ejscheepers:
Closes #2329


ejscheepers:
Thanks I see you fixed my pull request.

PR #2342, opened 2025-04-17
Add context parameter to findUserByEmail calls, exposed custom findUser function

PodkopovP:


packages/better-auth/src/types/options.ts
```diff
@@ -362,6 +362,12 @@ export type BetterAuthOptions = {
 			 */
 			deleteTokenExpiresIn?: number;
 		};
+		/**
+		 * A function that is used to override the default user finding logic. The return type of this function should match the User type given to Better Auth.
+		 *
+		 * @default adapter.findOne({ model: "user", where: [{value: email.toLowerCase(), field: "email"}]})
+		 */
+		findUser?: (ctx: GenericEndpointContext) => Promise<any>;
```
himself65, line 370: I think this is not a good practice. It adds too much complexity and we have to bypass the ctx to cross all functions.

PR #2354, opened 2025-04-18
feat(adapter): export kysely adapter.

chaozwn:
export kysely adapter.

### DIFF
--- a/packages/better-auth/package.json
+++ b/packages/better-auth/package.json
@@ -591,6 +591,9 @@
       "adapters/drizzle": [
         "./dist/adapters/drizzle-adapter/index.d.ts"
       ],
+      "adapters/kysely": [
+        "./dist/adapters/kysely-adapter/index.d.ts"
+      ],
       "adapters/mongodb": [
         "./dist/adapters/mongodb-adapter/index.d.ts"
       ],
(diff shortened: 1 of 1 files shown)


ping-maxwell:
Any particular reason to export the kysely adapter?

chaozwn:
> Any particular reason to export the kysely adapter?

Because I use kysely as my orm, I can 't use it because it' s not exported.
image

ping-maxwell:
It's built-in
image

chaozwn:
> It's built-in image

Then I don't quite understand why a separate adaptation is needed here. I think this is the entrance for automatically inserting login information when logging in.
image

ping-maxwell:
It's used internally.

chaozwn:
> It's used internally.

Thank you for giving the relevant hints. I used the solution given in the above picture, but the table authentication that comes with better auth seems to be camel-case. Can it be modified to snake_case.

image

chaozwn:
> It's used internally.

Frankly speaking, I don't think so. If I have a database connection for Kysely that can be converted for use by Better-Auth's database, that's the role of the Kysely adapter. Because the method given in the diagram above actually provides Better-Auth with a new connection, without handing it over to Kysely to manage.

ping-maxwell:
Could this help? 
image

PR #2361, opened 2025-04-19
feat(anonymous): Custom anonymous names

ping-maxwell:
In some cases, `name` is unique in the DB, so this allows the dev to configure generating a random name via the plugin config.
If suitable, in the future we can consider customizing `name`  (and maybe `image` too)  straight from the client's `signIn` method- but it's up for debate.

image

- [x] Docs
- [x] Tests

For further context:
https://discord.com/channels/1288403910284935179/1363063581784018974

### DIFF
--- a/packages/better-auth/src/plugins/anonymous/index.ts
+++ b/packages/better-auth/src/plugins/anonymous/index.ts
@@ -14,6 +14,7 @@ import type {
 import { parseSetCookieHeader, setSessionCookie } from "../../cookies";
 import { getOrigin } from "../../utils/url";
 import { mergeSchema } from "../../db/schema";
+import { z } from "zod";
 
 export interface UserWithAnonymous extends User {
 	isAnonymous: boolean;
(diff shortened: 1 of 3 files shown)


packages/better-auth/src/plugins/anonymous/index.ts
```diff
@@ -43,6 +43,12 @@ export interface AnonymousOptions {
 	 * Disable deleting the anonymous user after linking
 	 */
 	disableDeleteAnonymousUser?: boolean;
+	/**
+	 * A hook to generate a name for the anonymous user.
+	 * Useful if you want to have random names for anonymous users, or if `name` is unique in your database.
+	 * @returns The name for the anonymous user.
+	 */
+	generateName?: () => string;
```
Bekacru: I think it should accept the ctx object or at least the reques obj

ping-maxwell:
@Bekacru It's ready.

Issue #2263, opened 2025-04-13
Anonymous plugin unique name


ywyher:
when you create a user the name field is always Anonymous
what if i want the name field to be unique ?
it would be better if we have an option to enable randomly generated names something like `Anonymous${random}` or any other way to customize the name filed

drago1520:
yes!!!! yes!

drago1520:
I searched through their docs. It is not supported currently. Having unique usernames is such a basic feature.

drago1520:
I fixed it. Now it generates a UUID string with postfix, but you can also add custom default username.

ping-maxwell:
https://github.com/better-auth/better-auth/pull/2361

ping-maxwell:
@drago1520 Sorry, just only noticed now you have a PR for this.
I made mine from a request via the Better Auth Discord server.

Bekacru:
#2361

PR #2362, opened 2025-04-19
feat(anonymous): ability to add custom default username for anonymous users.

drago1520:
As a database engineer, I want to have a UNIQUE constraint on my users' names so they do not duplicate. 
Before, every new Anonymous user had the same username: "Anonymous." Now, a unique string with a postfix is generated for each user's name.

#2263


drago1520:
This is a super quick fix. Ping me and I will be here immediately.

PR #2373, opened 2025-04-20
update(api-key): Disable hashing API Keys

ping-maxwell:
Allows the user to disable hashing of api keys.

- [x] docs
- [x] tests 

For further context:
https://discord.com/channels/1288403910284935179/1288403910284935182/1363454834740559952


packages/better-auth/src/plugins/api-key/utilts.ts
```diff
@@ -0,0 +1,12 @@
+import { base64Url } from "@better-auth/utils/base64";
+import { createHash } from "@better-auth/utils/hash";
+
+export const defaultKeyHasher = async (key: string) => {
```
Bekacru: I'd prefer if we don't have a utils file for plugins. Let's keep it the old way :))

packages/better-auth/src/plugins/api-key/types.ts
```diff
@@ -7,6 +7,15 @@ export interface ApiKeyOptions {
 	 * @default "x-api-key"
 	 */
 	apiKeyHeaders?: string | string[];
+	/**
+	 * Disable hashing of the API key.
+	 *
+	 * **Security Warning:** It's strongly recommended to not disable hashing.
```
Bekacru: let's use 
⚠️ Security Warning: //...
instead of '**'

PR #2381, opened 2025-04-21
Update api-key.mdx

awesomepandapig:
Modify documentation to reflect correct HTTP method.


ping-maxwell:
Hey @awesomepandapig, internally we do use `POST` as that's what Better-Auth supports for now.
Maybe in the future when BA supports `DELETE` we will move over to that.

In the mean time I'm going to close this as `POST` is correct.

PR #2399, opened 2025-04-22
Support literal/enum type for drizzle schema generator

ywyher:
currently the literal type for additional fields doesn't work for the drizzle schema generator and just throws an error
this PR add support for it


ywyher:
added `pgEnum()` instead of `text({ enum: [] })`
if you think that `convertToCamalCase()` isn't necessary i could remove it

himself65:
I think we already have enum support now

PR #1742, opened 2025-03-09
Fix bugs in SvelteKit example

phanect:
This PR fixes bugs in the SvelteKit example, which is not working properly at all now.

I have not fixed all the TypeScript errors yet, but they probably do not affect the behavior as far as I checked. I guess they are unused components. If I have a change, I will fix those errors or clean up unused code later.

PR #2404, opened 2025-04-22
Fix SvelteKit example

DePasqualeOrg:
The SvelteKit example was broken and wrong in a number of ways. It could still be cleaned up, but at least it builds now.


examples/svelte-kit-example/src/routes/sign-up/+page.svelte
```diff
@@ -1,33 +1,35 @@
...
-		callbackURL: "/",
-		fetchOptions: {
-			onError(context) {
-				alert(context.error.message);
-			},
-		},
-	});
-};
+  // Function to handle form submission
+  const handleSignUp = async () => {
```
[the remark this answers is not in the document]
DePasqualeOrg: AI slop

examples/svelte-kit-example/src/routes/sign-up/+page.svelte
```diff
@@ -66,12 +68,12 @@
...
           placeholder="m@example.com"
           required
-          bind:value={$email}
+          bind:value={email}
         />
       </div>
       <div class="grid gap-2">
         <Label for="password">Password</Label>
-        <Input id="password" type="password" bind:value={$password} />
+        <Input id="password" type="password" bind:value={password} />
```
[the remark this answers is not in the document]
DePasqualeOrg: AI slop

DePasqualeOrg:
@Bekacru @Kinfe123, your SvelteKit example is completely broken, and some of the code examples are not correct. I spent a lot of time fixing this for you, and I just resolved some merge conflicts that emerged after I submitted this PR for review. Could you please review and merge this before more conflicts emerge?

Kinfe123:
@DePasqualeOrg will look into it.

vladshcherbin:
Why does this example need a hundred of UI components and packages? Just show some basic UI and actual auth setup without bloat.

It's insane 🤯

DePasqualeOrg:
It doesn't need most them. It only uses a few, so I removed all the unused ones.

DePasqualeOrg:
@Bekacru @Kinfe123, I submitted this fix more than 6 weeks ago, and you have ignored it. It's a pretty big red flag to me that something like this got merged in the first place, and that the fix simply gets ignored.

DePasqualeOrg:
The SvelteKit example was broken, among other reasons, because Biome is not compatible with Svelte files and broke those files whenever the linter fixes were applied. I have added separate linting and formatting (eslint and prettier) just for the SvelteKit example. The CI tests are now passing.

@Bekacru @Kinfe123, it would be great to review this soon, since already more than 6 weeks have passed since I submitted it, and the more time that goes by, the more merge conflicts there are to be resolved. I think it's important to have a working SvelteKit example, and I've invested a lot of time in making that happen. Please take a look at this.

DePasqualeOrg:
@Bekacru @Kinfe123, over three months after I submitted this fix, your SvelteKit example is still completely broken.

DePasqualeOrg:
@Bekacru @Kinfe123, I see that you merged https://github.com/better-auth/better-auth/pull/1742 without engaging here. However, the SvelteKit example is still broken due to the issues with Biome that I mentioned above. I have rebased this PR and fixed those issues here, so that the example now builds and runs. However, I noticed that there are still issues with the sign-up flow. I would appreciate it if you could finally address this PR.

Bekacru:
@DePasqualeOrg sorry about leaving you hanging here. Can you please fix the CI?

DePasqualeOrg:
It looks like it's failing because of a frozen lockfile. How can this be fixed?

Bekacru:
re-install locally and make sure to commit the lock file

DePasqualeOrg:
The CI tests are now passing.

package.json
```diff
@@ -9,9 +9,9 @@
         "dev": "turbo --filter \"./packages/*\" dev",
         "dev:dts": "turbo --filter \"./packages/*\" dev:dts",
         "clean": "turbo --filter \"./packages/*\" clean && rm -rf node_modules",
-        "format": "biome format . --write",
-        "lint": "biome check .",
-        "lint:fix": "biome check . --apply",
+        "format": "biome format . --write && pnpm --filter @better-auth/svelte-kit-example format",
```
Bekacru, line 12: We should avoid applying filters inline like this, they should be defined in the Biome config

examples/svelte-kit-example/eslint.config.js
```diff
@@ -0,0 +1,42 @@
+import prettier from 'eslint-config-prettier';
```
Bekacru, line 1: We don’t want to introduce ESLint or any other formatter/linter on top of Biome. There should be only one linter and formatter for the whole project. If Biome turns out to be incompatible with Svelte, feel free to open a new issue suggesting we switch to an ESLint/Prettier setup or if there is a fix for biome that'd be apperciated

examples/svelte-kit-example/.prettierrc
```diff
@@ -0,0 +1,16 @@
+{
```
Bekacru, line 1: we wouldn't want to introduce prettier on top of biome

Bekacru:
There should be no reason to remove the shadcn components. We want to have all of them there so it'd be easier for testing

examples/svelte-kit-example/eslint.config.js
DePasqualeOrg: Biome is not compatible with Svelte, so if you want to lint and format the SvelteKit example, you need to add prettier and eslint to the SvelteKit example. Are you sure you want to switch to prettier and eslint for the entire repo rather than just using them in the SvelteKit example, which will be ignored by Biome?

DePasqualeOrg:
@Bekacru, what do you mean having extra, unused components makes testing easier? On the contrary, I don't see any reason to keep this dead code. If you want to keep unnecessary components, why just these ones and not all the shadcn-svelte components from the registry? It doesn't make sense to me.

Bekacru:
I just tried building the current SvelteKit example, and it seems to build fine. Biome also appears to format it properly. Could you let me know exactly where it's broken?

Bekacru:
> @Bekacru, what do you mean having extra, unused components makes testing easier? On the contrary, I don't see any reason to keep this dead code. If you want to keep unnecessary components, why just these ones and not all the shadcn-svelte components from the registry? It doesn't make sense to me.

these are every component from the svelte shadcn registry if I recall properly :)

examples/svelte-kit-example/eslint.config.js
Bekacru: yes. Probably at this point if the example is such an issue and there is no way to fix through Biome, we'd probably move it to a different repo to avoid introducing another linter

DePasqualeOrg:
No, there are many more in the registry. You still haven't stated why you want to include unnecessary components, except for saying that they make testing easier, which doesn't make sense.

examples/svelte-kit-example/eslint.config.js
DePasqualeOrg: So to be clear, instead of keeping this example with all the others, which can have its own dependencies just like the others, you prefer to have a separate repo just for the SvelteKit example.

DePasqualeOrg:
> I just tried building the current SvelteKit example, and it seems to build fine. Biome also appears to format it properly. Could you let me know exactly where it's broken?

Biome doesn't work with SvelteKit and was stripping out required code, which is why the SvelteKit example was broken. The only reason it works now is because I fixed it by making Biome ignore the SvelteKit example.

Bekacru:
I'm saying, the one in the main branch works, Biome isn't stripping any code. I'm not sure what is broken about it.

examples/svelte-kit-example/eslint.config.js
Bekacru: yes

Bekacru:
Anyway, I feel like you're not engaging in good faith. Thanks for flagging this, we'll improve the SvelteKit example soon and if there's anything to fix, we'll address it then. In the meantime, there are several better-auth SvelteKit examples out there. Feel free to check those out. This one (like some of the others examples) hasn't been updated recently, but hopefully all of them will be soon.

DePasqualeOrg:
@Bekacru, what are you talking about? You're the one who's not engaging in good faith by closing this PR, which fixes a legitimate issue.

Your claim that the SvelteKit example already works is not true: I just cloned the main branch and verified that the SvelteKit example does not build or run in dev mode. Run `pnpm run check` in the SvelteKit example and you will see all the missing code that has already been stripped by Biome.

You gave a reason for keeping dead code that doesn't make sense, and then instead of engaging in a rational discussion, you simply closed this PR.

To anyone else watching this PR, what do you think about this behavior?

PR #2416, opened 2025-04-24
feat/saml plugin

### LINKED ISSUE #96: SAML authentication
For enterprise usecases; it is possible to achieve this with OAuth and running SAML Jackson, but would be nice to not need that extra dependency.

https://github.com/boxyhq/jackson

Kinfe123:
closes #96


.vscode/settings.json
```diff
@@ -2,7 +2,7 @@
 	"biome.enabled": true,
 	"editor.defaultFormatter": "biomejs.biome",
 	"[typescript]": {
-		"editor.defaultFormatter": "biomejs.biome"
+		"editor.defaultFormatter": "esbenp.prettier-vscode"
```
Bekacru: remove this change

Bekacru: and all prettier formattings

PR #2430, opened 2025-04-24
docs: Add guide for Sign In with Apple

TheGenio:
I noticed the docs were lacking in regards to apple sign in integration so i wanted to contribute to make sure others have an easier time than me.

I Noticed that step 1 was longer than all other steps in other providers but imo apple makes it harder so it makes sense why.

I also wanted to add a component to create the apple JWT right in the docs. Just like how the secret is created in the instalation part. Is that allowed? I would use crypto to create the jwt. User would insert all the needed details to generate the apple JWT and i would do it in the website directly.

Thank you, this is my first open source contribution so be kind :)

### DIFF
--- a/docs/content/docs/authentication/apple.mdx
+++ b/docs/content/docs/authentication/apple.mdx
@@ -2,22 +2,70 @@
 title: Apple
 description: Apple provider setup and usage.
 ---
-
 <Steps>
     <Step>
         ### Get your OAuth credentials
         To use Apple sign in, you need a client ID and client secret. You can get them from the [Apple Developer Portal](https://developer.apple.com/account/resources/authkeys/list).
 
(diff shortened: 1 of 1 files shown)


TheGenio:
@Bekacru Thank you, glad to see this merged.
I have two questions however.
1- In trusted origins. Bette auth url is automatically added, right? 
No need to do this 
```
  trustedOrigins: [
    process.env.BETTER_AUTH_URL as string,
    "https://appleid.apple.com",
  ],
```
Cause in the docs i said only added apple's domain in the array

2-I also wanted to add a component to create the apple JWT right in the docs. Just like how the secret is created in the installation part. Is that allowed? if yes then lmk so i can work on creating the component for it :)

Bekacru:
1. yes no need to do this
2. yeah go for it

PR #2432, opened 2025-04-25
fix(api-key): pass real rateLimitvalue from ctx.body

Siumauricio:
Ensures the `rateLimitEnabled` property in the ApiKey schema accurately reflects the provided input value, prioritizing it over defaults.

I noticed the value in the database was always true, linked issue to dokploy https://github.com/Dokploy/dokploy/issues/1757

### DIFF
--- a/packages/better-auth/src/plugins/api-key/routes/create-api-key.ts
+++ b/packages/better-auth/src/plugins/api-key/routes/create-api-key.ts
@@ -407,7 +407,7 @@ export function createApiKey({
 				refillInterval: refillInterval ?? null,
 				rateLimitEnabled:
 					rateLimitEnabled === undefined
-						? opts.rateLimit.enabled || true
+						? opts.rateLimit.enabled ?? true
 						: rateLimitEnabled,
 				requestCount: 0,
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/plugins/api-key/routes/create-api-key.ts
```diff
@@ -406,9 +406,9 @@ export function createApiKey({
 				refillAmount: refillAmount ?? null,
 				refillInterval: refillInterval ?? null,
 				rateLimitEnabled:
-					rateLimitEnabled ?? opts.rateLimit.enabled === undefined
-						? true
-						: opts.rateLimit.enabled,
+					rateLimitEnabled === undefined
+						? opts.rateLimit.enabled ?? true
```
Bekacru, line 410: I think this will ignore `opts.rateLimit.enabled` when it's set to false

Siumauricio: Added a new test for this case specifically

majidrazaee2:
خواهش میکنم کمکم کنید مرا مسدود نکنید

PR #2435, opened 2025-04-25
docs: Fix the color of the tooltip arrow in night mode

LioRael:
Before:
image
After:
78d056ce-f385-4664-9d00-660922a984de


Kinfe123:
it is already fixed

PR #2450, opened 2025-04-26
feat: ability to change default username for anonymous users.

drago1520:
As a database engineer, I want to have UNIQUE constrain on my users' names, so they do not dublicate.
Before, Every new Anonymous user had the same username "Anonymous".

Now, a unique string is generated for each user's name with a postfix.


drago1520:
#2361 #2362

Bekacru:
#2361

Issue #2434, opened 2025-04-25
nextCookies() do not set Cookies with auth.api


drago1520:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

I cannot log in with server methods like `auth.api` because they do not set the cookie headers in my Chrome browser on Next.js. Retrieving the session always returns `null`. I am using nextCookie(), but the server methods do not work. I can see Better-auth creating new sessions and users in the DB, but no cookies are returned to my browser.

**Goal:** Create a user on the server.

1. I am using Next.js
2. I followed the installation instructions for `better-auth`
3. I have the nextCookie() plugin enabled 
4. Here is my call in an isolated RSC page:
```js
import { auth } from '@/lib/auth';
import { headers } from 'next/headers';

export default async function BlogLayout({ children }: { children: React.ReactNode }) {
  const session = await auth.api.getSession({ headers: await headers() });
  console.log('session :', session);
  if (!session) {
    const data = await auth.api.signUpEmail({
        body: {
            email: `${crypto.randomUUID()}@example.com`,
            password: crypto.randomUUID(),
            name: crypto.randomUUID(),
        },
        headers: await headers(),
    });
    console.log('data :', data); //Creates users successfully; I see in the log & db.
    const session = await auth.api.getSession({ headers: await headers() });
    console.log('session :', session); //!Reads null even when I refresh the page
  }

  return (
    <>
      {children}
    </>
  );
}
```
5. I expect the second ` console.log('session :', session); //!Reads null even when I refresh the page to read the session, but it returns `null` even when I refresh the page, and a new user is being created.
6. Instead, no cookies are being sent to my Chrome browser

### Current vs. Expected behavior

I expect Chrome to sync with the newly created user using auth.api. Instead, no cookies are being set.

### What version of Better Auth are you using?

1.2.7

### Provide environment information

```bash
- OS: [Windows 10]
- Browser: [Chrome]
```

### Which area(s) are affected? (Select all that apply)

Client

### Auth config (if applicable)

```typescript
import { db, schema } from "@/models";
import { betterAuth } from "better-auth";
import { drizzleAdapter } from "better-auth/adapters/drizzle";
import { anonymous, openAPI } from "better-auth/plugins";
import { nextCookies } from "better-auth/next-js";

export const auth = betterAuth({
  database: drizzleAdapter(db, {
    provider: "pg",
    usePlural: true,
    schema,
  }),
  emailAndPassword: {
    enabled: true,
  },
[...]

Issue #2055, opened 2025-03-30
auth.api.getSession returns null in RSC Layout.tsx with valid session


ian:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

I'm trying to debug why `auth.api.getSession` will return null, when the session is valid after a query to /api/auth/get-session.

Consider the following src/app/layout.tsx:

```ts
import { Providers } from '@/components/providers';
import { auth } from '@/auth';
import { Analytics } from '@vercel/analytics/react';
import { headers } from 'next/headers';

export default async function RootLayout({
  children,
}: Readonly<{
  children: React.ReactNode;
}>) {
  const session = await auth.api.getSession({
    headers: await headers(),
  });

  console.debug('RootLayout', { session });

  return (
    <html lang="en" suppressHydrationWarning>
      <body>
        <Providers>{children}</Providers>
        <Analytics />
      </body>
    </html>
  );
}
```

### Current vs. Expected behavior

One would expect to see "RootLayout: { session: { ... } }" but instead, the session is null.

I can see the session token in the requests:

![Image](https://github.com/user-attachments/assets/bc0fd419-0de8-460a-9189-d769c868293f)

And requests to /api/auth/get-session return a valid session:

![Image](https://github.com/user-attachments/assets/2c8b73d2-0362-4264-ac7b-0e3912a765da)

### What version of Better Auth are you using?

1.2.5

### Provide environment information

```bash
- OS: MacOS
- Browser: Chrome
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
export const auth = betterAuth({
  basePath: '/auth',
  database: prismaAdapter(prisma, {
    provider: 'postgresql',
  }),
  plugins: [
    emailOTP({
      sendVerificationOTP,
    }),
    nextCookies(),
    expo(),
  ],
  socialProviders: {
    google: {
      clientId: process.env.GOOGLE_CLIENT_ID as string,
      clientSecret: process.env.GOOGLE_CLIENT_SECRET as string,
    },
  },
  user: {
    additionalFields: {
      role: {
        type: 'string',
        required: false,
        defaultValue: 'user',
        input: false,
      },
      lang: {
        type: 'string',
        required: false,
        defaultValue: 'en',
      },
    },
  },
});
```

BjoernRave:
[...]

PR #2451, opened 2025-04-26
docs(error-handling): added error message when next.js tries to set cookies in a RSC or other enviroments where cookies() is read-only.

drago1520:
Currently, setting a cookie in a read-only cookie enviroment fails silently, which is horrible. The Client is out of sync of the database and you have no idea why.

I had this problem and wasted 6 days working on it. It would save me all this time if I just had this error handling.


drago1520:
This fixes #2434 #2055

Bekacru:
The reason we didn't do this is cause, this will make a lot of error logs since most likely you'll have a bunch of `auth.api.getSession` calls in RSC

PR #2453, opened 2025-04-26
Docs-feat: added apple sign in JWT  generation in docs

TheGenio:
Related to https://github.com/better-auth/better-auth/pull/2430

To keep users in better auth docs more and to avoid them getting confused with how to create the jwt. I Added this component to help users generate the JWT right in their browsers as a client component.

Tested this and it works .
WIth this component  there it make the process of opening the docs to having sign in with apple work as a sign in option happen and production ready in less than 3 mins :)


Bekacru:
Screenshot 2025-04-26 at 10 36 42 PM
👀

TheGenio:
@Bekacru Got your notification about the error. For some reason i cant find your reply here but I will double check my implementation and see whats wrong

TheGenio:
@Bekacru  Update: Its a safari issue with jose and EC keys.
I will try to find an alternative for it

TheGenio:
@Bekacru Fixed. Works on safari and chrome :)

Production testing required for JWT/Apple integration to make sure apple accepts it, but confident as jose's similar method worked. Only diff is i changed the function.

 jwt.io yields same JWT's so i think it should work

TheGenio:
@Bekacru Bump, is there anything else i need to do so this gets merged?

PR #2511, opened 2025-05-02
fix: return local IP in development mode

DiiiaZoTe:
In the getIp function, added a condition to return the loopback IP address (127.0.0.1) when running in development mode. This ensures that Better Auth's rate limiting functions correctly during local development by providing a consistent IP address. This change addresses issues where the client's IP address isn't available in local development environments because it is set to ::1 or another value, leading to the rate limiter not functioning as intended as it will not pass the isValidIp check.


himself65:
move to https://github.com/better-auth/better-auth/pull/4591

himself65:
Oh, I successfully rebased the commit. let's back to here

PR #2514, opened 2025-05-02
fix(api-key): rate limits not working

ntgussoni:
This PR fixes a bug where rate limits were not enforced when using api-keys and sessions.

- Extracted the verification step so that it can be reused
- Added new tests


packages/better-auth/src/plugins/api-key/routes/verify-api-key.ts
```diff
@@ -167,9 +167,9 @@ export async function validateApiKey({
 	});
 
 	if (!newApiKey) {
-		throw new APIError("UNAUTHORIZED", {
-			message: ERROR_CODES.USAGE_EXCEEDED,
-			code: "USAGE_EXCEEDED" as const,
+		throw new APIError("INTERNAL_SERVER_ERROR", {
+			message: "Failed to update API key",
```
Bekacru: let's put this under ERROR_CODES

ntgussoni: done @Bekacru

packages/better-auth/src/plugins/api-key/index.ts
```diff
@@ -165,21 +167,15 @@ export const apiKey = (options?: ApiKeyOptions) => {
...
+							opts,
+							schema,
 						});
 
-						if (!apiKey) {
-							throw new APIError("UNAUTHORIZED", {
-								message: ERROR_CODES.INVALID_API_KEY,
-							});
-						}
+						await routes._deleteAllExpiredApiKeys(ctx.context);
```
rxri: just use `await deleteAllExpiredApiKeys(ctx.context)`. I don't see a reason for duplicating the function in this case tbh

ntgussoni: @rxri the endpoint and the middleware take a different context, that's why i made them separate. Also, I think it's better to keep them isolated, because I don't need the whole route, just the method which is shared all across.
image

rxri: I see, but wouldn't it be better to just move it off from `createApiKeyRoutes` and then export the function and import it when you need it...? That's just my opinion in the end

ntgussoni: I was gonna say that the method needs to be in the closure, together with lastUsed but it's the same if its at the module level.

I've moved it out, removed the alias _ i created and reused the same function.  Let me know what you think

rxri: yea that's fine for sure ^^

CodeWithShreyans:
this also fixes `lastRequest`

@Bekacru please merge

PR #2527, opened 2025-05-03
fix: session cookie refresh on update and verify

Kinfe123:


packages/better-auth/src/plugins/phone-number/index.ts
```diff
@@ -580,9 +580,23 @@ export const phoneNumber = (options?: PhoneNumberOptions) => {
 							},
 							ctx,
 						);
+						const newSession = await ctx.context.internalAdapter.createSession(
```
Bekacru: We don't need to create a new session

packages/better-auth/src/plugins/phone-number/index.ts
```diff
@@ -580,9 +580,23 @@ export const phoneNumber = (options?: PhoneNumberOptions) => {
...
+							user.id,
+							ctx,
+						);
+						if (!newSession) {
+							throw new APIError("INTERNAL_SERVER_ERROR", {
+								message: BASE_ERROR_CODES.FAILED_TO_CREATE_SESSION,
+							});
+						}
+
+						await setSessionCookie(ctx, {
```
Bekacru, line 583: here we just need to update the user object

ma13gagan:
@Kinfe123, Any update on this PR?

himself65 (changes requested):
Could you please add a test case?

PR #2545, opened 2025-05-05
Allow getUserInfo return null

isnifer:
Main `ProviderOptions.getUserInfo` do not allow to return `null` as value. But each oauth2 does.
VSCode complains about it:
image

I've added `null` as a possible value for `ProviderOptions.getUserInfo` to `types.ts`


Jarva:
I'm encountering this issue also.

@isnifer Was this handled in another PR?

isnifer:
@Jarva sorry, my bad. I just cleared up forks and didn't notice — this one didn't merge yet.
No this wasn't handled. I just have `// @ts-ignore` above facebook's `getUserInfo`

PR #2557, opened 2025-05-06
feat: add getAccessToken api for oauth accounts

BlankParticle:
Added a `getAccessToken` api that returns a valid access token after refreshing if necessary.
Useful for server side logic which needs to use the access token to call apis on behalf of users.


packages/better-auth/src/api/routes/get-access-token.ts
```diff
@@ -0,0 +1,145 @@
...
+			throw new APIError("BAD_REQUEST", {
+				message: "Account not found",
+			});
+		}
+		const provider = ctx.context.socialProviders.find(
+			(p) => p.id === providerId,
+		);
+		if (!provider) {
+			throw new APIError("BAD_REQUEST", {
+				message: `Provider ${providerId} not found.`,
```
ping-maxwell: Should probably check if the provider exists before checking if a given account has that provider id

ping-maxwell (approved):

packages/better-auth/src/api/routes/get-access-token.ts
BlankParticle: I copied the refreshToken endpoint as base, so need to change that too?

ping-maxwell: Oh sure, that would be great.

splicebison:
@BlankParticle @Bekacru This is a good feature to have however the current implementation doesn't account for generic oauth clients as it's explicitly checking only the social providers list. It would be better if this could accommodate non-social oauth clients used in the generic oauth plugin

BlankParticle:
> @BlankParticle @Bekacru This is a good feature to have however the current implementation doesn't account for generic oauth clients as it's explicitly checking only the social providers list. It would be better if this could accommodate non-social oauth clients used in the generic oauth plugin

created pr #2643

IhsenBouallegue:
is it just me or the authClient methods are missing? It works on the server side. But contrary to what to the docs the client side methods are missing.

image
image

BlankParticle:
> is it just me or the authClient methods are missing? It works on the server side. But contrary to what to the docs the client side methods are missing.
> 
> image image

`SERVER_ONLY` is set to `true` on the method, so the method doesn't show up on the client

PR fixing the issue https://github.com/better-auth/better-auth/pull/2662

PR #2578, opened 2025-05-07
update(forget-password): sets user's emailVerified to true upon sucessful password-reset

ping-maxwell:


packages/better-auth/src/api/routes/forget-password.ts
```diff
@@ -284,6 +284,13 @@ export const resetPassword = createAuthEndpoint(
 			hashedPassword,
 			ctx,
 		);
+
+		await ctx.context.internalAdapter.updateUser(
```
Bekacru, line 288: we should check if it's already verified before updating

Bekacru (changes requested):

packages/better-auth/src/api/routes/forget-password.ts
ping-maxwell: Wouldn't that end up slower having to call DB twice? In either case their emailVerified status would end up as `true`

himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #2579, opened 2025-05-07
feat: SIWE plugin

### LINKED ISSUE #41: Siwe support
sample: 
https://docs.reown.com/appkit/vue/core/siwe

rokitgg:
Aims to close #41

Tried my best to implement SIWE within the boundaries of better-auth, and seems like this draft may be a good starting point for it! Although I do feel like a bit of discussion is needed on some topics.

Things to consider:

1) As of this draft, the plugin requires the official [siwe](https://www.npmjs.com/package/siwe) npm package as a dependency, which may not be ideal
2) Comprehensive testing could include mocking a wallet connection to generate a real signature - as of now tests are bit too shy in my opinion, should be improved.
3) Happy to help with documentation on how to actually implement a web3 login feature using this plugin, which absolutely needs some proper docs to help beginners.

Any feedback/contributions are highly welcomed!


reslear:
we need remove `siwe` packages and use user functions for siwe `createmessage` or `generateNonce` 
for example for use 
 https://viem.sh/docs/siwe/utilities/createSiweMessage

or https://docs.reown.com/appkit/vue/core/siwe

8times4:
Given that viem directly implements siwe, using siwe's bloat is not so great imho. 

it's be also beneficial to make the `chainId` or chain itself a plugin option, given that the message can contain the `chainId` of any other chain, however there'd be no direct way to modify that from where the message is crafted. 
(Meaning you could not verify a message using the snippet below for any other `chainId`.)

Ps; 've also made the rpc url an option in mine, in case I need to use a custom rpc when validating via publicClient.

Here's a small snippet of my local plugin I've been using:
```ts
import { generateSiweNonce, parseSiweMessage } from "viem/siwe";
import { createPublicClient, http, type Address, type Hex } from "viem";

const publicClient = createPublicClient({
  chain: mainnet,
  transport: http()
})

...

const nonce = generateSiweNonce();

...

const parsedMessage = parseSiweMessage(message);

...


            const isValid = await publicClient.verifyMessage({
              message,
              signature: signature as Hex,
              address: parsedMessage.address,
            });
```



So pretty much almost the same what comes with siwe.
docs for viem: https://viem.sh/docs/siwe/actions/verifySiweMessage

rokitgg:
> we need remove `siwe` packages and use user functions for siwe `createmessage` or `generateNonce` for example for use https://viem.sh/docs/siwe/utilities/createSiweMessage
> 
> or https://docs.reown.com/appkit/vue/core/siwe

Agree but packages like viem use siwe under the hood so isn't that quite the same problem after all?

reslear:
> > we need remove `siwe` packages and use user functions for siwe `createmessage` or `generateNonce` for example for use [viem.sh/docs/siwe/utilities/createSiweMessage](https://viem.sh/docs/siwe/utilities/createSiweMessage)
> > or [docs.reown.com/appkit/vue/core/siwe](https://docs.reown.com/appkit/vue/core/siwe)
> 
> Agree but packages like viem use siwe under the hood so isn't that quite the same problem after all?

its no problem, i mean let the user do it :

```ts

// import viem or siwe or other libs ...
 
    // ... other config options
    plugins: [
        siwe({ 
                async generateSiweNonce() { 
					// Implement 
				}, 
				
				async parseSiweMessage() {
				
				}
        }) 
    ]
})
```

rokitgg:
> > > we need remove `siwe` packages and use user functions for siwe `createmessage` or `generateNonce` for example for use [viem.sh/docs/siwe/utilities/createSiweMessage](https://viem.sh/docs/siwe/utilities/createSiweMessage)
> > > or [docs.reown.com/appkit/vue/core/siwe](https://docs.reown.com/appkit/vue/core/siwe)
> > 
> > 
> > Agree but packages like viem use siwe under the hood so isn't that quite the same problem after all?
> 
> its no problem, i mean let the user do it :
> 
> ```ts
> import { betterAuth } from "better-auth"
> 
> // import viem or siwe or other libs ...
>  
> export const auth = betterAuth({
>     // ... other config options
>     plugins: [
>         siwe({ 
>                 async generateSiweNonce() { 
> 					// Implement 
> 				}, 
> 				
> 				async parseSiweMessage() {
> 				
> 				}
>         }) 
>     ]
> })
> ```

Ohh okay now I get what u mean by user functions, that def sounds like the best solution!

rokitgg:
So plugin options could look something the likes of

```ts
interface SiweOptions {
	domain: string;
	generateSiweNonce: () => Promise<string>;
	parseSiweMessage: (message: string, signature: string) => Promise<boolean>;
}
```

Since the siwe message should be generated on the client, that's probably all it takes in terms of plugin user functions - lmk if missed something

reslear:
@rokitgg and need `verifyMessage`

byronpolley:
@rokitgg I was about to start writing this as a community plugin 'cause I need it to migrate a project from Auth.js 😅
Happy to assist you here though.

On your question with external deps, I think adding `siwe` is the preferred option over writing your own functions that will be complex to maintain as new standards emerge and since any auth with eth will always need this package it's kinda an essential part of the flow.

We should however add it to peer deps so the user is expected to install it to keep bundle size down and add a note in the docs. 

Not sure what @8times4 means by bloat as it's 76.4 kB unpacked compared to `viem` 16.4 MB, so don't use that.

If going with @reslear solution (user defines the functions) then add a custom RPC to the config as the `http()` transport is rate limited and therefore not production ready.

```
import { http, createPublicClient } from "viem";
import { mainnet, sepolia, type Chain } from "viem/chains";

export function getChainPublicClient(chain: Chain) {
  const apiKey = process.env.PROVIDER_API_KEY;
  const rpcUrl =
    chain === mainnet
      ? "https://api.etherscan.io/api"
      : "https://sepolia.etherscan.io/api";
  const useCustomRpc = (chain === mainnet || chain === sepolia) && !!apiKey;
  return createPublicClient({
    chain: chain,
    transport: useCustomRpc ? http(`${rpcUrl}/${apiKey}`) : http(),
  });
}
```

packages/better-auth/src/plugins/siwe/index.ts
```diff
@@ -0,0 +1,156 @@
...
+								{
+									field: "publicKey",
+									operator: "eq",
+									value: ctx.body.publicKey,
+								},
+							],
+						});
+
+						if (!user) {
+							const tempEmail = `${ctx.body.publicKey}@${process.env.NEXT_PUBLIC_BASE_URL}`;
```
byronpolley: Use 
```
const { emailDomainName = getOrigin(ctx.context.baseURL) } = options || {};`
const email = `${publicKey}@${emailDomainName}`;
```

packages/better-auth/src/plugins/siwe/index.ts
```diff
@@ -0,0 +1,156 @@
...
+							user = await ctx.context.internalAdapter.createUser({
+								name: ctx.body.publicKey,
```
byronpolley: Missed opportunity to retrieve ENS name, take it in as an optional param from the client and fallback to something else other than `publicKey`. Check the Anonymous plugin `AnonymousOptions` interface for some other ideas.

packages/better-auth/src/plugins/siwe/types.ts
```diff
@@ -0,0 +1,5 @@
+import type { User } from "../../types";
+
+export interface SiweUser extends User {
+	publicKey?: string;
```
byronpolley: `walletAddress` would be better as publicKey is too general IMO.

8times4:
> @rokitgg I was about to start writing this as a community plugin 'cause I need it to migrate a project from Auth.js 😅 Happy to assist you here though.
> 
> On your question with external deps, I think adding `siwe` is the preferred option over writing your own functions that will be complex to maintain as new standards emerge and since any auth with eth will always need this package it's kinda an essential part of the flow.
> 
> We should however add it to peer deps so the user is expected to install it to keep bundle size down and add a note in the docs.
> 
> Not sure what @8times4 means by bloat as it's 76.4 kB unpacked compared to `viem` 16.4 MB, so defintely don't use that.

ah, my bad then, I stand corrected on bloat there.
My thought process went along as `viem` being used anyways in a web3 project(given that most wallet connector adapters like rainbow/appkit/family rely on `viem` anyways), so having it built in would not really add size.
On top of that, I was under the assumption that the `siwe` package is not maintained as well as viem, given their ecosystem, but I was wrong there as well. :) 

unrelated; I just noticed that siwe is also implemented in  `ox`, which is wevm's Ethereum StdLib - wevm is the parent company of `viem` - 
https://oxlib.sh/api/Siwe

byronpolley:
@8times4 Yeah `viem` is industry standard at this point so most will be using that for sure I get your thinking. But better to decouple in case users want to use `ethers` etc.

You're actually not wrong on SIWE maintenance it has seen better days 😅 Maybe we use it for now and write our own utils phasing it out 1:1 at some point. This industry moves really fast though so we will be stuck on maintaining it ofc. Tradeoffs.

rokitgg:
> @rokitgg I was about to start writing this as a community plugin 'cause I need it to migrate a project from Auth.js 😅 Happy to assist you here though.
> 
> On your question with external deps, I think adding `siwe` is the preferred option over writing your own functions that will be complex to maintain as new standards emerge and since any auth with eth will always need this package it's kinda an essential part of the flow.
> 
> We should however add it to peer deps so the user is expected to install it to keep bundle size down and add a note in the docs.
> 
> Not sure what @8times4 means by bloat as it's 76.4 kB unpacked compared to `viem` 16.4 MB, so don't use that.
> 
> If going with @reslear solution (user defines the functions) then add a custom RPC to the config as the `http()` transport is rate limited and therefore not production ready.
> 
> ```
> import { http, createPublicClient } from "viem";
> import { mainnet, sepolia, type Chain } from "viem/chains";
> 
> export function getChainPublicClient(chain: Chain) {
>   const apiKey = process.env.PROVIDER_API_KEY;
>   const rpcUrl =
>     chain === mainnet
>       ? "https://api.etherscan.io/api"
>       : "https://sepolia.etherscan.io/api";
>   const useCustomRpc = (chain === mainnet || chain === sepolia) && !!apiKey;
>   return createPublicClient({
>     chain: chain,
>     transport: useCustomRpc ? http(`${rpcUrl}/${apiKey}`) : http(),
>   });
> }
> ```

Thank you so much for this comments! I really do agree with your vision.

packages/better-auth/src/plugins/siwe/index.ts
rokitgg: Yep, I had ENSName and ENSAvatar in a previous implementation but dropped it before drafting PR because it depended on wagmi. But happy to add that!

rokitgg:
Damn this [jazzberry](https://github.com/apps/jazzberry-ai) bot really is smart 😂

reslear:
Thank @rokitgg you for your hard work

@8times4 @b-bot Still disagree, this is not consistent with the whole library:

1. there should be no additional packages neither view nor siwe nor ox (this should be user selectable).
2. all functions should be customizable by the user, giving the user a choice

Imagine if we added only 1 provider to the plugin https://www.better-auth.com/docs/concepts/email

rokitgg:
I certainly think @reslear makes a really strong point here.

If i had no other option but to add any deps (as peer) my personal choice would be `siwe` since it's such a tiny package in comparison and does the job we need here pretty well, however I do agree with giving users - in this case developers - the choice to bring their own implementation via viem, ox or whatever choice they wish to.

byronpolley:
Yeah let's go with typing the function and leaving implementation to the user to pass in the config 👌🏼

As a final note (and why I mentioned Anonymous plugin) we need to handle account linking. ie. user signs in with google while signed in with eth and it should replace the fake email. This would allow retrieval of the account through multiple methods.

packages/better-auth/src/plugins/siwe/index.ts
```diff
@@ -0,0 +1,153 @@
...
+
+						if (!user) {
+							const emailDomainName =
+								options.emailDomainName ?? getOrigin(ctx.context.baseURL);
+							const email = `${walletAddress}@${emailDomainName}`;
+							user = await ctx.context.internalAdapter.createUser({
+								name: ensName ?? walletAddress, // TODO: should fallback to something else other than walletAddress
+								email,
+								walletAddress,
+								avatar: "", // TODO: implement ens avatar (?)
```
byronpolley: import { ethers } from 'ethers';

const provider = new ethers.providers.JsonRpcProvider('https://mainnet.infura.io/v3/YOUR_INFURA_KEY');

async function getEnsAvatar(ensName) {
  const resolver = await provider.getResolver(ensName);
  const avatar = await resolver.getText('avatar');
  return avatar;
}

getEnsAvatar('vitalik.eth').then(console.log);

rokitgg: Shall we make this getEnsAvatar function a user function - I guess so (?)

reslear:
Next point - it will be necessary to check how it works when a user has many addresses, and maybe switching or linking

packages/better-auth/src/plugins/siwe/index.ts
byronpolley: Since it's correlated to retrieving the ens name (via the provider) we could get both in a single function with optional returns. `ensLookup(walletAddress) => { name, avatar }`

rokitgg:
Alright so just removed the `siwe` dependency and refactored related functions into being plugin options - for the most part just a mock but surely moving in the right direction

Where I'm a bit lost is on how to write exhaustive tests for this plugin - If anyone is willing to share a bit of light there I would very much appreciate that!

byronpolley:
@rokitgg You're going to want to finalise everything before writing test cases but I would say that because it's more of an open implementation we can mock most things. I don't know the policy about adding packages for testing (since dev deps aren't shipped in bundle) but this exists: https://wagmi.sh/react/api/connectors/mock

Use AI to gen some cases after finishing everything and we can look for gaps.

There's also examples of how SIWE is testing to give you ideas: https://github.com/spruceid/siwe/blob/main/packages/siwe/lib/client.test.ts

PS. I thought of one more thing we might want to add to config.

An `anonymous` field defaulting to true. If not it requires an email to be passed into the sign in call with the payload. Because Better Auth is so dependant on email this is a good idea and will help with notifications and account linking.

[Edit:](https://www.better-auth-kit.com/docs/libraries/tests)

rokitgg:
Thank you so much for the suggestions @b-bot - already implemented the anonymous flag

Regarding tests, the issue here when using the `mock` feature from wagmi it's that I would be likely falling into the same issue again of having `wagmi` as a dep, so that's concerning still, but really nice that you pointed that out.

Will be taking a closer look at the siwe tests you mentioned there, that's probably quite helpful.

byronpolley:
@rokitgg looking good here, excited to use this!
I can pull this to test and write the docs if you like?

packages/better-auth/src/plugins/siwe/schema.ts
```diff
@@ -0,0 +1,13 @@
+import type { AuthPluginSchema } from "../../types";
+
+export const schema = {
+	user: {
+		fields: {
+			walletAddress: {
```
8times4: regarding this, wouldn't it possibly make sense to call it `evmAddress`? 
only reason i'm asking is because SIWS (Sign In With Solana) could not use `publicKey` as it being too generic, - but solAddress would not be so "streamlined" compared to walletAddress. 

(This is to future-proof as I'm planning to add SIWS later)

byronpolley: @8times4 wouldn't that be another plugin as this is tightly coupled to ENS etc?
(I don't know SIWS)

8times4: correct @b-bot although [SIWX](https://docs.reown.com/appkit/next/core/siwx) exists, that would be overkill most likely. 

[SIWS](https://phantom.com/learn/developers/sign-in-with-solana) has SNS instead of ENS, only proposing this `evmAddress` change due to: https://github.com/better-auth/better-auth/pull/2579#discussion_r2079555101
so that it's somewhat similar/understandable in the schema which does what if someone uses both `SIWE` and `SIWS`

rokitgg:
> @rokitgg looking good here, excited to use this! I can pull this to test and write the docs if you like?

Sure! Go ahead whenever u wish to.

packages/better-auth/src/plugins/siwe/schema.ts
rokitgg: Don't rlly have any strong views on this so absolutely open to rename the field into whatever feels the most convenient for everyone - naming things is truly hard and it shows

vimtor:
Hey! Just chiming in to suggest that the plugin should probably store the checksummed version of the wallet address instead of the raw input. Not all wallets send the address properly checksummed, and this could lead to inconsistencies when trying to match the user given a wallet.

rokitgg:
@vimtor thanks for your suggestion, totally missed that one! Will absolutely look to implement the checksum parsing.

The issue I've found here is that the hashing algorithm `Keccak-256` is not natively supported by `node:crypto`, so this will led us into adding a peer-dep to the plugin, which is not ideal even tho the [keccak](https://bundlephobia.com/package/keccak@3.0.4) package is so tiny.

`node:crypto` supports `sha3-256` which is very similar, but if we use `sha3-256` instead of `Keccak-256` for Ethereum addresses, we will get the wrong checksum.

If we want strict EIP-55 compliance, we need a **_Keccak256_** implementation somehow - happy to discuss the go to way for this.

For reference:

https://ethereum.stackexchange.com/questions/550/which-cryptographic-hash-function-does-ethereum-use

https://eips.ethereum.org/EIPS/eip-55

As defined in the EIP-55 reference (link above), this is the appropriate implementation:

```ts
const createKeccakHash = require('keccak')

function toChecksumAddress (address) {
  address = address.toLowerCase().replace('0x', '')
  var hash = createKeccakHash('keccak256').update(address).digest('hex')
  var ret = '0x'

  for (var i = 0; i < address.length; i++) {
    if (parseInt(hash[i], 16) >= 8) {
      ret += address[i].toUpperCase()
    } else {
      ret += address[i]
    }
  }

  return ret
}
```

reslear:
@rokitgg 
viem use small module https://github.com/paulmillr/noble-hashes
we can add Tree-shakeable keccak to https://github.com/better-auth/utils

based64-eth:
@rokitgg im looking to quickly implement siwe signin to my better auth app. do you have a working example anywhere that uses the viem/siwe packages without the hassle of reimplementing their functions from scratch?

rokitgg:
> @rokitgg 
> 
> viem use small module https://github.com/paulmillr/noble-hashes
> 
> we can add Tree-shakeable keccak to https://github.com/better-auth/utils 
> 
> 
> 
> 

Cool, should I make a PR for that or...?

Sorry for the absence I've been busy lately but I'll be advancing this PR today.

rokitgg:
Cool so @vimtor suggestion has been implemented, appreciate you for that one.

Overall I'll say we're pretty close to a reviewable PR.

Any other suggestions/changes please let me know so we can add the final touches - cheers!

reslear:
@rokitgg about `keccak256`:

| [keccak](https://pkg-size.dev/keccak) | [@noble/hashes](https://pkg-size.dev/@noble/hashes) |
|--------|--------|
|pkg-size dev_keccak(iPhone 12 Pro) (1) | pkg-size dev_@noble_hashes(iPhone 12 Pro) (1)|

reslear:
1. 
By the way, it would be cool to do a future installment. 
that a user could have multiple addresses 
walletAddress[] or even a whole table for this purpose
 
2. 
walletAddress should always go with the network address chainId, we're past that problem https://docs.reown.com/appkit/javascript/core/siwe

rokitgg:
@reslear thank you for that buddy. Will make the change(s) just now!

packages/better-auth/src/plugins/siwe/index.ts
```diff
@@ -0,0 +1,250 @@
...
+						identifier: `siwe:${walletAddress}:${chainId}`,
+						value: nonce,
+						expiresAt: new Date(Date.now() + 15 * 60 * 1000),
+					});
+
+					return { nonce };
+				},
+			),
+			// Verify siwe payload
+			verify: createAuthEndpoint(
```
ping-maxwell: Try to use more *specific* naming, instead of `verify`.
This is because everything will be combined into one auth.api, thus `auth.api.verify` wouldn't be very descriptive.

packages/better-auth/src/plugins/siwe/index.ts
```diff
@@ -0,0 +1,250 @@
...
+	ensLookup?: (args: ENSLookupArgs) => Promise<ENSLookupResult>;
+}
+
+export const siwe = (options: SIWEPluginOptions) =>
+	({
+		id: "siwe",
+		schema,
+		endpoints: {
+			// Generate nonce endpoint
+			nonce: createAuthEndpoint(
```
ping-maxwell: Same as the comment I left below, try to use more descriptive names.

packages/better-auth/src/plugins/siwe/index.ts
```diff
@@ -0,0 +1,250 @@
...
+					const nonce = await options.getNonce();
+					
+					// Store nonce with wallet address and chain ID context
+					await ctx.context.internalAdapter.createVerificationValue({
+						identifier: `siwe:${walletAddress}:${chainId}`,
+						value: nonce,
+						expiresAt: new Date(Date.now() + 15 * 60 * 1000),
+					});
+
+					return { nonce };
```
ping-maxwell: I'm not exactly sure what would occure here, but I would recommend using `ctx.json({nonce})`

packages/better-auth/src/plugins/siwe/index.ts
```diff
@@ -0,0 +1,250 @@
...
+						// Create new user if none exists
+						if (!user) {
+							const domain = options.emailDomainName ?? getOrigin(ctx.context.baseURL);
+							const userEmail = !isAnon && email ? email : `${walletAddress}@${domain}`;
+							const { name, avatar } = (await options.ensLookup?.({ walletAddress })) ?? {};
+							
+							user = await ctx.context.internalAdapter.createUser({
+								name: name ?? walletAddress,
+								email: userEmail,
+								avatar: avatar ?? "",
```
ping-maxwell: we use `image`, not `avatar`

packages/better-auth/src/plugins/siwe/index.ts
rokitgg: Agree! I thought this would end up looking something the likes of auth.api.siwe.verify that's why but thank you for the review.

rokitgg:
@ping-maxwell super thankful for the review, fixes have been commited!

Bekacru:
hey @rokitgg can you add the docs for it?

byronpolley:
@rokitgg I said I would write the docs - instead of complicating the PR I added a draft for you here
https://gist.github.com/b-bot/a824c2e5e49f43ad4de448a94424a7d2

Just check through it and make any modifications in case I missed anything.

rokitgg:
> hey @rokitgg can you add the docs for it?

Sure, more than happy to.

Thanks to @b-bot for the draft!

rokitgg:
@Bekacru I've just commited the docs, or at least the first draft for it.

If anyone feels that something is missing, please comment below!

Thanks again @b-bot for doing pretty much all of the docs-related work.

rokitgg:
I am also working on a production-ready SIWE implementation repo leveraging all of the plugin features like the ENS lookup, multiple wallets on different chains, etc .... with a live preview deployed on Vercel.

Will surely link it here once it's ready!

byronpolley:
So keen to integrate this! 

As a side note, @8times4 can we get a draft up for Sign In With Solana?
Happy to assist but more familiar with EVM.

reslear:
@b-bot its concept move from `siwe`->`siwx`(you can even do SUI)
https://docs.reown.com/appkit/javascript/core/siwx-custom

byronpolley:
@reslear will this be compatible with this PR or supersede it?

8times4:
> So keen to integrate this!
> 
> As a side note, @8times4 can we get a draft up for Sign In With Solana? Happy to assist but more familiar with EVM.

Sure thing, will draft one up soon for SIWS. 

@reslear regarding SIWX, my only concern is that as per AppKit, it’s not production ready, so idk how stable it is.
https://github.com/reown-com/appkit/tree/main/packages/siwx

byronpolley:
@Bekacru What else left to get this merged?

byronpolley:
@rokitgg Been using this in a production app and just a couple of things for either this PR or after merge:

1. We should create an entry in the accounts table for each wallet (accountId maybe referencing the address) because as it stands if a user links a social account, they will not be able to unlink it as it requires at least one left to prevent lockout.

2. Plugin breaks if using table name mapping eg. `wallet_addresses` which is best practice on SQL DB's. Didn't look into why.

3. Might be wrong but is there any reason the dummy email prefix is not checksummed? Might come in handy for deriving wallet.

Other than that it's working great you can check it out here: https://poly.link/auth/login

rokitgg:
> @Bekacru What else left to get this merged?

I think we're pretty much there! Huge thanks to everyone who contributed.

Can't wait to make use of this plugin in my production apps 🙏

byronpolley:
Let's gooooo

0xNegative:
Timing couldn't have been better. Thought I was about to have to roll my own SIWE better auth solution. Thanks to everyone that helped here.

PR #2397, opened 2025-04-22
feat(admin): update user

Bekacru:
close #2394

PR #2587, opened 2025-05-08
feat(admin): get user

### LINKED ISSUE #2585: Allow admin to get a user

The admin plugin does not currently support getting a user.

New endpoint to get the user by id.

Creating a custom route in a standalone plugin

### Additional context

_No response_

0xJJW:
closes #2585

1. Added user role `get`
2. Added error code `YOU_ARE_NOT_ALLOWED_TO_GET_USER`
3. Add endpoint `/admin/get-user`
3. Updated admin tests to follow #2397


0xJJW:
This PR only adds support for getting a user by `id`.

After some consideration, it might be worth adding support for `id` and `email. ' If you agree, let me know how you would like to proceed.

e.g one endpoint or separate endpoints for each?

I kept this PR simple as it's my first for this project.

Kinfe123 (approved):

reslear:
wow really need

ping-maxwell:
@reslear Can you add docs?

reslear:
@ping-maxwell 
[Patch.patch](https://github.com/user-attachments/files/21201528/Patch.patch)

packages/better-auth/src/plugins/admin/admin.ts
```diff
@@ -5,23 +5,103 @@ import {
...
+	banned?: boolean | null;
+	banReason?: string | null;
+	banExpires?: Date | null;
+}
+
+export interface SessionWithImpersonatedBy extends Session {
+	impersonatedBy?: string;
+}
+
+export interface AdminOptions {
```
himself65: why do we duplicate the code here?

himself65 (changes requested):

packages/better-auth/src/plugins/admin/admin.ts
ping-maxwell: Resolved!

himself65 (approved):

Laurin-Notemann:
Maybe I am wrong, but to me it seems like the return type is hard coded? meaning if I added additional fields, or use the admin plugins it does not show those types. unless there is something i don't know about how to use this function

PR #2611, opened 2025-05-10
feat(docs): expo -> add baseUrl explanation

casperleerink:
Hi I got stuck with the expo intergration and it took me a while to figure it out, but the issue is that the baseUrl `localhost:8081` (lets pretend thats where your server is), of course doesn't work on the phone (server is not running on the phone)! This pr adds an accordion explanation to help other people get the correct `baseUrl` with a function. 

Especially helpful, since there is no error message for when the server url is not found.

```ts
import Constants from "expo-constants";

export const getBaseUrl = () => {
  const debuggerHost = Constants.expoConfig?.hostUri;
  const localhost = debuggerHost?.split(":")[0];

  if (!localhost) {
   // Production
   return "https://[your-better-auth-backend-url]";
  }
  // Development
  return `http://${localhost}:8081`; // this should be the port of your Better Auth backend
};
```

### DIFF
--- a/docs/content/docs/integrations/expo.mdx
+++ b/docs/content/docs/integrations/expo.mdx
@@ -18,7 +18,7 @@ Expo is a popular framework for building cross-platform apps with React Native. 
 
         ```ts title="app/api/auth/[...auth]+api.ts"
         import { auth } from "@/lib/auth"; // import Better Auth handler
-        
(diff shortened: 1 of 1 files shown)


himself65:
I think you can use environment variable to set the base url

PR #2623, opened 2025-05-12
Allow filtering users from using 'contains' operator

### LINKED ISSUE #2588: Filter user by role in authClient.admin.listUsers()

### To Reproduce

Use the official admin plugin and call authClient.admin.listUsers() on the frontend

The exemple in the [docs](https://www.better-auth.com/docs/plugins/admin#list-users) shows:
```js
const users = await authClient.admin.listUsers({
    query: {
        searchField: "email",
        searchOperator: "contains",
        searchValue: "@example.com",
        limit: 10,
        offset: 0,
        sortBy: "createdAt",
        sortDirection: "desc",
        filterField: "role",
        filterOperator: "eq",
        filterValue: "admin",
    }
});
```
And by using the filter fields in the example we dont get user that have more roles than just the admin role.
Let say whe have a user that has the roles ["user", "admin"]. Those are not returned and i believe that is not the result most people expect.

Supporting 'contains' operator could solve the problem.

^1.2.7

```bash
- OS: Windows 10
- Browser: Firefox
```

Package, Other

```typescript

  plugins: [
    admin()
  ],
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

thisisleobro:
Motivation to allow 'contains' was so that i could filter by role when calling authClient.admin.listUsers().
Currently 'eq' (equals) is the only valid operator for filtering by role but it only works if role is the unique role set on user.
A user with roles 'admin' and 'user' can not be filtered using 'admin' nor 'user' as filter value

Closes #2588

### DIFF
--- a/packages/better-auth/src/plugins/admin/admin.ts
+++ b/packages/better-auth/src/plugins/admin/admin.ts
@@ -518,7 +518,7 @@ export const admin = <O extends AdminOptions>(options?: O) => {
 							.or(z.boolean())
 							.optional(),
 						filterOperator: z
-							.enum(["eq", "ne", "lt", "lte", "gt", "gte"], {
+							.enum(["eq", "ne", "lt", "lte", "gt", "gte", "contains"], {
 								description: "The operator to use for the filter",
 							})
 							.optional(),


Kinfe123:
it is already added along with the other changes. thanks!

PR #2626, opened 2025-05-12
docs: fix getFullOrganization example to use correct query parameter structure

### LINKED ISSUE #2593: Organizations: getFullOrganization() has incorrect example in docs.
Currently in [docs ](https://www.better-auth.com/docs/plugins/organization) we have this:

```
const organization = await authClient.organization.getFullOrganization({
    organizationId: "organization-id" // optional, by default it will use the active organization
})
//you can also use organizationSlug instead of organizationId
const organization = await authClient.organization.getFullOrganization({
    organizationSlug: "organization-slug"
})
```

This doesn't work, **but this works**:

```
const organization = await authClient.organization.getFullOrganization({
    query: { organizationId: "organization-id" } // optional, by default it will use the active organization
})
//you can also use organizationSlug instead of organizationId
const organization = await authClient.organization.getFullOrganization({
    query: { organizationSlug: "organization-slug" }
})
```

Kinfe123:
closes #2593

### DIFF
--- a/docs/content/docs/plugins/organization.mdx
+++ b/docs/content/docs/plugins/organization.mdx
@@ -384,12 +384,12 @@ To get the full details of an organization, you can use the `getFullOrganization
   <Tab value="client">
   ```ts title="auth-client.ts"
   const organization = await authClient.organization.getFullOrganization({
-      organizationId: "organization-id" // optional, by default it will use the active organization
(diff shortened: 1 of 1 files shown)


leebenson:
Docs still show the old example: https://www.better-auth.com/docs/plugins/organization#get-full-organization

PR #2628, opened 2025-05-12
fix: encode/decode magic link callbackURL

### LINKED ISSUE #2592: Magic Link callbackurl query params are not merged

### Issue

In short the url ends up being this (noticed the two ? in the query params)
```
http://localhost:5173/login?from_better_auth=true?error=INVALID_TOKEN
```

### To Reproduce
Send a magic link with a query param in callbackURL:
```ts
    const {error: err} = await authClient.signIn.magicLink({
      email,
      callbackURL: "http://localhost:5173/login?from_better_auth=true"
    });
```

The magic link url is this, which is correct
```
http://localhost:3000/api/auth/magic-link/verify?token=lqqowxkhhubcfveaxgpaztvgdmwdmsar&callbackURL=http://localhost:5173/login?authenticated=true
```

Better auth then adds another query param when it redirects from the verify endpoint, but as an extra query not as an extra param:
```
http://localhost:5173/login?from_better_auth=true?error=INVALID_TOKEN
```

This should be with an "&":
```
http://localhost:5173/login?from_better_auth=true&error=INVALID_TOKEN
```

Currently better auth appends query param by adding "?error=asd" when it should be appending the param by using the URL constructor.

1.1.21

```bash
Windows 10
Firefox
```

Backend, Client

```typescript
      plugins: [
        magicLink({
          expiresIn: AUTH_MAGIC_LINK_EXPIRY_SECONDS, // 1 hour
          sendMagicLink: async ({email, url}, request) => {
            await email_login_magic_link(email, url);
          }
        }),
      ],
});
```

philipp-lampert:
Fixes an issue preventing **magic link** authentication from correctly redirecting to **callbackURL**s with **query parameters**. This is crucial for restoring post-login redirection to specific states.

Closes #2592

### DIFF
--- a/packages/better-auth/src/plugins/magic-link/index.ts
+++ b/packages/better-auth/src/plugins/magic-link/index.ts
@@ -127,9 +127,9 @@ export const magicLink = (options: MagicLinkOptions) => {
 					);
 					const url = `${
 						ctx.context.baseURL
-					}/magic-link/verify?token=${verificationToken}&callbackURL=${
+					}/magic-link/verify?token=${verificationToken}&callbackURL=${encodeURIComponent(
 						ctx.body.callbackURL || "/"
-					}`;
+					)}`;
 					await options.sendMagicLink(
 						{
 							email,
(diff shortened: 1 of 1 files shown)


philipp-lampert:
Hi @Bekacru

would be great if you could have a look at this very short PR. It is a best practice to URI-encode any URLs passed as query params themselves, as is the case for callback URLs. Very simple fix!

Thanks!

PR #2634, opened 2025-05-13
docs(cli): Warn to not use Bun

ping-maxwell:
Lots of users run into errors when using the Better-Auth CLI to run `generate` or `migrate`. This adds a warning in the docs to advise against it.


Bekacru:
should work with bun

PR #2636, opened 2025-05-13
Support postgres array types

JeremyMoeglich:
I have a custom plugin with a string[] type and that resulted in a warning despite me using a postgres array type.
```
2025-05-13 10:26:10.237 WARN  servercode/src/lib/auth/auth.ts:107 Field roles in table user has a different type in the database. Expected string[] but got _text.
```

Looking at the code it only allowed json previously, this is a small patch that adds various array types into the check.


himself65:
Closing as staling

Issue #2519, opened 2025-05-03
Custom error page (errorURL in onAPIError) not used for all errors


ErikPetersenDev:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

This occurs with several issues, but this is one consistent way of doing it:

1. Set a custom error page in Better Auth options
2. Log in with Google.
3.  Hit the back button to go back to the Google sign-in screen.
4.  Choose the same account again.

Default Better Auth error page is shown.

### Current vs. Expected behavior

Expected:  If a custom error page is defined it should be used for all API errors.

### What version of Better Auth are you using?

1.2.7

### Provide environment information

```bash
- OS: MacOS
- Browser: Chrome
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
relevant portion:


  onAPIError: {
    throw: false,
    onError: (error) => {
      console.error("Auth error:", error);
    },
    errorURL: "/auth-error",
  },
```

### Additional context

I found code in `node_modules` for the specific error I was generating (see reproduction steps above).  The code has the error url hardcoded to the base Better Auth error page instead of honoring the custom URL in auth config.

From the code - see it's redirecting to baseURL/error:
https://github.com/better-auth/better-auth/blob/2a246aeb8735246afb405f705cc5e535f330e303/packages/better-auth/src/oauth2/state.ts#L63
```
  if (!data) {
    c.context.logger.error("State Mismatch. Verification not found", {
      state
    });
    throw c.redirect(
      `${c.context.baseURL}/error?error=please_restart_the_process`
    );
  }
```

Here's an example from elsewhere in the codebase that does honor the override.  It checks options for "onAPIError"
```
          const defaultErrorURL = ctx.context.options.onAPIError?.errorURL || `${ctx.context.baseURL}/error`;
          if (ctx.query.error || !ctx.query.code) {
            throw ctx.redirect(
              `${defaultErrorURL}?error=${ctx.query.error || "oAuth_code_missing"}&error_description=${ctx.query.error_description}`
            );
          }
```

ErikPetersenDev:
Helpful search to find places in the code with this bug:  https://github.com/search?q=repo%3Abetter-auth%2Fbetter-auth%20baseURL%7D%2Ferror%3F&type=code

coopbri:
[...]

PR #2640, opened 2025-05-13
Use custom errorURL when available

ErikPetersenDev:
This PR resolves: https://github.com/better-auth/better-auth/issues/2519

### Summary
There are a few places in the code where the custom errorURL is not checked/used, resulting in the default Better Auth error page.  This PR copies the way this is done elsewhere to check for the custom errorURL in context before deciding which URL to use. 

### Notes
Future improvement opportunities:  centralize the logic so it isn't duplicated throughout the codebase as it is today.  Or potentially have a higher level error function that handles throwing the redirect with the error URL.  I kept it simple for this bug fix though.

### DIFF
--- a/packages/better-auth/src/oauth2/link-account.ts
+++ b/packages/better-auth/src/oauth2/link-account.ts
@@ -31,9 +31,9 @@ export async function handleOAuthUserInfo(
 				"Better auth was unable to query your database.\nError: ",
 				e,
 			);
-			throw c.redirect(
-				`${c.context.baseURL}/error?error=internal_server_error`,
-			);
+			const errorURL =
+				c.context.options.onAPIError?.errorURL || `${c.context.baseURL}/error`;
+			throw c.redirect(`${errorURL}?error=internal_server_error`);
 		});
 	let user = dbUser?.user;
 	let isRegister = !user;
--- a/packages/better-auth/src/oauth2/state.ts
+++ b/packages/better-auth/src/oauth2/state.ts
@@ -62,9 +62,8 @@ export async function parseState(c: GenericEndpointContext) {
 		c.context.logger.error("State Mismatch. Verification not found", {
 			state,
 		});
-		throw c.redirect(
-			`${c.context.baseURL}/error?error=please_restart_the_process`,
-		);
+		const errorURL = c.context.options.onAPIError?.errorURL || `${c.context.baseURL}/error`;
+		throw c.redirect(`${errorURL}?error=please_restart_the_process`);
 	}
 	const parsedData = z
 		.object({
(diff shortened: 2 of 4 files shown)


ItsWendell:
I also just posted a message in the Discord, but also sharing here for context:

I run better-auth as a separate service from my main backend applications. My goal was to improve the user experience during error scenarios by redirecting users back to the originating application with clearer error messages.

To try to achieve this, I enabled `onApiError.throw` and implemented a custom `onApiError.onError` handler. The intention was to use the context available in onError to determine the type of request and then redirect the user with a specific error message / code to the appropriate application, instead of using errorURL.

Quickly scanning through this code, I noticed that in error cases the oauth plugin simply `throws`  a redirect, I think for this reason onError is never even called.

Could we explore this further? It seems that the error propagation mechanism within better-auth, the OAuth plugin, and other plugins, might need to be properly reviewed to ensure it works as intended and before bad / unintentional design patterns emerge into the fast growing ecosystem of better-auth community plugins.

ErikPetersenDev:
@ItsWendell Yeah, that makes sense.  This PR was intended as a quick fix without me digging into the codebase much.  In my original "Note" I mentioned that it seemed like a broader change might be needed, and it seems like you've looked at it enough to find more of the root cause.

Without looking into how it works now, it makes sense from your discussion that it _should_ work something like this:

1. Code anywhere in BA, including plugins, hits an unrecoverable error and throws.
2. A centralized place in the stack catches those errors.
3. First checks if the user has configured throwing on errors -> rethrows if yes, for the user to manage.
4. Then checks custom error URL -> redirect to custom or default errorURL based on that.

That seems like a decent pattern, which would mean there should be no places in the code that throw redirects other than that centralized location (barring exceptions, or maybe there are a couple places where we'd need the catch).  I would actually benefit from this too, since at least one of the current OAuth "redirect to error page" errors is something where I'd rather redirect the user somewhere else (not an error page) but the hardcoded redirect doesn't allow that.  So effectively there are two issues:  1) custom errorURL is not being respected due to the implementation of redirects (fixed by this PR), and 2) throw on API error not being respected due to redirects where they don't belong (not addressed by this PR).

I'll leave it to the maintainers to decide what to do next:

1. Merge this PR as-is first -->  narrow scope: ensures current places where we redirect respect the custom errorURL, without addressing _whether_ those redirects should be there at all.  Then open a second issue/PR to fix the broader issue.
2. Close this PR and someone opens another PR to fix the broader issue, which would still address the original Issue.

ItsWendell:
@ErikPetersenDev Great response and totally agree, this quick fix would already be beneficial. How error handling is handled design-wise within better-auth should be a separate issue!

FalconiZzare:
Any update on this yet?

PR #2641, opened 2025-05-13
Fix OIDC Post-Login Flow

### LINKED ISSUE #984: OIDC-Provider issue with redirects when signing in using the client.signIn method

### To Reproduce

1. Set up the oidcProvider Plugin with next.js
2. Create an account and register a oidc application
3. I'm using postman to test the oauth flow
4. Go through the oauth flow using the registered oidc application, make sure that you're not signed in to the provider
5. You will be redirected to the login page, now sign in and watch the network tab
6. It follows the redirect after login, but should instead use the frameworks router?

Currently i have to use the following to redirect to the consent page after signing in because fetch otherwise would follow the redirect.

```ts
  const searchParams = useSearchParams();
  const router = useRouter();

  async function onSubmit(values: z.infer<typeof loginSchema>) {
    try {
      await authClient.signIn.email({
        ...values,
        fetchOptions: {
          redirect: "error",
          onError(context) {
            toast.error(context.error.message);
          },
        },
      });
    } catch {
      const clientId = searchParams.get("client_id");
      const scope = searchParams.get("scope");

      if (clientId && scope) {
        router.push(`/consent?client_id=${clientId}&scope=${scope}`);
      }
    }
  }
```

The ideal solution would be to use the frameworks redirect or document this behaviour

1.1.1

```bash
- OS: macOS Sonoma 14.5
- Browser: Chrome
```

Backend, Client

```typescript
import { oidcProvider } from "better-auth/plugins";
import { db } from "@/db";

  emailAndPassword: {
    enabled: true,
  },
    provider: "pg",
  }),
  plugins: [
    nextCookies(),
    oidcProvider({
      loginPage: "/sign-in",
      consentPage: "/consent",
    }),
  ],
});
```

### Additional context

_No response_

### LINKED ISSUE #1160: OIDC redirect issue/

### To Reproduce

1. Create OIDC service (central-auth-app);

```
    provider: "pg",
    schema: schema,
  }),
  emailAndPassword: {
    enabled: true,
  },
  baseUrl: process.env.BETTER_AUTH_URL,
  plugins: [
    oidcProvider({
      loginPage: "/login",
      requirePKCE: true,
      consentPage: "/oauth/consent",
    }),

    openAPI(),
  ],
});
```

  baseURL: "https://central-auth-app.com",
  plugins: [oidcClient()],
});

2. Register a client "client-app" oauth/register.

3. Create a "client-app" which uses "central-auth-app.com" in generic oauth plugin.

```

    provider: 'pg',
    schema: schema,
  }),
  emailAndPassword: {
    enabled: true,

  },

  account: {

    accountLinking: {
      enabled: true,
      trustedProviders: ['test'],
    },
  },

  baseUrl: 'http://client-app.com',
  plugins: [
    genericOAuth({
      config: [
        {
          providerId: 'test',
          clientId: 'nvkafqizujtfacldfdvrettvdkvpiajr',
          clientSecret: 'ixndgtbzrnnkncylkfvmorytwibfselp',
          authorizationUrl: 'https://https://central-auth-app.com/api/auth/oauth2/authorize',
          tokenUrl: 'https://https://central-auth-app.com/api/auth/oauth2/token',
          scopes: ['openid', 'profile', 'email'],
          pkce: true,
        },
      ],
    }),

  ],
})
```

4.  In client app, button onClick event
 ```
 const data = await authClient.signIn.oauth2(
            {
              providerId: 'test',
              callbackURL: '/reports/property',
            },
            {
              throw: false,
            },
          )
```
This will redirect to the https://central-auth-app.com/login?client_id=things...other query params.

5. In central auth app, when login button is clicked this event should redirect to consent or client screen correctly.

```
 const res = await authClient.signIn.email({
        email: data.email,
        password: data.password,
      });

```

I get the 302. but redirect doesn't happen. It fetches the consent document instead. and stay stucked.

image

I am able to fix this redirect using fetchOptions.
```
  const res = await authClient.signIn.email({
        email: data.email,
        password: data.password,
        fetchOptions: {
          onSuccess(context) {
            if (context.response.redirected) {
              window.location.href = context.response.url;
            }
          },
        },
      });
```
It goes to consent screen after this. and everything works fine. 

> Tried, redirect:"follow", it didn't work.

Now, consent is accepted. I sign out and try to sign in again. It doesn't have to go through consent screen again.
Instead it can just redirect.

But, i get cors error. Instead of redirecting its try to fetch the redirect url. And the fetch options doesn't work here.

image

Is this how it supposed to be or i am missing some step here?

1.1.10

```bash
OS: macos
Browser: chrome, safari and firefox.
```

Client

_No response_

### Additional context

Tried on localhost and vercel subdomain.

IhsenBouallegue:
## Fix OIDC Post-Login Flow and Consent Handling

Closes #1160 
Closes #984 

**Addresses:**
-   CORS errors when an unauthenticated user logs in via the OIDC provider, and the provider attempts to redirect back to the client application.

**Summary of Issues:**

1.  **CORS Error on Post-Login Redirect:**
    When an unauthenticated user was redirected to the OIDC provider's internal login page, upon successful login (e.g., via `signIn.email`), the OIDC plugin's `after` hook would attempt to resume the OIDC flow. It did this by re-running the authorization logic and directly issuing an HTTP 302 redirect from the `POST /api/auth/sign-in/email` response. This 302 redirected to the client application's callback URL. Browsers have strict CORS policies for redirects originating from `fetch` POST requests, especially if they appear cross-origin or involve complex redirect chains, leading to the observed CORS errors and the redirect failing.

2.  **Consent Screen Skipped After Internal Login:**
    The OIDC `authorize` endpoint's decision to show a consent screen was primarily based on an explicit `prompt=consent` parameter in the URL. When the OIDC flow was resumed by the `after` hook post-login, the original `prompt` parameter (which might have been undefined or different) was used. This often led to the consent check being bypassed, even if OIDC best practices suggest prompting for consent at least the first time a user/client combination requests specific scopes.

**Solution Implemented:**

1.  **Modified OIDC Provider `after` Hook (`src/plugins/oidc-provider/index.ts`):**
    -   Instead of the `after` hook directly causing a 302 redirect, it now returns a 200 OK JSON response to the `signIn.email` (or equivalent login) call.
    -   This JSON response includes:
        -   Standard user and session token information.
        -   A flag `oidcContinuation: true`.
        -   An `oidcAuthorizeURL`: This is the URL to the OIDC provider's own `/oauth2/authorize` endpoint, constructed with the original OIDC request parameters (from the `oidc_login_prompt` cookie).
    -   Crucially, when constructing this `oidcAuthorizeURL`, the hook now intelligently sets `prompt=consent` if the original OIDC request did not specify `prompt=none`. This ensures that the subsequent call to `/oauth2/authorize` will properly evaluate the need for consent.
    -   This change ensures that the `Set-Cookie` headers for the new session and for clearing the `oidc_login_prompt` are correctly handled by the framework via `ctx.json()`.

2.  **Client-Side Change on OIDC Provider's Login Page:**
    -   The client-side JavaScript that calls `authClient.signIn.email(...)` on the OIDC provider's login page must be updated in its `onSuccess` handler.
    -   It now checks for `context.data.oidcContinuation === true`.
    -   If true, it performs a `window.location.href = context.data.oidcAuthorizeURL;`. This action by the client initiates a new, full-page GET navigation to the OIDC provider's `/oauth2/authorize` endpoint.

3.  **OIDC `authorize` Endpoint (`src/plugins/oidc-provider/authorize.ts`):**
    -   When this endpoint is hit via the GET request (from the client-side navigation):
        -   It finds the active session (established by the preceding `signIn.email`).
        -   It now correctly receives `prompt=consent` (if set by the `after` hook).
        -   The existing logic for `if (query.prompt === 'consent')` then correctly proceeds to check `hasAlreadyConsented`. If consent is needed (user hasn't consented to these scopes for this client before), the consent page is shown.
        -   If consent has already been granted, it issues a 302 redirect to the client application's callback URL. This redirect is now in response to a GET request and is handled correctly by browsers without the previous CORS issues.

**Impact:**

-   The CORS error during the post-login redirect in the OIDC flow is resolved because the redirect to the client application now originates from a GET request to the `/oauth2/authorize` endpoint.
-   The consent screen is more reliably presented after a user logs in via the OIDC provider's login page if they haven't previously consented to the requested scopes for the client, or if `prompt=consent` was part of the original request. The system now also respects an explicit `prompt=none` if present in the original request to allow for silent authentication attempts.
-   The solution aligns with standard web practices by having the client initiate navigations that change the top-level browser context, rather than trying to achieve this via redirects from background `fetch` POST requests.
-   Logging has been reviewed and refined in both modified files to be more concise for production while retaining essential information for debugging.


zackify:
im seeing the consent screen come up even if its none, and if i do not specify one in the oidc config.

so then i ended up setting it to a page, and on that page, just doing this on the backend:

```
 const data = await auth.api.oAuthConsent({
        body: { accept: true },
        query: req.query,
        headers: req.headers,
      });
      res.redirect(data.redirectURI);
```

and then it says that consent isnt required and breaks. i am wondering if this PR fixes the problem. i dont want a consent screen but there is no way to turn it off, it errors if you do not specify the page or html in the config.

IhsenBouallegue:
Hey @zackify . Thanks for trying it out. 
Regarding not specifying the consent screen in the config resulting in an error, that's a current bug and has nothing to do with the PR. I encourage you to test it on main.
Now the other problem.
I did set consent as the default (as that's what I expect from an identity provider). So when specifying nothing it should add prompt=consent to the URL. But it should definitely not override none.
I will double check that behavior.

The PR mainly focuses on fixing the navigation and CORS problem that was appearing because of the structure of the requests.

zackify:
Thanks for the reply, sounds good! I did make a PR myself that seems to fix the ability to mark consent as given. Curious if you have any thoughts on it 😁 it’s very small. 

https://github.com/better-auth/better-auth/pull/2723

The big you mention of skipping consent completely would also avoid this. 

But my PR seems to make it possible to have a custom consent screen, then mark it as accepted server side. In my case I don’t want to render frontend JS if I do add a consent screen. I just want to post to my server and mark it accepted to avoid any client js in the auth service, so this server method is great if only it worked without state. Sorry if I hijacked your pr!

PR #2643, opened 2025-05-13
fix: use dynamic list of social providers to allow generic oauth

### LINKED ISSUE #2610: Support getAccessToken for genericOAuth

Consider an `auth.ts` using `genericOAuth` plugin:
```ts
  plugins: [
    genericOAuth({
      config: [
        {
          providerId: "keycloak",
          ...
        },
      ],
    }),
  ],
  ...
});
```

Then try to get the access token with:
```ts
await auth.api.getAccessToken({
  body: {
    providerId: "keycloak",
  },
  headers: Astro.request.headers,
});
```

Gives error:
```
[ERROR] Provider keycloak is not supported.
```

I would love for `auth.api.getAccessToken` to work with the `genericOAuth` plugin.

Based on existing source code, I made this function that can be used instead:

```ts
import { refreshAccessToken } from "better-auth/oauth2";

export const getAccessToken = async (userId: string) => {
  const ctx = await auth.$context;
  const [account] = await ctx.internalAdapter.findAccountByUserId(userId);
  if (
    account.refreshToken &&
    account.accessTokenExpiresAt &&
    account.accessTokenExpiresAt.getTime() - Date.now() < 5_000
  ) {
    try {
      const res = await refreshAccessToken({
        refreshToken: account.refreshToken,
        tokenEndpoint: OIDC_TOKEN_URL,
        options: {
          clientId: OIDC_CLIENT,
          clientSecret: OIDC_SECRET,
        },
      });
      await ctx.internalAdapter.updateAccount(account.id, {
        accessToken: res.accessToken,
        accessTokenExpiresAt: res.accessTokenExpiresAt,
        refreshToken: res.refreshToken,
        refreshTokenExpiresAt: res.refreshTokenExpiresAt,
      });
      console.log("Refreshed access token.");
      return res.accessToken;
    } catch (e: any) {
      console.error("Failed to refres access token:", e);
      return null;
    }
  }
  return account.accessToken;
};
```

### Additional context

Using Astro integration.
```
"astro": "^5.7.12",
"better-auth": "1.2.8-beta.7"
```

BlankParticle:
fixes #2610
ref https://github.com/better-auth/better-auth/pull/2557#issuecomment-2876979264

the endpoints now use the soclal provider list in context so that generic oauth plugin providers are also allowed in token endpoints

### DIFF
--- a/packages/better-auth/src/api/routes/account.ts
+++ b/packages/better-auth/src/api/routes/account.ts
@@ -1,9 +1,5 @@
 import { z } from "zod";
 import { createAuthEndpoint } from "../call";
-import {
-	socialProviderList,
-	type SocialProvider,
-} from "../../social-providers";
 import { APIError } from "better-call";
 import { generateState, type OAuth2Tokens } from "../../oauth2";
 import {
@@ -12,6 +8,7 @@ import {
 	sessionMiddleware,
 } from "./session";
 import { BASE_ERROR_CODES } from "../../error/codes";
+import { SocialProviderListEnum } from "../../social-providers";
 
 export const listUserAccounts = createAuthEndpoint(
 	"/list-accounts",
@@ -106,9 +103,7 @@ export const linkSocialAccount = createAuthEndpoint(
 			/**
 			 * OAuth2 provider to use
 			 */
-			provider: z.enum(socialProviderList, {
-				description: "The OAuth2 provider to use",
-			}),
+			provider: SocialProviderListEnum,
 			/**
 			 * Additional scopes to request when linking the account.
 			 * This is useful for requesting additional permissions when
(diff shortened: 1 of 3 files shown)


IhsenBouallegue:
I was going to implement this right now and I saw it was already done. Amazing job. My app is depending on it so would love to see it on main soon 👍

IhsenBouallegue:
would close #2610

Bekacru:
the issue with this is, we'll lose the type inference for the built in providers which we don't want. Maybe, we can make it union of either those providers or string so it at least auto completes for the builtin ones.

BlankParticle:
> the issue with this is, we'll lose the type inference for the built in providers which we don't want. Maybe, we can make it union of either those providers or string so it at least auto completes for the builtin ones.

I can try doing a union if that works

BlankParticle:
image
image

ok now it works as you wanted 
@Bekacru

alexandre-hallaine:
Amazing work!

PR #2666, opened 2025-05-16
feat: MCP plugin

Bekacru:


kziemski:
@Bekacru managed to get this to work in rrv7/cf-workers and mcpagent from cf's agents sdk  by inserting into the fetch handler and checking the request pathname sending it.  
withMcpAuth acts as a guard but i'm wondering if it should ignore anything other than /mcp/* or /sse/* .well-known/oauth-auth.... in the block or have a default handler. that way you can wrap withmcpagent at the root and it would work for non mcp and mcpauth related routes.

kziemski:
@Bekacru ty man this is awesome either way. i got stuck trying to get oidcprovider plugin to work for this and realized i missed something there.

zackify:
Curious, what is the reason for an mcp plugin inside better auth? I setup an mcp server using the official protocol last night and used better auth as my oauth2 server all in express. 

Since the official protocol has the proxy provider methods, it all worked very well.

kziemski:
@zackify what version of better-auth did you use?
I initially went down that route and was getting an error with the flow.
did you use the router and the proxyprovider from ts-sdk server/auth? 


mcp plugin works @Bekacru just a heads up #2703 
Was going to submit a pr but was curious if it made sense to be able to override that and not just append the baseUrl

zackify:
There's literally 3 different bugs with the official mcp inspector and the proxy provider in the sdk. 

First the inspector doesn't do the oauth flow unless you downgrade to v11. 

Then the well known file from the real integration hard codes the oauth url so I had to make my own well known route above the middleware. It's bad haha. 

I can share specific code if you need later.

kziemski:
@zackify its amazing that the entire ecosystem is adopting MCP but here we are. 
need to push down the screaming voice in the back of my head about better... and just grind.
would be greatly appreciated if you post a link to see what you did it if only to figure out what i did wrong ;)

submitted an issue on the hardcode for the plugin

zackify:
i dont have time yet and my project isnt open source, but it will be when im done.

looks like the new update fixes the problems: oauth works for streaming servers correctly now: https://github.com/modelcontextprotocol/inspector/releases/tag/0.13.0 give it a try

dead8309:
for someone looking on how to use this plugin on cloudflare-workers
https://github.com/dead8309/better-auth-mcp-hono

kziemski:
> i dont have time yet and my project isnt open source, but it will be when im done.
> 
> looks like the new update fixes the problems: oauth works for streaming servers correctly now: https://github.com/modelcontextprotocol/inspector/releases/tag/0.13.0 give it a try

thnx, just wondering what version of better-auth you're using and did you use the proxy provider from the sdk server to implement it.

ya was going to tackle it again not that 0.13 is out. I think the plugin works fine they just need to fix the WWW-Authenticate-Value not show stopper.

zackify:
The latest version. 

Proxy provider from the sdk. 

And streamable on a different express route. So I’m not sure how mcp fits into this library when it works that way. 

Check the console in the inspector if with isn’t working and let me know what part you’re stuck at I can try to share that part

kziemski:
@zackify as a guard its ok. 
I mean it works 100%, just have to patch the unauthorized returned value which is hardcoded not to the baseurl. But ya thank you i'll let you know. Moving over to tanstack this weekend so will take another crack at.

packages/better-auth/src/plugins/mcp/index.ts
```diff
@@ -0,0 +1,928 @@
...
+		});
+	}
+	return {
+		issuer,
+		authorization_endpoint: `${baseURL}/mcp/authorize`,
+		token_endpoint: `${baseURL}/mcp/token`,
+		userinfo_endpoint: `${baseURL}/mcp/userinfo`,
+		jwks_uri: `${baseURL}/mcp/jwks`,
+		registration_endpoint: `${baseURL}/mcp/register`,
+		scopes_supported: ["openid", "profile", "email", "offline_access"],
```
zackify, line 51: for claude max and adding a remote mcp, the scope actually must menton "claudeai" as one of them

packages/better-auth/src/plugins/mcp/index.ts
```diff
@@ -0,0 +1,928 @@
...
+		response_types_supported: ["code"],
+		response_modes_supported: ["query"],
+		grant_types_supported: ["authorization_code"],
```
zackify, line 54: also we need ""refresh_token"" support or else the client just logs in constantly each hour

PR #2678, opened 2025-05-16
Verify password in auth.api.verifyPassword

izadoesdev:
### Added

- New `verifyPassword` API endpoint to check if a user's password is valid.
  - Accepts either the authenticated user or an email.
  - Returns `{ valid: boolean }` in the response.

### Tests

- Added comprehensive tests for the `verifyPassword` endpoint, covering:
  - Authenticated and unauthenticated (email) flows
  - Valid and invalid password checks
  - Error cases for missing email and non-existent users


izadoesdev:
@ping-maxwell 😉 (i doubt it'll get approved since i noticed you had something similar in re-verify, but would be nice to have this)

PR #2679, opened 2025-05-16
feat: admin emails

mmkal:


mmkal:
@himself65 do you think this will get merged?

himself65:
I don't think this is a good case. You can wrap the API to support multiple users by email, but adding this will have extra complexity.

mmkal:
Hmm , I don't think you can define email patterns that will make certain accounts admins declaratively, ahead of them existing though can you?

It's not that big a deal, just a bit fiddly to have to edit the db to make certain accounts admins even in test envs.

PR #2695, opened 2025-05-18
docs: custom password validation

matt-shipman:
Added documentation for how to implement custom password validation for email and password authentication.


Bekacru:
Thanks for the PR but I suggest using database hooks instead, since password can be updated with different endpoints. We'll have a guide about it soon.

matt-shipman:
Good to know. Thanks for the response. Looking forward to the guide.

matt-shipman:
@Bekacru I tried to switch to using `databaseHooks.account.create.before`, however `account.password` is the hashed version of the password. The only way I can seem to get the raw password is through `ctx?.body.password`. Is there a better method for getting the raw password in a database hook?

Issue #2501, opened 2025-05-01
Stripe webhooks on cloudflare (opennextjs) - No webhook payload was provided


zpg6:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Deploy to a opennextjs-cloudflare project with stripe plugin configured and send a webhook request.

### Current vs. Expected behavior

I expect that it works out of the box when I configure the webhook in Stripe dashboard with `mysite.com/api/auth/stripe/webhook` and put `STRIPE_WEBHOOK_SECRET` and other env vars in.

In my case (opennextjs-cloudflare) Cloudflare worker logs the request body isn't there
```
[Better Auth]: No webhook payload was provided.
```

And in Stripe logs I get 400 with:
```json
{
  "code": "WEBHOOK_ERROR_NO_WEBHOOK_PAYLOAD_WAS_PROVIDED",
  "message": "Webhook Error: No webhook payload was provided."
}
```

When I go to print the ctx I get
```
{
  headers: {},
  params: {},
  request: {},
  body: {
    data: {
      object: {...}, // The actual object payload
    },
    livemode: false,
    id: "evt_1RJmXVETs...............",
    object: "event",
    api_version: "2025-03-31.basil",
    type: "checkout.session.completed",
    ...
  }
}
``` 

### What version of Better Auth are you using?

v1.2.8-beta.3

### Provide environment information

```bash
N/A
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
// stripe.ts
export const stripe = new Stripe(process.env.STRIPE_SECRET_KEY || "sk_test", {
    apiVersion: "2025-03-31.basil",
    httpClient: Stripe.createFetchHttpClient(),
});

// in the config
stripePlugin({
    stripeClient: stripe,
    stripeWebhookSecret: process.env.STRIPE_WEBHOOK_SECRET,
    createCustomerOnSignUp: true,
    onCustomerCreate: async ({ customer, user }) => {
        console.log(`Customer ${customer.id} created for user ${user.id}`);
    },
    subscription: {
        enabled: true,
        plans: plans,
        ...
    },
}),
...
```

zpg6:
I was able to get it working (200 on stripe) by changing to this:

```typescript
let event: Stripe.Event;

if (ctx.body?.object === "event") {
	event = ctx.body;
	console.log("event", event);
}
else {
         // ... how it is now
         const buf = await ctx.request.text();
[...]

PR #2706, opened 2025-05-18
disableBody option in Stripe webhook endpoint

deestt:
This PR is to fix a plugin that doesn't work on Cloudflare Workers (#2501) because CF Workers behaves differently when duplicating a request (gives empty text). Here's the Pull Request better-call repo: https://github.com/Bekacru/better-call/pull/18


f1shy-dev:
+1 on this getting merged, is really affecting me in prod

PR #2717, opened 2025-05-19
email otp verification guide added

snehendu098:


Kinfe123:
can you please fix most the indentation issue on the code snippet since it should be consistent with others snippets with in the docs.

PR #2736, opened 2025-05-21
fix: sidebar placement issue with header

johurul-haque:
Before: Top dropdown gets cut off by the header and the sidebar border does not stretch fully to the bottom of the screen

image

After: Top dropdown is not cut off by header as the top value of it is now set to the header height explicitly. The sidebar border now stretches to the bottom of the screen by setting the height to `calc(100dvh - 56px)`, 56px is the current header height.

image

Getting the header height via Javascript and then setting the value via CSS custom properties would have been a more future proof solution. But at the moment, this is a simple fix without tweaking much.

### DIFF
--- a/docs/components/side-bar.tsx
+++ b/docs/components/side-bar.tsx
@@ -39,7 +39,7 @@ export default function ArticleLayout() {
 			<aside
 				className={cn(
 					"md:transition-all",
-					"border-r border-lines top-[43px] md:flex hidden md:w-[268px] lg:w-[286px] overflow-y-auto absolute h-[calc(92dvh-7px)] flex-col justify-between w-[var(--fd-sidebar-width)]",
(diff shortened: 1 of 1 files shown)


Kinfe123:
Thanks for the PR, but it has already been fixed earlier. Good catch!

johurul-haque:
The updated fix does address the raised issues, but introduces new ones.

The sidebar's height is set to 100dvh, the full height of the viewport, and a top value of 53px. So, the full viewport height sidebar is offset by 53px from the top, and as a result, its contents go below the screen.  

In the screenshot below, the Reference menu item is not showing, even when I've scrolled to the bottom, due to being offset below the screen by 53px. You can reproduce this by opening the Authentication menu and scrolling down below.

The offset value from the top needs to be accounted for in the height. Setting `h-[calc(100dvh - {TOP_VALUE})`

image

Additionally, the dropdown at the top is also cut off by 3px, as the header size is rendered as 56px, not 53px.

The green background from devtools indicates that the top padding of the dropdown is half below the header. Notice the header border between the green background.

image

Issue #2729, opened 2025-05-21
Log in with apple idToken. The callbackURL in the received email verification URL is undefined.


gee1k:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

![Image](https://github.com/user-attachments/assets/2b58017b-de76-4e29-8751-2baa5affa578)

Verification URL: https://xxx.com/api/auth/verify-email?token=_m_2yxOf_8tl9bsrBFePZrT6o&callbackURL=undefined

### Current vs. Expected behavior

Verification URL: https://xxx.com/api/auth/verify-email?token=_m_2yxOf_8tl9bsrBFePZrT6o&callbackURL=undefined

When registering with an email, you will receive the correct URL.

Verification URL: https://xxx.com/api/auth/verify-email?token=_m_2yxOf_8tl9bsrBFePZrT6o&callbackURL=exp:///verify-email

### What version of Better Auth are you using?

1.2.8

### Provide environment information

```bash
Expo SDK 53
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
import { expo } from '@better-auth/expo';
export const auth = betterAuth({
  plugins: [expo()],
  trustedOrigins: ['exp://'],
  database: createPool({
    host: process.env.DATABASE_HOST,
    port: parseInt(process.env.DATABASE_PORT, 10) || 3306,
    user: process.env.DATABASE_USERNAME,
    password: process.env.DATABASE_PASSWORD,
    database: process.env.DATABASE_NAME,
  }),
  emailAndPassword: {
    enabled: true,
    requireEmailVerification: true,
    autoSignIn: true,
  },
  emailVerification: {
    sendOnSignUp: true,
    autoSignInAfterVerification: true,
    sendVerificationEmail: async ({ user, url, token }, request) => {
      await sendMail({
        to: user.email,
        subject: 'Verify your email address',
        body: `Click the link to verify your email: ${url} `,
      });
    },
  },
  session: {
    freshAge: 10,
    modelName: 'sessions',
  },
  user: {
    modelName: 'users',
    additionalFields: {
      role: {
        type: 'string',
        defaultValue: 'user',
      },
      clientId: {
        type: 'number',
        defaultValue: '1',
      },
    },
  },
  account: {
    modelName: 'accounts',
    accountLinking: {
      enabled: true,
      trustedProviders: ['google', 'apple', 'email-password'],
[...]

PR #2740, opened 2025-05-22
fix: Improve the callbackURL parameter for social, oauth, SSO

gee1k:
Fix the issue where the callbackURL parameter is missing when emailVerified is false using the idToken method for social login.

Fixed: #2729

### DIFF
--- a/packages/better-auth/src/api/routes/sign-in.ts
+++ b/packages/better-auth/src/api/routes/sign-in.ts
@@ -285,6 +285,7 @@ export const signInSocial = createAuthEndpoint(
 					accountId: userInfo.user.id,
 					accessToken: c.body.idToken.accessToken,
 				},
+				callbackURL: c.body.callbackURL,
 				disableSignUp:
 					(provider.disableImplicitSignUp && !c.body.requestSignUp) ||
 					provider.disableSignUp,
--- a/packages/better-auth/src/plugins/generic-oauth/index.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/index.ts
@@ -706,6 +706,7 @@ export const genericOAuth = (options: GenericOAuthOptions) => {
 							...tokens,
 							scope: tokens.scopes?.join(","),
 						},
+						callbackURL: callbackURL,
 						disableSignUp:
 							(provider.disableImplicitSignUp && !requestSignUp) ||
 							provider.disableSignUp,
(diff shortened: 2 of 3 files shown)


gee1k:
I also found that there are the same missing parameter issues in two other places.

https://github.com/better-auth/better-auth/blob/f64fcc82880ffc8713e8b250b4748c9b41c1ab87/packages/better-auth/src/plugins/sso/index.ts#L848-L868

https://github.com/better-auth/better-auth/blob/f64fcc82880ffc8713e8b250b4748c9b41c1ab87/packages/better-auth/src/plugins/generic-oauth/index.ts#L698-L713

gee1k:
The issue of missing callbackURL in the verification email after logging in with the genericOAuth and SSO plugin has also been synchronized fixed.

gee1k:
Additionally, I found that the social and oauth login methods all have a `newUserCallbackURL` parameter, which can be used to distinguish from the `callbackURL` parameter when a new user is created.

But when I checked the code, it seems that this parameter is not actually used anywhere. For example, in `handleOAuthUserInfo`, when `dbUser` does not exist, it should first use the `newUserCallbackURL` parameter.

I really want to help, but since I have only recently started using better-auth, I don't have a comprehensive understanding of the project. I hope you can check if there is this issue in the project.


https://github.com/better-auth/better-auth/blob/f64fcc82880ffc8713e8b250b4748c9b41c1ab87/packages/better-auth/src/oauth2/link-account.ts#L122-L170

Bekacru:
> Additionally, I found that the social and oauth login methods all have a `newUserCallbackURL` parameter, which can be used to distinguish from the `callbackURL` parameter when a new user is created.
> 
> But when I checked the code, it seems that this parameter is not actually used anywhere. For example, in `handleOAuthUserInfo`, when `dbUser` does not exist, it should first use the `newUserCallbackURL` parameter.
> 
> I really want to help, but since I have only recently started using better-auth, I don't have a comprehensive understanding of the project. I hope you can check if there is this issue in the project.
> 
> https://github.com/better-auth/better-auth/blob/f64fcc82880ffc8713e8b250b4748c9b41c1ab87/packages/better-auth/src/oauth2/link-account.ts#L122-L170

other than for idToken the newUserCallbackUrl is used when state is generated https://github.com/better-auth/better-auth/blob/69ef90dfd7f86b43defafa30dddaefeb774119f7/packages/better-auth/src/oauth2/state.ts#L25-L26

PR #2756, opened 2025-05-23
fix(oidc): await for getAdditionalUserInfoClaim

NefixEstrada:
The signature of the function is either a sync function or an async function. Without this change, if you use an async function (that in theory should be correct), the response gets returned without the promise being finished.

With this change, if:
  - is an async function, will wait for the promise to finish before returning
  - is a sync function, will create a promise, run the function, and wait for the promise to finish


zackify:
Haha I just made this same pr! Nice

PR #2761, opened 2025-05-23
fix(types): add missing 'nullable' property to FieldAttributeConfig

bytaesu:
code
code

While developing, I extracted BetterAuthOptions into a separate file to make it reusable across different environments (e.g., Better Auth CLI and Cloudflare Wrangler). After doing so, I encountered a type error specifically related to the `nullable` property — other properties were working as expected.

Upon investigation, I found that the `nullable` field was missing from the type definition.
This PR fixes the issue by adding the missing `nullable` property back to the relevant type.


Kinfe123:
There is nothing with the nullable field attribute; it was mistakenly mentioned in the docs, which has been fixed here - #2808

PR #2763, opened 2025-05-23
fix(oidc-provider): remove need for client secret in `authorization_code` flows

### LINKED ISSUE #2655: bug: OIDC client secret required in PKCE flow

### To Reproduce

1. Install the OIDC plugin
2. Register a client
3. Initiate oauth flow with PKCE

The OIDC provider cannot work correctly with public clients as pkce is not supported correctly. A client secret is always required in the token endpoint

1.2.7

```bash
OS: macOS Sonoma 14.3
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

ieedan:
Fixes #2655 

In `authorization_code` flows the `client_secret` should not be required because it wouldn't be a secret anyways. This PR removes the need for a `client_secret` when calling the `/token` endpoint in oidc flows.

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/index.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/index.ts
@@ -463,7 +463,7 @@ export const oidcProvider = (options: OIDCOptions) => {
 					await ctx.context.internalAdapter.deleteVerificationValue(
 						verificationValue.id,
 					);
-					if (!client_id || !client_secret) {
+					if (!client_id) {
 						throw new APIError("UNAUTHORIZED", {
 							error_description: "client_id and client_secret are required",
 							error: "invalid_client",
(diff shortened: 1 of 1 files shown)


ieedan:
> # Bug Report
> Name	Severity	Example test case	Description
> OIDC Provider Authorization Code Flow Client Secret Bypass	Critical	Register a client, initiate authorization code flow, obtain authorization code, and exchange the code for an access token using the correct client_id but an incorrect or missing client_secret.	The patch removes the client secret validation in the authorization code flow, allowing attackers to impersonate any client by providing their client_id, leading to unauthorized access and potential data breaches.
> Comments? [Email us](mailto:support@jazzberry.ai).

I could be mistaken but from what I understand that's just how this flow works. It's why you use PKCE and consent flows.

Bekacru:
thanks for the PR! closed in favor of https://github.com/better-auth/better-auth/pull/3091

PR #2780, opened 2025-05-24
fix: prevent parseState from overriding custom error url

sebastiangollob:
This PR addresses an issue where `parseState` unconditionally sets a default `errorURL` if it's not found in the state object.

This behavior causes redirection to the default error page instead of the one configured via `onAPIError.errorURL`.


sebastiangollob:
Duplicate of #2640

PR #2797, opened 2025-05-26
feat: file-storage plugin (WIP)

zpg6:
Adding a file-storage plugin to tie file uploads to users that upload them. This should be usable for profile images, file sharing, organization logos, and other use cases in your application.

Includes:
-  `StorageProvider` interface so that we can easily add more providers or users can bring their own.
-  `schema` options and ability to add `additionalFields` to the database table
- `canUploadFile` for your custom pre-upload checks of whether the user can upload a file
- `onFileUploaded` callback for your handling of file uploads completed

### DIFF
--- a/packages/file-storage/build.config.ts
+++ b/packages/file-storage/build.config.ts
@@ -1,0 +1,18 @@
+import { defineBuildConfig } from "unbuild";
+
+export default defineBuildConfig({
+	declaration: true,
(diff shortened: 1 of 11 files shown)


zpg6:
Added ability to list files and get a user's storage usage (each with filter options on the query). List files can be paginated

PR #2816, opened 2025-05-28
Fix Entra token refresh

### LINKED ISSUE #2769: EntraID token refresh not working

### To Reproduce

social sign on user with entra ID, attempt to refresh the access token. get bad request 

following the previous steps, i expected the token refresh to behave normally. but it does not. we get 400 bad request error from Microsoft 

1.2.8

```bash
- OS windows 11
- Browser latest chrome canary (not important)
```

Package

```typescript

    provider: 'sqlserver',
  }),
  socialProviders: {
    microsoft: {
      clientId: env.MICROSOFT_ENTRA_ID_ID,
      clientSecret: env.MICROSOFT_ENTRA_ID_SECRET,
      tenantId: env.MICROSOFT_ENTRA_ID_TENANT_ID,
    },
  },
  account: {
    updateAccountOnSignIn: true,
  },
  plugins: [nextCookies()],
  session: {
    cookieCache: {
      enabled: true,
      maxAge: 5 * 60, // Cache duration in seconds
    },
  },
});
```

### Additional context

Looking at the guide here: https://learn.microsoft.com/en-us/graph/auth-v2-user?tabs=http#step-4-use-the-refresh-token-to-renew-an-expired-access-token

and from my previous experiences writing this logic manually, we are missing the scope url parameters. we cannot omit this. This looks like a fairly simple fix here: https://github.com/better-auth/better-auth/blob/43513e98d1973c15dd2285404955e6475145aa49/packages/better-auth/src/social-providers/microsoft-entra-id.ts#L119 

also, the documentation on this is a little obscure. But, in order to get a refresh token, we need to have offline_access in our scopes. Stated here: https://learn.microsoft.com/en-us/graph/auth-v2-user?tabs=http#authorization-request

One other small thing is that if we chose to disable the profile picture, we still are opening and establishing a connection with microsoft at least until we finish getting the headers here: https://github.com/better-auth/better-auth/blob/43513e98d1973c15dd2285404955e6475145aa49/packages/better-auth/src/social-providers/microsoft-entra-id.ts#L87

This operation is cheap (although not free) since we are fully establishing a TLS session only to tare it down. although we dont finish reading the body. there isn't a good reason i see to be doing this.

CarbonNeuron:
This closes #2769 by fixing the token refresh mechanism for Entra ID.

From Microsoft's documentation [here](https://learn.microsoft.com/en-us/graph/auth-v2-user?tabs=http#step-4-use-the-refresh-token-to-renew-an-expired-access-token) we need to include the "scope" parameter when refreshing the access token.

We also need to include the `offline_access` scope when requesting token, which is documented [here](https://learn.microsoft.com/en-us/graph/auth-v2-user?tabs=http#response-body-properties) so I have added that to the default scopes.

### DIFF
--- a/packages/better-auth/src/social-providers/microsoft-entra-id.ts
+++ b/packages/better-auth/src/social-providers/microsoft-entra-id.ts
@@ -120,8 +120,8 @@ export const microsoft = (options: MicrosoftOptions) => {
 			? options.refreshAccessToken
 			: async (refreshToken) => {
 					const scopes = options.disableDefaultScope
-					? []
-					: ["openid", "profile", "email", "User.Read", "offline_access"];
+						? []
+						: ["openid", "profile", "email", "User.Read", "offline_access"];
 					options.scope && scopes.push(...options.scope);
 
 					return refreshAccessToken({
@@ -131,7 +131,7 @@ export const microsoft = (options: MicrosoftOptions) => {
 							clientSecret: options.clientSecret,
 						},
 						extraParams: {
-							scope: scopes.join(' ') //Include the scopes in request to microsoft
+							scope: scopes.join(" "), // Include the scopes in request to microsoft
 						},
 						tokenEndpoint,
 					});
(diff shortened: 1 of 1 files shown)


CarbonNeuron:
I just tested this on my own private project, without specifying any custom auth, and it seems like it works great! Which is not surprising, because I tested this before committing them lol.

PR #2821, opened 2025-05-29
fix: refresh access token on valid refresh token and expiry

Kinfe123:


ping-maxwell (approved):

ping-maxwell:
Actually I just found out a PR is already merged which solves this
https://github.com/better-auth/better-auth/pull/2764

maybe we should close this

PR #2829, opened 2025-05-29
fix: api key model map

Kinfe123:


himself65:
This would be a breaking change. Should consider after discussion

PR #2862, opened 2025-06-01
docs(stripe): document switching plans on a subscription

rgodha24:
fwiw, I think that making the user pass the `subscriptionId` parameter to make sure that they don't double charge their users is a really bad footgun. imo, it makes a lot more sense to default to updating subscriptions when there is already an active subscription.

needs #2864 to be merged before it actually works


rgodha24:
@Bekacru can you take a look at this whenever you get the chance please? I fixed the bug that made upgrading subs impossible in #2864, and it would be great to document how to switch subscriptions! 

also, i think there needs to be a conversation around switching subscriptions being the default, but i think this is a good patch until then

PR #2864, opened 2025-06-01
fix(stripe): inconsistency preventing subscription upgrades

### LINKED ISSUE #2860: bug(stripe): inconsistency between subscription.stripeSubscriptionId and subscription.id

### To Reproduce

here, we use the subscriptionId field on the body as the `subscription.id` field on the database
https://github.com/better-auth/better-auth/blob/96634ceefd4605761eb183ae8ed3529b9eb1db3e/packages/stripe/src/index.ts#L200-L205

here, we use the subscriptionId field on the body as the `subscription.stripeSubscriptionId`
https://github.com/better-auth/better-auth/blob/96634ceefd4605761eb183ae8ed3529b9eb1db3e/packages/stripe/src/index.ts#L252-L264

this prevents us from switching subscriptions to new plans

rgodha24:
fixes #2860. 

an alternative way of doing this that doesnt depend on the `connector: "OR"` pattern that I haven't seen anywhere else in the repo would be updating this line 
https://github.com/better-auth/better-auth/blob/96634ceefd4605761eb183ae8ed3529b9eb1db3e/packages/stripe/src/index.ts#L260
to this: `(subscription) => subscription.id === subscriptionToUpdate?.stripeSubscriptionId,`

### DIFF
--- a/packages/stripe/src/index.ts
+++ b/packages/stripe/src/index.ts
@@ -200,7 +200,18 @@ export const stripe = <O extends StripeOptions>(options: O) => {
 				const subscriptionToUpdate = ctx.body.subscriptionId
 					? await ctx.context.adapter.findOne<Subscription>({
 							model: "subscription",
-							where: [{ field: "id", value: ctx.body.subscriptionId }],
+							where: [
+								{
+									field: "id",
+									value: ctx.body.subscriptionId,
+									connector: "OR",
+								},
+								{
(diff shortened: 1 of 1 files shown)


rgodha24:
fwiw im running this in prod on [brainrot.mov](brainrot.mov) thru the pkg-pr link and it fixed the issue (with the drizzle adapter using postgres)

PR #2865, opened 2025-06-01
chore: removed custom path options

Kinfe123:


CrutchTheClutch:
Why was this removed?

PR #2878, opened 2025-06-02
(feat:oidcProvider): Trusted Client implementation

BadPirate:
- Add trustedClients configuration to OIDCOptions
- Add skipConsent property to Client interface
- Implement unified getClient() function for layered lookup
- Update authorize and token flows to support trusted clients
- Add comprehensive documentation and examples

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/types.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/types.ts
@@ -121,6 +121,11 @@ export interface OIDCOptions {
 		user: User & Record<string, any>,
 		scopes: string[],
 	) => Record<string, any> | Promise<Record<string, any>>;
+	/**
+	 * Trusted clients that are configured directly in the provider options.
+	 * These clients bypass database lookups and can optionally skip consent screens.
+	 */
+	trustedClients?: (Client & { skipConsent?: boolean })[];
 }
 
 export interface AuthorizationQuery {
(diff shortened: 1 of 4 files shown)


packages/better-auth/src/plugins/oidc-provider/types.ts
```diff
@@ -284,6 +289,11 @@ export interface Client {
 	 * Whether the client is disabled or not.
 	 */
 	disabled: boolean;
+	/**
+	 * Whether to skip the consent screen for this client.
+	 * Only applies to trusted clients.
+	 */
+	skipConsent?: boolean;
```
NormalGaussian, line 296:
```suggestion
	skipConsentForTrustedClients?: boolean;
```

Suggestion: use a more explicit name to make it both a less concerning option, and more obvious.

BadPirate: This is a field on interface Client, not an overall toggle.

BadPirate:
The flag could be removed, and we could just fix the type -- Currently documentation says that jwks is incomplete and I'm pretty sure that HS256 doesn't work with jwks anyway... but without the flag this will force oidcProvider to have jwt plugin setup, and might mess with expectations as it would change accepted key types.

packages/better-auth/src/plugins/oidc-provider/types.ts
BadPirate: Though it's important to teach proper consent :)

PR #2887, opened 2025-06-03
feat(stripe): Reuse existing subscription instead of creating one every time

Konixy:
I don't know if it's intentional but every time a user proceed to checkout and then cancel, an "incomplete" subscription is created on the db and never used again. So I updated the plugin to use only one subscription by user. Obviously a user can only have one unique subscription now, but I think it can be nice if in the future a config option is implemented to allow a user to have multiple subscriptions.


bytaesu:
Closing this PR. Please comment if the problem persists. (seems like already resolved)

PR #2895, opened 2025-06-03
fix: oauth proxy between http and https

juliusmarminge:
Shows a POC of an issue with the oauth proxy. We're setting secure headers, even if the proxied url is a localhost without ssl.

This is hacky, probably a much nicer solution exists, as it won't cover custom cookie names etc. The `isSecureContext` check might also not be the best, but this is what I did to make it work. Happy to continue with guidance on proper helpers etc to use as well as adding tests


Bekacru:
Added a support for prefix in the current PR. But I'll check if we can find a better alternative

PR #2897, opened 2025-06-03
fix(docs): replace error with status in elysia example

LuggaPugga:
This PR updates the example code to reflect the deprecation of error in resolve and instead uses status. This aligns with the latest Elysia context structure as seen [here](https://github.com/elysiajs/elysia/blob/7d921a2c90b99a83240b20d01e8673e49e4327fd/src/context.ts#L190).


LuggaPugga:
Also used in the Elysia integration docs
https://elysiajs.com/integrations/better-auth.html#macro

PR #2912, opened 2025-06-05
fix: Use jwt plugin and RSA instead of HS256 for oidcProvider

### LINKED ISSUE #2726: OIDC: Code grant id_token cannot be decoded

### To Reproduce

1. Create a better-auth server that uses the OIDC  and JWT plugin
2. Create a client that uses the Authorization Code Flow
3. Complete the login of a user through login and consent steps, we're using the "openid email profile offline_access" scope
4. Use the POST auth/token endpoint to get the access_token, (refresh_token) and id_token
5. Verify the id_token

The id_token is encoded with the HS256 symmetrical algorithm, this cannot be verified unless you also have the secret key. I expected the RS256 algorithm to be used to encode the token instead, so it can be validated with  keys from the JWKS endpoint.

1.2.8

```bash
- OS: Windows 11 (client)
- OS: Ubuntu / Docker / NodeJS 22
- Browser: Chrome
```

Backend

```typescript
import { drizzleAdapter } from "better-auth/adapters/drizzle"
import db from "../drizzle/postgres.ts"
import { jwt, oidcProvider, openAPI } from "better-auth/plugins"

const BASE_PATH = "/api/auth"
const issuer = process.env.BETTER_AUTH_URL + BASE_PATH

    provider: "pg",
    usePlural: true,
  }),
  emailAndPassword: {
    enabled: true,
  },
  plugins: [
    oidcProvider({
      loginPage: "/dialog/login",
      consentPage: "/dialog/consent",
      metadata: {
        issuer,
      },
      // requirePKCE: true,
    }),
    openAPI(),
    jwt(),
  ],
  trustedOrigins: [
    ...
  ],
  basePath: BASE_PATH,
})
```

### Additional context

We're using a NodeJS server to request the tokens using the [openid-client package](https://github.com/panva/openid-client).

BadPirate:
Current implementation for jwks in oidcProvider is incomplete / broken it ignores jwt plugin and uses HS256, which is not normally a supported algorithm for jwks.

Since changing it to the right algorithm would break any existing implementations, I added a flag to oidcProvider for "useJWTPlugin" if that flag is set to true, and jwt plugin is setup, oidcProvider will use the same key as the one reported in jwks endpoint from jwt generating tokens that correctly validate.

Resolves #2726 

Tested with config:

```js
jwt({
  jwks: {
    keyPairConfig: {
      alg: 'RS256',
      modulusLength: 2048,
    },
  },
  ...
  }),
  oidcProvider({
    useJWTPlugin: true,
    loginPage: '/auth/signin',
    ...
  })
```

Created a jwt for a genericOauth, validated header:

```
{
  "alg": "RS256",
  "kid": "cA1t6gJISVPhQBIoYbNnrIPx3Bo4wNao"
}
```

Verified signature using the `api/auth/jwks` endpoint payload and jwt.io


BadPirate:
The flag could be removed, and we could just fix the type -- Currently documentation says that jwks is incomplete and I'm pretty sure that HS256 doesn't work with jwks anyway... but without the flag this will force oidcProvider to have jwt plugin setup, and might mess with expectations as it would change accepted key types.

dvanmali:
Fixed through #4161, #3219. Same as #3226.

PR #2920, opened 2025-06-05
fix(docs): include required schema object in drizzle adapter example

Akashamba:


Kinfe123:
I don't think this is the right place to put it, and everything related to Drizzle is already documented here: https://www.better-auth.com/docs/adapters/drizzle

Akashamba:
Thank you for reviewing the PR, @Kinfe123. While I agree that the Drizzle adapter is well-documented on the dedicated page you linked, I believe there’s a discoverability issue. Currently, users following the installation guide who choose Drizzle aren’t explicitly directed to that page, which might lead to confusion during setup.

It might be helpful to reference or link the Drizzle adapter page within the adapters section of the installation guide to ensure smoother onboarding for Drizzle users.

Happy to hear your thoughts on this suggestion!

PR #2924, opened 2025-06-05
feat(drizzle-adapter): add customIdGenerator config

tacomanator:
Seems to me like build-in custom adapters should take in more options than these, but in any case... starting here.


ping-maxwell:
Hello, you can generate a custom id in the auth config already by passing `advanced.database.generateId`

https://www.better-auth.com/docs/reference/options#advanced

tacomanator:
Thanks @ping-maxwell. I found [`customIdGenerator`](https://www.better-auth.com/docs/guides/create-a-db-adapter#customidgenerator) in docs first and didn't see there are multiple ways to get it done.

Confirmed the option mentioned works. Closing PR.

Issue #2890, opened 2025-06-03
bug(stripe): Cannot upgrade number of seats


yavens:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Subscribe to a plan
2. Attempt to subscribe to the same plan with a different number of seats

### Current vs. Expected behavior

When attempting to upgrade your existing plan with more seats, a `YOURE_ALREADY_SUBSCRIBED_TO_THIS_PLAN` error is thrown. Instead, your plan should be upgraded to have more seats.

### What version of Better Auth are you using?

1.2.9-beta.8

### Provide environment information

```bash
- MacOS
- chrome
```

### Which area(s) are affected? (Select all that apply)

Package

### Auth config (if applicable)

```typescript

```

PR #2939, opened 2025-06-07
fix(stripe): allow the user to upgrade a plan with an amount of seats

TheYoxy:
Closes: #2890

When trying to upgrade an existing subscription, use the quantity from the request instead of using the default 1 value.

I've also added a check to verify that the seats are not the same, since it tries to change the subscription and I wouldn't be useful to update with the same amount of seats.


Bekacru:
in favor of https://github.com/better-auth/better-auth/pull/2967

PR #2947, opened 2025-06-07
fix: rename `forgetPassword` APIs to `requestPasswordReset`

### LINKED ISSUE #2946: `forgetPassword` is grammatically incorrect
Is there a specific reason the function is named `forgetPassword` instead of `forgotPassword`?

The name `forgetPassword` sounds like it's intended to delete or remove a password, rather than initiate password recovery. This becomes especially confusing when it's used alongside `resetPassword` and `sendResetPassword`, as the naming doesn’t follow a consistent or intuitive pattern. It also affects the professionalism and clarity of the codebase.

Renaming it to `requestPasswordReset` or maybe `forgotPassword` would better reflect its purpose and align with common industry conventions.

Bekacru:
closes #2946

### DIFF
--- a/demo/nextjs/app/(auth)/forget-password/page.tsx
+++ b/demo/nextjs/app/(auth)/forget-password/page.tsx
@@ -29,7 +29,7 @@ export default function Component() {
 		setError("");
 
 		try {
-			const res = await client.forgetPassword({
+			const res = await client.requestPasswordReset({
 				email,
 				redirectTo: "/reset-password",
 			});
--- a/demo/nextjs/app/(auth)/forget-password/page.tsx
+++ b/demo/nextjs/app/(auth)/forget-password/page.tsx
@@ -29,7 +29,7 @@ export default function Component() {
 		setError("");
 
 		try {
-			const res = await client.requestPasswordReset({
+			await client.requestPasswordReset({
 				email,
 				redirectTo: "/reset-password",
 			});
--- a/docs/components/builder/code-tabs/index.tsx
+++ b/docs/components/builder/code-tabs/index.tsx
@@ -22,7 +22,7 @@ export default function CodeTabs() {
 				? `emailAndPassword: {
 		enabled: true,
 ${
-	options.forgetPassword
+	options.requestPasswordReset
 		? `async sendResetPassword(data, request) {
 			// Send an email to the user with a link to reset their password
 		},`
--- a/docs/components/builder/index.tsx
+++ b/docs/components/builder/index.tsx
@@ -372,11 +372,11 @@ export function Builder() {
 												</div>
 												<Switch
 													id="email-provider-forget-password"
-													checked={options.forgetPassword}
+													checked={options.requestPasswordReset}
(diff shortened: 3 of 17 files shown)


iatomic1:
Hi @Bekacru i think this broke things 
trying to use `authClient.requestPasswordReset` now and its not working

```
   │   ├╴  Property 'requestPasswordReset' does not exist on type '{ emailOtp: { sendVerificationOtp: <FetchOptions extends { mode?: RequestMode | undefined; method?: string | undefined; headers?: (HeadersInit & (HeadersInit | CommonHeaders)) | undefined; ... 32 more ...; disableValidation?: boolean | undefined; }>(data_0: Prettify<...>, data_1?: FetchOptions | undefined) => Promis...'. ts (2339) [27, 46]
```
and when i call it anyway in the server logs the page doesn't exist
image

other auth routes are working
BA version: 1.2.8


```typescript
import { emailOTP, openAPI, twoFactor } from "better-auth/plugins";
import { db } from "./db/drizzle";
import { schema } from "./db/schema";
import { bearer, jwt } from "better-auth/plugins";
import { sendEmail } from "./email/send";
import { reverify } from "@better-auth-kit/reverify";

  user: {
    additionalFields: {
      type: {
        type: "string",
        fieldName: "type",
        returned: true,
        required: true,
        defaultValue: "APP",
      },
    },
  },
    provider: "pg",
    schema: schema,
  }),
  account: { accountLinking: { enabled: true } },
  emailAndPassword: {
    enabled: true,
    autoSignIn: true,
    minPasswordLength: 4,
    sendResetPassword: async ({ user, url }) => {
      await sendEmail(user.email, "forget-password", url);
    },
  },

  plugins: [
    openAPI(),
    reverify(),
    emailOTP({
        console.log(`Sending OTP ${otp} to ${email} for ${type}`);
        await sendEmail(email, type, otp);
      },
      expiresIn: 300,
      otpLength: 6,
      disableSignUp: true,
      sendVerificationOnSignUp: true,
    }),
    bearer(),
    jwt({
      jwt: {
        expirationTime: "15m",
      },
    }),
    twoFactor({
      otpOptions: {
        async sendOTP(data) {
          const user = data.user;
          await sendEmail(user.email, "sign-in", data.otp);
        },
        digits: 6,
      },
    }),
    nextCookies(),
  ],
});
```

jcubic:
BTW: this is a breaking change. Do you follow semver?

shiny:
I’m getting an error in my project: Property 'forgetPassword' does not exist. My suggestion is that, as a responsible project, we should mark it as `@deprecated` first, then roll out the change gradually.

jcubic:
Yeah it was really odd. This is big project, with a lot of users, you can't just change a method like this.

Bekacru:
> Yeah it was really odd. This is big project, with a lot of users, you can't just change a method like this.

This was initially marked as deprecated, as you can see in the PR. It was only removed in v1.4, and the change was made about 7 months ago. We loosely follow semever and often we may introduce "breaking" changes in .1 releases

Issue #2703, opened 2025-05-18
[MCP Plugin] Bearer resource_metadata hardcoded


kziemski:
https://github.com/better-auth/better-auth/blob/ec51c4050ce8f58c05bbbd685968798938639b71/packages/better-auth/src/plugins/mcp/index.ts#L882

On the face should point to baseUrl from context.
but Wondering if it should be overiddable in plugin config so that

instead of https://example.com/api/auth/.well-known/oauth-authorization-server
you can set it to https://example.com/.well-known/oauth-authorization-server

make it a developer responsibility?

PR #2971, opened 2025-06-10
fix issue with hardcoded baseURL in withMcpAuth

lazakrisz:
hey,
I came across an issue while integrating the mcp plugin with hono using version 1.2.9. The resource_metadata hardcodes the localhost:3000 host, see the issue at https://github.com/better-auth/better-auth/issues/2703 .

This PR fixes it, although I had to extend the Auth interface with the `options` property with type of `BetterAuthOptions`, but I think that should work since, the documentation mentions passing the `auth` instance in which should have the options property on it. (see docs at https://www.better-auth.com/docs/plugins/mcp#mcp-session-handling)

Let me know if anything else is needed, I'd be happy to adjust! 🙏

### DIFF
--- a/packages/better-auth/src/plugins/mcp/index.ts
+++ b/packages/better-auth/src/plugins/mcp/index.ts
@@ -886,8 +886,7 @@ export const withMcpAuth = <
 		const session = await auth.api.getMcpSession({
 			headers: req.headers,
 		});
-		const wwwAuthenticateValue =
-			`Bearer resource_metadata=${baseURL}/api/auth/.well-known/oauth-authorization-server`;
+		const wwwAuthenticateValue = `Bearer resource_metadata=${baseURL}/api/auth/.well-known/oauth-authorization-server`;
 		if (!session) {
 			return Response.json(
 				{
--- a/packages/better-auth/src/plugins/mcp/index.ts
+++ b/packages/better-auth/src/plugins/mcp/index.ts
@@ -18,10 +18,13 @@ import { generateRandomString } from "../../crypto";
 import { createHash } from "@better-auth/utils/hash";
 import { subtle } from "@better-auth/utils";
 import { SignJWT } from "jose";
-import type { GenericEndpointContext } from "../../types";
+import type { BetterAuthOptions, GenericEndpointContext } from "../../types";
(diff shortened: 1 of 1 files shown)


KrisTemmerman:
bumping this up

K-Mistele:
+1

ping-maxwell (approved):

K-Mistele:
Hi guys, on further reading of [the RFC](https://datatracker.ietf.org/doc/html/rfc9728#name-www-authenticate-response), I do not believe that this implementation is standards-compliant. 

The standard indicates that the `oauth-protected-resource` and `oauth-authorization-server` paths should be defined under the well-known path defined in the [Well-Known Resource URIs Standard (RFC5785)](https://datatracker.ietf.org/doc/html/rfc5785)

From [Section 3](https://datatracker.ietf.org/doc/html/rfc5785#section-3)
> A well-known URI is a URI [[RFC3986](https://datatracker.ietf.org/doc/html/rfc3986)] whose path component begins with the characters "/.well-known/", and whose scheme is "HTTP", "HTTPS", or another scheme that has explicitly been specified to use well-known URIs.

Further clarification from [Appendix B]():
>    4. Why aren't per-directory well-known locations defined? 
> 
>        Allowing every URI path segment to have a well-known location (e.g., "/images/.well-known/") would increase the risks of colliding with a pre-existing URI on a site, and generally these solutions are found not to scale well, because they're too "chatty".

It seems like it would be best to define a handler that can be exported from better-auth and mounted as the route handler at `/.well-known/oauth-authorization-server` and `/.well-known/oauth-protected-resource` 

While this might work in many cases, it risks breaking client compatibility since the MCP specification explicitly refers to these standards as defining valid implementation parameters.

kriszlazar:
Hi @K-Mistele, have you seen this: https://www.better-auth.com/docs/plugins/mcp#oauth-discovery-metadata there is a handler that you can use to expose the discovery endpoint. If that is not what you are looking for could you create an issue, and we can continue the conversation there 😁.

JackatDJL:
@lazakrisz the better-auth docs only show the configuration for the oauth-authorization-server route and not the oauth-protected-resource, how are we supposed to continue there?

PR #2976, opened 2025-06-11
Update tanstack start setup instruction from api route to server route

HimanshuKumarDutt094:
the api route convention is deprecated in tanstack start, they now have createServerFileRoute().methods()({}) for creating server routes


SrikanthKumarC:
Your code is throwing an error, as shown below:
image

Refer to #2984, where I used ServerRoute as the variable name while creating the route method, and it worked.
Not sure if the variable name itself is what's causing the issue, but that approach resolved it in my case.

rasibn:
@HimanshuKumarDutt094 now `createServerFileRoute` is deprecated too. Maybe we can close this?

PR #2990, opened 2025-06-11
feat(stripe): pass context obejct to stripe plugin callbacks

Bekacru:


packages/stripe/src/types.ts
```diff
@@ -1,6 +1,7 @@
 import type { InferOptionSchema, Session, User } from "better-auth";
 import type Stripe from "stripe";
 import type { subscriptions, user } from "./schema";
+import type { GenericEndpointContext } from "../../better-auth/dist/index.cjs";
```
ping-maxwell: is this import path intentional?

ping-maxwell (approved):
Interesting that the hero was updated in this PR too, but the stripe code looks good to me!

ping-maxwell:
I'll fix tests.

PR #2994, opened 2025-06-11
feat: onPasswordReset callback

Kinfe123:


ping-maxwell:
hey I made a [PR](https://github.com/better-auth/better-auth/pull/2973) for this already, I'll just close mine.

It's worth considering that if the user throws an error in the onPasswordReset it can potentially cause a 500 server error if not caught in a try/catch.

Kinfe123:
They can catch the error on callback itself. and it is also do have the same implementation [here](https://github.com/better-auth/better-auth/blob/c78574311bba4fd9deeaa109a8bf5e60c9b01e6e/packages/better-auth/src/api/routes/reset-password.ts#L126C2-L126C3)

SaviruFr:
Can we get this PR merged?

Bekacru:
@Kinfe123 can we add docs for this

SaviruFr:
@ping-maxwell

PR #3013, opened 2025-06-13
Fix scope issues for Google and Microsoft

jafri:
For Microsoft: Allows incremental scope expansion
For Google: Return all granted scopes to current key, so that signIn does not overwrite previously granted scopes in db


arthberman:
+1 face this issue, thanks for the PR! hope it will be merged soon

chrisdadev13:
+1 I face this issue too!!!!

Issue #1587, opened 2025-02-27
Plugin: Passkey - Signin callback doesn't exist


char-ptr:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. setup better auth to use passkeys
2. make signin page
3. attempt to use passkeys to signin and add a callback url

### Current vs. Expected behavior

current: `authClient.signIn.passkey` function does not have a param to accept callback url
expected: `authClient.signIn.passkey` function does has param to accept callback url

### What version of Better Auth are you using?

1.1.21

### Provide environment information

```bash
- OS: Windows 10
- Browser: Firefox
```

### Which area(s) are affected? (Select all that apply)

Documentation, Client

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
import { passkey } from "better-auth/plugins/passkey"

export const auth = betterAuth({
   plugins: [
    passkey(),
  ]
});
// client
import { createAuthClient } from "better-auth/react"
import { passkeyClient } from "better-auth/client/plugins"
const authClient = createAuthClient({
  baseURL: "https://localhost:3000",
  plugins: [
    passkeyClient()
  ]
})
```

### Additional context

In documentation it specifies that there should be a callback field, yet there is no such field:
https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/passkey/client.ts#L27-L32
![Image](https://github.com/user-attachments/assets/73e54f97-ecfd-41e0-856f-76a85bb233ae)
- https://better-auth.vercel.app/docs/plugins/passkey#signin-with-a-passkey

codedusting:
It also doesn't work without registration and registration gives "SESSION_IS_NOT_FRESH" error. I think this passkey feature is broken?

codedusting:
> It also doesn't work without registration and registration gives "SESSION_IS_NOT_FRESH" error. I think this passkey feature is broken?

Is there any way around for the above?

DimitarY:
> > It also doesn't work without registration and registration gives "SESSION_IS_NOT_FRESH" error. I think this passkey feature is broken?
> 
> Is there any way around for the above?

You need to disable the freshAge on the session or you need to refresh the current session

https://www.better-auth.com/docs/concepts/session-management#session-freshness

DimitarY:
I tested the passkey on v1.2.5 and it is still missing the callbackURL as described in docs
https://www.better-auth.com/docs/plugins/passkey#signin-with-a-passkey

[...]

PR #3014, opened 2025-06-13
docs(passkey): Fixed signIn passkey props

ping-maxwell:
callbackURL doesn't exist.


ping-maxwell:
https://github.com/better-auth/better-auth/issues/1587

saturnonearth:
Am I missing something, this still seems to not be present?

PR #3021, opened 2025-06-14
fix: schema generate mismatch

Kinfe123:


packages/better-auth/src/plugins/organization/organization.test.ts
```diff
@@ -18,15 +18,9 @@ describe("organization", async (it) => {
 					membershipLimit: 6,
 					async sendInvitationEmail(data, request) {},
 					schema: {
-						organization: {
+						team: {
```
Bekacru: remove this change

bayromest:
is this pr included in 1.3.7 ? i have that version this is not on yet fixing the bugs

maelp:
I've been sent here by @Dosu but not sure if it's relevant..

PR #3030, opened 2025-06-15
Merge branch 'main' of https://github.com/better-auth/better-auth

RikhiSingh:


Rieranthony:
Thanks for this @RikhiSingh curious to know if the team of better auth have some eye balls on this? This makes the library unusable in production for many many people.

Would love to keep using the lib! Thanks 🙏

@Bekacru maybe?

HenriqueBragaMoreira:
Thanks for the fix @RikhiSingh! 🙌 Really curious if the @better-auth team has any plans to review and merge this soon? This issue completely breaks production setups when frontend and backend are on different domains.

It’s been reported multiple times, and the PR looks solid. Would love to keep using the library without hacks!

@Bekacru maybe you could take a look?

HenriqueBragaMoreira:
@Kinfe123 Can you help us here?

docs/content/docs/concepts/cookies.mdx
```diff
@@ -50,6 +50,70 @@ export const auth = betterAuth({
...
+        defaultCookieAttributes: {
+            secure: true,
+            httpOnly: true,
+            sameSite: "none",  // Allows cross-domain cookie sharing
+            partitioned: true, // New browser standards will mandate this for foreign cookies
+        },
+    },
+    trustedOrigins: [
+        'https://app.mydomain.com',
+        'https://admin.mydomain.com',
```
himself65: note: this example is still cross-subdomain, not cross-domain

himself65 (changes requested):

docs/content/docs/concepts/cookies.mdx
HenriqueBragaMoreira: @himself65 Thanks for the clarification! @RikhiSingh could you apply this update so we can move forward with the approval?

PR #3046, opened 2025-06-16
fix: callback on successfull email verification

### LINKED ISSUE #870: Callback to run after successful emailVerification
N/A

The `sendVerificationEmail()` function is super useful for streamlining the email verification process. That said, I believe there is need for expanding the `emailVerification` object inside `betterAuth()` so that we can tie in more of custom logic to email verification. More specifically, there are certain fields that I would like to execute after the user verifies their email address via database calls. This could look like the following:
```typescript
// auth.ts

  emailVerification: {
    sendVerificationEmail: async ({ user, url, token ), request) => {
      // verification email logic here
    },
    // proposed callback function
    afterVerification: async ({ user, url, token }, request) => {
      // allow for any custom logic here
    },
  }
});
```

I have considered using webhooks to take care of this, and I believe it is the only viable/easy option to handle something like this, but having this natively available in the package would be a much better option in my opinion.

**Additional context**
I run the `sendVerificationEmail()` function with `sendOnSignUp: true`.

Kinfe123:
closes #870

### DIFF
--- a/docs/components/builder/index.tsx
+++ b/docs/components/builder/index.tsx
@@ -28,7 +28,6 @@ import { useAtom } from "jotai";
 import { optionsAtom } from "./store";
 import { useTheme } from "next-themes";
 import { ScrollArea } from "../ui/scroll-area";
-import styles from "./builder.module.css";
 const frameworks = [
 	{
 		title: "Next.js",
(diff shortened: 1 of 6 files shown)


iatomic1:
@Kinfe123 Would this run only once?

Issue #2010, opened 2025-03-26
Create organization on user sign-up


CodeWithShreyans:
I want to create an org whenever a user signs up.

Is it possible to call `auth().api.createOrganization()` inside the hook?

```js
		hooks: {
			after: createAuthMiddleware(async (ctx) => {
				if(ctx.path.startsWith("/sign-up")){
					const newSession = ctx.context.newSession;
					if(newSession){
						// Create org here
					}
				}
			}),
		}
```

AlexanderHott:
this is how I'm doing it, although it could be made much better

```ts
  databaseHooks: {
    user: {
      create: {
        async after(user) {
          // FIXME: make this better
          const slug =
            user.name.toLowerCase().replaceAll(" ", "-") +
            "-" +
            Math.floor(Math.random() * 10_000).toString();
          const org = await auth.api.createOrganization({
            body: {
              name: `${user.name}'s Organization`,
              slug,
              userId: user.id,
            },
          });
          if (org === null) {
            console.error("failed to create organization on account creation");
            return;
          }

          await auth.api.setActiveOrganization({
            body: { organizationId: org.id },
          });
        },
      },
    },
```

CodeWithShreyans:
@Bekacru opinion on this?

matthewlynch:
The above code works for me but I consistently see an error logged when I call `createOrganization`:

```
[APIError] {
  status: 'UNAUTHORIZED',
  body: undefined,
  headers: {},
  statusCode: 401
}
```

<img width="2036" alt="Image" src="https://github.com/user-attachments/assets/4fc59287-5c3c-4714-8ac9-445eb6e23bcb" />

tehnrd:
I would love to get guidance on correctly handling this from better-auth maintainers/experts. I'm building a B2B app, and all users will be part of an organization, so every user that signs up and logs in needs to be associated with an organization.

It may simply be a documentation update to use hooks, as business logic may be required when creating an organization, but if there is a gap in functionality, that would be noteworthy too.

tehnrd:
I didn't want to create an org for every user, so what I did was have some logic to determine if I should prompt the user to create an org after signing in.

https://github.com/better-auth/better-auth/issues/1023#issuecomment-2809527284

k3dom:
> The above code works for me but I consistently see an error logged when I call `createOrganization`:
> 
> ```
> [APIError] {
>   status: 'UNAUTHORIZED',
>   body: undefined,
[...]

PR #3076, opened 2025-06-18
feat: allow create org on signup and set active org on sign in 

Kinfe123:


fahmiidris:
What if user logs out? Most web applications store the last organization and when user logs back in, the user will be back to the previous organization.

ping-maxwell:
closes: https://github.com/better-auth/better-auth/issues/2010

himself65:

To update your PR:
```bash
git checkout your-branch
```

Issue #1495, opened 2025-02-19
Enable multiple client ids for google provider


samducker:
### Is this suited for github?

- [Y] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

Currently you can only add one client ID for google, unlike other providers like clerk and supabase where you can add a comma seperated list. This is so you can use the native sdks on an android device, ios, and web. 

### Describe the solution you'd like

Ability to add multiple client ids for google provider, and also add skip nonce checks as an option.

### Describe alternatives you've considered

Using web only login approach, but then you lack the beautiful native mobile login!

matheusmichels:
+1 to this, I can't get Google sign in working on web + iOS/Android because of this limitation

li-kai:
I believe this issue is still relevant.

A-Boudi:
I think this is still relevant.

maxvaljan:
relevant

PR #3087, opened 2025-06-19
Accept multiple google clientID

FireLord:
This fixes: https://github.com/better-auth/better-auth/issues/1495

### DIFF
--- a/packages/better-auth/src/oauth2/create-authorization-url.ts
+++ b/packages/better-auth/src/oauth2/create-authorization-url.ts
@@ -41,8 +41,9 @@ export async function createAuthorizationURL({
 	scopeJoiner?: string;
 }) {
 	const url = new URL(authorizationEndpoint);
+	const clientId = Array.isArray(options.clientId) ? options.clientId[0] : options.clientId;
 	url.searchParams.set("response_type", responseType || "code");
-	url.searchParams.set("client_id", options.clientId);
+	url.searchParams.set("client_id", clientId);
 	url.searchParams.set("state", state);
 	url.searchParams.set("scope", scopes.join(scopeJoiner || " "));
 	url.searchParams.set("redirect_uri", options.redirectURI || redirectURI);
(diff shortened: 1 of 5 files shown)


FireLord:
While integrating the iOS app and following the OpenAPI flow, I wasn’t able to pass the `idToken` as expected https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/api/routes/sign-in.ts#L75. After inspecting the implementation https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/social-providers/google.ts#L139, I noticed it only handled a single `clientId`.

To fix this, I updated the code to accept **multiple client IDs**. The first one (`clientId[0]`) is used for web (to support `refreshToken` logic), and the rest can be used for iOS, Android, etc. This makes the system more flexible across platforms.

I've tested the following config locally and confirmed that it works as expected:

```ts
socialProviders: {
  google: {
    enabled: true,
    clientId: ["web-client", "ios-client"],
    clientSecret: envConfig.GOOGLE_CLIENT_SECRET as string,
  }
}
```

> I haven’t updated the documentation yet — I’d like to hear your thoughts on the approach first, especially regarding the `refreshToken` handling, which I haven’t fully explored.

packages/better-auth/src/oauth2/create-authorization-url.ts
```diff
@@ -41,8 +41,9 @@ export async function createAuthorizationURL({
 	scopeJoiner?: string;
 }) {
 	const url = new URL(authorizationEndpoint);
+	const clientId = Array.isArray(options.clientId) ? options.clientId[0] : options.clientId;
```
himself65, line 44: what's the reason to only use the clientId[0]?

FireLord: Hello!

I'm using clientId[0] so the existing setup stays unchanged, most projects already have a single web client ID with a secret, and this keeps that flow working without breaking anything. 

The login flow works fine for the mobile client, but since it uses a different ID and doesn’t have a secret, things like idToken and refreshToken aren’t generated for it. I still need to add proper handling for the mobile client.

professorkolik:
@FireLord are you aware that your generated secret is tight to clientId? You cannot use single secret for multiple clients

himself65:
Closing as outdated

PR #3118, opened 2025-06-21
feat(username): support additional signup paths

ping-maxwell:
The invite-app plugin allows the user to sign up upon accepting the invitation. If a dev is also using the username plugin, they expect to be able to pass username during signup process too.


Bekacru:
not sure if this make sense...closing

PR #3125, opened 2025-06-21
feat: add configurable OAuth state management

### LINKED ISSUE #1819: Proposal: Stateless OAuth State Management for Cross-Environment Authentication

When using the OAuth proxy plugin with different environments (e.g., development server proxying to production), authentication fails because:

- The OAuth flow starts in one environment (development), where generateState() stores state data in the development database
- After OAuth provider authentication, the callback is processed in another environment (production), where parseState() attempts to retrieve the state from the production database
- The verification fails because the state exists in a different database

This creates a consistent authentication failure when testing across environments with separate databases.

Implement a stateless state management approach that doesn't rely on database storage:

- Encrypt all OAuth state data using the existing symmetricEncrypt utility
- Pass the encrypted data directly in the state parameter
- On callback, decrypt and validate the state parameter instead of querying the database

- Not using preview databases
- Implementing a Vercel Integration that dynamically adds redirect URIs

### Additional context

Based on my limited understanding of how the package works I think this change could be made by simply updating the `generateState` and `parseState` functions

Here is a quick mock made by AI

```ts
// Modified generateState function
export async function generateState(c: GenericEndpointContext, link?: { email: string; userId: string; }) {
  const callbackURL = c.body?.callbackURL || c.context.options.baseURL;
  if (!callbackURL) {
    throw new APIError("BAD_REQUEST", { message: "callbackURL is required" });
  }
  
  const codeVerifier = generateRandomString(128);
  const stateData = {
    callbackURL,
    codeVerifier,
    errorURL: c.body?.errorCallbackURL,
    newUserURL: c.body?.newUserCallbackURL,
    link,
    expiresAt: Date.now() + 10 * 60 * 1000,
    requestSignUp: c.body?.requestSignUp
  };
  
  // Encrypt the state data instead of storing in DB
  const encryptedState = await symmetricEncrypt({
    key: c.context.secret,
    data: JSON.stringify(stateData)
  });
  
  return {
    state: encodeURIComponent(encryptedState),
    codeVerifier
  };
}

// Modified parseState function
export async function parseState(c: GenericEndpointContext) {
  const encryptedState = decodeURIComponent(c.query.state || c.body.state);
  try {
    const decryptedState = await symmetricDecrypt({
      key: c.context.secret,
      data: encryptedState
    });
    
    const parsedData = z.object({
      callbackURL: z.string(),
      codeVerifier: z.string(),
      errorURL: z.string().optional(),
      newUserURL: z.string().optional(),
      expiresAt: z.number(),
      link: z.object({
        email: z.string(),
        userId: z.string()
      }).optional(),
      requestSignUp: z.boolean().optional()
    }).parse(JSON.parse(decryptedState));
    
    if (!parsedData.errorURL) {
      parsedData.errorURL = `${c.context.baseURL}/error`;
    }
    
    if (parsedData.expiresAt < Date.now()) {
      throw new Error("State expired");
    }
    
    return parsedData;
  } catch (error) {
    c.context.logger.error("Failed to parse state", error);
    throw c.redirect(`${c.context.baseURL}/error?error=please_restart_the_process`);
  }
}
```

shaug:
This PR addresses the cross-environment OAuth state management issue raised in #1819 by introducing a configurable state management system that enables stateless OAuth flows while maintaining full backward compatibility.

## Changes

- Add `oauth.stateManagement` configuration option with `generateState` and `parseState` functions
- Add `StateManagement` interface with comprehensive JSDoc examples
- Extract database-based state logic into `generateVerificationState` and `parseVerificationState` functions
- Modify `generateState` and `parseState` to support custom state management with DB fallback
- Add comprehensive test coverage for both custom and fallback scenarios

## Backward Compatibility

- When no custom state management is configured, behavior is identical to before
- All existing OAuth flows continue to work unchanged
- Database-based state management remains the default

## Usage

```ts
// Enable stateless state management
const auth = betterAuth({
  oauth: {
    stateManagement: {
      generateState: async (ctx, payload) => {
        const encryptedState = await symmetricEncrypt({
          key: ctx.context.secret,
          data: JSON.stringify(payload)
        });
        return encodeURIComponent(encryptedState);
      },
      parseState: async (ctx, state) => {
        try {
          const decryptedState = await symmetricDecrypt({
            key: ctx.context.secret,
            data: decodeURIComponent(state)
          });
          return JSON.parse(decryptedState);
        } catch {
          return undefined; // Fallback to database
        }
      }
    }
  }
});
```

This enables cross-environment OAuth flows where state can be passed between different environments without requiring shared database access (especially when used with the oAuthProxy plugin).

Closes #1819


shaug:
I'm closing this PR, because after trying to use it in practice I don't think it is an adequate solution to #1819 when used with better-auth's oAuthProxy plugin, mostly because the oAuthProxy plugin still requires that the originating service and the proxy service share the same BetterAuth secret config, which is not good security practice, in my opinion.

I believe the library's proxy plugin could be vastly improved. In the above ticket, I've included source for the one we're now using for our project. It is simpler than the 'standard' oAuthProxy plugin, as it doesn't create new endpoints -- it only intercepts the existing OAuth endpoints. It also doesn't require sharing secrets (beyond the provider's client's secret, which is necessary due to the provider's requirements, not BetterAuth's), and doesn't require sharing storage. It intercepts the outbound POST request to the provider and replaces the state parameter with an encrypted payload that includes the original state and the baseURL, and it intercepts the redirect on the proxy server to decrypt that state to rewrite the redirect back to the originating server.

It currently only works with the existing socialProviders config, but with some minor updates it could also support the 'generic oauth' plugin.

PR #3141, opened 2025-06-23
chore: update @types/react to match react 19

bytaesu:
I think it would be better to separate modules like “react” into their own package and declare them as peerDependencies instead of devDependencies. However, that would involve a structural change, so for now, I’m just submitting this PR.


Kinfe123:
yeah. i have noticed that and it is causing dual react problem on some client interaction and i will try to adjust that as peerDeps with some metadata including those structural change. thanks for noticing this.

PR #3144, opened 2025-06-23
fix(mssql-datatypes): "text" datatype is not working with mssql

### LINKED ISSUE #3143: SQL error when exchanging the token (Microsoft Auth + MSSQL + NextJS)

### To Reproduce

1. Create a Microsoft Entry ID app [configured following this recommendations](https://www.better-auth.com/docs/authentication/microsoft)
2. Create a NextJS app
3. Create an MSSQL database
4. Configure your auth.ts file to connect your mssql database [following this recommendations](https://www.better-auth.com/docs/adapters/mssql)
6. Configure the Microsoft social provider
7. Expose the `better-auth/next-js` handler route API [following this recommendations](https://www.better-auth.com/docs/integrations/next)
8. Implement a `signIn.social` call with `microsoft `as a provider using `createAuthClient`
9. Try to sign in with microsoft 
10. You are going to be redirected to Microsoft login
11. Fill the form with your Microsoft credentials
12. You are going to be redirected to the microsoft callback url with a code, and the error will happen here on the server side: `# SERVER_ERROR: [Error: The data types text and nvarchar are incompatible in the equal to operator.] `

I was expecting better-auth to be able to exchange the code with an access_token, but it failed due to a SQL errror that seems to be an incompatibility between the better-auth SQL schema and what the better-auth handle is doing behing the scene during the token exchange, trying to perform some invalid operation in MSSQL.

1.2.9

```bash
- OS: WSL2
- Browser: Google Chrome
```

Backend, Client

```typescript
import { MssqlDialect } from "kysely";
import * as Tarn from "tarn";
import * as Tedious from "tedious";
 
const dialect = new MssqlDialect({
  tarn: {
    ...Tarn,
    options: {
      min: 1,
      max: 10,
    },
  },
  tedious: {
    ...Tedious,
    connectionFactory: () =>
      new Tedious.Connection({
        server: "localhost",
        authentication: {
          type: "default",
          options: {
            userName: "sa",
            password: "StrongP@ssw0rd!",
          },
        },
        options: {
          database: "betterauth_db",
          port: 1433,
          trustServerCertificate: true,
        },
      }),
  },
});
 
  database: {
    dialect,
    type: "mssql",
  },
  emailAndPassword: {
    enabled: true,
  },
  socialProviders: {
    microsoft: {
      clientId: "*******",
      clientSecret: "*******",
      tenantId: "*******",
      prompt: "select_account"
    },
  },
});
```

### Additional context

_No response_

ludoblues:
Fix #3143


ludoblues:
I created a PR that fixed this issue but still had other mssql specific ones when trying to store the access token

ludoblues:
I solved the access_token issue i had.
I am testing with Microsoft Entra ID and the size of their access_token is 2200+ characters.

packages/better-auth/src/db/get-migration.ts
```diff
@@ -180,7 +180,7 @@ export async function getMigrations(config: BetterAuthOptions) {
 						? "varchar(255)"
 						: field.references
 							? "varchar(36)"
-							: "text",
+							: "varchar(4000)",
```
ping-maxwell: Can you change it to use `varchar(max)` instead? This was the recommended alternative after `text` getting deprecation

ping-maxwell: Actually I can do it, dw

AugustasPaliulis:
@ping-maxwell Heyy :)), are there are plans releasing this PR? I don't know the whole process, how it works. But basically this should solve pretty important error for our project, as the client insisted that we use MS SQL

ping-maxwell:
cc @Bekacru

PR #3159, opened 2025-06-24
feat: add password strength plugin

moshetanzer:


moshetanzer:
Hey @ping-maxwell this would be cool but needs a drop of love. Any thoughts? Biggest issue is better auth built in min password strength I think that was being run before this plug-in check ?

sovetski:
@moshetanzer is it normal that this PR is on draft mode?

Also, in addition to these messages `Password must contain at least ${options.minUppercase} uppercase letter(s).` I suggest to add error codes, by this way users can catch the error codes and show their custom error message in the language they want, if not it will be only in english

moshetanzer:
@sovetski left it in draft mode becuase it needs a bit of love as explained above - it always runs only after built in basic password checker so not sure if somehting in core needs to change or if there is a better way to do this. Def do hear your point about error codes - not sure if will have time to do it right now 🙈

himself65:

To update your PR:
```bash
git checkout your-branch
```

Issue #3150, opened 2025-06-23
better-auth is contributing to memory leak with Sentry and Sveltekit


tehnrd:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

This repo has full steps and code to reproduce: https://github.com/tehnrd/sentry-mem-leak

### Current vs. Expected behavior

Memory continues to grow and grow. Memory should not continue to grow.

### What version of Better Auth are you using?

1.2.10

### Provide environment information

```bash
Mac OSX
Node 22 LTS
"@sentry/sveltekit": "9.30.0",
"better-auth": "1.2.10"
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({});
```

### Additional context

The problem appears to be this line: https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/integrations/svelte-kit.ts#L23

After upgrading Sentry to 9.28.1 this starts to cause memory leaks. I do not entirely understand why.

I think the fix is simply moving the following code outside of the svelteKitHandler function as there is no reason this import needs to run on every single request:
```javascript
//@ts-expect-error
	const { building } = await import("$app/environment")
		.catch((e) => {})
		.then((m) => m || {});
```

I will try to open a PR soon, but wanted to log the issue if someone else can jump on this. 

Related sentry issue: https://github.com/getsentry/sentry-javascript/issues/16674#issuecomment-2996188619

tehnrd:
It looks like `import { building } from "$app/environment";` used to be [implemented as a top-level import](https://github.com/better-auth/better-auth/blame/c9644cc9861e4ac09d06029ba8d963fa8082c690/packages/better-auth/src/integrations/svelte-kit.ts), but then it was moved [here](https://github.com/better-auth/better-auth/commit/78978d4a32978c5c952bec924dce9f5a91bb54ee).

tehnrd:
@ping-maxwell @Bekacru Sorry to @ mention, I know how hectic startup life is, but can you take a look at this one? PR is ready to go. There is a nasty memory leak with better-auth with [Sentry](https://github.com/getsentry/sentry-javascript) and the way better-auth is doing dynamic imports in ESM builds for every single http request with SvelteKit.

ping-maxwell:
Hello, can you try installing this and LMK if it solves your issue?
[...]

PR #3161, opened 2025-06-24
fix(svelte): Memory leak caused by dynamic import every req

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/3150

### DIFF
--- a/packages/better-auth/src/integrations/svelte-kit.ts
+++ b/packages/better-auth/src/integrations/svelte-kit.ts
@@ -7,6 +7,8 @@ export const toSvelteKitHandler = (auth: {
 	return (event: { request: Request }) => auth.handler(event.request);
 };
 
+let building: any | undefined = undefined;
+
 export const svelteKitHandler = async ({
 	auth,
 	event,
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/integrations/svelte-kit.ts
```diff
@@ -19,10 +21,13 @@ export const svelteKitHandler = async ({
...
-	//@ts-expect-error
-	const { building } = await import("$app/environment")
-		.catch((e) => {})
-		.then((m) => m || {});
+	if (typeof building === "undefined") {
+		//@ts-expect-error
+		const envModule = await import("$app/environment")
+			.catch((e) => {})
+			.then((m) => m || {});
+		building = resolve.building;
```
tehnrd: @ping-maxwell what is thought here? resolve is a callback function, resolve.building will always be undefined and then this loops as `if (typeof building === "undefined")` is alway true.

tehnrd: This works and eliminates the memory leak, https://github.com/better-auth/better-auth/pull/3152/files#diff-df9cd9043b34567fe39871e30a2ca2358cb33437339ff8b85baceb90ef6aab6aR31

packages/better-auth/src/integrations/svelte-kit.ts
```diff
@@ -7,6 +7,8 @@ export const toSvelteKitHandler = (auth: {
 	return (event: { request: Request }) => auth.handler(event.request);
 };
 
+let building: any | undefined = undefined;
```
tehnrd, line 10: Why any type? it will only ever be boolean or undefined.

PR #3171, opened 2025-06-25
docs: logger option

Kinfe123:


rebasecase:
With plugin context hooks could you enable otel tracing do you think? I haven’t looked deeply enough

ping-maxwell (approved):

Issue #3172, opened 2025-06-25
database mongodbAdapter options.advanced.generateId is deprecated and is the only option checked in the create method


Ngorror:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Set your betterAuth server with the following options:
```js
export const auth = betterAuth({
  advanced: {
    database: {
      generateId: createId,
    },
  /* do not uncomment the following line since it is deprecated
   * @see https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/types/options.ts#L733
   */
  // generateId: createId   
  },
  database: mongodbAdapter(db),
  emailAndPassword: { enabled: true },
});
```
2. Sign up a new user
3. Check the `user` collection for the newly created user - there is no `id` field in it, as is expected to be generated by your function

### Current vs. Expected behavior

the user to  have my custom id

Although the check is only on the `options.advanced.generatedId` 

```js
const hasCustomId = options.advanced?.generateId;
```
see [implementation](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/adapters/mongodb-adapter/mongodb-adapter.ts#L230)

### What version of Better Auth are you using?

1.2.9

### Provide environment information

```bash
- OS: ubuntu (mint)
- node: 24.2.0
- browser: Chrome
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { createId } from '@paralleldrive/cuid2';
import { betterAuth } from 'better-auth';
import { mongodbAdapter } from 'better-auth/adapters/mongodb';

const client = new MongoClient('mongodb://localhost:27017/my-db');

export const db = client.db();

export const auth = betterAuth({
  advanced: {
    database: {
      generateId: createId,
    },
  },
  database: mongodbAdapter(db),
  emailAndPassword: { enabled: true },
});
```

### Additional context

Possible fix:
just check for both `options.advanced.generateId` and `options.advanced.database.generateId` like

```js
	const customIdGen =
		options.advanced?.database?.generateId || options.advanced?.generateId;
```
 see [implementation](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/adapters/mongodb-adapter/mongodb-adapter.ts#L11)

PR #3178, opened 2025-06-26
fix(adapter-mongodb): honor custom generateId in create

tinkerer-shubh:
fixes: https://github.com/better-auth/better-auth/issues/3172


tinkerer-shubh:
would appreciate a review when time permits.

this pr and https://github.com/better-auth/better-auth/pull/3179

cc @Kinfe123

Issue #3138, opened 2025-06-23
[@better-auth/stripe] updatedAt field missing from schema in plugin


joaoseidel:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

A webhook call triggers the following error:

```shell
2025-06-23T06:00:57.826Z ERROR [Better Auth]: Stripe webhook failed. Error: Could not find property pointer for update shape key: 'updatedAt'
```

This happens in better-auth/packages/stripe/src/index.ts at:

```ts
await ctx.context.adapter.update({
  model: "subscription",
  update: {
    cancelAtPeriodEnd: false,
    updatedAt: new Date(), // <------
  },
  where: [
    {
      field: "id",
      value: subscription.id,
    },
  ],
});
```

The updatedAt field is not mapped in either schema.ts or types.ts of the Stripe plugin.

### Current vs. Expected behavior

**Current Behavior**
The plugin attempts to update a field (updatedAt) that is not declared in the schema definition. This results in runtime errors, especially for databases that do not use the default camelCase conversion.

**Expected Behavior**
The updatedAt field should be explicitly defined in the schema.ts and included in types.ts, ensuring compatibility with custom or snake_case naming strategies used in various databases.

### What version of Better Auth are you using?

1.2.10

### Provide environment information

```bash
- OS: Windows 11
- Browser: Chrome
```

### Which area(s) are affected? (Select all that apply)

Package

### Auth config (if applicable)

```typescript

```

PR #3179, opened 2025-06-26
Fix/3138 stripe updatedAt

tinkerer-shubh:
What & Why
Fixes missing timestamp columns that caused `updatedAt` write errors in the Stripe plugin and ensures the field is kept in-sync on every mutation.

Changes
schema.ts – added `createdAt` & `updatedAt` fields to `subscription` model (default `new Date()`).
types.ts – added optional `createdAt` / `updatedAt` props to `Subscription` interface.
index.ts – stamped `updatedAt: new Date()` in the few `adapter.update` paths that previously omitted it (customer creation, cancel-at-period-end, manual cancel fallback, restore).

closes: https://github.com/better-auth/better-auth/issues/3138

### DIFF
--- a/packages/stripe/src/index.ts
+++ b/packages/stripe/src/index.ts
@@ -256,6 +256,7 @@ export const stripe = <O extends StripeOptions>(options: O) => {
 							model: "user",
 							update: {
 								stripeCustomerId: stripeCustomer.id,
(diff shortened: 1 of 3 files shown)


himself65:

To update your PR:
```bash
git checkout your-branch
```

Issue #3133, opened 2025-06-22
[better-auth/vue] authClient.signUp.email() sends invalid JSON payload – causes "[object Object]" parsing error on backend


Jeromearsene:
### Is this suited for github?

- [ ] Yes, this is suited for github

### To Reproduce

Hi 👋

I'm using better-auth/vue and ran into a blocking issue when calling authClient.signUp.email().

## 🐞 Problem
When sending the following request:

```typescript
await authClient.signUp.email({
  email: "test@example.com",
  password: "password1234",
  name: "test",
  image: "https://example.com/image.png",
});
```

The backend (using better-auth/node and toNodeHandler(auth)) throws the following error:

```
SyntaxError: "[object Object]" is not valid JSON
```

## ✅ What works
If I send the exact same data using curl:

```bash
curl -X POST http://localhost:3000/api/auth/sign-up/email \
  -H "Content-Type: application/json" \
  -d '{"email":"test@example.com","password":"password1234","name":"test"}'
```

...everything works fine. This confirms the backend is set up correctly.

## 🔍 Investigation
After intercepting the outgoing request, it turns out the body is passed directly as a plain JS object, not a JSON string. This causes the browser to serialize it as [object Object], which of course is invalid JSON.

## 🛠️ Suggested fix
In the implementation of authClient.signUp.email, the payload should be passed through JSON.stringify(data) before being set as the body in the fetch call.
You must stringify data before.

## ✅ Temporary workaround
I overrode the signUp.email method manually to fix the issue locally, but it would be ideal to have this fixed in the library itself.

Thanks for your work! Let me know if you need a repro or more details. 🙌

## Current

`SyntaxError: "[object Object]" is not valid JSON`

### What version of Better Auth are you using?

1.2.10

### Provide environment information

```bash
- OS: Mac
- Backend: Express (5.1.0) with drizzle-orm (0.44.2) and trpc (11.4.2)
```

### Which area(s) are affected? (Select all that apply)

Client

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  onAPIError: {
    throw: true,
    onError: (error, ctx) => {
      // Custom error handling
      console.error("=== AUTH ERROR DETAILS ===");
      console.error("Error:", error);
      console.error("Error message:", error.message);
      console.error("Error stack:", error.stack);
      console.error("Context:", ctx);
      console.error("=== END ERROR DETAILS ===");
    },
    logger: {
[...]

PR #3160, opened 2025-06-24
Update to-auth-endpoints.ts

Adityakk9031:
This PR addresses issue [#3126](https://github.com/%3Cyour-org%3E/%3Crepo%3E/issues/3126):
"Sign up additionalFields with array type items gets duplicated when using before hooks"

✅ Root Cause:
The defu library is used to merge contexts, and it concatenates arrays by default, leading to duplication in fields like tags.

✅ Fix:
Replaced defu with a custom defuReplaceArrays instance using createDefu, which replaces arrays instead of merging them.

Applied defuReplaceArrays in:

toAuthEndpoints (line 73)

runBeforeHooks (line 157)

🧪 Testing
Manual and unit testing confirmed that:

Arrays in ctx.body such as tags are no longer duplicated.

All other fields are still merged correctly.

📦 Commands used
bash
Copy
Edit
pnpm i
pnpm build
pnpm -F docs dev
Related Issue
Closes #3126

PR #3182, opened 2025-06-26
Fix: JSON body not serialized in authClient.signUp.email (#3133)

Adityakk9031:
This PR fixes a bug where authClient.signUp.email() sends a plain JS object instead of a serialized JSON string in the request body, causing the backend to receive [object Object].

Fix:

Wraps the body in JSON.stringify() before sending via fetch

Ensures proper "Content-Type": "application/json" header is included

Adjusts createDynamicPathProxy to serialize only for non-GET methods

📸 Reproduction
ts
Copy
Edit
await authClient.signUp.email({
  email: "test@example.com",
  password: "password1234",
  name: "test",
});
Before: throws SyntaxError: "[object Object]" is not valid JSON
After: works as expected, backend receives correct payload.

🔬 Test Plan
 Manually tested with backend better-auth/node

 Confirmed body is now valid JSON

 Verified no regression for GET requests

 Tested with $fetch override — behaves identically

📦 Affected Packages
better-auth/vue (client)

🧩 Related Issues
Closes [#3133](https://github.com/your-org/better-auth/issues/3133)

### DIFF
--- a/packages/expo/src/expo.test.ts
+++ b/packages/expo/src/expo.test.ts
@@ -189,7 +189,6 @@ describe("expo with cookieCache", async () => {
 			expires: expect.any(String),
 		});
 	});
-
 	it("should refresh session_data when it expired without erasing session_token", async () => {
 		vi.advanceTimersByTime(1000);
 		const { data } = await client.getSession();


Adityakk9031:
@cubic-dev-ai ay suggestion

Adityakk9031:
@cubic-dev-ai can you please ask @Kinfe123 to look in my pr

ping-maxwell:
This... doesn't look right..

ping-maxwell:
The link to the issue you says it solves doesn't exists, and you shouldn't need to JSON.serialize since should be done already under the hood

Adityakk9031:
@ping-maxwell could you please check this pr Update to-auth-endpoints.ts #3160

Kinfe123:
I have checked the issue. it is working fine. make sure to provide us any repro that will help us to experience what you have faced. closing now. will reopen once we have found that this is actually an issue.

PR #3187, opened 2025-06-27
feat: add support for custom script url in open-api plugin

### LINKED ISSUE #3186: Add support for custom script URL in openAPI plugin

The current implementation of the openAPI plugin uses a hardcoded CDN URL to load the Scalar API Reference script. This makes it impossible to use it in environments where internet access is restricted or self-hosting is required.

This issue proposes adding support for a new optional configuration option: `apiReferenceScriptSrc`. When provided, this will be used instead of the default hard-coded CDN URL, allowing users to self-host the script and still benefit from Scalar.

As far as I understand, there are no alternatives to this in environments without internet access.

### Additional context

A user may want to:
- Serve the Scalar API Reference Script from their own internal server.
- Avoid relying on external CDNs due to security policies or lack of connectivity.

itsyoboieltr:
Fixes #3186 

This PR adds support for customizing the script src used in the OpenAPI reference HTML. This allows users to self-host the API Reference script instead of relying on the default CDN, which is important for restricted or offline environments.

Changes Introduced:
- Introduced a new apiReferenceScriptSrc option in OpenAPIOptions.
- Updated getHTML() to accept and apply apiReferenceScriptSrc.
- Modified the openAPIReference endpoint to respect the apiReferenceScriptSrc value when rendering the HTML.

Tests Added:
- Added a test to verify that the provided apiReferenceScriptSrc appears in the generated HTML.

This change is fully backward-compatible. If no apiReferenceScriptSrc is specified, it defaults to the original CDN.

### DIFF
--- a/packages/better-auth/src/plugins/open-api/index.ts
+++ b/packages/better-auth/src/plugins/open-api/index.ts
@@ -68,9 +68,9 @@ export interface OpenAPIOptions {
 
 export const openAPI = <O extends OpenAPIOptions>(options?: O) => {
 	const path = (options?.path ?? "/reference") as "/reference";
-	const apiReferenceScriptSrc = options?.apiReferenceScriptSrc
-		? new URL(options.apiReferenceScriptSrc).toString()
-		: "https://cdn.jsdelivr.net/npm/@scalar/api-reference";
+	const apiReferenceScriptSrc =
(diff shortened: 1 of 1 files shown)


himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #3196, opened 2025-06-27
feat: Steam social provider

ping-maxwell:
Better-Auth does not currently support Steam as a provider because Steam authentication does not comply with OAuth or OIDC standards. For example, it does not respect state, does not require a client ID, client secret, or application registration with redirect whitelisting, and lacks an OIDC endpoint—instead, you must use your own API key to call ISteamUser/GetPlayerSummaries/v002/ after receiving the callback. Additionally, Steam does not return an email address, as user information is only accessible via a public profile endpoint. Due to these incompatibilities, integrating Steam with a spec-compliant adapter is not feasible, so a custom plugin is required.

Big thanks to the initial implementation from https://gist.github.com/Whats-A-MattR/5bce5574e568e8d8e6be55cf692df3a1 while a lot of the logic is re-written, this was a good head start for me!

## Status: in-development

- [x] docs

# IMPORTANT

I'm closing this PR because of merge conflict issues, was easier to just open a new PR:
https://github.com/better-auth/better-auth/pull/4877


Whats-A-MattR:
nice

D3visionNL:
Damn this sounds like a headache. Been following the thread in the post in the discord about this for a while so I can't wait to see this work!

Whats-A-MattR:
> Damn this sounds like a headache. Been following the thread in the post in the discord about this for a while so I can't wait to see this work!

Yeah, steam auth is non-standard and a bit of an arse to wrangle within standardised confines.

packages/better-auth/src/plugins/steam/index.ts
```diff
@@ -0,0 +1,288 @@
...
+					const profile = profileRes.data.response.players[0];
+
+					if (!profile) {
+						throw ctx.redirect(`${errorURL}?error=steam_profile_not_found`);
+					}
+
+					let account = await ctx.context.internalAdapter.findAccount(steamid);
+					let user: User | null = null;
+					let isNewUser = false;
+					if (!account) {
```
TimFL: No plans to add a way to disable implicit sign ups here?

ping-maxwell: Good callout

himself65 (commented):
We have lots of document CSS changes. Is this expected?

ping-maxwell:
Are you sure? I can see it
image

PR #3206, opened 2025-06-28
feat: add SoundCloud social provider

8times4:
This adds SoundCloud as a social login provider. 

The provider gives full-scope access, as it's not fine-tunable on SC side, but in return the site can support liking tracks/playlists etc via the signed in user's account. 

Note: SC does not return an email address, so a workaround is used for that as `username@soundcloud`

Would appreciate some help on where Better-Auth gets the sidebar logos, as I've omitted them for now. 

https://developers.soundcloud.com/docs/api/guide#authentication

### DIFF
--- a/docs/content/docs/authentication/soundcloud.mdx
+++ b/docs/content/docs/authentication/soundcloud.mdx
@@ -1,0 +1,63 @@
+---
+title: SoundCloud
+description: SoundCloud provider setup and usage.
+---
+
+<Steps>
+    <Step>
+        ### Get your SoundCloud credentials
+        To use SoundCloud sign-in, you need a **Client ID** and a **Client Secret**.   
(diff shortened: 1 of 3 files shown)


packages/core/src/social-providers/soundcloud.ts
```diff
@@ -0,0 +1,96 @@
...
+
+			const userMap = await options.mapProfileToUser?.(profile);
+			return {
+				user: {
+					id: profile.id.toString(),
+					name: profile.full_name || profile.username,
+					// @note SoundCloud does not provide email addresses, so this becomes username@soundcloud
+					email: profile.username + "@soundcloud",
+					image: profile.avatar_url,
+					emailVerified: false,
```
8times4, line 88: I've hardcoded this as `false`, even though SC returns a bool for  `primary_email_confirmed` but we are unable to get the email from any endpoint, therefore I felt like it's useless.

himself65:
Please update the code to use generic oauth format

See https://www.better-auth.com/docs/plugins/generic-oauth

Issue #3235, opened 2025-07-01
Incorrect PriceId after upgrading subscription from monthly to yearly


MikeCodeur:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

When upgrading a subscription from monthly to yearly via Better Auth, the `priceId` returned by the `subscription.list()` API call is not updated and remains the one from the monthly version. This prevents correctly determining the billing mode (yearly/monthly) of the subscription.

## Steps to Reproduce

1. **Create a monthly subscription**
   - Create a new Pro subscription in monthly mode
   - Verify that the `priceId` corresponds to the monthly price (e.g., `price_123_monthly`)

2. **Upgrade to yearly**
   - Use Better Auth's `upgrade` function with `annual: true`
   - Choose the same plan (Pro) but in yearly mode
   - Complete the payment process

3. **Check priceId after upgrade**
   - Retrieve the subscription via `authClient.subscription.list()`
   - Observe that the `priceId` remains the one from the monthly price
   - The Stripe subscription is correctly yearly but Better Auth keeps the old priceId

> 

## Expected Behavior

- The `priceId` in Better Auth should be updated with the new yearly priceId
- The `isYearlyPrice(priceId)` function should return `true` for a yearly subscription
- The interface should correctly display "Yearly" in the summary
- The monthly/yearly toggle should initialize to the correct value

## Actual Behavior

- The `priceId` remains the one from the original monthly price
- `isYearlyPrice(priceId)` returns `false` even for a yearly subscription
- The interface displays "Monthly" instead of "Yearly"
- The toggle initializes incorrectly

## Workaround Implemented

To work around this issue, we have implemented:

1. **Retrieve the real priceId via Stripe**
   ```typescript
   const realPriceId = await getPriceIdFromSubscriptionIdAction(
     activeSubscription.stripeSubscriptionId
   )
   ```

2. **Use the real priceId for business logic**
   ```typescript
   const activeSubscriptionIsYearly = realPriceId ? isYearlyPrice(realPriceId) : false
   ```

## Impact

- **Functionality**: Subscription management works correctly with the workaround
- **Performance**: An additional Stripe API call is required
- **Maintenance**: More complex code to handle data inconsistencies

## Labels

- `bug`
- `better-auth`
- `stripe`
- `subscription`
- `data-sync`

## Notes

[...]

PR #3239, opened 2025-07-01
fix(stripe): update priceId on subscription upgrade and webhook sync

tinkerer-shubh:
related issue: https://github.com/better-auth/better-auth/issues/3235

### 1. File: `packages/stripe/src/hooks.ts`
- **On Checkout Session Completion and Subscription Update:**  
  - The `priceId` is now included when updating the subscription object in the database.  
  - This ensures the correct Stripe price ID is stored whenever a subscription is completed or updated, supporting better synchronization.

### 2. File: `packages/stripe/src/index.ts`
- **Subscription Object Construction:**  
  - When constructing the subscription object, `priceId` is now set to `sub.priceId` if it exists, otherwise it falls back to `plan?.priceId`.
  - Previously, it always took `plan?.priceId`. This makes the assignment more robust by preferring the subscription's own price ID, which is important after upgrades or webhook syncs.

### 3. File: `packages/stripe/src/schema.ts`
- **Subscription Schema Update:**  
  - The `subscriptions` schema now includes `priceId` as a (non-required) string field.
  - This update allows the `priceId` to be stored in the database alongside other subscription details, supporting the changes above.

### DIFF
--- a/packages/stripe/src/hooks.ts
+++ b/packages/stripe/src/hooks.ts
@@ -40,6 +40,7 @@ export async function onCheckoutSessionCompleted(
 						update: {
 							plan: plan.name.toLowerCase(),
 							status: subscription.status,
+							priceId,
 							updatedAt: new Date(),
 							periodStart: new Date(
 								subscription.items.data[0].current_period_start * 1000,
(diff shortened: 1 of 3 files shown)


yihui-zhou:
please fix this

niklashansson:
@yihui-zhou +1

niklashansson:
> please fix this

+1

kubk:
Bump. Would be nice to get this bug fixed

himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #3246, opened 2025-07-02
fix: react dual reference

Kinfe123:
Declaring react as a peerDependency with peerDependenciesMeta.optional avoids bundling it and prevents dual React issues and referenced on mutliple instances.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -6,10 +6,10 @@ packages:
   - demo/*
 catalogs:
   react18:
-    '@types/react': ^18.3.14
-    '@types/react-dom': ^18.3.2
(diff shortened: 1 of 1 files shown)


bytaesu:
https://github.com/better-auth/better-auth/blob/83d5adffb0c232eabeb072f8d31a1ec7735c5874/packages/better-auth/package.json#L691

How about upgrading the `@types/react` version as well, unless there’s a specific reason not to? Because the code has already been written based on v19.

The MUI repository follows the same approach -> [file link](https://github.com/mui/material-ui/blob/35a95be46899fb105ea2cfb87977e5f3b76d299e/packages/mui-material-nextjs/package.json#L44)

PR #3256, opened 2025-07-03
fix(cli): simplify and correct comma insertion logic in plugin array

bytaesu:
---

This PR fixes an edge case bug by simplifying and cleaning up the plugin insertion logic.

---

For example:

```
◇  Would you like to set up plugins?
│  Yes
│
◇  Select your new plugins
│  none
│
◇  It looks like you're using NextJS. Do you want to add the next-cookies plugin? (Recommended)
│  Yes
```

Result:

```ts

    appName: "testing",
    plugins: [, nextCookies()],
});

```


bytaesu:
I’ll rewrite it clean.

PR #3274, opened 2025-07-05
docs(plugin): update schema key to optional for username plugin

H7ioo:
Updated the `username` plugin docs to reflect the current behavior, where username is not required during registration. Also, `session.user.username` can be `string | null | undefined`, so the schema fields should be optional.


himself65:

To update your PR:
```bash
git checkout your-branch
```

Issue #3276, opened 2025-07-06
SCIM Provisioning on SSO Plugin


Whats-A-MattR:
### Is this suited for github?

- [x] Yes, this is suited for github

### Describe the solution you'd like

As SSO implementation is now in Beta, it seems like a good time to introduce SCIM provisioning. 
This is a great feature to support, particularly in B2B SaaS use cases. 

### Describe alternatives you've considered

n/a - new functionality

ntaraujo:
Any updates/plans on this?

gunta:
Definitely looking forward to this

nadilas:
How can we help?

paulgould:
+1

tnkuehne:
https://x.com/imbereket/status/1985780735715029254 seems to be happening

PR #3277, opened 2025-07-06
SCIM Provisioning Enhancement on SSO Plugin

Whats-A-MattR:
Implementing SCIM provisioning on the SSO plugin. This is great functionality to have if Better-Auth looks to support Enterprise deployments (like you'd see with SAML). 

My opinion is that this is really tightly bound to enterprise auth, and it makes sense to extend the SSO Plugin rather than create an additional plugin for this particular feature/enhancement. 

Still very very early, completely untested, just wanted to get it on the table. 
Addresses: #3276


packages/better-auth/src/plugins/sso/index.ts
```diff
@@ -938,6 +980,380 @@ export const sso = (options?: SSOOptions) => {
...
+						schemas: ["urn:ietf:params:scim:api:messages:2.0:ListResponse"],
+						totalResults: users.length,
+						startIndex: 1,
+						itemsPerPage: users.length,
+						Resources: users.map((user) => ({
+							id: user.id,
+							userName: user.email,
+							active: true,
+							schemas: ["urn:ietf:params:scim:schemas:core:2.0:User"],
+						})),
```
ping-maxwell, line 1065: Maybe we can have a config option to configure what to return in the `Resources`?
For example some users may have the `username` plugin, so they could add a `displayName` here.

ping-maxwell (commented):
And maybe an endpoint for `GET /scim/v2/Users/:id` would be nice

Whats-A-MattR:
All good call-outs, will work on those.

packages/better-auth/src/plugins/sso/index.ts
```diff
@@ -938,6 +980,380 @@ export const sso = (options?: SSOOptions) => {
...
+							scimType: "uniqueness",
+						});
+					}
+
+					const user: User = await ctx.context.adapter.create({
+						model: "user",
+						data: {
+							email,
+							name: scim.name?.givenName ?? email,
+							organizationId: provider.organizationId,
```
ping-maxwell, line 1144: I'm pretty sure this shouldn't be in the user table

ping-maxwell:
Skim is kind of a challenge because Better Auth's schemas can be so dynamic from plugins. The createUser endpoint doesn't take into account that other plugins may require additional checks or fields to be part of the user table. We also need all of the schema endpoints (`/scim/v2/Schemas`) as well in order for services to know what fields can come from something like the User data.

ping-maxwell:
I suppose that you can have another config option to configure what else is required for the user creation data.
And then let plugins use hooks or something to run  logic to check if it's valid based on the data to allow user creation

ping-maxwell:
Kind of a challenge, but I suppose this is all we need as a minimum, great contributions though, we really appreciate your work here!

Whats-A-MattR:
> Skim is kind of a challenge because Better Auth's schemas can be so dynamic from plugins. The createUser endpoint doesn't take into account that other plugins may require additional checks or fields to be part of the user table. We also need all of the schema endpoints (`/scim/v2/Schemas`) as well in order for services to know what fields can come from something like the User data.
> 
> 
> 
> 

Yeah I'm just smashing through code at the moment, just to try and get the idea fledged out. 
I did implement '/ServiceProviderConfig' but still need to do ReaourceTypes and Schemas

> Kind of a challenge, but I suppose this is all we need as a minimum, great contributions though, we really appreciate your work here!

My thinking is to basically uncouple the implicit user creation and stick that on an endpoint. 

I'm still pretty surface level in terms of lib knowledge with better-auth, but is there already a method for validating user shape considering the dynamism of better-auth's tables/schemas? 

Also, thank you. I quite like better-auth, happy to contribute wherever I can.

Bekacru:
Hey, could you please re-open the PR? It was closed by accident when the branch got removed

kziemski:
@Whats-A-MattR can you reopen this?

PR #3280, opened 2025-07-06
chore(docs): overriding default styling to match with the current branding

Kinfe123:


ping-maxwell (approved):

Bekacru:
This was intentional as a sharp code block, and the accordion looks a little uncomfortable

PR #3285, opened 2025-07-07
Add Discord OAuth email scope configuration field

Summerlytz:
Collecting the real email address is a privacy concern. By implementing this simple fix, it allows users to replace said address with a dummy address easily. **I understand you can write over the email field, but this removes the SCOPE which is important as the user understands their email is being sent.**

### DIFF
--- a/packages/better-auth/src/social-providers/discord.ts
+++ b/packages/better-auth/src/social-providers/discord.ts
@@ -75,6 +75,8 @@ export interface DiscordProfile extends Record<string, any> {
 
 export interface DiscordOptions extends ProviderOptions<DiscordProfile> {
 	prompt?: "none" | "consent";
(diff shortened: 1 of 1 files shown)


ping-maxwell:
Why not just let the user to pass the scopes to overwrite the `email` scope?

Summerlytz:
Could you elaborate? Right now the email scope is hardcoded and I can't see a method to overwrite. This pr just lets you remove that one scope which is a big concern for a lot of people depending on the use case of better-auth.

ping-maxwell:
Oh email is forced, right of course. My fault!

Summerlytz:
Hi @ping-maxwell, any update on this?

Summerlytz:
@ping-maxwell :)

ping-maxwell:
Hey, I don't merge PRs. It would help if your CI is passing tho :)

Issue #2661, opened 2025-05-15
Line (TW, JP, TH) , Kakao Talk (KR), Naver (KR) provider wished


faithli-ouo:
Hope for add these native provide as default. As some of them have some extra steps for applying oidc.

PR #3286, opened 2025-07-07
feat(social): add Kakao provider

bytaesu:
---

This PR adds support for [Kakao](https://www.kakaocorp.com/) as a social provider.
Kakao is a platform with over 45 million MAU in South Korea, and Kakao Login is used by nearly all local services.
I hope this PR encourages more Korean services to adopt Better Auth.

Related to #2661

---

kakao-login


bytaesu:
-> https://github.com/better-auth/better-auth/pull/3287

PR #3302, opened 2025-07-08
fix: Use twoFactorEnabled flag instead of database lookup for OTP validation

bairdj:
Currently, the OTP part of the 2FA plugin checks that there is an existing entry in the `twoFactor` table before allowing OTP operations to continue.

However, for OTP (not TOTP) there is actually no need for the two factor table, because the generated code is not dependent on the secret. Therefore this call is redundant, and instead the enabled check should use `twoFactorEnabled` as in other parts of the code.

This PR: 
- Changes the check in `/send-otp` to check for `twoFactorEnabled`
- Removes the check altogether in `/verify-otp`, which already has behaviour to update the `twoFactorEnabled` setting to true when a valid code is provided, therefore the check is unnecessary


gleandroj:
This PR broke better auth OTP flow very nice contribution :)

PR #3312, opened 2025-07-08
feat: add custom user schema support

saintparish4:
- Allow options.user.schema to override default user schema
- Maintain backward compatibility with additionalFields
- Update TypeScript inference for custom schemas
- Add comprehensive tests and examples
- Update documentation with schema option
- Create detailed example file with various use cases

### DIFF
--- a/docs/.env.example
+++ b/docs/.env.example
@@ -1,1 +1,0 @@
-NEXT_PUBLIC_URL=http://localhost:3000
--- a/docs/content/docs/reference/options.mdx
+++ b/docs/content/docs/reference/options.mdx
@@ -224,6 +224,8 @@ User configuration options.
 
 ```ts
 import { betterAuth } from "better-auth";
+import { z } from "zod";
+
 export const auth = betterAuth({
 	user: {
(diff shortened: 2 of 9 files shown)


saintparish4:
## 🚀 Add Custom User Schema Support

### Problem
Currently, all custom user fields must be defined using `additionalFields`, which creates several developer experience issues:

- **Poor TypeScript inference**: Custom fields don't integrate naturally with the user type
- **Unintuitive API**: Forces separation of core user properties from their natural schema location
- **Requires pattern knowledge**: Developers must understand the `additionalFields` pattern for basic customization

### Solution
This PR introduces `options.user.schema` to allow complete user schema customization while maintaining full backward compatibility.

### Changes Made

#### 1. **Extended BetterAuthOptions Interface**
```typescript
// Before: Only additionalFields supported
user?: {
  additionalFields?: Record<string, FieldAttribute>;
}

// After: Custom schema support added
user?: {
  additionalFields?: Record<string, FieldAttribute>; // Still works
  schema?: z.ZodObject<any>; // New option
}

PR #3327, opened 2025-07-10
stripe metered billing with v2 metered events

dagmawibabi:


packages/stripe/src/index.ts
```diff
@@ -923,6 +925,365 @@ export const stripe = <O extends StripeOptions>(options: O) => {
...
+						description:
+							"Map a meter event to a customer by passing a customer ID in the event's payload.",
+					}),
+				}),
+			},
+			async (ctx) => {
+				// Check for required fields
+				const { displayName, eventName, aggregationFormula, stripeCustomerId } =
+					ctx.body;
+				if (
```
ping-maxwell: Any reason we need this? The `body` of the endpoint options does the validation for you

packages/stripe/src/index.ts
```diff
@@ -923,6 +925,365 @@ export const stripe = <O extends StripeOptions>(options: O) => {
...
+					}),
+					updatedDisplayName: z.string({
+						description: "The meter's new name. Not visible to the customer.",
+					}),
+				}),
+			},
+			async (ctx) => {
+				// Check for required fields
+				const { meterId, updatedDisplayName } = ctx.body;
+				if (!meterId || !updatedDisplayName) {
```
ping-maxwell: Same as before, do we really need this?

packages/stripe/src/index.ts
```diff
@@ -923,6 +925,365 @@ export const stripe = <O extends StripeOptions>(options: O) => {
...
+				});
+			},
+		),
+
+		listMeteredBilling: createAuthEndpoint(
+			"/metered-billing/list",
+			{
+				method: "GET",
+			},
+			async (ctx) => {
```
ping-maxwell, line 1097: I'm not too familiar with Stripe billing APIs, but would this endpoint mean that anyone can query it and get a list of metered billings?
Sorry if it's a dumb question

dagmawibabi: No, not anyone can get a list of billing meters. Access to Stripe's billing meters is restricted and requires proper authentication. 

    Only authenticated Stripe account holders with the appropriate permissions can retrieve billing meter information.
    You need a valid API key to make requests to the Stripe API, including retrieving billing meter data.
    The API request to list billing meters would typically be made server-side in a secure environment, not from client-side code accessible to the public.

packages/stripe/src/index.ts
dagmawibabi: thank you, resolved these kinds of issues

packages/stripe/src/index.ts
ping-maxwell: I can understand the stripe's client.billing.meters requiring the API key to call, but I mean for the `/metered-billing/list` endpoint. Couldn't someone just call that endpoint and then they will get the data or am I missing something?

dagmawibabi: Okay I got what you meant, thanks, it now checks for session. Let me know if it's something I can improve.

packages/stripe/src/index.ts
```diff
@@ -923,6 +925,328 @@ export const stripe = <O extends StripeOptions>(options: O) => {
...
+				const session = await ctx.context.session;
```
ping-maxwell: I don't mean to be nit-picky, but just a tip for you 👀

I recommend using the session-middleware (see [here](https://github.com/better-auth/better-auth/blob/8bcc35285613fe9ef59f6bf3c93e30c9f5ec93fb/packages/better-auth/src/plugins/organization/routes/crud-members.ts#L499)) this makes it so that the request must have a session otherwise it's rejected automatically.

Alternatively if you want to write it manually I recommend using `getSessionFromCtx` helper function  to get the session rather than `ctx.context.session`

dagmawibabi: I used getSessionFromCtx in the previous commit but for some reason it didn't pass the tests, but will use the session middleware now. 

(always nit pick please whenever u review my code, only way I can improve)

dagmawibabi: @ping-maxwell check current commit and lmk

ping-maxwell: Yeah looks good

himself65:

To update your PR:
```bash
git checkout your-branch
```

Issue #3309, opened 2025-07-08
Add support for organizationSlug in hasPermission API to allow permission checks using slugs


brunowego:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

Currently, the hasPermission API only supports permission checks using the organizationId. However, many applications use organizationSlug as the primary identifier in routes and UI logic, making it more convenient and user-friendly.

We propose adding support for passing organizationSlug directly to the hasPermission API, allowing developers to verify permissions without needing to resolve or store the organizationId first.

### Describe the solution you'd like

```
const { success: hasAccess } = await auth.api.hasPermission({
  headers: await headers(),
  body: {
    permissions,
    organizationSlug: workspace,
  },
});
```

### Describe alternatives you've considered

I’m currently using organizationId by performing an additional query beforehand to resolve the organizationSlug to its corresponding ID. While this works, it introduces extra database or API calls, adds complexity to the code, and slightly increases latency in permission checks.

### Additional context

Benefits

• Simplifies permission checks in apps that use slugs as primary identifiers.
• Reduces the need for extra database lookups or additional API calls to resolve slugs to IDs.
• Improves developer experience and reduces boilerplate.

ceolinwill:
I think this feature is still relevant. AFAIK, we can't use `organizationSlug` with `hasPermission` yet.

PR #3329, opened 2025-07-10
feat(organization): support organizationSlug in hasPermission

tinkerer-shubh:
closes https://github.com/better-auth/better-auth/issues/3309

The pull request introduces the following changes:

1. **In `packages/better-auth/src/plugins/organization/adapter.ts`:**
   - Adds a new method `findMemberByOrgSlug` to fetch a member based on the user's ID and the organization's slug. This method optimizes the lookup by resolving the organization via its slug and then finding the member directly, rather than loading all members.

2. **In `packages/better-auth/src/plugins/organization/organization.ts`:**
   - Updates the API to accept an optional `organizationSlug` in addition to `organizationId`.
   - Refactors the logic in the permission check route to:
     - Prefer using `organizationSlug` if provided, falling back to `organizationId` or the active organization ID.
     - Use the new `findMemberByOrgSlug` method for slug-based lookups.
     - Return a more precise error if the organization is not found or if the user is not a member.
   - Handles both slug-based and ID-based organization member lookups.
   - Ensures the returned member role is used in the permission check.


frectonz:
@tinkerer-shubh can you resolve the merge conflicts and also add tests

packages/better-auth/src/plugins/organization/adapter.ts
```diff
@@ -161,6 +161,54 @@ export const getOrgAdapter = <O extends OrganizationOptions>(
 				},
 			};
 		},
+		// Optimized lookup: fetch member directly by organization slug instead of loading all members
+		findMemberByOrgSlug: async (data: {
```
frectonz, line 165: Instead of doing this we should just use the `findOrganizationBySlug` function to resolve the `slug` into an `organizaionId`.

packages/better-auth/src/plugins/organization/adapter.ts
```diff
@@ -161,6 +161,54 @@ export const getOrgAdapter = <O extends OrganizationOptions>(
 				},
 			};
 		},
+		// Optimized lookup: fetch member directly by organization slug instead of loading all members
```
frectonz, line 164: Also remove these unnecessary comments.

packages/better-auth/src/plugins/organization/adapter.ts
```diff
@@ -161,6 +161,54 @@ export const getOrgAdapter = <O extends OrganizationOptions>(
...
+		findMemberByOrgSlug: async (data: {
+			userId: string;
+			organizationSlug: string;
+		}) => {
+			// First, resolve the organization ID from the slug. We only need the id field.
```
frectonz, line 169: same here

packages/better-auth/src/plugins/organization/adapter.ts
```diff
@@ -161,6 +161,54 @@ export const getOrgAdapter = <O extends OrganizationOptions>(
...
+			const organization = await adapter.findOne<Organization>({
+				model: "organization",
+				where: [
+					{
+						field: "slug",
+						value: data.organizationSlug,
+					},
+				],
+				// We don't need additional fields, but the underlying adapter returns the full row anyway.
```
frectonz, line 178: same

packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -708,14 +710,53 @@ export const organization = <O extends OrganizationOptions>(
 					},
 				},
 				async (ctx) => {
-					const activeOrganizationId =
-						ctx.body.organizationId ||
+					const adapter = getOrgAdapter(ctx.context);
+
+					// Determine the organization identifier (ID from body, slug, or active organization ID)
```
frectonz, line 715: same here this comment is unnecessary

packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -708,14 +710,53 @@ export const organization = <O extends OrganizationOptions>(
...
+								});
+							}
+
+							throw new APIError("UNAUTHORIZED", {
+								message:
+									ORGANIZATION_ERROR_CODES.USER_IS_NOT_A_MEMBER_OF_THE_ORGANIZATION,
+							});
+						}
+
+						memberRole = member.role;
```
frectonz, line 753: Instead of having a custom implementation for this you should just resolve the `slug` into an `organizationId` when no `organizationId` and `activeOrganizationId` is found.

himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #3340, opened 2025-07-10
fix(username): ensure `APIError` code matches `USERNAME_ERROR_CODES`

aleclarson:
The `throw new APIError` statements in the Username plugin do not specify the `code` option, so it's [generated by `better-call`](https://github.com/Bekacru/better-call/blob/b6f1948a1ab08833aef717fd0fac1c3871fda0f3/src/error.ts#L136-L144) from the `message` option. To ensure that [client-side error handling](https://www.better-auth.com/docs/concepts/client#error-codes) can function properly, the values of `USERNAME_ERROR_CODES` must use the same words/spelling as its keys.

### DIFF
--- a/packages/better-auth/src/plugins/username/error-codes.ts
+++ b/packages/better-auth/src/plugins/username/error-codes.ts
@@ -2,8 +2,8 @@ export const USERNAME_ERROR_CODES = {
 	INVALID_USERNAME_OR_PASSWORD: "invalid username or password",
 	EMAIL_NOT_VERIFIED: "email not verified",
 	UNEXPECTED_ERROR: "unexpected error",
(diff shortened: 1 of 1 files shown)


Kinfe123 (commented):
You are trying to compare the value of the error code with the error code itself. thats why the test is failing.

aleclarson:
I've updated this PR to set `code` directly instead of reworking the error messages.

The current failing test appears to be unrelated to my changes.

Bekacru:
Thanks for your contribution! could you open the pr against canary instead of main?

aleclarson:
Looks like `canary` already solved the issue. You can close this or merge as a backported fix.

himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #3341, opened 2025-07-11
fix(oidc): use JWT plugin if enabled to sign keys with tests

kziemski:
This is a draft PR added a test to it in a separate file oidc-jwt.test.ts
in case original pr doesn't update with a test case
---
With this PR, if the JWT plugin is enabled, the OIDC plugin will use it to issue valid JWT tokens, following the published JWKS keys as the metadata endpoint suggests it should.


Bekacru:
in favor of the original PR

kziemski:
ya sry i should have closed it myself

PR #3343, opened 2025-07-11
fix(sso): saml redirection

Kinfe123:
## Summary by cubic
Fixed SAML redirection to use HTTP redirects instead of returning a JSON response, ensuring proper client navigation after SSO login.

- **Bug Fixes**
  - Updated SAML callback to throw an HTTP redirect.
  - Improved SAML test to check for correct redirect behavior.


packages/sso/src/index.ts
```diff
@@ -1328,10 +1328,9 @@ export const sso = (options?: SSOOptions) => {
 					let session: Session =
 						await ctx.context.internalAdapter.createSession(user.id, ctx);
 					await setSessionCookie(ctx, { session, user });
-					return ctx.json({
-						redirect: true,
-						url: RelayState || `${parsedSamlConfig.issuer}/dashboard`,
-					});
+					throw ctx.redirect(
+						RelayState || `${parsedSamlConfig.issuer}/dashboard`,
```
Bekacru: why do we do `/dashboard`? also shouldn't we use the `callbackURL` that was provided on signIn

Kinfe123: The /dashboard was for testing. This was supposed to be a draft PR.

natetewelde:
Just want to say thank you for making this change. Was struggling to understand why my new SAML SSO sign on flow was not working. After looking through all the index code, I noticed this too and was minutes from submitting my own PR before I saw this. Appreciate your work!

packages/sso/src/saml.test.ts
```diff
@@ -709,25 +752,33 @@ describe("SAML SSO", async () => {
...
+		let redirectLocation = "";
+		await betterFetch("http://localhost:8081/api/sso/saml2/callback", {
+			method: "POST",
+			redirect: "manual",
+			headers: {
+				"Content-Type": "application/x-www-form-urlencoded",
+			},
+			body: new URLSearchParams({
+				SAMLResponse: samlResponse.samlResponse,
+				RelayState: "http://localhost:3000",
```
Bekacru: we don't want to hardcode relay state here, if im understanding this correctly. We need to redirect to localhost:3000/dashboard since that's what's provided in the callback url on signIn.sso

natetewelde:
Any ideas as to when this fix will be merged into 1.3? Really need this fix in order to use SSO for the project I’m building.

natetewelde:
@Bekacru which beta release will this be a part of? Just saw the new release didn’t include these changes so just curious.

Issue #3337, opened 2025-07-10
Stripe plugin fails to validate subscription when using lookupKey in price


ow-ph:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

```typescript
//auth.ts

export const auth = betterAuth({
  // some config
  
  plugins: [
    stripe({
       subscription: {
         plans: [
         {
            name: "basic", // the name of the plan, it'll be automatically lower cased when stored in the database
            // priceId: "price_1234567890", // the price ID from stripe
            lookupKey: "default_price",
            // annualDiscountPriceId: "price_1234567890", // (optional) the price ID for annual billing with a discount
            annualDiscountLookupKey: "annual_price",
            limits: {
                projects: 5,
                storage: 10
            }
        }
         ]
       }
    })
  ]
})

```

```typescript
// checkout page

await client.subscription.upgrade({
    plan: "basic",
    successUrl: "/dashboard",
    cancelUrl: "/pricing",
    annual: true, // Optional: upgrade to an annual plan
    referenceId: "org_123" // Optional: defaults to the current logged in user ID
    seats: 5 // Optional: for team plans
});
```

### Current vs. Expected behavior

💥 Current Behavior

The user is redirected to Stripe, where the product and price are displayed correctly. After accepting the payment, the user is redirected back to the application. However, in the database, the subscription status remains "incomplete".

✅ Expected Behavior

After completing the payment, the user should start the trial period as intended. This behavior works correctly when using a priceId directly instead of a lookup_key.

### What version of Better Auth are you using?

1.2.10

### Provide environment information

```bash
macos with chrome, stripe in sandbox
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
export const auth = betterAuth({
  database: new Pool({
    connectionString: process.env.BETTER_AUTH_DATABASE_URL,
  }),
  advanced: {
    database: {
      generateId: false,
    },
  },

  emailAndPassword: {
    enabled: true,
    requireEmailVerification: true,
  },

  emailVerification: {
    autoSignInAfterVerification: true, // Automatically signIn the user after verification
  },

  plugins: [
    emailOTP({
      sendVerificationOnSignUp: true,
      disableSignUp: true,

      async sendVerificationOTP({ email, otp, type }) {
        if (type === 'email-verification') {
          sendOtpConfirmEmail({
            Email: email,
            Code: otp,
          })
        }
[...]

PR #3351, opened 2025-07-12
Stripe: resolve lookupKey → price-ID so subscriptions validate correctly

tinkerer-shubh:
closes https://github.com/better-auth/better-auth/issues/3337

### DIFF
--- a/packages/stripe/src/utils.ts
+++ b/packages/stripe/src/utils.ts
@@ -1,4 +1,45 @@
 import type { StripeOptions } from "./types";
+import type Stripe from "stripe";
+
+const priceIdCache = new Map<string, string>();
+
+const CACHE_MAX_SIZE = 100;
+
+function cacheSet(key: string, value: string) {
+	if (priceIdCache.size >= CACHE_MAX_SIZE) {
+		const oldestKey = priceIdCache.keys().next().value as string | undefined;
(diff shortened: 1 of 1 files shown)


packages/stripe/src/utils.ts
```diff
@@ -1,4 +1,45 @@
 import type { StripeOptions } from "./types";
+import type Stripe from "stripe";
+
+const priceIdCache = new Map<string, string>();
```
Bekacru: I think for the cache we should rely on secondary storage instead

tinkerer-shubh: thanks! addressed in https://github.com/better-auth/better-auth/pull/3351/commits/ffa4396aeec6cf5d1702c1492c3edc2ff198a7cf

packages/stripe/src/types.ts
```diff
@@ -314,6 +319,10 @@ export interface StripeOptions {
 		};
 	};
 	onEvent?: (event: Stripe.Event) => Promise<void>;
+	/**
+	 * Optional secondary storage to cache Stripe priceId lookups and other shared data
+	 */
+	secondaryStorage?: Storage;
```
Bekacru, line 325: It shouldn't take a secondary storage. It should rely on the global secondary storage if provided

PR #2755, opened 2025-05-22
fix(oidc): use JWT plugin if enabled to sign keys

NefixEstrada:
With this PR, if the JWT plugin is enabled, the OIDC plugin will use it to issue valid JWT tokens, following the published JWKS keys as the metadata endpoint suggests it should.

PR #3354, opened 2025-07-12
Fix `getJwtToken` to look for `session` or `newSession`.

### LINKED ISSUE #3353: Error with `getJwtToken` on newSession: TypeError: Cannot read properties of null (reading 'user')
I copied the setup @Bekacru did and ran into the same issue: https://github.com/rocicorp/ztunes/pull/9#issuecomment-3026768995

Code:
```ts
after: createAuthMiddleware(async ctx => {
  if (ctx.path.indexOf('/callback/') !== -1) {
    const headers = must(ctx.context.responseHeaders);
    const session = ctx.context.newSession;
    const token =
      ctx.context.responseHeaders?.get('set-auth-jwt') ||
      (await getJwtToken(ctx));

    if (session && token) {
      setCookies(headers, {
        userid: session.user.id,
        email: session.user.email,
        jwt: token,
      });
    }
    return;
  }

  if (ctx.path.indexOf('/sign-out') !== -1) {
    const headers = must(ctx.context.responseHeaders);
    setCookies(headers, {
      userid: '',
      email: '',
      jwt: '',
    });
    return;
  }
})
```

Error:
```
TypeError: Cannot read properties of null (reading 'user')
    at getJwtToken (file:///Users/izakfilmalter/Projects/FaithBase/openfaith/node_modules/better-auth/dist/plugins/jwt/index.mjs:108:70)
    at async onSome (/Users/izakfilmalter/Projects/FaithBase/openfaith/packages/auth/auth.ts:143:68)
    at async eval (/Users/izakfilmalter/Projects/FaithBase/openfaith/packages/auth/auth.ts:132:7)
```

Source:
https://github.com/better-auth/better-auth/blob/3fed4c0f70202bfadda0a0619016a721dacd7312/packages/better-auth/src/plugins/jwt/index.ts#L179

This issue comes from not having `session` but having `newSession`. I have done the following to work around the issue for now:
```ts
    const session = ctx.context.newSession;
    ctx.context.session = session // Force the newSession as the session so that `getJwtToken` doesn't error.
    const token =
      ctx.context.responseHeaders?.get('set-auth-jwt') ||
      (await getJwtToken(ctx));

```

izakfilmalter:
Resolves #3353

### DIFF
--- a/packages/better-auth/src/plugins/jwt/index.ts
+++ b/packages/better-auth/src/plugins/jwt/index.ts
@@ -175,9 +175,10 @@ export async function getJwtToken(
 		options?.jwks?.keyPairConfig?.alg ?? "EdDSA",
 	);
 
+	const session = ctx.context.session || ctx.context.newSession
 	const payload = !options?.jwt?.definePayload
-		? ctx.context.session!.user
-		: await options?.jwt.definePayload(ctx.context.session!);
+		? session!.user
+		: await options?.jwt.definePayload(session!);
 
 	const jwt = await new SignJWT(payload)
 		.setProtectedHeader({
@@ -189,8 +190,8 @@ export async function getJwtToken(
 		.setAudience(options?.jwt?.audience ?? ctx.context.options.baseURL!)
 		.setExpirationTime(options?.jwt?.expirationTime ?? "15m")
 		.setSubject(
-			(await options?.jwt?.getSubject?.(ctx.context.session!)) ??
-				ctx.context.session!.user.id,
+			(await options?.jwt?.getSubject?.(session!)) ??
+				session!.user.id,
 		)
 		.sign(privateKey);
 	return jwt;


LightTab2:
It might bypass account linking. For example: `anonymous user` is not merged with `newSession`, and the JWT is invalid, because it doesn't contain data from the `newSession`. There might be more cases where we want to have session after it passed other hooks. This leads to potential false positive: we get JWT, but that is JWT we didn't expect. Also this is the default behaviour now, so this does not happen because you changed something.

I'm not saying this solution is bad, but I think it's imperfect (and this problem might be deeper than it seems). Still, I made many assumptions when investigating code, so someone more competent might be able to tell better.

izakfilmalter:
@LightTab2 those are great points, @Bekacru was showing that this is how it should be done here: https://github.com/rocicorp/ztunes/pull/9, but it errors.

saturnonearth:
I can't seem to import `getJwtToken` anymore?

LightTab2:
@saturnonearth
Why do you need it?

According to [the docs](https://www.better-auth.com/docs/plugins/jwt), you should fetch the token. But if you want to embed the token into session, you need this.

It has been moved by #2755 and cannot be imported now I think.

saturnonearth:
> @saturnonearth Why do you need it?
> 
> According to [the docs](https://www.better-auth.com/docs/plugins/jwt), you should fetch the token.

It seems I needed the `bearer()` plugin, it seems to be working now - thanks!

himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #3355, opened 2025-07-12
fix(reset-password): update emailVerified

ping-maxwell:
## Summary by cubic
After a user resets their password, their email is now marked as verified. This ensures users who reset passwords can log in without extra email verification steps.


LightTab2:
There is a niche case, where verification might require additional extra steps (since we can override the function, which sends the email), but password reset not. Shouldn't this be toggleable?

I guess before hook in `authMiddleware` is there for that case, so it's ok, but should at least be noted in docs I think, as this is a potential vurn for the unaware user, especially if they are used to the old way. But then again, they might not read the docs after upgrade, so...

PR #3361, opened 2025-07-13
fix(client): Fix logic around sending FormData

ping-maxwell:
This PR fixes the logic around endpoints which require/use FormData. Primarily needed for the file-storage pluign in the future where something like SignUPEmail would require a FormData. (Since we can't send Blob/File via JSON we must use FormData)

---

## Summary by cubic
When FormData is passed to an authClient API method, the request now uses POST by default instead of GET. This ensures endpoints expecting FormData receive the correct HTTP method.


ping-maxwell:
@cubic-dev-ai test

ping-maxwell:
@cubic-dev-ai are you able to update the PR description?

Issue #3264, opened 2025-07-05
rate limiting not working on magic links


Bruh-Codes:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. create a new nextjs project and install better auth. you can follow https://www.better-auth.com/docs/integrations/next
2. config and add rate limiting https://www.better-auth.com/docs/concepts/rate-limit

### Current vs. Expected behavior

1. on multiple request rate limiting errors are not logged or does not even work.
2. even after updating rate limit storage to database, i dont see any data in db
3. session and everything is working i even toggled the enabled property on the reate limiting for dev environment

### What version of Better Auth are you using?

^1.2.12

### Provide environment information

```bash
- OS [Windows 11]
- Browser [Chrome]
```

### Which area(s) are affected? (Select all that apply)

Client, Documentation, Package

### Auth config (if applicable)

```typescript
import EmailTemplate from "@/components/email-template";
import { betterAuth } from "better-auth";
import { admin } from "better-auth/plugins/admin";
import { magicLink } from "better-auth/plugins/magic-link";
import { Pool } from "pg";
import { Resend } from "resend";

const resend = new Resend(process.env.RESEND_API_KEY);

export const auth = betterAuth({
	rateLimit: {
		enabled: true,
		window: 60, // 1 minute
		max: 3, // Max 3 requests per minute
		storage: "database",
	},
	trustedOrigins: ["http://localhost:3000"],
	socialProviders: {
		// google: {
		// 	clientId: process.env.GOOGLE_CLIENT_ID!,
		// 	clientSecret: process.env.GOOGLE_CLIENT_SECRET!,
		// },
		// twitter: {
		// 	clientId: process.env.TWITTER_CLIENT_ID!,
		// 	clientSecret: process.env.TWITTER_CLIENT_SECRET!,
		// },
	},
	plugins: [
		magicLink({
			async sendMagicLink({ email, url }) {
				console.log(url);
				// const { error } = await resend.emails.send({
				// 	from: "YieldEdu <onboarding@support.yieldedu.xyz>",
				// 	to: [email],
				// 	subject: "Your Magic Sign-In Link for YieldEdu",
				// 	react: EmailTemplate({ link: url }),
				// });
				// if (error) console.log(error);
			},
		}),

		admin({
			adminRoles: ["admin", "partner admin"],
		}),
	],
	/** if no database is provided, the user data will be stored in memory.
	 * Make sure to provide a database to persist user data **/
[...]

PR #3365, opened 2025-07-14
Extend Rate Limiting Docs and add small warning

work-in-progress-danny:
I fell for something like what[ this issue](https://github.com/better-auth/better-auth/issues/3264) is mentioning but my issue was it wasn't clear how rate limiting was working as the docs didn't mention it uses the connecting IP and wasn't throwing any issues and failing silently.

This change is to explicitly mention the use of IP in the rate limiting docs and if it's trying to be used but doesn't find an IP address to `console.warn`. 

I'm using better-auth with cloudflare workers, then the `cf-connecting-ip` is what cloudflare provides with the request. 

Also a small clean up of `createDBStorage` as the param was not being used inside the function body and removing the unnecessary variable assignment `testIP`


StashOfCode:
@work-in-progress-danny, looks like this warning message is getting removed https://github.com/better-auth/better-auth/pull/3691. I was confused by it too

PR #3399, opened 2025-07-16
refactor(perf): zodv4 migration

Kinfe123:
## Summary by cubic
Migrated all packages from Zod v3 to Zod v4 to ensure compatibility and future support.

- **Refactors**
  - Updated all imports and type references to use `zod/v4`.
  - Replaced deprecated or changed Zod APIs and options across the codebase.
  - Adjusted validation schemas and error messages to match Zod v4 requirements.
  - Updated dependencies and lockfiles to use Zod v4.


packages/better-auth/src/api/routes/email-verification.ts
```diff
@@ -192,11 +190,11 @@ export const verifyEmail = createAuthEndpoint(
 		method: "GET",
 		query: z.object({
 			token: z.string({
-				description: "The token to verify the email",
+				error: "The token to verify the email",
 			}),
 			callbackURL: z
 				.string({
-					description: "The URL to redirect to after email verification",
+					error: "The URL to redirect to after email verification",
```
Bekacru: this should be removed altogether. It's not an error message

Bekacru: It should be a js doc comment instead

PR #3426, opened 2025-07-17
fix(organization): allow org owner to update their own roles

### LINKED ISSUE #3010: Owner of organization cannot update their own roles

### To Reproduce

See: https://stackblitz.com/edit/stackblitz-starters-wsvsumaj?file=index.js (`node ./index.js` to start)

1. Use the Organization plugin
2. Create an organization with a custom role
3. Whilst authenticated as the owner, try to call `updateMemberRole` to assign the custom role to the owner

Expected: Owner is allowed to modify their own roles
Actual: Get the error below
```
[APIError: You are not allowed to update this member] {
  status: 'FORBIDDEN',
  body: {
    code: 'YOU_ARE_NOT_ALLOWED_TO_UPDATE_THIS_MEMBER',
    message: 'You are not allowed to update this member'
  },
  headers: {},
  statusCode: 403
}
```

1.2.9

```bash
- OS: Windows 11 Home, v10.0.26100 Build 26100, running WSL2
- WSL2: Linux 6.6.87.1-microsoft-standard-WSL2
- Debian trixie
```

Backend, Package, Client

```typescript
See Stackblitz repro
```

### Additional context

Possibly to related to the fix for #1518

frectonz:
Closes #3010 

Instead of only allowing users with the `member:update` permission to update roles, this PR also checks if the user who is updating the `role` has the `creatorRole` in which case they are allowed to update the `role` of the given user.

This also means that a owner of an org is allowed to remove their own `creatorRole` which may or may not be sth we want.
    

---

## Summary by cubic
Org owners can now update their own roles, including adding or removing the owner role, without needing extra permissions.

- **Bug Fixes**
  - Updated role update checks to allow org owners to change their own roles.
  - Added tests to cover owner role updates and edge cases.

### DIFF
--- a/packages/better-auth/src/plugins/organization/has-permission.ts
+++ b/packages/better-auth/src/plugins/organization/has-permission.ts
@@ -18,13 +18,22 @@ export const hasPermission = (
 	input: {
 		role: string;
 		options: OrganizationOptions;
+		allowCreatorAllPermissions?: boolean;
 	} & PermissionExclusive,
 ) => {
 	if (!input.permissions && !input.permission) {
 		return false;
 	}
+
 	const roles = input.role.split(",");
 	const acRoles = input.options.roles || defaultRoles;
+
+	const creatorRole = input.options.creatorRole || "owner";
+	const isCreator = roles.includes(creatorRole);
+
+	const allowCreatorsAllPermissions = input.allowCreatorAllPermissions || false;
+	if (isCreator && allowCreatorsAllPermissions) return true;
+
 	for (const role of roles) {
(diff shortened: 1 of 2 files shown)


frectonz:
I have moved the logic for, checking to see if the user is a creator and then allowing them the requested permissions, to the `hasPermission` function. This behavior can be activated by calling `hasPermission` with `allowCreatorAllPermissions` set to `true`. For example

```typescript
const canUpdateMember = hasPermission({
	role: member.role,
	options: ctx.context.orgOptions,
	permissions: {
		member: ["update"],
	},
	allowCreatorAllPermissions: true,
});
```

PR #3446, opened 2025-07-18
feat(org): multi team support

frectonz:
## Summary by cubic
Added multi-team support to organizations, including setting an active team, listing user teams, and managing team membership.

- **New Features**
  - Users can set or unset their active team.
  - Added endpoints to list all teams a user belongs to.
  - Team membership is now tracked and updated when accepting invitations.


packages/better-auth/src/plugins/organization/adapter.ts
```diff
@@ -549,6 +551,111 @@ export const getOrgAdapter = (
...
+						field: "userId",
+						value: data.userId,
+					},
+				],
+			});
+
+			return member;
+		},
+
+		findOrCreateTeamMember: async (data: {
```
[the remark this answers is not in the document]
frectonz: The better auth adapter doesn't seem to support upsetting values atomically at this point.

packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -302,6 +332,11 @@ export const organization = <O extends OrganizationOptions>(
 						required: false,
 						fieldName: options?.schema?.session?.fields?.activeOrganizationId,
 					},
+					activeTeamId: {
```
Bekacru: should check if team is enabled before adding the field

packages/better-auth/src/plugins/organization/routes/crud-team.ts
```diff
@@ -449,3 +450,227 @@ export const listOrganizationTeams = createAuthEndpoint(
...
+
+		const member = await adapter.findMemberById(session.user.id);
+		if (!member) {
+			throw new APIError("BAD_REQUEST", {
+				message:
+					ORGANIZATION_ERROR_CODES.USER_IS_NOT_A_MEMBER_OF_THE_ORGANIZATION,
+			});
+		}
+
+		if (!member.teamId) {
```
Bekacru: teamId should be removed now that it's many to many this wouldn't be needed

Bekacru: this should be done from the schema level

Bekacru: and also from the impl. level everywhere

packages/better-auth/src/plugins/organization/routes/crud-invites.ts
```diff
@@ -422,6 +423,23 @@ export const acceptInvitation = createAuthEndpoint(
 					}
 				: {}),
 		});
+
+		if (acceptedI.teamId) {
```
Bekacru: When creating an invite, `teamId` should now support an array of team IDs. When storing them in the database, they can be concatenated into a single string separated by commas. For each team ID in the array, a `teamMember` entry should be created.

packages/better-auth/src/plugins/organization/organization.ts
frectonz: I have fixed this issue. Now `activeTeamId` will not set if `team` is not enabled.

packages/better-auth/src/plugins/organization/routes/crud-team.ts
frectonz: I have fixed this issue. I have removed the `teamId` field on the `member` object.

packages/better-auth/src/plugins/organization/routes/crud-invites.ts
frectonz: `teamId` now supports being give a list of team IDs of the invited user to join.

frectonz:
I have fixed all the merge conflicts and the failing typechecks and test.

frectonz:
@cubic-dev-ai update the summary

frectonz:
These are the functions I have added in this PR.

1. `setActiveTeam`
2. `listUserTeams`
3. `listTeamMembers`
4. `addTeamMember`
5. `removeTeamMember`

PR #3449, opened 2025-07-18
refactor: Pass both building and getRequestEvent from outside of library

RobiMez:
- Removes the need for importing from virtual modules at the expense of more user boilerplate
    

---

## Summary by cubic
Refactored the SvelteKit integration to require passing both the building flag and getRequestEvent function from outside the library, removing internal imports from virtual modules.

- **Refactors**
  - Users now provide building and getRequestEvent, reducing reliance on SvelteKit virtual imports.


Bekacru:
@RobiMez can you add/update the docs

RobiMez:
> @RobiMez can you add/update the docs

yup , on it

PR #3458, opened 2025-07-19
fix(oidc): upgrade the oidc to be oauth 2.1 compliant

dvanmali:
**Introduces the following improvements:**

## OIDC

- OAuth 2.1 by default
    - Properly supports authorization_code, refresh_token, and client_credentials grants
    - PKCE by default (removes `plain` completely)
- JWT verifiable by JWKS through JWT Plugin
- Consolidation of Refresh Token onto Session
- `oauthAccessToken` strictly deals with opaque tokens
    - Opaque tokens are given only when resource parameter (aka audience) is not provided
- Option to Encode and Decode refresh tokens
- `allowDynamicClientRegistration` with `allowUnauthenticatedClientRegistration` flags
- Separation of default expiration times
- Proper creation of public and confidential clients
- Prevents misconfiguration between `.well-known/openid-configuration` endpoint and plugin settings
- Tests Files: 7, Tests: 54

## JWT

- Remote signing
- Remote JWKS
- Compatibility with OIDC (removes /token endpoint)
- Tests: 14

## MCP

- Separates MCP logic from authentication logic. OIDC package handles the Auth 2.1 server (ie OIDC just uses "openid" scope)
- `checkMcp` function to support more platforms beyond react
- `oAuthDiscoveryMetadata` endpoint is configurable per endpoint and is hosted by the resource server instead of the authorization server.
- Tests: 9


dvanmali:
Will break this PR into smaller chunks to manage

XavierGeerinck:
This is great! +1 on client_credentials as it's needed for proper service-to-service authentication

Issue #3226, opened 2025-06-30
Invalid keyPairConfig options available to user


LightTab2:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Pass
```js
 {
	alg: "EdDSA",
	crv: "Ed448",
}
```
or
```js
 {
	alg: "ECDH-ES",
	crv: any of ["P-256" | "P-384" | "P-521"],
}
```

to
```js
plugins: [
	jwt({
		jwks: {
			keyPairConfig: {
				...
			},
		}),
	],
```

### Expected:
These options should not be allowed
### Current:
```js
 {
	alg: "EdDSA",
	crv: "Ed448",
}
```
uses "**Ed25519**" instead.

```js
 {
	alg: "ECDH-ES",
	crv: any of ["P-256" | "P-384" | "P-521"],
}
```
cannot be used to sign tokens.

### What version of Better Auth are you using?

1.2.12

### Provide environment information

```bash
Irrelevant
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript

```

### Additional context

According to JOSE [Algorithm Key Requirements](https://github.com/panva/jose/issues/210) **EdDSA** algorithm needs to be used with **Ed25519** curve.

These bugs were found while writing tests for PR #3219 solving another issue #3218 , related comment:
> Tests concerning 'ECDH-ES' are also failing.
>
> I inspected the **private keys** and they have no 'sing' usage assigned to them:
```js
CryptoKey {
  type: 'private',
  extractable: false,
  algorithm: { name: 'ECDH', namedCurve: 'P-256' },
  usages: [ 'deriveBits' ]
}
```
> example of another key:
```js
CryptoKey {
  type: 'private',
  extractable: false,
  algorithm: {
    name: 'RSASSA-PKCS1-v1_5',
    modulusLength: 2048,
    publicExponent: [Uint8Array],
    hash: [Object]
  },
  usages: [ 'sign' ]
}
```
> This one has that field.
> 
> It is apparently pre-Diffie-Hellman, we would need to use deriveBits to create a common secret and then an actual key for signing.
> I will drop test support for this case (not sure if that's a valid move eww)

_Originally posted by @LightTab2 in https://github.com/better-auth/better-auth/issues/3219#issuecomment-3017310172_

LightTab2:
It was fixed by #4161

PR #3464, opened 2025-07-19
feat: remote jwks, remote signing, and exportable signJwt function

dvanmali:
**feat**: remoteUrl disablesjwks endpoint and uses this endpoint in discovery mode

**feat**: remote sign payloads to such as Google KMS, AWS KMS, and Azure Key Vault

**feat** `usesOauthProvider` adapts the plugin for use with OIDC and MCP

**fix**: signJwt now accepts payload as a parameter

**chore**: combine shared createJwks functionality

__Partial #3458__

### DIFF
--- a/packages/better-auth/src/plugins/jwt/index.ts
+++ b/packages/better-auth/src/plugins/jwt/index.ts
@@ -1,207 +1,319 @@
 import type {
+	AuthContext,
 	BetterAuthPlugin,
+	HookEndpointContext,
 	InferOptionSchema,
 	Session,
 	User,
 } from "../../types";
-import { type Jwk, schema } from "./schema";
+import { schema } from "./schema";
 import { getJwksAdapter } from "./adapter";
-import { getJwtToken } from "./sign";
-import { exportJWK, generateKeyPair } from "jose";
+import { createJwk, signJwt } from "./sign";
+import type { JWTPayload } from "jose";
 import {
 	createAuthEndpoint,
 	createAuthMiddleware,
 	sessionMiddleware,
 } from "../../api";
-import { symmetricEncrypt } from "../../crypto";
 import { mergeSchema } from "../../db/schema";
+import { BetterAuthError } from "../../error";
+import type { Awaitable } from "../../types/helper";
 
-type JWKOptions =
+// Asymmetric (JWS) Supported (https://github.com/panva/jose/issues/210)
+export type JWKOptions =
 	| {
-			alg: "EdDSA"; // EdDSA with either Ed25519 or Ed448 curve
-			crv?: "Ed25519" | "Ed448";
+			alg: "EdDSA"; // EdDSA with Ed25519 key
+			crv?: "Ed25519";
 	  }
 	| {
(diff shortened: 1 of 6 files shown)


LightTab2 (approved):
This looks very good!
Especially getting `client.$fetch` to work

LightTab2:
Fixes #3226

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -53,23 +59,79 @@ export async function getJwtToken(
...
+		.setIssuedAt(payload.iat)
+		.setIssuer(
+			payload.iss ?? options?.jwt?.issuer ?? ctx.context.options.baseURL!,
+		)
+		.setAudience(
+			payload.aud ?? options?.jwt?.audience ?? ctx.context.options.baseURL!,
 		)
-		.sign(privateKey);
-	return jwt;
+		.setAudience(options?.jwt?.audience ?? ctx.context.options.baseURL!)
```
LightTab2: `setAudience` called twice, this is probably unintended

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -1,40 +1,46 @@
...
 
-export async function getJwtToken(
+/**
+ * Signs a payload in jwt format
+ *
+ * @param ctx - endpoint context
+ * @param payload - payload to sign
+ * @param options - Jwt signing options. If not provided, uses the jwtPlugin options
+ */
+export async function signJwt(
```
LightTab2, line 23: We should still have `getJwtToken` function for backwards compability

LightTab2: It used to be exported in `index.ts` and probably will be again if #3555 is merged

LightTab2 (approved):

packages/better-auth/src/plugins/jwt/sign.ts
LightTab2: Fixed

packages/better-auth/src/plugins/jwt/sign.ts
LightTab2: Added

packages/better-auth/src/plugins/jwt/index.ts
```diff
@@ -161,6 +162,25 @@ export const getJwtPlugin = (ctx: AuthContext) => {
...
+ * @deprecated - prefer signJwt
+ *
+ * @param ctx - endpoint context
+ * @param options - Jwt signing options. If not provided, uses the jwtPlugin options
+ */
+export async function getJwtToken(
+	ctx: GenericEndpointContext,
+	options?: JwtPluginOptions,
+) {
+	const payload = !options?.jwt?.definePayload
```
LightTab2: Not checking if options are valid and falling back to:

```js
if (!options) {
		options = getJwtPlugin(ctx.context).options;
	}
```
if necessary

dvanmali:
Gonna close this PR and move to another since build step is failing even though it passes on LightTab2's version

Issue #3218, opened 2025-06-29
Cannot generate JWKS with other algorithm than the default one


LightTab2:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Clone my [minimal example repository](https://github.com/LightTab2/better-auth-jwtbug-minimalexample)
2. `pnpm -i`
3. `pnpm dev`
4. Try to access `http://localhost:3000/api/auth/jwks`

## Expected:
JSON response that would look something like this:
```json
{"keys":[{"kty":"EC","x":"KEY_x","y":"KEY_y","crv":"P-256","alg":"ES256"}]}
```

## Current:
```
SERVER_ERROR:  TypeError: non-extractable CryptoKey cannot be exported as a JWK
    at keyToJWK (D:\better-auth-jwtbug-minimalexample\node_modules\jose\dist\webapi\lib\key_to_jwk.js:23:15)
    at Object.exportJWK (D:\better-auth-jwtbug-minimalexample\node_modules\jose\dist\webapi\key\export.js:10:12)
    at D:\better-auth-jwtbug-minimalexample\node_modules\better-auth\dist\plugins\jwt\index.cjs:214:46
    at async internalHandler (D:\better-auth-jwtbug-minimalexample\node_modules\better-call\src\endpoint.ts:332:20)
    at async api.<computed> (D:\better-auth-jwtbug-minimalexample\node_modules\better-auth\dist\api\index.cjs:520:22)
    at async processRequest (D:\better-auth-jwtbug-minimalexample\node_modules\better-call\src\router.ts:179:22)
    at async handler (D:\better-auth-jwtbug-minimalexample\node_modules\better-call\src\router.ts:200:16)
    at async <anonymous> (D:\better-auth-jwtbug-minimalexample\node_modules\better-call\src\adapters\node\index.ts:11:20)
```

### What version of Better Auth are you using?

1.2.12

### Provide environment information

```bash
- OS: Windows 10 IoT Enterprise LTSC 21H2 build 19044.5965
- Node: v24.3.0
- NPM:  11.4.2
- PNPM: 10.12.4
First encountered when working with TanStack Start.
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
import { jwt } from "better-auth/plugins";

export const auth = betterAuth({
  plugins: 
    [
        jwt(
        {
            jwt: 
            {
                expirationTime: '5min',
[...]

PR #3471, opened 2025-07-19
fix(jwt): Reimplement fix that allows to generate JWKS with other algorithm than the default one

LightTab2:
I'm afraid the test cases I've made in previous PR were insufficient. PR #2755 reintroduced issue #3218.

This is because tests checked internals, but didn't test the function used in the **jwt plugin endpoint**.

I've added test case for this and reimplemented the fix. Hope this never occurs again :)
    

---

## Summary by cubic
Fixed JWKS generation in the JWT plugin to support custom algorithms, not just the default. Added tests to ensure the endpoint works with different algorithms.

- **Bug Fixes**
  - Reimplemented key generation logic to respect custom algorithm settings.
  - Added endpoint-level tests to prevent regression.

### DIFF
--- a/packages/better-auth/src/plugins/jwt/jwt.test.ts
+++ b/packages/better-auth/src/plugins/jwt/jwt.test.ts
@@ -186,44 +186,7 @@ describe("jwt", async (it) => {
 				alg: "RS256",
 			},
 		},
-		// We cannot sign using key exchange protocol, need to establish a key first (only allowed usage for these keys is `deriveBits`)
-		/*
-		{
-			keyPairConfig: {
-				alg: "ECDH-ES",
-				crv: "P-256",
-			},
-			expectedOutcome: {
-				ec: "EC",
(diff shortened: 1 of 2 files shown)


LightTab2:
This is awesome, CI fails at a test case of code I didn't touch nor does reference the code I changed.

LightTab2:
There's apparently a race condition in testing of `social-providers`

LightTab2:
Also I tried to do `client.$fetch` for more elegant test case, but it fails with `"UNAUTHORIZED"` response (why?) and after some fighting I gave up. Split the function in `sign.ts` into two and tested that.

This is bad, because bug can still be possibly reintroduced in future. Current solution only strengthens testing against it.

But shouldn't `client.token()` call the `/token` endpoint's method? If that's true, tests shouldn't pass for #2755

I'm too bad at this and I don't know really what's the core issue.

Issue #3461, opened 2025-07-19
Invalid Token for every reset password request


JuanPabloGilA:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

With this repository you can reproduce it.

https://github.com/LoannPowell/hono-react-boilerplate``

### Current vs. Expected behavior

Requesting token:

```
Auth route accessed: POST /api/auth/forget-password
URL: http://localhost:3000/api/auth/reset-password/jLVNWxorayJxBDHCa2iQnKAv?callbackURL=%2Freset-password
Token: jLVNWxorayJxBDHCa2iQnKAv
user myemail@gmail.com
--> POST /api/auth/forget-password 200 779ms
```

Sending token with password:

```
<-- POST /api/auth/reset-password
Auth route accessed: POST /api/auth/reset-password
--> POST /api/auth/reset-password 400 311ms
```

In the database:

```

, {
    id: "4WSXCjrRuzU0GD2FACsnvNVorv1uh9L7",
    identifier: "reset-password:5NiJmwlsBv9stz936obLVfVg",
    value: "WsHGoub6GjEvqh0rTyGHNbP1sBvjV7Zv",
    expiresAt: 2025-07-19T07:05:48.000Z,
    createdAt: 2025-07-19T06:05:48.000Z,
    updatedAt: 2025-07-19T06:05:48.000Z,
  }, {
    id: "85Dn9APZo8q4DhPpUHfzWzgcjb3wrV0S",
    identifier: "reset-password:jLVNWxorayJxBDHCa2iQnKAv",
    value: "WsHGoub6GjEvqh0rTyGHNbP1sBvjV7Zv",
    expiresAt: 2025-07-19T07:05:57.000Z,
    createdAt: 2025-07-19T06:05:57.000Z,
    updatedAt: 2025-07-19T06:05:57.000Z,
  }
]
[
  {
    id: "WsHGoub6GjEvqh0rTyGHNbP1sBvjV7Zv",
    name: "juan",
    email: "myemail@gmail.com",
    emailVerified: true,
    image: null,
    createdAt: 2025-07-17T00:51:03.000Z,
    updatedAt: 2025-07-17T00:51:03.000Z,
  }
]


```
### What version of Better Auth are you using?

1.2.12

### Provide environment information

```bash
- Os [Windows 11]
= Browser Chrome
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
[...]

PR #3477, opened 2025-07-19
fix(drizzle): use timestamp with time zone for PostgreSQL

LoannPowell:
### What does this PR do?

Fixes a bug where PostgreSQL schemas were being generated with `timestamp` (without timezone), causing expiration checks to break due to UTC mismatches.

### Why is it needed?

BetterAuth tokens like `reset-password` or `verification-token` were expiring immediately if the server and database timezones differed. This fix ensures that expiration date fields in Drizzle schemas for PostgreSQL use `timestamp with time zone`, making comparisons consistent regardless of environment.

### Changes

- Updated Drizzle schema generator to use `timestamp(..., { withTimezone: true })` for PostgreSQL when field type is `date`.
- Updated snapshot files for Drizzle and Prisma.
- Verified all test snapshots reflect the correct schema output.

### Notes

- Untracked temporary folder `packages/cli/getConfig_test-*` was ignored locally and not committed.
- No changes were needed to the test definitions—only snapshots were updated.

Closes https://github.com/better-auth/better-auth/issues/3461

### DIFF
--- a/packages/better-auth/src/adapters/prisma-adapter/test/state.txt
+++ b/packages/better-auth/src/adapters/prisma-adapter/test/state.txt
@@ -1,1 +1,1 @@
-IDLE
+RUNNING
--- a/packages/cli/getConfig_test-Oujd4M/test/server/auth/auth.ts
+++ b/packages/cli/getConfig_test-Oujd4M/test/server/auth/auth.ts
@@ -1,0 +1,12 @@
+import {betterAuth} from "better-auth";
+			 import {prismaAdapter} from "better-auth/adapters/prisma";			
+			 import {db} from "../db/db";
+
+			 export const auth = betterAuth({
+					database: prismaAdapter(db, {
+							provider: 'sqlite'
+					}),
+					emailAndPassword: {
+						enabled: true,
+					}
+			 })
--- a/packages/cli/src/generators/drizzle.ts
+++ b/packages/cli/src/generators/drizzle.ts
@@ -100,7 +100,7 @@ export const generateDrizzleSchema: SchemaGenerator = async ({
 				},
 				date: {
 					sqlite: `integer('${name}', { mode: 'timestamp' })`,
-					pg: `timestamp('${name}')`,
+					pg: `timestamp('${name}', { withTimezone: true })`,
 					mysql: `timestamp('${name}')`,
 				},
 				"number[]": {
--- a/packages/cli/test/__snapshots__/auth-schema-number-id.txt
+++ b/packages/cli/test/__snapshots__/auth-schema-number-id.txt
@@ -15,10 +15,10 @@ export const user = pgTable("user", {
     .$defaultFn(() => false)
     .notNull(),
   image: text("image"),
-  createdAt: timestamp("created_at")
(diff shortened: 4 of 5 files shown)


pranavmappoli:
facing the same issue, Are we gonna merge this ?

JuanPabloGilA:
> facing the same issue, Are we gonna merge this ?

Since this is an issue with the generator, you can just add to your dates in the drizzle schema an additional parameter to every date. So it will be like this:

`timestamp('nameofthetable', { withTimeZone:  true })`

And then push to your database, with the drizzle command.

pranavmappoli:
@LoannPowell so this will store user timezone or server timezone ?

pranavmappoli:
For me even after add timezone doesnt works,
Are they comparing the timezone as well when the user go through the link ?

JuanPabloGilA:
> For me even after add timezone doesnt works,
> 
> Are they comparing the timezone as well when the user go through the link ?

You should open an issue or in discord with your whole setup, so you can get help.

packages/better-auth/src/adapters/prisma-adapter/test/state.txt
```diff
@@ -1 +1 @@
-IDLE
\ No newline at end of file
+RUNNING
```
frectonz: This shouldn't be part of the commit, it should be removed.

packages/cli/getConfig_test-Oujd4M/test/server/auth/auth.ts
```diff
@@ -0,0 +1,12 @@
+import { betterAuth } from "better-auth";
```
frectonz: This is also not being used and therefore should be removed.

packages/stripe/src/index.ts
```diff
@@ -12,7 +12,6 @@ import {
 	originCheck,
 	getSessionFromCtx,
 } from "better-auth/api";
-import { generateRandomString } from "better-auth/crypto";
```
frectonz: This is an unrelated change so it shouldn't be part of this PR.

frectonz:
@LoannPowell can you also fix the merge conflicts so that the tests can run.

packages/better-auth/src/adapters/prisma-adapter/test/state.txt
JuanPabloGilA: done

packages/cli/getConfig_test-Oujd4M/test/server/auth/auth.ts
JuanPabloGilA: done

packages/stripe/src/index.ts
JuanPabloGilA: done

JuanPabloGilA:
> @LoannPowell can you also fix the merge conflicts so that the tests can run.

I closed this one and opened this one https://github.com/better-auth/better-auth/pull/4440 it was easiest.

JuanPabloGilA:
I closed this one and opened this one https://github.com/better-auth/better-auth/pull/4440 it was easiest.

PR #3511, opened 2025-07-21
fix(teams): listTeams should prioritize the provided org ID

ping-maxwell:
Use the provided org id in the params, then if not provided, fallback to session active org id

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-team.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-team.ts
@@ -502,7 +502,7 @@ export const listOrganizationTeams = <O extends OrganizationOptions>(
 		async (ctx) => {
 			const session = ctx.context.session;
 			const organizationId =
-				session.session.activeOrganizationId || ctx.query?.organizationId;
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/organization/routes/crud-team.ts
```diff
@@ -502,7 +502,7 @@ export const listOrganizationTeams = <O extends OrganizationOptions>(
 		async (ctx) => {
 			const session = ctx.context.session;
 			const organizationId =
-				session.session.activeOrganizationId || ctx.query?.organizationId;
+				ctx.query?.organizationId || session.session.activeOrganizationId;
```
Bekacru, line 505: we need to make sure the user has access to the organization id in the query param

ping-maxwell: Wait what? We don't do that already??
Alright I'll get on to that.

ping-maxwell: image

Isn't this it?

himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #3512, opened 2025-07-21
fix(drizzle-cli): check for undefined defaultValue instead of truthy value

eni4sure:
## Summary by cubic
Fixed defaultValue check in the Drizzle schema generator to handle values like false or 0 correctly.


eni4sure:
Any update getting this merged? @Kinfe123 @Bekacru

PR #3542, opened 2025-07-22
feat(cli): added --yes option, deprecated --y, as per #1937

nktnet1:
See #1937.

This change adds the `--yes` option to the `migrate` and `generate` command.

`--y` will be kept, although treated as deprecated.

### DIFF
--- a/docs/content/changelogs/1.0.mdx
+++ b/docs/content/changelogs/1.0.mdx
@@ -20,7 +20,7 @@ npx @better-auth/cli@latest generate
 
 - `--output` - Where to save the generated schema. For Prisma, it will be saved in prisma/schema.prisma. For Drizzle, it goes to schema.ts in your project root. For Kysely, it’s an SQL file saved as schema.sql in your project root.
 - `--config` - The path to your Better Auth config file. By default, the CLI will search for a better-auth.ts file in **./**, **./utils**, **./lib**, or any of these directories under `src` directory.
-- `--y` - Skip the confirmation prompt and generate the schema directly.
+- `--yes` - Skip the confirmation prompt and generate the schema directly.
 
 ## Migrate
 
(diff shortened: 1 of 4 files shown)


docs/content/changelogs/1.0.mdx
```diff
@@ -20,7 +20,7 @@ npx @better-auth/cli@latest generate
 
 - `--output` - Where to save the generated schema. For Prisma, it will be saved in prisma/schema.prisma. For Drizzle, it goes to schema.ts in your project root. For Kysely, it’s an SQL file saved as schema.sql in your project root.
 - `--config` - The path to your Better Auth config file. By default, the CLI will search for a better-auth.ts file in **./**, **./utils**, **./lib**, or any of these directories under `src` directory.
-- `--y` - Skip the confirmation prompt and generate the schema directly.
+- `--yes` - Skip the confirmation prompt and generate the schema directly.
```
frectonz: We should probably not change the change log for `1.0` which was released a while ago. Remove this change.

docs/content/changelogs/1.0.mdx
```diff
@@ -33,7 +33,7 @@ npx @better-auth/cli@latest migrate
 ### Options
 
 - `--config` - The path to your Better Auth config file. By default, the CLI will search for a better-auth.ts file in **./**, **./utils**, **./lib**, or any of these directories under `src` directory.
-- `--y` - Skip the confirmation prompt and apply the schema directly.
+- `--yes` - Skip the confirmation prompt and apply the schema directly.
```
frectonz: Same here, we don't want to update an old changelog. Revert this change

docs/content/changelogs/1.0.mdx
Bhadboiab: Okay ✅ 
Actually is been awhile sir

nktnet1: Not sure what you mean @Bhadboiab.

Fixed, thanks @frectonz.

docs/content/changelogs/1.0.mdx
nktnet1: Fixed, thanks @frectonz.

Bhadboiab (commented):
Loading

PR #3554, opened 2025-07-23
fix(stripe): don't update incomplete stripe session in the db

### LINKED ISSUE #3545: Stripe plugin incorrectly setting status to active in DB without checking in Stripe

### To Reproduce

1. Call authClient.subscription.upgrade from the client, it redirects to Stripe hosted UI for checkout. This creates a stripe customer if does not exist and creates an entry in the DB subscriptions table with status incomplete. 
2. Click on the back button on the Stripe hosted checkout without doing a checkout. You'll be redirected back to the app
3. Make a call again to authClient.subscription.upgrade from the client, it redirects to Stripe hosted UI for checkout

Now, the status in the DB is changed to active without checking the actual status in Stripe.

Subscription is being set to active without checking the stripe_subscription_id in Stripe. If a row is not active or trialing, its status is set to active if a second checkout session is created

https://github.com/better-auth/better-auth/blob/main/packages/stripe/src/index.ts#L351-L353
https://github.com/better-auth/better-auth/blob/main/packages/stripe/src/index.ts#L399-L413

Expected. whenever a checkout session is created, if an entry exists in the db and stripe_subscription_id is NULL, it should not update the status without checking for subscriptions in stripe

1.3.2

```bash
- MacOS
```

Package

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

Bekacru:
closes #3545
    

---

## Summary by cubic
Stopped updating existing incomplete Stripe subscription sessions in the database. Now, only new incomplete subscriptions are created when needed.

### DIFF
--- a/packages/better-auth/src/plugins/jwt/index.ts
+++ b/packages/better-auth/src/plugins/jwt/index.ts
@@ -361,3 +361,5 @@ export const jwt = (options?: JwtOptions) => {
 		schema: mergeSchema(schema, options?.schema),
 	} satisfies BetterAuthPlugin;
 };
+
+export { getJwtToken };
--- a/packages/stripe/src/index.ts
+++ b/packages/stripe/src/index.ts
@@ -396,52 +396,18 @@ export const stripe = <O extends StripeOptions>(options: O) => {
 					});
 				}
 
-				let subscription = existingSubscription;
-				if (!subscription) {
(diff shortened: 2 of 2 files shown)


nowy:
Hey @Bekacru @docu this fixed the issue but brings back the original issue where a new subscription is created everytime a customer goes to the checkout section but cancels
 
image

PR #3555, opened 2025-07-23
fix(JWT): Re-export the getJwtToken function

ping-maxwell:
In a previous RP, we removed the export for that function which caused an unintentional breaking change. This will fix it.
    

---

## Summary by cubic
Re-exported the getJwtToken function from the JWT plugin to fix a breaking change caused by its previous removal.


dvanmali:
Fixed this in ~~#3464~~ #3572

ping-maxwell:
@dvanmali Your PR has additional things to consider, I made this PR to fix breaking change from a previous release, so it's purpose is to be small and merged as soon as possible.

Issue #3559, opened 2025-07-23
Use of null breaking things


Micnubinub:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

I think this commit here is braking things in my DB, and potentially others too:
https://github.com/better-auth/better-auth/commit/7d034604

```js
export function setTokenUtil(
	token: string | null | undefined,
	ctx: AuthContext,
) {
	if (ctx.options.account?.encryptOAuthTokens && token) {
		return symmetricEncrypt({
			key: ctx.secret,
			data: token,
		});
	}
	return token ?? null;
}
```
the set token util that is now being useed everywhere is setting values that would have been set as `undefined` before to be `null` and `null !== undefined`, so as time goes on I think more people will start to get issues

### Current vs. Expected behavior

expecting `undefined` but getting `null`. Might have to put it up as a breaking change

### What version of Better Auth are you using?

1.3.3

### Provide environment information

```bash
- MacOS 15.5
```

### Which area(s) are affected? (Select all that apply)

Backend, Types

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});
```

Micnubinub:
(SurrealDB users) I fixed my DB by changing`option<string>` to `option<string | NULL>`, other DBs can do something similar

```sql
DEFINE FIELD OVERWRITE ipAddress ON session TYPE option<string | NULL>;
DEFINE FIELD OVERWRITE location ON session TYPE option<string | NULL>;
DEFINE FIELD OVERWRITE userAgent ON session TYPE option<string | NULL>;
DEFINE FIELD OVERWRITE idToken ON account TYPE option<string | NULL>;
DEFINE FIELD OVERWRITE accessToken ON account TYPE option<string | NULL>;
DEFINE FIELD OVERWRITE refreshToken ON account TYPE option<string | NULL>;
DEFINE FIELD OVERWRITE accessTokenExpiresAt ON account TYPE option<datetime | NULL>;
DEFINE FIELD OVERWRITE refreshTokenExpiresAt ON account TYPE option<datetime | NULL>;
DEFINE FIELD OVERWRITE name ON passkey TYPE option<string | NULL>;
```

Bekacru:
Generally, an optional field is treated as nullable by the library, so setting it to null would be expected to work.

[...]

PR #3565, opened 2025-07-23
fix: preserve undefined tokens when encrypting

tinkerer-shubh:
possibly closes https://github.com/better-auth/better-auth/issues/3559

    

---

## Summary by cubic
Fixed token handling to preserve undefined values during encryption, preventing accidental overwrites of missing tokens.

- **Bug Fixes**
  - Updated token filtering to distinguish between null and undefined values.
  - Ensured setTokenUtil returns undefined tokens as-is.


Bekacru:
in favor of #3577

PR #3566, opened 2025-07-23
feat: implement getting the last used social provider

frectonz:
This PR introduces a new function called `lastUsedSocialProvider`, which returns the last social provider used for sign-in. This function can be utilized to display an indicator to users, allowing them to choose to always use the same social provider for sign-in.
    

---

## Summary by cubic
Added a new API endpoint and cookie to track and return the last social provider a user used to sign in.

- **New Features**
  - Stores the last used social provider in a cookie after sign-in.
  - Provides a GET endpoint to retrieve the last used social provider for the user.


Bekacru:
We wouldn't want to implement this within the core library. but could be a cool utility plugin :)

PR #3572, opened 2025-07-23
feat(jwt): remote jwks, remote signing, and exportable signJwt function

dvanmali:
## Features

**feat**: `remoteUrl` option disables jwks endpoint and uses this endpoint in oAuth metadata

**feat**: remote `sign` payloads using a server-side privateKey environment variable or remote signing service like Google KMS, AWS KMS, Azure Key Vault, etc.

**feat**: `signJwt` (previously `getJwtToken`) now accepts payload as a parameter. Those who utilized exported `getJwtToken` should utilize `signJwt` and its payload parameter instead of through options. `getJwtToken` marked deprecated. `signJwt` now has safeguards that is enabled by default and can be disabled per plugin implementation.

**feat**: Added `disableSettingJwtHeader` to disable setting jwt in header. Useful for oAuth compliance.

**feat**: exports `getJwtPlugin` for other plugins

**chore**: combine shared createJwks functionality

__Duplicate #3464__
__Partial #3458__

## Breaking Changes

**MINOR**
- style: `JwtOptions` renamed to `JwtPluginOptions` to match other plugin syntax.
- fix: `getJwtToken` deprecated, prefer `signJwt` instead

### DIFF
--- a/docs/content/docs/plugins/jwt.mdx
+++ b/docs/content/docs/plugins/jwt.mdx
@@ -307,13 +307,16 @@ jwt({
 
 ### OAuth Provider Mode
 
-If you are making your system oAuth compliant (such as when utilizing the OIDC or MCP plugins), you **MUST** disable the `/token` endpoint to prevent issuing JWT tokens without oAuth checks.
+If you are making your system oAuth compliant (such as when utilizing the OIDC or MCP plugins), you **MUST** disable the `/token` endpoint (oAuth equivalent `/oauth2/token`) and disable setting the jwt header (oAuth equivalent `oauth2/userinfo`).
 
 ```ts title="auth.ts"
 betterAuth({
   disabledPaths: [
     "/token",
   ],
+  plugins: [jwt({
+    disableSettingJwtHeader: true,
+  })]
 })
 ```
 
@@ -335,14 +338,6 @@ jwt({
   }
 })
 ```
-
-### JWT Safety Options
-
-**allowLongerExpTime** @default false - Set true to allow signing tokens past `expirationTime`
-
-**allowIssuerMismatch** @default false - Set true to allow issuer to mismatch jwt plugin option setting
-
-**allowAudienceMismatch** @default false - Set true to allow audience to mismatch jwt plugin option setting
 
 ### Custom Signing
 
--- a/docs/content/docs/plugins/jwt.mdx
+++ b/docs/content/docs/plugins/jwt.mdx
@@ -307,11 +307,13 @@ jwt({
 
 ### OAuth Provider Mode
 
-If you are making your system an oAuth provider by using the OIDC or MCP plugins, you **MUST** set this parameter to disable the `/token` endpoint and the ability to obtain tokens via the `set-auth-jwt`.
+If you are making your system oAuth compliant (such as when utilizing the OIDC or MCP plugins), you **MUST** disable the `/token` endpoint to prevent issuing JWT tokens without oAuth checks.
 
 ```ts title="auth.ts"
-jwt({
-  usesOAuthProvider: true,
+betterAuth({
+  disabledPaths: [
+    "/token",
+  ],
 })
 ```
 
@@ -321,25 +323,16 @@ Disables the `/jwks` endpoint and uses this endpoint in any discovery such as OI
 
 Useful if your JWKS are not managed at `/jwks` or if your jwks are signed with a certificate and placed on your CDN.
 
-```ts title="auth.ts"
-jwt({
-  jwks: {
-    remoteUrl: "https://example.com/.well-known/jwks.json",
-  }
-})
-```
-
-NOTE: If you are also using the OIDC plugin, you **MUST** specify which algorithm is used for signing as well.
+NOTE: you **MUST** specify which asymmetric algorithm is used for signing.
 
 ```ts title="auth.ts"
 jwt({
-  usesOAuthProvider: true,
   jwks: {
     remoteUrl: "https://example.com/.well-known/jwks.json",
     keyPairConfig: {
       alg: 'ES256',
     },
-  },
+  }
 })
 ```
 
@@ -366,6 +359,9 @@ Implementers:
 jwt({
   jwks: {
     remoteUrl: "https://example.com/.well-known/jwks.json",
+    keyPairConfig: {
+      alg: 'EdDSA',
+    },
   },
   jwt: {
     sign: async (jwtPayload: JWTPayload) => {
@@ -390,6 +386,9 @@ Useful if you are using a remote Key Management Service such as Google KMS, Amaz
 jwt({
   jwks: {
     remoteUrl: "https://example.com/.well-known/jwks.json",
+    keyPairConfig: {
+      alg: 'ES256',
+    },
   },
   jwt: {
     sign: async (jwtPayload: JWTPayload) => {
--- a/docs/content/docs/plugins/jwt.mdx
+++ b/docs/content/docs/plugins/jwt.mdx
@@ -345,13 +345,13 @@ jwt({
 
 ### JWT Safety Options
 
-**disallowFutureIatTime** @default true - Set false to allow setting issue times in the future.
+**allowFutureIatTime** @default false - Set true to allow setting issue times in the future.
 
-**disallowLongerExpTime** @default true - Set false to allows signing tokens past `expirationTime`
+**allowLongerExpTime** @default false - Set true to allow signing tokens past `expirationTime`
 
-**disallowIssuerMismatch** @default true - Allows issuer to mismatch jwt plugin option setting
+**allowIssuerMismatch** @default false - Set true to allow issuer to mismatch jwt plugin option setting
 
-**disallowAudienceMismatch** @default true - Allows audience to mismatch jwt plugin option setting
+**allowAudienceMismatch** @default false - Set true to allow audience to mismatch jwt plugin option setting
 
 ### Custom Signing
 
--- a/docs/content/docs/plugins/jwt.mdx
+++ b/docs/content/docs/plugins/jwt.mdx
@@ -345,8 +345,6 @@ jwt({
 
 ### JWT Safety Options
 
-**allowFutureIatTime** @default false - Set true to allow setting issue times in the future.
-
 **allowLongerExpTime** @default false - Set true to allow signing tokens past `expirationTime`
 
 **allowIssuerMismatch** @default false - Set true to allow issuer to mismatch jwt plugin option setting
@@ -377,7 +375,7 @@ jwt({
           alg: "EdDSA",
           kid: process.env.currentKid,
           typ: "JWT",
-        });
+        })
         .sign(process.env.clientPrivateKey);
     },
   },
--- a/docs/content/docs/plugins/jwt.mdx
+++ b/docs/content/docs/plugins/jwt.mdx
@@ -304,3 +304,112 @@ jwt({
   }
 })
 ```
+
+### Oauth Provider Mode
+
+If you are making your system an oAuth provider by using the OIDC or MCP plugins, you **MUST** set this parameter to disable the `/token` endpoint and the ability to obtain tokens via the `set-auth-jwt`.
+
+```ts title="auth.ts"
+jwt({
+  usesOauthProvider: true,
+})
+```
+
+### Remote JWKS Url
+
+Disables the `/jwks` endpoint and uses this endpoint in any discovery such as OIDC.
+
+Useful if your JWKS are not managed at `/jwks` or if your jwks are signed with a certificate and placed on your CDN.
+
+```ts title="auth.ts"
+jwt({
+  jwks: {
+    remoteUrl: "https://example.com/.well-known/jwks.json",
+  }
+})
+```
+
+NOTE: If you are also using the OIDC plugin, you **MUST** specify which algorithm is used for signing as well.
+
+```ts title="auth.ts"
+jwt({
+  usesOauthProvider: true,
+  jwks: {
+    remoteUrl: "https://example.com/.well-known/jwks.json",
+    keyPairConfig: {
+      alg: 'ES256',
+    },
+  },
+})
+```
+
+### JWT Safety Options
+
+**disallowFutureIatTime** @default true - Set false to allow setting issue times in the future.
+
+**disallowLongerExpTime** @default true - Set false to allows signing tokens past `expirationTime`
+
+**disallowIssuerMismatch** @default true - Allows issuer to mismatch jwt plugin option setting
+
+**disallowAudienceMismatch** @default true - Allows audience to mismatch jwt plugin option setting
+
+### Custom Signing
+
+This is an advanced feature! Configuration outside of this plugin MUST be provided.
+
+Implementers:
+- `remoteUrl` must be defined if using the `sign` function. This shall store all active keys, not just the current one.
+- If using localized approach, ensure server uses the latest private key when rotated. Depending on deployment, the server may need to be restarted.
+- When using remote approach, verify the payload is unchanged after transit. Use integrity validation like CRC32 or SHA256 checks if available.
+
+#### Localized Signing
+
+```ts title="auth.ts"
+jwt({
+  jwks: {
+    remoteUrl: "https://example.com/.well-known/jwks.json",
+  },
+  jwt: {
+    sign: async (jwtPayload: JWTPayload) => {
+      // this is pseudocode
+      return await new SignJWT(jwtPayload)
+        .setProtectedHeader({
+          alg: "EdDSA",
+          kid: process.env.currentKid,
+          typ: "JWT",
+        });
+        .sign(process.env.clientPrivateKey);
+    },
+  },
+})
+```
+
+#### Remote Signing
+
+Useful if you are using a remote Key Management Service such as Google KMS, Amazon KMS, or Azure Key Vault.
+
+```ts title="auth.ts"
+jwt({
+  jwks: {
+    remoteUrl: "https://example.com/.well-known/jwks.json",
+  },
+  jwt: {
+    sign: async (jwtPayload: JWTPayload) => {
+      // this is pseudocode
+      const headers = JSON.stringify({ kid: '123', alg: 'ES256', typ: 'JWT' })
+      const payload = JSON.stringify(jwtPayload)
+      const encodedHeaders = Buffer.from(headers).toString('base64url')
+      const encodedPayload = Buffer.from(payload).toString('base64url')
+      const hash = createHash('sha256')
+      const data = `${encodedHeaders}.${encodedPayload}`
+      hash.update(Buffer.from(data))
+      const digest = hash.digest()
+      const sig = await remoteSign(digest)
+      // integrityCheck(sig)
+      const jwt = `${data}.${sig}`
+      // verifyJwt(jwt)
+      return jwt
+    },
+  },
+})
+```
--- a/docs/content/docs/plugins/jwt.mdx
+++ b/docs/content/docs/plugins/jwt.mdx
@@ -305,13 +305,13 @@ jwt({
 })
 ```
 
-### Oauth Provider Mode
+### OAuth Provider Mode
 
 If you are making your system an oAuth provider by using the OIDC or MCP plugins, you **MUST** set this parameter to disable the `/token` endpoint and the ability to obtain tokens via the `set-auth-jwt`.
 
 ```ts title="auth.ts"
 jwt({
-  usesOauthProvider: true,
+  usesOAuthProvider: true,
 })
 ```
 
@@ -333,7 +333,7 @@ NOTE: If you are also using the OIDC plugin, you **MUST** specify which algorith
 
 ```ts title="auth.ts"
 jwt({
-  usesOauthProvider: true,
+  usesOAuthProvider: true,
   jwks: {
     remoteUrl: "https://example.com/.well-known/jwks.json",
     keyPairConfig: {
--- a/packages/better-auth/src/plugins/jwt/index.ts
+++ b/packages/better-auth/src/plugins/jwt/index.ts
@@ -1,228 +1,173 @@
-import type {
-	BetterAuthPlugin,
-	InferOptionSchema,
-	Session,
-	User,
-} from "../../types";
-import { type Jwk, schema } from "./schema";
+import type { BetterAuthPlugin, HookEndpointContext } from "../../types";
+import { schema } from "./schema";
 import { getJwksAdapter } from "./adapter";
-import { getJwtToken } from "./sign";
-import { exportJWK, generateKeyPair, type JWK } from "jose";
 import {
 	createAuthEndpoint,
 	createAuthMiddleware,
 	sessionMiddleware,
 } from "../../api";
-import { symmetricEncrypt } from "../../crypto";
 import { mergeSchema } from "../../db/schema";
+import { BetterAuthError } from "../../error";
+import { createJwk, signJwt } from "./sign";
+import type { JwtPluginOptions } from "./types";
+export type * from "./types";
+export { createJwk, getJwtToken } from "./sign";
+export { getJwtPlugin } from "./utils";
 
-type JWKOptions =
-	| {
-			alg: "EdDSA"; // EdDSA with either Ed25519 or Ed448 curve
-			crv?: "Ed25519" | "Ed448";
-	  }
-	| {
-			alg: "ES256"; // ECDSA with P-256 curve
-			crv?: never; // Only one valid option, no need for crv
-	  }
-	| {
-			alg: "RS256"; // RSA with SHA-256
-			modulusLength?: number; // Default to 2048 or higher
-	  }
-	| {
-			alg: "PS256"; // RSA-PSS with SHA-256
-			modulusLength?: number; // Default to 2048 or higher
-	  }
-	| {
-			alg: "ECDH-ES"; // Key agreement algorithm with P-256 as default curve
-			crv?: "P-256" | "P-384" | "P-521";
-	  }
-	| {
-			alg: "ES512"; // ECDSA with P-521 curve
-			crv?: never; // Only P-521 for ES512
-	  };
+export const jwt = (options?: JwtPluginOptions) => {
+	const endpoints: BetterAuthPlugin["endpoints"] = {};
+	const hooks: BetterAuthPlugin["hooks"] = {};
 
-export interface JwtOptions {
-	jwks?: {
-		/**
-		 * Key pair configuration
-		 * @description A subset of the options available for the generateKeyPair function
-		 *
-		 * @see https://github.com/panva/jose/blob/main/src/runtime/node/generate.ts
-		 *
-		 * @default { alg: 'EdDSA', crv: 'Ed25519' }
-		 */
-		keyPairConfig?: JWKOptions;
+	// Remote url must be set when using signing function
+	if (options?.jwt?.sign && !options.jwks?.remoteUrl) {
+		throw new BetterAuthError(
+			"jwks_config",
+			"jwks.remoteUrl must be set when using jwt.sign",
+		);
+	}
 
-		/**
-		 * Disable private key encryption
-		 * @description Disable the encryption of the private key in the database
-		 *
-		 * @default false
-		 */
-		disablePrivateKeyEncryption?: boolean;
-	};
+	// Alg is required to be specified when using oidc plugin and remote url (needed in openid metadata)
+	if (
+		options?.usesOauthProvider &&
+		options.jwks?.remoteUrl &&
+		!options.jwks?.keyPairConfig?.alg
+	) {
+		throw new BetterAuthError(
+			"jwks_config",
+			"must specify alg when using the oidc plugin and jwks.remoteUrl",
+		);
+	}
 
-	jwt?: {
-		/**
-		 * The issuer of the JWT
-		 */
-		issuer?: string;
-		/**
-		 * The audience of the JWT
-		 */
-		audience?: string;
-		/**
-		 * Set the "exp" (Expiration Time) Claim.
-		 *
-		 * - If a `number` is passed as an argument it is used as the claim directly.
-		 * - If a `Date` instance is passed as an argument it is converted to unix timestamp and used as the
-		 *   claim.
-		 * - If a `string` is passed as an argument it is resolved to a time span, and then added to the
-		 *   current unix timestamp and used as the claim.
-		 *
-		 * Format used for time span should be a number followed by a unit, such as "5 minutes" or "1
-		 * day".
-		 *
-		 * Valid units are: "sec", "secs", "second", "seconds", "s", "minute", "minutes", "min", "mins",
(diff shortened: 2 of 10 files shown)


LightTab2:
Fixes #3226

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -53,23 +60,101 @@ export async function getJwtToken(
...
+			typ: "JWT",
 		})
-		.setIssuedAt()
-		.setIssuer(options?.jwt?.issuer ?? ctx.context.options.baseURL!)
-		.setAudience(options?.jwt?.audience ?? ctx.context.options.baseURL!)
-		.setExpirationTime(options?.jwt?.expirationTime ?? "15m")
-		.setSubject(
-			(await options?.jwt?.getSubject?.(ctx.context.session!)) ??
-				ctx.context.session!.user.id,
+		.setIssuedAt(payload.iat)
```
LightTab2: If options are defined and payload has different parameters, I'm not sure which one should take precedence or even if a mismatch is a case for an error throw. Chosen behaviour should be stated in function's description I think

packages/better-auth/src/plugins/jwt/types.ts
```diff
@@ -0,0 +1,135 @@
...
+			modulusLength?: number; // Default to 2048 or higher
+	  }
+	| {
+			alg: "RS256"; // RSA with SHA-256
+			modulusLength?: number; // Default to 2048 or higher
+	  };
+
+export type JWSAlgorithms = JWKOptions["alg"];
+
+export interface JwtPluginOptions {
```
LightTab2, line 31: This is a breaking change to type, add this to PR description :)

packages/better-auth/src/plugins/jwt/index.ts
```diff
@@ -1,315 +1,172 @@
...
 	createAuthEndpoint,
 	createAuthMiddleware,
 	sessionMiddleware,
 } from "../../api";
-import { symmetricEncrypt } from "../../crypto";
 import { mergeSchema } from "../../db/schema";
+import { BetterAuthError } from "../../error";
+import { createJwk, signJwt } from "./sign";
+import type { JwtPluginOptions } from "./types";
+export * from "./types";
```
LightTab2: Will `export type * from ".types";` work here?

packages/better-auth/src/plugins/jwt/sign.ts
LightTab2: Also include in description that `options?.jwt?.definePayload` is ignored

LightTab2: Honestly I can't shake off feeling that calling `setIssuedAt` to other time than current one is a bad idea. I can't logically explain it why though and even see cases when its desired...

LightTab2: Maybe at least a safeguard boolean parameters `disallowOptionsMismatch`, `disallowFutureIssueTime` with defaulted to `true`, that does not allow expiration time (and issuer?) to differ from options' one and issued time to be later than now.

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -53,23 +60,101 @@ export async function getJwtToken(
...
+		)
+		.setAudience(
+			payload.aud ?? options?.jwt?.audience ?? ctx.context.options.baseURL!,
 		)
-		.sign(privateKey);
-	return jwt;
+		.setExpirationTime(payload.exp ?? options?.jwt?.expirationTime ?? "15m");
+	const sub =
+		(await options?.jwt?.getSubject?.(ctx.context.session!)) ??
+		payload.sub ??
```
LightTab2, line 129: Currently payload arguments take precedence everywhere except here

LightTab2 (changes requested):
I have some security concerns and think some safeguards could be used to make sure user  wants to override critical settings with their payload. I can imagine a case when someone creates a sloppy API endpoint that signs anything and the payload could be maliciously crafted

packages/better-auth/src/plugins/jwt/sign.ts
dvanmali: JWT structure is `${protectedHeader}.${jwtPayload}.${signature}`. Typically the payload should never have `alg` defined as this is entirely determined by the config.

If you put `alg` then technically the output would be in a payload claim not a header claim. In which verification from packages like `jose` aren't designed to verify. Plus it'll be redundant if defined in both header and payload.

dvanmali: Adding the `definePayload` description.

dvanmali: only if `payload.iat` is defined will it issue that value instead otherwise the default is now. I include this because I need to control the iat (and exp) in a future commits for the OIDC package where sessions/tokens should equate to the same value on the database. [Looking ahead](https://github.com/dvanmali/better-auth/blob/9bacc1b8bed9cda0ae7b5d65539ea21a246f22ba/packages/better-auth/src/plugins/oidc-provider/token.ts#L251)

dvanmali: I think `disallowFutureIssueTime` would be a good, simple safeguard to add. I'm guessing `disallowOptionsMismatch` is for issuer and audience which would also be good to add as well.

Will add.

LightTab2: There are 2 missing possible fields:
```js
	if (payload.nbf) jwt.setNotBefore(payload.nbf);
	if (payload.jti) jwt.setJti(payload.jti);
```

dvanmali: For these, if they are defined inside payload, it'll work in the same way as setting them individually.

dvanmali:
@cubic-dev-ai review again

dvanmali:
@cubic-dev-ai review again

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -1,42 +1,95 @@
...
+	const disallowLargerExpTime = !(options?.jwt?.allowLongerExpTime ?? false);
+	const disallowIssuerMismatch = !(options?.jwt?.allowIssuerMismatch ?? false);
+	const disallowAudienceMismatch = !(
+		options?.jwt?.allowAudienceMismatch ?? false
+	);
+
+	// Iat safety check
+	let iat = payload.iat;
+	if (disallowFutureIatTime && iat && iat > Date.now() / 1000) {
+		throw new Error("unable to set a future iat time");
```
LightTab2: Prefer `BetterAuthError` in place of `Error`

packages/better-auth/src/plugins/jwt/types.ts
```diff
@@ -0,0 +1,159 @@
...
+	 * Valid units are: "sec", "secs", "second", "seconds", "s", "minute", "minutes", "min", "mins",
+	 * "m", "hour", "hours", "hr", "hrs", "h", "day", "days", "d", "week", "weeks", "w", "year",
+	 * "years", "yr", "yrs", and "y". It is not possible to specify months. 365.25 days is used as an
+	 * alias for a year.
+	 *
+	 * If the string is suffixed with "ago", or prefixed with a "-", the resulting time span gets
+	 * subtracted from the current unix timestamp. A "from now" suffix can also be used for
+	 * readability when adding to the current unix timestamp.
+	 *
+	 * @default 1hr
```
LightTab2: Changing the default from "15m" to "1h" is debatable, this might be done in another PR to not slow down reviewing of this PR by questioning ourselves which one is better

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -1,42 +1,95 @@
...
+ * @param ctx - endpoint context
+ * @param payload - payload to sign
+ */
+export async function signJwtPayload(
 	ctx: GenericEndpointContext,
-	options?: JwtOptions,
+	payload: JWTPayload,
+	options?: JwtPluginOptions,
 ) {
+	const disallowFutureIatTime = !(options?.jwt?.allowFutureIatTime ?? false);
```
LightTab2: Can just check `!options?.jwt?.allowFutureIatTime` where it is needed, these variables are used only once.

`undefined` is **falsy**.

packages/better-auth/src/plugins/jwt/index.ts
```diff
@@ -1,315 +1,172 @@
...
-		 * Key pair configuration
-		 * @description A subset of the options available for the generateKeyPair function
-		 *
-		 * @see https://github.com/panva/jose/blob/main/src/runtime/node/generate.ts
-		 *
-		 * @default { alg: 'EdDSA', crv: 'Ed25519' }
-		 */
-		keyPairConfig?: JWKOptions;
+	// Remote url must be set when using signing function
+	if (options?.jwt?.sign && !options.jwks?.remoteUrl) {
```
LightTab2, line 20:
```js
	const jwtOpts: JwtOptions | undefined = options?.jwt;
	const jwksOpts: JwksOptions | undefined = options?.jwks;
```
Might add these to simplify reading

packages/better-auth/src/plugins/jwt/utils.ts
```diff
@@ -1,5 +1,83 @@
...
+		| (Omit<BetterAuthPlugin, "options"> & { options?: JwtPluginOptions })
+		| undefined = ctx.options.plugins?.find((plugin) => plugin.id === "jwt");
+	if (!plugin) {
+		throw new BetterAuthError("jwt_config", "jwt plugin not found");
+	}
+	return plugin;
+};
+
+export function toExpJWT(input: number | Date | string, iat?: number): number {
+	const now = iat ?? Math.floor(Date.now() / 1000); // current Unix timestamp in seconds
```
LightTab2: Adding `iat` here allows to create correct JWT tokens that are to be send in future, like caching them... It's weird but some might want to use this scenario, however if `iat > Date.now()/1000` and this token gets leaked, somebody basically got their hands on a token that has longer Expiration Time than allowed... So I propose another safeguard:
`allowFutureIatWithoutNbf`. 
Logic of the safeguard would be:
> Set `nbf` to `iat`, if `iat` is set to future AND `nbf `is `undefined` or **lesser** than `iat` unless `allowFutureIatWithoutNbf` is `true`

packages/better-auth/src/plugins/jwt/utils.ts
```diff
@@ -1,5 +1,83 @@
...
+
+	if (typeof input === "number") {
+		return now + input;
```
LightTab2: This is wrong, when `exp` is a `number`, it is a **UNIX timestamp**.

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -55,23 +108,120 @@ export async function getJwtToken(
...
+		})
+		.setIssuedAt(iat)
+		.setExpirationTime(exp ?? defaultExp)
+		.setIssuer(iss ?? defaultIss)
+		.setAudience(aud ?? defaultAud);
+	const sub =
+		payload.sub ??
+		(await options?.jwt?.getSubject?.(ctx.context.session!)) ??
+		ctx.context.session?.user.id;
+	if (sub) jwt.setSubject(sub);
```
LightTab2, line 132:
```js
	if (payload.nbf) jwt.setNotBefore(payload.nbf);
	if (payload.jti) jwt.setJti(payload.jti);
```
Any reason to not set these?

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -55,23 +108,120 @@ export async function getJwtToken(
...
+	return key;
+}
+
+/**
+ * Creates a new JWK (JSON Web Key) on the database.
+ *
+ * @param ctx - endpoint context
+ */
+// Plugin exportable
+export async function createJwk(ctx: GenericEndpointContext) {
```
LightTab2, line 235: Note lack of `options` as breaking change.

packages/better-auth/src/plugins/jwt/utils.ts
```diff
@@ -1,5 +1,83 @@
...
+		throw new Error(`Unsupported unit: ${unit}`);
+	}
+
+	const totalSeconds = value * seconds;
+	const isSubtraction = negativePrefix || suffix?.toLowerCase() === "ago";
+
+	return Math.floor(isSubtraction ? now - totalSeconds : now + totalSeconds);
+}
 
 async function deriveKey(secretKey: string): Promise<CryptoKey> {
```
LightTab2: Are `deriveKey`, `encryptPrivateKey`, `decryptPrivateKey` functions used anywhere?

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -1,42 +1,95 @@
...
+	// Iat safety check
+	let iat = payload.iat;
+	if (disallowFutureIatTime && iat && iat > Date.now() / 1000) {
+		throw new Error("unable to set a future iat time");
+	}
+
+	// Exp safety check
+	let exp = payload.exp;
+	const defaultExp = options?.jwt?.expirationTime ?? "1h";
+	if (disallowLargerExpTime && exp && exp > toExpJWT(defaultExp, iat)) {
```
LightTab2: Correct to
```js
if (disallowLargerExpTime && (!exp || exp > toExpJWT(defaultExp, iat)))
```

Missing `exp` creates an infinite token, can't have that. In fact there should be a comment as precaution that `allowLargerExpTime` set to `true` will allow to create infinite tokens.

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -1,42 +1,95 @@
...
+			}
+		}
+	}
+
+	// Custom/remote signing function
+	if (options?.jwt?.sign) {
+		payload = {
+			...payload,
+			iat,
+			exp: exp ?? toExpJWT(defaultExp, iat),
```
LightTab2: Don't know if other params are needed. Honestly, I don't know if anything related to remote signing could be improved upon, because I would have to try using it in the first place. Someone could take a second look at this and oidc code.

LightTab2 (changes requested):
`exp` check needs to be fixed. Other comments are not so important and opinion-biased.

packages/better-auth/src/plugins/jwt/types.ts
dvanmali: It's required in this one because of the reliance of oidc-provider on the jwt package. Either all instances of `oidc-provider` needs to either lower the access token expiration time to "15 min" or set `allowLongerExpTime: true`, or `jwt` package needed to increase to `1hr`. I went with the latter since we are dealing only with `jwt` changes in this PR.

packages/better-auth/src/plugins/jwt/sign.ts
dvanmali: Likely redundant but I can add it in if `SignJWT` has any built in checkers there

packages/better-auth/src/plugins/jwt/sign.ts
dvanmali: I don't believe the jwt function needs to verify or default anything beyond `iat`, `exp`, `iss`, and `aud`. The payload in oidc plugin will be able to add custom claims to `id_token` and JWT `access_tokens` in which verification of those claims should be completed in oidc plugin.

packages/better-auth/src/plugins/jwt/sign.ts
LightTab2: I was wrong (didn't look ahead), the `exp` will be set to default one if its `undefined`, so the check was alright. Infinite tokens are forbidden by design.

packages/better-auth/src/plugins/jwt/utils.ts
dvanmali: ~~Good catch! For future iat issuing, I'd propose a strict policy on the JWT package that no future token shall be allowed unless `nbf` is set and is lesser than exp. I couldn't see a reason why `nbf` should be flagged to be disabled for future tokens. I will add the check to the `signJwtPayload` instead.~~

Update: `iat` should never be allowed in the future, by definition it is the issued time, thus can only be the current_time or in the past! Only `nbf` should be allowed in the future (but less than `exp`). Removed the `allowFutureIatTime` flag.

packages/better-auth/src/plugins/jwt/utils.ts
dvanmali: Not in this plugin, nor other plugins, nor exported from the plugin. I think it can safely be removed.

dvanmali: Looking at the history, it appears the dependency both moved away into `better-auth/utils` package and changed down the road to use the `jose` package and its `generateKeyPair`.

packages/better-auth/src/plugins/jwt/index.ts
dvanmali: Feels easier to work directly with options to make tracing easier as options are used everywhere, style choice imo, will leave as is for now.

packages/better-auth/src/plugins/jwt/types.ts
dvanmali: @LightTab2 if this sounds good to you, you or I can mark this as resolved.

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -1,42 +1,100 @@
...
+	const allowLargerExpTime = options?.jwt?.allowLongerExpTime;
+	const defaultExp = toExpJWT(options?.jwt?.expirationTime ?? "1h", iat ?? nowSeconds);
+	if (!allowLargerExpTime && exp && exp > defaultExp) {
+		throw new BetterAuthError("unable to set future exp time");
+	}
+	exp = exp ?? defaultExp;
+
+	// Nbf strict checks
+	const nbf = payload.nbf;
+	if (nbf &&  !((iat && nbf < iat) || nbf > exp)) {
```
dvanmali: Implementation Note: Using `iat && nbf < iat` instead of `nbf < (iat ?? nowSeconds)` because nbf can only be checked against iat if sent, nowSeconds is local and may fail if used in this comparison.

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -1,42 +1,100 @@
...
 	ctx: GenericEndpointContext,
-	options?: JwtOptions,
+	payload: JWTPayload,
+	options?: JwtPluginOptions,
 ) {
+	// Iat strict checks
+	const nowSeconds = Math.floor(Date.now() / 1000);
+	const iat = payload.iat;
+	if (iat && iat > nowSeconds) {
+		throw new BetterAuthError("unable to set a future iat time");
```
dvanmali, line 34: By definition `iat` is the current issued time. It can never be set to a future time nor should it be bypassed. Utilize `nbf` if the issuance is set for the future.

dvanmali:
@cubic-dev-ai review again

packages/better-auth/src/plugins/jwt/sign.ts
dvanmali: Shouldn't create infinite token. It'll go to the default.

packages/better-auth/src/plugins/jwt/sign.ts
dvanmali: This code is a new function entirely. Not breaking. Equivalent to `signJwt` in which no options is provided and uses the option. Intended use is for a future admin plugin to be able to create/rotate jwks, thus is exported in index.ts

packages/better-auth/src/plugins/jwt/types.ts
LightTab2: I don't think it's about personal preference. I understand the concern that `oicd-provider` asks for '1h' JWT, but the most common practice is to set their expiration time to 5 - 15 minutes. Because even if user logs out and JWT token gets leaked, it still can be used to permit actions in the name of the user who just logged out. Someone from Better Auth Team needs to approve this

packages/better-auth/src/plugins/jwt/utils.ts
LightTab2: Yes it shouldn't, but if we allow to "sign" any payload, we already don't care if the contents are valid, but if they are safe despite being created by other system which might not adhere to standards (hence safeguards, if we still expect standards to be followed). This gives us much control in some uncommon cases where it is needed. 

If the design choice is to sign data (not payload) with our JWT plugin, the function signature should be:
```js
async function signJwt(
	ctx: GenericEndpointContext,
	data: readonly {[propName: string]: unknown},
)
```

And we populate the `Payload` fields with our JWT Plugin Settings.

LightTab2 (commented):
I didn't review OIDC changes; custom sign function tries to achieve golden ratio between customization and safety, personally I'd like to have 2 separate functions, one which simply signs using plugin's JWT settings and one which allows for every shenanigans possible (with togglable safeguards), but this is a good solution all in all and here I talked only about my preferences.

I won't mark this review as approved, because I want somebody more knowledgable to take notion of the default time change from "15m" to "1h" and decide whether its a good move first and if we need to make a workaround around OIDC instead.

packages/better-auth/src/plugins/jwt/types.ts
dvanmali: I agree access tokens are the most secure with lower expiration times at `15 min`, which should be used in highly secure systems. But `1 hr` works best for most web applications and this setting serves as a default absolute max for other plugins. Other plugins can always lower their absolute expiration (this JWT-plugin setting) or manually lower the exp field in their plugin (usually scope dependent). There's just a lot more breaking changes and higher likeliness to setup improper settings if `15 min` is kept.

Note for better-auth, the alternative to setting `15 min` here requires setting it also on OIDC for all default JWT token times (`idTokens` and `accessTokens`).

packages/better-auth/src/plugins/jwt/utils.ts
dvanmali: `signJwt` must at least follow the strict standards of [JWT RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519#section-4) which specifies the `iat`, `exp`, and `nbf` claims. Thus, `signJwt` should only sign a **valid** JWT payload (not just data) makes the most valid descriptive sense to me.

dvanmali:
Ready for review @Bekacru

dvanmali:
@Bekacru?

dvanmali:
@ping-maxwell @Bekacru

packages/better-auth/src/plugins/jwt/index.ts
```diff
@@ -321,23 +180,59 @@ export const jwt = (options?: JwtOptions) => {
 					},
 				},
 				async (ctx) => {
-					const jwt = await getJwtToken(ctx, options);
+					// Disables endpoint if using oAuth provider
+					if (options?.usesOAuthProvider) {
```
Bekacru: I'm not sure if see the point of using this flag. The goal seems to be only to disable the token endpoint or am I missing something? If that's the case, they can already do that in other ways, including with the `disabledPaths` option.

packages/better-auth/src/plugins/jwt/index.ts
```diff
@@ -321,23 +180,59 @@ export const jwt = (options?: JwtOptions) => {
...
+						token,
 					});
 				},
 			),
 		},
 		hooks: {
 			after: [
 				{
-					matcher(context) {
+					matcher(context: HookEndpointContext) {
```
Bekacru: we don't need to provide the type here, it should be inferred

packages/better-auth/src/plugins/jwt/index.ts
```diff
@@ -321,23 +180,59 @@ export const jwt = (options?: JwtOptions) => {
...
 						return context.path === "/get-session";
 					},
 					handler: createAuthMiddleware(async (ctx) => {
+						// Disables middleware if using oAuth provider
+						if (options?.usesOAuthProvider) {
```
Bekacru: here it might make sense to use this flag but I think another flag to specifically disable setting the jwt token on a hook is better like `disableSettingJWTHeader`

packages/better-auth/src/plugins/jwt/index.ts
```diff
@@ -321,23 +180,59 @@ export const jwt = (options?: JwtOptions) => {
...
+							// Convert context into user payload
+							let payload: Record<string, any>;
+							if (options?.jwt?.definePayload) {
+								payload = await options?.jwt.definePayload(
+									ctx.context.session!,
+								);
+							} else {
+								payload = {
+									...(ctx.context.session?.user ?? {}),
+									id: undefined, // id becomes sub in Sign Function
```
Bekacru: this is I think a breaking change, given there would be users who relied on the `id` field. Also keeping the id should be fine even if it's already on the `sub`

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -1,42 +1,103 @@
...
+	const allowIssuerMismatch = options?.jwt?.allowIssuerMismatch;
+	if (!allowIssuerMismatch && iss && iss !== defaultIss) {
+		throw new BetterAuthError(`iss ${iss} not allowed`);
+	}
+
+	// Aud safety checks (for non-oAuth mode, audience checking shall be performed in oAuth plugin instead)
+	const aud = payload.aud;
+	const defaultAud = options?.jwt?.audience ?? ctx.context.options.baseURL!;
+	const allowAudienceMismatch = options?.jwt?.allowAudienceMismatch;
+	if (!options?.usesOAuthProvider && !allowAudienceMismatch && aud) {
```
Bekacru: here also we should check regardless of `useOAuthProvider` if there is allowed audiences list

packages/better-auth/src/plugins/jwt/index.ts
dvanmali: That is correct, `usesOAuthProvider` disables the `/token` endpoint because an oAuth provider should use their implemented version of `/token` instead. This endpoint would return a token without the proper checks required by oAuth and MUST be disabled. I'll look into a way to perform this through `disabledPaths` instead of creating the additional flag and move the check into OIDC instead.

packages/better-auth/src/plugins/jwt/index.ts
dvanmali: Valid, will implement

packages/better-auth/src/plugins/jwt/sign.ts
dvanmali: That is supplied by `defaultAud` which is always type `string` or `string[]` (defaults to the baseUrl). Thus, audience check is always performed when `payload.aud` is supplied.

packages/better-auth/src/plugins/jwt/index.ts
dvanmali: Sounds good, will return to original state

packages/better-auth/src/plugins/jwt/index.ts
dvanmali: NOTE: this requires all `oidcProvider` plugins to set `allowAudienceMismatch: true` in order to sign id tokens

dvanmali:
@cubic-dev-ai review

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -1,42 +1,103 @@
...
+	const allowIssuerMismatch = options?.jwt?.allowIssuerMismatch;
+	if (!allowIssuerMismatch && iss && iss !== defaultIss) {
+		throw new BetterAuthError(`iss ${iss} not allowed`);
+	}
+
+	// Aud safety checks (for non-oAuth mode, audience checking shall be performed in oAuth plugin instead)
+	const aud = payload.aud;
+	const defaultAud = options?.jwt?.audience ?? ctx.context.options.baseURL!;
+	const allowAudienceMismatch = options?.jwt?.allowAudienceMismatch;
+	if (!allowAudienceMismatch && aud) {
```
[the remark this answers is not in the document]
dvanmali: Flag was removed and linked plugin should set `allowAudienceMismatch` instead

packages/better-auth/src/plugins/jwt/index.ts
```diff
@@ -1,145 +1,37 @@
...
+	// Remote url must be set when using signing function
+	if (options?.jwt?.sign && !options.jwks?.remoteUrl) {
+		throw new BetterAuthError(
+			"jwks_config",
+			"jwks.remoteUrl must be set when using jwt.sign",
+		);
+	}
+
+	// Alg is required to be specified when using oidc plugin and remote url (needed in openid metadata)
+	if (options?.jwks?.remoteUrl && !options.jwks?.keyPairConfig?.alg) {
```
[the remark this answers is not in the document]
dvanmali: Chose to specify `alg` everytime `remoteUrl` since `usesOAuthProvider` flag is removed. Updated comment to align.

packages/better-auth/src/plugins/jwt/types.ts
dvanmali: Update, I was able to keep the `15min` using `safeguards` parameter. Marking it as resolved.

packages/better-auth/src/plugins/jwt/index.ts
dvanmali: Was able to remove `usesOAuthProvider` by adding a `safeguards` parameter instead of going through options.

docs/content/docs/plugins/jwt.mdx
```diff
@@ -304,3 +304,104 @@ jwt({
   }
 })
 ```
+
+### OAuth Provider Mode
+
+If you are making your system oAuth compliant (such as when utilizing the OIDC or MCP plugins), you **MUST** disable the `/token` endpoint (oAuth equivalent `/oauth2/token`) and disable setting the jwt header (oAuth equivalent `oauth2/userinfo`).
```
Bekacru, line 310: I'm not sure why it's a must that they disable this?

dvanmali: It should be a must because oAuth requires scope checking. This jwt plugin does not require scope checking. (ie having scopes `openid` vs `openid email` provides different information on the userinfo endpoint).

packages/better-auth/src/plugins/jwt/sign.ts
```diff
@@ -55,23 +116,123 @@ export async function getJwtToken(
...
+	safeguards?: JwtSignSafeguards,
+) {
+	const options = getJwtPlugin(ctx.context).options;
+	return signJwtPayload(ctx, payload, options, safeguards);
+}
+
+/**
+ * Backwards compatable version of signJwt
+ *
+ * @deprecated - prefer signJwt to prevent option injection
```
dvanmali, line 158: Actually does not prevent option injection because updating the `ctx.context` directly would do the same.

dvanmali, line 158: Actually this doesn't prevent option injection - updating `ctx.context.options.plugins` will do the same. The `plugin` variable must be `readonly` after it is initialized. I'd consider this a _minor_ security concern since package implementers control the options and its fix would be considered a _major_ package fix if implemented.

gustavovalverde:
@dvanmali this was partially shipped in `next`, I'm closing this PR, but if you'd like to continue with the the work please open a new updated PR

PR #3580, opened 2025-07-24
feat(organization): add membersLimit param to allow to fetch more or less members than the membership limit

### LINKED ISSUE #3467: `findFullOrganization` should not be limited by `membershipLimit`

### To Reproduce

Currently, `findFullOrganization` (and `findTeamById` & `listMembers`) limit their member queries to `membershipLimit` (default `100`). If there are more valid memberships in the database* this results (a) in an inconsistent state and (b) let's `setActiveOrganization` calls fail (related: #3466, #3465).

*Potential Reasons:
- The limit was previously higher and got reduced
- `members` where added to the database w/o better-auth

Even though if the argumentation is something like "but we want to never have a state where an organization has more members than `membershipLimit`", I would respond:

1. This results in state inconsistencies between DB & API
2. This limit should be mainly enforced when adding new members via the API
3. And even If the limitation is decided to be kept, there should be a clear sorting/filtering strategy & documentation on **which** members to return and **which not**.

Current: `findFullOrganization`, `findTeamById`, and `listMembers` limit their result based on `membershipLimit`
Expected: Always take all valid memberships from the database into account

1.2.12

```bash
n/a
```

Backend, Client

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

Bekacru:
closes #3467 
    

---

## Summary by cubic
Added a membersLimit parameter to the getFullOrganization API, allowing clients to control how many organization members are returned.

- **New Features**
  - Supports custom member limits per request, overriding the default membershipLimit value.
  - Updated documentation to explain the new parameter.

### DIFF
--- a/packages/better-auth/src/plugins/organization/adapter.ts
+++ b/packages/better-auth/src/plugins/organization/adapter.ts
@@ -375,7 +375,7 @@ export const getOrgAdapter = <O extends OrganizationOptions>(
 				adapter.findMany<InferMember<O>>({
 					model: "member",
 					where: [{ field: "organizationId", value: org.id }],
-					limit: membersLimit || options?.membershipLimit || 100,
+                    limit: membersLimit ?? options?.membershipLimit ?? 100,
 				}),
 				includeTeams
 					? adapter.findMany<InferTeam<O>>({


santiagojuarez2715-maker:
Gracias, soy nuevo.

hakai2k:
### Related to 8a75cc3 and 26965c9

I think this doesn't work if we do 
`await authClient.organization.getFullOrganization({ query:{ membersLimit: 0 }})` 
It still retrieves 100 by default because the expression `limit: membersLimit ?? options?.membershipLimit ?? 100` is probably evaluating to `false` and defaulting to `100`?

hakai2k:
@Bekacru

Issue #3560, opened 2025-07-23
Upgrading / switching plans does not cancel original subscription


nowy:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Setup `better-auth` with 2 Stripe plans.
2. Subscribe to the 1st Stripe plan.
3. Upgrade to the 2nd Stripe plan.

### Current vs. Expected behavior

Following the steps of the previous section, I would expect there to be one entry in the database `subscription` table, and 1 entry in the Stripe subscription page for the user.

**What actually happens**

- There is one entry in the database `subscription` table with the updated `plan` ✅

<img width="1196" height="135" alt="Image" src="https://github.com/user-attachments/assets/2089d3c8-18ca-4cd8-bc78-535290d23658" />

- There are 2 subscriptions in Stripe

<img width="901" height="107" alt="Image" src="https://github.com/user-attachments/assets/5da32008-4da4-4ff8-ae22-f61cb2917877" />

### What version of Better Auth are you using?

1.3.3

### Provide environment information

```bash
- OS: Mac OS
- Browser: Irrelevant
- Using next.js
```

### Which area(s) are affected? (Select all that apply)

Package

### Auth config (if applicable)

```typescript
export const auth = betterAuth({
  database: prismaAdapter(prisma, {
    provider: "postgresql",
  }),
  session: {
    cookieCache: {
      enabled: true,
      maxAge: 5 * 60, // Cache duration in seconds
    },
  },
  emailVerification: {
    // ...
  },
  user: {
    // ...
  },
  account: {
    // ...
  },
  plugins: [
    admin(),
    organization({
      // ...
    }),

    emailOTP({
      // ...
    }),

    stripe({
      stripeClient: stripeInstance,
      stripeWebhookSecret: process.env.STRIPE_WEBHOOK_SECRET!,
      createCustomerOnSignUp: true,

      subscription: {
        enabled: true,
        plans: pricingData
          .filter((plan) => !!plan.stripeIds)
          .map((plan) => ({
            name: plan.name === "Semi-Pro" ? "Enthusiast" : plan.name,
            priceId: plan.stripeIds!.priceId,
            freeTrial: {
              days: 30,
            },
          })),
        getCheckoutSessionParams: async ({ user }) => {
          return {
            params: {
              allow_promotion_codes: true,
            },
          };
        },

        onSubscriptionComplete: async ({ subscription, plan }) => {
          try {
            const user = subscription.stripeCustomerId
[...]

PR #3583, opened 2025-07-24
fix(stripe): reuse existing trial subscription on upgrade and prevent duplicate subscriptions

tinkerer-shubh:
closes https://github.com/better-auth/better-auth/issues/3560
    

---

## Summary by cubic
Fixed an issue where upgrading a subscription could create duplicate Stripe subscriptions by reusing any existing active or trial subscription during the upgrade process.

- **Bug Fixes**
  - Checks for existing active or trialing subscriptions before creating a new one.
  - Prevents duplicate subscriptions when upgrading.


tinkerer-shubh:
@cubic-dev-ai  re-run the review

PR #3600, opened 2025-07-24
fix: organization allowUserToCreateOrganization user argument typing

TimurBas:
When using the organization plugin and if we have:
```ts
allowUserToCreateOrganization(user) {
  return user.role === "internalAdmin";
},
```
The user argument does not have `role` in its type definition. 
Looks like it just looks at `schema.ts` and uses the  `userSchema` const:
```ts
export const userSchema = z.object({
	id: z.string(),
	email: z.string().transform((val) => val.toLowerCase()),
	emailVerified: z.boolean().default(false),
	name: z.string(),
	image: z.string().nullish(),
	createdAt: z.date().default(() => new Date()),
	updatedAt: z.date().default(() => new Date()),
});
```
If used in conjuction with the admin plugin I would expect I could get the user's role to check whether it is a regular user or admin. If it is an internal admin role (a custom role) then I would allow organization creation otherwise not.

The jwt plugin's `definePayload` method deconstructs a `user` argument which also accepts any fields on the user.
```ts
definePayload?: (session: {
            user: User & Record<string, any>;
            session: Session & Record<string, any>;
        }) => Promise<Record<string, any>> | Record<string, any>;
```
Taken inspiration from that, obviously.

### DIFF
--- a/packages/better-auth/src/plugins/organization/types.ts
+++ b/packages/better-auth/src/plugins/organization/types.ts
@@ -25,7 +25,7 @@ export interface OrganizationOptions {
 	 */
 	allowUserToCreateOrganization?:
 		| boolean
-		| ((user: User) => Promise<boolean> | boolean);
+		| ((user: User & Record<string, any>) => Promise<boolean> | boolean);
 	/**
 	 * The maximum number of organizations a user can create.
 	 *


TimurBas:
Come to think about it.. I haven't checked whether we have the value available in the `allowUserToCreateOrganization` method.

TimurBas:
I tested it and it looks like we have the values in the method.

ping-maxwell:
Yo @TimurBas can you check why CI failing?

TimurBas:
@ping-maxwell Of course 

> 🦋  error Some packages have been changed but no changesets were found. Run `changeset add` to resolve this error.
> 🦋  error If this change doesn't need a release, run `changeset add --empty`.

I suppose I need to add a changeset? Haven't worked with it before. 
Please confirm.

TimurBas:
Nvm, I see it is a requirement, will add it later.
https://github.com/better-auth/better-auth/blob/main/CONTRIBUTING.md#development-workflow

petrbela:
Maybe I'm missing something but if `role` is a legitimate column on the users table, shouldn't it be added with `{ user: { additionalFields }}` and then the User type would be resolved properly? What's proposed here isn't really a solution, it's just bailing out on proper type safety.

TimurBas:
@petrbela AFAIK, the role column/property is inherited from the use of the admin plugin. Therefore, it looks like it is not a part of the base `user` table/schema definition as we see in https://github.com/TimurBas/better-auth/blob/4cff983e5c6e19100135fe93bd315f74de19f98c/packages/better-auth/src/types/models.ts#L92.

I agree that this looks like a common issue (as mentioned earlier with the `definePayload` method) which have been quick-fixed.

The ideal fix would be to infer the additionalFields on the user table dynamically, based on which plugins are used in the auth setup. This PR is more of a quick patch to silence the error rather than a full resolution.

PR #3601, opened 2025-07-24
fix(jwt): ensure alg is added to the jwks when generating via /token endpoint

elliottminns:
When using the jwt plugin I noticed there was an issue with the public key missing the alg field whenever the JWKS was created via the `/token` endpoint compared to when it was created using the `/jwks` endpoint 

For example:

`/api/auth/token`
```
apply=# SELECT public_key from jwks;
                                   public_key
---------------------------------------------------------------------------------
 {"crv":"Ed25519","x":"fi_qr9jILwhtL71fSwZad3RMrdbZqHCpd6NgCATXPEA","kty":"OKP"}
(1 row)
```

`/api/auth/jwks`
```
apply=# SELECT public_key from jwks;
                                          public_key
-----------------------------------------------------------------------------------------------
 {"alg":"EdDSA","crv":"Ed25519","x":"Apvgc4vNpmXtpUfjAxGlcqMeBW1FPVWGlcTNR8PQGZE","kty":"OKP"}
(1 row)
```

As you can see the `alg` field is created in the second example.

This PR resolves the issue, causing both endpoints to have the same outcome in the public_key structure.

For reference, below is how the `/jwks` endpoint adds the algo which is what I've mimicked in the fix.

https://github.com/better-auth/better-auth/blob/68f09f15d3ef3339ece01f39c9a51bebd8f97010/packages/better-auth/src/plugins/jwt/index.ts#L258-L271
    

---

## Summary by cubic
Fixed an issue where the public key generated by the /token endpoint was missing the alg field, so both /token and /jwks now include alg in the public key.

- **Bug Fixes**
 - Added alg to the public key structure in /token endpoint responses.

### DIFF
--- a/packages/better-auth/src/plugins/jwt/jwt.test.ts
+++ b/packages/better-auth/src/plugins/jwt/jwt.test.ts
@@ -85,7 +85,7 @@ describe("jwt", async (it) => {
 		const jwks = await client.jwks();
 
 		expect(jwks.data?.keys).length.above(0);
-		expect(jwks.data?.keys[0].alg).toBe("EdDSA")
+		expect(jwks.data?.keys[0].alg).toBe("EdDSA");
 	});
 
 	it("Signed tokens can be validated with the JWKS", async () => {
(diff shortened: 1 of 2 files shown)


elliottminns:
This appears to also be fixed in https://github.com/better-auth/better-auth/pull/3572 but considering that's a much larger PR this one might be a little easier to review an merge sooner.

If https://github.com/better-auth/better-auth/pull/3572 is merged first then this one can be closed

PR #3605, opened 2025-07-25
refactor: change apple type and name in order to support multiple audiences

dagyu:
In cross-platform applications like Flutter, the `aud` (audience) claim in Apple ID tokens can differ between platforms:

- On **iOS**, the audience is usually the app's **bundle identifier** (e.g., `com.example.myapp`)
- On **Android** (using web-based Sign in with Apple), the audience is the **Services ID** (e.g., `com.example.myapp.signin`)

Currently, `better-auth` define the audience in this way:

https://github.com/better-auth/better-auth/blob/f99305b29eb096186a0f85100e01f721d208f2f0/packages/better-auth/src/social-providers/apple.ts#L119

In this way I can't support the case where I need multiple string as audience.

To address this, I changed the name of`appBundleIdentifier` into `audiences` and the type from `string` to `string[]`. This makes it possible to support both audience values depending on which platform the user is authenticating from.
    

---

## Summary by cubic
Updated Apple Sign In to support multiple audience values, allowing cross-platform authentication for apps using both bundle identifiers and service IDs.

- **Refactors**
 - Renamed the option from `appBundleIdentifier` to `audience` and changed its type from `string` to `string[]`.

### DIFF
--- a/docs/content/docs/plugins/username.mdx
+++ b/docs/content/docs/plugins/username.mdx
@@ -185,6 +185,7 @@ const auth = betterAuth({
                 if (username === "admin") {
                     return false
                 }
+                return true
             }
         })
     ]


dagyu:
@Kinfe123 I didn't received any answer, please why you didn't consider this PR? Who needs multiple audience can't actually use your library without rewriting `verifyIdToken` function

Kinfe123:
Hey. It is already [here](https://github.com/better-auth/better-auth/pull/3710). There was a conflict that's why.

PR #3636, opened 2025-07-26
fix(username): add missing normalization

bortoz:


himself65:
I should add a test case before merging

PR #3648, opened 2025-07-27
Fix API key remaining initialization issue #3640

### LINKED ISSUE #3640: Bug: remaining defaults to refillAmount when null in create-api-key.ts

### To Reproduce

There is a bug in packages/better-auth/src/plugins/api-key/routes/create-api-key.ts 

In the code, the following line assigns remaining: `remaining: remaining || refillAmount || null,`

So setting the `remaining` as null to remove the cap on API usage causes it to set remaining to the refillAmount value.  This creates a key usage cap when there shouldn't be.

To reproduce, create an ApiKey with the `remaining` set to null as follows:

```
const apiKeyData = await auth.api.createApiKey({
          body: {
            name: SYSTEM_TITLE_ABRV,
            expiresIn: null, // seconds or never
            userId: newUser.user.id,
            prefix: API_KEY_PREFIX,
            metadata: metadata,
            // permissions,

            // Enable rate limit throttling
            rateLimitEnabled: true,
            rateLimitTimeWindow: RATE_LIMIT_TIME_WINDOW,
            rateLimitMax: RATE_LIMIT_MAX,
            remaining: null, // There is no cap on usage.
            // Refill the API key one request per second
            refillAmount: 1,
            refillInterval: 1000,
          },
        });
```

This will cause `remaining` to be set from null to 1.

Setting `remaining` to null should keep it null. 

1.3.3

```bash
Node 20
```

Backend

```typescript
Not relevant.
```

### Additional context

Tested on the latest version of better-auth.  
The bug is on line 422 of https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/api-key/routes/create-api-key.ts

eaoliver:
Minor bug fix to Closes #3640 related to the re-initialization of the `remaining` property of an API key.  The existing code was ignoring the user and default value of `null` and changing it to the value of `refillAmount`.  

Additional tests added, which are consistent with documentation.  All tests pass.  No updates to documentation.

### DIFF
--- a/.changeset/salty-banks-heal.md
+++ b/.changeset/salty-banks-heal.md
@@ -1,0 +1,11 @@
+---
+"better-auth": patch
+---
+
+Fixes API key remaining initialization issue #3640.
+This is a one line change to fix the initialiation of the remaining variable
+when creating a new API key. Setting remaining to null during the creation of a
+key to signify 'no cap' on key usage should propagate to the database. The
+current code is treating null as an invalid value and replacing it with the
+value of remaining. Setting the value to remaining introduces a cap on API key
+usage, which is incorrect behaviour.
--- a/packages/better-auth/src/plugins/api-key/api-key.test.ts
+++ b/packages/better-auth/src/plugins/api-key/api-key.test.ts
@@ -600,6 +600,72 @@ describe("api-key", async () => {
 		expect(apiKey.remaining).toEqual(remaining);
 	});
 
+	it("should create API Key with remaining explicitly set to null", async () => {
+		const apiKey = await auth.api.createApiKey({
+			body: {
+				remaining: null,
+				userId: user.id,
+			},
+		});
+
+		expect(apiKey).not.toBeNull();
+		expect(apiKey.remaining).toBeNull();
+	});
+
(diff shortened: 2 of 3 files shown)


Kinfe123 (approved):

.changeset/salty-banks-heal.md
```diff
@@ -0,0 +1,11 @@
...
+"better-auth": patch
+---
+
+Fixes API key remaining initialization issue #3640.
+This is a one line change to fix the initialiation of the remaining variable
+when creating a new API key. Setting remaining to null during the creation of a
+key to signify 'no cap' on key usage should propagate to the database. The
+current code is treating null as an invalid value and replacing it with the
+value of remaining. Setting the value to remaining introduces a cap on API key
+usage, which is incorrect behaviour.
```
Bekacru, line 11: fix(api key): key creation bug where remaining was incorrectly defaulting to a capped value instead of allowing null

Bekacru:
```suggestion
fix(api key): key creation bug where remaining was incorrectly defaulting to a capped value instead of allowing null
```

PR #3655, opened 2025-07-27
feat: expose verify id token on generic oauth

ThallesP:
Recently, I had to integrate a generic OAuth2 system that uses identity tokens for validations (https://web3auth.io/docs/authentication/id-token). However, the generic OAuth options did not expose the `verifyIdToken` in the available options. This PR exposes this functionality so it can be passed.

Additionally, the support for identity tokens could be improved for generic OAuth. I am considering adding an option to accept a JWKS endpoint (instead of only accepting a discovery URL) and also an option to decode the token directly into the `getUserInfo` without requiring installation of a full JWT library. This would allow developers to input the necessary `userInfo` from the `idToken`'s payload.

If this approach makes sense, please let me know so I can open another PR. For now, I wanted to implement something quickly as I need this functionality for my work.

### DIFF
--- a/docs/content/docs/plugins/generic-oauth.mdx
+++ b/docs/content/docs/plugins/generic-oauth.mdx
@@ -182,6 +182,8 @@ interface GenericOAuthConfig {
 
 **mapProfileToUser**: (Optional) A function to map the provider's user profile to your app's user object. Useful for custom field mapping or transformations.
 
+**verifyIdToken**: (Optional) A function to verify the ID token returned by the provider. Takes the token string and optional nonce as parameters and returns a Promise<boolean>.
+
 **authorizationUrlParams**: (Optional) Additional query parameters to add to the authorization URL. These can override default parameters.
 
 **disableImplicitSignUp**: (Optional) If true, disables automatic sign-up for new users. Sign-in must be explicitly requested with sign-up intent.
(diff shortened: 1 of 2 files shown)


docs/content/docs/plugins/generic-oauth.mdx
```diff
@@ -244,6 +246,60 @@ genericOAuth({
...
+#### Sign In with ID Token
+
+To sign in with an OAuth provider using an ID Token, you can use the `signIn.social` function to pass the ID Token directly.
+
+<Callout>
+ If an ID token is provided no redirection will happen, and the user will be signed in directly.
+</Callout>
+
+```ts title="auth-client.ts"
+await authClient.signIn.social({
```
Bekacru, line 294: I think we should change this to `signIn.oauth2` and implement ID token sign‑in within the generic OAuth plugin.

ThallesP: > I think we should change this to `signIn.oauth2` and implement ID token sign‑in within the generic OAuth plugin.

@Bekacru so basically deprecate the `.social` and add `.oauth2`? Also agree, `social` keyword makes sense in a apple/google context. Should I go through with it?

And what do you mean by `implement ID token sign‑in within the generic OAuth plugin`?

himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #3660, opened 2025-07-28
feat(sso): defaultSSO options and ACS endpoint

Kinfe123:


natetewelde:
@Kinfe123 i don’t feel like my PR was a duplicate of this? We are doing two separate things?

the issue I was coming up against was with mappings being required outside of their respective configs (even though that’s where the callbacks are looking for them), you lose the capability to use first name, last name mappings because its defaulting to oidc mappings. When you make it so that mappings are defined inside of their respective configs you can enhance the type safety and allow for proper mappings. Also mappings don’t have good fallbacks. NameID is often times just the email and since there is no fallback for that the account provisioning/creation throws null errors when trying to create an account for the user

Kinfe123:
Not the tittle but I have included this issue in my implementation. Will be moved to ready for review soon.

Kinfe123:
@natetewelde  yeah. i have noticed your PR. thats why i closed it. since it is already been in mind while opening this PR.

natetewelde:
any update on this being added? SAML is still unusable at the moment

flowhorn:
Hey @Kinfe123, I tried your approach, I almost got it running, when running the callback callbackSSO Request i get an error: error=invalid_provider&error_description=provider not found. Is this function using the defaultProviders?

Kinfe123:
on the implementation the callback also handles the defaultSSO config that you have added. can you please send me the your defaultSSO config or your sso config ?

flowhorn:
> on the implementation the callback also handles the defaultSSO config that you have added. can you please send me the your defaultSSO config or you sso config ?

Thank you for your quick response. I got it to work using the generic-oauth plugin. It is my mistake, that I missinterpreted, that I would use the sso plugin for this. So I think my comment is redundant. Apprently I can't share my configuration as it contains private secrets. But thank you!

natetewelde:
What is the hold up on getting this pushed? I’ve been waiting to try and use better-auth for my company’s SAML SSO setup and got excited when 1.3 dropped saying it was ready. Just curious what the delay has been for over 2 months now?

PR #3662, opened 2025-07-28
fix(anonymous): prevent false positive error on first anonymous sign-in

### LINKED ISSUE #3658: Anonymous plugin's 'after' hook incorrectly rejects successful user creation requests (Auto anonymous user sign in fails)

### Environment Setup
- **Framework**: Next.js 15 with React 19
- **Database**: Convex with @convex-dev/better-auth integration
- **Better Auth Version**: 1.2.12
- **@convex-dev/better-auth Version**: 0.7.11

### 1. Create Better Auth Configuration

```typescript
// lib/auth.ts
import { convexAdapter } from "@convex-dev/better-auth";
import { convex } from "@convex-dev/better-auth/plugins";
import { anonymous } from "better-auth/plugins"

export const createAuth = (ctx: GenericCtx) =>
  betterAuth({
    baseURL: process.env.SITE_URL,
    database: convexAdapter(ctx, betterAuthComponent),
    trustedOrigins: [process.env.SITE_URL as string],
    plugins: [
      convex(),
      anonymous({
          // Account linking logic here
        }
      }),
    ],
  });
```

### 2. Create Anonymous Sign-in Component

```typescript
// app/components/auth/anonymous-sign-in.tsx
'use client';

import { useEffect, useRef } from 'react';
import { authClient } from '@/lib/auth-client';

export function AnonymousSignIn() {
  const attemptedAnon = useRef(false);

  useEffect(() => {
    if (!attemptedAnon.current) {
      attemptedAnon.current = true;

      const signInAnonymously = async () => {
        try {
          const result = await authClient.signIn.anonymous();
          console.log('Success:', result);
        } catch (error) {
          console.error('Error:', error);
          // Error: "Anonymous users cannot sign in again anonymously"
        }
      };

      signInAnonymously();
    }
  }, []);

  return <div>Loading...</div>;
}
```

### 3. Reproduce the Issue

1. Open application in incognito/private browsing mode
2. Navigate to page that renders `AnonymousSignIn` component
3. Observe network request: `POST /api/auth/sign-in/anonymous` returns `400 Bad Request`
4. Check database: Anonymous user record was successfully created
5. Reload the page: Application works perfectly (no new sign-in attempt)

**Current Behavior:**
When attempting to sign in anonymously for the first time:
1. Server successfully creates anonymous user and session
2. Client receives `400 Bad Request` with error: "Anonymous users cannot sign in again anonymously"
3. User creation actually succeeds (visible in database)
4. Page reload works perfectly (because session cookie exists)

**Expected Behavior:**
- Anonymous sign-in should succeed on first attempt
- Client should receive success response and be properly authenticated
- No page reload should be required

1.2.12

```bash
- **OS**: macOS
- **Browser**: Chrome (tested in incognito mode)
- **Package Manager**: Bun
- **Better Auth**: 1.2.12
- **@convex-dev/better-auth**: 0.7.11
- **Integration**: Convex database with Better Auth adapter
```

Client, Backend

```typescript
export const createAuth = (ctx: GenericCtx) =>
  betterAuth({
    baseURL: process.env.SITE_URL,
    database: convexAdapter(ctx, betterAuthComponent),
    trustedOrigins: [process.env.SITE_URL as string],
    socialProviders: {
      google: { 
        prompt: "select_account", 
        clientId: process.env.AUTH_GOOGLE_ID as string,
        clientSecret: process.env.AUTH_GOOGLE_SECRET as string,
      },
    },
    account: {
      accountLinking: {
        enabled: true, 
        allowDifferentEmails: true,
      }
    },
    plugins: [
      convex(),
      anonymous({
          // Account linking implementation
        }
      }),
    ],
  });
```

### Bug Location
The issue is located in the Better Auth anonymous plugin's `after` hook:
- **File**: `packages/better-auth/src/plugins/anonymous/index.ts`

### Hook Execution Flow (Problematic)
```javascript
// Simplified version of the problematic code
handler: createAuthMiddleware(async (ctx) => {
  const setCookie = ctx.context.responseHeaders?.get("set-cookie");
  const sessionCookie = parseSetCookieHeader(setCookie || "").get(sessionTokenName)?.value.split(".")[0];
  
  if (!sessionCookie) {
    return; // No session cookie, continue
  }
  
  const session = await getSessionFromCtx(ctx, { disableRefresh: true });
  
  if (!session || !session.user.isAnonymous) {
    return; // No existing anonymous session, continue  
  }
  
  if (ctx.path === "/sign-in/anonymous") {
      message: ERROR_CODES.ANONYMOUS_USERS_CANNOT_SIGN_IN_AGAIN_ANONYMOUSLY
    });
  }
})
```

### The Timing Bug Explained

1. **User Creation Succeeds**: Anonymous sign-in endpoint successfully creates user and session
2. **Session Cookie Set**: Response headers include the new session cookie
3. **After Hook Executes**: The `after` hook runs and detects the newly created session cookie
4. **Session Detection**: Hook calls `getSessionFromCtx()` which finds the session (just created in this request)
5. **Incorrect Rejection**: Since `session.user.isAnonymous` is true and path is `/sign-in/anonymous`, the hook throws the error
6. **Request Fails**: The hook rejects the very request that successfully created the anonymous user

### Why Page Reload Works
- On reload, the session cookie exists from the previous (successful) creation
- The component detects an existing session and doesn't attempt anonymous sign-in
- No new `/sign-in/anonymous` request is made

### Client-Side Debug Output
```
[anon-1753664152484-tshj8s9c7] About to call authClient.signIn.anonymous()
POST http://localhost:3000/api/auth/sign-in/anonymous 400 (Bad Request)
Error: "Anonymous users cannot sign in again anonymously"

authClient.getSession(): {data: null, error: null}
Auth-related cookies: []
```

### Server-Side Debug Output
```
🔍 SERVER DEBUG - onCreateUser called: {
  userIsAnonymous: true,
  userName: null,
  userEmail: null,
  timestamp: "2025-07-28T02:05:43.123Z",
  hasAuthIdentity: true  // ← Shows server detects auth identity
}
```

**Key Insight**: Server successfully creates the user (`onCreateUser` called) but client receives error response due to the hook's timing issue.

ajanraj:
**Fixes #3658**

## Summary
Resolves timing issue where the anonymous plugin incorrectly rejected first-time anonymous sign-in attempts, causing users to receive 400 “Anonymous users cannot sign in again anonymously” errors despite successful user creation.

## Root Cause
The issue occurred when apps automatically sign in anonymous users (e.g., in React useEffect on page load). The original implementation checked for existing sessions in an `after` hook, which couldn't distinguish between:
- Sessions created in the current request (`ctx.context.newSession`)
- Pre-existing sessions from previous requests.

## Solution
Moved the session check from the `after` hook to the sign-in endpoint handler itself, where it properly detects existing anonymous sessions **before** attempting to create new users. This approach is:
- ✅ More direct and reliable
- ✅ Eliminates timing dependencies
- ✅ Follows existing authentication patterns.

## Changes
- **Core Fix**: Added session check in `/sign-in/anonymous` endpoint handler (lines 124-133)
- **Comprehensive Tests**: Added test cases for both first-time success and subsequent rejection scenarios
- **Maintained Compatibility**: All existing functionality preserved, only fixed the false positive case

## Test Results
```
✓ should sign in anonymously
✓ link anonymous user account  
✓ should link in social sign on
✓ should work with generateName
✓ should not reject first-time anonymous sign-in
✓ should reject subsequent anonymous sign-in attempts once signed in
```

## Impact
- **Fixes**: First-time anonymous sign-in now works correctly in automatic scenarios (useEffect, app load)
- **Maintains**: Security - still prevents actual repeat anonymous sign-ins
- **Improves**: Developer experience - no more confusing false positive errors

## Files Changed
- `packages/better-auth/src/plugins/anonymous/index.ts` - Core fix implementation
- `packages/better-auth/src/plugins/anonymous/anon.test.ts` - Added comprehensive test coverage
- `.changeset/slimy-months-yawn.md` - Changeset for version tracking

## Breaking Changes
None - this is a pure bug fix that maintains all existing behavior while fixing the false positive case.
    

---

## Summary by cubic
Fixed a bug where first-time anonymous sign-in attempts were incorrectly rejected with a 400 error, even though the user was new.

- **Bug Fixes**
  - Moved the session check to the sign-in endpoint to properly allow first-time anonymous sign-ins.
  - Added tests to cover both successful first sign-ins and correct rejection of repeat attempts.

### DIFF
--- a/.changeset/slimy-months-yawn.md
+++ b/.changeset/slimy-months-yawn.md
@@ -1,5 +1,0 @@
----
-"better-auth": minor
----
-
-Fix timing issue where anonymous plugin incorrectly rejected automatic anonymous sign-in when used on app load


Bhadboiab (approved):
Yeah

Kinfe123 (approved):

motz0815:
I just tested your PR using PkgPR's npm package.
It seems as though I can now log in multiple times, and the session itself doesn't seem to get populated (even after a page refresh)

Here's my page.tsx:
<details><summary>page.tsx</summary>
<p>

```tsx
"use client"

import { Button } from "@/components/ui/button"
import { authClient } from "@/lib/auth-client"
import { toast } from "sonner"

export default function Home() {
    const { data: session } = authClient.useSession()

    return (
        <div>
            <pre>{JSON.stringify(session, null, 2)}</pre>

            <Button
                onClick={() => {
                    if (!session) {
                        authClient.signIn.anonymous({
                            fetchOptions: {
                                onSuccess: () => {
                                    toast.success("Signed in")
                                },
                                onError: (e) => {
                                    toast.error(e.error.message)
                                },
                            },
                        })
                    } else {
                        toast.error("You are already signed in")
                    }
                }}
            >
                Anon Sign in
            </Button>
        </div>
    )
}
``` 

</p>
</details>

`session` is always null, but after each click on "Anon Sign in", I receive the success callback toast.

Cookies are correctly set, but there's nothing in the session. Note that I'm also using convex.
In the database, no user is created. (That's probably why there is no session to show)

ajanraj:
Hey, that's really strange - for me it worked sometimes but other times I have to manually refetch the session after the success callback toast, then it worked perfectly. The user is getting populated in the db on my end though and on reload it worked when I tested this.

But yeah, I actually moved back to convex auth since I was running into a lot of issues, especially with anon users. The inconsistency was becoming too much of a headache.

Kinfe123:
i will address this issue.

Kinfe123:
The timing is actually works but this is not the issue that is related to the anon users but the reactivity. can you please share some repro of your setup that we can try out may be `page.tsx` only file is not enough on debugging those kind of issue.

motz0815:
I found out that the issue also persists on the latest better-auth (1.3.4), so I'm assuming the issue stems from `@get-convex/better-auth`, because it only supports better-auth up to version 1.2.12.

I guess I'll just have to wait for the compatibility to get better.

motz0815:
Any update on this?

motz0815:
What's preventing this PR from getting merged?

PR #3675, opened 2025-07-28
fix: changelog formatting

Kinfe123:
- **update**
- **update**

    

---

## Summary by cubic
Improved changelog formatting by cleaning up commit messages and removing unnecessary prefixes for better readability.

- **Bug Fixes**
  - Stripped commit prefixes like "fix:" and "feat:" from changelog entries.
  - Fixed spacing and formatting issues in changelog content.

### DIFF
--- a/docs/app/changelogs/_components/default-changelog.tsx
+++ b/docs/app/changelogs/_components/default-changelog.tsx
@@ -25,19 +25,122 @@ const ChangelogPage = async () => {
 		}[]
 	>("https://api.github.com/repos/better-auth/better-auth/releases");
 
-	const messages = releases
+	function extractVersion(title: string): string {
+		if (title.slice(2).includes("@")) {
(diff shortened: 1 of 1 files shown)


Bekacru:
Screenshot 2025-07-28 at 12 44 48 PM
We want to keep the package name that got updated in the title so we don't have 2 same version with no description. And actually it might be a better idea to merge all changes under 1.3.4 (like we've been doing and list them per pkg instead)

Kinfe123:
I have noticed that. and i was trying to do some kind subheading telling what package is patched and merging it since it kinda of create a confusion. but yeah will update you on this.

Bekacru:
Screenshot 2025-07-29 at 10 00 59 PM

seems like "Fix typo on" is cut off

Bekacru:
@Kinfe123 could you revert this to work with the previous way of changelog format instead now that we removed changeset

PR #3676, opened 2025-07-28
fix: refactored mappings to be inside of respective configs for types…

natetewelde:
## Summary by cubic
Refactored SSO attribute mappings to be defined inside their respective config objects for better type safety and clarity. Improved SAML account provisioning to handle cases where standard fields like nameID may not be present.

- **Refactors**
  - Moved OIDC and SAML mapping definitions into their config objects.
  - Updated docs and tests to match the new structure.
  - Enhanced SAML user extraction logic to use alternative fields if nameID is missing.


Kinfe123:
close as duplicate via - #3660

PR #3679, opened 2025-07-28
fix(client): prevent proxy from exposing `then` property to avoid promise-like behavior (#3657)

### LINKED ISSUE #3657: `createAuthClient` in a Vitest fixture hangs Vitest indefinitely due to `then` property always available

### To Reproduce

Try to put better-auth client in a [Vitest fixture](https://vitest.dev/guide/test-context.html#test-extend):
```typescript
import {test as baseTest} from 'vitest';
import {createAuthClient} from 'better-auth/client';

export const test = baseTest.extend<{
    authClient: ReturnType<typeof createAuthClient>
}>({
   authClient: ({}, use) => use(createAuthClient({
        baseURL: process.env.BACKEND_URL,
        basePath: '/auth'
    }))
});
```

Any test using the fixture hangs/timeouts never actually stepping into the test itself:
```typescript
import {test} from '<the file with the fixture above>';

test('Standard sign-up is disallowed', ({authClient}) => {
});
```

1.3.4

Doesn't matter.

Client

Probably doesn't matter. If needed, I can set up a reproduction repo.

### Additional context

The issue is basically this: `Promise.resolve(createAuthClient())` returns a never-resolving Promise instead of a resolved one because **`createAuthClient()` returns a thenable**.

1. [Vitest internally resolves a Promise with the fixture value](https://github.com/vitest-dev/vitest/blob/a4df683d108369ce274052beea87b51b87c8c497/packages/runner/src/fixture.ts#L286), the value being the client in our case
2. However, `createAuthClient` effectivery returns a thenable due to [createProxy's `get`](https://github.com/better-auth/better-auth/blob/3afb25a780dc28a2f5b10a557f711d85864d22f4/packages/better-auth/src/client/proxy.ts#L41) always returning something even when the path doesn't exist on the client
3. [One funny Promise behaviour](https://medium.com/@davidcallanan/a-peculiar-promises-and-proxy-bug-that-cost-me-5-hours-javascript-3a11e1fcd713) happens, and instead of the client Vitest gets a new Promise which will never be resolved

I guess `createAuthClient` should just return `undefined` for any access to `authClient.then`.

<sub>This issue wasted my workday 🫤.</sub>

Aditya-ingole21:
Summary  
Fixes #3657  

The client proxy was unintentionally exposing a `then` property, which caused Vitest to treat the proxy as a Promise.  
This resulted in unexpected test behavior.

Changes Made  
- Updated proxy implementation to ensure `.then` is not exposed  
- Added a regression test:

  it("should not expose a 'then' property on the proxy", async () => {
    const proxy = (client as any).test;
    expect(proxy.then).toBeUndefined();
  });

- Removed unused test artifact (state.txt)

Why  
This fix ensures the proxy cannot be mistaken for a Promise-like object,  
preventing runtime and test inconsistencies.

Checklist  
- Bug fix implemented  
- New test added and passing

    

---

## Summary by cubic
Fixed the client proxy so it no longer exposes a `then` property, preventing it from being treated as a Promise and avoiding unexpected test behavior.

- **Bug Fixes**
  - Updated the proxy to return `undefined` for `then`.
  - Added a regression test to ensure `then` is not exposed.
  - Removed an unused test artifact.


Aditya-ingole21:
Hi @Bekacru, this PR is ready for review. Could you please take a look when you get a chance?

packages/better-auth/src/client/proxy.ts
```diff
@@ -39,6 +39,10 @@ export function createDynamicPathProxy<T extends Record<string, any>>(
 	function createProxy(path: string[] = []): any {
 		return new Proxy(function () {}, {
 			get(target, prop: string) {
+
+				if (prop === "then") {
```
himself65: it is possible that user has a path called `then`. IMO this is vitest issue. If you want workaround this, maybe you can wrap the client into an object

himself65: https://github.com/vitest-dev/vitest/issues/5122#issuecomment-2406488169

> We think this should be better handled by user/package side, so closing as wontfix.

KazimirPodolski: Well, one could argue that user/package side here is actually better-auth.

Also not only Vitest could be affected, but anything `await`ing the client.

himself65: make sense. I think a better way is we check do a special check for 'then' whether it exist in user's paths. if not, we can easily returns undefined without breaking

KazimirPodolski: Bonus points for throwing a warning when 'then' path exists. It's a very weird _top-level_ path anyway.

IMO the amount of users who will face Vitest issue >>> the amount of users with a `then` top-level path together with `signIn`, `signUp`,...

himself65:
```patch
Subject: [PATCH] feat: allow pass trusted origin as function
---
Index: packages/better-auth/src/client/proxy.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/packages/better-auth/src/client/proxy.ts b/packages/better-auth/src/client/proxy.ts
--- a/packages/better-auth/src/client/proxy.ts	(revision 91cae8aa1e567034aacc11e316508ca8fed52393)
+++ b/packages/better-auth/src/client/proxy.ts	(date 1754071259541)
@@ -29,6 +29,8 @@
 	matcher: (path: string) => boolean;
 };
 
+const kSkipSymbol = Symbol("skip");
+
 export function createDynamicPathProxy<T extends Record<string, any>>(
 	routes: T,
 	client: BetterFetch,
@@ -38,9 +40,13 @@
 ): T {
 	function createProxy(path: string[] = []): any {
 		return new Proxy(function () {}, {
-			get(target, prop: string) {
+			get(target, prop: string, receiver) {
 				const fullPath = [...path, prop];
 				let current: any = routes;
+				if (prop === 'then' && Reflect.get(target, prop, receiver) === kSkipSymbol) {
+					// If the property is 'then', we return undefined to avoid treating it as a promise
+					return undefined;
+				}
 				for (const segment of fullPath) {
 					if (current && typeof current === "object" && segment in current) {
 						current = current[segment];
@@ -103,5 +109,14 @@
 			},
 		});
 	}
-	return createProxy() as T;
+	const proxy = createProxy() as T;
+	// patch root `then` method to avoid be treated as a promise, this is a special case for some downstream libraries like vitest
+	if (!('then' in knownPathMethods)) {
+		Object.defineProperty(proxy, 'then', {
+			value: kSkipSymbol,
+			configurable: true,
+			enumerable: false
+		})
+	}
+	return proxy;
 }
```

My suggestion would sth like

himself65:
ok, now I join better auth. let's pick this PR up

PR #3680, opened 2025-07-28
fix(api-key): is-username-available endpoint should also validate usernames

ping-maxwell:
Adds username validation in the `isUsernameAvailable` endpoint - such as username length & custom validators
    

---

## Summary by cubic
The isUsernameAvailable endpoint now checks if usernames meet length and custom validation rules before confirming availability.

- **Bug Fixes**
 - Added checks for minimum and maximum username length.
 - Applied custom username validators to ensure valid input.

### DIFF
--- a/.changeset/wild-pianos-join.md
+++ b/.changeset/wild-pianos-join.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Fixed username plugins isUsernameAvailable endpoint to run validation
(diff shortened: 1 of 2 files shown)


.changeset/wild-pianos-join.md
```diff
@@ -0,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Fixed username plugins isUsernameAvailable endpoint to run validation
```
Bekacru: let's use standard format from this on:

fix(username): fixed username plugins isUsernameAvailable endpoint to run validation
```suggestion
fix(username): `isUsernameAvailable` should run validation
```

PR #3696, opened 2025-07-30
fix(stripe): fallback to subscription id instead of picking the first if subId is provided

### LINKED ISSUE #3690: Regression: Active subscription detection may select wrong subscription if multiple are present

### To Reproduce

**Description**
Since commit c2fb1aa3, the logic for retrieving the active subscription has changed. Previously, the code would strictly look for the subscription matching subscriptionToUpdate?.stripeSubscriptionId or ctx.body.subscriptionId among active subscriptions.
Now, it fetches all subscriptions with status active or trialing, and if no subscriptionToUpdate?.stripeSubscriptionId is provided, it simply picks the first result (.find(... ? sub.id === ... : true)).

**Steps to Reproduce**

Create several active/trialing subscriptions for the same Stripe customer.
Call the feature without providing a subscriptionToUpdate?.stripeSubscriptionId.
Notice that the chosen subscription may not be the expected one.

```js
const activeSubscriptions = await client.subscriptions
  .list({
    customer: customerId,
  })
  .then((res) =>
    res.data.filter(
      (sub) => sub.status === "active" || sub.status === "trialing",
    ),
);

const activeSubscription = activeSubscriptions.find((sub) =>
  subscriptionToUpdate?.stripeSubscriptionId
    ? sub.id === subscriptionToUpdate?.stripeSubscriptionId
    : true, // <== Here the first subscription is selected
  );
```

If a customer has several active or trialing subscriptions and no subscriptionToUpdate?.stripeSubscriptionId is provided, the current logic just picks the first subscription returned by Stripe.
This can result in selecting the wrong subscription, leading to unexpected behavior when updating or managing subscriptions.

1.3.4

```bash
- OS Mac OS
- Browser Chrome
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true,
    organization(),
    stripe(),
  },
});
```

### Additional context

_No response_

Bekacru:
closes #3690 
    

---

## Summary by cubic
Updated Stripe logic to use the provided subscription ID as a fallback when selecting the active subscription, instead of always picking the first one. This ensures the correct subscription is used when a specific ID is given.

### DIFF
--- a/.changeset/angry-items-boil.md
+++ b/.changeset/angry-items-boil.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/stripe": patch
+---
+
+should fallback to subscription id when it's present in the body
--- a/packages/stripe/src/index.ts
+++ b/packages/stripe/src/index.ts
@@ -328,8 +328,9 @@ export const stripe = <O extends StripeOptions>(options: O) => {
 						),
 					);
 				const activeSubscription = activeSubscriptions.find((sub) =>
-					subscriptionToUpdate?.stripeSubscriptionId
-						? sub.id === subscriptionToUpdate?.stripeSubscriptionId
+					subscriptionToUpdate?.stripeSubscriptionId || ctx.body.subscriptionId
+						? sub.id === subscriptionToUpdate?.stripeSubscriptionId ||
+							sub.id === ctx.body.subscriptionId
 						: true,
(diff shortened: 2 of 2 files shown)


joe10833 (changes requested):
packages/stripe/src/index.ts

PR #3706, opened 2025-07-30
feat/telemetry-support

frectonz:


packages/better-auth/src/__snapshots__/init.test.ts.snap
```diff
@@ -116,7 +123,7 @@ exports[`init > should match config 1`] = `
       "readonly": false,
     },
     "plugins": [],
-    "secret": "better-auth-secret-123456789",
+    "secret": "MKHwop6MjXXscHqrlKxfkmj2ypxNeTLW",
```
Bekacru: why is this changed?

packages/better-auth/src/config/index.ts
```diff
@@ -0,0 +1,80 @@
...
+
+import { loadConfig } from "./load-config";
+import { getConfigDir } from "./config-dir";
+
+interface GlobalConfigOptions {
+	name?: string;
+}
+type RealizedGlobalConfigOptions = Required<GlobalConfigOptions>;
+
+export class GlobalConfig {
```
Bekacru: lets avoid using class unless it's very necessary

packages/better-auth/src/telemetry/index.ts
```diff
@@ -0,0 +1,336 @@
...
+	] as any[];
+
+	sendDeviceState(plugins, authConfig);
+}
+
+async function sendDeviceState(plugins: { name: string }[], config?: any) {
+	if (!isTelemetryEnabled()) return;
+
+	try {
+		const telemetryId = getTelemetryId();
```
Bekacru: telemetry id should probably be drivered from base url hash or something instead of an env variable

packages/better-auth/src/telemetry/index.ts
```diff
@@ -0,0 +1,336 @@
...
+			nodeVersion: typeof process !== "undefined" ? process.version : undefined,
+			runtime: getRuntime(),
+			framework: detectFramework(),
+			database: detectDatabaseType(),
+			authConfig: config
+				? JSON.stringify(sanitizeAuthConfig(config))
+				: undefined,
+		};
+
+		await fetch(TELEMETRY_ENDPOINT, {
```
Bekacru: use betterFetch instead

Bekacru: for all fetch instances

packages/cli/src/commands/telemetry.ts
```diff
@@ -0,0 +1,25 @@
+import { Command } from "commander";
+import { GlobalConfig } from "../../../better-auth/src/config";
+import { TELEMETRY_CONFIG_KEY } from "../../../better-auth/src/telemetry/config-key";
+
+export const telemetry = new Command("telemetry").action(() =>
```
Bekacru: disabling telemetry should probably done through an env variable. I don't see the need for the CLI

Bekacru: that said maybe we should display in one of the CLI commands or post install that they can disable this behavior by passing DISBLAE_AUTH_TELEMETRY=1 or something

Bekacru:
we should also have docs about it

packages/better-auth/src/__snapshots__/init.test.ts.snap
frectonz: I had set a `BETTER_AUTH_SECRET` env var for another project, that may have been overriding it, i have fixed it now.

packages/better-auth/src/config/index.ts
frectonz: I have updated to use a `createConfig` function instead now.

packages/cli/src/commands/telemetry.ts
frectonz: I took they idea of having the CLI from Astro's telemetry implementation. (https://astro.build/telemetry/) Some users may want a way to disable telemetry for all better auth projects globally.

They can also still disable telemetry on a per project basis with `BETTER_AUTH_TELEMETRY` env var.

packages/better-auth/src/telemetry/index.ts
frectonz: We are now using better-fetch to send telemetry events.

packages/better-auth/src/telemetry/index.ts
frectonz: The telemetry ID is an anonymous ID meant to be used to track unique installations of better auth per device. I am using it to mark that this user is known and has this `telemetryID`.

We could use another anonymous ID. We can call that a `projectId` to track unique projects using better auth. We could use a hash of the `baseUrl` for that.

frectonz: I have added support for a `projectId`.

packages/better-auth/src/telemetry/index.ts
frectonz: Nice with better fetch we have retries 🎉

packages/better-auth/package.json
```diff
@@ -689,7 +689,10 @@
     "@simplewebauthn/browser": "^13.0.0",
     "@simplewebauthn/server": "^13.0.0",
     "better-call": "catalog:",
+    "ci-info": "^4.3.0",
```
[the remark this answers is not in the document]
frectonz: are you sure `ci-info` doesn't seem to be causing an issue in the build

frectonz:
@cubic-dev-ai generate a summary of the pr

frectonz:
This is all the data we are currently collecting.

- Detect Auth Config
  - Sanitized version of the user's auth config
  - A list of all the plugins that have been enabled
- Detect Better Auth
  - The version of better auth being used
- Detect Database Type
  - The database the project is using
  - The version of the database driver being used
- Detect Framework
  - The framework the project is built on
  - The version number of the framework
- Detect Production (if the project is in prod)
- Detect Project Info
  - Is the project using Git
  - Anonymous project ID
  - The name and version of the package manger being used
- Detect Runtime
  - The JS runtime being used
  - The version of the JS runtime
- Detect System Info
  - The OS
  - The CPU architecture
  - The CPU count
  - The CPU model
  - Total Memory
  - Check if the project is running in CI
  - Check if the project is running in WSL
  - Check if the project is running in Docker
  - Check if the project is running in a TTY

frectonz:
### Configuring Telemetry Behaviour

#### 1. Via Auth config

```typescript
betterAuth({
  telemetry: {
    enabled: true,
    address: "https://custom.telemetry.com/track"
  }
});
```

#### 2. Via CLI

This will control telemetry globally for all installations of better auth.

Enable Telemetry
```
npx @better-auth/cli telemetry enable
```

Disable Telemetry
```
npx @better-auth/cli telemetry disable
```

#### 2. Via Environment Variables

Enable Telemetry
```
BETTER_AUTH_TELEMETRY=1
```

Disable Telemetry
```
BETTER_AUTH_TELEMETRY=0
```

Enable Telemetry Debug
_This means that instead of sending the collected info to the address specified, the collected info will be logged._
```
BETTER_AUTH_TELEMETRY_DEBUG=1
```

.changeset/full-hotels-serve.md
```diff
@@ -0,0 +1,6 @@
+---
+"better-auth": patch
+"@better-auth/cli": patch
+---
+
+Initial telemetry implementation
```
Bekacru: I don't think we need to say initial. Something like `feat: telemetry` or something should suffice

packages/better-auth/src/config/index.ts
```diff
@@ -0,0 +1,80 @@
...
+import { loadConfig } from "./load-config";
+import { getConfigDir } from "./config-dir";
+
+export interface GlobalConfigOptions {
+	name?: string;
+}
+
+type RealizedGlobalConfigOptions = Required<GlobalConfigOptions>;
+
+export async function createGlobalConfig(options?: GlobalConfigOptions) {
```
Bekacru: I honestly don't see the point of storing this in a global config, assuming we're just a library and not a full-on web framework. It's mostly useful for local dev, unless you're also running a VPS in production. For serverless, since there's no persistence, this seems pretty useless (I'm assuming).

Bekacru: But I haven't checked. If everyone does this, might make sense

packages/better-auth/src/types/options.ts
```diff
@@ -1046,4 +1046,21 @@ export type BetterAuthOptions = {
...
+		 *
+		 * @default true
+		 */
+		enabled?: boolean;
+		/**
+		 * Telemetry endpoint to send events to
+		 *
+		 * @default - "https://telemetry.better-auth.com/v1/track"
+		 */
+		endpoint?: string;
```
Bekacru: we shouldn't have to expose this as an option

packages/better-auth/src/telemetry/index.ts
```diff
@@ -0,0 +1,95 @@
...
+
+		return telemetryEndpoint({
+			event,
+			payload,
+			anonymousId: await anonymousId(),
+		});
+	};
+
+	const report = async () => {
+		if (!(await isEnabled())) return;
```
Bekacru: let's not await inline for if statements

packages/better-auth/src/telemetry/index.ts
```diff
@@ -0,0 +1,95 @@
...
+
+		telemetryId = await config.getWithFallback(TELEMETRY_ID_CONFIG_KEY, () =>
+			generateId(32),
+		);
+
+		return telemetryId;
+	};
+
+	const publish = async (event: string, payload: any) => {
+		if (!(await isEnabled())) return;
```
Bekacru: I feel like is enabled should be checked at the top and the actions shouldn't be required to check

packages/better-auth/src/config/index.ts
frectonz: Yeah kinfish also suggest it, i will remove the global config.

Don't we need an anonymous ID that will be reused across projects, this can be thought of as a proxy to the number of users better auth has. We will need an anonymous ID that can be reused across different projects. I am currently storing the anonymous ID in this global config.

This will let use track situations in which even though i might have 3 projects that use better-auth in my computer, all of them will have the same anonymous ID when a telemetry event is published. 

To differentiate between the projects we have a separate project anonymous ID, but the global anonymous ID let's us count the number of devices that have better auth running in some way.

packages/better-auth/src/types/options.ts
frectonz: oh i can remove it, i was thinking some user might want to publish to their own telemetry endpoint, i also thought it would be useful during testing.

packages/better-auth/src/config/index.ts
Bekacru: but this id will be useless in prod, since you can't really persisit the config. For ID as I suggested, you should rely on a hashed base url or something

packages/cli/src/commands/telemetry.ts
```diff
@@ -0,0 +1,25 @@
+import { Command } from "commander";
+import { createGlobalConfig } from "../../../better-auth/src/config";
+import { TELEMETRY_CONFIG_KEY } from "../../../better-auth/src/telemetry/config-key";
+
+export const telemetryCommand = new Command("telemetry").action(() =>
```
Bekacru: We don't need to add a cli command. Setting the env should be enough

packages/better-auth/src/telemetry/index.ts
frectonz: fixed

packages/cli/src/commands/telemetry.ts
frectonz: removed

packages/better-auth/src/config/index.ts
frectonz: that sounds good, i will use the project Id as the anonymous ID

Bekacru: nvm, I get what you mean

Bekacru: But yeah, I don't think we care about per-device. It would be misleading, especially since most people will host their project in a serverless environment where the same thing could generate a lot of new IDs.

.changeset/full-hotels-serve.md
frectonz: fixed

packages/better-auth/src/telemetry/index.ts
frectonz: yeah you are right i have updated the implementation to check if telemetry is enabled only once, if telemetry is disabled the returned telemetry object will have no op functions, instead of the real `publish` and `report` functions.

frectonz:
_UPDATED_

### Configuring Telemetry Behaviour

#### 1. Via Auth config

```typescript
betterAuth({
  telemetry: {
    enabled: true
  }
});
```

#### 2. Via Environment Variables

Enable Telemetry
```
BETTER_AUTH_TELEMETRY=1
```

Disable Telemetry
```
BETTER_AUTH_TELEMETRY=0
```

Enable Telemetry Debug
_This means that instead of sending the collected info to the address specified, the collected info will be logged._
```
BETTER_AUTH_TELEMETRY_DEBUG=1
```

Issue #3695, opened 2025-07-29
Setting Active Organization not working with MongoDB


TheUntraceable:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Setup better auth with mongodb adapter and organization plugin
2. Create a new org
3. Try to set that org as the active one
https://codesandbox.io/p/devbox/l6c97p <- Fully setup, try the preview in a new tab (the MongoDB url setup there is newly made)

### Current vs. Expected behavior

Current - It says that member is not part of the organization.
Expected - Sets the active organization just fine.

### What version of Better Auth are you using?

1.3.4

### Provide environment information

```bash
- OS: Windows 11
- Browser: Brave
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});

https://codesandbox.io/p/devbox/l6c97p
```

### Additional context

I made this codesandbox for testing: https://codesandbox.io/p/devbox/l6c97p

Kinfe123:
should be fixed by now via #3757. if not let me now i will reopen this.

PR #3540, opened 2025-07-22
fix(organization): resolve slug when setting active organization

tinkerer-shubh:
closes https://github.com/better-auth/better-auth/issues/3528
    

---

## Summary by cubic
Fixed an issue where setting the active organization did not resolve organization slugs correctly. Now, both organization IDs and slugs are properly handled when updating the active organization.

PR #3708, opened 2025-07-30
Fix/3695 set active org slug resolve step 

RobiMez:
Adds a slug resolution step where slugs get converted to ids in the case where a slug is provided but not an id on setactiveorg
    

---

## Summary by cubic
Added a step to resolve organization slugs to IDs when only a slug is provided in setActiveOrganization, fixing cases where the ID is missing.

- **Bug Fixes**
 - Ensures setActiveOrganization works when given a slug instead of an ID.

### DIFF
--- a/.changeset/odd-apples-judge.md
+++ b/.changeset/odd-apples-judge.md
@@ -2,4 +2,4 @@
 "better-auth": patch
 ---
 
-fix(3695): Adds a slug resolution step in the case where slug is provided but not id
+fix(organization): Adds a slug resolution step in the case where slug is provided but not id


RobiMez:
Fixes #3695

Kinfe123 (commented):
i think we dont need this much comment. it is enough if we do have a comment on block that is important.

.changeset/odd-apples-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+fix(3695): Adds a slug resolution step in the case where slug is provided but not id
```
Bekacru:
```suggestion
fix(organization): Adds a slug resolution step in the case where slug is provided but not id
```

packages/better-auth/src/plugins/organization/routes/crud-org.ts
```diff
@@ -703,35 +706,47 @@ export const setActiveOrganization = <O extends OrganizationOptions>(
...
+			if (!organization) {
+				throw new APIError("BAD_REQUEST", {
+					message: ORGANIZATION_ERROR_CODES.ORGANIZATION_NOT_FOUND,
+				});
+			}
+
 			const isMember = await adapter.checkMembership({
 				userId: session.user.id,
-				organizationId,
+				organizationId: organization.id,
```
Bekacru, line 740: assuming this could be null when organizationSlug is only provided, we should probably update how checkMembersShip works instead so it can check for both slug and id

Bekacru: also in the spirit of improving our test, make sure to have a test for both with slug and id and test all the possible cases under `crud-org.test.ts`

TheUntraceable:
I think this a duplicate of #3540, right?

Bekacru:
@TheUntraceable you're right. closing this in favor of that

PR #3712, opened 2025-07-31
Fix: accountId condition being ignored and OAuth providerId type mismatch issue

j-token:
### 1. Issue with all accounts being fetched when `accountId` is `undefined`

* **Issue Description**:
  When `accountId: undefined` is included in the `where` condition in Prisma, the condition is ignored, and all accounts are fetched.
  As a result, when a user signs up via OAuth with the first account (`a@google.com`) in an empty database, the `accountId` of that account is always returned, causing incorrect login matching for all subsequent logins.

* **Solution**:
  The `accountId` condition is only included when it explicitly exists, and the `providerId` filter is also applied to ensure accurate user matching.

---

### 2. OAuth provider `id` type mismatch (`number` → `string`)

* **Issue Description**:
  Some OAuth providers (such as Kakao) return `profile.id` as a `number`, but internal logic and database schemas expect it to be a `string`, causing a type mismatch.

* **Solution**:
  Explicitly convert `providerId` values to strings using `.toString()` to maintain consistent handling across all providers.

    

---

## Summary by cubic
Fixed an issue where accountId conditions were ignored in OAuth logins, causing incorrect account matching, and resolved a type mismatch by converting providerId values to strings.

- **Bug Fixes**
  - Only include accountId in queries when defined to prevent fetching all accounts.
  - Ensure providerId is always a string for consistent matching across OAuth providers.

### DIFF
--- a/demo/nextjs/app/features.tsx
+++ b/demo/nextjs/app/features.tsx
@@ -59,28 +59,6 @@ const Card = ({
 	);
 };
 
-const AceternityIcon = () => {
-	return (
-		<svg
-			width="66"
-			height="65"
-			viewBox="0 0 66 65"
-			fill="none"
-			xmlns="http://www.w3.org/2000/svg"
-			className="h-10 w-10 text-black dark:text-white group-hover/canvas-card:text-white "
-		>
-			<path
-				d="M8 8.05571C8 8.05571 54.9009 18.1782 57.8687 30.062C60.8365 41.9458 9.05432 57.4696 9.05432 57.4696"
-				stroke="currentColor"
-				strokeWidth="15"
-				strokeMiterlimit="3.86874"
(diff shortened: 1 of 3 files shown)


packages/better-auth/src/db/internal-adapter.ts
```diff
@@ -536,15 +536,19 @@ export const createInternalAdapter = (
 			const account = await adapter
```
Bekacru, line 535: If the account ID isn't defined, we should probably just return an error. For providers that don't return an ID, we might need to fall back to the user ID or another consistent, unique identifier

j-token:
@Bekacru

himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #3719, opened 2025-07-31
Add username validation for length and format in availability check

emersonlaurentino:
Implement validation to ensure usernames meet specified length requirements and format criteria during availability checks. 

This enhancement improves error handling for invalid usernames.
    

---

## Summary by cubic
Added username length and format validation to the availability check, returning clear errors for invalid usernames.

- **Bug Fixes**
  - Returns specific errors for usernames that are too short, too long, or have invalid characters.


max-lancaster:
Not sure this really makes sense, valid and available are two different things and should probably be done in separate steps.

max-lancaster:
Well I guess by returning errors for invalid usernames you're effectively forcing users to validate before checking availability.

emersonlaurentino:
I disagree with you @max-lancaster, because there's no point in checking availability if I can't register it because it's invalid. 

Imagine the frustration of a user checking the availability of `XPTO` and the API reporting that it's available, but when they go to register, the validation won't allow it because it only allows lowercase usernames.

packages/better-auth/src/plugins/username/index.ts
```diff
@@ -268,12 +268,39 @@ export const username = (options?: UsernameOptions) => {
...
+					if (username.length > maxUsernameLength) {
+						ctx.context.logger.error("Username too long", { username });
+						throw new APIError("UNPROCESSABLE_ENTITY", {
+							message: ERROR_CODES.USERNAME_TOO_LONG,
+						});
+					}
+
+					const validator =
+						options?.usernameValidator || defaultUsernameValidator;
+
```
Bekacru:
```suggestion
const isValidUsername = await validator(username)
if (!isValidUsername) {
	throw new APIError("BAD_REQUEST", {
		message: ERROR_CODES.INVALID_USERNAME,
	});
}
```

packages/better-auth/src/plugins/username/index.ts
```diff
@@ -268,12 +268,39 @@ export const username = (options?: UsernameOptions) => {
...
+					if (!(await validator(username))) {
+						throw new APIError("UNPROCESSABLE_ENTITY", {
+							message: ERROR_CODES.INVALID_USERNAME,
+						});
+					}
+
```
Bekacru:
```suggestion
```

packages/better-auth/src/plugins/username/index.ts
```diff
@@ -268,12 +268,39 @@ export const username = (options?: UsernameOptions) => {
 							message: ERROR_CODES.INVALID_USERNAME,
 						});
 					}
+
+					const minUsernameLength = options?.minUsernameLength || 3;
+					const maxUsernameLength = options?.maxUsernameLength || 30;
+
+					if (username.length < minUsernameLength) {
+						ctx.context.logger.error("Username too short", { username });
+						throw new APIError("UNPROCESSABLE_ENTITY", {
```
Bekacru, line 277: We should use `BAD_REQUEST` instead of `UNPROCESSABLE_ENTITY`

max-lancaster:
> Imagine the frustration of a user checking the availability of `XPTO` and the API reporting that it's available, but when they go to register, the validation won't allow it because it only allows lowercase usernames.

The developer still needs to display a different validation message in either case so they might as well just do it in two steps.  If the developer only checks this endpoint for 200 status code, the end user will see a username is unavailable message instead of a username is invalid message with instructions on what a valid username is.  Also the endpoint is called `is-username-available` not `is-username-valid-and-available` so the public api interface doesn't fully express what the endpoint does.

himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #3726, opened 2025-08-01
fix(jwt): add options to plugin

grant0417:
Currently the OIDC plugin relies on the jwt options, but they are currently not added to the plugin as they are optional which led to inconsistent behavior. I have added the options and updated the type in the oidc plugin which would have caught this with typechecks.

    

---

## Summary by cubic
Added the jwt plugin options to ensure consistent behavior and updated the OIDC plugin type to catch related issues with type checks.

### DIFF
--- a/.changeset/slimy-paws-sin.md
+++ b/.changeset/slimy-paws-sin.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+fix(jwt): add options to plugin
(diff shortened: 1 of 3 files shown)


runreal-warman:
+1 just debugged exact same issue. Without this fix the oidc plugin will ignore your jwt configuration when `useJWTPlugin` is enabled.

Issue #3686, opened 2025-07-29
Organization additional fields not returning new fields


senyosimpson:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

To reproduce, just use my configuration down below and try get the field `publicId` on the data returned by `useActiveOrganization` or `authClient.organization.list()`

### Current vs. Expected behavior

The expected behaviour is that the field `publicId` exists on the returned data from the function _and_ that type inference works.

The current behaviour is that neither work.

### What version of Better Auth are you using?

1.3.4

### Provide environment information

```bash
MacOS
```

### Which area(s) are affected? (Select all that apply)

Types, Client

### Auth config (if applicable)

```typescript
export const auth = betterAuth({
  user: {
    additionalFields: {
      publicId: {
        type: "string",
        fieldName: "public_id",
        required: false,
        input: false,
        unique: true,
        returned: true,
      },
    },
  },
  plugins: [
    organization({
      teams: {
        enabled: true,
        allowRemovingAllTeams: false,
      },
      schema: {
        organization: {
          additionalFields: {
            publicId: {
              type: "string",
              fieldName: "public_id",
              required: false,
              input: false,
              unique: true,
              returned: true,
            },
          },
        },
        member: {
          additionalFields: {
            publicId: {
              type: "string",
              fieldName: "public_id",
              required: false,
              input: false,
              unique: true,
              returned: true,
            },
          },
        },
        team: {
          additionalFields: {
            publicId: {
              type: "string",
              fieldName: "public_id",
              required: false,
              input: false,
              unique: true,
              returned: true,
            },
          },
        },
        invitation: {
          additionalFields: {
            publicId: {
              type: "string",
              fieldName: "public_id",
              required: false,
              input: false,
              unique: true,
              returned: true,
            },
          },
        },
      },
    }),
  ],
});

And for the client config

export const authClient = createAuthClient({
  plugins: [
    inferAdditionalFields<typeof auth>(),
    organizationClient({
      teams: {
        enabled: true,
      },
      schema: inferOrgAdditionalFields<typeof auth>(),
    }),
  ],
});
```

### Additional context

I tried the `useActiveOrganization` function and the `authClient.organization.list()` function and neither return the  `publicId`. The user additional field  _does_ return the `publicId`. Interestingly, when creating an organization, the field is known. If `required` is `true` and I try create an organization without inputting the `publicId`, I get an error. So some part of it is working.

[...]

PR #3731, opened 2025-08-01
fix(organization): some endpoints do not return additional fields

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/3686
linear: https://linear.app/better-auth/issue/ENG-131/fix-ensure-organization-additionalfields-are-returned-and-inferred-in

    

---
## Summary by cubic
Ensures organization endpoints and client hooks include and correctly infer additional fields for organizations, members, teams, and invitations with client/server-safe typing. Addresses Linear ENG-131.

- **Bug Fixes**
  - Unified responses and types to include additional fields across endpoints and hooks (including nested members/teams/invitations), while keeping hidden fields typed but not exposed to clients.
  - Fixed role parsing and team inference; improved SSR detection in useAuthQuery to avoid server-side fetches.

<sup>Written for commit c343970431d078d0d6f0db2ac9118bb87b0e8000. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/package.json
+++ b/packages/better-auth/package.json
@@ -1,6 +1,6 @@
 {
   "name": "better-auth",
-  "version": "1.4.0-beta.18",
+  "version": "1.4.0-beta.19",
   "description": "The most comprehensive authentication framework for TypeScript.",
   "type": "module",
   "license": "MIT",
--- a/packages/better-auth/src/plugins/oidc-provider/authorize.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/authorize.ts
@@ -3,6 +3,7 @@ import { APIError } from "better-call";
 import { getSessionFromCtx } from "../../api";
 import { generateRandomString } from "../../crypto";
 import { getClient } from "./index";
+import { getAuthorizePromptSet } from "./middlewares/check-prompt";
 import type { AuthorizationQuery, OIDCOptions } from "./types";
 
 function formatErrorURL(url: string, error: string, description: string) {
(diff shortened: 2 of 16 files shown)


packages/better-auth/src/client/query.ts
```diff
@@ -6,7 +6,7 @@ import {
 import { atom, onMount, type PreinitializedWritableAtom } from "nanostores";
 
 // SSR detection
-const isServer = typeof window === "undefined";
+const isServer = () => typeof window === "undefined";
```
Bekacru, line 10: is this change related to a different issue?

ping-maxwell: This allows me to run hooks / atoms in tests: https://github.com/better-auth/better-auth/pull/3731/files#diff-7bbb92899636716deb61f6792bbcd85618249fa54e0067edb9d12943d0ff558fR2166

packages/better-auth/src/plugins/organization/routes/crud-members.ts
```diff
@@ -262,7 +266,7 @@ export const removeMember = <O extends OrganizationOptions>(options: O) =>
 					message: ORGANIZATION_ERROR_CODES.MEMBER_NOT_FOUND,
 				});
 			}
-			const roles = toBeRemovedMember.role.split(",");
+			const roles = (toBeRemovedMember.role as string).split(",");
```
Bekacru: to avoid casting here what if we move `InferMember<O, false> ` on the return type instead...see below

Bekacru: instead of 
```ts
 ctx.json({
	member: toBeRemovedMember,
});
```

```ts
 ctx.json({
	member: toBeRemovedMember as InferMember<O, false>,
});
```

ping-maxwell: All of these type asserting were due to this:

image

the `keyof` makes all of the role be type of `string | Symbol | number` or something like that, so I will just fix that and then remove all the type asserting.

statusunknown418:
should be added here that `useActiveOrganization` does not return `additionalFields` either

PabloDeCortes:
Hi @ping-maxwell,
Why this PR was marked as draft? This issue still persists

ping-maxwell:
cc @Bekacru / @himself65 ready for re-review + merge

senyosimpson:
just a heads up, might be handled in this PR already but it seems that `getFullOrganization` doesn't return the additional fields either

laduniestu:
when this PR will be merged?

ping-maxwell:
cc @Bekacru @himself65

laduniestu:
i tought that this PR will be includes on 1.3.31

senyosimpson:
@ping-maxwell any update on when this will be merged?

ping-maxwell:
> @ping-maxwell any update on when this will be merged?

It should be before or on 1.4

cc @Bekacru @himself65

packages/better-auth/src/client/query.ts
```diff
@@ -7,7 +7,7 @@ import { atom, onMount, type PreinitializedWritableAtom } from "nanostores";
 import type { SessionQueryParams } from "./types";
 
 // SSR detection
-const isServer = typeof window === "undefined";
+const isServer = () => typeof window === "undefined";
```
himself65, line 10: nit: this won't work for expo

ping-maxwell: You mean it can't be an arrow function?

Bekacru: typeof window

Bekacru (changes requested):

PR #3740, opened 2025-08-01
fix(client): prevent Promise unwrapping interference with proxy paths

### LINKED ISSUE #3650: Dynamic Import of `createAuthClient` Breaks with Proxy: Triggers Unexpected `.then` Call and 404
Hi,

I’m facing a significant issue when using `better-auth` in a public package that must be ESM/CJS compatible. To avoid CJS/ESM incompatibility (which causes the entire package to break at import time, even if the code path is never reached), I **must** use dynamic import (`await import('better-auth/client')`) for `createAuthClient`. See related [issue 3613 comment](https://github.com/better-auth/better-auth/issues/3613#issuecomment-3123375720).

**Minimal Example:**

```ts
// (Context: I cannot use static import for createAuthClient, because it breaks CJS consumers.)

let authClientInstance = null;

const getAuthClient = async () => {
  if (authClientInstance) return authClientInstance;
  const { createAuthClient } = await import("better-auth/client");
  authClientInstance = createAuthClient({ baseURL: "http://localhost:3100" });
  return authClientInstance;
};

const signInEmail = async (...args) => {
  const client = await getAuthClient();
  return client.signIn.email(...args);
};
```

**Issue:**
When I call one of this function, I get the following network error in the browser:

```
GET http://localhost:3100/api/auth/then 404 (Not Found)
```

After investigating, it appears that the proxy returned by `createAuthClient` is being treated as a `Promise` by the JavaScript runtime (or some internal code), and thus tries to access the `.then` property (Promise-like). Because the proxy does not implement a `then` method, this causes the proxy to treat `then` as a regular API path, resulting in an HTTP request to `/api/auth/then`.

`createDynamicPathProxy` code path: `packages/better-auth/src/client/vanilla.ts`

**Why dynamic import?**
If I use a static import of `better-auth/client`, it breaks CJS consumers *even if the code is never called* due to how Node/ESM interop works. (Static import loads the ESM module at load-time.)
Using dynamic import should solve this, but due to the proxy, it breaks functionally.

**Request:**

* Can you ensure the object returned by `createAuthClient` is *not* mistaken for a Promise (i.e., it should **not** have a `then` property unless it actually implements Promise behavior)?
* Or alternatively, is there a recommended pattern for using `createAuthClient` *only* with dynamic import in public packages without breaking the proxy API?

Kinfe123:
The pr fixes issue regarding to proxy returned by createAuthClient was unintentionally exposing a .then property by design ,resolving it like any other dynamic path (e.g., /api/auth/then). As a result, the runtime or libraries attempting to unwrap the client as a promise would trigger unintended requests to /api/auth/then, leading to unintended behavior. 
This could one way introducing a backward compat becasue of the ESM/CJS compat. 

closes #3650 or may be #3679


Bekacru:
can you add a test for it?

Bekacru:
closed in favor of #3679

PR #3741, opened 2025-08-01
docs: Remove non-existent invitation.createdAt field, add missing invitation.teamId field

acusti:
The current docs list a `createdAt` field for the `invitation` table that doesn’t exist and are missing the optional `teamId` field that does exist on the [invitation schema](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/organization/schema.ts#L32-L41)

### DIFF
--- a/docs/content/docs/plugins/organization.mdx
+++ b/docs/content/docs/plugins/organization.mdx
@@ -1512,6 +1512,13 @@ Table Name: `invitation`
       isForeignKey: true
     },
     {
+      name: "teamId",
+      type: "string",
+      description: "The ID of the team",
+      isOptional: true,
+      isForeignKey: true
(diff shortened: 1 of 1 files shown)


docs/content/docs/plugins/organization.mdx
```diff
@@ -1511,6 +1511,13 @@ Table Name: `invitation`
       description: "The ID of the organization",
       isForeignKey: true
     },
+    {
```
Bekacru: Unless team is enabled this wouldn't apply, so maybe it's better to mention this under teams section instead

acusti: that makes sense to me

PR #3748, opened 2025-08-02
chore: fix TypeScript to a correct version

himself65:
Before, the typescript version was using 5.6.1-rc
    

---

## Summary by cubic
Updated TypeScript to version 5.9.2 across the workspace to ensure consistency and use the latest stable release.


himself65:
btw, ts 5.9 has a breaking change https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/#lib.d.ts-changes. I tried and it breaks all better-auth buffer related type. Not sure how it affect the user side

@Bekacru

Bekacru:
I fixed the errors from the type checking. I don't think it'll break anything from the user side

PR #3755, opened 2025-08-02
feat: add support for `team` params for slack oauth

### LINKED ISSUE #3752: Slack `team` option not implemented but mentioned in docs

### To Reproduce

1. Add slack social provider
2. Add `team` configuration option
3. Get TS error `Object literal may only specify known properties, and 'team' does not exist in type 'Prettify<SlackOptions & { enabled?: boolean | undefined; }>'`

The docs mention this team option under [Workspace-Specific Sign In](https://www.better-auth.com/docs/authentication/slack#workspace-specific-sign-in), but looking at the provider code, this option is not implemented

1.3.4

```bash
n/a
```

Backend

```typescript

```

### Additional context

_No response_

Kinfe123:
closes #3752
    

## Summary by cubic
Added support for an optional team parameter in Slack OAuth to allow restricting sign-in to a specific Slack workspace.
more - https://api.slack.com/authentication/oauth-v2

### DIFF
--- a/.changeset/beige-scissors-tickle.md
+++ b/.changeset/beige-scissors-tickle.md
@@ -1,5 +1,0 @@
----
-"better-auth": patch
----
-
-feat: add support for `team` params for slack oauth


packages/better-auth/src/social-providers/slack.ts
```diff
@@ -33,7 +33,12 @@ export interface SlackProfile extends Record<string, any> {
 	"https://slack.com/team_image_default": boolean;
 }
 
-export interface SlackOptions extends ProviderOptions<SlackProfile> {}
+export interface SlackOptions extends ProviderOptions<SlackProfile> {
+	/**
+	 * Optional Slack workspace ID to restrict sign in to a specific workspace
+	 */
+	team?: string;
```
Bekacru: is this supposed to be a static value?

Kinfe123: Yeah. it can be like passing this parameter ensures the user will auth against that workspace. you can check the docs link i attached on the pr.

Bekacru: But shouldn't that be provided as a query param or something, so it's dynamic? I'm not sure how passing the team as a plugin option makes sense, since you'd probably want different users to choose different team or workspace IDs

Kinfe123: Yeah. also the team option is useful when you want to restrict Slack OAuth to a specific workspace which is ideal for internal tools or to avoid showing the workspace picker during login. you can see how their are constructing their team on their docs.

Bekacru: I think make it `string | (ctx)=>string` type. So if they want to change the team based on something from the request they can easily do that

Kinfe123: hmm..make sense.

Kinfe123: but i dont think we can pass them the ctx object from the slack itself. may be let them pass a function that returns string

Bekacru: The function needs some kind of input, like a request object or ctx. I know createAuthorizationURL doesn’t accept context right now, but maybe we should make it accept one so we can pass it to the callback. Otherwise, the function is useless

PR #3757, opened 2025-08-02
fix(org): incorrect use of find org by slug

Kinfe123:
- **use org slug from context body**
- **lint**

    

---

## Summary by cubic
Fixed an issue where the organization slug was not correctly read from the request body when finding an organization by slug.

### DIFF
--- a/packages/better-auth/src/plugins/organization/organization.test.ts
+++ b/packages/better-auth/src/plugins/organization/organization.test.ts
@@ -55,6 +55,7 @@ describe("organization", async (it) => {
 	});
 
 	let organizationId: string;
+	let organization2Id: string;
 	it("create organization", async () => {
 		const organization = await client.organization.create({
 			name: "test",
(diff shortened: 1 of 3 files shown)


himself65 (commented):
can you add a test case for this?

Bekacru (approved):

packages/better-auth/src/plugins/organization/organization.test.ts
```diff
@@ -116,11 +116,11 @@ describe("organization", async (it) => {
 			},
 		});
 
+		organization2Id = organization?.id as string;
```
himself65, line 119:
```suggestion
		organization2Id = organization?.id!;
```

packages/better-auth/src/plugins/organization/organization.test.ts
```diff
@@ -178,6 +178,23 @@ describe("organization", async (it) => {
...
+			fetchOptions: {
+				headers,
+			},
+		});
+		const session = await client.getSession({
+			fetchOptions: {
+				headers,
+			},
+		});
+		expect((session.data?.session as any).activeOrganizationId).toBe(
```
himself65, line 194: As any here? Could we improve the type

Issue #3772, opened 2025-08-04
[STRIPE] Plugin should support stripe subscription id for organizations


sigma-andex:
Currently the stripe subscription id is always on the user. In case of using organizations, the stripe subscription id should however be on the organization, or at least it should be possible to set it on the organization.

# Problem

Here is a simple scenario:

Imagine freelancer Bob Wilson working for two corporations Acme Corporation and Krusty Krab. 
If he buys a subscription for Acme Corp., a stripe user will be created in stripe and the billing address will be set to Acme Corp's billing address. Now when he buys another subscription for Krusty Krab, it will use the existing stripe customer id from the customer (since it was already created previously) and the the billing address for the user will be overwritten with the Krusty Krab's billing address. 

This can potentially lead to some nasty problems with invoicing. 

# Solution

The plugin should support stripe customer ids on the organization and use these when organizations are enabled.

# Temporary workaround

We have created a simple patch that allows to pass in an additional stripeCustomerId  to the request (see attached pull request #3773 ). However this should only be done when using better-auth only backend-side. So we added a field stripe_customer_id on the subscription, then manually generate a new stripe customer id for each generation and pass in the organization's stripe customer id to the initial subscription request.

saturnonearth:
Yea this is a desperately needed feature - personally I can't proceed with using Stripe without it. In my mind, this would be the least confusing way to manage this. Organizations are a business, and therefore should be tied to a Stripe customer. This would make it easier to switch owners of the business as well or hand off.

kylegwalsh:
Just came across this as well. In my application, I'm forcing all users to be associated with organizations (creating a default "Personal" organization during signup). Having everything attached to organizations greatly simplifies the database relations and it would be great to have stripe supported in the same way.

sigma-andex:
Still relevant and hopefully getting solved in #5952

bytaesu:
Hello @sigma-andex,

I haven’t finished #5952 yet due to other tasks, but I haven’t forgotten. I’ll wrap it up soon 🙂
I discussed something in that PR, could you share your thoughts on the preferred approach? I’ll complete it soon.

PR #3773, opened 2025-08-04
Add custom stripeCustomerId to request & fix missing trial fields

sigma-andex:
Solves #3772

Additionally, the trial fields were missing from the schema, so added these as well
    

---

## Summary by cubic
Added support for passing a custom Stripe customer ID when upgrading subscriptions and included missing trial fields in the subscription schema.

- **New Features**
  - Accepts a custom Stripe customer ID in the upgrade flow.
  - Adds trialStart, trialEnd, and updatedAt fields to the subscription schema.

### DIFF
--- a/packages/stripe/src/index.ts
+++ b/packages/stripe/src/index.ts
@@ -167,6 +167,16 @@ export const stripe = <O extends StripeOptions>(options: O) => {
 						})
 						.optional(),
 					/**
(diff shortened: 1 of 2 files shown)


sigma-andex:
Closing this, as this is not the proper way to do it. The stripe customer id should be attached to the organization table

PR #3778, opened 2025-08-04
fix(adapter): kysely adapter renamed fields

Kinfe123:
The fix addresses the issue where the Kysely adapter couldn't find field definitions for renamed fields, preventing proper data transformation like date and other and causing database binding errors as attached below
Screenshot 2025-08-04 at 2 10 28 PM
. The solution addresses create-adapter's getDefaultFieldName function to properly handle renamed fields.

### DIFF
--- a/packages/better-auth/src/adapters/kysely-adapter/kysely-adapter.ts
+++ b/packages/better-auth/src/adapters/kysely-adapter/kysely-adapter.ts
@@ -93,6 +93,14 @@ export const kyselyAdapter = (db: Kysely<any>, config?: KyselyAdapterConfig) =>
 				const { type = "sqlite" } = config || {};
 				let f = schema[model]?.fields[field];
 				if (!f) {
+					const foundField = Object.values(schema[model]?.fields || {}).find(
+						(f) => f.fieldName === field,
+					);
+					if (foundField) {
+						f = foundField;
+					}
+				}
+				if (!f) {
 					//@ts-expect-error - The model name can be a sanitized, thus using the custom model name, not one of the default ones.
 					f = Object.values(schema).find((f) => f.modelName === model)!;
 				}
--- a/packages/better-auth/src/adapters/kysely-adapter/test/normal/adapter.kysely.test.ts
+++ b/packages/better-auth/src/adapters/kysely-adapter/test/normal/adapter.kysely.test.ts
@@ -46,6 +46,11 @@ export const opts = ({
 		session: {
 			modelName: "sessions",
 		},
+		verification: {
+			fields: {
+				expiresAt: "expires_at",
+			},
+		},
 		advanced: {
 			database: {
 				useNumberId: isNumberIdTest,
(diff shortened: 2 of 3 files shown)


packages/better-auth/src/adapters/kysely-adapter/kysely-adapter.ts
```diff
@@ -92,6 +92,14 @@ export const kyselyAdapter = (db: Kysely<any>, config?: KyselyAdapterConfig) =>
...
 				const { type = "sqlite" } = config || {};
 				let f = schema[model]?.fields[field];
+				if (!f) {
+					const foundField = Object.values(schema[model]?.fields || {}).find(
+						(f) => f.fieldName === field,
+					);
+					if (foundField) {
+						f = foundField;
+					}
+				}
```
astahmer, line 110: hey, this needs to be AFTER the next condition (`f = Object.values(schema).find((f) => f.modelName === model)!;`) otherwise it will still fail to find the field if the table name (model) was ALSO was renamed (ex: `verification` -> `auth_verification`)

you might want to add a test for that

astahmer:
for anyone stumbling on this, here's the patch that you can add in `pnpm-workspace.yaml`

```yaml
patchedDependencies:
  better-auth@1.3.7: patches/better-auth@1.3.7.patch
```

```diff
diff --git a/dist/shared/better-auth.gn8iAL5-.mjs b/dist/shared/better-auth.gn8iAL5-.mjs
index ae7a12a1b861c26fd08cdc893210dd4b8fffe4cb..a5aa2a3b2a553ca1027a0e5f6992ed5aa55ac6dc 100644
--- a/dist/shared/better-auth.gn8iAL5-.mjs
+++ b/dist/shared/better-auth.gn8iAL5-.mjs
@@ -122,9 +122,15 @@ const kyselyAdapter = (db, config) => createAdapter({
         return value;
       }
       const { type = "sqlite" } = config || {};
-      let f = schema[model]?.fields[field];
+      const schemaModel = schema[model]?? Object.values(schema).find((f2) => f2.modelName === model)
+      let f = schemaModel?.fields[field];
       if (!f) {
-        f = Object.values(schema).find((f2) => f2.modelName === model);
+        const foundField = Object.values(schemaModel?.fields || {}).find(
+          (f) => f.fieldName === field,
+        );
+        if (foundField) {
+          f = foundField;
+        }
       }
       if (f.type === "boolean" && (type === "sqlite" || type === "mssql") && value !== null && value !== void 0) {
         return value ? 1 : 0;

```

Kinfe123:
@astahmer  we dont need to pull iteration everytime for finding if the model is renamed which might the odd for happening. we should be fallback to that instead.

wamujlb:
hi @Kinfe123 are there any news regarding when this patch is going to be released? Or maybe you know another solution that we can use to fix this issue using BetterAuth API, e.g. databaseHooks, plugins, etc.

CC: @astahmer

PR #3783, opened 2025-08-04
fix: await `ctx` in middleware

himself65:
`ctx` could be a promise in some cases
    

---

## Summary by cubic
Fixed middleware to await the context if it is a promise, ensuring correct context merging before passing to plugins.

- **Bug Fixes**
 - Updated middleware to handle promise-based contexts.
 - Added tests to verify correct context handling.


himself65:
btw im trying to update lint rules. But noticed that the biome version is outdated

PR #3785, opened 2025-08-04
feat(organization): add option for requiring email verificaiton

Bekacru:
## Summary by cubic
Added an option to require users to verify their email before accepting an organization invitation.

- **New Features**
  - Added requireEmailVerificationOnAcceptInvitation setting to OrganizationOptions.
  - Blocked invitation acceptance if the user's email is not verified and the option is enabled.
  - Added a new error message for unverified email attempts.

### DIFF
--- a/docs/content/docs/plugins/organization.mdx
+++ b/docs/content/docs/plugins/organization.mdx
@@ -584,6 +584,25 @@ type acceptInvitation = {
 ```
 </APIMethod>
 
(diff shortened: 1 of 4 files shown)


NiklasDah:
@Bekacru, thank you for adding this! 👍🏻 
I believe that this option should be true by default as it otherwise enables a very simple attack vector (if you know someone gets invited you can hijack their account)

Issue #2715, opened 2025-05-19
useSession with React JS - unable to get the updated values


Tholkappiar:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

**React client side code :** 

**React version** : ^19.1.0
```
import { authClient } from "../lib/auth-client";

export function Temp() {
    const { data: session, isPending, error } = authClient.useSession();

    return (
        <div>
            {session ? (
                <>
                    {/* after the session expiry which in cookies, i am still getting 
                    these values - meaning not re-rendered with updated null state when the session expired */}

                    <div>Logged in as: {session.user.name}</div>
                    <div>is Pending : {isPending}</div>
                    <div>is error : {error?.error}</div>
                </>
            ) : (
                <>
                    <div>Not logged in</div>
                    <div>is Pending : {isPending}</div>
                    <div>is error : {error?.error}</div>
                </>
            )}
        </div>
    );
}
```

**Express server side code :** 

**Express version** : ^5.1.0
**Better auth version** : ^1.2.7

```
import { betterAuth } from "better-auth";
import { prismaAdapter } from "better-auth/adapters/prisma";
import { PrismaClient } from "../../generated/prisma/client";
import { openAPI } from "better-auth/plugins";
import { CONFIG } from "./config";

const prisma = new PrismaClient();
export const auth = betterAuth({
    database: prismaAdapter(prisma, {
        provider: "postgresql",
    }),
    emailAndPassword: { enabled: true },
    trustedOrigins: [CONFIG.FRONTEND_URL],
    session: {
        expiresIn: 30, // 30 sec for test
    },
});

```

**Steps :** 

**Important and Easier to reproduce the issue with this below repo  :** 
Reproduced the exact same issue with readme file : https://github.com/Tholkappiar/better-auth-issue/

1. use the above repo which contains the React and express code
2. use the ```authClient.useSession();``` like usual to reproduce this issue !

### Current vs. Expected behavior

**Current behaviour :**
- The component doesn't re-render or returning any new updated values 

**Expected behaviour :** 
- The component needs to re-render after the session expired with the updated null values 

### What version of Better Auth are you using?

^1.2.7

### Provide environment information

```bash
- OS : macOS
- Browser : Brave (Brave is up to date Version 1.78.102 Chromium: 136.0.7103.113 (Official Build) (arm64))
```

### Which area(s) are affected? (Select all that apply)

Client

### Auth config (if applicable)

```typescript
[...]

Issue #1006, opened 2024-12-23
`useSession()` not always triggering a state change


kedom1337:
### Is this suited for github?

- [X] Yes, this is suited for github

### To Reproduce

1. Sign in with email and password
2. Use `useSession()` in the following component:

```ts
export function AuthLayout({ children }: AppLayoutProps) {
  const sidebar = useConfigStore((state) => state.sidebar)
  const { data: session } = useSession()

  return (
    <SidebarProvider defaultOpen={sidebar}>
      <Sidebar collapsible="icon" variant="inset">
        <SidebarHeader>Test</SidebarHeader>
        <SidebarContent>
          <NavMain />
        </SidebarContent>
        <SidebarFooter>
          <NavUser
            name={session?.user.name ?? ''}
            email={session?.user.email ?? ''}
          />
        </SidebarFooter>
        <SidebarRail />
      </Sidebar>
      <SidebarInset>
        <NavHeader />
        <div className="m-4 mt-0 h-full">{children}</div>
      </SidebarInset>
    </SidebarProvider>
  )
}
```

### Current vs. Expected behavior

- **Current**:
The session state always starts with `null` (as it should). It then does a network request to `get-session` which succeeds. The problem now is that the state does not update maybe 50% to the time resulting in me not being able to show the user's name in this example. If I then trigger a component rerender by e.g. opening the navbar, the session information is properly reflected in the UI

- **Expected**:
I expect the state to always update after a successfull `get-session` call

### What version of Better Auth are you using?

1.1.1

### Provide environment information

```bash
OS: Linux 6.12.4-arch1-1
Browser: Chrome 130.0.6723.116
```

### Which area(s) are affected? (Select all that apply)

Client

### Auth config (if applicable)

```typescript
betterAuth({
      trustedOrigins: c.env.BETTER_AUTH_TRUSTED_ORIGINS.split(','),
      database: drizzleAdapter(c.get('db'), {
        provider: 'pg',
        usePlural: true,
      }),
      emailAndPassword: {
        enabled: true,
      },
      advanced: {
        cookiePrefix: 'viaptic',
      },
    })
```

### Additional context

I tested this in both production and development and on both Chrome and Firefox. The issue persists everywhere

cem2ran:
I'm also seeing this. version 1.1.3. Not seeing useSession updating after logging in with apple in expo app.

Nicolab:
Because you should use useEffect to trigger a React rerender when the session object change

[...]

PR #3795, opened 2025-08-05
fix: revalidate session after expire

himself65:
checklist:
- [ ] Unit Test

### DIFF
--- a/packages/better-auth/src/client/session-atom.ts
+++ b/packages/better-auth/src/client/session-atom.ts
@@ -10,6 +10,16 @@ export function getSessionAtom($fetch: BetterFetch) {
 		session: Session;
 	}>($signal, "/get-session", $fetch, {
 		method: "GET",
+		onSuccess: (ctx) => {
+			// todo: we should implement the full revalidation logic for async resources, like SWR or React Query
+			// revalidate the session signal after expiration
+			if ("session" in ctx.data && "expiresAt" in ctx.data.session) {
+				const expiresAt = new Date(ctx.data.session.expiresAt);
+				setTimeout(() => {
+					$signal.set(false);
+				}, expiresAt.getTime() - Date.now());
+			}
+		},
 	});
 	return {
 		session,


himself65:
Just tried to add a test using React, and it's pretty hard now

- Vitest version is tooooo outdated, it's v1, now it's v3
- Idk why setting up the DB in happy-dom is much harder than I thought. The cookie is missing. It's always 401

Bekacru:
@himself65 
- we can update the vitest version (latest in September since we never updated vitest)
- I think happy dom strictly tries to emulate browser (node APIs won't work)

Bekacru:
And regarding the PR, I'm not sure if we should try to revalidate once it’s expired, since at that point the user will probably be signed out. But, I was thinking of adding a way that sends a request to getSession at intervals so it can be updated at some point. It shouldn't happen like onSuccess callback or something but rather a process that runs throughout the app to keep the session updated. This is mostly useful if, for example people only call getSession in server components or in places where it can’t update cookies, which can cause issues (there are a bunch of issues regarding that)

himself65:
> And regarding the PR, I'm not sure if we should try to revalidate once it’s expired, since at that point the user will probably be signed out. But, I was thinking of adding a way that sends a request to getSession at intervals so it can be updated at some point. It shouldn't happen like onSuccess callback or something but rather a process that runs throughout the app to keep the session updated. This is mostly useful if, for example people only call getSession in server components or in places where it can’t update cookies, which can cause issues (there are a bunch of issues regarding that)

Yeah, true. I think the ideal behavior is sth like SWR or react-query, we might need to manually write much logic for that

himself65:
I think maybe we should throw the error or log a warning right now in useSession if it expires? Or add an extra parameter called `autoRefetch` or something

Bekacru:
> Yeah, true. I think the ideal behavior is sth like SWR or react-query, we might need to manually write much logic for that

Yeah, I wanted to avoid reinventing SWR but at some point it might be a necessary evil unless we move to framework specific packages

himself65:
> I think happy dom strictly tries to emulate the browser (node APIs won't work)

background: to reproduce https://github.com/better-auth/better-auth/issues/2715 and https://github.com/better-auth/better-auth/issues/1006

Yeah, I set up a real playwright + nodejs test environment (similar to what I did in [waku](https://github.com/wakujs/waku/tree/main/e2e)), but I kinda feel it's too complicated, so I tried vitest + playwright, but it cannot run nodejs at the same time, then I tried vitest + happy-dom, well, it stuck because fake DOM environment. I might need to go back to the first one... 😢

Bekacru:
> I think maybe we should throw the error or log a warning right now in useSession if it expires? Or add an extra parameter called `autoRefetch` or something

I think expiration is rare when you call `useSession` since sessions are updated based on updateAge rather than only when they expire.

himself65:
Maybe we could integrate with https://github.com/nanostores/query ?

PR #3811, opened 2025-08-05
feat: support device authorization

himself65:
Support device authorization, following [RFC8628](https://datatracker.ietf.org/doc/html/rfc8628)

## Checklist
- [ ] Unit test
- [ ] Add demo on next


himself65:
I was trying to use https://github.com/vadimdemedes/ink/, but it just throws a weird runtime error when I use it. I doubt it's node_modules related. I give up using it now

packages/better-auth/src/plugins/device-authorization/index.ts
```diff
@@ -0,0 +1,858 @@
...
+							{
+								field: "userCode",
+								value: cleanUserCode,
+							},
+						],
+					});
+
+					if (!deviceCodeRecord) {
+						return ctx.json(
+							{
```
Bekacru: I think here you might want to use `throw ctx.error` or `throw new APIError` so the return type isn’t inferred. This should be done for all error returns

packages/better-auth/src/plugins/device-authorization/index.ts
```diff
@@ -0,0 +1,855 @@
...
+	 */
+	formatUserCode?: boolean;
+	/**
+	 * Enable rate limiting for token polling.
+	 * @default true
+	 */
+	enableRateLimiting?: boolean;
+}
+
+const deviceCodeSchema: AuthPluginSchema = {
```
Bekacru: change this to satisfy instead

packages/better-auth/src/plugins/device-authorization/index.ts
```diff
@@ -0,0 +1,855 @@
...
+		// Format with hyphen for readability if enabled (e.g., "WDJB-MJHT")
+		if (opts.formatUserCode && code.length === 8) {
+			return `${code.slice(0, 4)}-${code.slice(4)}`;
+		}
+		return code;
+	};
+
+	return {
+		id: "device-authorization",
+		schema: deviceCodeSchema,
```
Bekacru: See other plugin examples on how to make this schema configurable from the user side by accepting a schema fields in the plugin options

himself65:
I'm doing a CLI personally, so on-device is a killer feature to me. But I'm also thinking about a better API for that

packages/better-auth/src/plugins/device-authorization/index.ts
```diff
@@ -0,0 +1,855 @@
...
+		for (let i = 0; i < opts.userCodeLength; i++) {
+			code += chars[Math.floor(Math.random() * chars.length)];
+		}
+		return code;
+	};
+
+	const formatUserCode = (code: string) => {
+		// Format with hyphen for readability if enabled (e.g., "WDJB-MJHT")
+		if (opts.formatUserCode && code.length === 8) {
+			return `${code.slice(0, 4)}-${code.slice(4)}`;
```
himself65: This is not good, need to remove `-` from the code side. It should be user-land defined behavior

packages/better-auth/src/plugins/device-authorization/index.ts
```diff
@@ -0,0 +1,855 @@
...
+	};
+
+	const generateUserCode = async () => {
+		if (opts.generateUserCode) {
+			return opts.generateUserCode();
+		}
+		const chars = opts.userCodeCharset;
+		let code = "";
+		for (let i = 0; i < opts.userCodeLength; i++) {
+			code += chars[Math.floor(Math.random() * chars.length)];
```
himself65: math random is not good. use `crypto.getRandomValues`

packages/better-auth/src/plugins/device-authorization/index.ts
```diff
@@ -0,0 +1,855 @@
...
+		if (opts.generateDeviceCode) {
+			return opts.generateDeviceCode();
+		}
+		return generateRandomString(opts.deviceCodeLength, "a-z", "A-Z", "0-9");
+	};
+
+	const generateUserCode = async () => {
+		if (opts.generateUserCode) {
+			return opts.generateUserCode();
+		}
```
himself65, line 172: not good

packages/better-auth/src/plugins/device-authorization/index.ts
```diff
@@ -0,0 +1,855 @@
...
+		userCodeCharset: "ABCDEFGHJKLMNPQRSTUVWXYZ23456789", // Excluding similar looking characters
+		formatUserCode: true,
+		enableRateLimiting: true,
+		...options,
+	};
+
+	const generateDeviceCode = async () => {
+		if (opts.generateDeviceCode) {
+			return opts.generateDeviceCode();
+		}
```
himself65, line 165: not good

ripgrim:
need this 🙏 @Bekacru

himself65:
> need this 🙏 @Bekacru

I will continue on this next week!

himself65:
I decided not to add a CLI helper function, but add it to document:

```ts

type InferResponse<T extends Endpoint> = T extends (ctx: infer C) => infer R
	? C extends InputContext<any, any>
		? (...data: any[]) => Promise<
				BetterFetchResponse<
					NonNullable<Awaited<R>>,
					T["options"]["error"] extends StandardSchemaV1
						? // InferOutput
							NonNullable<T["options"]["error"]["~standard"]["types"]>["output"]
						: {
								code?: string;
								message?: string;
							}
				>
			>
		: never
	: never;

type InferEndpointSuccessType<T extends Endpoint> = InferResponse<T> extends (
	...args: any[]
) => Promise<BetterFetchResponse<infer S, any>>
	? S
	: never;

type InferEndpointErrorType<T extends Endpoint> = InferResponse<T> extends (
	...args: any[]
) => Promise<BetterFetchResponse<any, infer E>>
	? E
	: never;

type EndPoints = ReturnType<typeof deviceAuthorization>["endpoints"];
type DeviceCodeSuccessResponse = InferEndpointSuccessType<
	EndPoints["deviceCode"]
>;
type DeviceCodeErrorResponse = InferEndpointErrorType<EndPoints["deviceCode"]>;
type TokenSuccessResponse = InferEndpointSuccessType<EndPoints["deviceToken"]>;
type TokenErrorResponse = InferEndpointErrorType<EndPoints["deviceToken"]>;

type RequestLoginOptions = {
	onOpenURL?: (url: string) => Awaitable<void>;
	onError?:
		| ((step: "deviceCode", error: DeviceCodeErrorResponse) => Awaitable<void>)
		| ((step: "token", error: TokenErrorResponse) => Awaitable<void>);
};
```

himself65:
Gonna try this in beta and see if it's good

thisismert:
There is an issue about invertal.

In plugin options:
- Using `"5s"` (as shown in the docs) causes a DB error, since the column expects a number.
- Using `"5"` does not throw an error, but the returned value when called authClient.device.code ends up being `0`.

So the current behavior is inconsistent with both the docs and the schema.

DB error when using `"5s"` is like this:
```
  query: 'insert into "device_code" ("id", "device_code", "user_code", "user_id", "expires_at", "status", "last_polled_at", "polling_interval", "client_id", "scope") values ($1, $2, $3, default, $4, $5, default, $6, $7, $8) returning "id", "device_code", "user_code", "user_id", "expires_at", "status", "last_polled_at", "polling_interval", "client_id", "scope"',
  params: [Array],
  [cause]: [Error [NeonDbError]: invalid input syntax for type integer: "5s"] {
    severity: 'ERROR',
```

himself65:
> There is an issue about invertal.
> 
> In plugin options:
> 
> * Using `"5s"` (as shown in the docs) causes a DB error, since the column expects a number.
> * Using `"5"` does not throw an error, but the returned value when called authClient.device.code ends up being `0`.
> 
> So the current behavior is inconsistent with both the docs and the schema.
> 
> DB error when using `"5s"` is like this:
> 
> ```
>   query: 'insert into "device_code" ("id", "device_code", "user_code", "user_id", "expires_at", "status", "last_polled_at", "polling_interval", "client_id", "scope") values ($1, $2, $3, default, $4, $5, default, $6, $7, $8) returning "id", "device_code", "user_code", "user_id", "expires_at", "status", "last_polled_at", "polling_interval", "client_id", "scope"',
>   params: [Array],
>   [cause]: [Error [NeonDbError]: invalid input syntax for type integer: "10s"] {
>     severity: 'ERROR',
> ```

Thanks for the feedback, I will check this

himself65:
https://github.com/better-auth/better-auth/pull/4243

PR #3815, opened 2025-08-05
fix(oauth2): compare accountId as forced string to prevent duplicates

### LINKED ISSUE #3814: Duplicated oauth accounts because of type mismatch

### To Reproduce

1. Add generic oauth plugin and setup a provider that returns accountId's as only numbers.
2. Try to login multiple times with that oauth provider.

Expected to have only one account item for the same providerId accountId combination but instead every time user logins, that creates new account in the database with same providerId and accountId.

When I dug deeper, I found that in this section https://github.com/better-auth/better-auth/blob/e57a49f99dbf72d462d94e8f8ccaec747eb1a725/packages/better-auth/src/oauth2/link-account.ts#L43-L47 accountId returns as a string from database but other one returns as a number from the endpoint. So they dont align and cause `hasBeenLinked` to be undefined.

in line 46 comparing like this `a.accountId === String(account.accountId)` seems like solved the issue.

1.3.4 or 1.3.5-beta2 both

```bash
(not relevant)
- OS: MacOS
- Browser: Chrome
```

Backend

```typescript

```

### Additional context

might be related to #2062 too.

cemreinanc:
in oauth2/link-account.ts file: `accountId` returns as a string from database but other one returns as a number from the endpoint. So they dont align and cause `hasBeenLinked` to be `undefined`.
in line 46 comparing like this `a.accountId === String(account.accountId)` seems like solved the issue.

Closes #3814
    

---

## Summary by cubic
Fixed an issue where OAuth account linking could create duplicate accounts by ensuring accountId is always compared as a string.

### DIFF
--- a/packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
@@ -18,6 +18,7 @@ describe("oauth2", async () => {
 	const { customFetchImpl, auth } = await getTestInstance({
 		plugins: [
 			genericOAuth({
+
 				config: [
 					{
 						providerId,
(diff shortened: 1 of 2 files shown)


Kinfe123 (approved):

packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
himself65, line 1: What is the real case causing the accountId to be a number? Could there be a deeper issue behind this? since the type annotation around is all string

himself65: I realized that issue is from https://github.com/better-auth/better-auth/pull/3816/files#diff-16e9f39c92f582849ffc6ba9be9b14f8f04a8137bf025e3c876159c01d254564R307 if `getUserInfo` returns the number. we don't give any error about this

Bekacru:
closed in favor of #3816

Issue #3819, opened 2025-08-06
secondaryStorage not deleting session in `active-sessions` after signing out


StashOfCode:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Add Redis as `secondaryStorage`
2. Sign in / Sign out multiple times
3. Execute a `GET active-sessions-{user_id}` command and inspect the result

<img width="394" height="510" alt="Image" src="https://github.com/user-attachments/assets/570dd73f-a3d1-4e03-a1b1-be08698fb8ba" />

### Current vs. Expected behavior

A session in `active-sessions` is not deleted on signout. Items in `active-sessions` keep infinitely growing until expiration.

`signOut` calls `deleteSession`, but there's no code that deletes a signed out session from `active-sessions`

https://github.com/better-auth/better-auth/blob/e57a49f99dbf72d462d94e8f8ccaec747eb1a725/packages/better-auth/src/db/internal-adapter.ts#L447-L458

### What version of Better Auth are you using?

1.3.4

### Provide environment information

```bash
-
```

### Which area(s) are affected? (Select all that apply)

Backend, Client

### Auth config (if applicable)

```typescript
export const auth = betterAuth({
  appName: "App",
  baseURL: process.env.NEXT_PUBLIC_BASE_URL!,
  database: drizzleAdapter(db, {
    provider: "pg",
    schema,
  }),
  secondaryStorage: redisSecondaryStorage,
  emailAndPassword: {
    enabled: true,
    autoSignIn: false,
    requireEmailVerification: true,
    resetPasswordTokenExpiresIn: 600, // 10 minutes
    // minPasswordLength: 8,
    // maxPasswordLength: 24,
    sendResetPassword: async ({ user, url }) => {
      // TODO: wrap in try-catch to handle errors?
      resend.emails.send({
        from: process.env.RESEND_FROM_EMAIL as string,
        to: user.email,
        subject: "Reset your password",
        react: ForgotPasswordEmail({
          userEmail: user.email,
          resetUrl: url,
        }),
      });
    },
    // onPasswordReset: async ({ user }, request) => {
    //   // your logic here
    //   console.log(`Password for user ${user.email} has been reset.`);
    // },
  },
  emailVerification: {
    sendVerificationEmail: async ({ user, url }) => {
      // TODO: wrap in try-catch to handle errors?
      resend.emails.send({
        from: process.env.RESEND_FROM_EMAIL as string,
        to: user.email,
[...]

PR #3820, opened 2025-08-06
feat: remove revoked session from active list

himself65:
Fixes: https://github.com/better-auth/better-auth/issues/3819
    

---

## Summary by cubic
Revoked sessions are now removed from the user's active session list, so listing sessions will no longer show revoked tokens.

- **Bug Fixes**
  - Updated session deletion to also remove the session from the active sessions record.
  - Added tests to confirm revoked sessions are excluded from active lists.

### DIFF
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -9,7 +9,7 @@ import prompts from "prompts";
 import fs from "fs/promises";
 import chalk from "chalk";
 import { getAdapter } from "better-auth/db";
-import { getGenerator } from "../generators";
+import { generateSchema } from "../generators";
 
 export async function generateAction(opts: any) {
 	const options = z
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/db/internal-adapter.ts
```diff
@@ -446,6 +448,42 @@ export const createInternalAdapter = (
...
+				if (data) {
+					const { session } =
+						safeJSONParse<{
+							session: Session;
+							user: User;
+						}>(data) ?? {};
+					if (!session) {
+						logger.error("Session not found in secondary storage");
+						return;
+					}
```
himself65, line 462: For defense programming, the user could clean up the secondary storage, since it's a KV store. I think we should just ignore them.

btw early return is wrong here

packages/better-auth/src/api/routes/session-api.test.ts
```diff
@@ -363,11 +363,15 @@ describe("session storage", async () => {
 		},
 	});
 
+	beforeEach(() => {
+		store.clear();
+	});
```
himself65, line 368: We should cleanup the store

PR #3822, opened 2025-08-06
feat: telemetry

frectonz:
## Summary by cubic
Added anonymous telemetry to Better Auth to collect usage data for improving performance and features. Users can audit or disable telemetry through config or environment variables.

- **New Features**
  - Collects environment, system, and config info (with sensitive data redacted) on startup.
  - Sends telemetry events to a remote endpoint or logs locally in debug mode.
  - Added documentation and UI in the docs to explain telemetry and opt-out options.


packages/better-auth/src/init.ts
```diff
@@ -91,7 +92,10 @@ export const init = async (options: BetterAuthOptions) => {
 		return generateId(size);
 	};
 
-	const ctx: AuthContext = {
+	const telemetry = await createTelemetry({ logger, options });
+	telemetry.report();
```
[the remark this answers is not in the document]
frectonz: we want the telemetry report to be run in the background, so we don't await it here

packages/better-auth/src/telemetry/project-id.ts
```diff
@@ -0,0 +1,46 @@
+import { execSync } from "child_process";
+
+import { generateId } from "../utils";
+import { hashToBase64 } from "../crypto";
+import { getNameFromLocalPackageJson } from "../utils/package-json";
+
+let projectIdCached: string | null = null;
```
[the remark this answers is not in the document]
frectonz: memoized properly now

packages/better-auth/src/telemetry/endpoint.ts
```diff
@@ -0,0 +1,25 @@
...
+const TELEMETRY_ENDPOINT = "https://telemetry.better-auth.com/v1/track";
+export const realEndpoint: TelemetryEndpoint = async (event) => {
+	try {
+		await betterFetch(TELEMETRY_ENDPOINT, {
+			method: "POST",
+			body: JSON.stringify(event),
+			headers: { "content-type": "application/json" },
+			retry: 5,
+		});
+	} catch {}
```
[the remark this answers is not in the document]
frectonz: that's exactly what we want we don't want telemetry publishing errors to be seen

packages/better-auth/src/telemetry/detectors/detect-project-info.ts
```diff
@@ -0,0 +1,40 @@
+import { execSync } from "child_process";
```
himself65: `node:child_process`, or bun, deno won't detect correctly

frectonz: oh thanks for catching this

docs/content/docs/reference/telemetry.mdx
```diff
@@ -0,0 +1,47 @@
+---
+title: Telemetry
+description: "Better Auth now collects anonymous telemetry data about general usage."
```
himself65:
```suggestion
description: Better Auth now collects anonymous telemetry data about general usage.
```

packages/better-auth/src/init.ts
himself65: No, AI comment is kinda correct. We should handle promises correctly, or there will be an unhandled promise rejection for nodejs

It should be like:

```ts
promise.catch(err => logger.error(err));
```

packages/better-auth/src/telemetry/detectors/detect-database.ts
```diff
@@ -0,0 +1,28 @@
...
+	pg: "postgresql",
+	mysql: "mysql",
+	mariadb: "mariadb",
+	sqlite3: "sqlite",
+	"better-sqlite3": "sqlite",
+	"@prisma/client": "prisma",
+	mongoose: "mongodb",
+	mongodb: "mongodb",
+	"drizzle-orm": "drizzle",
+};
```
himself65, line 14: use `as const` is better

packages/better-auth/src/telemetry/detectors/detect-framework.ts
```diff
@@ -0,0 +1,30 @@
...
+	"@remix-run/server-runtime": "remix",
+	astro: "astro",
+	"@sveltejs/kit": "sveltekit",
+	"solid-start": "solid-start",
+	"tanstack-start": "tanstack-start",
+	hono: "hono",
+	express: "express",
+	elysia: "elysia",
+	expo: "expo",
+};
```
himself65, line 15: same

packages/better-auth/src/utils/package-json.ts
```diff
@@ -0,0 +1,47 @@
...
+		);
+
+		const raw = await fs.readFile(pkgJsonPath, "utf-8");
+		const json = JSON.parse(raw);
+		return json.version as string;
+	} catch {}
+	return undefined;
+}
+
+export async function getVersionFromLocalPackageJson(pkg: string) {
```
himself65: Note: The package.json version might be outdated with the real version in the lock file.

packages/better-auth/src/utils/package-json.ts
```diff
@@ -0,0 +1,47 @@
...
+import path from "node:path";
+
+export async function readPackageJson(pkg: string) {
+	try {
+		const pkgJsonPath = path.join(
+			process.cwd(),
+			"node_modules",
+			pkg,
+			"package.json",
+		);
```
himself65, line 32: This only works with `nodeLinker: node-modules` (npm, pnpm default, yarn with node_modules). 

And this will break for monorepo (packages are in the global or top-level), it's better to use require.resolve to get the real path of a package

```
const require = createRequire(import.meta.url) // ESM
const path = require.resolve(pkg)
```

Refs: https://nodejs.org/api/modules.html#requireresolverequest-options

packages/better-auth/src/telemetry/detectors/detect-system-info.ts
```diff
@@ -0,0 +1,116 @@
+import fs from "node:fs";
+import os from "node:os";
+import process from "node:process";
+
+import type { SystemInfo } from "../types";
+
+export async function detectSystemInfo(): Promise<SystemInfo> {
+	const cpus = os.cpus() || [];
```
himself65: cpus is must be array, || is unnecessary

packages/better-auth/src/init.ts
frectonz: yeah we don't want to interrupt the user with an error, we don't care if the telemetry event reporting fails, and it is wrapped with a try/catch block internally so this is not needed

packages/better-auth/src/telemetry/detectors/detect-database.ts
frectonz: we don't really care about the type it is not gonna be returned to anyone, we just use the object to loop over and check for each db driver

packages/better-auth/src/utils/package-json.ts
frectonz: I don't know how common that is but this would mean we would need to implement parsing the lock files of `yarn`, `pnpm` and `bun`. I think we can trust the version defined in the `package.json`.

packages/better-auth/src/telemetry/detectors/detect-system-info.ts
frectonz: good catch

packages/better-auth/src/init.ts
frectonz: I have prefixed it with `void` to note that we want the promise to run in the background.

packages/better-auth/src/telemetry/endpoint.ts
```diff
@@ -0,0 +1,25 @@
...
+type Logger = AuthContext["logger"];
+
+export const debugEndpoint =
+	(logger: Logger): TelemetryEndpoint =>
+	async (event) => {
+		logger.info("telemetry event", JSON.stringify(event, null, 2));
+	};
+
+const TELEMETRY_ENDPOINT = "https://telemetry.better-auth.com/v1/track";
+export const realEndpoint: TelemetryEndpoint = async (event) => {
```
Bekacru: this should be called `sendEvent` or something and the endpoint for debug and the real one should be passed as a parameter to the function instead

Bekacru: or even just move this inside create telemetry inside publish or something

packages/better-auth/src/telemetry/endpoint.ts
```diff
@@ -0,0 +1,25 @@
+import type { AuthContext } from "../types";
+import type { TelemetryEvent } from "./types";
+import { betterFetch } from "@better-fetch/fetch";
+
+export type TelemetryEndpoint = (event: TelemetryEvent) => Promise<void>;
```
Bekacru: we shouldn't need a telemetry endpoint type

packages/better-auth/src/telemetry/endpoint.ts
frectonz: I have remove the endpoint file and moved everything into `createTelemetry`.

rebasecase:
>Collects environment, system, and config info (with sensitive data redacted) on startup.

lol.

PR #3824, opened 2025-08-06
fix: client secret should be optional in configuring the generic oauth plugin

frectonz:
Not all oauth2 authorization flows require a client secret some use client assertions, this PR make the `clientSecret` field on the generic oauth plugin optional.
    

---

## Summary by cubic
Made the clientSecret field optional in the generic OAuth plugin to support providers that do not require a client secret. 

- **Bug Fixes**
  - Updated type definitions and logic to handle cases where clientSecret is not provided.

### DIFF
--- a/packages/better-auth/src/oauth2/validate-authorization-code.ts
+++ b/packages/better-auth/src/oauth2/validate-authorization-code.ts
@@ -47,7 +47,9 @@ export async function validateAuthorizationCode({
 			`${options.clientId}:${options.clientSecret ?? ""}`,
 		);
 		requestHeaders["authorization"] = `Basic ${encodedCredentials}`;
-	} else if (options.clientSecret !== undefined) {
+	}
+
+	if (options.clientSecret) {
 		body.set("client_secret", options.clientSecret);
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/oauth2/validate-authorization-code.ts
```diff
@@ -44,10 +44,10 @@ export async function validateAuthorizationCode({
...
 	if (authentication === "basic") {
 		const encodedCredentials = base64.encode(
-			`${options.clientId}:${options.clientSecret}`,
+			`${options.clientId}:${options.clientSecret ?? ""}`,
 		);
 		requestHeaders["authorization"] = `Basic ${encodedCredentials}`;
-	} else {
+	} else if (options.clientSecret !== undefined) {
 		body.set("client_secret", options.clientSecret);
 	}
```
Bekacru: let's avoid else if...
```suggestion
	}
	if(authentication !== "basic" && options.clientSecret){
	    body.set("client_secret", options.clientSecret);
	}
```

PR #3827, opened 2025-08-06
Remove session middleware from Organization Slug Check.

emersonlaurentino:
In some cases we want to create a organization in signup moment, so we need to check the slug availabe and for some reason it's authenticated, but it shouldn't.

Don't have any reason for it be authenticated route once the response is just a boolean.

---

This pull request makes a minor update to the organization routes by removing the unnecessary `requestOnlySessionMiddleware` from the middleware stack in the `checkOrganizationSlug` route. This simplifies the middleware usage and ensures only the required organization middleware is applied.

- Middleware simplification:
  * Removed `requestOnlySessionMiddleware` from the `use` array in the `checkOrganizationSlug` route, leaving only `orgMiddleware` to be used.
  * Cleaned up the import statement to remove the unused `requestOnlySessionMiddleware`.

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-org.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-org.ts
@@ -5,7 +5,7 @@ import { orgMiddleware, orgSessionMiddleware } from "../call";
 import { APIError } from "better-call";
 import { setSessionCookie } from "../../../cookies";
 import { ORGANIZATION_ERROR_CODES } from "../error-codes";
-import { getSessionFromCtx, requestOnlySessionMiddleware } from "../../../api";
+import { getSessionFromCtx } from "../../../api";
 import type { OrganizationOptions } from "../types";
 import type {
 	InferInvitation,
@@ -280,7 +280,7 @@ export const checkOrganizationSlug = <O extends OrganizationOptions>(
 					description: 'The organization slug to check. Eg: "my-org"',
 				}),
 			}),
-			use: [requestOnlySessionMiddleware, orgMiddleware],
+			use: [orgMiddleware],
 		},
 		async (ctx) => {
 			const orgAdapter = getOrgAdapter<O>(ctx.context, options);


TheUntraceable:
I have some security concerns with making this route unauthenticated. This change introduces an organization enumeration vulnerability, where an attacker could automate requests to build a list of all organizations using our platform. This is conceptually similar to why we avoid 'user not found' errors on login pages, opting for a generic 'incorrect email or password' message instead.

emersonlaurentino:
> I have some security concerns with making this route unauthenticated. This change introduces an organization enumeration vulnerability, where an attacker could automate requests to build a list of all organizations using our platform. This is conceptually similar to why we avoid 'user not found' errors on login pages, opting for a generic 'incorrect email or password' message instead.

What's the difference between username availability?

himself65 (commented):
We should probably add `allowAnonymousCheckSlug` to the option?

TheUntraceable:
> > I have some security concerns with making this route unauthenticated. This change introduces an organization enumeration vulnerability, where an attacker could automate requests to build a list of all organizations using our platform. This is conceptually similar to why we avoid 'user not found' errors on login pages, opting for a generic 'incorrect email or password' message instead.
> 
> What's the difference between username availability?

They both risk enumeration but the nature and impact differ. Usernames *can* be personally identifiable (John.Smith123), but can also be random (AmazingGreatVibeCoder123). Organization slugs are typically much more identifiable (acme-corp, vibe-coders-inc).  I don't think this should affect all apps using better auth. Adding an option in the plugin is a good option for this I guess but also opens the door for adding more precise options that would add bloat. Would be open to hearing any other thoughts on this

emersonlaurentino:
I like the approach of has an option in the plugin to turn this part a public path.

himself65:
> I like the approach of has an option in the plugin to turn this part a public path.

yeah, please update the code. i appreciate it

PR #3830, opened 2025-08-06
fix: secondary storage session deletion

dvanmali:
Deletion of a session removed its token-based secondary storage but left a hanging token in the user-based secondary storage.

Fix: deletion of session also removes token on user-based secondary storage.
    

---

## Summary by cubic
Fixed session deletion so that tokens are now removed from both token-based and user-based secondary storage, preventing leftover tokens after a session is deleted.


dvanmali:
@cubic-dev-ai review

Bekacru:
I think this is duplicate of #3820

dvanmali:
Thanks @Bekacru just saw it, closing

PR #3835, opened 2025-08-06
fix(organization): fix convert emails to lower case

gwoodbridge:
Both the findMemberByEmail and findPendingInvitation functions can accept email addresses that haven't been converted to lowercase. Currently searching for a member or pending invitations with any uppercase characters won't find the appropriate user or invitations.


himself65 (commented):
can we have a test case for this?

PR #3836, opened 2025-08-06
fix: ttl sessions list expiration

dvanmali:
- Logic for ttl expiration should mimic existing `expiresAt` sessions list.
- Secondary storage never added token to secondary storage when calling `createSession`

Partial Fix: #4034, #4472 
Additional PR: #4961

---

## Summary by cubic
Updated session list logic to remove expired sessions and set TTL based on the latest session expiration.

### DIFF
--- a/packages/better-auth/src/db/internal-adapter.test.ts
+++ b/packages/better-auth/src/db/internal-adapter.test.ts
@@ -1,17 +1,25 @@
 import { beforeAll, expect, it, describe, vi, afterEach } from "vitest";
-import type { BetterAuthOptions, BetterAuthPlugin } from "../types";
+import type {
+	BetterAuthOptions,
+	BetterAuthPlugin,
+	GenericEndpointContext,
+	Session,
+	User,
+} from "../types";
 import Database from "better-sqlite3";
 import { init } from "../init";
 import { betterAuth } from "../auth";
 import { getMigrations } from "./get-migration";
 import { Kysely, SqliteDialect } from "kysely";
 import { getTestInstance } from "../test-utils/test-instance";
+import { safeJSONParse } from "../utils/json";
 
 describe("adapter test", async () => {
 	const sqliteDialect = new SqliteDialect({
 		database: new Database(":memory:"),
 	});
 	const map = new Map();
+	const expirationMap = new Map();
 	let id = 1;
 	const hookUserCreateBefore = vi.fn();
 	const hookUserCreateAfter = vi.fn();
@@ -31,12 +39,14 @@ describe("adapter test", async () => {
 		secondaryStorage: {
 			set(key, value, ttl) {
 				map.set(key, value);
+				expirationMap.set(key, ttl);
 			},
 			get(key) {
 				return map.get(key);
 			},
 			delete(key) {
 				map.delete(key);
+				expirationMap.delete(key);
 			},
 		},
 		advanced: {
@@ -102,6 +112,7 @@ describe("adapter test", async () => {
 	});
 	afterEach(async () => {
 		vi.clearAllMocks();
+		map.clear();
 	});
 	const ctx = await init(opts);
 	const internalAdapter = ctx.internalAdapter;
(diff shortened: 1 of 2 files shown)


Bekacru:
thanks for the PR! can we add tests for this?

dvanmali:
Fixes #4034

packages/better-auth/src/db/internal-adapter.ts
```diff
@@ -320,16 +320,59 @@ export const createInternalAdapter = (
 									list = list.filter((session) => session.expiresAt > now);
 								}
 
-								list.push({
+								const filtered = [];
```
Bekacru, line 323: this should have some sort type annotation

packages/better-auth/src/db/internal-adapter.test.ts
```diff
@@ -411,4 +422,104 @@ describe("adapter test", async () => {
 		expect(finalTTL).toBeLessThanOrEqual(7199);
 		expect(finalTTL).toBeGreaterThanOrEqual(7198); // Allow for test execution time
 	});
+
+	it("should create on secondary storage", async () => {
+		// Create session
```
Bekacru, line 427: generally we should avoid writing too many comments

packages/better-auth/src/db/internal-adapter.ts
```diff
@@ -320,16 +320,59 @@ export const createInternalAdapter = (
...
+								let furthestSessionExp = now;
+								for (const item of list) {
+									if (item.expiresAt > now) {
+										filtered.push(item);
+										if (furthestSessionExp < item.expiresAt) {
+											furthestSessionExp = item.expiresAt;
+										}
+									}
+								}
```
Bekacru, line 332: why do we need this where we're already filtering here `list = list.filter((session) => session.expiresAt > now);`

Bekacru: For `furthestSessionExp` it can be the part of the filter callback

packages/better-auth/src/db/internal-adapter.ts
```diff
@@ -320,16 +320,59 @@ export const createInternalAdapter = (
...
+
+								filtered.push({
 									token: data.token,
-									expiresAt: now + sessionExpiration * 1000,
+									expiresAt: data.expiresAt.getTime(),
 								});
+								if (furthestSessionExp < data.expiresAt.getTime()) {
```
Bekacru, line 338: I think you can refactor this by doing `.sort` on list on expiresAt and taking the first index instead of all these change

Bekacru:
```ts
const furthestSessionExp = list.sort((a, b)=> a.expiresAt.getTime() -  b.expiresAt.getTime())[0].getTime();
//get the ttl....
```

packages/better-auth/src/db/internal-adapter.ts
```diff
@@ -541,13 +584,28 @@ export const createInternalAdapter = (
 					if (currentList) {
 						let list: { token: string; expiresAt: number }[] =
 							safeJSONParse(currentList) || [];
-						list = list.filter((s) => s.token !== token);
+						const now = Date.now();
+
+						const filtered = [];
```
Bekacru, line 589: here we need type annontation

packages/better-auth/src/db/internal-adapter.test.ts
dvanmali: Generally I'm a fan to comment as needed - this organizes steps into breakable sections especially if it is not broken into task-specific functions. Rollup build won't put test files in the build.

packages/better-auth/src/db/internal-adapter.ts
dvanmali: This is different, this ensures the secondary session expiration is set to the last/furthest expiring session in the storage list to prevent it from dying early.

packages/better-auth/src/db/internal-adapter.ts
dvanmali: Good catch 👍

packages/better-auth/src/db/internal-adapter.ts
dvanmali: I'll use this syntax, not sure why i didn't do this in the first place

packages/better-auth/src/db/internal-adapter.ts
Bekacru: I'm asking about `filtered` and for furthestSession please use `.sort` as suggested below

dvanmali: expires filter is performed only on set and list functions. I added it on delete to remove sessions if i'm already reading through the list anyway, then I should also filter

varundeva:
Hey @dvanmali  

which issue this will fix? 
"better-auth": "^1.3.18", is using now and noticed those few records are there in redis database without TTL 
image
can someone help me to rectify this? 
because if 100 user logged in then 100 different ips with multiple records comes in redis

cc @Bekacru

dvanmali:
Apologies @varundeva. This fixed ttl for sessions and session tokens stored on secondary storage. PR #4961 will additionally fix rate limit requests on secondary storage. Please wait for that to merge as well.

varundeva:
> Apologies @varundeva. This fixed ttl for sessions and session tokens stored on secondary storage. PR #4961 will additionally fix rate limit requests on secondary storage. Please wait for that to merge as well.

Yes
Another key which I'm showing on screenshot also creating memory issue
But I'm not getting from where it is setting

Issue #2062, opened 2025-03-31
Duplicate social sign-on accounts via `authClient.linkSocial` is allowed


stefan1575:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Create a new user that uses `emailAndPassword` authentication using `authClient.signUp.email`
2. While logged in, Link an account using the Google social provider using `authClient.linkSocial`.
3. Invoke another `authClient.linkSocial` call, with a successful linked google account.

### Current vs. Expected behavior

Following the steps from the previous section:

- The additional `authClient.linkSocial` call to and either fail silently/throw an error, 
- But I observed that the additional `authClient.linkSocial` calls still succeed and creates multiple accounts in the `accounts` table generated by the better-auth cli resulting duplicate accounts despite having the same `user_id`.

### What version of Better Auth are you using?

1.2.5

### Provide environment information

```bash
- OS: Windows 10
- Browser: Google Chrome
```

### Which area(s) are affected? (Select all that apply)

Backend, Client

### Auth config (if applicable)

```typescript
import { db } from "@/db";
import * as schema from "@/db/schema";
import { betterAuth } from "better-auth";
import { drizzleAdapter } from "better-auth/adapters/drizzle";

export const auth = betterAuth({
  database: drizzleAdapter(db, {
    provider: "pg",
    schema: schema, // this comes from the better-auth cli
  }),
  account: {
    accountLinking: {
      enabled: true,
    },
  },
  emailAndPassword: {
    enabled: true,
  },
  socialProviders: {
    google: {
      clientId: process.env.GOOGLE_CLIENT_ID!,
      clientSecret: process.env.GOOGLE_CLIENT_SECRET!,
      redirectURI: process.env.BETTER_AUTH_URL! + "/api/auth/callback/google",
    },
  },
});
```

Kinfe123:
Pls refer to this pr - #1803

stefan1575:
> Pls refer to this pr - [#1803](https://github.com/better-auth/better-auth/pull/1803)

The relevant tables for this discussion are `user` and `account`.

I guess I could have worded it better :sweat_smile:.

What I meant was in the same `user`, I could register the _same_ google account multiple times and it would reflect on the database where I have two `user@gmail.com`. 

To illustrate this, one user may have two duplicate google accounts which is shown by the `account_id`. 

![Image](https://github.com/user-attachments/assets/6835fde3-ed19-4826-b73e-7ac6a098759e)

[...]

PR #3839, opened 2025-08-06
fix: avoid general oauth flow duplicate user

himself65:
Fixes: https://github.com/better-auth/better-auth/pull/3816#issuecomment-3161748770, 
Upstream: https://github.com/better-auth/better-auth/issues/3814, https://github.com/better-auth/better-auth/issues/2062

Update the general OAuth plugin to support numbers as user.id and reduce any type in the codebaes
    

---

## Summary by cubic
Updated the generic OAuth plugin to support numeric user IDs and prevent duplicate user accounts when OAuth providers return numbers as IDs.

- **Bug Fixes**
  - Ensured all user and account IDs are handled as strings to avoid duplicate accounts.
  - Added tests for numeric user IDs in different OAuth flows.

### DIFF
--- a/packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
@@ -10,7 +10,8 @@ import { parseSetCookieHeader } from "../../cookies";
 
 let server = new OAuth2Server();
 
-describe("oauth2", async () => {
+// each test suite relies on the same port, so we need to enable sequential execution
+describe.sequential("oauth2", async () => {
 	const providerId = "test";
 	const clientId = "test-client-id";
 	const clientSecret = "test-client-secret";
--- a/packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
@@ -766,4 +766,99 @@ describe("oauth2", async () => {
 
 		expect(accounts[0].accountId).toBe(String(numericProfileId));
 	});
+
+	it("should handle Strava OAuth with custom mapProfileToUser", async () => {
+		const stravaUserId = 12345678;
+		const stravaProfile = {
+			id: stravaUserId,
+			firstname: "John",
+			lastname: "Doe",
+			profile: "https://example.com/strava-avatar.jpg",
+			email_verified: true,
+		};
+
+		server.service.once("beforeUserinfo", (userInfoResponse) => {
+			userInfoResponse.body = stravaProfile;
+			userInfoResponse.statusCode = 200;
+		});
+
+		const { customFetchImpl, auth } = await getTestInstance({
+			plugins: [
+				genericOAuth({
+					config: [
+						{
+							providerId: "strava",
+							authorizationUrl: `http://localhost:${port}/authorize`,
+							tokenUrl: `http://localhost:${port}/token`,
+							userInfoUrl: `http://localhost:${port}/userinfo`,
+							clientId: "STRAVA_CLIENT_ID",
+							clientSecret: "STRAVA_CLIENT_SECRET",
+							scopes: ["read", "activity:read_all"],
+							pkce: true,
+							mapProfileToUser: (profile) => {
+								const fullName = `${profile.firstname} ${profile.lastname}`;
+								return {
+									id: profile.id,
+									email: `${profile.id}@strava.local`,
+									name: fullName,
+									image: profile.profile,
+									emailVerified: true,
+								};
+							},
+						},
+					],
+				}),
+			],
+		});
+
+		const authClient = createAuthClient({
+			plugins: [genericOAuthClient()],
+			baseURL: "http://localhost:3000",
+			fetchOptions: {
+				customFetchImpl,
+			},
+		});
+
+		const signInRes = await authClient.signIn.oauth2({
+			providerId: "strava",
+			callbackURL: "http://localhost:3000/dashboard",
+			newUserCallbackURL: "http://localhost:3000/new_user",
+		});
+
+		expect(signInRes.data?.url).toContain(`http://localhost:${port}/authorize`);
+		// we missed the `activity:read_all`
+		expect(signInRes.data?.url).toContain("scope=read+activity");
+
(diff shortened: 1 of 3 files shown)


packages/better-auth/src/plugins/generic-oauth/index.ts
```diff
@@ -736,13 +727,15 @@ export const genericOAuth = (options: GenericOAuthOptions) => {
 					}
 
 					const result = await handleOAuthUserInfo(ctx, {
+						// @ts-expect-error userInfo and mapUser type mismatch to User
```
himself65: We should have a stronger type in the codebase, I would prefer to add ts-expect-error here instead of `as any`

packages/better-auth/src/plugins/generic-oauth/index.ts
```diff
@@ -736,13 +727,15 @@ export const genericOAuth = (options: GenericOAuthOptions) => {
...
 						userInfo: {
 							...userInfo,
 							...mapUser,
+							id: mapUser.id ? String(mapUser.id) : String(userInfo.id),
 						},
 						account: {
 							providerId: provider.providerId,
-							accountId: userInfo.id,
+							accountId: String(userInfo.id),
```
Bekacru: here also let's do `mapUser.id ? String(mapUser.id) : String(userInfo.id)`

packages/better-auth/src/plugins/generic-oauth/index.ts
Bekacru: why is there a mismatch?

cemreinanc:
here is the config example I tried with:

```
 // ...
  plugins: [
    genericOAuth({
      config: [
        {
          providerId: 'strava',
          clientId: env.STRAVA_CLIENT_ID,
          clientSecret: env.STRAVA_CLIENT_SECRET,
          authorizationUrl: 'https://www.strava.com/oauth/authorize',
          tokenUrl: 'https://www.strava.com/oauth/token',
          userInfoUrl: 'https://www.strava.com/api/v3/athlete',
          scopes: [['read', 'activity:read', 'activity:read_all'].join(',')], // some workaround for strava scopes only works with comma delimited but plugin sends these with + sign. (their bad probably not standard) 
          mapProfileToUser(profile) {
            return {
              name: `${profile.firstname} ${profile.lastname}`,
              image: profile.profile,
              email: `${profile.id}@users.strava.com`, // fake email just to test registration with it as strava doesnt return user email address
            };
          },
        },
      ],
    }),
  ],
});
```

maybe I can write a strava social provider and send pr for it after discovering all other quirks

himself65:
> here is the config example I tried with:
> 
> ```
> export const auth = betterAuth({
>  // ...
>   plugins: [
>     genericOAuth({
>       config: [
>         {
>           providerId: 'strava',
>           clientId: env.STRAVA_CLIENT_ID,
>           clientSecret: env.STRAVA_CLIENT_SECRET,
>           authorizationUrl: 'https://www.strava.com/oauth/authorize',
>           tokenUrl: 'https://www.strava.com/oauth/token',
>           userInfoUrl: 'https://www.strava.com/api/v3/athlete',
>           scopes: [['read', 'activity:read', 'activity:read_all'].join(',')], // some workaround for strava scopes only works with comma delimited but plugin sends these with + sign. (their bad probably not standard) 
>           mapProfileToUser(profile) {
>             return {
>               name: `${profile.firstname} ${profile.lastname}`,
>               image: profile.profile,
>               email: `${profile.id}@users.strava.com`, // fake email just to test registration with it as strava doesnt return user email address
>             };
>           },
>         },
>       ],
>     }),
>   ],
> });
> ```
> 
> maybe I can write a strava social provider and send pr for it after discovering all other quirks

yeah sure that would be help, thank u

packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
```diff
@@ -8,22 +8,26 @@ import { betterFetch } from "@better-fetch/fetch";
...
+		await server.stop();
+	});
 
 	const { customFetchImpl, auth } = await getTestInstance({
 		plugins: [
 			genericOAuth({
 				config: [
 					{
 						providerId,
-						discoveryUrl:
```
himself65, line 24: `server.issuer.url` is wrong

packages/better-auth/src/plugins/generic-oauth/index.ts
himself65:
```
Types of property email are incompatible.
Type string | null | undefined is not assignable to type string
Type undefined is not assignable to type string

```

Bekacru: we should probably make sure email isn't null or undefined before passing it to the the handler

himself65: yeah let me see how to make that

himself65: It's not possible right now, `getUserInfo` is OAuthUserInfo, where only `id` is required. This makes sense because some social providers may not require email.

Then `mapProfileToUser` is an optional function

So the email here could be undefined. How do you think

himself65: Oh sorry you are correct, it's there

https://github.com/better-auth/better-auth/blob/99b111835f580d2649b16857c537a616aac57a16/packages/better-auth/src/plugins/generic-oauth/index.ts#L662

himself65: updated

packages/better-auth/src/plugins/generic-oauth/index.ts
```diff
@@ -660,32 +648,51 @@ export const genericOAuth = (options: GenericOAuthOptions) => {
...
-					}
-					const mapUser = provider.mapProfileToUser
-						? await provider.mapProfileToUser(userInfo)
-						: userInfo;
-					if (!mapUser?.email) {
-						ctx.context.logger.error("Unable to get user info", userInfo);
-						throw redirectOnError("email_is_missing");
-					}
+					const userInfo: Omit<User, "createdAt" | "updatedAt"> =
+						await (async function handleUserInfo() {
```
himself65, line 652: Use IIFE here to isolate the variable

himself65:
@cemreinanc thank u I added you input as a test case

PR #3842, opened 2025-08-06
chore: enable `noFloatingPromises` lint rule

himself65:
Upstream comment: https://github.com/better-auth/better-auth/pull/3822#discussion_r2258508346
    

## Summary by cubic
Enabled the `noFloatingPromises` lint rule to catch unhandled promise rejections and updated codebase to handle or explicitly ignore floating promises.

- **Refactors**
  - Added error handling or lint ignores for floating promises in API key routes, CLI entrypoint, and UI components.
  - Updated tests and utility functions to await or handle async calls as needed.

### DIFF
--- a/packages/better-auth/src/adapters/test.ts
+++ b/packages/better-auth/src/adapters/test.ts
@@ -98,9 +98,9 @@ async function adapterTest(
 	test.skipIf(disabledTests?.CREATE_MODEL)(
 		`${testPrefix ? `${testPrefix} - ` : ""}${adapterTests.CREATE_MODEL}`,
 		async ({ onTestFailed }) => {
-			resetDebugLogs();
-			onTestFailed(() => {
-				printDebugLogs();
+			await resetDebugLogs();
+			onTestFailed(async () => {
+				await printDebugLogs();
 			});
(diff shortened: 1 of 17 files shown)


packages/better-auth/src/adapters/test.ts
```diff
@@ -54,7 +54,7 @@ const numberIdAdapterTests = {
 	SHOULD_RETURN_A_NUMBER_ID_AS_A_RESULT:
 		"Should return a number id as a result",
 	SHOULD_INCREMENT_THE_ID_BY_1: "Should increment the id by 1",
-} as const;
+};
```
Bekacru: why remove as const?

himself65: IDE blames me for deleting a variable on a constant object. I think this is from my default ESLint setup? anyway `as const` is not necessary anyway

image

packages/better-auth/src/plugins/api-key/routes/create-api-key.ts
himself65, line 1: Maybe we should await all `deleteAllExpiredApiKeys`? I don't know what's the design is behind this

PR #3845, opened 2025-08-07
feat(oidc): pass oauth consent code via query params

grant0417:
This PR adds support for passing consent codes via URL parameters in addition to the existing cookie-based approach.

## What changed

The `/oauth2/consent` endpoint now accepts the consent code in two ways:

```typescript
// New: URL parameter method
const res = await client.oauth2.consent({
  accept: true,
  consent_code: consentCode // from URL params
});

// Existing: Cookie-based method  
const res = await client.oauth2.consent({
  accept: true
  // consent_code read from cookie
});
```

When redirecting to the consent page, we now include `consent_code` in the URL:
```
/consent?consent_code=abc123&client_id=...&scope=...
```

## Why?

The cookie-based approach doesn't work well in environments with third-party cookie restrictions (iframe).

## Breaking changes

None - existing implementations continue to work as before.
    

---

## Summary by cubic
Added support for passing the OAuth consent code via URL parameters, in addition to the existing cookie-based method. This improves compatibility with environments that block third-party cookies, such as iframes.

- **New Features**
  - The `/oauth2/consent` endpoint now accepts the consent code from either a URL parameter or a cookie.
  - Consent page redirects now include the `consent_code` in the URL.
  - Documentation updated to show both consent code methods.

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/authorize.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/authorize.ts
@@ -271,8 +271,11 @@ export async function authorize(
 		});
 
 		// Pass the consent code as a URL parameter to support URL-based consent flows
-		// This enables both cookie-based and URL parameter-based implementations
-		const consentURI = `${options.consentPage}?consent_code=${encodeURIComponent(code)}&client_id=${encodeURIComponent(client.clientId)}&scope=${encodeURIComponent(requestScope.join(" "))}`;
+		const urlParams = new URLSearchParams();
+		urlParams.set("consent_code", code);
+		urlParams.set("client_id", client.clientId);
+		urlParams.set("scope", requestScope.join(" "));
+		const consentURI = `${options.consentPage}?${urlParams.toString()}`;
 
 		return handleRedirect(consentURI);
 	}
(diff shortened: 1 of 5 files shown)


packages/better-auth/src/plugins/oidc-provider/oidc.test.ts
```diff
@@ -233,7 +233,7 @@ describe("oidc", async () => {
 				newHeaders.append("Cookie", headers.get("Cookie") || "");
 			},
 		});
-		expect(redirectURI).toContain("/oauth2/authorize?client_id=");
+		expect(redirectURI).toContain("/oauth2/authorize?consent_code=");
```
himself65: Can we make these two separate test cases, covering both `client_id` and `consent_code`?

packages/better-auth/src/plugins/oidc-provider/authorize.ts
```diff
@@ -263,14 +263,16 @@ export async function authorize(
...
 			path: "/",
 			sameSite: "lax",
 		});
-		const consentURI = `${options.consentPage}?client_id=${
-			client.clientId
-		}&scope=${requestScope.join(" ")}`;
+
+		// Pass the consent code as a URL parameter to support URL-based consent flows
+		// This enables both cookie-based and URL parameter-based implementations
+		const consentURI = `${options.consentPage}?consent_code=${encodeURIComponent(code)}&client_id=${encodeURIComponent(client.clientId)}&scope=${encodeURIComponent(requestScope.join(" "))}`;
```
himself65: Should we use URLSearchParams here?

Refs: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

himself65 (approved):
LGTM

PR #3850, opened 2025-08-07
fix(client): properly sync session state when activeOrg changes

Kinfe123:


docs/components/landing/section-svg.tsx
himself65, line 1: why there's a change

packages/better-auth/src/plugins/organization/client.ts
```diff
@@ -161,6 +161,32 @@ export const organizationClient = <CO extends OrganizationClientOptions>(
...
+									...currentSession.data.session,
+									activeOrganizationId: data.organizationId,
+								},
+							},
+						};
+
+						sessionAtom.set(updatedSession);
+					}
+
+					$store.notify("$activeOrgSignal");
```
himself65, line 186: note: I think the string-based atom system is not good.

PR #3866, opened 2025-08-07
chore: `noConsole` path based lint rule

Kinfe123:
This PR introduces a new noConsole: "error" rule that removes all console.log statements when running pnpm lint:fix. The rule is path-based, applying to all files within the core package, including tests, with specific exceptions configured through Biome where necessary.


himself65:
Can you rebase the code

himself65:
Way too many conflicts now. I think the issue here is that some tests need `console.log`, and some scripts in the demo and docs also need that. And CLI now also has `console.log`

I'm closing this now, but feel free to re-open it when you fix the issues

PR #3867, opened 2025-08-07
feat: schema check for user side input

himself65:
This is my idea for validating user-side input. We should have more of both type checking and runtime checking to make sure the logic is working correctly.

    

---

## Summary by cubic
Added schema validation for user input in getSchema to catch invalid configurations at runtime.

- **Bug Fixes**
 - Throws a clear error if the user or plugin schema is invalid.
 - Added tests to cover valid and invalid schema cases.


himself65:
This is maybe unnecessary

Issue #3855, opened 2025-08-07
Multi Session listDeviceSessions to respect customSession


BjoernRave:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

I just started using the multi session plugin and when calling `authClient.multiSession.listDeviceSessions()` what I configured in `customSession` does not get respected.

It would be nice if it would

### Describe the solution you'd like

Calling `listDeviceSession` to return the customized sessions

### Describe alternatives you've considered

-

PR #3868, opened 2025-08-07
fix(custom-session): also mutate multi-session response

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/3855

### DIFF
--- a/packages/better-auth/src/plugins/custom-session/index.ts
+++ b/packages/better-auth/src/plugins/custom-session/index.ts
@@ -1,12 +1,19 @@
 import * as z from "zod/v4";
-import { createAuthEndpoint, getSession } from "../../api";
+import {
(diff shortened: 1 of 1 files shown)


himself65 (approved):

Bekacru:
I think this should be opt-in, because listing device sessions is mostly used to show a user’s profile image and name so they can switch. Retrieving the entire session data always feels unnecessary

ping-maxwell:
> I think this should be opt-in, because listing device sessions is mostly used to show a user’s profile image and name so they can switch. Retrieving the entire session data always feels unnecessary

Understandable, will do.

ping-maxwell:
Added opt-in plugin options

himself65 (approved):

PR #3869, opened 2025-08-07
feat: support `node:sqlite`

himself65:
I'm a Node.js SQLite WG member, so I'm happy to integrate this into better-auth. Even though it's not stable yet but it's okay to do sth like testing and small demo

Document: https://nodejs.org/docs/v22.18.0/api/sqlite.html#sqlite
    

---

## Summary by cubic
Added support for the experimental Node.js built-in SQLite driver (`node:sqlite`) in Better Auth, allowing users to connect using DatabaseSync. Updated documentation and added tests for the new dialect.

- **New Features**
 - Implemented a NodeSqliteDialect for Kysely to support `node:sqlite`.
 - Added integration tests to verify compatibility with Better Auth.
 - Updated docs to show usage examples for both better-sqlite3 and node:sqlite.

### DIFF
--- a/demo/nextjs/package.json
+++ b/demo/nextjs/package.json
@@ -92,7 +92,7 @@
   },
   "devDependencies": {
     "@types/canvas-confetti": "^1.9.0",
-    "@types/node": "^24.2.0",
+    "@types/node": "^24.3.0",
     "@types/react": "^19.0.0",
     "@types/react-dom": "^19.0.0",
     "@types/three": "^0.168.0",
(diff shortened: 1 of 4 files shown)


Bekacru:
lets go

docs/content/docs/adapters/sqlite.mdx
```diff
@@ -20,6 +23,28 @@ export const auth = betterAuth({
 });
 ```
 
+### Node.js Built-in SQLite (Experimental)
```
Bekacru, line 31: can we also add Bun sqlite example. It's already supported

PR #3870, opened 2025-08-08
chore: bump zod deps across the workspaces

Kinfe123:
## Summary by cubic
Upgraded all workspaces to use zod v4 and updated imports to reference the new version.


Bekacru:
I think we should also use `catalogs` for zod

PR #3873, opened 2025-08-08
Use defaultNow() for drizzle timestamp fields

Badbird5907:
This PR modifies the Drizzle schema generator to use `.defaultNow()` instead of `.defaultFn(() => new Date())` for timestamp fields that are defaulted to the current date.

I haven't been able to test this on a real project, however I have modified the tests with the expected results, and everything looks ok.
    

---

## Summary by cubic
Updated the Drizzle schema generator to use .defaultNow() for timestamp fields instead of .defaultFn(() => new Date()), making default values clearer and more consistent.

### DIFF
--- a/packages/cli/src/generators/drizzle.ts
+++ b/packages/cli/src/generators/drizzle.ts
@@ -153,11 +153,14 @@ export const generateDrizzleSchema: SchemaGenerator = async ({
 								typeof attr.defaultValue !== "undefined"
 							) {
 								if (typeof attr.defaultValue === "function") {
-                  if (attr.type === "date" && attr.defaultValue.toString().trim() === "new Date()") {
-                    type += `.defaultNow()`;
-                  } else {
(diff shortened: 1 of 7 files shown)


himself65:
please fix the lint

Kinfe123:
here defaultNow() will only affect the where the current timestamp should be defined means either on the app or db level. unless you dont have any thing outside the application logic affecting the row, you should be fine. if you have a case for that how are you using this ?

himself65:
I think there's no obvious difference with defaultFn, but I think using internal function is a better way

Badbird5907:
> I think there's no obvious difference with defaultFn, but I think using internal function is a better way

To clarify, this does change drizzle's behavior. The generated SQL migration by drizzle will use `DEFAULT now()`.
With `defaultFn`, drizzle handles populating the columns (instead of the database)

image

Kinfe123:
There is indeed a difference.

PR #3890, opened 2025-08-08
feat: add query parameter to useSession().refetch() for cache control consistency

### LINKED ISSUE #3889: Add query parameter support to `useSession().refetch()` for cache control consistency

Currently, `getSession()` supports query parameters like `{ disableCookieCache: true }`, but `useSession().refetch()` doesn't. This creates an inconsistent API where I have to import and use both methods just to bypass cookie cache after updating user data. When using optimistic updates with cookie caching enabled, the session remains stale even after calling `refetch()`, requiring a workaround with `getSession({ query: { disableCookieCache: true }})`.

I'd like `useSession().refetch()` to accept the same query parameters as `getSession()`:

```ts
const { refetch } = useSession();
await refetch({ query: { disableCookieCache: true }});
```

This would provide API consistency and eliminate the need to import both `useSession()` and `getSession()` for cache invalidation.

Current workaround:

```ts
import { useSession, getSession } from '@/lib/auth-client';

const { data: session, refetch } = useSession();
// After mutation...
const newSession = await getSession({ query: { disableCookieCache: true }});
```

Other alternatives considered:

- Disabling cookie cache entirely (loses performance benefits)
- Using shorter cache duration (still has timing issues)
- Manual state management (defeats the purpose of the session hook)

adriandlam:
Closes #3889

This PR adds query parameter (`disableCookieCache` and `disableRefresh`) to `useSession().refetch()` for consistency with `getSession()`. 

Currently, `getSession()` supports query parameters like `{ disableCookieCache: true }`, but `useSession().refetch()` doesn't, creating an inconsistent API where developers need to import both methods.

Now you can use:

```ts
const { refetch } = useSession();
await refetch({ query: { disableCookieCache: true }});
```

### DIFF
--- a/packages/better-auth/src/client/client.test.ts
+++ b/packages/better-auth/src/client/client.test.ts
@@ -306,4 +306,23 @@ describe("type", () => {
 			} | null>
 		>();
 	});
+	it("should support refetch with query parameters", () => {
+		const client = createReactClient({
+			plugins: [testClientPlugin()],
+			baseURL: "http://localhost:3000",
+			fetchOptions: {
+				customFetchImpl: async (url, init) => {
+					return new Response();
+				},
+			},
+		});
+
+		type UseSessionReturn = ReturnType<typeof client.useSession>;
+		expectTypeOf<UseSessionReturn>().toMatchTypeOf<{
+			data: any;
+			isPending: boolean;
+			error: BetterFetchError | null;
+			refetch: (queryParams?: { query?: Record<string, any> }) => void;
+		}>();
+	});
 });
--- a/packages/better-auth/src/client/query.ts
+++ b/packages/better-auth/src/client/query.ts
@@ -27,18 +27,18 @@ export const useAuthQuery = <T>(
 		error: null | BetterFetchError;
 		isPending: boolean;
 		isRefetching: boolean;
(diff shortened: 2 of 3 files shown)


packages/better-auth/src/client/query.ts
himself65, line 1: should have a stronger type

packages/better-auth/src/client/client.test.ts
```diff
@@ -306,4 +306,23 @@ describe("type", () => {
...
+			fetchOptions: {
+				customFetchImpl: async (url, init) => {
+					return new Response();
+				},
+			},
+		});
+
+		type UseSessionReturn = ReturnType<typeof client.useSession>;
+		expectTypeOf<UseSessionReturn>().toMatchTypeOf<{
+			data: any;
```
TheUntraceable: Shouldn't this be of type Session?

adriandlam: Yup, missed this one. Thanks

frectonz:
Please revert the reformatting in some tests. You can run `pnpm format` to restore the original formatting.

himself65 (approved):
I will merge this for now. But we will have a better client side hook in the future

PR #3902, opened 2025-08-10
fix(api): replace defu array concatenation with override in hook context merges

### LINKED ISSUE #3126: Sign up additionalFields with array type items gets duplicated when using before hooks

### To Reproduce

```js
const auth = betterAuth({
  user: {
    additionalFields: {
      tags: {
        type: "string[]",
        required: true
      }
    }
  },
  hooks: {
      if (ctx.path !== "/sign-up/email") {
        return
      }
      return {
        context: {
          ...ctx,
          body: {
            ...ctx.body,
            someOtherField: "someValue", // Example of additional field
          },
        }
      }
    }),
  },
})
```

`tags` array gets duplicated in the database (might affect other features)

`defu` dependency concatenates array values by default
See sample:
```
console.log(defu({ array: ["b", "c"] }, { array: ["a"] }));
// => { array: ['b', 'c', 'a'] }
```

Problematic code in better-auth source:
https://github.com/better-auth/better-auth/blob/2fe1be0e5f023be77b07206dd960f9fd38abba8c/packages/better-auth/src/api/to-auth-endpoints.ts#L73

Or here:
https://github.com/better-auth/better-auth/blob/2fe1be0e5f023be77b07206dd960f9fd38abba8c/packages/better-auth/src/api/to-auth-endpoints.ts#L157

Expected: I expect array items to not get duplicated.

1.2.7

```bash
chrome
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

Kinfe123:
This pr fixes on the Use a custom defu merger (createDefu) that replaces arrays instead of concatenating
when merging contexts from before hooks into internalContext/modifiedContext.
Prevents duplication for array additionalFields (e.g., user.tags).

Closes #3126

### DIFF
--- a/packages/better-auth/src/api/to-auth-endpoints.test.ts
+++ b/packages/better-auth/src/api/to-auth-endpoints.test.ts
@@ -117,6 +117,39 @@ describe("before hook", async () => {
 				key: "value",
 			});
 		});
+		it("should replace existing array when hook provides another array", async () => {
+			const endpoints3 = {
+				body: createAuthEndpoint(
+					"/body-array-replace",
+					{ method: "POST" },
+					async (c) => c.body as any,
+				),
+			};
+
+			const authContext3 = init({
+				hooks: {
+					before: createAuthMiddleware(async (c) => {
+						if (c.path === "/body-array-replace") {
+							return {
+								context: {
+									body: {
+										tags: ["a"],
+									},
+								},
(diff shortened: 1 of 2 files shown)


Bekacru:
closed in favor of #3160

Issue #2970, opened 2025-06-10
[Bug] Duplicate cookies causing "Invalid Base64 character" error with @better-auth/expo


quuentinho:
### Is this suited for github?

- [x] Yes, this is suited for github

## Environment
- **Better Auth version**: `1.2.8-beta.3` (web) / `1.2.8` (expo)
- **Framework**: Next.js 15.3.1 + Expo SDK 53
- **Platform**: React Native iOS (testing on simulator)

## Problem Description

I'm experiencing an issue where duplicate `better-auth.session_data` cookies are being sent from my Expo app to my Next.js API, causing a "Invalid Base64 character: ," error on the server side.

### Next.js API (working fine with web frontend)
```javascript
// src/app/api/mobile/me/route.js
export async function GET(request) {
    const session = await auth.api.getSession({
        headers: request.headers,
        query: {
            disableCookieCache: true
        }
    })
    
    if (!session) {
        return NextResponse.json({ error: "Unauthorized" }, { status: 401 })
    }
    
    // ... rest of the code
}
```

### Expo Auth Client
```javascript
// lib/auth-client.js
import { createAuthClient } from "better-auth/react"
import { expoClient } from "@better-auth/expo/client"
import * as SecureStore from "expo-secure-store"
import Constants from 'expo-constants'

const baseURL = Constants.expoConfig.extra.API_URL

export const authClient = createAuthClient({
    baseURL: baseURL,
    plugins: [
        expoClient({
            scheme: "myapp",
            storagePrefix: "myapp",
            storage: SecureStore
        })
    ]
})
```

### Axios Configuration
```javascript
// lib/axios.js
import axios from 'axios'
import Constants from 'expo-constants'
import { authClient } from './auth-client'

const baseURL = Constants.expoConfig.extra.API_URL

const axiosInstance = axios.create({
  baseURL,
  headers: {
    "Cookie": authClient.getCookie()
  }
})

export default axiosInstance
```

### Mobile Side (authClient.getCookie() output):
```javascript
{
  "better-auth.session_token": {
    "value": "TOKEN_VALUE_HERE",
    "expires": "2025-06-10T12:03:53.703Z"
  },
  "better-auth.session_data": {
    "value": "ENCODED_SESSION_DATA_1",
    "expires": "2025-06-10T12:07:27.251Z"
  },
  "better-auth.dont_remember": {
    "value": "",
    "expires": "2025-06-10T11:53:45.571Z"
  }
}
```

[...]

PR #3909, opened 2025-08-10
fix(cookies): use base64url for session_data and Set-Cookie parsing

Kinfe123:
This fixes an issue on Base64URL/Base64 mismatch in session_data decoding and harden Set-Cookie parsing to handle Expires commas, preventing malformed/duplicate cookies from Expo clients. Add tests in core utils and Expo client for the new splitter. Resolves “Invalid Base64 character” errors.
    

---

## Summary by cubic
Fixed session_data decoding to use base64url and improved Set-Cookie parsing to handle Expires commas, preventing cookie errors from Expo clients.

- **Bug Fixes**
  - Switched session_data decoding to base64url to resolve "Invalid Base64 character" errors.
  - Updated Set-Cookie parsing logic to correctly split cookies with Expires attributes containing commas.
  - Added tests for the new Set-Cookie splitter in both core utils and Expo client.


Bekacru:
is this solving an opened issue?

Kinfe123:
There are inconsistencies on how we handles encoding and decoding. as i mentioned on pr this would make things consistent and at the same time make up on the standard on how should we handles padding leaks and headers. it also potentially solves this problem #2970 . but clearly there are inconsistencies out there thats needs to be addressed.

PR #3911, opened 2025-08-10
fix(username): don't normalize displayusernames

ping-maxwell:


Bekacru:
closed in favor of #3912

PR #3931, opened 2025-08-11
chore: fix typecheck check

Kinfe123:
## Summary by cubic
Added @types/jest to the stripe package to fix type checking errors during development.

- **Dependencies**
  - Updated pnpm-lock.yaml to include @types/jest and related type dependencies.


Bekacru:
we're not using jest

Kinfe123:
The ci was failing because of missing types of jest package that's why I added that. Ig this is because of the peerDep since I saw that fixed on your ci fix.

PR #3934, opened 2025-08-11
fix(org): prevent cross-org role updates in updateMemberRole

Kinfe123:


Bekacru:
closed in favor of https://github.com/better-auth/better-auth/pull/3932

PR #3937, opened 2025-08-11
feat(api): support glob patterns in disabledPaths matching

### LINKED ISSUE #3767: disabledPaths accepting a glob pattern

I am frustrated with the way better-auth handles enabling/disabling specific api routes.

Currently, Better Auth requires you to mount its global router/handler in order to access its full functionality. If you don’t mount the router, you lose to features, as some things like callbacks are only available when you mount an api route. This makes it difficult to implement custom authentication flows or integrate authentication logic directly into your own API endpoints.

Additionally, while there is a disabledPaths configuration option, it is annoying to use because you have to explicitly specify every route you want to disable. This makes it hard to selectively override or customize specific auth routes, and increases the risk of missing a route or introducing inconsistencies in your authentication logic.

My proposed solution is to allow the disabledPaths configuration option to accept not just an array of strings, but also glob patterns. Currently, the code simply checks on each request if the path is included in disabledPaths. It would not be a significant change to support glob patterns—the internal logic can remain mostly the same—but this would allow developers to be much more precise about which routes they want to enable or disable. This added flexibility would make it easier to customize authentication flows and override specific routes as needed.

### Additional context

A issue was already report about the same problem, but with a different solution: https://github.com/better-auth/better-auth/issues/2622

Kinfe123:
closes #3767
    

---

## Summary by cubic
Added support for glob patterns in the disabledPaths option, allowing more flexible path matching for disabling auth routes.

- **New Features**
 - Disabled paths now accept wildcards like * and ** for single and multi-level matching.
 - Updated documentation and tests to show usage and verify behavior.


packages/better-auth/src/api/to-auth-endpoints.test.ts
```diff
@@ -433,3 +433,17 @@ describe("disabled paths", async () => {
 		expect(error?.status).toBe(404);
 	});
 });
+
+describe("disabled paths with glob patterns", async () => {
+	const { client } = await getTestInstance({
+		disabledPaths: ["/sign-in/*", "/callback/**"],
+	});
+
+	it("should return 404 for single-level wildcard", async () => {
```
Bekacru, line 448: add more endpoints to test properly

peterjarian:
When will this be merged?

PR #3941, opened 2025-08-12
fix(org): make plugin schema compatible with exactOptionalPropertyTypes

Kinfe123:
With ts's exactOptionalPropertyTypes enabled, the org plugin was returning a schema object whose keys (e.g., team, teamMember) were sometimes present with undefined values due to conditional construction. This violates the AuthPluginSchema index signature because undefined is not assignable to the table shape, causing the “Property 'team' is incompatible with index signature” error. 
    

## Summary by cubic
Updated the organization plugin schema to avoid undefined optional keys and ensure compatibility with TypeScript's exactOptionalPropertyTypes setting.

- **Refactors**
 - Built schema objects without spreading undefined values.
 - Merged team tables and session fields only when team support is enabled.


packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -559,7 +557,8 @@ export const organization = <O extends OrganizationOptions>(
 		...options?.roles,
 	};
 
-	const teamSchema = teamSupport
+	// Build team schema in a way that never introduces undefined values when spreading
+	const teamSchema: AuthPluginSchema = teamSupport
```
Bekacru: `AuthPluginSchema` - would break the type since it wouldn't be able to infer the structure

try customizing the schema through the plugin options

packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -620,7 +619,139 @@ export const organization = <O extends OrganizationOptions>(
 					},
 				},
 			} satisfies AuthPluginSchema)
-		: undefined;
+		: ({} as AuthPluginSchema);
```
Bekacru: you don't want to do as

packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -620,7 +619,139 @@ export const organization = <O extends OrganizationOptions>(
...
+						model: "user",
+						field: "id",
+					},
+					fieldName: options?.schema?.invitation?.fields?.inviterId,
+					required: true,
+				},
+				...(options?.schema?.invitation?.additionalFields || {}),
+			},
+		},
+	};
```
Bekacru, line 755: this probably need satisfies

packages/better-auth/src/plugins/organization/organization.ts
Kinfe123: was trying to stub on it down the line but ig make sense to isolate it.

PR #3944, opened 2025-08-12
fix(telemetry): ensure telemetry consent notice displays correctly

frectonz:
This PR fixes two bugs:

- Telemetry debug mode was being enabled by default because `getBooleanEnvVar` has a `true` set as its default fallback, so this means telemetry debug was being enabled by default and telemetry notice was being disabled by default.
- `hasShownNoticeBefore` was returning true if it failed to read the `telemetry.json` file, which means for the first install it will always return true, and the notice will never be printed.
    

---

## Summary by cubic
Fixed telemetry consent notice so it displays correctly on first install and respects debug and disable settings.

- **Bug Fixes**
 - Telemetry debug mode and notice are now disabled by default unless explicitly enabled.
 - Consent notice now shows on first install as intended.

### DIFF
--- a/packages/better-auth/src/telemetry/index.ts
+++ b/packages/better-auth/src/telemetry/index.ts
@@ -23,24 +23,42 @@ You can also debug what would be sent by setting:
 
 Learn more in the docs: https://www.better-auth.com/docs/reference/telemetry\n\n`;
 
+async function configFilePath() {
+	const path = await import("path");
+	const os = await import("os");
+	const baseDir =
+		process.platform === "win32"
+			? process.env.APPDATA || path.join(os.homedir(), "AppData", "Roaming")
+			: path.join(os.homedir(), ".config");
+	const dir = path.join(baseDir, "better-auth");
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/telemetry/index.ts
```diff
@@ -23,46 +23,60 @@ You can also debug what would be sent by setting:
...
-	} catch {
-		//if we can't read the file, we don't want to show the notice
+	} catch (err: unknown) {
+		if (
+			err &&
+			typeof err === "object" &&
+			"code" in err &&
+			(err as NodeJS.ErrnoException).code === "ENOENT"
+		) {
+			// if the file doesn't exist we know that the notice hasn't been shown before
```
Bekacru, line 57: This means it'll always not exist so we keep showing the notice on every init, no?

frectonz: no it won't the `telemetry.json` file will be updated to include the current session id at the end of the `maybeShowTelemetryNotice` function

PR #3948, opened 2025-08-12
feat: async baseURL support

Kinfe123:
## Summary by cubic
Added support for async and per-request baseURL resolution in Better Auth, allowing dynamic baseURL values using a function.

- **New Features**
  - baseURL can now be a function that receives the request and context, and returns a string or Promise.
  - All internal usage and plugins updated to handle dynamic baseURL values.
  - New tests added for async baseURL scenarios.

### DIFF
--- a/packages/better-auth/src/plugins/jwt/sign.ts
+++ b/packages/better-auth/src/plugins/jwt/sign.ts
@@ -33,11 +33,14 @@ export async function signJWT(
 
 	// Iss
 	const iss = payload.iss;
-	const defaultIss = options?.jwt?.issuer ?? ctx.context.options.baseURL!;
-
+	const baseURL =
+		typeof ctx.context.options.baseURL === "function"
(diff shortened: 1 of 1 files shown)


himself65 (commented):
what's the benefit behind this? Do we really need async function to get `baseURL`?

Why not just wrap the whole betterAuth into an async function, because baseURL is too fundamental

Kinfe123:
Yeah there are the use case when the base urn could be resolved from external async function and that would be the case where this could be important.

himself65:
If you check this, we used a factory function to create a better auth instance.

 https://github.com/better-auth/better-auth/blob/466a23683d972aed74c36a6f050dbf99ef4bf286/integration-tests/cloudflare/src/auth.ts#L5

So, I think why not use a similar way to support async base URL?

himself65:
I think this PR is good, maybe I will cherry pick in another PR

PR #3955, opened 2025-08-12
fix(organization): was possible to remove sole org owner

### LINKED ISSUE #3651: Organization can have no owner or members

### To Reproduce

1. Install sveltekit
2. Install better-auth
3. Install organization from better-auth/plugins
4. implement adding/removing organization members and changing their roles

1. After creating an organization, if I try to remove myself from that org (as the owner), I get the error.message: "You cannot leave the organization as the only owner" That's great and expected behavior and a reasonable default for 99.9% of use cases.

2. Ok, so then I change my own role, as the only member of the org to 'admin' and that works. This is the issue. I don't believe you should be able to have an organization with no owner. Now the org has no owner. I cannot change myself back to an owner. I get error.message "You are not allowed to update this member" which makes sense

3. If I then try to remove myself from the org as an admin, I get the error.message "You are not a member of this organization" this is a strange error because really it was a success, I was removed from the organization, but I probably should not have been able to do that because there is now an org in the database with no members.

So I think the fix here, that would be the expected behavior and suit almost every setup is: Do not allow owners to change their role unless there is another owner. This will also fix the problem of organizations with no members, because as the sole owner, I can't remove myself I have to delete the team.

1.3.4

```bash
svelte@5.28.2
@sveltejs/kit@2.26.1
```

Client

```typescript
	appName: PUBLIC_SITE_NAME,
		provider: 'pg' // or "mysql", "sqlite"
	}),
	emailAndPassword: {
		enabled: true,
		sendResetPassword: async ({ user, url, token }, request) => {
			await sendEmail({
				to: user.email,
				subject: 'Reset your password',
				html: `Click the link to reset your password: ${url}`
			});
		},
		minPasswordLength: Number(PUBLIC_MIN_PASSWORD_LENGTH),
		maxPasswordLength: Number(PUBLIC_MAX_PASSWORD_LENGTH)
	},
	socialProviders: {
		google: {
			prompt: 'select_account',
			clientId: GOOGLE_CLIENT_ID,
			clientSecret: GOOGLE_CLIENT_SECRET
		},
		facebook: {
			clientId: FACEBOOK_APP_ID,
			clientSecret: FACEBOOK_APP_SECRET as string
		}
	},
	emailVerification: {
			await sendEmail({
				to: user.email,
				subject: 'Verify Your Email',
				html: `Click the link to verify your email: ${url}`
			});
		}
	},
	plugins: [
		sveltekitCookies(getRequestEvent),
		twoFactor(),
		organization({
			async sendInvitationEmail(data) {
				const inviteLink = `${PUBLIC_SITE_URL}/accept-invite/${data.id}`;
				sendEmail({
					to: data.email,
					subject: `You've been invited to ${data.organization.name}`,
					html: `To accept this invitation, follow this link: ${inviteLink}`
				});
			}
		}),
		admin({
			adminUserIds: [SUPER_ADMIN_USER_ID]
		}),
		passkey({
			rpName: PUBLIC_SITE_NAME,
			origin: PUBLIC_SITE_URL
		}),
		emailOTP({
			disableSignUp: true,
			expiresIn,
				let subject = '';
				let html = '';
				switch (type) {
					case 'sign-in':
						subject = 'One-Time Login Code';
						html = `
							<p>Here is your one-time login code: ${otp}</p>
							<p>Or you can follow this link: <a href="${PUBLIC_SITE_URL}/one-time-password?email=${email}&otp=${otp}">${PUBLIC_SITE_URL}/one-time-password?email=${email}&otp=${otp}</a>
							<p>This code will expire in ${expiresInMinutes} minutes.</p>
						`;
						break;
				}
				await sendEmail({
					to: email,
					subject,
					html
				});
			}
		})
	]
});
```

### Additional context

_No response_

gwoodbridge:
Removed possibility of removing the sole org owner by checking if the user calling update-member-role is an owner, if there is only one owner, if they are updating their own roles, and if they are not including an owner/creator role and then throwing a new error message.

Fixes #3651 
    

---

## Summary by cubic
Prevented removal of the sole organization owner to ensure every organization always has at least one owner, addressing Linear issue #3651.

- **Bug Fixes**
 - Added a check to block owners from removing their own owner role if they are the only owner.
 - Updated error messages for clarity.
 - Added tests to cover this scenario.


packages/better-auth/src/plugins/organization/routes/crud-members.ts
```diff
@@ -471,6 +473,26 @@ export const updateMemberRole = <O extends OrganizationOptions>(option: O) =>
 				});
 			}
 
+			if (updaterIsCreator) {
```
Bekacru: we should probably check here if `memberIsUpdatingThemselves` before making listMembers call

packages/better-auth/src/plugins/organization/routes/crud-members.ts
```diff
@@ -471,6 +473,26 @@ export const updateMemberRole = <O extends OrganizationOptions>(option: O) =>
...
+				const { members } = await adapter.listMembers({
```
Bekacru: Listing members currently makes 2 calls. It’s better to have a custom adapter call here, since we don’t need the count.

packages/better-auth/src/plugins/organization/routes/crud-members.ts
gwoodbridge: Moved check for `memberIsUpdatingThemselves`

packages/better-auth/src/plugins/organization/routes/crud-members.ts
gwoodbridge: Removed listMembers call.

gwoodbridge:
When making changes, I noticed that the check for if a member is an owner/creator in the leaveOrganization function wouldn't handle if an owner/creator had any additional roles correctly. I updated the function and test case.

himself65 (approved):

PR #3957, opened 2025-08-12
chore: Standardize error code capitalization and improve USER_ALREADY_EXISTS error code

jonas-s-s-s:
# USER_ALREADY_EXISTS 

This changes the USER_ALREADY_EXISTS error code from `"User already exists"` to `"User already exists. Use another email."` this signifies that the error was caused by entering an email that's already in the database (duplicate emails aren't allowed). It also helps to differentiate it from the USERNAME_IS_ALREADY_TAKEN error when using the username plugin. 

# Username plugin error codes

All strings in the USERNAME_ERROR_CODES struct have been changed to start with an uppercase letter, this ensures that the capitalization is the same as in the BASE_ERROR_CODES struct. 

    

---

## Summary by cubic
Updated the USER_ALREADY_EXISTS error code to give clearer guidance and standardized error code capitalization in the username plugin.

- **Error Code Improvements**
 - Changed USER_ALREADY_EXISTS to "User already exists. Use another email."
 - Updated all username plugin error messages to start with an uppercase letter for consistency.

### DIFF
--- a/packages/better-auth/src/error/codes.ts
+++ b/packages/better-auth/src/error/codes.ts
@@ -16,7 +16,7 @@ export const BASE_ERROR_CODES = {
 	EMAIL_NOT_VERIFIED: "Email not verified",
 	PASSWORD_TOO_SHORT: "Password too short",
 	PASSWORD_TOO_LONG: "Password too long",
-	USER_ALREADY_EXISTS: "User already exists",
+	USER_ALREADY_EXISTS: "User already exists. Use another email.",
 	EMAIL_CAN_NOT_BE_UPDATED: "Email can not be updated",
 	CREDENTIAL_ACCOUNT_NOT_FOUND: "Credential account not found",
 	SESSION_EXPIRED: "Session expired. Re-authenticate to perform this action.",
--- a/packages/better-auth/src/plugins/admin/error-codes.ts
+++ b/packages/better-auth/src/plugins/admin/error-codes.ts
@@ -1,6 +1,6 @@
 export const ADMIN_ERROR_CODES = {
 	FAILED_TO_CREATE_USER: "Failed to create user",
-	USER_ALREADY_EXISTS: "User already exists",
+	USER_ALREADY_EXISTS: "User already exists. Use another email.",
 	YOU_CANNOT_BAN_YOURSELF: "You cannot ban yourself",
 	YOU_ARE_NOT_ALLOWED_TO_CHANGE_USERS_ROLE:
 		"You are not allowed to change users role",
--- a/packages/better-auth/src/plugins/username/error-codes.ts
+++ b/packages/better-auth/src/plugins/username/error-codes.ts
@@ -1,10 +1,10 @@
 export const USERNAME_ERROR_CODES = {
-	INVALID_USERNAME_OR_PASSWORD: "invalid username or password",
-	EMAIL_NOT_VERIFIED: "email not verified",
-	UNEXPECTED_ERROR: "unexpected error",
-	USERNAME_IS_ALREADY_TAKEN: "username is already taken. please try another.",
-	USERNAME_TOO_SHORT: "username is too short",
-	USERNAME_TOO_LONG: "username is too long",
-	INVALID_USERNAME: "username is invalid",
-	INVALID_DISPLAY_USERNAME: "display username is invalid",
+	INVALID_USERNAME_OR_PASSWORD: "Invalid username or password",
+	EMAIL_NOT_VERIFIED: "Email not verified",
+	UNEXPECTED_ERROR: "Unexpected error",
+	USERNAME_IS_ALREADY_TAKEN: "Username is already taken. Please try another.",
+	USERNAME_TOO_SHORT: "Username is too short",
+	USERNAME_TOO_LONG: "Username is too long",
+	INVALID_USERNAME: "Username is invalid",
+	INVALID_DISPLAY_USERNAME: "Display username is invalid",
 };


Kinfe123 (approved):

lukebarton:
@Bekacru @Kinfe123 @jonas-s-s-s 

## 1. Bug introduced?
Heads up that this change causes the `authClient` to return the `error.code` `USER_ALREADY_EXISTS_USE_ANOTHER_EMAIL` because `better-call` translates the error `message` into uppercase snake_case here: https://github.com/Bekacru/better-call/blob/1baf90934b215658f9cd761d4278e86f803337d9/src/error.ts#L211-L219

This breaks usage like:
```ts
if(res.error.code){
    message = authClient.$ERROR_CODES[res.error.code];
}
```
because the `code` (`USER_ALREADY_EXISTS_USE_ANOTHER_EMAIL`) no longer matches the object `key` (`USER_ALREADY_EXISTS`) after `better-call` has transformed it (for whatever reason?).

It feels like there's a design issue here that needs unpicking, because translating an `error message` into something that looks like an `error code` isn't a choice I've seen made very often.

## 2. Design improvements
On a related design note, as a user of TypeScript, I kinda expected to be able to match the `error.code`s, to the values of the `auth.$ERROR_CODES` like so:

```ts
if (error.code == authClient.$ERROR_CODES.USER_ALREADY_EXISTS) {
  // ...
}
```

But the value of `authClient.$ERROR_CODES.USER_ALREADY_EXISTS` is the _message_ (`"User already exists. Use another email."`), rather than a string matching (i.e. `"USER_ALREADY_EXISTS"`) as is typically done in the TypeScript community for these kinds of string enums, so it doesn't match even if it weren't broken as mentioned above.

I'd usually expect code like this to look more like this:

```ts
enum ERROR_CODE {
  USER_ALREADY_EXISTS = "USER_ALREADY_EXISTS",
}

const builtin_messages: Record<ERROR_CODE, string> = {
  [ERROR_CODE.USER_ALREADY_EXISTS]: "User already exists. Use another email."
}

const error: { code: ERROR_CODE } = { code: ERROR_CODE.USER_ALREADY_EXISTS }

// user land:

const my_custom_messages: Record<ERROR_CODE, string> = {
  [ERROR_CODE.USER_ALREADY_EXISTS]: "Try another email, foo!"
}

builtin_messages[error.code]; // User already exists. Use another email.
my_custom_messages[error.code]; // Try another email, foo!

// and, further still:
if (error.code == ERROR_CODE.USER_ALREADY_EXISTS) {
  // ...
}

// and additionally
switch (error.code) {
  case ERROR_CODE.USER_ALREADY_EXISTS:
      // ...
    break;
}
```

Above is a _very_ standard implementation for this kind of code.

jonas-s-s-s:
> @Bekacru @Kinfe123 @jonas-s-s-s
> ## 1. Bug introduced?
> 
> Heads up that this change causes the `authClient` to return the `error.code` `USER_ALREADY_EXISTS_USE_ANOTHER_EMAIL` because `better-call` translates the error `message` into uppercase snake_case here: https://github.com/Bekacru/better-call/blob/1baf90934b215658f9cd761d4278e86f803337d9/src/error.ts#L211-L219
> 
> This breaks usage like:
> 
> ```ts
> if(res.error.code){
>     message = authClient.$ERROR_CODES[res.error.code];
> }
> ```
> 
> because the `code` (`USER_ALREADY_EXISTS_USE_ANOTHER_EMAIL`) no longer matches the object `key` (`USER_ALREADY_EXISTS`) after `better-call` has transformed it (for whatever reason?).
> 
> It feels like there's a design issue here that needs unpicking, because translating an `error message` into something that looks like an `error code` isn't a choice I've seen made very often.
> ## 2. Design improvements
> 
> On a related design note, as a user of TypeScript, I kinda expected to be able to match the `error.code`s, to the values of the `auth.$ERROR_CODES` like so:
> 
> ```ts
> if (error.code == authClient.$ERROR_CODES.USER_ALREADY_EXISTS) {
>   // ...
> }
> ```
> 
> But the value of `authClient.$ERROR_CODES.USER_ALREADY_EXISTS` is the _message_ (`"User already exists. Use another email."`), rather than a string matching (i.e. `"USER_ALREADY_EXISTS"`) as is typically done in the TypeScript community for these kinds of string enums, so it doesn't match even if it weren't broken as mentioned above.
> 
> I'd usually expect code like this to look more like this:
> 
> ```ts
> enum ERROR_CODE {
>   USER_ALREADY_EXISTS = "USER_ALREADY_EXISTS",
> }
> 
> const builtin_messages: Record<ERROR_CODE, string> = {
>   [ERROR_CODE.USER_ALREADY_EXISTS]: "User already exists. Use another email."
> }
> 
> const error: { code: ERROR_CODE } = { code: ERROR_CODE.USER_ALREADY_EXISTS }
> 
> // user land:
> 
> const my_custom_messages: Record<ERROR_CODE, string> = {
>   [ERROR_CODE.USER_ALREADY_EXISTS]: "Try another email, foo!"
> }
> 
> builtin_messages[error.code]; // User already exists. Use another email.
> my_custom_messages[error.code]; // Try another email, foo!
> 
> // and, further still:
> if (error.code == ERROR_CODE.USER_ALREADY_EXISTS) {
>   // ...
> }
> 
> // and additionally
> switch (error.code) {
>   case ERROR_CODE.USER_ALREADY_EXISTS:
>       // ...
>     break;
> }
> ```
> 
> Above is a _very_ standard implementation for this kind of code.

Then it must have already been an issue before this change? You have several error messages that don't match the key, like:

SESSION_EXPIRED: "Session expired. Re-authenticate to perform this action.",

or

USERNAME_IS_ALREADY_TAKEN: "Username is already taken. Please try another."

Therefore I didn't think it would be a problem to have a message that differs from the key for USER_ALREADY_EXISTS as well.

PR #3966, opened 2025-08-13
feat(passkey): error codes in passkey client

### LINKED ISSUE #3879: Include error code in error types for passkey result in client auth

### To Reproduce

1. Create a backend
2. Use Passkey plugin
```ts
 ...,
  plugins: [
    passkey({}),
  ],
```
3. Add Passkey plugin to client auth
```ts
 ...,
  plugins: [
    passkeyClient(),
  ]
})
```
4. Call the addPasskey function on the auth client
```ts
const result = await authClient.passkey.addPasskey({
      name: values.passkey_name,
      authenticatorAttachment: "cross-platform",
    });
```
5. You will need to trigger some kind of failure, in my case, the session was not fresh. Log the `result.error`
```ts
// the type of result.error
error: {
    message?: string | undefined;
    status: number;
    statusText: string;
} | undefined
```

CURRENT BEHAVIOR
When I log the error from the results of adding a passkey, it has a code property.

EXPECTED BEHAVIOR
The type of the error property in the results of adding a passkey should have a code property.

1.3.4

```bash
- OS: Windows 11
- Browser: Google Chrome
```

Types

```typescript
import { authAdapter } from "./firebase";
import { oneTap } from "better-auth/plugins";
import { headers } from "next/headers";
import { redirect } from "next/navigation";
import { cache } from "react";
import { siteInfo, siteLinks } from "@/config/site";
import { passkey } from "better-auth/plugins/passkey";

  database: authAdapter,
  plugins: [
    passkey({
      rpName: siteInfo.title,
      rpID: "better-auth-with-firebase",
      origin: process.env.BETTER_AUTH_URL,
      authenticatorSelection: {
        userVerification: "required",
        residentKey: "preferred",
      },
    }),
    oneTap({
      clientId: process.env.NEXT_PUBLIC_GOOGLE_CLIENT_ID as string,
      disableSignup: false,
    }),
    nextCookies(),
  ],
  emailAndPassword: {
    enabled: true,
    minPasswordLength: 12,
    autoSignIn: true,
  },
  account: {
    accountLinking: {
      enabled: true,
      allowDifferentEmails: false,
    },
  },
  socialProviders: {
    google: {
      clientId: process.env.NEXT_PUBLIC_GOOGLE_CLIENT_ID as string,
    },
    github: {
      clientId: process.env.GITHUB_CLIENT_ID as string,
      clientSecret: process.env.GITHUB_CLIENT_SECRET as string,
    },
  },
  rateLimit: {
    storage: "database",
    modelName: "rateLimit",
    enabled: true,
  },
  session: {
    expiresIn: 60 * 60 * 24 * 7, // 7 days
    updateAge: 60 * 60 * 24, // 1 day (every 1 day the session expiration is updated)
    freshAge: 60 * 30, // 30 minutes (the session is fresh if created within the last 30 minutes)
    cookieCache: {
      enabled: true,
      maxAge: 15 * 60, // Cache duration in seconds
    },
  },
});
```

### Additional context

I tried to review the code and it looks like some of the errors are handled manually and then an error is returned without the code. Will it be possible for us to add the code or will it break anything? I can create a pull request but I am not sure what impact it will ahve.

In my use case for example, the error code I wanted to handle was `SESSION_NOT_FRESH` and the message in the `result.error` was something like `Session is not fresh`. This is good but I think that as the library evolves, messages from errors might change so checking for specific messages or text strings like that is not the best but the code is more dependable since it will hardly change.

frectonz:
closes #3879
    

---

## Summary by cubic
Added error codes to passkey client error responses to make it easier to identify and handle specific authentication errors. This addresses the requirements in Linear issue #3879.


himself65 (commented):
should we put ERROR_CODE top-level?

PR #3969, opened 2025-08-13
docs: `authClient` should not be used from a content script

### LINKED ISSUE #3872: best practice to call useSession in content script of browser extension

### To Reproduce

The related issue is #1381 

I'm thinking what's the best practice to use better auth in content script which send request from the host origin.

Have several ideas to solve (haven't verified):
1. send message to background script to let background request api from better auth backend, and let the auth backend whitelist the background origin `extension://xxxxx`
2. change `fetch` in `createAuthClient`, and write the logic in this `fetch` to send message to the backgroud script and return the results from there.
```
  baseURL: WEBSITE_URL,
  fetch: customFetch
})
```

current behavior: 

Image

expected behavior:

no CORS problem

1.2.12

```bash
- OS: macOS
- Browser: Chrome
```

Client

```typescript
import type { BetterAuthOptions } from 'better-auth'
import { APP_NAME } from '@repo/constants'
import { env } from './env'

/**
 * Custom options for Better Auth
 *
 * Docs: https://www.better-auth.com/docs/reference/options
 */
export const betterAuthOptions: BetterAuthOptions = {
  /**
   * The name of the application.
   */
  appName: APP_NAME,
  /**
   * Base path for Better Auth.
   * @default "/api/auth"
   */
  // basePath: "/api",

  /**
   * Enable email and password authentication
   */
  emailAndPassword: {
    enabled: true,
    requireEmailVerification: false, // TODO: set to be true later
    minPasswordLength: 8,
    maxPasswordLength: 128,
  },
  socialProviders: {
    google: {
      prompt: 'select_account',
      clientId: env.GOOGLE_CLIENT_ID,
      clientSecret: env.GOOGLE_CLIENT_SECRET,
    },
  },
  // .... More options
}
```

### Additional context

_No response_

frectonz:
closes #3872
    

---

## Summary by cubic
Added a warning to the browser extension guide explaining that `authClient` should not be used directly in content scripts, with recommendations for safer usage through background scripts.

### DIFF
--- a/docs/content/docs/guides/browser-extension-guide.mdx
+++ b/docs/content/docs/guides/browser-extension-guide.mdx
@@ -213,8 +213,8 @@ If you would like to view a completed example, you can check out the <Link href=
 
 ## Usage Notes
 
-<Accordions>
-<Accordion title="Avoid Using authClient Directly in Content Scripts">
+### Avoid Using `authClient` Directly in Content Scripts
+
 Content scripts run in an isolated world and cannot directly communicate with your authentication server in the same way as background scripts or UI scripts. For security and stability, do not call authClient directly from a content script.
 
 For security and stability, **do not** call authClient directly from a content script.
@@ -239,9 +239,6 @@ chrome.runtime.sendMessage({ type: "getSession" }, (session) => {
   console.log("Session from background:", session) // [!code highlight]
 })
 ```
-</Accordion>
-</Accordions>
-
 
 ## Wrapping Up
 


docs/content/docs/guides/browser-extension-guide.mdx
```diff
@@ -83,6 +83,17 @@ If you would like to view a completed example, you can check out the <Link href=
             plugins: [],
         });
         ```
+
+        <Callout type="warn">
```
Bekacru: this seems a bit out of place to be included here. Maybe let's move it in the bottom or have a section dedicated on how to use it with customFetchImpl...

frectonz: ok will do that

frectonz: I have moved the callout to a usage notes sections that looks like this.

image

docs/content/docs/guides/browser-extension-guide.mdx
```diff
@@ -211,6 +211,37 @@ If you would like to view a completed example, you can check out the <Link href=
 
 </Steps>
 
+## Usage Notes
+
+<Accordions>
```
Bekacru: we dont need to do an accordion for one section

frectonz: I have removed the accordion.

PR #3974, opened 2025-08-13
  fix(cli): clean up unused `pg` and `mysql` import in drizzle schema generator

### LINKED ISSUE #3971: Remove unused 'integer' import from drizzle schema (pg-core)

When generating the Drizzle schema for PostgreSQL (pgTable), the code currently imports 'integer' from "drizzle-orm/pg-core" even though it is not used in the resulting schema output. This creates unnecessary imports in generated files and may cause confusion for new users.

Update the code generation logic so that 'integer' is only imported if it is actually needed for the table definition. For standard auth tables (user, session, account, verification), only 'text', 'timestamp', and 'boolean' should be imported unless integer columns exist in the schema.

Alternatively, always check the schema for any integer fields before determining the imports. Or allow customization to remove unused imports.

### Additional context

This cleanup would improve clarity for code generated via 'npx @better-auth/cli generate' and prevent unnecessary imports for users who only use text/timestamp/boolean fields.

Kinfe123:
closes #3971    

## Summary by cubic
Removed unused `integer` import from generated Drizzle schemas for PostgreSQL when not needed, keeping imports clean and accurate.

### DIFF
--- a/packages/cli/src/generators/drizzle.ts
+++ b/packages/cli/src/generators/drizzle.ts
@@ -212,7 +212,18 @@ function generateImport({
 	imports.push(hasBigint ? (databaseType !== "sqlite" ? "bigint" : "") : "");
 	imports.push(databaseType !== "sqlite" ? "timestamp, boolean" : "");
 	if (databaseType === "mysql") {
-		imports.push("int");
+		// Only include int for MySQL if actually needed
+		const hasNonBigintNumber = Object.values(tables).some((table) =>
+			Object.values(table.fields).some(
+				(field) =>
+					(field.type === "number" || field.type === "number[]") &&
+					!field.bigint,
+			),
+		);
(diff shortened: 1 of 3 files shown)


packages/cli/src/generators/drizzle.ts
```diff
@@ -211,7 +211,32 @@ function generateImport({
 	);
 	imports.push(hasBigint ? (databaseType !== "sqlite" ? "bigint" : "") : "");
 	imports.push(databaseType !== "sqlite" ? "timestamp, boolean" : "");
-	imports.push(databaseType === "mysql" ? "int" : "integer");
+	if (databaseType === "mysql") {
+		imports.push("int");
```
Bekacru: I think we should also check for mysql here

Kinfe123:
for sqlite they can reference the `integer` for creating a timestamp values.

PR #3975, opened 2025-08-13
refactor(cli): remove unused integer import from drizzle schema gener…

### LINKED ISSUE #3971: Remove unused 'integer' import from drizzle schema (pg-core)

Boby900:
…ation

- Add check for integer fields before including import
- Only import integer when useNumberId is true or number fields exist
- Add tests to verify import behavior

Fixes #3971

### DIFF
--- a/packages/cli/src/generators/drizzle.ts
+++ b/packages/cli/src/generators/drizzle.ts
@@ -145,32 +145,29 @@ export const generateDrizzleSchema: SchemaGenerator = async ({
 		)}", {
 					id: ${id},
 					${Object.keys(fields)
-						.map((field) => {
-							const attr = fields[field]!;
-							let type = getType(field, attr);
-							if (attr.defaultValue) {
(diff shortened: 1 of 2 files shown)


Kinfe123:
closing in favor of #3974

PR #3977, opened 2025-08-13
fix: improve error handling for unsupported additionalFields on generate

Kinfe123:


packages/better-auth/src/db/get-migration.ts
```diff
@@ -369,6 +369,11 @@ export async function getMigrations(config: BetterAuthOptions) {
 		if (Array.isArray(type)) {
 			return "text";
 		}
+		if (!(type in typeMap)) {
```
himself65, line 372: Do we already support JSON now?

PR #3983, opened 2025-08-13
feat: add `returnHeaders` to `getSession`

### LINKED ISSUE #3780: Type issue when using "auth.api.getSession"
### To Reproduce

1. Create a backend
2. Use `auth.api.getSession`
3. The endpoint signature does not allow to use `returnHeaders` option

**Current** : 

- No `returnHeaders` option for `getSession` endpoint
 
Image

*__Note__: Still, we can force the `returnHeaders` option (even if not typed) and it will work as expected but with many typscript errors !* 👇

Image

**Expected** : 

- Having `returnHeaders` option + correct typings when using the option, like other endpoints, here is an example with `signUpEmail` endpoint :

Image

We should have no problem using the following code :

```ts
const { headers, response } = await auth.api.getSession({
  headers,
  returnHeaders: true
}

console.log(headers) // Should be of type Headers
console.log(response) // Should be of type { session: ..., user: ... }
```

1.3.4

```bash
OS: Ubuntu
```

Backend

```typescript

  emailAndPassword: {
    enabled: true,
  },
})
```

### Additional context

https://github.com/better-auth/better-auth/issues/2115
https://github.com/better-auth/better-auth/issues/2115#issuecomment-3069526380

It looks like there a big type issue, as using "Go to Type Definition" on `getSession` leads to this :

Image

Meanwhile other endpoints lead to a definition where there is all other endpoints :

Image

And the big surprise, there is another `getSession` definition there that looks correctly typed :

Image

So overall I think there's something in the types that override this correctly typed `getSession` endpoint by a wrong typed  `getSession` endpoint (the one with `InferAPI` thing)

frectonz:
closes #3780 
    

---

## Summary by cubic
Added the `returnHeaders` option to the `getSession` endpoint, allowing callers to include response headers in the returned session object when requested.


Bekacru (approved):

Bekacru:
Lint and also can you add a type test for all `auth.api.getSession` related stuff here

frectonz:
I have fixed the lint issues and added some tests, I have also added type tests for the `auth.api.getSession` function.

kevinmarrec:
Would be grateful to have some update on this one and be able to use this soon :)

himself65 (approved):

himself65:
> Would be grateful to have some update on this one and be able to use this soon :)

Sorry I forget this PR, I will merge it soon

kevinmarrec:
@himself65 We may need to double check, dunno, I tried to override my local betterAuth types with the ones of this PR :

image

I was expecting the usage to be the same as other endpoints, and have a `headers` aside a `response` like this :

image

frectonz:
@kevinmarrec you can try out the changes in this PR by installing it like this

```
npm i https://pkg.pr.new/better-auth/better-auth@3983
```

check this for reference https://github.com/better-auth/better-auth/pull/3983#issuecomment-3185673861

frectonz:
You may have copied the types incorrectly. 

`signUpEmail` as shown in the screenshot, infers the types correctly when `returnHeaders` is passed in. It returns the `headers`.

#### `signUpEmail `

image


#### `getSession`

image

kevinmarrec:
@frectonz Yes, the tests are also correct, but headers is added to the object already containing session and user, AFAIK.

`getSession` is special and types are defined separately from other endpoints

kevinmarrec:
image

I think I'm confused with the difference with other endpoints implementations :

image

UPDATE:

I've tried to output the result of `getSession` call with `returnHaders: true`, and here you go :

image

kevinmarrec:
One of the main issue of the types is that `getSession` is automatically correctly typed by `createAuthEndpoint` here :

https://github.com/better-auth/better-auth/blob/cf2cac4dd9eaefb01b8f08cbb096585b31962aad/packages/better-auth/src/api/routes/session.ts#L47-L80

But types have been explicitly overridden in another place :

https://github.com/better-auth/better-auth/blob/cf2cac4dd9eaefb01b8f08cbb096585b31962aad/packages/better-auth/src/types/api.ts#L4-L15

EDIT : AFAIK it has been done this way to be able to infer custom session response type from custom session plugin :
https://github.com/better-auth/better-auth/pull/579

frectonz:
Yeah you are right calling `getSession` even `asResponse` doesn't return the headers, this might be a specific design choice and not a bug will investigate this further.

image

image

frectonz:
The type modification are not needed since the implementation itself doesn't return the headers.

kevinmarrec:
@frectonz AFAIK it return the headers, you need to have a low `session.updateAge`, so that when you're logged in and call `getSession` after this age, the API will refresh cookie thanks to this header

kevinmarrec:
image

Btw no idea why this landed, the endpoint never returns `options` or `path`

frectonz:
Yeah it does return the headers but only when the session needs to be refreshed. Your comment here made me think there was a bug.

https://github.com/better-auth/better-auth/pull/3983#issuecomment-3294035158

packages/better-auth/src/types/api.ts
```diff
@@ -32,18 +32,29 @@ export type InferSessionAPI<API> = API extends {
...
+												path: E["path"];
+												headers: Headers;
+											}
+										>
+									: Promise<
+											PrettifyDeep<Awaited<ReturnType<E>>> & {
+												options: E["options"];
+												path: E["path"];
+											}
+										>
```
kevinmarrec: I think correct types could be
```suggestion
									? Promise<{
									        headers: Headers;
											response: PrettifyDeep<Awaited<ReturnType<E>>>;
										}>
									: Promise<PrettifyDeep<Awaited<ReturnType<E>>>>
```

My guess is that `options` / `path` stuff is non sense to me, it seems to be either a mistake OR some hack to infer them from another place (not in user land)

kevinmarrec:
> Yeah it does return the headers but only when the session needs to be refreshed. Your comment here made me think there was a bug.
> 
> [#3983 (comment)](https://github.com/better-auth/better-auth/pull/3983#issuecomment-3294035158)

@frectonz It's indeed not bugged, to be clear what I was pointing is only the structure of the response (`{ headers, response }`) and not its content (like empty headers), sorry If you got confused

packages/better-auth/src/types/api.ts
frectonz: Yeah I think you are right this the correct way to do it but with two modifications.

- the `returnHeaders` generic should default to `false`
- this PR shouldn't remove the `options` and `path` option, I am not clear on what their use is here but I think handling them in a separate PR would be better.

kevinmarrec: I agree with both, totally makes sense

packages/better-auth/src/types/api.ts
```diff
@@ -32,18 +32,29 @@ export type InferSessionAPI<API> = API extends {
...
-										options: E["options"];
-										path: E["path"];
-									}
+								? H extends true
+									? Promise<{
+											headers: Headers;
+											response: PrettifyDeep<Awaited<ReturnType<E>>>;
+										}>
+									: Promise<
+											PrettifyDeep<Awaited<ReturnType<E>>> & {
```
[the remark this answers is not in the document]
frectonz: `options` and `path` are not actually returned

frectonz:
I think the PR is ready now.

kevinmarrec:
@frectonz Just tried the last version, works perfectly.

image

kevinmarrec:
@himself65 I let you check again, LGTM

PR #3985, opened 2025-08-13
fix(cli): create telemetry object before doing CLI tasks

frectonz:


Kinfe123 (approved):

packages/cli/src/commands/generate.ts
```diff
@@ -43,6 +47,10 @@ export async function generateAction(opts: any) {
 		process.exit(1);
 	});
 
+	// Unmark CLI execution before creating telemetry.
+	process.env.BETTER_AUTH_CALLED_FROM_CLI = "false";
```
himself65:
```suggestion
	delete process.env.BETTER_AUTH_CALLED_FROM_CLI;
```

packages/cli/src/commands/migrate.ts
```diff
@@ -45,13 +48,16 @@ export async function migrateAction(opts: any) {
 		process.exit(1);
 	}
 
+	// Unmark CLI execution before creating telemetry.
+	process.env.BETTER_AUTH_CALLED_FROM_CLI = "false";
```
himself65: same

Kinfe123: yeah that make sense

packages/cli/src/commands/generate.ts
frectonz: image

i don't think biome likes delete.

PR #3990, opened 2025-08-13
feat: waku integration guide

### LINKED ISSUE #3976: Docs: integration guide for Waku

Waku is a minimal React framework. Developers building projects on Waku are looking for guides to integrate authentication. Better Auth seems to be very easy to integrate with Waku.

I'd like to write an integration guide - like the [remix guide](https://www.better-auth.com/docs/integrations/remix) - for Waku.

I have published an example. https://github.com/rmarscher/waku-better-auth-example

### Additional context

_No response_

rmarscher:
Fixes #3976. I made an example repo with most of the patterns mentioned in the guide. https://github.com/rmarscher/waku-better-auth-example


docs/content/docs/integrations/waku.mdx
```diff
@@ -0,0 +1,470 @@
...
+```
+
+<Callout type="warn">RSCs that run after the response has started streaming cannot set cookies. The [cookie cache](/docs/concepts/session-management#cookie-cache) will not be refreshed until the server is interacted with from the client via Server Actions or Route Handlers.</Callout>
+
+### Server Action Cookies
+
+When you call a function that needs to set cookies, like `signInEmail` or `signUpEmail` in a server action, cookies won’t be set.
+
+We can create a plugin that works together with our middleware to set cookies.
+
```
himself65, line 115: Do we already have the helper for that? no?

docs/content/docs/integrations/waku.mdx
```diff
@@ -0,0 +1,470 @@
...
+  } satisfies BetterAuthPlugin;
+}
+```
+
+See below for the middleware to create to add the `contextData.betterAuthSetCookie` cookies to the response.
+Now, when you call functions that set cookies, they will be automatically set.
+
+```ts
+"use server";
+import { auth } from "@/lib/auth"
```
himself65: use relative path, since waku doesn't supports tsconfig path by default

docs/content/docs/integrations/waku.mdx
rmarscher: @himself65 I don't think so. It would be a helpful addition after the default middleware refactor is complete.

docs/content/docs/integrations/waku.mdx
rmarscher: Thanks. I missed that one.

docs/content/docs/integrations/waku.mdx
himself65: I think any is okay since it just the document. Idk the progress on waku now, but I hope it supports alias built-in afterall

rmarscher: Yeah, I think it should be in the default template. There is just a canonical example in the waku repo:
- https://github.com/wakujs/waku/blob/v0.26.1/examples/41_path-alias/waku.config.ts
- https://github.com/wakujs/waku/blob/v0.26.1/examples/41_path-alias/tsconfig.json#L17

Issue #2772, opened 2025-05-24
Organization plugin lacking metadata type-safety and JSON parsing


ohkimur:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Create a new organization.
2. Add metadata fields.
3. Try to access this metadata from `useActiveOrganization` and/or `useListOrganizations` -> No type-safety and it's raw `string` instead of parsed JSON.

### Current vs. Expected behavior

The current behavior is almost there, but we should be able to define a type schema for the metadata in an intuitive way. Additionally the metadata object although stored as a string in the db should be parsed by default with the right type.

Additionally we should have the ability to customize the organization schema through `additionalFields` such that we don't just have metadata, but we have the power of actually extending the organization schema properly. There are plenty of usecases where this would be very valuable.

### What version of Better Auth are you using?

1.2.8

### Provide environment information

```bash
MacOS, Arc Browser
```

### Which area(s) are affected? (Select all that apply)

Types

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
import { organization } from 'better-auth/plugins'

export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
  plugins: [organization()]
});
```

Kinfe123:
Metadata is designed to store a wide range of complex and dynamic data based on various use cases. This gives you greater flexibility when storing information in the metadata field, and you can handle typing in your application code, which feels quite intuitive.

PR #4000, opened 2025-08-14
feat(organization): make stringifying metadata optional

Drewsapple:
I was confused by the documentation's use of the organization metadata field as an object, and set it up in my database as a jsonb column. While I understand @Kinfe123 reason of "flexibility" in #2772, allowing an object to be saved to the db enables flexibility to do much more with raw sql, including adding indices on metadata fields.

Is there a strong reason why stringifying before saving to the DB should be required? This PR provides the option to directly write the incoming Record<string, any> to the db, so that the flexibility extends to direct db queries as well, instead of forcing the flexibility to exist only within application code.

I'd be happy to update the docs as well to describe the behavior here, but I'm unsure if that should be a commit in this PR or a separate one with a docs header. LMK. 
    

---

## Summary by cubic
Added an option to control whether organization metadata is stringified before saving to the database, allowing direct storage of objects for more flexible querying.

- **New Features**
 - Added the stringifyOrganizationMetadata option to OrganizationOptions.
 - Metadata can now be stored as either JSON strings or raw objects based on this setting.

### DIFF
--- a/packages/better-auth/src/plugins/organization/adapter.ts
+++ b/packages/better-auth/src/plugins/organization/adapter.ts
@@ -50,17 +50,22 @@ export const getOrgAdapter = <O extends OrganizationOptions>(
 				data: {
 					...data.organization,
 					metadata: data.organization.metadata
-						? JSON.stringify(data.organization.metadata)
+						? (options?.stringifyOrganizationMetadata ?? true)
+							? JSON.stringify(data.organization.metadata)
+							: data.organization.metadata
 						: undefined,
 				},
 			});
 
 			return {
 				...organization,
-				metadata:
-					organization.metadata && typeof organization.metadata === "string"
+				metadata: organization.metadata
+					? typeof organization.metadata === "string"
 						? JSON.parse(organization.metadata)
-						: undefined,
+						: typeof organization.metadata === "object"
+							? organization.metadata
+							: undefined
+					: undefined,
 			} as typeof organization;
 		},
 		findMemberByEmail: async (data: {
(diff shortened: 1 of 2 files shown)


Kinfe123:
Right now we don't have a standard way of handling JSON variant as field type on across the adapters. But at the same time flexibility on storing as string should be fine. Since either way it is reversible at the application level regardless of any client consuming the result. Once we have a stable way of handling JSON we can do that with the option for you to fallback even to serialized one. But it is in our roadmap.

Drewsapple:
> But at the same time flexibility on storing as string should be fine.

It isn't fine. I need to be able to be able to use json/jsonb in my db to access properties on the fields. This doesn't happen in my application code, after better-auth parses it.

What do you mean by:
> a stable way of handling json

With directly writing the object to the db (at least in drizzle), it just works.

Kinfe123:
Im not saying they don't support that. But We are working on making sure JSON fields behave consistently across adapters, mapping cleanly to the appropriate JSON primitives with a solid fallback to a serialized form. This is already on our roadmap. We have run into this issue before and we plan to make JSON handling a first-class feature, ensuring the underlying databases handle it properly as well.

Drewsapple:
Is there a tracking issue or an open PR for json support in adapters? If there's any way I can help get that across the line, I'd be happy to help. Storing structured data in my db is pretty important to me.

Kinfe123:
Yeah we welcome contributions. Happy to review your code.

treepo1:
Hi guys, is there any update on this? I'm trying to use the metadata field to create organization-scoped apiKeys and this would be important to safe query my db

PR #4002, opened 2025-08-14
fix: never type still requires setting clientId to never

szcharlesji:
Right now `clientId` is still required. We have to use `clientId: undefined as never` right now. The fix makes it optional.

- No Type Errors: Legacy users can keep clientId in their config without TypeScript complaining.
- No Requirement for New Users: New users can set up the provider without clientId.
- Runtime Safety: The logic ensures the obsolete clientId is never actually used, making the integration robust.
    

---

## Summary by cubic
Made the clientId field optional in the TikTok provider config so it's no longer required for TypeScript users or new integrations.

- **Migration**
 - Legacy configs with clientId will not cause type errors.
 - New setups do not need clientId at all.


himself65 (approved):

himself65 (commented):
just realized that in the current type system, clientId is has to be string, or the type checking will fail

szcharlesji:
Thanks, this should be the complete type fix

PR #4018, opened 2025-08-15
fix: hero section alignment with feature section on big screens

Kinfe123:
## Summary by cubic
Improved the alignment and width of the hero and feature sections for better layout consistency on large and ultrawide screens. 

- **UI Improvements**
 - Matched container widths and spacing between the hero and feature sections.
 - Adjusted padding and max-width values for both sections to align content precisely.


Bekacru:
now it's too centered for all screens. We want to fix the hero section from being centered on bigger screens

PR #4036, opened 2025-08-15
Add proofkit/better-auth DB adapter

eluce2:
FileMaker is an application platform that includes a built-in database. This adapter allows us to use that built-in database directly for authentication without having to spin up a new database just for auth. It communicates using the OData protocol
    

---

## Summary by cubic
Added @proofkit/better-auth to the Community Adapters docs. This enables using FileMaker’s built-in database for Better Auth via OData, so you don’t need a separate auth database.


himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #4037, opened 2025-08-16
fix(stripe): prevent undefined assignment to optional properties

Kinfe123:
## Summary by cubic
Prevents undefined being assigned to optional Stripe schema properties by stripping disabled subscription fields before merge and returning a correctly typed schema.

- **Bug Fixes**
  - Build a base schema from user (+subscriptions when enabled) and merge without mutating options.schema.
  - Remove subscription from overrides when subscriptions are disabled instead of setting it to undefined.
  - Add explicit AuthPluginSchema typing and satisfies assertions for safer returns.

### DIFF
--- a/packages/better-auth/package.json
+++ b/packages/better-auth/package.json
@@ -1,6 +1,6 @@
 {
   "name": "better-auth",
-  "version": "1.3.7-beta.1",
+  "version": "1.3.7-beta.3",
   "description": "The most comprehensive authentication library for TypeScript.",
   "type": "module",
   "license": "MIT",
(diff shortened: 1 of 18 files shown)


packages/stripe/src/schema.ts
```diff
@@ -58,18 +58,27 @@ export const user = {
...
+	}
+
 	if (
 		options.schema &&
 		!options.subscription?.enabled &&
 		"subscription" in options.schema
 	) {
-		options.schema.subscription = undefined;
+		const { subscription, ...restSchema } = options.schema;
+		return mergeSchema(baseSchema, restSchema) satisfies AuthPluginSchema;
```
Bekacru: do we need to do this `satisfies AuthPluginSchema;` here?

Kinfe123: yeah. i just did that for case of undefined. but as long as the merge is capable of handling optional field. i dont think need it but it is more safer.

Kinfe123:
since it already extending and asserting the `AuthPluginSchema` on the merge. i dont think we need satisfies here. it seems a redundant handles.

PR #4049, opened 2025-08-17
feat(organization): organization life cycle hooks

Bekacru:


dobrynin:
I think this PR broke the hook functionality. Unless there is some magic going on, I think there needs to be a replacement for the `...(hookResponse?.data || {}),` line that was removed. I had to revert to 1.3.7.

PR #4062, opened 2025-08-18
feat: database transaction support

jslno:
## Summary by cubic
Add first-class database transactions to the adapter layer using an async context, so all CRUD calls inside a transaction use the correct tx-bound client. Exposes adapter.transaction(cb) and getContext() to enable nested transactions and safer multi-step operations.

- New Features
  - Added Adapter.transaction(cb) to run operations atomically.
  - Introduced async context and getContext() to pass the tx-bound db through adapter methods.
  - Implemented transaction support for Drizzle, Prisma, Kysely, and MongoDB (via session).
  - Updated adapters to read db from context, and added tests for nested transactions and error handling.

## Todo
- [x] opt-out transactions (D1 Compatibility)


ping-maxwell:
Awesome PR, I started work on this months ago but it's incomplete.
I know you're still working on it, but if I understand the current progress, it might have a flaw.
Say you create a transaction, and within there are some adapter operations. But outside of the transaction there were more adapter operations running synchronously to the transaction, would the transaction catch the operations that were ran outside of the transaction itself incorrectly thus when rolled-back cause incorrect rollback operations?

jslno:
> Say you create a transaction, and within there are some adapter operations. But outside of the transaction there were more adapter operations running synchronously to the transaction, would the transaction catch the operations that were ran outside of the transaction itself incorrectly thus when rolled-back cause incorrect rollback operations?

You're right, it's not possible to properly isolate concurrent operations without relying on `node:async_hooks`. I need to refactor to explicitly pass the context.

mw10013:
Cloudflare D1 database seems to only support a transaction around multiple statements with `batch()` (https://developers.cloudflare.com/d1/build-with-d1/d1-client-api/#batch-statements) and does not have a construct like a transaction context. Would this PR break D1 compatibility?

Wondering if `batch` along with `upsert` would be able to cover all the transaction cases needed by better-auth. Drizzle supports batch: https://orm.drizzle.team/docs/batch-api

jslno:
> Cloudflare D1 database seems to only support a transaction around multiple statements with `batch()` (https://developers.cloudflare.com/d1/build-with-d1/d1-client-api/#batch-statements) and does not have a construct like a transaction context. Would this PR break D1 compatibility?
> 
> Wondering if `batch` along with `upsert` would be able to cover all the transaction cases needed by better-auth. Drizzle supports batch: https://orm.drizzle.team/docs/batch-api

You're correct. It'd break D1 compatibility. I need to add an option to opt-out of transactions and (eventually) support batching.

packages/better-auth/src/adapters/create-adapter/test/create-adapter.test.ts
```diff
@@ -66,57 +70,66 @@ async function createTestAdapter(
 			},
 			config,
 		),
+		context,
+		// @ts-ignore
```
himself65: use `ts-expect-error`

packages/better-auth/src/adapters/create-adapter/index.ts
```diff
@@ -75,6 +78,10 @@ export const createAdapter =
 		// End-user's Better-Auth instance's schema
 		const schema = getAuthTables(options);
 
+		const getContext = (ctx: AdapterContext | undefined) => {
+			return (ctx || baseContext) as C;
+		};
```
himself65, line 83: is this utility needed?

packages/better-auth/src/adapters/mongodb-adapter/mongodb-adapter.ts
```diff
@@ -228,20 +272,29 @@ export const mongodbAdapter = (db: Db, config?: MongoDBAdapterConfig) => {
...
+				async create({ model, data: values }, context) {
+					const ctx = serializeContext(context);
+					const res = await ctx.db.collection(model).insertOne(values, {
+						session: ctx.session,
+					});
 					const insertedData = { _id: res.insertedId.toString(), ...values };
 					return insertedData as any;
 				},
-				async findOne({ model, where, select }) {
+				async findOne({ model, where, select }, context) {
```
himself65, line 298: should we move context into db? What is the difference between db and ctx.db?

packages/better-auth/src/adapters/create-adapter/index.ts
jslno: Not necessarily, but its keeping the code clean, and helps with type safety. `|| baseContext` is needed to fallback to the default db instance, when transactions aren't supported.

packages/better-auth/src/adapters/mongodb-adapter/mongodb-adapter.ts
jslno: `ctx` is required because it holds the current active db instance. During a transaction, the context swaps in a different db instance (or in this case a session) so queries run inside the transaction.
So `ctx.db.`  runs with the transaction adapter inside transactions while `db.` would always use the default db instance. In this case you'd probably get away with using `db.` since only the session changes and not the db instance.

packages/better-auth/src/adapters/mongodb-adapter/mongodb-adapter.ts
```diff
@@ -15,9 +25,36 @@ export interface MongoDBAdapterConfig {
...
+				database?: string;
+				options?: DbOptions;
+		  };
+	client: MongoClient | undefined;
+	session: ClientSession | undefined;
+};
+
+export const mongodbAdapter = (
+	db:
+		| Db
```
jslno, line 51: Should i add a deprecated notice for `Db`?

himself65:
will check this today

himself65:
I implement in another way, see https://github.com/better-auth/better-auth/pull/4414.

Basically, I don't want to expose `ctx` to the application side. The abstraction layer should handle the trx correctly. Let me know if you'd like to continue to work based on that. Feel free to have other opinions. I think I won't continue on that PR for days

jslno:
> Basically, I don't want to expose ctx to the application side. The abstraction layer should handle the trx correctly. Let me know if you'd like to continue to work based on that. Feel free to have other opinions. I think I won't continue on that PR for days

Yeah I agree. I'm quite busy right now so I just took a quick look at your PR. If I understand you're approach correctly you'd only be required to pass the trx adapter to internalAdapter methods instead of any adapter method, which is way safer. I might update this PR or open a new one merging into yours when I have time.

PR #4073, opened 2025-08-18
feat: export jwt plugin with error

dvanmali:
Exports `getJwtPlugin` for other plugins. Errors early via `init` to prevent need of checks at runtime.

Removes unused JWT utils from [utils.ts](packages/better-auth/src/plugins/jwt/utils.ts). Non-breaking since it is never exported by [index.ts](packages/better-auth/src/plugins/jwt/index.ts).

Partial https://github.com/better-auth/better-auth/pull/3572

Type: **PATCH**
    

---

## Summary by cubic
Exports a safe getJwtPlugin helper that throws when the JWT plugin is missing, and updates OIDC to fail fast during init when useJWTPlugin is true. Also removes unused JWT crypto helpers.

- **New Features**
  - Exported getJwtPlugin from the JWT plugin; throws BetterAuthError if not found.
  - OIDC provider init checks for JWT plugin when useJWTPlugin is true to prevent runtime failures.

- **Refactors**
  - Removed unused JWT encryption/decryption utilities.
  - Replaced OIDC’s local JWT lookup with shared getJwtPlugin.
  - Added a test to ensure init fails without the JWT plugin.

### DIFF
--- a/packages/better-auth/src/plugins/jwt/utils.ts
+++ b/packages/better-auth/src/plugins/jwt/utils.ts
@@ -1,80 +1,11 @@
-import { subtle, getRandomValues } from "@better-auth/utils";
-import { base64 } from "@better-auth/utils/base64";
+import type { AuthContext } from "../../types";
+import { BetterAuthError } from "../../error";
+import type { jwt } from "./index";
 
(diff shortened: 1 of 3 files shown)


dvanmali:
Closing, each plugin can do this themselves and fail if they wish

PR #4075, opened 2025-08-18
fix (oidc): payload times should be in sec

dvanmali:
According to the [OAuth spec](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2), `iat` and `exp` should be in seconds. Database values should reflect these values as well.

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/oidc.test.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/oidc.test.ts
@@ -683,8 +683,8 @@ describe("oidc-jwt", async () => {
 					jwkSet,
 				);
 				expect(checkSignature).toBeDefined();
-				expect(Number.isInteger(checkSignature.payload.iat)).toBeTruthy()
-				expect(Number.isInteger(checkSignature.payload.exp)).toBeTruthy()
+				expect(Number.isInteger(checkSignature.payload.iat)).toBeTruthy();
+				expect(Number.isInteger(checkSignature.payload.exp)).toBeTruthy();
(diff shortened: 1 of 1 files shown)


himself65 (commented):
looks fair. but we might need to upgrade the database.

@Bekacru How do we handle user side database upgrade?

dvanmali:
Database shouldn't need changes, the object storage is still in Date format. It is just floored to the nearest second:

```ts
new Date(iat * 1000)
```

himself65:
> Database shouldn't need changes, the object storage is still in Date format. It is just floored to the nearest second:
> 
> ```ts
> new Date(iat * 1000)
> ```

Yeah you are right... 

But I just realized we don't have any migration logic in the current codebase, not sure if we're going to have something breaking in the future.

himself65 (commented):
Can you make a test case for this? to make sure signjwt is in second (not ms)

dvanmali:
> Can you make a test case for this? to make sure signjwt is in second (not ms)

I don't think there is a good way to do this. Best I can do is ensure it is an integer.

dvanmali:
> But I just realized we don't have any migration logic in the current codebase, not sure if we're going to have something breaking in the future.

If yours is currently stored as a number, the logic here shouldn't affect the stored value. Even if the Date value is stored via an database adapter that doesn't support dates, (stored as a string or number), it will still be stored properly.

> Database values should reflect these values as well.

I wrote this statement because database values should reflect the payload not a `new Date()` value which changes in ns every call. It is just a precision change.

Issue #4080, opened 2025-08-19
1.3.7 broke auth with existing API keys


tehnrd:
Existing API Keys have stopped working with 1.3.7.

This is with SvelteKit, I haven't been able to test other frameworks.

This feels like a critical issue.

### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Use an API Key for auth on 1.3.6, works.

Upgraded to 1.3.7.

Use the same API Key, it starts throwing:
```
[APIError] {
  status: 'UNAUTHORIZED',
  body: undefined,
  headers: {},
  statusCode: 401
}
```

### Current vs. Expected behavior

API Keys should continue to work after package upgrades.

### What version of Better Auth are you using?

1.3.7

### System info

```bash
System:
    OS: macOS 15.5
    CPU: (10) arm64 Apple M1 Pro
    Memory: 227.45 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 139.0.7258.128
    Safari: 18.5
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript

```

### Additional context

Probably related to https://github.com/better-auth/better-auth/commit/b386d5edc13b94adff3b2e39c2fd131ae2c786a2 but I haven't had time to investigate.

ping-maxwell:
Hey do you know which API endpoint specifically is throwing the error?

tehnrd:
Not yet. Do you know the best way to identify this? console.trace() isn't showing anything helpful or specific to better-auth. It's all server-side, so methods are being called directly and not through the HTTP router.

ping-maxwell:
As far as I know when checking changes between 1.3.6 and 1.3.7 is the logic change in the `x-api-key` in get session requests, but I thought there was a custom error message if it's unauthenticated, which I don't see in your issue post

EDIT: Looks like @himself65 is onto something 👀

himself65:
> Not yet. Do you know the best way to identify this? console.trace() isn't showing anything helpful or specific to better-auth. It's all server-side, so methods are being called directly and not through the HTTP router.

In the latest version we have a errorWithStack property in the APIError

himself65:
> As far as I know when checking changes between 1.3.6 and 1.3.7 is the logic change in the `x-api-key` in get session requests, but I thought there was a custom error message if it's unauthenticated, which I don't see in your issue post
> 
[...]

PR #4083, opened 2025-08-19
fix: check `x-api-key` for all auth endpoint

himself65:
Fixes: https://github.com/better-auth/better-auth/issues/4080
    

---

## Summary by cubic
Always set the session context when authenticating with an API key, not just for /get-session. Fixes API key requests failing on other endpoints and enables using x-api-key across the API.

- **Bug Fixes**
  - Set ctx.context.session for all API key–authenticated requests after user validation.
  - Keep /get-session behavior (returns session) while other endpoints now work with API key auth.
  - Added tests: authenticate with x-api-key on list/verify endpoints and ensure session context is set.


tehnrd:
This PR fixed the issue for me.

Issue #2743, opened 2025-05-22
Dynamic Roles and Permissions


gerkim62:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

 Currently, better-auth only supports hardcoded roles in the codebase, which makes it impossible to manage access control dynamically without deploying code changes or when the permissions and roles are not already known beforehand.

### Describe the solution you'd like

I’d like better-auth to support dynamic roles and permissions that can be created and managed at runtime—ideally via an API or admin UI. These roles be stored in  database.

### Describe alternatives you've considered

Building a separate, custom access control layer on top of better-auth, but this duplicates effort and adds complexity.

### Additional context

See this related discussion: [https://www.answeroverflow.com/m/1357026607033880727](https://www.answeroverflow.com/m/1357026607033880727)

yavuz1205:
I'm experiencing the exact same issue and would really appreciate this feature being implemented.

brettfullstack:
I have the same problem and because of this problem I am thinking of changing the package no matter how good it is. I tried to update it with my own additions but I would be very happy if the package updates itself in this area.

Bruhun:
Same issue, can't really go forward with it...

ohabash:
We need this too. 

When my app starts cant we just remove hard coded things and get the data from our own database instead? then when there is a change in those databases we would just restart the access controls?

Please let me know if you think this would work.

sovetski:
It seems like a good idea, but with dynamic roles, wouldn’t we lose the benefits of TypeScript?

ohabash:
there is no type safety in any RBAC system where there is dynamic control. Unless the values are predefined. Or unless they are generated or handled manually. Kind of like prisma generate

gerkim62:
This is a reasonable compromise

thereis:
Looking forward for this feature!

its4lion:
Would be awesome to add this to the admin plugin too!

yavuz1205:
@gerkim62 @Bekacru Since this feature was added in version 1.3.8, we can close this issue.

PR #4087, opened 2025-08-19
feat(org): Dynamic Access Control

ping-maxwell:
https://github.com/better-auth/better-auth/issues/2743

- [x] additional fields support
- [x] tests
- [x] [documentation](https://better-auth-docs-git-fork-ping-maxwell-feat-75babf-better-auth.vercel.app/docs/plugins/organization#dynamic-access-control)

image
image

    

---

## Summary by cubic
Add Dynamic Access Control for organizations so you can create, read, update, and delete org-specific roles at runtime, with async permission checks that include those dynamic roles. This keeps static roles working and unlocks flexible, DB-backed role management.

- New Features
  - New organizationRole table and endpoints: create-role, delete-role, list-roles, get-role, update-role.
  - hasPermission is now async and loads dynamic roles from the DB; new clientSideHasPermission for client-only checks.
  - Added ac resource to default statements; owner/admin get create/read/update/delete; member gets read.
  - Config: dynamicAccessControl.enabled, validateRoleName, allowCreating/Updating/Deleting/Listing/GettingRole, normalizeRoleName, maximumRolesPerOrganization.
  - organizationRole supports additionalFields; infer on client via inferOrgAdditionalFields.
  - Updated org/team/invite/member routes to await hasPermission.

- Migration
  - Run DB migration to add the organizationRole table.
  - Server: provide a shared ac instance and enable dynamicAccessControl in the organization plugin.
  - Client: enable dynamicAccessControl in organizationClient; optionally use inferOrgAdditionalFields.
  - Update code to await hasPermission and pass organizationId when available.
  - Note: checkRolePermission runs on the client and does not include dynamic roles; use hasPermission for dynamic role checks.


packages/better-auth/src/adapters/kysely-adapter/test/state.txt
himself65, line 1: is this expected?

packages/better-auth/src/plugins/organization/has-permission.ts
```diff
@@ -14,20 +19,113 @@ type PermissionExclusive =
...
+
+	return hasPermissionFn(input, acRoles);
+};
+
+const hasPermissionFn = (
+	input: {
+		role: string;
+		options: OrganizationOptions;
+		allowCreatorAllPermissions?: boolean;
+	} & PermissionExclusive,
```
himself65: Can we move this to a single type?

```
{
		role: string;
		options: OrganizationOptions;
		allowCreatorAllPermissions?: boolean;
	}
```

packages/better-auth/src/adapters/kysely-adapter/test/state.txt
ping-maxwell: yeah

packages/better-auth/src/plugins/organization/has-permission.ts
ping-maxwell: Sure

antonvishal:
Guys @ping-maxwell  @himself65 , I wanna ask as a continuation to this does it make sense to give an ability like fine grained access control for org plugin?
For example, an AC like "team.create" where only org members with this permission can create a team. Basically, a toggle on/off per user, managed by the org owner/admin.

If that’s already on the timeline, ignore this.

packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -621,7 +642,52 @@ export const organization = <O extends OrganizationOptions>(options?: O) => {
...
+							required: true,
+							defaultValue() {
+								return new Date();
+							},
+							fieldName: options?.schema?.organizationRole?.fields?.createdAt,
+						},
+						updatedAt: {
+							type: "date",
+							required: false,
+							fieldName: options?.schema?.organizationRole?.fields?.updatedAt,
```
himself65, line 679: onUpdate?

packages/better-auth/src/plugins/organization/routes/crud-access-control.ts
```diff
@@ -0,0 +1,1377 @@
...
+
+			// -----
+			// Make sure the organization doesn't have too many roles.
+			const maximumRolesPerOrganization =
+				typeof options.dynamicAccessControl?.maximumRolesPerOrganization ===
+				"function"
+					? await options.dynamicAccessControl.maximumRolesPerOrganization(
+							organizationId,
+						)
+					: options.dynamicAccessControl?.maximumRolesPerOrganization || 25;
```
himself65:
```suggestion
					: options.dynamicAccessControl?.maximumRolesPerOrganization ?? 25;
```

packages/better-auth/src/plugins/organization/types.ts
```diff
@@ -56,6 +62,184 @@ export interface OrganizationOptions {
...
+		 */
+		enabled?: boolean;
+		/**
+		 * Validate the name of the role. You can throw an error or return `false` to deny that role name.
+		 *
+		 * @param role - The role to validate.
+		 * @param organizationId - The organization id of the organization the role is being created in.
+		 * @returns - A boolean indicating if the role name is valid. If nothing is returned, the role will be created.
+		 */
+		validateRoleName?: ({
```
Bekacru, line 82: We don't need to provide this kind of option. We should have a default way of validating role names, and if they want to add more they should be able to use hooks. I'm aware we have done this before for another plugin but I don't think this is necessary

packages/better-auth/src/plugins/organization/types.ts
```diff
@@ -56,6 +62,184 @@ export interface OrganizationOptions {
...
+		 * Validate the creation of a role. You can throw an error or return `false` to prevent the role from being created.
+		 *
+		 * @param role - The role to validate.
+		 * @param organizationId - The organization id of the organization the role is being created in.
+		 * @param session - The session of the user who is creating the role.
+		 * @param member - The member who is creating the role.
+		 * @param ctx - The context of the request.
+		 * @returns - A boolean indicating if the role is valid. If nothing is returned, the role will be created.
+		 */
+		allowCreatingRole?: (
```
Bekacru, line 99: Why do we have callbacks for the entire CRUD of roles instead of just checking a permission list like we do for other actions? I agree we should have a callback to determine whether an Organization is allowed to create custom roles (so you can, for example, gate role creation based on pricing), but beyond that, why do we need a whole list of callbacks for permission check?

packages/better-auth/src/plugins/organization/types.ts
```diff
@@ -56,6 +62,184 @@ export interface OrganizationOptions {
...
+			ctx: GenericEndpointContext,
+		) => boolean | void | Promise<boolean | void>;
+
+		/**
+		 * Normalize a role name.
+		 *
+		 * @default - role names will be normalized to lowercase.
+		 * @returns - The normalized role.
+		 */
+		normalizeRoleName?: (data: {
```
Bekacru, line 230: Same here role name normalization...shouldn't be a thing. We should have a sane default and going for this option should bea very rare situation and in that case they should be able to use hooks instead

packages/better-auth/src/plugins/organization/types.ts
```diff
@@ -56,6 +62,184 @@ export interface OrganizationOptions {
...
+			role: string;
+			organizationId: string;
+		}) => string | Promise<string>;
+		/**
+		 * The maximum number of roles that can be created for an organization.
+		 *
+		 * @default 25
```
Bekacru, line 237: 25 is kinda random. We should do either do unlimited (so they'll know to limit it) or just 5 or max 10

packages/better-auth/src/plugins/organization/routes/crud-access-control.ts
```diff
@@ -0,0 +1,1377 @@
...
+					},
+				);
+				throw new APIError("FORBIDDEN", {
+					message: "You are not a member of this organization.",
+				});
+			}
+
+			// -----
+			// Check if the user has the permission to create a role by checking against `ac` resource on the `create` permission.
+			const canCreateRole = await hasPermission(
```
Bekacru, line 168: if we check this why do we have options for `allowCreatingRole`, `allowUpdatingRole`...

packages/better-auth/src/plugins/organization/routes/crud-access-control.ts
```diff
@@ -0,0 +1,1377 @@
...
+			// Validate the role name.
+			await validateRoleName({
+				role: roleName,
+				organizationId,
+				options,
+			});
+
+			// -----
+			// Check if the role name is already taken by a pre-defined/hard-coded role.
+			await checkIfRoleNameIsTakenByPreDefinedRole({
```
Bekacru, line 229: this should be done early

packages/better-auth/src/plugins/organization/routes/crud-access-control.ts
```diff
@@ -0,0 +1,1377 @@
...
+			// Check if the role name is already taken by a role in the database.
+			await checkIfRoleNameIsTakenByRoleInDB({
+				ctx,
+				organizationId,
+				role: roleName,
+			});
+
+			// -----
+			// Make sure the organization doesn't have too many roles.
+			const maximumRolesPerOrganization =
```
Bekacru, line 246: this shoudl be done early

packages/better-auth/src/plugins/organization/routes/crud-access-control.ts
```diff
@@ -0,0 +1,1377 @@
...
+			}
+
+			// -----
+			// Return the roles
+			return ctx.json(roles);
+		},
+	);
+};
+
+// todo: endpoint additional fields vvvvv
```
Bekacru, line 715: ?

maelp:
Nice! What would be cool too is more fine-grained permissions than role, eg something that lets you say "user has edit on posts which match post.author_id === user.id" etc, is that possible?

PR #4090, opened 2025-08-19
docs(google): Fix the prompt value to be the expected typescript comp…

tharakadesilva:
…liant value

### DIFF
--- a/docs/content/docs/authentication/google.mdx
+++ b/docs/content/docs/authentication/google.mdx
@@ -123,7 +123,7 @@ This will trigger a new OAuth flow that requests the additional scopes. After co
 Google only issues a refresh token the first time a user consents to your app.
 If the user has already authorized your app, subsequent OAuth flows will only return an access token, not a refresh token.
 
-To always get a refresh token, you can set the `accessType` to `offline`, and `prompt` to `select_account consent` in the provider options.
+To always get a refresh token, you can set the `accessType` to `offline`, and `prompt` to `select_account+consent` in the provider options.
 
 ```ts
 socialProviders: {
@@ -131,7 +131,7 @@ socialProviders: {
         clientId: process.env.GOOGLE_CLIENT_ID as string,
         clientSecret: process.env.GOOGLE_CLIENT_SECRET as string,
         accessType: "offline", // [!code highlight]
-        prompt: "select_account consent", // [!code highlight]
+        prompt: "select_account+consent", // [!code highlight]
     },
 }
 ```


docs/content/docs/authentication/google.mdx
```diff
@@ -123,15 +123,15 @@ This will trigger a new OAuth flow that requests the additional scopes. After co
 Google only issues a refresh token the first time a user consents to your app.
 If the user has already authorized your app, subsequent OAuth flows will only return an access token, not a refresh token.
 
-To always get a refresh token, you can set the `accessType` to `offline`, and `prompt` to `select_account consent` in the provider options.
+To always get a refresh token, you can set the `accessType` to `offline`, and `prompt` to `select_account+consent` in the provider options.
```
himself65, line 126: You guys are in conflict with each other

https://github.com/better-auth/better-auth/pull/4027/files#diff-5cfbfb96b38fe9d21d0f216d2eadb2d04ffd54939c335d2e72ee811a5474ed52R126

Which one is correct? Can you show me the document?

tharakadesilva: `select_account consent` is correct for Google.

But, BA's type system is using `select_account+consent`. Here's another related MR: https://github.com/better-auth/better-auth/pull/3303

Happy to update the type files instead if you prefer.

tharakadesilva: Code here: https://github.com/ping-maxwell/better-auth/blob/86abb8f340754baef323b9e93115d5073ae69d8b/packages/better-auth/src/oauth2/types.ts#L174

> Happy to update the type files instead if you prefer.

I don't think we should do this, it might break other oauth stuff.

tharakadesilva: The doc for Google is at: https://developers.google.com/identity/protocols/oauth2/web-server

> A space-delimited, case-sensitive list of prompts to present the user

himself65: @ping-maxwell can you look at this?

ping-maxwell: I think the issue is `main` is using the old code which has `select_account+consent` as a type.
But recently I think we changed that in favour of `select_account consent` as a type in canary or newer versions and updated the docs on main
@tharakadesilva are you on latest?

tharakadesilva: Ah, that makes sense!!

I am on `1.3.4` (have to be pinned because of Convex).

I'll close the ticket, thanks folks for the help!

PR #4091, opened 2025-08-19
fix: show error stack in debug mode

himself65:
DX improvement: show the error stack when the user enables debug log.

Right now, it's pretty hard to trace which module throws the error

Related feedback: https://github.com/better-auth/better-auth/issues/4080#issuecomment-3201743847

### DIFF
--- a/packages/better-auth/src/api/to-auth-endpoints.ts
+++ b/packages/better-auth/src/api/to-auth-endpoints.ts
@@ -8,6 +8,7 @@ import {
 import type { AuthEndpoint, AuthMiddleware } from "./call";
 import type { AuthContext, HookEndpointContext } from "../types";
 import { createDefu } from "defu";
+import { shouldPublishLog } from "../utils";
 
 type InternalContext = InputContext<string, any> &
 	EndpointContext<string, any> & {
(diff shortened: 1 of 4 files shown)


packages/better-auth/src/api/to-auth-endpoints.ts
himself65, line 1: I think we should have a better way to handle the error with stack. might do this later

packages/better-auth/src/api/to-auth-endpoints.ts
```diff
@@ -109,12 +121,31 @@ export function toAuthEndpoints<E extends Record<string, AuthEndpoint>>(
 			internalContext.context.returned = result.response;
 			internalContext.context.responseHeaders = result.headers;
 
-			const after = await runAfterHooks(internalContext, afterHooks);
+			const after = await runAfterHooks(internalContext, afterHooks).catch(
```
Bekacru, line 124: not sure if this doing what you're intended to do since `runAfterHooks` already does the catch

himself65: yeah I see, let me fix it

PR #4117, opened 2025-08-21
fix: CLI recognizes default export auth config

Kinfe123:
## Summary by cubic
Fixes CLI config loading to recognize default-exported auth instances, so projects exporting auth as default no longer fail to load. Also supports configs exposed directly via options.

- **Bug Fixes**
  - getConfig now resolves from auth.options, default export objects (default.options), and top-level options.
  - Applies the same resolution logic for both the main and fallback config paths.
  - Adds a test for default-exported auth configuration.


Kinfe123:
closes since it already fixed.

PR #4122, opened 2025-08-21
feat/last-used-social-provider

frectonz:
This PR introduces a new plugin `lastSocialProvider` which records which social provider a user used to log in the last time they logged in.


himself65:
What's the reason you closed the PR? I feel like it's okay

himself65:
Ok, it's here https://github.com/better-auth/better-auth/pull/4128

Issue #3082, opened 2025-06-19
Issue with set-auth-token acquisition not working on OnSuccess behavior


EasyDevv:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Create a function for each login on the frontend to test behavior

### Current vs. Expected behavior

Only authClient.signIn.email can get the bearer token properly and,
Not obtained when using authClient.getSession, authClient.signIn.social, or createAuthClient.

### What version of Better Auth are you using?

1.2.9

### Provide environment information

```bash
- Windows11
- Chrome
- Astrojs + Svelte5
```

### Which area(s) are affected? (Select all that apply)

Client

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
  socialProviders: {
        google: {
            prompt: "consent",
            clientId: process.env.GOOGLE_CLIENT_ID!,
            clientSecret: process.env.GOOGLE_CLIENT_SECRET!,
        },
    },
  plugins: [bearer()]
});
```

## ✅ Successful code
```ts
import { authClient } from "@lib/auth-client";

let bearerToken = null

authClient.signIn.email(
    {
        email: email,
        password: password,
    },
    {
        onSuccess:  (ctx) => {
            bearerToken = ctx.response.headers.get("set-auth-token");
            localStorage.setItem("bearer_token", bearerToken ?? "");
        },
    }
);
```

## ❌ Failed code
result is `null`

```ts
import { authClient } from "@lib/auth-client";

let bearerToken = null

authClient.getSession({
    fetchOptions: {
        onSuccess: (ctx) => {
            bearerToken = ctx.response.headers.get("set-auth-token");
            localStorage.setItem("bearer_token", bearerToken ?? "");
        },
    },
});

authClient.signIn.social(
        {
            provider: "google",
        },
        {
            onSuccess: (ctx) => {
                bearerToken = ctx.response.headers.get("set-auth-token");
                localStorage.setItem("bearer_token", bearerToken ?? "");
            },
        },
    );
```

```ts
export const authClient = createAuthClient({
    fetchOptions: {
        onSuccess: (ctx) => {
            const authToken = ctx.response.headers.get("set-auth-token") // get the token from the response headers
            // Store the token securely (e.g., in localStorage)
            if (authToken) {
                localStorage.setItem("bearer_token", authToken);
            }
        }
    }
})
```

[...]

PR #4123, opened 2025-08-21
fix(bearer): certain sign-in endpoints won't give bearer token

ping-maxwell:
Closes https://github.com/better-auth/better-auth/issues/3082

## What's this PR for?
Normal sign-in endpoints will return a response which includes the `set-auth-token` header from the bearer plugin, however some sign-in endpoints such as social logins or magic links where the set-cookie response comes from a separate `/api/auth` page will cause the authClient not pick up on the bearer token. This PR addresses this issue.

## How does it address it?
We can address this issue by checking if the `set-cookie` response in the hook also contains a `location` header thus meaning it will be redirected. (For example social logins where the user is redirected to /callback which will give a `set-cookie` as well as a `location` header) If we find out that it does include a location value, then we can set a cookie called `bearer-token-confirmation` which is just `true`. Then, on the client bearer plugin, on-start-up it will check if that cookie exists, and if so, hit a newly `/get-bearer-value` endpoint. This endpoint will then remove the `bearer-token-confirmation` cookie and then return the bearer token by just grabbing it from the session cookie value. Once it's returned, the authClient's `onSuccess` global hook would pickup on that response and thus the new bearer token is returned.

An example flow will look like:
image

## Breaking Changes / Migrations?

Nothing will break, however the need to add `bearerClient` plugin on the authClient will be required in order for social logins/magic link or other similar sign-ins to work.

    

---

## Summary by cubic
Fixes missing bearer tokens after social login and magic link sign-ins that redirect, so clients reliably receive a token on startup. Uses a temporary cookie and a new endpoint to recover the token after redirects.

- **Bug Fixes**
  - Add bearer client plugin that, on load, checks for a bearer-token cookie and calls /get-bearer-token.
  - Introduce GET /get-bearer-token to read the session cookie, set set-auth-token, and clear the bearer-token cookie.
  - On sign-in responses that include Set-Cookie and a Location header (redirect flows), set a temporary bearer-token-confirmation=true cookie to trigger recovery.

- **Migration**
  - Add bearerClient() to the auth client plugins to enable token recovery on startup.


packages/better-auth/src/plugins/bearer/index.ts
```diff
@@ -21,6 +21,43 @@ interface BearerOptions {
...
+					method: "GET",
+					metadata: {
+						client: false,
+					},
+					requireHeaders: true,
+				},
+				async (ctx) => {
+					const cookieString = ctx.headers.get("cookie");
+					if (!cookieString) return null;
+					const cookies = cookieString.split("; ");
```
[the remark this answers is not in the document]
ping-maxwell: how would you validate it?

packages/better-auth/src/plugins/bearer/index.ts
```diff
@@ -114,6 +151,19 @@ export const bearer = (options?: BearerOptions) => {
...
+							ctx.context.responseHeaders?.get("location") ||
+							ctx.context.responseHeaders?.get("Location");
+						// If location exists, it likely means the authClient isn't able to pick up the bearer token.
+						// We will store a "bearer-token" cookie so that when the authClient loads it can hit the
+						// `/get-bearer-token` endpoint and check for it, then delete it and return the bearer token.
+						if (location) {
+							// set a temporary cookie that will be used to get the bearer token
+							ctx.setHeader(
+								"set-cookie",
+								`bearer-token=true; Path=/; SameSite=Strict`,
```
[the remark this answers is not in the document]
ping-maxwell: It's intentionally not httpOnly, to be read on the client plugin

packages/better-auth/src/plugins/bearer/client.ts
```diff
@@ -0,0 +1,17 @@
+import type { BetterAuthClientPlugin } from "../../types";
+
+export const bearerClient = () => {
+	return {
+		id: "bearer",
+		getActions($fetch, $store, options) {
```
himself65, line 6: We shouldn't call the side effect inside `getActions`. Is there any other way?

packages/better-auth/src/plugins/bearer/index.ts
```diff
@@ -13,14 +13,76 @@ interface BearerOptions {
...
+				async (ctx) => {
+					const cookieString = ctx.headers.get("cookie");
+					if (!cookieString) {
+						return ctx.json({
+							success: false,
+						});
+					}
+					const cookies = cookieString
+						.split(";")
+						.map((cookie) => cookie.trim());
```
himself65, line 51: here you manually parse the cookie, but I guess there's `ctx.getCookie` utility to use

packages/better-auth/src/plugins/bearer/index.ts
```diff
@@ -13,14 +13,76 @@ interface BearerOptions {
...
+							sessionCookie["max-age"] === 0
+						) {
+							return;
+						}
+						const token = sessionCookie.value;
+						ctx.setHeader("set-auth-token", token);
+						ctx.setHeader(
+							"set-cookie",
+							`${bearerConfirmationCookieName}=; Path=/; HttpOnly; SameSite=Strict; Max-Age=0; Expires=Thu, 01 Jan 1970 00:00:00 GMT`,
+						);
```
himself65, line 75: use ctx utilities

packages/better-auth/src/test-utils/test-instance.ts
```diff
@@ -247,6 +248,10 @@ export async function getTestInstance<
 
 	const client = createAuthClient({
 		...(config?.clientOptions as C extends undefined ? {} : C),
+		plugins: [
+			bearerClient(),
```
himself65, line 252: Don't put this into the top level. And you changed the top-level property, but no new test was added. Which means this plugin didn't affect anything. Could you add some tests for the bearer token?

packages/better-auth/src/plugins/bearer/client.ts
ping-maxwell: It's done so that the authClient's `onSuccess` will get triggered, not sure of another way but open for suggestions

PR #4128, opened 2025-08-21
feat/last-login-method

frectonz:


frectonz:
I think what you’re thinking of is something like this https://github.com/miki-tebe/remember-last-method , but I think if we go in that direction, this PR doesn't make sense, and this feature shouldn't be part of better auth itself and should be made a separate package.

I agree with the point you made about having to handle each kind of sign-in method. I didn't like doing that too. I initially implemented this PR to only handle OAuth login, but since we want to support different kinds of logins, I had to go down this direction. I asked about different ways of doing this but wasn't able to find any.

But then again, the only dependency the hooks have on the separate signing methods is the endpoint path, and that rarely changes, if ever. As you can see, I have also added tests for each kind of login, so if the plugin breaks, we will know about it.

I opted to store the data in a signed cookie. To have it be more secure, we don't want to leak info about user sign-ins after they sign out. If we had stored it in local storage, any script on the website would be able to know that a user has signed in before using the existence of the local storage value, but if we have it stored in a cookie, they won't have access to that information.

I will go over this PR once again and try to come up with better ways to do it.

frectonz:
Another point I would want to make is the benefit to the user when they use this plugin is that they don't have to set the `last-login-method` in an `onSuccess` hook or something like that. The offer to the user is that better auth will keep track of the `last-login-method` for them automatically.

maelp:
I would be interested in just a "lastLoginAt" feature, which is set by all the email/magic-link/social providers

packages/better-auth/src/plugins/siwe/siwe.test.ts
```diff
@@ -352,24 +353,40 @@ describe("siwe", async (it) => {
 							);
 						},
 					}),
+					lastLoginMethod({ storage: "cookie" }),
 				],
 			},
 			{
 				clientOptions: {
-					plugins: [siweClient()],
+					plugins: [siweClient(), lastLoginMethodClient({ storage: "cookie" })],
```
Bekacru: we should generally avoid testing other plugins inside other plugins test. If you need to test it with SIWE, add the SIWE test in last login method test

packages/better-auth/src/plugins/phone-number/phone-number.test.ts
```diff
@@ -197,11 +203,25 @@ describe("phone auth flow", async () => {
...
+			},
+			{
+				onSuccess(context) {
+					cookieSetter(signInHeaders)(context);
+				},
+			},
+		);
 		expect(res.data?.token).toBeDefined();
+
+		const lastUsedLogin = await client.lastUsedLoginMethod(
```
Bekacru, line 217: same here

Bekacru: remove all login method tests you added on other plugins tests

packages/better-auth/src/plugins/siwe/siwe.test.ts
frectonz: I chose this approach because setting up most the sign in methods requires a whole lot of boilerplate, I thought that was unwarranted for a simple plugin that sets a single cookie.

Plus if I just copy over the boilerplate and set up a separate test it would mean that that test would also need to be update every time the `siwe` plugin is updated. Changes like adding an option would need to duplicated twice.

Issue #2307, opened 2025-04-16
APIKey Table schema generation is incorrect


djt:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Create backend
2. Add apiKey plugin
3. Add custom modelName for user table in config:
`
export const auth = betterAuth({
  ...
  user: {
    modelName: 'users'
  },
  ...
});
`
4. npx @better-auth/cli generate

The reference for the user Id will reference the default table name of user.

`
userId: text('user_id').notNull().references(()=> user.id, { onDelete: 'cascade' }),
`

### Current vs. Expected behavior

Current:
`
userId: text('user_id').notNull().references(()=> user.id, { onDelete: 'cascade' }),
`
Expected:
`
userId: text('user_id').notNull().references(()=> users.id, { onDelete: 'cascade' }),
`

### What version of Better Auth are you using?

1.2.7

### Provide environment information

```bash
- OS: Mac OSX
- Browser: Chrome
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});
```

ping-maxwell:
@djt if you update to latest, do you still run into this issue?

johnrjj:
Happening to me as well

changalberto:
Thought I was going crazy. But I'm having this issue as well. Renaming the core user, and session table name cause the cli to generate bad schemas. The other core tables, account, verification are unaffected by the model name change.

Here's my config:

```
export const auth = betterAuth({
  adapter: drizzleAdapter(db, {
    provider: 'pg', // PostgreSQL
    schema: {
      user: users,
      session: sessions,
    },
    usePlural: true, // Explicitly enforce plural naming convention
  }),
  database: new Pool({
    connectionString: env.DATABASE_URL_DIRECT,
  }),
  user: {
    modelName: 'users',
    fields: {
      emailVerified: 'email_verified',
      createdAt: 'created_at',
      updatedAt: 'updated_at',
    },
  },
  session: {
    modelName: 'sessions',
    fields: {
      userId: 'user_id',
      expiresAt: 'expires_at',
      ipAddress: 'ip_address',
      userAgent: 'user_agent',
      createdAt: 'created_at',
      updatedAt: 'updated_at',
    },
  },
  account: {
    modelName: 'accounts',
    fields: {
[...]

PR #4132, opened 2025-08-21
fix(cli): Incorrect drizzle schema gen

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/2307
    

---

## Summary by cubic
Fixes incorrect Drizzle schema generation for foreign key references. The generator now resolves the correct model name from the tables map, preventing broken .references() targets.

- **Bug Fixes**
  - Use tables[ref.model]?.modelName || ref.model when building .references() so custom model names are respected.


himself65 (commented):
let's make a test for this?

ping-maxwell:
> let's make a test for this?

I don't think we have any tests for our CLI, it might be too big overall for this PR (want to keep it small for quick fix)
I think we should open a separate/new PR to introduce tests

himself65:
> > let's make a test for this?
> 
> I don't think we have any tests for our CLI, it might be too big overall for this PR (want to keep it small for quick fix) I think we should open a separate/new PR to introduce tests

kk, we can do this later

Bekacru:
hmm we have tests for the CLI @ping-maxwell

ping-maxwell:
> hmm we have tests for the CLI @ping-maxwell

Oh what, how did I miss this

okay will add tests

himself65 (approved):

PR #4139, opened 2025-08-21
feat: optimize vercel build on avoiding depends

Kinfe123:
## Summary by cubic
Optimize Vercel builds by scoping Turbo to demo and docs, and skipping unnecessary monorepo builds. Also pin demo to published betas and make docs routes dynamic for reliable deploys.

- **Refactors**
  - Add vercel.json to build only @better-auth/demo and @better-auth/docs.
  - Set build.dependsOn: [] in demo/docs and disable workspace linking in .npmrc.
  - Force dynamic rendering for docs pages.
  - Silence a type check in the demo sign-in button to unblock builds.

- **Dependencies**
  - Pin better-auth and @better-auth/stripe in the demo to 1.3.8-beta.2 and update the lockfile.


demo/nextjs/components/sign-in-btn.tsx
```diff
@@ -4,6 +4,7 @@ import { auth } from "@/lib/auth";
 import { headers } from "next/headers";
 
 export async function SignInButton() {
+	// @ts-ignore
```
himself65, line 7: ts-expect-error would be better

himself65:
This is not what we expected. We need build from source code for demoing

Kinfe123:
I have not finished it as it is already on draft I'm experimenting on how can I optimize it by avoid the workspace mentions getting build for no reason.

himself65:
Sorry. But that’s sounds good. Feel free to re-open it when you are ready

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #4140, opened 2025-08-21
demo: fix domain issue on preview env

himself65:


himself65:
It basically works. But we might need `trustedOrigins` to allow all Vercel domains. But right now it's okay

PR #4142, opened 2025-08-22
Renamed `auth-client.ts` to `auth.client.ts` to match actual usage references in the docs.

ooyay:
## Summary by cubic
Fixed a filename typo in the Remix integration docs so examples match the actual file name. Renamed auth-client.ts to auth.client.ts in the text and code block title.


Kinfe123:
the .server.ts is just there for server only gurantees. there is no such for the client. so updated it according to that.

Issue #4150, opened 2025-08-22
Tests on canary are failing


dvanmali:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Run any test. For example:

```sh
pnpm test src/plugins/jwt/jwt.test.ts
```

```
# SERVER_ERROR:  TypeError: Cannot read properties of undefined (reading 'stack')
    at api.<computed> (.../better-auth/packages/better-auth/src/api/to-auth-endpoints.ts:125:60)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
    at processRequest (.../better-auth/node_modules/better-call/src/router.ts:194:22)
    at handler (.../better-auth/node_modules/better-call/src/router.ts:215:16)
    at betterFetch (.../better-auth/node_modules/@better-fetch/fetch/src/fetch.ts:77:17)
    at $fetch (.../better-auth/node_modules/@better-fetch/fetch/src/create-fetch/index.ts:95:10)
    at Object.apply (.../better-auth/packages/better-auth/src/client/proxy.ts:80:12)
    at .../better-auth/packages/better-auth/src/plugins/oauth-provider/register.test.ts:57:20
    at .../better-auth/node_modules/@vitest/runner/dist/chunk-hooks.js:752:20
```

### Current vs. Expected behavior

Began within the past ~15 commits

### What version of Better Auth are you using?

canary

### System info

```bash
na
```

### Which area(s) are affected? (Select all that apply)

Package

### Auth config (if applicable)

```typescript

```

NalinDalal:
uhh, can i approach this one?

Kinfe123:
please make sure to update better-call to the latest and do pnpm install

PR #4152, opened 2025-08-22
fix: prevent TypeError when errorWithStack is undefined in to-auth-endpoints

NalinDalal:
## Summary

This PR fixes a bug where tests would fail with a `TypeError: Cannot read properties of undefined (reading 'stack')` in `to-auth-endpoints.ts` when `errorWithStack` is undefined.

## Details

- Adds a nullish check before accessing `errorWithStack.stack`.
- Prevents server errors during test runs and improves error handling.

## Reproduction

Run any test (e.g. `pnpm test src/plugins/jwt/jwt.test.ts`) on canary before this fix to see the error.

## Related

- Traced to PR #4094
- Issue #4150

---

## Summary by cubic
Prevents a TypeError in to-auth-endpoints by guarding stack access when errorWithStack is undefined. Stops test crashes and keeps debug stacks when available.

- **Bug Fixes**
  - Use optional chaining when assigning APIError.stack from errorWithStack in endpoint and before/after hooks.
  - Avoids server/test failures and preserves stack traces in debug mode when present.

### DIFF
--- a/packages/better-auth/src/api/to-auth-endpoints.ts
+++ b/packages/better-auth/src/api/to-auth-endpoints.ts
@@ -1,9 +1,9 @@
 import {
-	APIError,
-	toResponse,
-	type EndpointContext,
-	type EndpointOptions,
-	type InputContext,
+  APIError,
+  toResponse,
+  type EndpointContext,
+  type EndpointOptions,
+  type InputContext,
 } from "better-call";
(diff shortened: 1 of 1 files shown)


Kinfe123:
please make sure to update better-call to the latest and do `pnpm install`

PR #4182, opened 2025-08-23
fix(docs): pin fumadocs-typescript to exact version 4.0.6

### LINKED ISSUE #4181: Bug: createGenerator not exported from fumadocs-typescript
### Description
When running the docs with pnpm run dev, the build fails because fumadocs-typescript does not export createGenerator.

It fails with following error : 

Image

### Steps to Reproduce
1. Clone the repository  
2. Install dependencies with `pnpm install`  
3. Run the docs with `pnpm run docs`  
4. Observe the error  

### Expected Behavior
The docs should start successfully without throwing any errors.  

### Actual Behavior
The docs fail to run and display the above error.  

### Environment
- **OS:** [Fedora Linux]  
- **Node.js version:** [v20.19.1]  
- **Package manager:** pnpm [10.11.0]

ShobhitPatra:
closes #4181 for now.
    

---

## Summary by cubic
Pin fumadocs-typescript in docs to exact version 4.0.6 to prevent unexpected upgrades that could break the docs build. Updates package.json and pnpm-lock.yaml to lock the version.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -48,7 +48,7 @@
     "fumadocs-core": "15.7.1",
     "fumadocs-docgen": "2.1.0",
     "fumadocs-mdx": "11.8.0",
-    "fumadocs-typescript": "^4.0.6",
+    "fumadocs-typescript": "4.0.6",
     "fumadocs-ui": "15.7.1",
     "gray-matter": "^4.0.3",
     "input-otp": "^1.4.1",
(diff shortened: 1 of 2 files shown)


himself65:
Do you know if this is the correct solution? Lock file hash doesn't chang,e which means it still installs the same version

ShobhitPatra:
@himself65  you are absolutely right,thanks for pointing it out.
After investigating further, the issue wasn't actually with the package version - the lock file hash didn't change, so we were already installing the exact same version.
The real problem was corrupted build cache and node_modules. The fix was simply:
```shell
rm -rf .next
rm -rf node_modules  
rm pnpm-lock.yaml
pnpm install
```
This resolved the ERR_PACKAGE_PATH_NOT_EXPORTED error and pnpm run docs now works perfectly.

ShobhitPatra:
@himself65  I'll close this PR since the version pinning wasn't the actual solution. Thanks for helping me understand how lock files work

himself65:
> @himself65 I'll close this PR since the version pinning wasn't the actual solution. Thanks for helping me understand how lock files work

Yeah. I will check if our doc build is failing

PR #4192, opened 2025-08-24
fix(stripe): metadata not passing to stripe subscriptions upon create

### LINKED ISSUE #4104: [Stripe] Metadata not passed through to Stripe subscription on upgrade

### To Reproduce

When creating a subscription using the `authClient.subscription.upgrade` method, the `metadata` object does not get applied to the resulting Stripe subscription. Normally Stripe supports attaching metadata to subscriptions, but it seems this plugin does not pass the values through.

Steps to Reproduce:

1. Call authClient.subscription.upgrade with a metadata object.
2. Complete the hosted checkout flow.
3. Check the created subscription in Stripe.

``` typescript
const { error: upgradeError } = await authClient.subscription.upgrade({
  plan,
  successUrl,
  cancelUrl,
  annual: isAnnual,
  metadata: {
    ga_client_id: clientId,
  },
  fetchOptions: {
    headers: {
      'x-currency': currency,
      'x-ga-client-id': clientId
    },
  },
});
```

I also attempted to add the metadata, initially passing it via the headers, and then again in the `getCheckoutSessionParams` hook as part of the metadata.
``` typescript
getCheckoutSessionParams: (_, request) => {
  const currency = request?.headers?.get('x-currency') ?? undefined;
  const clientId = request?.headers?.get('x-ga-client-id') ?? undefined;

  return {
    params: {
      allow_promotion_codes: true,
      currency,
      metadata: {
        test: 'test123',
        ga_client_id: clientId ?? 'no client id',
      },
    },
  };
},
```

Expected behavior:
The Stripe subscription should contain the metadata key/value pairs (e.g., ga_client_id) as provided in the upgrade call.

Actual behavior:
The subscription is created successfully, but the metadata field is missing on the subscription object in Stripe.

1.3.7

### System info

```bash
System:
    OS: Windows 11 10.0.26100
    CPU: (32) x64 AMD Ryzen 9 9950X3D 16-Core Processor
    Memory: 67.17 GB / 93.65 GB
  Browsers:
    Edge: Chromium (139.0.3405.86)
```

Package

```typescript

```

### Additional context

_No response_

melsonic:
Previously, while creating a Stripe subscription, the metadata provided by the user was not being passed to the subscription. This PR ensures that the user-defined metadata from authClient.subscription.upgrade or auth.api.upgradeSubscription is included in the subscription that is created.

context: https://www.better-auth.com/docs/plugins/stripe#creating-a-subscription

/fixes #4104
    

---

## Summary by cubic
Passes user-provided metadata to Stripe when creating a subscription, so values from authClient.subscription.upgrade and auth.api.upgradeSubscription are attached to the resulting subscription.

### DIFF
--- a/packages/stripe/src/index.ts
+++ b/packages/stripe/src/index.ts
@@ -487,6 +487,9 @@ export const stripe = <O extends StripeOptions>(options: O) => {
 							],
 							subscription_data: {
 								...freeTrial,
+								metadata: {
+									...ctx.body?.metadata,
+								}
 							},
 							mode: "subscription",
 							client_reference_id: referenceId,
--- a/packages/stripe/src/index.ts
+++ b/packages/stripe/src/index.ts
@@ -489,7 +489,7 @@ export const stripe = <O extends StripeOptions>(options: O) => {
 								...freeTrial,
 								metadata: {
 									...ctx.body?.metadata,
-								}
+								},
 							},
 							mode: "subscription",
 							client_reference_id: referenceId,


gustavovalverde:
Thanks for the fix. This is already resolved in `main`: `ctx.body.metadata` flows into both `subscription_data.metadata` and the top-level subscription `metadata` (`packages/stripe/src/routes.ts:1129` and 1143). Closing as already fixed.

PR #4196, opened 2025-08-24
Generic OAuth: make mapProfileToUser synchronous-only to prevent silent drops

kyrregjerstad:
### Summary
- Enforces that `mapProfileToUser` must return synchronously.
- Prevents accidental `async` functions from being passed, which previously led to fields being silently dropped when merged in the core callback path.

### Motivation
In the core OAuth callback, the result of `mapProfileToUser` is merged without `await`. When an async function was supplied, the returned Promise was spread (a no-op), causing mapped fields to be lost before `user.create.before`.

### What changed
- Type signature narrowed to disallow Promise return.
- No runtime logic changes. Strictly a DX/type-safety improvement.

### Impact
- Compile-time error if an async `mapProfileToUser` is provided.
- No behavioral changes for existing synchronous mappings.

### Migration guide
- If your `mapProfileToUser` is `async`, make it synchronous.
- Move any async work into `getUserInfo` and return a fully shaped user there.

Example (before, problematic):
```ts
mapProfileToUser: async (profile) => {
	const data = await something()

	return {
		...profile,
		...data
	}
}
```

Example (after, recommended):
```ts
getUserInfo: async (tokens) => {
  const profile = await fetchProviderProfile(tokens);
  return {
    id: profile.sub,
    email: profile.email,
    emailVerified: profile.email_verified,
    name: profile.name,
    firstName: profile.given_name,
  };
},
mapProfileToUser: (profile) => ({
  // Optional: do a final sync mapping if needed
  lastName: profile.family_name,
})
```
    

---

## Summary by cubic
Enforces mapProfileToUser to be synchronous-only in Generic OAuth to prevent silent field drops caused by async functions. Type-only change; existing synchronous mappings are unaffected.

- **Migration**
  - If you used an async mapProfileToUser, make it synchronous.
  - Move any async work to getUserInfo and return the final user shape there.


himself65 (commented):
LGTM, but can you add a test case for this

gustavovalverde:
Thanks for raising this. The silent-drop problem this PR narrows the type to prevent does not exist in current code: every provider in `packages/core/src/social-providers/` awaits `options.mapProfileToUser?.(...)`, so async mappers are fully supported. Enforcing sync-only would remove a pattern the codebase intentionally supports. Closing.

Issue #2682, opened 2025-05-17
Cannot import `expo-web-browser` with Expo for TV


jschoeny:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

- Follow the instructions to create a React Native TV app using Expo: https://docs.expo.dev/guides/building-for-tv/
- Follow the instructions to add Better Auth to the Expo app (client only): https://www.better-auth.com/docs/integrations/expo
- Run the app
- App will crash with the error `Cannot find native module 'ExpoWebBrowser'`, pointing back to `import { expoClient } from '@better-auth/expo/client'`

### Current vs. Expected behavior

Currently, the Better Auth Expo plugin does not distinguish as to whether the current platform is a TV or not. Because Expo for TV does not bundle the `expo-web-browser` package in the app, a crash occurs when Better Auth's `expoClient` plugin tries to import it.

Expected behavior would be to have the `expoClient` plugin not use the `expo-web-browser` package when the current platform is a TV. For reference, the following changes to [packages/expo/src/client.ts](https://github.com/better-auth/better-auth/blob/e7aec18d7d297da2d0c8e3b85c348a459b6d0ae0/packages/expo/src/client.ts#L173-L188) allows the Expo TV app to run without crashing:

```diff
  import type { BetterAuthClientPlugin, Store } from "better-auth";
- import * as Browser from "expo-web-browser";
  import * as Linking from "expo-linking";
  import { Platform } from "react-native";
  import Constants from "expo-constants";
  import { BetterFetchOption } from "@better-fetch/fetch";
```
```diff
      fetchPlugins: [
        {
          id: "expo",
          name: "Expo",
          hooks: {
            async onSuccess(context) {
              if (isWeb) return;
              const setCookie = context.response.headers.get("set-cookie");
              if (setCookie) {
                const prevCookie = await storage.getItem(cookieName);
                const toSetCookie = getSetCookie(
                  setCookie || "",
                  prevCookie ?? undefined,
                );
                await storage.setItem(cookieName, toSetCookie);
                store?.notify("$sessionSignal");
              }

              if (
                context.request.url.toString().includes("/get-session") &&
                !opts?.disableCache
              ) {
                const data = context.data;
                storage.setItem(localCacheName, JSON.stringify(data));
              }

-             if (
-               context.data?.redirect &&
-               context.request.url.toString().includes("/sign-in") &&
-               !context.request?.body.includes("idToken") // id token is used for silent sign-in
-             ) {
[...]

PR #4207, opened 2025-08-24
fix(expo): fix the inability to dynamically import a dependency

fax1ty:
After upgrading to Expo SDK 54 Beta, I encountered an issue with assembling a JS bundle that linked to the `expo-web-browser`. The error message indicated that the module could not be found.

image

This statement seems to imply that the Expo Web Browser is optional, but the code clearly indicates that it is mandatory.

P.S. Also may close https://github.com/better-auth/better-auth/issues/2682

### DIFF
--- a/packages/expo/src/client.ts
+++ b/packages/expo/src/client.ts
@@ -1,5 +1,4 @@
 import { BetterAuthClientPlugin, Store } from "better-auth/types";
-import * as Browser from "expo-web-browser";
 import * as Linking from "expo-linking";
 import { Platform } from "react-native";
 import Constants from "expo-constants";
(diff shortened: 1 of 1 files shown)


packages/expo/src/client.ts
```diff
@@ -209,6 +208,14 @@ export const expoClient = (opts: ExpoClientOptions) => {
...
 							const to = callbackURL;
 							const signInURL = context.data?.url;
+							let Browser: any;
+							try {
+								Browser = await import("expo-web-browser");
+							} catch (error) {
+								throw new Error(
+									"expo-web-browser is not installed as a dependency!",
+								);
+							}
```
himself65:
```suggestion
							const Browser = await import("expo-web-browser");
```

himself65 (commented):
is this good? I think bundler will handle the error correctly

packages/expo/src/client.ts
fax1ty: Unfortunately, the correct syntax for Metro is `try/catch` combined with `require` _(which is unsupported by library's test suite)_ or `import` statements
image

himself65 (approved):

PR #4215, opened 2025-08-25
fix(api): rollback upon failure to create account/session

### LINKED ISSUE #4193: Failure to create an account/session after user creation should rollback

### To Reproduce

1. Setup your schema and auth instance like below.
2. Create a user `await authClient.signUp.email`

```ts
// all tables plural (sessions, accounts)
export const users = pgTable('users', {
	id: text().primaryKey(),
	name: text().notNull(),
	email: text().notNull().unique(),
	emailVerified: boolean('email_verified').notNull(),
	image: text(),
	createdAt: timestamp('created_at', { withTimezone: true }).notNull(),
	updatedAt: timestamp('updated_at', { withTimezone: true }),
})

		provider: 'pg',
		schema,
		debugLogs: true,
	}),
	user: {
		modelName: 'users',
	},
	account: {
        // if you exclude this, user will create but account creation will fail
		// modelName: 'accounts', 
	},
	verification: {
		modelName: 'verifications',
	},
    // also missing session object
})
```

User creation will succeed but not account creation. This happens too if a session is attempted.

It would be expected that these operations work within a database transaction, so if an error occurs the user creation rolls back. 

Currently, the setup described would allow a user without an account or session to be created.

1.3.7

### System info

```bash
System:
    OS: macOS 15.6
    CPU: (10) arm64 Apple M1 Pro
    Memory: 196.30 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Brave Browser: 139.1.81.136
    Chrome: 139.0.7258.139
    Safari: 18.6
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

ShobhitPatra:
### Summary
Fixes #4193 - Ensures proper rollback when user creation succeeds but subsequent account or session creation fails, preventing orphaned user records.

### Problem
Previously, when user creation succeeded but account/session creation failed afterward, the user record would remain in the database without proper cleanup, leading to -Orphaned user records without associated accounts/sessions

### Solution
Wrapped account linking and session creation in comprehensive try-catch block
Implemented rollback mechanism that attempts to delete the created user if subsequent operations fail

    

---

## Summary by cubic
Prevents orphaned users by rolling back sign-up if account linking, email verification, or session creation fails. Fixes #4193.

- **Bug Fixes**
  - Wrapped account linking, email verification, and session creation in a try/catch.
  - On failure, delete the user when supported; otherwise remove sessions and accounts and log warnings to avoid orphans.
  - Early return when autoSignIn is disabled or email verification is required (no session/token).
  - Set session cookie only after successful session creation and propagate original errors.


himself65 (changes requested):
Can you make a test case for this?

himself65:
Do we have a transaction in the current DB adapter? /cc @Bekacru 

I think the best way to have that internally

ping-maxwell:
Yeah that is what I was about to mention @himself65 , we have transaction systems being built rn
Should leave this functionality to that

ping-maxwell:
It's on Linear, and you can find an active PR being built here:
https://github.com/better-auth/better-auth/pull/4062

ShobhitPatra:
@himself65 That was my initial approach actually... but went with this as could not find any  built-in transaction method in the internal adapter

himself65:
> It's on Linear, and you can find an active PR being built here: #4062

let me review that first

Kinfe123:
This should be included with the current transaction support.

packages/better-auth/src/api/routes/sign-up.ts
```diff
@@ -155,146 +155,188 @@ export const signUpEmail = <O extends BetterAuthOptions>() =>
...
-				...additionalFields
-			} = body;
-			const isValidEmail = z.email().safeParse(email);
-
-			if (!isValidEmail.success) {
-				throw new APIError("BAD_REQUEST", {
-					message: BASE_ERROR_CODES.INVALID_EMAIL,
-				});
-			}
+			return ctx.context.adapter.transaction(async (trx) => {
```
himself65, line 158: I updated the code to use `transaction` for all logic.

But issues here are

1. We need to pass the trx to all DB actions; otherwise, it will just be pending forever.
2. I'm not sure if this is a good idea, since we may change all logic within the transaction, the git blame will be broken with a useless diff.

So I might need to implement [AsyncLocalStorage](https://github.com/Bekacru/better-call/issues/47) first, then we will have a light migration to the transaction

PR #4224, opened 2025-08-25
a

TheUntraceable:
## Summary by cubic
Adds OIDC discovery to SSO provider registration and login to auto-populate endpoints and validate configs, improving reliability and spec compliance. Also aligns token endpoint auth method handling with the OIDC spec.

- **New Features**
  - Validate discovery endpoint and issuer; require token, authorization, userinfo, and JWKS endpoints.
  - Auto-fill authorization, token, user info, and JWKS endpoints from discovery while allowing overrides.
  - Added OIDCDiscoveryData type for stricter typing.

- **Bug Fixes**
  - Use token_endpoint_auth_methods_supported per OIDC spec and select the first supported method.
  - Clearer errors when discovery fails or required fields are missing.


TheUntraceable:
Opened this by accident

PR #4233, opened 2025-08-26
docs(admin): Fix calculation of total pages using pageSize

cosmosgenius:
Wrong variable was being used. It should be pageSize instead of limit.
    

---

## Summary by cubic
Corrected the pagination example in the admin plugin docs. totalPages now uses pageSize instead of limit, matching the API and preventing wrong page counts.


himself65 (approved):

himself65:
make sense, thank u

PR #4263, opened 2025-08-27
fix(oidc-provider): change updated_at to be a UNIX numeric timestamp

### LINKED ISSUE #4251: updated_at field in OpenID JWT should be a number

### To Reproduce

Right now, the `updated_at` field in a better-auth JWT is an ISO date string: https://github.com/better-auth/better-auth/blob/eca7fc5ec7f41013c167fa8c79909d27316e5bcc/packages/better-auth/src/plugins/oidc-provider/index.ts#L789

This doesn't match the OpenID spec, which calls for it to be [a JSON number](https://openid.net/specs/openid-connect-core-1_0-final.html#:~:text=updated.-,Its%20value%20is%20a%20JSON%20number%20representing%20the%20number%20of%20seconds%20from%201970%2D01%2D01T0:0:0Z%20as%20measured%20in%20UTC%20until%20the%20date/time.). The name [`updated_at`](https://www.iana.org/assignments/jwt/jwt.xhtml#:~:text=1.0%2C%20Section%205.1]-,updated_at,-Time%20the%20information) is reserved for OpenID.

I noticed this issue because it caused an error in my JWT client implementation, which I worked around.

I expected the `updated_at` field to be a number, but it is a string.

1.3.7

### System info

```bash
N/A
```

Backend

```typescript

```

### Additional context

_No response_

ShobhitPatra:
### closes #4251 

Reference :  https://openid.net/specs/openid-connect-core-1_0-final.html#:~:text=updated.-,Its%20value%20is%20a%20JSON%20number%20representing%20the%20number%20of%20seconds%20from%201970%2D01%2D01T0:0:0Z%20as%20measured%20in%20UTC%20until%20the%20date/time.
    

---
## Summary by cubic
Use a UNIX numeric timestamp (seconds since epoch) for the updated_at claim in OIDC and MCP, instead of an ISO string. This aligns with the OIDC Core spec and prevents client validation errors.

<sup>Written for commit 878a3f736f19b1bec26b1e447e52c9c44b3eb758. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/mcp/index.ts
+++ b/packages/better-auth/src/plugins/mcp/index.ts
@@ -585,7 +585,7 @@ export const mcp = (options: MCPOptions) => {
 						family_name: user.name.split(" ")[1]!,
 						name: user.name,
 						profile: user.image,
-						updated_at: user.updatedAt.toISOString(),
+						updated_at: Math.floor(new Date(user.updatedAt).getTime() / 1000),
 					};
 					const email = {
 						email: user.email,


himself65:
LGTM. please add a test case, thx

conradev:
@ShobhitPatra you likely also want to update this in the MCP JWT generation as well

himself65 (approved):

PR #4265, opened 2025-08-27
fix(drizzle-adapter): resolve fieldName mapping in field validation

### LINKED ISSUE #4211: “fieldName” Mapping in additionalFields Causes “Field Does Not Exist” Error with Drizzle Adapter

### Description
When configuring custom fields in Better Auth’s Drizzle ORM adapter using the fieldName option inside additionalFields, user creation fails with an error claiming the specified database column does not exist, despite the field being defined in the Drizzle schema. Removing the fieldName property makes the process succeed.

### Reproduction Steps

1. Define a custom field in Drizzle schema (db/schema.ts):

```ts
export const user = pgTable("user", {
  // …other columns…
  expiryDate: timestamp("expiry_date")
    .$defaultFn(() => {
      const d = new Date();
      d.setDate(d.getDate() + 7);
      return d;
    })
    .notNull(),
});
```

2. Configure Better Auth with Drizzle adapter and add expiryDate in additionalFields with fieldName:

```ts
import { db } from "@/db/drizzle";
import { schema } from "@/db/schema";

    provider: "pg",
    schema,
  }),
  user: {
    additionalFields: {
      expiryDate: {
        type: "date",
        required: true,
        fieldName: "expiry_date",  // ← Including this causes error
        defaultValue: () => {
          const d = new Date();
          d.setDate(d.getDate() + 7);
          return d;
        },
        input: false,
      },
    },
  },
  // …rest of config…
});
```

3. Attempt to register a new user:

```ts
ERROR [Better Auth]: Failed to create user BetterAuthError: The field "expiry_date" does not exist in the "user" schema. Please update your drizzle schema or re-generate using "npx @better-auth/cli generate".
```

4. Comment out or remove the fieldName line and restart. User creation succeeds without error.

### Expected Behavior

Better Auth should respect the fieldName mapping in additionalFields and correctly find and use the database column defined in the Drizzle schema.

### Actual Behavior

Better Auth ignores or misinterprets the fieldName property in additionalFields, leading to a “field does not exist” error even though the column is present in the schema.

1.3.7

### System info

```bash
System:
    OS: macOS 15.6.1
    CPU: (14) arm64 Apple M4 Max
    Memory: 152.14 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
Browsers:
    Chrome: 139.0.7258.139
    Safari: 18.6
```

Backend

```typescript

    provider: "pg",
    schema,
  }),
  user: {
    additionalFields: {
      expiryDate: {
        type: "date",
        required: true,
        fieldName: "expiry_date",  // ← Including this causes error
        defaultValue: () => {
          const d = new Date();
          d.setDate(d.getDate() + 7);
          return d;
        },
        input: false,
      },
    },
  },
  // …rest of config…
});
```

### Additional context

The error occurs consistently in both local development and CI environments; it is not isolated to a specific machine or network.

The issue manifests during server-side user registration flows and does not depend on any particular browser or client.

Removing the fieldName option in additionalFields immediately resolves the error, confirming the field exists in the Drizzle schema and database.

All other built-in and custom fields without fieldName mappings work correctly, including default Drizzle schema fields and additional fields defined without explicit column mapping.

Verified that the database table contains the expiry_date column and that Drizzle migrations are up to date.

The problem persists after clearing caches, reinstalling node_modules, and regenerating Better Auth artifacts with bunx @better-auth/cli generate.

ShobhitPatra:
### Closes #4211
    

---

## Summary by cubic
Fix field validation in the Drizzle adapter by mapping incoming keys via getFieldName before checking the schema. This prevents false “field does not exist” errors when models use renamed or aliased fields.

### DIFF
--- a/packages/better-auth/src/adapters/drizzle-adapter/test/adapter.drizzle.test.ts
+++ b/packages/better-auth/src/adapters/drizzle-adapter/test/adapter.drizzle.test.ts
@@ -152,3 +152,107 @@ describe("Drizzle Adapter Number Id Test", async () => {
 		},
 	});
 });
+
+describe("Drizzle Adapter fieldName Mapping Tests", async () => {
+	let pg: Pool;
+	let postgres: Kysely<any>;
+
+	beforeAll(async () => {
+		pg = createTestPool();
+		postgres = createKyselyInstance(pg);
+		await cleanupDatabase(postgres, false);
+	});
+
+	afterAll(async () => {
+		await cleanupDatabase(postgres);
+	});
+
+	it("should handle fieldName mapping in additionalFields during user creation", async () => {
+		const optsWithFieldNameMapping = {
+			database: pg,
+			user: {
+				fields: { email: "email_address" },
+				additionalFields: {
+					test: {
+						type: "string",
+						fieldName: "test",
+						defaultValue: "default-test-value",
+					},
+				},
+			},
+			session: {
+				modelName: "sessions",
+			},
+		} satisfies BetterAuthOptions;
+
+		const { runMigrations } = await getMigrations(optsWithFieldNameMapping);
+		await runMigrations();
+
+		const db = drizzle(pg);
+		const auth = betterAuth({
+			...optsWithFieldNameMapping,
+			database: drizzleAdapter(db, { provider: "pg", schema }),
+			emailAndPassword: {
+				enabled: true,
+			},
+		});
+
+		const testUser = {
+			email: "fieldname-test@example.com",
+			password: "password123",
+			name: "FieldName Test User",
+		};
+
+		const result = await auth.api.signUpEmail({ body: testUser });
+
+		expect(result).toBeDefined();
+		expect(result.user).toBeDefined();
+		expect(result.user.email).toBe(testUser.email);
+		expect(result.user.name).toBe(testUser.name);
+	});
+
+	it("should handle case where fieldName differs from property name", async () => {
+		const optsWithDifferentFieldName = {
+			database: pg,
+			user: {
+				fields: {
+					email: "email_address",
+				},
+				additionalFields: {
+					test: {
+						type: "string",
+						defaultValue: "test-value",
+					},
+				},
+			},
+			session: {
+				modelName: "sessions",
+			},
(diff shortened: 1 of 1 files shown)


himself65 (commented):
can you add test cases for this

ShobhitPatra:
@himself65 added the test cases, could you take a  look please . Thanks !

himself65 (approved):

packages/better-auth/src/adapters/drizzle-adapter/test/adapter.drizzle.test.ts
```diff
@@ -152,3 +152,107 @@ describe("Drizzle Adapter Number Id Test", async () => {
...
+			password: "password123",
+			name: "FieldName Test User",
+		};
+
+		const result = await auth.api.signUpEmail({ body: testUser });
+
+		expect(result).toBeDefined();
+		expect(result.user).toBeDefined();
+		expect(result.user.email).toBe(testUser.email);
+		expect(result.user.name).toBe(testUser.name);
```
himself65, line 211: there's no coverage for additional field `test`

packages/better-auth/src/adapters/drizzle-adapter/test/adapter.drizzle.test.ts
```diff
@@ -0,0 +1,258 @@
...
+			//@ts-ignore
+			opts.database = undefined;
+			const merged = merge(opts, customOptions);
+			merged.database = db;
+			return adapter(merged);
+		},
+	});
+});
+
+describe("Drizzle Adapter fieldName Mapping Tests", async () => {
```
ping-maxwell, line 156: We don't use these tests anymore, please use `normal.ts` (located [here](https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/adapters/tests/normal.ts))

If you believe this only makes sense to test specifically for drizzle than you can create a new test suite (follow how other test suites are made in other files like `normal.ts`, `auth-flow.ts`, etc) and name it to be specific to drizzle, then make sure all existing drizzle adapter test uses that newly define test-suite.

ping-maxwell:
Closing as original issue is resolved.

Issue #4261, opened 2025-08-27
Erroneous error returning in organization delete endpoint


pathconnected:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Call `/organization/delete` on an organization where the current user is not member of.

### Current vs. Expected behavior

The backend returns a `200` with empty body, but it should return an error:

```typescript
			if (!member) {
				return ctx.json(null, {
					status: 400,
					body: {
						message:
							ORGANIZATION_ERROR_CODES.USER_IS_NOT_A_MEMBER_OF_THE_ORGANIZATION,
					},
				});
			}
```

### What version of Better Auth are you using?

1.3.7

### System info

```bash
Does not matter
```

### Which area(s) are affected? (Select all that apply)

Package

### Auth config (if applicable)

Add organization plugin.

### Additional context

Seems like the way the error is returned is different from how it's done elsewhere, by throwing an APIError.

PR #4276, opened 2025-08-27
fix: member not exist on org

himself65:
Fixes: https://github.com/better-auth/better-auth/issues/4261
    

## Summary by cubic
Fixes org deletion to return BAD_REQUEST (400) with a standardized APIError when a non-member tries to delete an organization. Adds a test to cover this case.

- **Bug Fixes**
  - Throw APIError with USER_IS_NOT_A_MEMBER_OF_THE_ORGANIZATION when no membership is found in deleteOrganization.
  - Added test: non-member delete attempt returns 400 with the expected error code.

### DIFF
--- a/packages/better-auth/src/plugins/organization/organization.test.ts
+++ b/packages/better-auth/src/plugins/organization/organization.test.ts
@@ -743,6 +743,43 @@ describe("organization", async (it) => {
 		expect(hasMultiplePermissions.data?.success).toBe(true);
 	});
 
+	it("should return BAD_REQUEST when non-member tries to delete organization", async () => {
+		// Create an organization first
+		const testOrg = await client.organization.create({
+			name: "test-delete-org",
+			slug: "test-delete-org",
+			fetchOptions: {
+				headers,
+			},
+		});
+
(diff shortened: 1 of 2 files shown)


pathconnected:
@himself65 good work! Did you notice that there is a similar issue in the same function, when it checks whether the organization exists?

```typescript
			const organizationId = ctx.body.organizationId;
			if (!organizationId) {
				return ctx.json(null, {
					status: 400,
					body: {
						message: ORGANIZATION_ERROR_CODES.ORGANIZATION_NOT_FOUND,
					},
				});
			}
```

himself65:
> @himself65 good work! Did you notice that there is a similar issue in the same function, when it checks whether the organization exists?
> 
> ```ts
> 			const organizationId = ctx.body.organizationId;
> 			if (!organizationId) {
> 				return ctx.json(null, {
> 					status: 400,
> 					body: {
> 						message: ORGANIZATION_ERROR_CODES.ORGANIZATION_NOT_FOUND,
> 					},
> 				});
> 			}
> ```

oh, good catch. will fix that also

PR #4278, opened 2025-08-28
feat: add alias plugin

himself65:


packages/better-auth/src/plugins/alias/client.ts
```diff
@@ -0,0 +1,143 @@
...
+									return fetchPlugin.hooks!.onRequest!(context);
+								}
+							: undefined,
+					}
+				: undefined,
+		}));
+	}
+
+	// Wrap the $InferServerPlugin if it exists
+	if (plugin.$InferServerPlugin) {
```
himself65, line 167: this is wrong, $Inferserverplugin is not a runtime value

packages/better-auth/src/plugins/alias/client.ts
```diff
@@ -0,0 +1,143 @@
...
+ * })
+ * // Now endpoints will be:
+ * // - /api/auth/polar/checkout
+ * // - /api/auth/dodo/checkout
+ * ```
+ */
+export function aliasClient<T extends BetterAuthClientPlugin>(
+	prefix: string,
+	plugin: T,
+): T {
```
himself65: I guess type will incorrect

himself65 (commented):
Right now, because of client-side plugin design, support for aliases is pretty hard.

himself65:
It‘s bearly impossible even now. I think a workaround is to make the plugin endpoint customizable in the init option.

Closing this as no ETA for this PR

PR #4288, opened 2025-08-28
feat: check endpoint conflits

himself65:


packages/better-auth/src/api/index.ts
```diff
@@ -36,11 +37,67 @@ import {
...
+					`  - "${conflict.path}" used by plugins: ${conflict.plugins.join(", ")}`,
+			)
+			.join("\n");
+		logger.error(
+			`Endpoint path conflicts detected! Multiple plugins are trying to use the same endpoint paths:
+${conflictMessages}
+
+To resolve this, you can:
+	1. Use only one of the conflicting plugins
+	2. Configure the plugins to use different paths (if supported)
```
himself65, line 95: I think in the future we will have alias plugin to change the pathname if possible

PR #4298, opened 2025-08-29
Add supportsDates option to allow string as timestampz from db

steipete:
Better Auth currently looks only at the database type to decide if it should handle timestams as Date or string.
This PR extends this feature to allow customizing this behavior:

```ts
    // Force timestamp-like types to be returned as raw strings
    // to preserve full precision for downstream consumers
    types: {
      // PostgreSQL OIDs for timestamp-related types
      timestamptz: { from: [1184], to: 1184, parse: (s: string) => s, serialize: (s: string) => s },
      timestamp: { from: [1114], to: 1114, parse: (s: string) => s, serialize: (s: string) => s },
      date: { from: [1082], to: 1082, parse: (s: string) => s, serialize: (s: string) => s },
      time: { from: [1083], to: 1083, parse: (s: string) => s, serialize: (s: string) => s },
      timetz: { from: [1266], to: 1266, parse: (s: string) => s, serialize: (s: string) => s },
    },

    // later on I init betterAuth:
  // Use Better Auth's Kysely adapter with supportsDates: false for Neon compatibility
  // This prevents crashes when the database driver returns dates as strings
  database: kyselyAdapter(db, {
    type: 'postgres',
    supportsDates: false, // NEW: Fix for Neon/PostgreSQL driver returning dates as strings
  }),
```

Without that, I get `ERROR [Better Auth]: Failed to create user TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of Date`

Took me quite some time to debug, I used AI to write this fix, so the tests defo need work,
but I wanted to get the discussion going first if this is something you consider.

And since I'm fairly new to web tech, also if this pattern is common.
I did talk with @igalklebanov and he confirmed that this is a common pattern.

Alternative design if we wanna be even more explicit (but then I'd expect auto to do a one-time probe)
```ts
  dates: 'auto',           // default
  // dates: 'native',     // force Date passthrough
  // dates: 'string',     // force ISO strings at DB boundary
```

Why go in on all strings?
This allows my project to keep full precision on time and greatly simplify date/time conversion.
Everything is a zod-verified timestamp, date conversion happens at the UI layer.
Especially with React RSC that's super nice, so I can push my existing models to the frontend.

Why not read from the db layer?
I considered reading the config from the database but that gets messy fast - I use Neon in prod which again has a differert type configuration syntax - so a new option seemed easier.

No conversion?
I use `postgres.js` which returns [correct ISO 8601](https://github.com/porsager/postgres/blob/32feb259a3c9abffab761bd1758b3168d9e0cebc/src/types.js#L31), as does Neon. pg returns ISO (with a space instead of T + numeric offset) so folks using that still have to do a tiny conversion or a [json_object trick](https://onecompiler.com/postgresql/43v4ab5kn) to get ISO 8601.
    

---

## Summary by cubic
Adds a supportsDates option to the Kysely adapter to control how date/time fields are handled. This prevents crashes with drivers that return strings (e.g., Neon Postgres) and allows preserving full timestamp precision by using ISO strings.

- **New Features**
  - New config: supportsDates. When false, dates are handled as strings. Defaults remain: sqlite/mssql=false, postgres/mysql=true. Can be overridden per adapter.
  - Added tests to cover defaults and overrides.

- **Migration**
  - If your Postgres driver returns dates as strings (e.g., Neon), set supportsDates: false in kyselyAdapter({ type: 'postgres', supportsDates: false }).

### DIFF
--- a/packages/better-auth/src/adapters/kysely-adapter/test/normal/adapter.kysely.test.ts
+++ b/packages/better-auth/src/adapters/kysely-adapter/test/normal/adapter.kysely.test.ts
@@ -244,3 +244,56 @@ describe("mssql", async () => {
 		});
 	});
 });
+
+describe("supportsDates configuration", () => {
+	it("should respect explicit supportsDates: false for postgres", () => {
+		const adapter = kyselyAdapter(mysqlKy, {
+			type: "postgres",  // Usually defaults to true
+			supportsDates: false,  // Override to false
+		});
+		
+		// Access the internal config to verify the setting
+		const adapterInstance = adapter({
+			database: { db: mysqlKy, type: "postgres" },
+			user: { fields: {} },
+		});
+		
+		expect(adapterInstance.options.adapterConfig.supportsDates).toBe(false);
+	});
+
+	it("should respect explicit supportsDates: true for sqlite", () => {
+		const adapter = kyselyAdapter(sqliteKy, {
+			type: "sqlite",  // Usually defaults to false
+			supportsDates: true,  // Override to true
+		});
+		
+		const adapterInstance = adapter({
+			database: { db: sqliteKy, type: "sqlite" },
+			user: { fields: {} },
+		});
+		
+		expect(adapterInstance.options.adapterConfig.supportsDates).toBe(true);
+	});
+
+	it("should use default behavior when supportsDates is not specified", () => {
+		// Test postgres default (true)
+		const postgresAdapter = kyselyAdapter(mysqlKy, {
+			type: "postgres",
+		});
+		const postgresInstance = postgresAdapter({
+			database: { db: mysqlKy, type: "postgres" },
+			user: { fields: {} },
+		});
+		expect(postgresInstance.options.adapterConfig.supportsDates).toBe(true);
+
+		// Test sqlite default (false)
+		const sqliteAdapter = kyselyAdapter(sqliteKy, {
+			type: "sqlite",
+		});
+		const sqliteInstance = sqliteAdapter({
+			database: { db: sqliteKy, type: "sqlite" },
+			user: { fields: {} },
+		});
+		expect(sqliteInstance.options.adapterConfig.supportsDates).toBe(false);
+	});
+});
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/adapters/kysely-adapter/test/normal/adapter.kysely.test.ts
himself65, line 1: Can you test the final data structure? not just config

himself65 (changes requested):

packages/better-auth/src/adapters/kysely-adapter/test/normal/adapter.kysely.test.ts
steipete: Absolutely! I'm about to sleep, took me a few hours to understand and track down the nuances.

If you say this is sth you wanna merge, I'll put in more time and write actual useful tests (not this vibed slop)
Wanted to get the mood before spending more time on it.

himself65:
I think we have similar PR here? https://github.com/better-auth/better-auth/pull/4447

lknlknim:
I hope drizzle adapter will expose supportDates as well. I hit this with Payload CMS + D1 + Better Auth
Now I need to fork the adapter and set supportDates to false. Also set config.verification.disableCleanup to true to login without errors.
```
{
  query: 'insert into "user_verifications" ("id", "created_at", "updated_at", "identifier", "value", "expires_at") values (null, ?, ?, ?, ?, ?) returning "id", "created_at", "updated_at", "identifier", "value", "expires_at"',
  params: [Array],
  [cause]: [Error: D1_TYPE_ERROR: Type 'object' not supported for value 'Mon Oct 13 2025 18:40:45 GMT+0800 (Hong Kong Standard Time)'] {
    [cause]: Error: Type 'object' not supported for value 'Mon Oct 13 2025 18:40:45 GMT+0800 (Hong Kong Standard Time)'
        at <unknown> (cloudflare-internal:d1-api:285:24)
        at Array.map (<anonymous>)
        at D1PreparedStatement.bind (cloudflare-internal:d1-api:256:42)
  }
```

gustavovalverde:
Thanks for the proposal. `supportsDates` has since landed as a first-class adapter option (`packages/core/src/db/adapter/index.ts:106`, honored in `factory.ts` lines 264, 356, 544). Setting `supportsDates: false` on the adapter config gives the behavior this PR was reaching for. Closing as already implemented.

PR #4299, opened 2025-08-29
refactor: use `ms` from npm

himself65:
Related: https://github.com/vercel/ms/issues/184#issuecomment-3226506002
Fixes: https://linear.app/better-auth/issue/DEV-258/refactor-use-ms-from-upstream-npm-package

Remove duplicated code


dimitropoulos:
thanks for giving it a try! if you run into any problem at all please don't hesitate to let us know!

himself65:
> thanks for giving it a try! if you run into any problem at all please don't hesitate to let us know!

sure

Issue #4301, opened 2025-08-29
Add a "lastLoginAt" field to users


maelp:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

It's quite common in apps, or at least for the admins, to want to know the last time some user logged-in, to see which accounts are fresh or stale, etc

### Describe the solution you'd like

Add a default (or optional) lastLoginAt field

### Describe alternatives you've considered

Implement it myself with an additionalField and a hook

0xEbrahim:
I want to work on this

ping-maxwell:
You can infer the last time at which they logged in based on the most recently created session for a given user

maelp:
@ping-maxwell sure, but I guess it means a lot of joins to get the info when displaying the user, while it might be convenient to store it on the user itself?

ping-maxwell:
In most cases where a `lastLoginAt` value would be displayed in the UI, the user would already be logged in. That means the front-end likely already has the user’s information, so joins aren’t needed. Retrieving the most recent session by`userId` should only require a single database query.

Adding a new field to the user table introduces significantly more complexity, since it impacts every Better-Auth user. That would be a major breaking change for something that can already be achieved with a straightforward DB call.

For these reasons, I believe adding `lastLoginAt` is unnecessary.

maelp:
@ping-maxwell but that's what I'm saying: the feature here is **not** for the user himself to see what time he last logged in (he doesn't care)

the feature is for the **admin** to have the list of last login time for all accounts (to detect stale accounts that haven't logged in for a while, do audits, etc)

maelp:
It could also be something optional like adding a "lastLoginAtPlugin" so it doesn't change anything for existing users

ping-maxwell:
I think it makes most sense to update the admin plugin, for whichever endpoint that relates to this to include a `lastLoginAt` in the result. (Most likely the `listUsers` endpoint, but it's up for discussion)
Adding that field to the schema has a lot more hidden complexities than you might think

maelp:
Then indeed it could be donne on the admin plugin, it's a good idea. Yes, I'm sure it is complicated, that's why I wanted it to be included "by default" so we don't have to manually implement it each time we need it. I think the complication is mostly because of the variety of ways we can connect? eg either from /sign-in/email, or from oauth callbacks, etc?

himself65:
I think you can retrieve the last login date from the session database

[...]

PR #4318, opened 2025-08-30
Add lastLogin field to the user schema

0xEbrahim:
# Add `lastLoginAt` field to users

##  Resolves
fixes [#4301](https://github.com/better-auth/better-auth/issues/4301)

##  Description

This PR implements a `lastLoginAt` field for users to track when they last logged in. This is a common requirement for admin dashboards to identify active vs. inactive users.

### Problem
As described in the issue, it's quite common for apps (especially for admins) to want to know the last time a user logged in to see which accounts are fresh or stale.

### Solution
Added a default `lastLoginAt` field that automatically updates whenever a user creates a new session (logs in).

### 1. Database Schema Updates

**File: `packages/better-auth/src/db/schema.ts`**
- Added `lastLoginAt: z.date().nullish()` to the `userSchema`
- Field is nullable/optional to support existing users

**File: `packages/better-auth/src/db/get-tables.ts`**
- Added field definition in the user table configuration:
```typescript
lastLoginAt: {
    type: "date",
    required: false,
    fieldName: options.user?.fields?.lastLoginAt || "lastLoginAt",
},
```

### 2. Session Creation Logic

**File: `packages/better-auth/src/db/internal-adapter.ts`**
- Added `updateWithHooks` call in `createSession` function to update user's `lastLoginAt` timestamp
- Updates both `lastLoginAt` and `updatedAt` fields when sessions are created
- Works with both database and secondary storage configurations

### 3. API Response Updates

**File: `packages/better-auth/src/api/routes/sign-in.ts`**
- Added `lastLoginAt` field to the user object in sign-in response

**File: `packages/better-auth/src/api/routes/sign-up.ts`**
- Added `lastLoginAt` field to the user object in sign-up response (both auto-sign-in and non-auto-sign-in cases)

### 4. Tests

**File: `packages/better-auth/src/db/last-login-at.test.ts`**
- Comprehensive test suite covering:
  - Initial state (null for new users)
  - Session creation updates
  - Multiple login scenarios
  - Database persistence
  - Custom field configurations

**File: `packages/better-auth/src/db/internal-adapter.test.ts`**
- Added test to verify the field exists in schema

##  How It Works

1. **New users**: `lastLoginAt = null` initially
2. **User signs in**: `createSession()` is called → `updateWithHooks()` updates `lastLoginAt` to current timestamp
3. **Database**: Field is automatically persisted
4. **API response**: `lastLoginAt` is included in user object returned to client
5. **Subsequent logins**: `lastLoginAt` gets updated to new timestamp

##  Features

- **Automatic updates** - No manual intervention needed
- **Database persistence** - Changes are committed to database
- **Hook integration** - Works with existing database hooks
- **Configurable** - Supports custom field names and model names
- **Type safe** - Properly typed with TypeScript
- **Backward compatible** - Existing users will have `null` values

##  Testing

All tests pass:
- ✅ Initial state test: New users have `lastLoginAt = null`
- ✅ Session creation test: `lastLoginAt` gets updated when user signs in
- ✅ Multiple logins test: `lastLoginAt` gets updated on subsequent logins
- ✅ Database persistence test: Field is properly stored in database

##  Migration Notes

For existing databases, the `lastLoginAt` field will be added as nullable, so existing users will have `null` values until their next login.

##  Alternative Implementation

As mentioned in the original issue, this could have been implemented using `additionalFields` and hooks, but providing it as a default field offers better developer experience and consistency across applications.

##  Usage Example

```typescript
// After user signs in
const signInResponse = await client.signIn.email({
  email: "user@example.com",
  password: "password123",
});

// lastLoginAt will be automatically set
console.log(signInResponse.data?.user.lastLoginAt); // Date object

// For admin queries
const users = await db.findMany({
  model: "user",
  sortBy: { field: "lastLoginAt", direction: "desc" }
});
```

##  Benefits

- **Admin dashboards**: Easily identify active vs. inactive users
- **User analytics**: Track user engagement patterns
- **Account management**: Identify stale accounts for cleanup
- **Security**: Monitor suspicious login patterns
- **Zero configuration**: Works out of the box with sensible defaults

    

---

## Summary by cubic
Add lastLoginAt to the user model. It auto-updates on session creation and is returned by sign-in/sign-up APIs to track recent logins.

- **New Features**
  - user schema: nullable date field lastLoginAt
  - createSession updates lastLoginAt (and updatedAt) via updateWithHooks
  - sign-in and sign-up responses include lastLoginAt
  - Tests cover initial null, updates on login, and DB persistence

- **Migration**
  - No action needed; field is nullable and defaults to null for existing users
  - Supports custom field name via options.user.fields.lastLoginAt

### DIFF
--- a/docs/components/landing/testimonials.tsx
+++ b/docs/components/landing/testimonials.tsx
@@ -1,6 +1,8 @@
 import { cn } from "@/lib/utils";
 import { Icons } from "../icons";
 import Link from "next/link";
+import { noSSR } from "foxact/no-ssr";
+import { Suspense } from "react";
 
 const testimonials = [
 	{
@@ -179,17 +181,20 @@ holy sh** is it good, works so nice with typescript + drizzle`,
 		avatar: "/people-say/egoist.png",
 		social: <Icons.x />,
 	},
-].sort(() => Math.random() - 0.5);
+];
 
 type TestimonialProps = (typeof testimonials)[number];
 
 const TestimonialItem = ({
 	reverse = false,
 	testimonials,
+	noSsr,
 }: {
 	reverse?: boolean;
 	testimonials: TestimonialProps[];
+	noSsr?: boolean;
 }) => {
+	noSsr && noSSR();
 	const animeSeconds = testimonials.length * 10;
 	return (
 		<div className="max-w-full mx-auto">
@@ -259,12 +264,31 @@ export const Testimonial = () => {
 					}}
 					className="relative flex justify-around gap-5 overflow-hidden shrink-0"
 				>
-					<TestimonialItem
-						reverse
-						testimonials={Array(15)
-							.fill(testimonials.slice(0, Math.floor(testimonials.length / 2)))
-							.flat()}
-					/>
+					<Suspense
+						fallback={
+							<TestimonialItem
+								testimonials={Array(15)
+									.fill(
+										testimonials.slice(
+											Math.floor(testimonials.length / 2) + 1,
+											testimonials.length - 1,
+										),
+									)
+									.flat()}
+							/>
+						}
+					>
+						<TestimonialItem
+							noSsr
+							reverse
+							testimonials={Array(15)
+								.sort(() => Math.random() - 0.5)
+								.fill(
+									testimonials.slice(0, Math.floor(testimonials.length / 2)),
+								)
+								.flat()}
+						/>
+					</Suspense>
 				</div>
 				<div
 					style={{
@@ -273,16 +297,33 @@ export const Testimonial = () => {
 					}}
 					className="relative flex justify-around gap-5 overflow-hidden shrink-0"
 				>
-					<TestimonialItem
-						testimonials={Array(15)
-							.fill(
-								testimonials.slice(
-									Math.floor(testimonials.length / 2) + 1,
-									testimonials.length - 1,
-								),
-							)
-							.flat()}
-					/>
+					<Suspense
+						fallback={
+							<TestimonialItem
+								testimonials={Array(15)
+									.fill(
+										testimonials.slice(
+											Math.floor(testimonials.length / 2) + 1,
+											testimonials.length - 1,
+										),
+									)
+									.flat()}
+							/>
+						}
+					>
+						<TestimonialItem
+							noSsr
+							testimonials={Array(15)
+								.sort(() => Math.random() - 0.5)
+								.fill(
+									testimonials.slice(
+										Math.floor(testimonials.length / 2) + 1,
+										testimonials.length - 1,
+									),
+								)
+								.flat()}
+						/>
+					</Suspense>
 				</div>
 			</div>
 		</div>
--- a/docs/package.json
+++ b/docs/package.json
@@ -45,6 +45,7 @@
     "cmdk": "1.1.1",
     "date-fns": "^3.6.0",
     "embla-carousel-react": "^8.5.1",
+    "foxact": "^0.2.49",
     "fumadocs-core": "15.7.1",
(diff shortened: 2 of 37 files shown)


0xEbrahim:
OK, now that was exhausting fr

himself65 (changes requested):
Why don't we use the session schema?

0xEbrahim:
> Why don't we use the session schema?

Good question, but i added the property inside user's schema as it is related more to the user, so i can access it easily querying the user's table/object.
If you want me to change it to the session schema i don't have problems

himself65:
> > Why don't we use the session schema?
> 
> Good question, but i added the property inside user's schema as it is related more to the user, so i can access it easily querying the user's table/object. If you want me to change it to the session schema i don't have problems

I think adding to the session is more reasonable. It's possible that one user has multiple device logins. So lastLoginAt is a wider concept

0xEbrahim:
> > > Why don't we use the session schema?
> > 
> > 
> > Good question, but i added the property inside user's schema as it is related more to the user, so i can access it easily querying the user's table/object. If you want me to change it to the session schema i don't have problems
> 
> I think adding to the session is more reasonable. It's possible that one user has multiple device logins. So lastLoginAt is a wider concept

I will work on that, thanks for your feedback

maelp:
@himself65 the idea would be that in an admin dashboard we can list all users and get the last time of their logins (this feature is more useful for admins than for the users themselves), so I guess it makes more sense to have it on users?

otherwise we can store it as the last created session, but then we have to do a lot of JOINS/MATERIALIZED VIEWS if we want to get for each user the last login time no?

@0xEbrahim does this work also for magic-link and social sign-ins?

0xEbrahim:
> @himself65 the idea would be that in an admin dashboard we can list all users and get the last time of their logins (this feature is more useful for admins than for the users themselves), so I guess it makes more sense to have it on users?
> 
> otherwise we can store it as the last created session, but then we have to do a lot of JOINS/MATERIALIZED VIEWS if we want to get for each user the last login time no?
> 
> @0xEbrahim does this work also for magic-link and social sign-ins?

Due to the changes I've done based on @himself65 recommendations so yeah it should work with anything that attached with sessions as i moved the `lastLoginAt` from user to session

jslno:
> otherwise we can store it as the last created session, but then we have to do a lot of JOINS/MATERIALIZED VIEWS if we want to get for each user the last login time no?

It might not be exactly what you want, but you could query the latest login by doing something like this:

```ts
const [session] = await ctx.adapter.findMany<Session>({
	model: "session",
	where: [{
		field: "userId",
		value: ctx.session.user.id,
	}],
	sortBy: {
		field: "lastLoginAt",
		direction: "desc"
	},
	limit: 1
});
```

Storing `lastLoginAt` on the sessions table is the only correct approach, to preverse a complete history across multiple devices.
I think the cleanest way would be if this was handled internally and a `latestLoginAt` was automatically added to `session.user`.

maelp:
@jslno what I'm saying is I want it for all users at once (eg to show a big table of ALL users — **not just the current user** — with the lastLoginAt so I can see which ones are logging recently, and which haven't logged in for a while)

jslno:
@maelp Yeah, I see the issue - my bad. Then maybe instead of adding `latestLoginAt` to `session.user` just add it as a schema field and update it accordingly?

maelp:
sure, that's what I did in my own code with an `additionalField`, this issue is just to suggest that it could be a useful addition as a default / plugin in BetterAuth since that's something that many people might use

jslno:
I agree, might make more sense to implement this into the admin plugin though.

PR #4330, opened 2025-08-31
fix(bearer): certain sign-in endpoints won't give bearer token v2

ping-maxwell:
The first PR was merged by accident (I assume) by Alex, then reverted. So this is the same PR but with new changes based on Alex's change request

## What's this PR for?
Normal sign-in endpoints will return a response which includes the `set-auth-token` header from the bearer plugin, however some sign-in endpoints such as social logins or magic links where the set-cookie response comes from a separate `/api/auth` page will cause the authClient not pick up on the bearer token. This PR addresses this issue.

## How does it address it?
We can address this issue by checking if the `set-cookie` response in the hook also contains a `location` header thus meaning it will be redirected. (For example social logins where the user is redirected to /callback which will give a `set-cookie` as well as a `location` header) If we find out that it does include a location value, then we can set a cookie called `bearer-token-confirmation` which is just `true`. Then, on the client bearer plugin, on-start-up it will check if that cookie exists, and if so, hit a newly `/get-bearer-value` endpoint. This endpoint will then remove the `bearer-token-confirmation` cookie and then return the bearer token by just grabbing it from the session cookie value. Once it's returned, the authClient's `onSuccess` global hook would pickup on that response and thus the new bearer token is returned.

An example flow will look like:
image

## Breaking Changes / Migrations?

Nothing will break, however the need to add `bearerClient` plugin on the authClient will be required in order for social logins/magic link or other similar sign-ins to work.

### DIFF
--- a/packages/better-auth/src/plugins/bearer/client.ts
+++ b/packages/better-auth/src/plugins/bearer/client.ts
@@ -1,0 +1,18 @@
+import type { BetterAuthClientPlugin } from "../../types";
+
+export const bearerClient = () => {
+	return {
+		id: "bearer",
+		getActions($fetch, $store, options) {
+			if (typeof document === "undefined") return {};
+			const cookie = document.cookie;
+			if (cookie.includes("bearer-token=true")) {
+				// This will hit the endpoint which would grab the bearer token cookie if it exists, then delete said cookie
+				// It will then return the bearer token in the response which should be caught on the authClient's `onSuccess` hook
+				$fetch("/get-bearer-token");
+			}
+			return {};
+		},
+	} satisfies BetterAuthClientPlugin;
+};
+ 
--- a/docs/content/docs/plugins/bearer.mdx
+++ b/docs/content/docs/plugins/bearer.mdx
@@ -22,6 +22,15 @@ export const auth = betterAuth({
 });
 ```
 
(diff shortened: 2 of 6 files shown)


packages/better-auth/src/plugins/bearer/client.ts
```diff
@@ -0,0 +1,27 @@
...
+		id: "bearer",
+		getActions($fetch) {
+			if (typeof document === "undefined") return {};
+			const cookie = document.cookie;
+			const cookieName = options?.cookieName || "bearer-token-confirmation";
+			if (cookie.includes(`${cookieName}=true`)) {
+				// This will hit the endpoint which would grab the bearer token cookie if it exists, then delete said cookie
+				// It will then return the bearer token in the response which should be caught on the authClient's `onSuccess` hook
+				$fetch("/get-bearer-token");
+			}
```
himself65, line 23: I need to design a better way to handle the side effect. Fetch shouldn't be in the getActions function anyway.

himself65 (approved):

packages/better-auth/src/plugins/bearer/client.ts
ping-maxwell: yeah I agree

PR #4337, opened 2025-09-01
update(organization): change requests for dynamic AC

ping-maxwell:
Fixed all change requests in this PR:
https://github.com/better-auth/better-auth/pull/4087
    

---

## Summary by cubic
Simplified dynamic access control for organization roles by removing per-endpoint hooks and relying on built-in checks. Role names are now always lowercased, the default role limit is unlimited, and get-role returns additional fields like color.

- **Refactors**
  - Removed dynamicAccessControl hooks: validateRoleName, normalizeRoleName, allowCreatingRole, allowUpdatingRole, allowDeletingRole, allowListingRoles, allowGettingRole.
  - Always normalize role names to lowercase internally.
  - Set default maximumRolesPerOrganization to Infinity; updated docs and examples.
  - Reordered create/update flow: normalize early, validate permissions/resources, verify member capability, then write.
  - get-role now returns additional fields (e.g., color) with correct typing; tests updated to assert color is a string.

- **Migration**
  - Remove the above hooks from your organization dynamicAccessControl config.
  - If you relied on the old default max (25), explicitly set maximumRolesPerOrganization to your desired limit.
  - If your code expects role names in mixed case, update it to use lowercase.
  - If you read role details, expect the color field in responses.


packages/better-auth/src/plugins/organization/routes/crud-access-control.ts
```diff
@@ -130,6 +131,20 @@ export const createOrgRole = <O extends OrganizationOptions>(options: O) => {
 				});
 			}
 
+			// -----
+			// Normalize the role name.
+			roleName = defaultNormalizeRoleName({
```
Bekacru: let's rename this to `normalizeRoleName` if we don't support custom not sure if it makes having `default` there

PR #4349, opened 2025-09-01
WIP: feat(sso): add support for pre-configured SSO providers during initilization

### LINKED ISSUE #4346: [sso] Allow static OIDC providers in SSO plugin config
### Is this suited for GitHub?
- [x] Yes

#### Problem
When using custom OpenID Connect (OIDC) providers, they currently have to be registered dynamically through the API.  
This is error-prone and makes working with them more difficult than necessary, especially in server-side or infrastructure-as-code setups.

#### Proposed solution
Allow OIDC providers to be declared directly in the `SSOOptions` at initialization.  
These providers should not be stored in the database, but instead behave like built-in social providers (GitHub, Google, etc.), so they can be easily changed and version-controlled.

#### Alternatives considered
- Improving the registration experience (e.g. removing the requirement for a logged-in user, or allowing true server-side admin registration).

#### Additional context
Currently, registering providers requires a logged-in session, which makes it hard to bootstrap providers from the server side at startup. This blocks clean deployments and automation.

flowhorn:
## Description

This PR adds support for **pre-configured SSO providers during initialization**.  
Instead of always relying on `registerSSOProvider` (which requires a logged-in session and persists providers to the database), you can now declare providers directly in your `SSOOptions` when setting up Better Auth.

This makes it possible to:

- Define static OIDC/SAML providers in code (e.g. Google Workspace, internal clinic SSO).
- Keep provider configs in **version control** and environment-based configs.
- Avoid database persistence and drift — providers exist only at runtime.
- Simplify deployments by removing the need for boot-time API calls.

Closes #4346 

## Motivation

Currently, custom OIDC providers must be registered dynamically through the API, which:
- Requires an authenticated session (not practical for server-side bootstrap).
- Is error-prone and difficult to automate in CI/CD or infra-as-code.
- Makes simple static use cases harder than necessary.

This PR addresses that by introducing **configuration-based providers** alongside the existing runtime registration.

## Changes

- Added `providers` option to `SSOOptions`.
- Updated provider resolution logic to check config-based providers before database providers.
- Extended documentation with configuration examples.
- Added tests to verify that pre-configured providers are available immediately.
- Provided example usage docs comparing config-based vs runtime-based providers.

## Usage

```ts
import { sso } from "@better-auth/sso"

const auth = betterAuth({
  plugins: [
    sso({
      providers: [
        {
          providerId: "google-workspace",
          issuer: "https://accounts.google.com",
          domain: "yourcompany.com",
          oidcConfig: {
            discoveryEndpoint: "https://accounts.google.com/.well-known/openid-configuration",
            pkce: true,
            scopes: ["openid", "email", "profile"],
            mapping: {
              id: "sub",
              email: "email",
              name: "name",
              image: "picture"
            }
          }
        }
      ]
    })
  ]
})

    
---

## Summary by cubic
Adds support for pre-configured SSO providers during initialization so you can define OIDC/SAML providers in code without database entries. This simplifies setup and works alongside runtime-registered providers.

- **New Features**
  - sso({ providers: [...] }) supports OIDC and SAML, with optional domain and organizationId.
  - Provider lookups now check config first, then DB (by providerId, domain, or organization slug).
  - SSOProvider.userId is optional; config-based providers have no owner.
  - Updated docs, examples, and tests to cover config-based providers.

- **Migration**
  - If your ssoProvider.userId column is NOT NULL, make it nullable to support config-backed providers.

### DIFF
--- a/docs/content/docs/plugins/sso.mdx
+++ b/docs/content/docs/plugins/sso.mdx
@@ -71,9 +71,140 @@ This plugin is in active development and may not be suitable for production use.
     </Step>
 </Steps>
 
+## Configuration
+
+The SSO plugin can be configured with pre-defined providers during initialization, allowing you to define your SSO providers directly in your auth configuration instead of using the `registerSSOProvider` endpoint later.
+
+### Basic Configuration
+
+```ts title="auth.ts"
+import { betterAuth } from "better-auth"
+import { sso } from "@better-auth/sso"
+
+const auth = betterAuth({
+    plugins: [
+        sso({
+            providers: [
+                {
+                    providerId: "google-workspace",
+                    issuer: "https://accounts.google.com",
+                    domain: "yourcompany.com",
+                    oidcConfig: {
+                        clientId: process.env.GOOGLE_CLIENT_ID,
+                        clientSecret: process.env.GOOGLE_CLIENT_SECRET,
+                        discoveryEndpoint: "https://accounts.google.com/.well-known/openid-configuration",
+                        pkce: true,
+                        scopes: ["openid", "email", "profile"],
+                        mapping: {
+                            id: "sub",
+                            email: "email",
+                            name: "name",
+                            image: "picture"
+                        }
+                    }
+                },
+                {
+                    providerId: "company-saml",
+                    issuer: "https://idp.yourcompany.com",
+                    domain: "yourcompany.com", 
+                    samlConfig: {
+                        entryPoint: "https://idp.yourcompany.com/sso",
+                        cert: process.env.SAML_CERT,
+                        // ... other SAML configuration
+                    }
+                }
+            ]
+        })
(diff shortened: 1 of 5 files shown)


Kinfe123:
ref #3660

flowhorn:
@Kinfe123 this looks great, is there a reason why you should'nt be able to use this in production environments? I think this would help setting up custom oicd providers.

captbunzo:
@flowhorn Why was this closed? I think this would be a great feature, a better way to do things, and more consistent with how other configuration is done with better-auth.

flowhorn:
@captbunzo For me, the Generic OAuth Plugin was the solution (https://www.better-auth.com/docs/plugins/generic-oauth), I did not need the sso plugin, as this is more for dynamically setting up sso.

PR #4360, opened 2025-09-02
fix: prevent build error for node:sqlite

bytaesu:
---

This PR fixes build errors caused by importing node:sqlite.

- Wraps the import in a server-only guard (`typeof window === "undefined"`)

---

**Build error:**  
```
"next": "^15.5.2"
"better-auth": "1.3.8-beta.10"
```
build-error
import-trace

---

> **NOTE: Even in beta/development, I want build errors to be caught early.**
> 
> I have been developing to use the Kakao Provider, and just before deployment, the production build was failing unexpectedly. After investigation, it seems this issue was the cause.

### DIFF
--- a/packages/better-auth/src/adapters/kysely-adapter/dialect.ts
+++ b/packages/better-auth/src/adapters/kysely-adapter/dialect.ts
@@ -106,7 +106,7 @@ export const createKyselyAdapter = async (config: BetterAuthOptions) => {
 		});
 	}
 
-	if ("createSession" in db) {
+	if ("createSession" in db && typeof window === "undefined") {
 		let DatabaseSync: typeof import("node:sqlite").DatabaseSync | undefined =
 			undefined;
 		try {


bytaesu:
@himself65, thank you for checking so quickly.

If it’s a Next.js issue, it might get complicated… Actually, there’s a problem with cookie serialization in Next.js, and I’ve fixed the bug and submitted a PR, but no one has reviewed it yet🥲

himself65:
> @himself65, thank you for checking so quickly.
> 
> If it’s a Next.js issue, it might get complicated… Actually, there’s a problem with cookie serialization in Next.js, and I’ve fixed the bug and submitted a PR, but no one has reviewed it yet🥲

which PR? let me see

bytaesu:
@himself65,

"@edge-runtime/cookie": https://github.com/vercel/edge-runtime/pull/1128

"next": https://github.com/vercel/next.js/pull/81873

The cookie serialization logic in Next.js is inconsistent. Because of this, the sameSite boolean does not work as described in the official docs.

I think this could cause difficult debugging later. It’s minor, but I studied using the Next.js source code before, so it stands out to me.

PR #4361, opened 2025-09-02
fix: prevent undefined from passed to adapter in username plugin

### LINKED ISSUE #4341: Prisma `strictUndefinedChecks` not working with `username()` plugin

### To Reproduce

I am hitting an issue when Prisma's `strictUndefinedChecks` is enabled. I see in the debug logs that `image` is being filtered out, but not `username` and `displayUsername`.

#### Simple Curl
```
$ curl -X POST http://localhost:5002/api/auth/sign-up/email \
    -H "Content-Type: application/json" \
    -d '{
      "email": "test7@example.com",
      "password": "SecurePassword123!",
      "name": "Test User"
    }'
{"code":"FAILED_TO_CREATE_USER","message":"Failed to create user","details":{"name":"PrismaClientValidationError","clientVersion":"6.4.1"}}%
```

#### Debug logs showing issue
```
{"level":30,"time":1756715373627,"pid":39763,"hostname":"qm.home","msg":"Server listening at http://192.168.97.0:5002"}
{"level":30,"time":1756715381176,"pid":39763,"hostname":"qm.home","reqId":"req-1","req":{"method":"POST","url":"/api/auth/sign-up/email","host":"localhost:5002","remoteAddr
2025-09-01T08:29:41.189Z INFO [Better Auth]: [Prisma Adapter] #0 [1/3] findOne: {
  model: "User",
  where: [
    {
      operator: "eq",
      connector: "AND",
      field: "email",
      value: "test7@example.com",
    }
  ],
  select: undefined,
}
2025-09-01T08:29:41.212Z INFO [Better Auth]: [Prisma Adapter] #0 [2/3] findOne (DB Result): {
  model: "User",
  data: null,
}
2025-09-01T08:29:41.212Z INFO [Better Auth]: [Prisma Adapter] #0 [3/3] findOne (Parsed Result): {
  model: "User",
  data: null,
}
2025-09-01T08:29:41.329Z INFO [Better Auth]: [Prisma Adapter] #1 [1/4] create (Unsafe Input): {
  model: "User",
  data: {
    createdAt: 2025-09-01T08:29:41.329Z,
    updatedAt: 2025-09-01T08:29:41.329Z,
    emailVerified: false,
    email: "test7@example.com",
    name: "Test User",
    image: undefined,
    username: undefined,
    displayUsername: undefined,
  },
}
2025-09-01T08:29:41.330Z INFO [Better Auth]: [Prisma Adapter] #1 [2/4] create (Parsed Input): {
  model: "User",
  data: {
    name: "Test User",
    email: "test7@example.com",
    emailVerified: false,
    createdAt: 2025-09-01T08:29:41.329Z,
    updatedAt: 2025-09-01T08:29:41.329Z,
    username: undefined,
    displayUsername: undefined,
  },
}
{"level":30,"time":1756715381338,"pid":39763,"hostname":"qm.home","reqId":"req-1","res":{"statusCode":422},"responseTime":161.23312499999884,"msg":"request completed"}
```

I can't really remove `strictUndefinedChecks` setting on Prisma, because the generated client then no longer has `Prisma.skip` which would require a much larger refactor throughout our codebase, and we like this prisma feature, so not interested.

# Workaround

I was able to get things working with the following "hack" that others may find useful. Since I'm creating a fresh prisma client just for the better-auth integration, I can extend that client to manually filter out the `undefined` values and replace them with `skip`. It could be cleaner I'm sure, but I think this should be fine for now. Maybe it helps someone else as well..

```typescript
import { PrismaClient, Prisma } from "@prisma/client";
import { username } from "better-auth/plugins";

const _prisma = new PrismaClient();


function skipUndefinedValues<D extends Record<string, unknown>>(data: D): D {
  return Object.entries(data).reduce(
    (acc, [key, value]) => {
      acc[key as keyof typeof data] = Prisma.skip.ifUndefined(value) as any;
      return acc;
    },
    {} as typeof data,
  );
}

const isPrismaSkip = (val: unknown): val is typeof Prisma.skip => {
  // Prisma.skip appears to be a singleton in the source code. 
  // However, this is relying on internal implementation which could change. 
  // The `isSkip` function is not exposed for our use, unfortunately.
  return Prisma.skip === val;
};

const prismaClientWithUndefinedSkipFix = _prisma.$extends({
  query: {
    $allModels: {
      async create({ args, query }) {
        if (args.data) {
          if (!isPrismaSkip(args.data)) {
            args.data = skipUndefinedValues(args.data);
          }
        }
        return query(args);
      },
      async update({ args, query }) {
        if (args.data) {
          if (!isPrismaSkip(args.data)) {
            args.data = skipUndefinedValues(args.data);
          }
        }
        return query(args);
      },
      async upsert({ args, query }) {
        if (args.create) {
          if (!isPrismaSkip(args.create)) {
            args.create = skipUndefinedValues(args.create);
          }
        }
        if (args.update) {
          if (!isPrismaSkip(args.update)) {
            args.update = skipUndefinedValues(args.update);
          }
        }
        return query(args);
      },
    },
  },
});

  database: prismaAdapter(prismaClientWithUndefinedSkipFix, {
    provider: "postgresql",
    debugLogs: true,
  }),
...
});

```

1.3.7

### System info

```bash
System:
    OS: macOS 15.6.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 117.73 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
```

Backend

```typescript
    provider: "postgresql",
    debugLogs: true,
  }),
  emailAndPassword: {
    enabled: true,
  },
  user: {
    modelName: "User",
  },
  plugins: [username()],
  advanced: {
    database: {
      useNumberId: true,
    },
  },
});
```

Kinfe123:
closes #4341
before 
Screenshot 2025-09-02 at 2 13 29 PM
after
Screenshot 2025-09-02 at 2 13 36 PM

### DIFF
--- a/packages/better-auth/src/adapters/create-adapter/index.ts
+++ b/packages/better-auth/src/adapters/create-adapter/index.ts
@@ -409,7 +409,9 @@ export const createAdapter =
 					});
 				}
 
-				transformedData[newFieldName] = newValue;
+				if (newValue !== undefined) {
+					transformedData[newFieldName] = newValue;
+				}
 			}
 			return transformedData;
 		};
--- a/packages/better-auth/src/plugins/username/index.ts
+++ b/packages/better-auth/src/plugins/username/index.ts
@@ -488,8 +488,12 @@ export const username = (options?: UsernameOptions) => {
 						);
 					},
 					handler: createAuthMiddleware(async (ctx) => {
-						ctx.body.displayUsername ||= ctx.body.username;
-						ctx.body.username ||= ctx.body.displayUsername;
+						if (ctx.body.username && !ctx.body.displayUsername) {
+							ctx.body.displayUsername = ctx.body.username;
+						}
+						if (ctx.body.displayUsername && !ctx.body.username) {
+							ctx.body.username = ctx.body.displayUsername;
+						}
 					}),
 				},
 			],


himself65 (approved):
LGTM, but I think we can improve the type to avoid this case in the future

Kinfe123:
Yeah. i will check on that on the follow-up PR.

PR #4363, opened 2025-09-02
feat(org): get organization

frectonz:
## Done

- Implement `getOrganization`
- Document `getOrganization`
- Add tests for it

## Summary by cubic
Add a lightweight GET /organization/get endpoint and SDK methods to fetch basic organization info by ID or slug for a signed-in member. This complements getFullOrganization with a faster, minimal payload.

- **New Features**
  - Endpoint: GET /organization/get
  - Query: organizationId or organizationSlug (one is required; if neither, returns 200 null)
  - Auth: requires headers; membership enforced via orgMiddleware and orgSessionMiddleware
  - Response: Organization object only (no members, invitations, or teams)
  - Errors: 400 if not found; 403 if user is not a member
  - SDK: auth.api.getOrganization and authClient.organization.getOrganization
  - Tests: added coverage for fetching an org and validating name/slug

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-org.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-org.ts
@@ -638,7 +638,9 @@ export const getOrganization = <O extends OrganizationOptions>(options: O) =>
 					query.organizationSlug,
 				);
 			} else {
-				return ctx.json(null, { status: 200 });
+				throw new APIError("FORBIDDEN", {
+					message: ORGANIZATION_ERROR_CODES.NO_ACTIVE_ORGANIZATION,
+				});
 			}
 
 			if (!organization) {


ping-maxwell:
Don't forget to check org by session.activeOrganizationId too!

frectonz:
oh yeah will add that

packages/better-auth/src/plugins/organization/routes/crud-org.ts
```diff
@@ -585,6 +585,84 @@ export const deleteOrganization = <O extends OrganizationOptions>(
...
+			} else if ("organizationSlug" in query && query.organizationSlug) {
+				organization = await adapter.findOrganizationBySlug(
+					query.organizationSlug,
+				);
+			} else {
+				return ctx.json(null, { status: 200 });
+			}
+
+			if (!organization) {
+				throw new APIError("BAD_REQUEST", {
```
ping-maxwell, line 647: Maybe `NOT_FOUND` makes more sense, rather than `BAD_REQUEST`? Wdyt?

packages/better-auth/src/plugins/organization/routes/crud-org.ts
```diff
@@ -585,6 +585,84 @@ export const deleteOrganization = <O extends OrganizationOptions>(
...
+
+			let organization: InferOrganization<O> | null = null;
+			if ("organizationId" in query && query.organizationId) {
+				organization = await adapter.findOrganizationById(query.organizationId);
+			} else if ("organizationSlug" in query && query.organizationSlug) {
+				organization = await adapter.findOrganizationBySlug(
+					query.organizationSlug,
+				);
+			} else {
+				return ctx.json(null, { status: 200 });
```
ping-maxwell: I think that it makes sense for get-organization to always return an org, either that or it's an error.
Maybe in this case we should throw a `BAD_REQUEST` saying that _a_ parameter is required in order to get an org in the first place?

ping-maxwell (changes requested):

packages/better-auth/src/plugins/organization/routes/crud-org.ts
frectonz: yeah this makes sense

packages/better-auth/src/plugins/organization/routes/crud-org.ts
frectonz: I used the `BAD_REQUEST` status code because it is the status code that is always used along the `ORGANIZATION_NOT_FOUND` error in the `organization` plugin.

Healios:
Just wanted to chime in and say this would be a great addition. A lightweight getOrganization would be very useful in real-world usage.

I’m not the most well-versed in reading PRs - is there anything currently blocking this from being merged?

PR #4373, opened 2025-09-02
feat(organization): allow client side requests to add-member

alliefitter:
This PR is intended to open a discussion about making `/admin/add-member` allow client side requests. I wasn't able to find anything in closed PRs that would explain why this is a server only endpoint, so I hoping to discuss allowing client side requests. Thanks!

---
## Summary by cubic
Enable client-side requests to `/admin/add-member` by removing the server-only restriction. Updated tests to use `client.organization.addMember`, confirming front-end apps can add members without a server proxy; org session middleware is not required.

<sup>Written for commit fd79cfb16d29aea2be5b440da4bb67954b77b15e. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-members.test.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-members.test.ts
@@ -383,12 +383,10 @@ describe("updateMemberRole", async () => {
 				}),
 			},
 		);
-		await auth.api.addMember({
-			body: {
-				organizationId: newOrg.data?.id as string,
-				userId: user.id,
-				role: "admin",
-			},
+		await client.organization.addMember({
+			organizationId: newOrg.data?.id as string,
+			userId: user.id,
+			role: "admin",
 		});
 		const updatedMember = await client.organization.updateMemberRole(
 			{
--- a/packages/better-auth/src/plugins/organization/routes/crud-members.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-members.ts
@@ -54,7 +54,7 @@ export const addMember = <O extends OrganizationOptions>(option: O) => {
 				...baseMemberSchema.shape,
 				...additionalFieldsSchema.shape,
 			}),
-			use: [orgMiddleware, orgSessionMiddleware],
+			use: [orgMiddleware],
 			metadata: {
 				$Infer: {
 					body: {} as {
--- a/packages/better-auth/src/plugins/organization/routes/crud-members.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-members.ts
@@ -54,7 +54,7 @@ export const addMember = <O extends OrganizationOptions>(option: O) => {
 				...baseMemberSchema.shape,
 				...additionalFieldsSchema.shape,
 			}),
-			use: [orgMiddleware],
+			use: [orgMiddleware, orgSessionMiddleware],
 			metadata: {
 				$Infer: {
 					body: {} as {


himself65 (commented):
can you provide more use case?

alliefitter:
@himself65 Sure, we're building an auth service with a dashboard for managing users, orgs, OIDC creds, etc... One of the requirements we have is that our customer success people should be able to set up and fix problems in customer organizations. All the other endpoints around org membership allow client side requests, and we had no problem using them to build out the user management portion of the dashboard, except for adding a member to an org. We can, of course, send an invitation email to add users, but we'd prefer to be able to set up all users a customer needs, with their appropriate roles ahead of time. We have a plugin that we're adding workarounds and one-of functionality to, and we have the below workaround in that plugin to allow client requests. However, if there isn't a great reason for making this endpoint server only, we prefer not maintain a workaround like this, even if it is simple.

```
export const addToOrg = createAuthEndpoint(
  "/dust/add-to-org",
  {
    method: "POST",
    use: [dustMiddleware],
    body: z.object({
      userId: z.coerce.string().meta({
        description:
          'The user Id which represents the user to be added as a member. If `null` is provided, then it\'s expected to provide session headers. Eg: "user-id"',
      }),
      organizationId: z.string().meta({
        description: "The organization the user should be added to.",
      }),
      role: z.string().default("member").optional().meta({
        description: "The role the user should have within the org.",
      }),
    }),
    metadata: {
      openapi: {
        description: "Remove a member from an organization",
        responses: {
          "200": {
            description: "Success",
            content: {
              "application/json": {
                schema: {
                  $ref: "#/components/schemas/User",
                },
              },
            },
          },
        },
      },
    },
  },
  async (ctx): Membership => {
    return auth.api.addMember({
      body: ctx.body,
    });
  },
```

himself65:
> @himself65 Sure, we're building an auth service with a dashboard for managing users, orgs, OIDC creds, etc... One of the requirements we have is that our customer success people should be able to set up and fix problems in customer organizations. All the other endpoints around org membership allow client side requests, and we had no problem using them to build out the user management portion of the dashboard, except for adding a member to an org. We can, of course, send an invitation email to add users, but we'd prefer to be able to set up all users a customer needs, with their appropriate roles ahead of time. We have a plugin that we're adding workarounds and one-of functionality to, and we have the below workaround in that plugin to allow client requests. However, if there isn't a great reason for making this endpoint server only, we prefer not maintain a workaround like this, even if it is simple.
> 
> ```
> export const addToOrg = createAuthEndpoint(
>   "/dust/add-to-org",
>   {
>     method: "POST",
>     use: [dustMiddleware],
>     body: z.object({
>       userId: z.coerce.string().meta({
>         description:
>           'The user Id which represents the user to be added as a member. If `null` is provided, then it\'s expected to provide session headers. Eg: "user-id"',
>       }),
>       organizationId: z.string().meta({
>         description: "The organization the user should be added to.",
>       }),
>       role: z.string().default("member").optional().meta({
>         description: "The role the user should have within the org.",
>       }),
>     }),
>     metadata: {
>       openapi: {
>         description: "Remove a member from an organization",
>         responses: {
>           "200": {
>             description: "Success",
>             content: {
>               "application/json": {
>                 schema: {
>                   $ref: "#/components/schemas/User",
>                 },
>               },
>             },
>           },
>         },
>       },
>     },
>   },
>   async (ctx): Membership => {
>     return auth.api.addMember({
>       body: ctx.body,
>     });
>   },
> ```

That makes sense. I'm good with this PR. just need double check the permission when expose to client side

himself65 (changes requested):
You will need to add `orgSessionMiddleware` here

himself65 (approved):
LGTM, thanks

himself65 (changes requested):
OK. I realized that (actually double check from Bekacru) there's still have security risk. and adding `orgSessionMiddleware` may break the server side... I think it's not easy just to flip the SERVER_ONLY now.

P.S. you can see test is failling

himself65:
> OK. I realized that (actually double check from Bekacru) there's still have security risk. and adding `orgSessionMiddleware` may break the server side... I think it's not easy just to flip the SERVER_ONLY now.

for me: I need to add some real test cases on e2e to make sure security is correct

alliefitter:
Is there any work you'd like me to do on this or is it on hold for your security analysis?

alliefitter:
@himself65 Removed `orgSessionMiddleware`, to fix the failing tests. I'm likely missing some context, but from the implementation of `addMember`, it doesn't seem necessary. The session is only used to get the active orgId as a fall back if an orgId isn't provided. 

```
const session = ctx.body.userId
	? await getSessionFromCtx<{
			session: {
				activeOrganizationId?: string | undefined;
			};
		}>(ctx).catch((e) => null)
	: null;
const orgId =
	ctx.body.organizationId || session?.session.activeOrganizationId;
```

Bekacru:
You can't just add member from the client to any org. That means anyone can join any org as they like

alliefitter:
@Bekacru That makes sense. This feature is in the same vein as #4390, so perhaps this should be on hold until when/if that gets merged. Then we could limit this to super admins and members with appropriate permissions. Btw, I was just looking at the code, and there doesn't currently seem to be any permission check in `addMember` as there is in `removeMember` and `updateMemberRole`. I'm assuming it being server only eschews the need, but I wanted to note it, as it will need to be added if my suggestion sounds good. Maybe, instead of waiting on #4390, we should just add permissions checks now and add a super admin check to that PR after merging this one. Let me know what you think. Thanks!

Bekacru:
addMember is only server side endpoint. It can't be triggered from the client

alliefitter:
@Bekacru I understand that's the current implementation. Is there no room for discussion about changing that? It seems odd to me that add member is server only when there are no other server only endpoints in the org plugin. I'm willing to do whatever development would need to be done in order to make this available on the client side, like the rest of the org plugin endpoints.

ping-maxwell:
closing as stale

PR #4390, opened 2025-09-03
feat(organization): Make membership optional

alliefitter:
I'm hoping to start discussion around allowing non-members to add, remove, change roles, etc... of org members. This PR currently has a very naive implementation, where the membership check is simply ignored based on a plugin option. However, I believe a better solution would be to tie this feature to the admin plugin and add an option to the organization plugin that allows users with the global/super admin role to edit org membership. But before I started real work on that feature, I wanted to get feedback from y'all on whether this feature is even desirable or if there's a better way of accomplishing it. Thanks!

### DIFF
--- a/packages/better-auth/src/plugins/organization/organization.ts
+++ b/packages/better-auth/src/plugins/organization/organization.ts
@@ -965,7 +965,7 @@ export const organization = <O extends OrganizationOptions>(options?: O) => {
 						userId: ctx.context.session.user.id,
 						organizationId: activeOrganizationId,
 					});
-					if (!member) {
+					if (!member && (options?.requireMembership ?? true)) {
 						throw new APIError("UNAUTHORIZED", {
 							message:
 								ORGANIZATION_ERROR_CODES.USER_IS_NOT_A_MEMBER_OF_THE_ORGANIZATION,
--- a/packages/better-auth/src/plugins/organization/types.ts
+++ b/packages/better-auth/src/plugins/organization/types.ts
@@ -52,6 +52,12 @@ export interface OrganizationOptions {
 	 */
 	ac?: AccessControl;
 	/**
+	 * Require membership in an organization to perform CRUD operations.
+	 *
+	 * @default true
+	 */
+	requireMembership?: boolean;
+	/**
 	 * Custom permissions for roles.
 	 */
 	roles?: {


packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -965,7 +965,7 @@ export const organization = <O extends OrganizationOptions>(options?: O) => {
 						userId: ctx.context.session.user.id,
 						organizationId: activeOrganizationId,
 					});
-					if (!member) {
+					if (!member && (options?.requireMembership ?? true)) {
```
TheUntraceable: I think of instead of simply letting anyone add users to any org (possibly themselves which would be a security issue i believe), check if the option is enabled, and if they do, check if the adder's role is set to admin (as per the admin plugin functionality), and allow users to specify other roles which can override this membership check)

alliefitter: Cool. That's exactly what I was thinking of for a real solution to this. This change is more of a stub to start the discussion more than anything. So, order of operations will be:
1. Check if the option is enabled. If not fail.
2. Get the admin plugin, if enabled. If not enabled, fail.
3. Check the current user's role against the admin plugin's `adminRoles` option.
4. If the user has one of the configured `adminRoles`, allow the operation to proceed.

alliefitter:
@TheUntraceable I'd like to get some feedback on my approach before getting too far along. So far, I've:
- Added `allowSuperAdmin` option.
- Added `admin.ts` which contains `isSuperAdmin()` function which is used to determine if the admin plugin is enabled, if `allowSuperAdmin` is enabled, and whether the current user has an admin role.
- Added `isSuperAdmin()` checks to most membership checks in `routes/crud-org.ts`. Some routes, like `/organization/set-active` were skipped, as they did not seem as though a super admin would need to call them.
- Added `/organization/list-all` endpoint to list all orgs, to get around `/organization/list` being filtered by membership. Maybe modify the existing list endpoint instead?
- Added tests in `admin.test.ts`

packages/better-auth/src/plugins/organization/types.ts
```diff
@@ -447,7 +452,7 @@ export interface OrganizationOptions {
 				[key: string]: any;
 			};
 			user: User & Record<string, any>;
-			member: Member & Record<string, any>;
+			member?: Member & Record<string, any>;
```
alliefitter: `member` needs to become optional in these hooks, but I was wondering if we'd want to pass in whether the user performing the update is a super admin.

packages/better-auth/src/plugins/organization/admin.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+describe("super admin", async (it) => {
+	const { customFetchImpl, signInWithTestUser, db } = await getTestInstance({
+		plugins: [organization({ allowSuperAdmin: true }), admin()],
+	});
+	const client = createAuthClient({
+		plugins: [organizationClient(), adminClient()],
+		baseURL: "http://localhost:3000/api/auth",
+		fetchOptions: {
+			customFetchImpl,
```
TheUntraceable, line 17: Shouldn't you remove the /api/auth?

packages/better-auth/src/plugins/organization/admin.test.ts
```diff
@@ -0,0 +1,88 @@
...
+		expect(data?.name).toBe("testAgain");
+	});
+	it("should be allowed to get a full org", async () => {
+		const { data, error } = await client.organization.getFullOrganization({
+			query: {
+				organizationId: orgId,
+			},
+			fetchOptions: {
+				headers,
+			},
```
TheUntraceable, line 62: Not sure if this pattern is expected in tests but why pass headers manually?

packages/better-auth/src/plugins/organization/admin.ts
```diff
@@ -0,0 +1,24 @@
+import type { GenericEndpointContext } from "../../types";
+import { admin } from "../admin";
+import type { OrganizationOptions } from "./types";
+
+const getAdminPlugin = (ctx: GenericEndpointContext) => {
+	return ctx.context.options.plugins?.find(
+		(plugin) => plugin.id === "admin",
+	) as ReturnType<typeof admin>;
+};
```
TheUntraceable, line 12: For this I would throw an error if not found

packages/better-auth/src/plugins/organization/routes/crud-org.ts
```diff
@@ -232,7 +232,7 @@ export const createOrganization = <O extends OrganizationOptions>(
 					teamId: defaultTeam.id,
 					userId: user.id,
 				});
-			} else {
+			} else if (!isSuperAdmin(options, ctx)) {
```
TheUntraceable: Can't really understand what this is doing. Might be because I'm on mobile. Can you explain please?

packages/better-auth/src/plugins/organization/admin.test.ts
alliefitter: This is copy pasta from [other tests in the org plugin](https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/plugins/organization/organization.test.ts#L47).

packages/better-auth/src/plugins/organization/admin.test.ts
alliefitter: This is [also copy pasta](https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/plugins/organization/organization.test.ts#L67). From what I've been able to glean from the existing tests, this is to ensure that the test users's session cookie is included in the headers.

packages/better-auth/src/plugins/organization/routes/crud-org.ts
alliefitter: The existing code adds the calling user as the owner of the newly created org. In the case of a super admin creating an org, it doesn't seem desirable that the super admin would implicitly be added as the owner.  The example use case we have in mind is a customer success person creating an org for a customer before onboarding them. So, we wouldn't want the customer success person to be the owner of the org--or be a member of it, for that matter. However, I'm not sure how that would affect other use cases, so I'm not married to it. If there are problems that would come from allowing orgs to be created without any users, we can revert this change.

packages/better-auth/src/plugins/organization/admin.ts
alliefitter: I started to work on this change, and I think changing the function to throw an error will make using it a bit more janky. For reference, this is adapted from how [the `oidc-provider` gets the `jwt` plugin](https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/plugins/oidc-provider/index.ts#L33). Here's a diff of how the `isSuperAdmin()` function would change. I think having the single check that all prerequisites are satisfied and returning `false` if any of them aren't is easier to follow than having one prerequisite check different from the other two.
```
diff --git a/packages/better-auth/src/plugins/organization/admin.ts b/packages/better-auth/src/plugins/organization/admin.ts
index 8719cd4e..b090049e 100644
--- a/packages/better-auth/src/plugins/organization/admin.ts
+++ b/packages/better-auth/src/plugins/organization/admin.ts
@@ -13,9 +13,14 @@ export const isSuperAdmin = <O extends OrganizationOptions>(
        ctx: GenericEndpointContext,
 ): boolean => {
        const user = ctx.context.session?.user;
-       const adminPlugin = getAdminPlugin(ctx);
+       let adminPlugin;
+       try {
+               adminPlugin = getAdminPlugin(ctx);
+       } catch (e) {
+               return false;
+       }
        let adminRoles = adminPlugin?.options?.adminRoles ?? ["admin"];
-       if (!options?.allowSuperAdmin || !adminPlugin || !user?.role) {
+       if (!options?.allowSuperAdmin || !user?.role) {
                return false;
        }
        adminRoles = Array.isArray(adminRoles) ? adminRoles : [adminRoles]
```

alliefitter:
@TheUntraceable replied to your comments. I don't feel incredibly strongly about any of my replies, so if you'd like to just dismiss them and have me change anything, just let me know. I'm gonna be on a stay-cation starting Thursday, so I'll probably start on adding `isSuperAdmin` to all the remaining `crud-*.ts` files under `plugins/organization` then. Would y'all like me to implement everything in this PR or create smaller PRs for each `crud-*.ts` file being changed? Either works for me.

packages/better-auth/src/plugins/organization/organization.ts
```diff
@@ -166,6 +168,22 @@ export const organization = <O extends OrganizationOptions>(options?: O) => {
 		 * @see [Read our docs to learn more.](https://better-auth.com/docs/plugins/organization#api-method-organization-get-full-organization)
 		 */
 		getFullOrganization: getFullOrganization(options as O),
+		/**
+		 * ### Endpoint
+		 *
+		 * GET `/organization/listAll`
```
himself65: use list-all? I forget we are camcel case or snake case for paths

alliefitter: Yup, the endpoint is actually `/organization/list-all`. Will update the comment.

escwxyz:
Hi, is there any update on this? I see only some type errors in the CI.

alliefitter:
Apologies. I'm writing this PR for work and I got re-prioritized. I back working on this now.

alliefitter:
@escwxyz I was finally able to get back to this PR. Got the latest from canary merged in and fixed the problems in CI. Let me know if there's anything y'all need from me.

PR #4394, opened 2025-09-03
feat(mcp): add protected-resource metadata endpoint

frectonz:


Kinfe123 (approved):

packages/better-auth/src/plugins/mcp/index.ts
```diff
@@ -83,6 +83,33 @@ export const getMCPProviderMetadata = (
 	};
 };
 
+export const getMCPProtectedResourceMetadata = (
+	ctx: GenericEndpointContext,
+	options?: OIDCOptions,
+) => {
+	const baseURL = ctx.context.baseURL;
```
himself65, line 90: I think this should always be set. maybe I need update type

himself65 (approved):

packages/better-auth/src/plugins/mcp/index.ts
frectonz: ok will remove the runtime check

docs/content/docs/plugins/mcp.mdx
```diff
@@ -66,6 +66,17 @@ import { auth } from "../../../lib/auth";
 export const GET = oAuthDiscoveryMetadata(auth);
 ```
 
+### OAuth Protected Resource Metadata
+
+Add a route to expose protected resource metadata for MCP clients:
+
+```ts title=".well-known/oauth-authorization-server/route.ts"
+import { oAuthProtectedResourceMetadata } from "better-auth/plugins";
+import { auth } from "../../../lib/auth";
```
himself65:
```suggestion
import { auth } from "@/lib/auth";
```

PR #4402, opened 2025-09-03
feat(logger): option to disable colors

martiinii:
This PR adds option to disable colors from default logger implementation.
I've made this change because convex doesn't support colors in logs (using https://convex-better-auth.netlify.app/)
Default logger implementations logs look like this:
image
image

In this environment colors are not needed nor supported in convex dashboard.

No breaking changes introduces, only optional `disableColors` property for logger.
I have found no issues or discussions for this setting

### DIFF
--- a/packages/better-auth/src/utils/logger.ts
+++ b/packages/better-auth/src/utils/logger.ts
@@ -11,6 +11,7 @@ export function shouldPublishLog(
 
 export interface Logger {
 	disabled?: boolean;
+	disableColors?: boolean;
 	level?: Exclude<LogLevel, "success">;
 	log?: (
 		level: Exclude<LogLevel, "success">,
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/utils/logger.ts
```diff
@@ -83,6 +93,12 @@ export const createLogger = (options?: Logger): InternalLogger => {
 	const enabled = options?.disabled !== true;
 	const logLevel = options?.level ?? "error";
 
+	const isDisableColorsSpecified = options?.disableColors !== undefined;
+	const terminalSupportsColor = process.stdout.getColorDepth() !== 1;
```
himself65: Please don't read `process` directly. It's not supported in Cloudflare and Vercel Edge runtime.

Please just copy the code I just gave u, and note that we use `getEnvVar` function in our utils

himself65 (changes requested):

packages/better-auth/src/utils/logger.ts
martiinii: Sorry, I now see failed tests, fixing!

packages/better-auth/src/utils/color-depth.ts
```diff
@@ -0,0 +1,187 @@
...
+		// See https://invisible-island.net/ncurses/terminfo.ti.html#toc-_Specials
+		getEnvVar("TERM") === "dumb"
+	) {
+		return COLORS_2;
+	}
+
+	if (typeof process !== "undefined" && process.platform === "win32") {
+		// Lazy load for startup performance.
+		if (OSRelease === undefined) {
+			const { release } = require("os") as typeof import("os");
```
martiinii: This is the only part I'm not sure I've done correctly, I've checked repo code for `process.platform` and found `detectSystemInfo` fn but it's async

himself65: let's prevent import `os`. Use [`globalThis.navigator.platform`](https://nodejs.org/docs/latest/api/globals.html#navigatorplatform) instead

himself65:
Or maybe we just forget about this now. We don't consider people who user earlier than windows 10

packages/better-auth/src/utils/color-depth.ts
martiinii: Reading through MDN it's not supported by Deno (https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform#browser_compatibility)

And also it gives different result than `release` from `os` (can't test on Window right now, I assume it would be similar difference):
image

himself65: https://github.com/better-auth/better-auth/pull/4402#issuecomment-3251134246

martiinii:
Yeah, checked those win versions, it has support since 2015, replacing

himself65 (approved):

martiinii:
Thank you @himself65 for support and guiding me and sorry for clumsy first PR. Next time I'll do better :saluting_face:

Issue #4408, opened 2025-09-03
Bug with admin plugin and renamed fields


maelp:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

I'm using the admin plugin, which adds a "role" field, I wanted to rename it "better_auth_role", so in the BetterAuth conf I did eg

  user: {
    modelName: "usersTable",
    fields: {
      emailVerified: "email_verified",
      createdAt: "created_at",
      updatedAt: "updated_at",
      role: "better_auth_role", // admin plugin
      banReason: "ban_reason", // admin plugin
      banExpires: "ban_expires", // admin plugin
    },
...
but when trying to create a user with auth.api.createUser it complains
The field "role" does not exist in the "usersTable" schema. Please update your drizzle schema or re-generate using "npx @better-auth/cli generate
so it correctly saw that I renamed my table to usersTable (also it works for the other fields like emailVerified, etc) but it seems the admin plugin fields are not translated?

### Current vs. Expected behavior

It should use the renamed fields from the config

### What version of Better Auth are you using?

latest

### System info

```bash
latest
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});
```

maelp:
the error seems to come from here https://github.com/better-auth/better-auth/blob/8466bfe2e499f4e594c272ee42b4d6d8bd424147/packages/better-auth/src/adapters/drizzle-adapter/drizzle-adapter.ts#L245

ping-maxwell:
@maelp I'm going to close this as I assume we solved your issue in the Discord server

maelp:
yes all good! perhaps you should add to the doc in either admin plugin, or the general doc mentioning the global schema, that the plugins have to use their own "local schemas" rather than the global one

PR #4426, opened 2025-09-04
fix(admin): admin plugin update renamed fields

LaurinBoettcher:
closes: [4408](https://github.com/better-auth/better-auth/issues/4408)
    

---

## Summary by cubic
Fixes admin plugin support for renamed user fields by mapping field aliases in getAuthTables. Ensures admin works with custom column names and keeps banned/banExpires fields available.

- **Bug Fixes**
  - Map options.user.fields aliases onto user field configs when building tables.
  - Add tests covering custom email, role, and banReason mappings and presence of banned/banExpires.

### DIFF
--- a/packages/better-auth/src/db/get-tables.ts
+++ b/packages/better-auth/src/db/get-tables.ts
@@ -168,10 +168,15 @@ export const getAuthTables = (
 					fieldName: options.user?.fields?.updatedAt || "updatedAt",
 				},
 
-		...(user?.fields? (() => {
+				...(user?.fields
+					? (() => {
 							const result: Record<string, FieldAttribute> = {};
-							const aliasMap = options.user?.fields as Record<string, string> | undefined;
-							for (const [fieldName, fieldConfig] of Object.entries(user.fields)) {
+							const aliasMap = options.user?.fields as
+								| Record<string, string>
+								| undefined;
+							for (const [fieldName, fieldConfig] of Object.entries(
+								user.fields,
+							)) {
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/db/get-tables.ts
```diff
@@ -167,7 +167,24 @@ export const getAuthTables = (
 					required: true,
 					fieldName: options.user?.fields?.updatedAt || "updatedAt",
 				},
-				...user?.fields,
+
+				...(user?.fields
+					? (() => {
+							const result: Record<string, FieldAttribute> = {};
+							const aliasMap = options.user?.fields as
```
himself65, line 174: no `as`

packages/better-auth/src/db/get-tables.ts
```diff
@@ -167,7 +167,24 @@ export const getAuthTables = (
...
+								| undefined;
+							for (const [fieldName, fieldConfig] of Object.entries(
+								user.fields,
+							)) {
+								const mapped = aliasMap?.[fieldName];
+								result[fieldName] = mapped
+									? { ...fieldConfig, fieldName: mapped }
+									: fieldConfig;
+							}
+							return result;
```
himself65, line 185: can we abstract this into a single function instead of IIFE

ping-maxwell:
@LaurinBoettcher when users use the admin plugin, it's expected for the schema changes to be configured right into the admin plugin's option's `schema` object, not directly in the global auth config.

PR #4431, opened 2025-09-04
test: add tests for `sendOnSignIn`

frectonz:
## Summary by cubic
Add tests to verify email verification behavior on sign-in based on emailVerification.sendOnSignIn.
Confirms the initial email is sent on user setup, sign-in returns EMAIL_NOT_VERIFIED, and sendVerificationEmail is called twice when enabled and once when disabled.

### DIFF
--- a/packages/better-auth/src/api/routes/sign-in.test.ts
+++ b/packages/better-auth/src/api/routes/sign-in.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect } from "vitest";
+import { describe, expect, vi } from "vitest";
 import { getTestInstance } from "../../test-utils/test-instance";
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/api/routes/sign-in.test.ts
```diff
@@ -44,4 +44,62 @@ describe("sign-in", async (it) => {
...
+		try {
+			await auth.api.signInEmail({
+				body: {
+					email: testUser.email,
+					password: testUser.password,
+				},
+			});
+		} catch (err: any) {
+			expect(err.body.code).toBe("EMAIL_NOT_VERIFIED");
+		}
```
himself65: expect().reject.toThrow

Issue #4422, opened 2025-09-04
Request: Update to jose ^6


kevcube:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

I'm trying to use some better-auth features in a react-native application, and I'm experiencing problems because of usage of jose 5.10.0, which depends on node:crypto.

As of jose 6, the webcryptoAPI is used, which would make everything more portable.

### Describe the solution you'd like

update to Jose ^6

### Describe alternatives you've considered

dependency overrides

varshithmee:
my app cant run because "subtle is undefined" error and I traced it to jose importing node:crypto.

edit: It is an expo app

kevcube:
@varshithmee I'm still testing this. a fix for now is to override to jose 6 and use `import 'react-native-webcrypto'` at the top level of your app.

varshithmee:
> [@varshithmee](https://github.com/varshithmee) I'm still testing this. a fix for now is to override to jose 6 and use `import 'react-native-webcrypto'` at the top level of your app.

How do you do that?

kevcube:
@varshithmee in package.json add 
```
  "overrides": {
    "jose": "^6.0.0"
  },
```

install `react-native-webcrypto` and import it before activating any better-auth plugins.

varshithmee:
> @varshithmee in package.json add 
> ```
>   "overrides": {
>     "jose": "^6.0.0"
>   },
> ```
> 
> install `react-native-webcrypto` and import it before activating any better-auth plugins.

Thannks a lot! Will try as soon as possible !

himself65:
> my app cant run because "subtle is undefined" error and I traced it to jose importing node:crypto.
> 
> edit: It is an expo app

hi can you try v1.3.10-beta.4

melv-n:
> > my app cant run because "subtle is undefined" error and I traced it to jose importing node:crypto.
> > edit: It is an expo app
> 
> hi can you try v1.3.10-beta.4

It removed some errors, but still getting:

> TypeError: Cannot read property 'subtle' of undefined

originating from the `'better-auth/client/plugins'` import.

It does work with importing `react-native-webcrypto` before the plugins, however.

PR #4433, opened 2025-09-04
chore: update jose and crypto dependencies

kevcube:
Fixes: https://github.com/better-auth/better-auth/issues/4422

these are problematic in expo projects, updating them resolves some warnings and issues.

jose causes issues with node:crypto not being available in react-native

the @noble packages are outdated and don't export correctly, causing warnings in metro bundler.

    

---

## Summary by cubic
Update jose and @noble crypto deps to fix React Native/Expo issues and silence Metro warnings. This removes node:crypto pitfalls in RN and stabilizes auth in mobile builds.

- **Dependencies**
  - jose: ^5.10.0 → ^6.1.0 (avoids node:crypto usage in React Native)
  - @noble/ciphers: ^0.6.0 → ^2.0.0 (fixes bad exports/warnings)
  - @noble/hashes: ^1.8.0 → ^2.0.0 (fixes bad exports/warnings)

- **Migration**
  - @noble/ciphers and @noble/hashes v2 require Node 20.19+ in dev/build environments.

### DIFF
--- a/packages/better-auth/src/crypto/index.ts
+++ b/packages/better-auth/src/crypto/index.ts
@@ -1,11 +1,11 @@
 import { createHash } from "@better-auth/utils/hash";
-import { xchacha20poly1305 } from "@noble/ciphers/chacha";
+import { xchacha20poly1305 } from "@noble/ciphers/chacha.js";
 import {
 	bytesToHex,
 	hexToBytes,
 	utf8ToBytes,
 	managedNonce,
-} from "@noble/ciphers/utils";
+} from "@noble/ciphers/utils.js";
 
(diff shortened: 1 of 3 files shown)


kevcube:
closes #4422

kevcube:
looks like there's some discrepancy between tsconfig and your testing setup, because you have to import by .js for tests to pass. outside of scope here but IMO nice to fix.

himself65:
> looks like there's some discrepancy between tsconfig and your testing setup, because you have to import by .js for tests to pass. outside of scope here but IMO nice to fix.

Yeah, that's another thing. Set it up correctly might take extra work. Won't do it now

PR #4440, opened 2025-09-04
fix: add missing closing parenthesis in PostgreSQL timestamp definition

LoannPowell:
- Updated test snapshots to reflect the corrected generated code
- All tests now pass successfully

Fixes issue where PostgreSQL timestamp fields had invalid syntax:
- Before: timestamp('name')
- After:  timestamp('name', { withTimezone: true })
    

---

## Summary by cubic
Fix PostgreSQL timestamp generation in the CLI by adding the missing closing parenthesis and emitting timestamp('name', { withTimezone: true }) for timestamp fields. Updated test snapshots to match the corrected output.

PR #4447, opened 2025-09-04
fix: add timezone support to PostgreSQL timestamps in Drizzle schema …

JuanPabloGilA:
- Update PostgreSQL timestamp generation to include { withTimezone: true }
- This ensures proper timezone handling for PostgreSQL databases
- Update test snapshots to reflect the new timestamp format

References:
Preview Pr's
https://github.com/better-auth/better-auth/pull/4440
https://github.com/better-auth/better-auth/pull/3477

Open issues:
https://github.com/better-auth/better-auth/issues/3461
    

---

## Summary by cubic
Make PostgreSQL timestamps timezone-aware in the Drizzle schema generator by using timestamp(name, { withTimezone: true }). This fixes incorrect time handling and updates generated snapshots for all timestamp fields.

- **Migration**
  - Regenerate the schema and run a migration to convert affected PostgreSQL columns to timestamp with time zone.

### DIFF
--- a/packages/cli/src/generators/drizzle.ts
+++ b/packages/cli/src/generators/drizzle.ts
@@ -104,7 +104,7 @@ export const generateDrizzleSchema: SchemaGenerator = async ({
 				},
 				date: {
 					sqlite: `integer('${name}', { mode: 'timestamp' })`,
-					pg: `timestamp('${name}')`,
+					pg: `timestamp('${name}', { withTimezone: true })`,
 					mysql: `timestamp('${name}')`,
 				},
 				"number[]": {
--- a/packages/cli/test/__snapshots__/auth-schema-number-id.txt
+++ b/packages/cli/test/__snapshots__/auth-schema-number-id.txt
@@ -13,8 +13,10 @@ export const custom_user = pgTable("custom_user", {
   email: text("email").notNull().unique(),
   emailVerified: boolean("email_verified").default(false).notNull(),
   image: text("image"),
-  createdAt: timestamp("created_at").defaultNow().notNull(),
-  updatedAt: timestamp("updated_at")
+  createdAt: timestamp("created_at", { withTimezone: true })
+    .defaultNow()
+    .notNull(),
+  updatedAt: timestamp("updated_at", { withTimezone: true })
     .defaultNow()
     .$onUpdate(() => /* @__PURE__ */ new Date())
     .notNull(),
@@ -25,10 +27,12 @@ export const custom_user = pgTable("custom_user", {
 
 export const custom_session = pgTable("custom_session", {
   id: serial("id").primaryKey(),
-  expiresAt: timestamp("expires_at").notNull(),
+  expiresAt: timestamp("expires_at", { withTimezone: true }).notNull(),
   token: text("token").notNull().unique(),
-  createdAt: timestamp("created_at").defaultNow().notNull(),
-  updatedAt: timestamp("updated_at")
+  createdAt: timestamp("created_at", { withTimezone: true })
+    .defaultNow()
+    .notNull(),
+  updatedAt: timestamp("updated_at", { withTimezone: true })
     .$onUpdate(() => /* @__PURE__ */ new Date())
     .notNull(),
   ipAddress: text("ip_address"),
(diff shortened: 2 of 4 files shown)


himself65 (commented):
I'm not sure if I'm correct, but I guess this will have a breaking change for people who use Drizzle DB?

JuanPabloGilA:
> I'm not sure if I'm correct, but I guess this will have a breaking change for people who use Drizzle DB?

The change is on the generator. So it's not a breaking change.

So it will be like if they dont want the issue happen, they should update their databases in drizzle with the new schema that contains withTimeZone on the dates. (that's the thing that i did, update my database with migrations, after make this same change).

Btw normally the issue is for devs that have to use server databases and are more than 3 hours away from the database server. In a production enviroment this shouldn't happen because the database and API should be in the same server or similar time zones.

himself65:
> So it will be like if they dont want the issue happen, they should update their databases in drizzle with the new schema that contains withTimeZone on the dates. (that's the thing that i did, update my database with migrations, after make this same change).

I'm not an expert on drizzle. if they migrate the database. do they need any extra step for that or drizzle will auto generate the upgrade SQL

JuanPabloGilA:
> > So it will be like if they dont want the issue happen, they should update their databases in drizzle with the new schema that contains withTimeZone on the dates. (that's the thing that i did, update my database with migrations, after make this same change).
> 
> I'm not an expert on drizzle. if they migrate the database. do they need any extra step for that or drizzle will auto generate the upgrade SQL

https://github.com/JuanPabloGilA/hono-react-boilerplate/blob/main/apps/api/db/schema.ts

You can see an example there, building this i encounter the bug. The only thing that they have to do is modify the schemas and push to the database, but if they don't do it, it will not have any effect in the functionallity (With that repo you can test, i have a prod one but i can't share at the moment).

himself65 (approved):
LGTM, I will hold this for days and see how other people think

himself65 (commented):
Does this cause DB to `timestampz`?

JuanPabloGilA:
> Does this cause DB to `timestampz`?

Yes

himself65:
I've discussed with the team, and we don't want to ship to `timezonez` for the user by default. SO I think the better way is to have an option here and by default disable it.

Bekacru:
We want to use `UTC` consistently everywhere. If you use `timestamp with timezone`, the query client will try to convert it to your local time, which causes a lot of issues. Storing times directly in `UTC` and strictly comparing them in `UTC` works reliably

If you want to use `withTimezone`, that’s fine - just make sure your database itself is set to `UTC`

JuanPabloGilA:
> We want to use `UTC` consistently everywhere. If you use `timestamp with timezone`, the query client will try to convert it to your local time, which causes a lot of issues. Storing times directly in `UTC` and strictly comparing them in `UTC` works reliably
> 
> If you want to use `withTimezone`, that’s fine - just make sure your database itself is set to `UTC`

Guys, I’m fine with the reasoning and I can close the PR if you prefer stability in the current better-auth status for avoid data migrations. 

Just to clarify: timestamptz in Postgres behaves the same as timestamp in MySQL both store values in UTC and apply timezone conversion on the client. If we want strict consistency across engines, then the MySQL side should probably use datetime instead of timestamp because both have wall time.

That way, Postgres timestamp and MySQL datetime line up (both are “naive” UTC), while Postgres timestamptz and MySQL timestamp line up (both are “UTC with client conversion”).

himself65:
> We want to use `UTC` consistently everywhere. If you use `timestamp with timezone`, the query client will try to convert it to your local time, which causes a lot of issues. Storing times directly in `UTC` and strictly comparing them in `UTC` works reliably
> 
> If you want to use `withTimezone`, that’s fine - just make sure your database itself is set to `UTC`

TBH, I think if you set the local machine timezone correctly, `timestampz` will handle the logic correctly. I don't think there's a comparison issue here

BaptisteCrouzet:
Letting the user choose between timestamp with timezone or without would be great, everyone then could choose.
**Maybe it is a better path here than setting something as default ?**

But `timestamp with timezone` is a bit misleading: PostgreSQL does not store the textual time zone, it stores an absolute point in time.
On input, PostgreSQL converts the value according to the provided time zone or the session time zone.
On output, PostgreSQL displays the value again using the session time zone.
So 2026-05-26 14:00:00+02 and 2026-05-26 12:00:00+00 represent the same instant.

I see it as something really normal when we talk about createAt/updatedAt fields. It allows some users to be more consistent with the rest of their DB in a multi timezone context.

ping-maxwell:
Closing as stale, and with new drizzle adapter around the corner this PR won't cover all grounds, we should create a new one.

PR #4480, opened 2025-09-06
feat: expose passkey to plugins

okisdev:
This PR exposes the Passkey authentication method to align with other plugins.
    

---

## Summary by cubic
Exposed the Passkey plugin via better-auth/plugins to match other plugins and simplify imports. Updated demo, docs, CLI, and tests to use the new path.

- **New Features**
  - Exported passkey from better-auth/plugins for direct import.

- **Migration**
  - Replace imports of "better-auth/plugins/passkey" with "better-auth/plugins".

### DIFF
--- a/demo/nextjs/lib/auth.ts
+++ b/demo/nextjs/lib/auth.ts
@@ -11,6 +11,7 @@ import {
 	customSession,
 	deviceAuthorization,
 	lastLoginMethod,
+	passkey,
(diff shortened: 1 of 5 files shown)


himself65 (approved):

Bekacru:
We intentionally don’t export passkeys under /plugins because, even when they’re not imported directly, they require additional packages and those packages don’t work in all environments

PR #4481, opened 2025-09-06
fix: sidebar content and mobile nav

bytaesu:
---

This PR removes incorrect href assignments in the sidebar content and fixes the mobile navigation to display the menu correctly.

Only the Docs website has been changed, the main packages remain unaffected.

---

### **Before:**

https://github.com/user-attachments/assets/5de20b81-4dee-4367-b4c2-855f8906eb15

### **After:**

https://github.com/user-attachments/assets/041d2df5-9d29-4746-9a89-bdb2cb8b25a0

    

---

## Summary by cubic
Stops accidental navigation in the docs sidebar and fixes the mobile menu rendering. Group headers are now non-clickable dividers; only real items navigate.

- **Bug Fixes**
  - Sidebar: cleared href on group headers to prevent navigation.
  - Mobile nav: render group headers as non-interactive and fix item borders; links still close the menu on tap.
  - Scope: docs site only; no package changes.

### DIFF
--- a/docs/components/nav-mobile.tsx
+++ b/docs/components/nav-mobile.tsx
@@ -161,28 +161,40 @@ function DocsNavBarContent() {
 							</div>
 						</AccordionTrigger>
 						<AccordionContent className="pl-5 divide-y">
-							{menu.list.map((child) => (
-								<Link
-									href={child.href}
-									key={child.title}
-									className="block py-2 text-sm border-b first:pt-0 last:pb-0 last:border-0 text-muted-foreground"
-									onClick={toggleNavbar}
-								>
-									{child.group ? (
-										<div className="flex flex-row items-center gap-2 ">
-											<div className="flex-grow h-px bg-gradient-to-r from-stone-800/90 to-stone-800/60" />
-											<p className="text-sm text-transparent bg-gradient-to-tr dark:from-gray-100 dark:to-stone-200 bg-clip-text from-gray-900 to-stone-900">
-												{child.title}
-											</p>
+							{menu.list.map((child, index) =>
+								child.group ? (
+									// Group header rendered as div (just a divider)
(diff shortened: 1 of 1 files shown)


bytaesu:
since this isn’t a docs content change, I removed ‘docs’ from the title

docs/components/sidebar-content.tsx
himself65, line 1: why delete all href?

himself65 (changes requested):

bytaesu:
hi @himself65,

It was assigned an href where there shouldn’t have been one in the first place.
Try opening it on mobile and clicking the menu divider -> you’ll see a 404 error.

https://github.com/better-auth/better-auth/blob/6d0048e2cfedf11376e050206434b67f0ed3ff50/docs/components/side-bar.tsx#L98

pic

docs/components/sidebar-content.tsx
bytaesu: refer to: https://github.com/better-auth/better-auth/pull/4481#issuecomment-3264231442

bytaesu:
> I'm approving this. may be for those non existed route like the parent route we might have to do the default redirect like for `docs/integrations` we might have to do a default redirect to the first item in the intergration list like `docs/integrations/next` or something like that

Yes, I was also going to link to the first item in the group, but I thought it might get complicated to maintain if the order changes. So I submitted the PR to remove the link from the group headers by default.

Kinfe123:
Yeah. technically those group headers should not have a link if so we either should do a home page for those group or needs to redirect to any of the child with in that group. for now this LGTM.

bytaesu:
@himself65, sorry for the ping.
Is it okay to merge this into canary?
I noticed the base branch was previously main.

PR #4492, opened 2025-09-07
fix: Cognito documentation

spoon16:
The Cognito documentation contains invalid code snippets.
    

---

## Summary by cubic
Fixed the Cognito docs to match the current API and prevent copy/paste setup errors. The example now uses socialProviders.cognito with CognitoOptions instead of the outdated socialProvider + cognito(...) call.

### DIFF
--- a/docs/content/docs/authentication/cognito.mdx
+++ b/docs/content/docs/authentication/cognito.mdx
@@ -28,8 +28,7 @@ description: Amazon Cognito provider setup and usage.
     Configure the `cognito` key in `socialProviders` key of your `auth` instance.
 
     ```ts title="auth.ts"
-    import { betterAuth } from "better-auth"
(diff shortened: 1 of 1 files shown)


docs/content/docs/authentication/cognito.mdx
```diff
@@ -25,20 +25,21 @@ description: Amazon Cognito provider setup and usage.
 
   <Step>
     ### Configure the provider
-    Import the Cognito provider and pass it to the `providers` option of your `auth` instance.
+    Configure the `cognito` key in `socialProviders` key of your `auth` instance.
 
     ```ts title="auth.ts"
+    import { type CognitoOptions } from 'better-auth/social-providers';
```
Kinfe123: you can remove the CognitoOptions for assertion

PR #4494, opened 2025-09-07
fix(cli): add missing JSON type to schema generation

TheGB0077:
PR #4297 Added support for JSON type to supported fields, but the cli did not receive the same updates. In addition, I replaced JSON to JSONb as it the preferred option in the context of pg databases.
    

---

## Summary by cubic
Adds JSON field support to the CLI schema generators and switches PostgreSQL to JSONB in migrations. This aligns the CLI with core JSON support introduced in #4297.

- **New Features**
  - Drizzle mapping: sqlite -> text, postgres -> jsonb, mysql -> json.
  - Prisma mapping: Json/Json? based on optionality.
  - Migration generator now outputs jsonb for postgres.
  - Tests added for pg, mysql, sqlite, and Prisma.


packages/cli/test/json-field.test.ts
```diff
@@ -0,0 +1,99 @@
+import { describe, expect, it, vi } from "vitest";
+import { generateDrizzleSchema } from "../src/generators/drizzle";
+import { generatePrismaSchema } from "../src/generators/prisma";
+import type { BetterAuthOptions } from "better-auth";
+
+describe("JSON field support in CLI generators", () => {
```
Bekacru: can we move this to `generate.test.ts` file instead

TheGB0077: Sure. I'll move it up.

Kinfe123 (approved):

TheGB0077:
Are we ready to merge? @Bekacru

PR #4502, opened 2025-09-07
feat(rate-limiter): allow disabling custom paths to not be rate limited

### LINKED ISSUE #4497: Rate limiting applies globally (cannot restrict to specific routes)

### Description

When enabling rateLimit in Better Auth, it currently applies globally to all routes (e.g., /get-session).
This results in unnecessary keys being created in secondary storage (see screenshot below).

I would like to restrict rate limiting to only certain routes (e.g., /callback/google or /sign-in/social). While customRules lets me override the window/max for a given route, it doesn’t allow me to fully opt-out of rate limiting on other routes.

### To Reproduce

Configure Better Auth with secondary storage (Cloudflare KV in my case).

Enable rateLimit with secondary storage.

Add a custom rule for /sign-in/social.

Make requests against /get-session.

### Result

Rate limiting is still applied to /get-session and keys are created in storage, even though I only want it on my social login routes.

Current: Rate limiting is applied to all routes by default, even ones I don’t want to rate-limit (e.g., /get-session).

Expected: Ability to apply rate limiting only to selected routes. For example, only /sign-in/social and /callback/google, while excluding /get-session.

What version of Better Auth are you using?

1.3.7

### System info

```bash
-
```

Backend

```typescript
### Auth config

import { drizzleAdapter } from "better-auth/adapters/drizzle"
import { db } from "~/db"
import { accounts, users, verifications } from "~/db/schema"
import { getCloudflareCtx } from "../cloudflare"

  session: {
    expiresIn: 60 * 60 * 24 * 7,
    updateAge: 60 * 60 * 24,
  },
  advanced: {
    ipAddress: {
      ipAddressHeaders: ["cf-connecting-ip"],
    },
  },
  rateLimit: {
    enabled: true,
    window: 60,
    max: 100,
    storage: "secondary-storage",
    modelName: "rateLimit",
    customRules: {
      "/sign-in/social": async request => {
        return {
          window: 60,
          max: 6,
        }
      },
    },
  },
    provider: "pg",
    schema: {
      verification: verifications,
      account: accounts,
      user: users,
    },
  }),
  user: {
    additionalFields: {
      role: {
        type: "string",
        fieldName: "role",
      },
      points: {
        type: "number",
        fieldName: "points",
      },
    },
  },
  secondaryStorage: {
    get: async key => {
      const cf = await getCloudflareCtx()
      const value = await cf.KV_PLATFORM.get(key)
      return value ? value : null
    },
      const cf = await getCloudflareCtx()
      if (ttl) await cf.KV_PLATFORM.put(key, value, { expirationTtl: ttl })
      else await cf.KV_PLATFORM.put(key, value)
    },
    delete: async key => {
      const cf = await getCloudflareCtx()
      await cf.KV_PLATFORM.delete(key)
    },
  },
  secret: process.env.BETTER_AUTH_SECRET!,
  baseURL: process.env.BETTER_AUTH_URL!,
  socialProviders: {
    github: {
      clientId: process.env.GITHUB_CLIENT_ID!,
      clientSecret: process.env.GITHUB_CLIENT_SECRET!,
    },
    google: {
    },
  },
})
```

### Additional context

_No response_

Bekacru:
closes #4497

### DIFF
--- a/docs/content/docs/concepts/rate-limit.mdx
+++ b/docs/content/docs/concepts/rate-limit.mdx
@@ -109,6 +109,21 @@ export const auth = betterAuth({
 })
 ```
 
+If you like to disable rate limiting for a specific path, you can set it to `false` or return `false` from the custom rule function.
+
+```ts title="auth.ts"
+import { betterAuth } from "better-auth";
+
+export const auth = betterAuth({
+    //...other options
+    rateLimit: {
+        customRules: {
+            "/get-session": false,
+        },
+    },
+})
+```
+
 ### Storage
 
 By default, rate limit data is stored in memory, which may not be suitable for many use cases, particularly in serverless environments. To address this, you can use a database, secondary storage, or custom storage for storing rate limit data.
--- a/packages/better-auth/src/api/rate-limiter/index.ts
+++ b/packages/better-auth/src/api/rate-limiter/index.ts
@@ -162,6 +162,10 @@ export async function onRequestRateLimit(req: Request, ctx: AuthContext) {
 				window = resolved.window;
 				max = resolved.max;
 			}
+
+			if (resolved === false) {
+				return;
+			}
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


himself65 (approved):

aarmful:
would it be possible to *only* rate-limit auth routes? it appears secondary storage is storing requests to static files in my kv

PR #4522, opened 2025-09-08
fix: refresh secondary storage sessions on user update

frectonz:
The `secondaryStorage` sessions were not being updated when a user update was performed using the `updateUserByEmail` function. This PR fixes that issue.


phuctm97:
Related https://github.com/better-auth/better-auth/issues/4721

When can this be merged and released?

PR #4414, opened 2025-09-04
feat: database transaction support

himself65:
Related: https://github.com/better-auth/better-auth/pull/4062
    

---

## Summary by cubic
Add first-class database transactions with a new Adapter.transaction API and an optional AdapterConfig.transaction. Drizzle and Kysely run real transactions; Memory, MongoDB, and Prisma are non-transactional for now.

- **New Features**
  - Added Adapter.transaction(callback) to run operations atomically.
  - Added AdapterConfig.transaction to enable/disable or wire custom transaction behavior.
  - Drizzle: uses db.transaction(cb).
  - Kysely: runs callback inside db.transaction().execute(trx) with a trx-scoped adapter.
  - Memory, MongoDB, Prisma: transaction: false (no transactional behavior yet).

PR #4532, opened 2025-09-08
feat: database transaction support

jslno:
Related: #4414
Closes: #4062 

## Todo

- [x] Fix MongoDB transactions not rolling back
     - `config.session` was not enough as its required to create a fresh session for every transaction.
        Removed `config.session` and added `config.client` instead.
     - disabled transaction tests for now. MongoDB transactions require a replica set or a sharded cluster
- [x] Documentation
    

---

## Summary by cubic
Add transaction support to the database adapters so auth operations can run within a single DB transaction. Implements this for Prisma and Drizzle and standardizes adapter creation.

- **New Features**
  - Provide a transaction(cb) hook that runs the callback with a transactional adapter instance.
  - Prisma: use $transaction and pass a tx-scoped adapter to the callback.
  - Drizzle: use db.transaction and pass a tx-scoped adapter to the callback.
  - Introduce CreateAdapterOptions and lazy BetterAuthOptions to safely build tx-scoped adapters; update Kysely to use the new type.

- **Bug Fixes**
  - Drizzle: ensure all operations inside transaction() use the transaction connection.


himself65:
Thank u so much, I will polish it and merge later

PR #4544, opened 2025-09-09
fix(organization): `beforeCreateOrganization` didn't pass new body

ping-maxwell:
When the dev returns a value from the `beforeCreateOrganization` fn, it should then be forwarded to the createOrganization fn. This PR implements such fix.

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-org.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-org.ts
@@ -194,10 +194,15 @@ export const createOrganization = <O extends OrganizationOptions>(
 					};
 				}
 			}
+			let {
+				keepCurrentActiveOrganization: ___,
+				userId: ____,
+				...orgData2
+			} = ctx.body;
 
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/organization/routes/crud-org.ts
```diff
@@ -194,10 +194,15 @@ export const createOrganization = <O extends OrganizationOptions>(
 					};
 				}
 			}
+			let {
```
himself65: use const?

himself65 (changes requested):

packages/better-auth/src/plugins/organization/routes/crud-org.ts
ping-maxwell: done

ping-maxwell:
Seems this was fixed recently by Beka here:
https://github.com/better-auth/better-auth/pull/4623
Closing this now.

PR #4546, opened 2025-09-09
feat(adapter): sortBy multiple fields

jslno:
---

## Summary by cubic
Enable multi-field sorting in findMany across all adapters by allowing sortBy to accept one or many fields. Introduces a SortBy type and updates implementations; this may affect custom adapters that assumed a single sort field.

- New Features
  - sortBy now accepts SortBy or SortBy[] for findMany.
  - Implemented in Drizzle, Kysely, Prisma, MongoDB, and Memory adapters.
  - Added a test for multi-field sorting (currently skipped).

- Migration
  - Update custom adapters to accept sortBy as SortBy | SortBy[] and apply order across fields in sequence.
  - Keep support for a single-object sortBy to remain backward compatible with callers.


ping-maxwell (changes requested):
We don't use the `test.ts` file anymore, please move those tests to the new tests. In this case it would be under `normal.ts` [located here](https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/adapters/tests/normal.ts).

PR #4554, opened 2025-09-10
feat(db): add withTransaction helper

jslno:
I guess this could be closed and put into a 3rd party library. I think it's useful for transaction heavy plugins but I also don't see a use case for this in the core library, as there aren't (at least for the current state) many transactions needed. (Maybe revert to a simpler version, like I had before).   

## Summary by cubic
Adds internalAdapter.withTransaction to run auth DB operations in a single transaction and reduce boilerplate. Provides a transaction-scoped internalAdapter where the transaction adapter is auto-injected.

- **New Features**
  - Added internalAdapter.withTransaction(cb): executes cb within adapter.transaction.
  - cb receives { adapter, internalAdapter }; internalAdapter methods omit the trx param.
  - Introduced shimLastParam utility to auto-inject the last parameter (used for trx) when missing.

### DIFF
--- a/packages/better-auth/src/__snapshots__/init.test.ts.snap
+++ b/packages/better-auth/src/__snapshots__/init.test.ts.snap
@@ -98,6 +98,7 @@ exports[`init > should match config 1`] = `
     "updateUser": [Function],
     "updateUserByEmail": [Function],
     "updateVerificationValue": [Function],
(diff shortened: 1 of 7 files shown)


jslno:
@cubic-dev-ai review this PR

jslno:
@cubic-dev-ai review this PR

PR #4561, opened 2025-09-10
refactor(organization): use transactions

jslno:


jslno:
@cubic-dev-ai review this PR

himself65 (commented):
I'm not sure if we should put a single request as a one transaction by default

jslno:
> I'm not sure if we should put a single request as a one transaction by default

Yeah I figured that. I'm probably closing this anyways, as I'm currently working on a bridge for internalAdapter & plugin adapters. It handles adding trxAdapter as last parameter and routing the right adapter to each method, by adding a withTransaction method that shims the last parameters. This would make it easier to implement transactions for plugins if needed and also cleans up the `(trxAdapter || adapter)`. Just let me know what you think of it, I'll push the latest changes later and reference the PR.

himself65:
Thanks. that's cool

PR #4563, opened 2025-09-10
docs(organization): fix list org endpoint requires session

ping-maxwell:


ping-maxwell:
I think CI should check if docs are going to main, not from canary/staging

himself65 (approved):

Issue #4564, opened 2025-09-10
DB: Incorrect Field Mapping


0xJJW:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

TLDR: Incorrect mapping in `getTables()`

https://github.com/better-auth/better-auth/blob/5b784698641bc68d7ad3ea537e6d564a44e60878/packages/better-auth/src/db/get-tables.ts#L228

**Reproduction steps**

1. Add a custom field name for `refreshTokenExpiresAt`

```
export const auth = betterAuth({
  account: {
    fields: {
      accessTokenExpiresAt: "access_token_expires_at",
      refreshTokenExpiresAt: "refresh_token_expires_at",
    },
  },
});
```

2. Depending on the adapter you are using you will see different results:

**Kysely:**
The `refresh_token_expires_at` column will be missing from the database. 

**Prisma:**
The generated schema will include two `@map()` values for the previous field which will prevent you from running migrate.

```
Error: Prisma schema validation - (validate wasm)
Error code: P1012
error: Attribute "@map" can only be defined once.
  -->  prisma/schema.prisma:62
   | 
61 |   id_token                String?   @map("idToken")
62 |   access_token_expires_at DateTime? @map("accessTokenExpiresAt") @map("refreshTokenExpiresAt")
   | 
```

### Current vs. Expected behavior

Current:
1. Migration fails when using Prisma || Kysely DB missing `refresh_token_expires_at`

Expected:
1. Migrations run successfully
2. Account table should include `refresh_token_expires_at`

### What version of Better Auth are you using?

1.3.6

### System info

```bash
N/A
Issue pointed out in `current`
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  account: {
    fields: {
      accessTokenExpiresAt: "access_token_expires_at",
      refreshTokenExpiresAt: "refresh_token_expires_at",
    },
  },
});
```

himself65:
have you tried with latest better-auth?

0xJJW:
@himself65, I have not tested the latest version as we run a custom implementation of most BetterAuth plugins, and I won't be updating until later this week.

As highlighted in the **TLDR**, I have pointed out the cause of the issue, which is still present in the **canary** branch.

[...]

PR #4569, opened 2025-09-10
fix: custom field for `refreshTokenExpiresAt`

himself65:
Fixes: https://github.com/better-auth/better-auth/issues/4564
    

---

## Summary by cubic
Fixes the custom field mapping for refreshTokenExpiresAt in getAuthTables so it uses the correct override and not accessTokenExpiresAt. Enables proper customization without breaking defaults.

- **Bug Fixes**
  - Map refreshTokenExpiresAt to options.account.fields.refreshTokenExpiresAt (fallback to "refreshTokenExpiresAt").
  - Add tests to verify custom mapping, no cross-over with accessTokenExpiresAt, and default names when unset.

### DIFF
--- a/packages/better-auth/src/db/get-tables.test.ts
+++ b/packages/better-auth/src/db/get-tables.test.ts
@@ -1,0 +1,60 @@
+import { describe, expect, it } from "vitest";
+import { getAuthTables } from "./get-tables";
+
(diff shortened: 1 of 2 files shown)


0xJJW:
Thanks @himself65 for the swift fix 🧙

PR #4573, opened 2025-09-10
fix(lastLoginMethod): add fallback support for customResolveMethod

lumpinif:
## Summary

  Previously, when users provided a `customResolveMethod`, it completely replaced the default resolution logic, causing them to lose built-in functionality for standard authentication paths like `/sign-in/email` and OAuth callbacks (`/callback/google`,
  `/oauth2/callback/github`, etc.).

  ## Problem

  Users who wanted to track custom authentication methods (like SAML, magic links, phone auth) had to choose between:
  - **Option A**: No `customResolveMethod` → only built-in paths work
  - **Option B**: Use `customResolveMethod` → lose ALL built-in path detection

  This forced users to reimplement all the default logic manually, which was error-prone and defeated the purpose of having built-in detection.

  ## Solution

  Implement a **fallback pattern** where:
  1. Try the user's `customResolveMethod` first
  2. If it returns `null`, fall back to the default resolution logic
  3. This allows extending functionality without losing built-in behavior

  ## Changes

  - Add fallback logic that preserves default path detection
  - Users can return `null` from `customResolveMethod` to use built-in detection
  - Comprehensive test suite covering all fallback scenarios
  - Updated documentation with improved examples and behavior explanation

  ## Test Plan

  - [x] All existing tests pass
  - [x] Added 5 new test cases covering:
    - Default behavior when no custom method provided
    - Custom method completely overriding (returns non-null always)
    - Custom method with fallback (returns null for standard paths)
    - Mixed custom + default path handling
    - Database storage compatibility with fallback

  ## Before/After

  **Before (Broken):**
  ```typescript
  // User loses ALL built-in detection
  lastLoginMethod({
    customResolveMethod: (ctx) => {
      if (ctx.path === "/custom-auth") return "custom";
      return null; // ❌ Standard paths like /sign-in/email return null
    }
  })

  After (Fixed):
  // User gets custom paths + built-in detection
  lastLoginMethod({
    customResolveMethod: (ctx) => {
      if (ctx.path === "/custom-auth") return "custom";
      return null; // ✅ Falls back to built-in: /sign-in/email → "email"
    }
  })
```
  Benefits

  - ✅ Extend without breaking: Add custom auth methods while keeping built-in functionality
  - ✅ No reimplementation needed: Users don't have to rewrite email/OAuth detection
  - ✅ Backward compatible: Existing custom methods that handle all paths still work
  - ✅ Intuitive behavior: Returning null means "use default logic" (expected behavior)
    

---

## Summary by cubic
Fixes lastLoginMethod so customResolveMethod now falls back to default detection when it returns null. Users can add custom auth methods without losing built-in email/OAuth detection.

- **Bug Fixes**
  - Implement fallback: try customResolveMethod first; if it returns null, use default resolver.
  - Preserve built-in paths like /sign-in/email and /callback/:provider.
  - Remains backward compatible when custom methods always return a value.
  - Add tests for no custom method, full override, fallback, mixed paths, and DB storage.
  - Update docs with examples and clearer behavior.

### CODE REVIEW

File: docs/content/docs/plugins/last-login-method.mdx [hunk 1]
```diff
@@ -328,27 +329,34 @@ export const auth = betterAuth({
...
     ]
 })
 ```
 
+**Benefits of this approach:**
+- ✅ Custom paths get your custom method names
+- ✅ Standard paths (`/sign-in/email`, `/callback/google`, etc.) still work automatically  
+- ✅ No need to reimplement built-in detection logic
+- ✅ Easy to extend without breaking existing functionality
+
```
Bekacru: let's remove the benefits section

Bekacru:
```suggestion

```

packages/better-auth/src/plugins/last-login-method/index.ts
```diff
@@ -49,16 +49,33 @@ export const lastLoginMethod = <O extends LastLoginMethodOptions>(
...
+		}
+		return null;
+	};
+
+	// Create the fallback resolve method
+	const resolveMethod = userConfig?.customResolveMethod
+		? (ctx: GenericEndpointContext) => {
+				// Try user's custom method first
+				const customResult = userConfig.customResolveMethod!(ctx);
+				if (customResult !== null) {
```
Bekacru: I think we should just check for `!customResult` otherwise `undefined` and `""` wouldn't use the fallback

lumpinif: nice catch! that's correct!

lumpinif:
fixed in https://github.com/better-auth/better-auth/pull/4821

PR #4576, opened 2025-09-10
feat(social-provider): multiple client ID support for social providers

Kinfe123:
## Summary by cubic
Add multiple client ID support for social providers (initially Google) so apps can choose a client per request (web, iOS, Android) while keeping current setups working by default. Token validation now accepts any configured client ID.

- New Features
  - signInSocial accepts optional clientId; stored in state and used in callback for code exchange.
  - Added SocialProviderMultipleClientOptions with configs[] for per-platform clientId/secret/scopes/redirectURI.
  - OAuth provider types now accept clientId in authorization URL and code validation flows.
  - Google provider resolves the requested client config, merges scopes, uses per-client redirectURI, and validates token aud against all configured client IDs.
  - Tests cover iOS/Android/web selection and fallback on invalid clientId.

- Migration
  - No breaking changes.
  - To use: define google.configs[] and pass clientId in signInSocial when selecting a specific client; otherwise the default client is used.

### DIFF
--- a/docs/content/docs/authentication/google.mdx
+++ b/docs/content/docs/authentication/google.mdx
@@ -141,3 +141,90 @@ socialProviders: {
   already authorized your app, you must have them revoke your app's access in
   their Google account settings, then re-authorize.
 </Callout>
+
+## Multiple Client ID Support
+
+Better Auth supports configuring multiple client IDs for the Google provider, which is useful when you need different OAuth configurations for different platforms (e.g., web, iOS, Android) or different redirect URIs.
+
+### Configuration
+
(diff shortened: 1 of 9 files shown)


himself65 (commented):
I think I committed something related to the multi-client ID. And [this PR](https://github.com/better-auth/better-auth/pull/4583) also adds the Google multi-client ID support.

If you have time, please double-check the code. Thanks

Kinfe123:
They both have a different approach on implementing it. Mine do expose config options from generic oauth and use that as way to add multiple config for multiple platforms and also easier to resolve compared to other as there is a client id per client secret config which also allowed in google credentials as well.

PR #4579, opened 2025-09-11
fix(one-time-token): typo and clean

gabrielmar:
## Summary by cubic
Fixes the one-time-token plugin by correcting the options type name and removing a duplicate token deletion call to avoid redundant DB writes. Also switches to the standard zod import.

- **Bug Fixes**
  - Delete verification value once (immediately after lookup) instead of twice.

- **Refactors**
  - Rename OneTimeTokenopts to OneTimeTokenOptions.
  - Import z from "zod" instead of "zod/v4".

### DIFF
--- a/packages/better-auth/src/plugins/one-time-token/index.ts
+++ b/packages/better-auth/src/plugins/one-time-token/index.ts
@@ -1,4 +1,4 @@
-import { z } from "zod";
+import * as z from "zod/v4";
 import {
 	createAuthEndpoint,
 	defaultKeyHasher,


packages/better-auth/src/plugins/one-time-token/index.ts
```diff
@@ -1,4 +1,4 @@
-import * as z from "zod/v4";
+import { z } from "zod";
```
himself65: we use v4 by default

himself65: it's better to revert this change, do that before merging

gabrielmar: I know, zod v4 is exported by default
https://github.com/colinhacks/zod/blob/main/packages/zod/src/index.ts#L3

himself65: make sense. But I think we could make that in a separate PR and replace all zod import.

gabrielmar: Import reverted.
I'll make another PR to change zod's imports

himself65: Thanks for doing that. I will appreciate it

PR #4583, opened 2025-09-11
feat: add support for multiple Google client IDs for cross-platform authentication

### LINKED ISSUE #4498: Support Multiple Google Client IDs for Cross-Platform Token Verification
Currently, `betterAuth` does not support verifying Google tokens from multiple client IDs within a single configuration. This limitation creates issues when building cross-platform applications where each platform (iOS, Android, Web) has its own Google OAuth client ID, but all need to authenticate with the same backend API.

**Use Case:**

* iOS app uses Google Client ID: `123-ios.googleusercontent.com`
* Android app uses Google Client ID: `456-android.googleusercontent.com`
* Both apps call the same `better-auth` API endpoint.

At present, only one client ID can be configured, which causes token verification to fail for other platform client ids.

**Expected Behavior:**
`better-auth` should allow configuring an **array of Google client IDs** and verify tokens against any of the configured client IDs or allow to use dynamic client id option at verification time.

This would enable seamless cross-platform authentication while maintaining security and consistency across apps.

RikhiSingh:
## Description

Adds support for configuring multiple Google OAuth client IDs in a single better-auth configuration, enabling cross-platform authentication where different platforms (iOS, Android, Web) can use their respective Google OAuth client IDs while authenticating with the same backend API.

## Problem

Currently, better-auth only supports a single Google client ID. This breaks cross-platform authentication where:
- iOS app uses: `123-ios.googleusercontent.com`
- Android app uses: `456-android.googleusercontent.com`
- Both need to authenticate with the same backend

## Solution

- Extended `GoogleOptions` interface to accept `clientId` as `string | string[]`
- Updated `verifyIdToken` to validate against any configured client ID
- Maintained backward compatibility and security

## Usage

```typescript
  socialProviders: {
    google: {
      clientId: [
        "123-ios.googleusercontent.com",
        "456-android.googleusercontent.com", 
        "789-web.googleusercontent.com"
      ],
      clientSecret: "your-client-secret",
    },
  },
});
```

## Testing

- ✅ 16 test cases covering all scenarios
- ✅ Backward compatible
- ✅ No breaking changes

Closes #4498
    

---

## Summary by cubic
Add support for multiple Google OAuth client IDs to enable cross-platform sign-in (iOS, Android, Web) against the same backend. Backward compatible; ID tokens are validated against any configured client ID.

- New Features
  - clientId accepts string | string[] in ProviderOptions (Google only)
  - Google provider uses the first client ID for auth URL, token exchange, and refresh
  - verifyIdToken checks aud against any configured client ID and validates issuer
  - Added tests to cover single and multiple client ID flows

- Refactors
  - Updated Biome schema to 2.2.3
  - Removed unused import in nav-mobile.tsx

### DIFF
--- a/packages/better-auth/src/social-providers/google.ts
+++ b/packages/better-auth/src/social-providers/google.ts
@@ -33,7 +33,8 @@ export interface GoogleProfile {
 	sub: string;
 }
 
-export interface GoogleOptions extends Omit<ProviderOptions<GoogleProfile>, "clientId"> {
+export interface GoogleOptions
+	extends Omit<ProviderOptions<GoogleProfile>, "clientId"> {
 	/**
 	 * The client ID(s) of your application
 	 * Can be a single client ID string or an array of client IDs for cross-platform support
@@ -77,7 +78,11 @@ export const google = (options: GoogleOptions) => {
 
 			// Use the first client ID for authorization URL creation
 			const primaryClientId = Array.isArray(options.clientId)
-				? options.clientId[0]
+				? options.clientId[0] ||
+					(() => {
+						logger.error("Google clientId array cannot be empty");
+						throw new BetterAuthError("CLIENT_ID_AND_SECRET_REQUIRED");
+					})()
 				: options.clientId;
 
 			const _scopes = options.disableDefaultScope
@@ -110,7 +115,11 @@ export const google = (options: GoogleOptions) => {
 		validateAuthorizationCode: async ({ code, codeVerifier, redirectURI }) => {
 			// Use the first client ID for token exchange
 			const primaryClientId = Array.isArray(options.clientId)
-				? options.clientId[0]
+				? options.clientId[0] ||
+					(() => {
+						logger.error("Google clientId array cannot be empty");
+						throw new BetterAuthError("CLIENT_ID_AND_SECRET_REQUIRED");
+					})()
 				: options.clientId;
 
 			return validateAuthorizationCode({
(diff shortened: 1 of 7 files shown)


packages/better-auth/src/social-providers/google.ts
```diff
@@ -33,7 +33,13 @@ export interface GoogleProfile {
 	sub: string;
 }
 
-export interface GoogleOptions extends ProviderOptions<GoogleProfile> {
+export interface GoogleOptions
+	extends Omit<ProviderOptions<GoogleProfile>, "clientId"> {
```
himself65, line 37: I think we might need to update the `ProviderOptions` type to allow different input. Let me do this in another PR

RikhiSingh: I approached it earlier differently, made clientId as string | string[] but then other OAuth providers expect the same, then, will have to  update all the OAuth providers to expect an array then, or create a new param only for Google. 

Let me know your thoughts.

himself65: im doing that

himself65: https://github.com/better-auth/better-auth/pull/4596#discussion_r2342587369

RikhiSingh:
> Can you show the document of the multiple client IDs by Google? I want to double check the authrozation url stuff if it's correct

Heres the docs I refered to:
- [Cross-Client Identity Documentation](https://developers.google.com/identity/protocols/oauth2/cross-client-identity)
- [OAuth 2.0 Web Server Flow](https://developers.google.com/identity/protocols/oauth2/web-server)

So:
1. Authorization URLs must use exactly one `client_id` parameter (not multiple)
2. ID Token Verification can accept tokens from any client ID in the same project
3. Shared Consent user consent applies to all client IDs in the project

So with comment I proposed on the #4596 
- Uses first client ID for authorization URLs
- Verifies tokens against all configured client IDs
- and it will follows Google's official guidelines

MarkLee425:
The same client secret cannot be shared in multiple clients (i.e. web, ios and android). This will not be working if only allowing single clientSecret.

vxshta:
Stuck for the past 3 days trying to find a workaround, What is the current status of this I am still not able to use an array of client id's

Issue #4586, opened 2025-09-11
Performance issue on large organization member lists


XavierGeerinck:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

When having large member lists (>10k members) in organizations, performance becomes slow due to way the `listMembers` and `listTeamMembers` (as part of https://github.com/better-auth/better-auth/pull/4582 but in analogy to `listMembers`) functions are implemented.

Currently they do not use `JOINS` but rather a 2-query approach where:
1. They fetch the members
2. They fetch the member details

When using a JOIN, the database would efficiently utilize existing indexes to handle this and filter out the non-required members, returning exactly and efficiently what we need.

### Current vs. Expected behavior

N/A

### What version of Better Auth are you using?

latest

### System info

```bash
N/A
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
N/A
```

himself65:
I think maybe we should have pagination API endpoint for that

maulik13:
I second the suggestion of introducing pagination with flexible page size. However, join would still be a good option as it sounds like a low hanging fruit.

PR #4587, opened 2025-09-11
chore: try working on join optimization in the adapter

XavierGeerinck:
This PR tries to tackle the missing JOIN operation in the adapters that can optimize some queries.
I'm currently mainly looking as well how the tests for the adapters should look like, but currently most of them are passing (besides the Drizzle one)
Issue Reference: https://github.com/better-auth/better-auth/issues/4586

### Before - Multiple Queries

```typescript
const members = await adapter.findMany({ model: "member", where: [...] });
const users = await adapter.findMany({ 
  model: "user", 
  where: [{ field: "id", operator: "in", value: memberIds }] 
});
```

### After - Single JOIN Query

```typescript
const membersWithUsers = await adapter.findMany({
  model: "member",
  where: [...],
  joins: [{
    type: "inner",
    table: "user",
    on: { left: "member.userId", right: "user.id" },
    select: ["id", "name", "email", "image"]
  }]
});
```
    

---

## Summary by cubic
Added JOIN support to adapters to run cross-table queries in a single call, improving performance for list and lookup operations. The organization plugin now fetches members with user data via one joined query.

- New Features
  - Added joins to Adapter.findOne, findMany, and count (inner/left/right/full, alias, select, on).
  - Implemented JOINs in Kysely, Prisma, MongoDB, and Memory adapters; Drizzle groundwork added.
  - Output preserves join-prefixed fields (table_or_alias_field) when joins are used.
  - Added shared JOIN test suite and adapter-specific tests.

- Refactors
  - Organization members listing uses a single JOIN to include user data.
  - Adapter factory and types updated to accept and forward joins without breaking existing calls.

### DIFF
--- a/packages/better-auth/src/adapters/create-adapter/index.ts
+++ b/packages/better-auth/src/adapters/create-adapter/index.ts
@@ -1,7 +1,7 @@
 import { safeJSONParse } from "../../utils/json";
 import { withApplyDefault } from "../../adapters/utils";
 import { getAuthTables } from "../../db/get-tables";
-import type { Adapter, BetterAuthOptions, Where } from "../../types";
+import type { Adapter, BetterAuthOptions, Join, Where } from "../../types";
 import { generateId as defaultGenerateId, logger } from "../../utils";
 import type {
 	AdapterConfig,
@@ -419,7 +419,7 @@ export const createAdapter =
 			data: Record<string, any> | null,
 			unsafe_model: string,
 			select: string[] = [],
-			joins?: any[],
+			joins?: Join[],
 		) => {
 			if (!data) return null;
 			const newMappedKeys = config.mapKeysTransformOutput ?? {};
@@ -752,7 +752,7 @@ export const createAdapter =
 				model: string;
 				where: Where[];
 				select?: string[];
-				joins?: import("../../types").Join[];
+				joins?: Join[];
 			}) => {
 				transactionId++;
 				let thisTransactionId = transactionId;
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/adapters/create-adapter/index.ts
```diff
@@ -419,6 +419,7 @@ export const createAdapter =
 			data: Record<string, any> | null,
 			unsafe_model: string,
 			select: string[] = [],
+			joins?: any[],
```
himself65: no any[]

packages/better-auth/src/adapters/create-adapter/index.ts
```diff
@@ -892,9 +919,11 @@ export const createAdapter =
 			count: async ({
 				model: unsafeModel,
 				where: unsafeWhere,
+				joins,
 			}: {
 				model: string;
 				where?: Where[];
+				joins?: import("../../types").Join[];
```
himself65: can we directly import the type on top level?

XavierGeerinck: Yes! Definitely. Changed

packages/better-auth/src/adapters/create-adapter/index.ts
XavierGeerinck: Changed! Good spot

himself65 (changes requested):
Hi, I just know our core team @ping-maxwell is also working on the `join` feature. I think we should have open discussion to avoid duplicate work

XavierGeerinck:
Seeing Join support was added and no comments were made I am closing this, sad to see no open discussion was held

PR #4593, opened 2025-09-11
fix: normalize zod imports

gabrielmar:
## Summary by cubic
Standardized Zod imports across the repo by replacing all "zod/v4" and mixed default/named imports with a single pattern: import * as z from "zod" (and type-only imports from "zod"). This removes subpath imports that can break bundlers and keeps types consistent, with no behavior changes.

- **Refactors**
  - Replaced all "zod/v4" imports with "zod" in src, plugins, and tests.
  - Updated OpenAPI generator to use z.ZodType, z.ZodObject, and z.ZodOptional.
  - Normalized default Zod imports to namespace form.

### DIFF
--- a/packages/better-auth/src/api/call.test.ts
+++ b/packages/better-auth/src/api/call.test.ts
@@ -8,7 +8,7 @@ import {
 } from ".";
 import { init } from "../init";
 import type { BetterAuthOptions, BetterAuthPlugin } from "../types";
-import * as z from "zod/v4";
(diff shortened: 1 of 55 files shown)


gabrielmar:
https://github.com/gabrielmar/better-auth/blob/353aedc7f73b193f3b18676681158eac9db0f219/packages/better-auth/src/index.ts#L6-L8

Are these types of zod exports necessary?

PR #4596, opened 2025-09-11
fix: set clientId in ProviderOptions to `unknown` by default

himself65:
Fixes: https://github.com/better-auth/better-auth/pull/4583/files#r2342548852
    

---

## Summary by cubic
Make clientId in ProviderOptions optional and typed as unknown. Require a string clientId in each social provider where applicable (TikTok remains no-clientId), improving type safety and aligning with provider specs.

- **Refactors**
  - ProviderOptions.clientId is now optional unknown with updated docs.
  - All providers now explicitly require clientId: string, except TikTok (clientId?: never).
  - Clarifies provider-specific typing without forcing a global string type.

- **Migration**
  - Ensure you pass clientId: string to all providers except TikTok.
  - For TikTok, remove any clientId usage.
  - Custom providers: set your desired clientId type or keep it as unknown if the provider varies.

### DIFF
--- a/packages/better-auth/src/oauth2/create-authorization-url.ts
+++ b/packages/better-auth/src/oauth2/create-authorization-url.ts
@@ -42,7 +42,14 @@ export async function createAuthorizationURL({
 }) {
 	const url = new URL(authorizationEndpoint);
 	url.searchParams.set("response_type", responseType || "code");
-	url.searchParams.set("client_id", options.clientId);
+	const clientId = options.clientId;
+	if (typeof clientId === "string") {
+		url.searchParams.set("client_id", clientId);
+	} else if (Array.isArray(clientId)) {
+		for (const id of clientId) {
+			url.searchParams.append("client_id", id);
+		}
+	}
 	url.searchParams.set("state", state);
 	url.searchParams.set("scope", scopes.join(scopeJoiner || " "));
 	url.searchParams.set("redirect_uri", options.redirectURI || redirectURI);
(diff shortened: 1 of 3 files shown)


packages/better-auth/src/oauth2/client-credentials-token.ts
```diff
@@ -37,7 +37,16 @@ export function createClientCredentialsTokenRequest({
...
-		body.set("client_id", options.clientId);
+		const clientId: unknown = options.clientId;
+		if (typeof clientId !== "undefined" && clientId !== null) {
+			if (typeof clientId === "string") {
+				body.set("client_id", clientId);
+			} else if (Array.isArray(clientId)) {
+				for (const id of clientId) {
+					body.append("client_id", String(id));
+				}
+			}
```
himself65, line 48: could you please tell me if the logic is correct?

/cc @RikhiSingh

RikhiSingh: Actually most OAuth2 servers will either:
   - Use only the first `client_id` and ignore the rest (whats been happening with Google Right now)
   - Reject the request as malformed (happened with me once using GItLab)
   - Return an error about duplicate parameters

Correct Fix:

```typescript
} else {
    const primaryClientId = Array.isArray(options.clientId) 
        ? options.clientId[0] 
        : options.clientId;
    body.set("client_id", primaryClientId);
    body.set("client_secret", options.clientSecret);
}
```


I think we should use `body.set()` instead of `body.append()` to ensure only one `client_id` parameter and use the first client ID from the array (consistent with Google provider behavior), it is only when we have Multiple client IDs which should only be used for **token verification**, not for making OAuth requests

himself65: ok, thanks for the feedback. woops I misclicked the merge... let me open another PR

RikhiSingh: 👍🏻, Let me know if you got any other queries regarding this :)

PR #4602, opened 2025-09-12
fix: correct MongoDB adapter import path in CLI

aajeeth-m:
The CLI was generating imports with `/mongo` instead of `/mongodb` for the MongoDB adapter, causing import errors. This fixes the import path to match the actual package structure.
    

---

## Summary by cubic
Fixes the CLI to generate the correct MongoDB adapter import path in auth config. Replaces better-auth/adapters/mongo with better-auth/adapters/mongodb to prevent import errors when selecting MongoDB.


aajeeth-m:
I'm sorry for the formatting noise - my IDE auto-formatted the file and I didn't notice before committing. This is my first contribution to open source. The actual fix is on line 428 where I corrected the import path to /mongodb

PR #4482, opened 2025-09-06
feat: support cross-domain cookie

HenriqueBragaMoreira:
This PR is based on the excellent work done by @RikhiSingh in PR #3030 
.
The implementation itself was already correct, but the related documentation contained some inconsistencies.

In this PR I’ve:

Kept the original implementation exactly as it was.

Fixed and clarified the documentation so it matches the behavior correctly.

All credits for the actual fix and implementation go to @RikhiSingh. I’ve only adjusted the docs to help get this merged and make it easier for others to use.

Note to maintainers:
Could you please review and validate this PR? Since the core logic was already resolved in the original contribution, this only aligns the docs with the actual behavior. Getting this merged would help the community move forward with a consistent reference. 🙏 @Bekacru @himself65 @Kinfe123 

Fixes #2962 
    

---

## Summary by cubic
Add cross-origin cookie support and enforce SameSite=None + Secure rules. Updates docs and tests to make cross-domain and subdomain setups safe and straightforward.

- **New Features**
  - advanced.crossOriginCookies with enabled, autoSecure (default true), and allowLocalhostUnsecure (default true).
  - Auto-sets Secure for SameSite=None (skips on localhost when allowed); supports Partitioned cookies.
  - Clear docs for cross-domain vs. cross-subdomain cookies, with examples, CORS/credentials guidance, and trustedOrigins usage.

- **Bug Fixes**
  - Throws when SameSite=None is used without Secure on non-localhost.
  - Applies __Secure- cookie prefix based on final merged attributes.
  - Merges defaultCookieAttributes and overrides correctly; adds tests covering all cases.

PR #4603, opened 2025-09-12
Revert "feat: support cross-domain cookie (#4482)"

himself65:
The issue here:

1. `partitioned` only works for some versions of Chrome, and it's not supported by any other browser. 3rd-party domain cookies are blocked from all modern browsers 
2. Imaging there are two servers on vercel.app and render.com, even render.com could set the cookie on vercel.app, but the Next.js server on vercel.app still cannot access the cookie. So this will lead server-side rendering issue. I thought e2e covers this, but it's not; it's okay for just a SPA app
3. api is confusing; the docs and examples might mislead people.

This reverts commit 61b6a874354cf080d4b6812c62346ae235b4c986.


himself65:
Landed at https://github.com/better-auth/better-auth/commit/22b78cfe7223374f80a50409167dbeadc890d416

PR #4604, opened 2025-09-12
docs: update cross domain cookie section

himself65:
## Summary by cubic
Updated the cookies docs with a new Cross-Domain Cookies section explaining that cookies cannot be shared across different domains in modern browsers. Added practical alternatives (rewrites/proxy, shared subdomains, SPA token auth, and auth providers) with clear warnings and guidance.

### DIFF
--- a/docs/content/docs/concepts/cookies.mdx
+++ b/docs/content/docs/concepts/cookies.mdx
@@ -81,6 +81,54 @@ export const auth = betterAuth({
     ],
 })
 ```
+### Cross-Domain Cookies (Different Domains)
+
+<Callout type="error">
+Cross-domain cookies (between completely different domains) are **not supported** and will not work reliably in modern browsers. This is a fundamental browser security limitation, not a framework limitation.
+</Callout>
+
+#### Why Cross-Domain Cookies Don't Work
+
+**1. Modern browsers block third-party cookies by default**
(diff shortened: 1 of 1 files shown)


docs/content/docs/concepts/cookies.mdx
```diff
@@ -81,6 +81,54 @@ export const auth = betterAuth({
...
+**3. Token-based authentication (SPA only)**
+- Store tokens in localStorage or sessionStorage
+- Send tokens in request headers instead of cookies
+- Note: This disables SSR and only works for client-side applications
+
+**4. Use authentication providers**
+- Consider using authentication services that handle cross-domain scenarios
+- These typically use OAuth flows or similar mechanisms designed for cross-domain authentication
+
+<Callout type="info">
```
Bekacru, line 128: I don't think we need this callout here

docs/content/docs/concepts/cookies.mdx
```diff
@@ -81,6 +81,54 @@ export const auth = betterAuth({
     ],
 })
 ```
+### Cross-Domain Cookies (Different Domains)
+
+<Callout type="error">
```
Bekacru, line 86: callout is uncessary just after a title...should be basic paragraph and change `#### Why Cross-Domain Cookies Don't Work` from title to a part of the paragraph and list the options then

PR #4612, opened 2025-09-12
fix: should infer types correctly when empty list of plugins is provided

frectonz:
Make sure the following infer the same types.

```typescript
const auth = betterAuth({
    plugins: []
});
```

and

```typescript
const auth = betterAuth({
});
```
    

---

## Summary by cubic
Ensure betterAuth({ plugins: [] }) infers the same types as betterAuth({}). This removes inconsistent $Infer results when the plugins array is empty.

- **Bug Fixes**
  - Return {} from InferFieldsFromPlugins when plugins is [].
  - Narrow BetterAuthOptions.plugins to [] | BetterAuthPlugin[] to preserve empty-array literal.
  - Add a type test to confirm both configurations produce equal types.

### DIFF
--- a/packages/better-auth/src/test-utils/index.ts
+++ b/packages/better-auth/src/test-utils/index.ts
@@ -93,7 +93,7 @@ export async function getTestInstanceMemory<
 			...options?.advanced,
 		},
 		plugins: [bearer(), ...(options?.plugins || [])],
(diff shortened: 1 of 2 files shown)


frectonz:
The test errors don't seem to be related to the changes in the PR.

PR #4617, opened 2025-09-12
fix(expo): missing peer deps

himself65:
## Summary by cubic
Declare missing Expo peer dependencies for the expo package and set a stable peer range for better-auth to avoid install/runtime warnings and ensure correct app setup.

- **Dependencies**
  - Added peerDependencies: expo-constants >=17, expo-crypto >=13, expo-linking >=7, expo-secure-store >=14, expo-web-browser >=14
  - Updated better-auth peer to >=1.0.0 (from workspace:*)

### DIFF
--- a/packages/expo/package.json
+++ b/packages/expo/package.json
@@ -61,7 +61,12 @@
     "unbuild": "^3.6.1"
   },
   "peerDependencies": {
-    "better-auth": "workspace:*"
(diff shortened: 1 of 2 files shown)


juan-altatech:
ummm... don't we need to install @better-auth/expo in the backend as well? why all the react native peer dependencies?

PR #4622, opened 2025-09-13
fix(cli): timestamp in schema for Drizzle with SQLite

zy1p:
fix: https://github.com/better-auth/better-auth/pull/4889

`.defaultNow()` which is `(cast((julianday('now') - 2440587.5)*86400000 as integer))` was storing timestamp in subsecond precision. 

refers to [implementation](https://github.com/drizzle-team/drizzle-orm/blob/8e8a9e902410ed2068a015d59180b225045dddf1/drizzle-orm/src/sqlite-core/columns/integer.ts#L136), retrieving timestamp will get wrong result.

using `(current_timestamp)` was storing text into integer field, use `(cast(unixepoch('subsecond') * 1000 as integer)` instead. see https://github.com/drizzle-team/drizzle-orm/issues/2810

image

image

image

### DIFF
--- a/packages/cli/src/generators/drizzle.ts
+++ b/packages/cli/src/generators/drizzle.ts
@@ -104,7 +104,7 @@ export const generateDrizzleSchema: SchemaGenerator = async ({
 						: `int('${name}')`,
 				},
 				date: {
-					sqlite: `integer('${name}', { mode: 'timestamp' })`,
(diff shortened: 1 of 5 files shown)


himself65:
Oh, I think we did some force push on Canary and messed up the PR

zy1p:
Now should be fine, I have force pushed wrong branch.

PR #4650, opened 2025-09-14
fix: respect `errorCallbackURL` in failed oauth flows

### LINKED ISSUE #1580: `errorCallbackURL` is not working it is ignored on error

- [x] #1581

### To Reproduce

Image

most probably this is due to callback api route and in my case code is not generated as i cancel the process to choose another account 

logger do print access_denied and error_description=undefined but it do not redirect it 

Current : 

it is ignoring errorCallbackURL in social signing process even if it is passed and redirecting it to api default error page

expected
it should redirect to the callbackurl passed in social loging 
```typescript
await authClient.signIn.social(
	{
		provider: "google",
		callbackURL: "/dashboard",
		errorCallbackURL: "/sign-in",
	},
	{
		onRequest: () => {
			setIsOAuthLogin(true);
		},
		onError: (ctx) => {
			setIsOAuthLogin(false);
			toast.error(ctx.error.message);
		},
	},
);
``` 
in my case it should redirect it to signing page with queries code and description as designed

1.1.21

```bash
- OS : Windows
- Browser : Chrome
```

Client

```typescript
	appName: "Notespace",
		provider: "pg",
		schema: {
			...schema,
			user: schema.UserTable,
			account: schema.ConnectedAuthProvidersTable,
			session: schema.UserAuthSessionTable,
			verification: schema.UserAuthVerificationTable,
		},
	}),
	emailAndPassword: {
		enabled: true,
		autoSignIn: true,
	},
	emailVerification: {
		sendOnSignUp: true,
		sendVerificationEmail: async ({ user, url }) => {
			await sendEmail({
				to: user.email,
				subject: "Verify your email address",
				text: `Click the link to verify your email: ${url}`,
			});
		},
	},
	socialProviders: {
		google: {
		},
	},
	plugins: [openAPI()],
} satisfies BetterAuthOptions);
```

### Additional context

Images for reference :

Image

Image

frectonz:
closes #1580
    

---

## Summary by cubic
Respect the errorCallbackURL for failed OAuth callbacks so users land on the app’s custom error page. Also forwards error_description when provided.

- **Bug Fixes**
  - Use errorCallbackURL from parsed state for provider errors, missing state, and missing code.
  - Centralize redirects via redirectOnError; builds URLs with URLSearchParams and preserves existing query strings, including error_description.
  - Reorder checks to parse state first so the custom error URL is available; keeps error logging.

### DIFF
--- a/packages/better-auth/src/api/routes/callback.ts
+++ b/packages/better-auth/src/api/routes/callback.ts
@@ -43,6 +43,13 @@ export const callbackOAuth = createAuthEndpoint(
 
 		const { code, error, state, error_description, device_id } = queryOrBody;
 
+		if (!state) {
+			c.context.logger.error("State not found", error);
+			const sep = defaultErrorURL.includes("?") ? "&" : "?";
+			const url = `${defaultErrorURL}${sep}state=state_not_found`;
+			throw c.redirect(url);
+		}
+
 		const {
 			codeVerifier,
 			callbackURL,
@@ -66,11 +73,6 @@ export const callbackOAuth = createAuthEndpoint(
 
 		if (error) {
 			redirectOnError(error, error_description);
-		}
-
-		if (!state) {
-			c.context.logger.error("State not found", error);
-			redirectOnError("state_not_found");
 		}
 
 		if (!code) {


packages/better-auth/src/api/routes/callback.ts
```diff
@@ -43,16 +43,6 @@ export const callbackOAuth = createAuthEndpoint(
 
 		const { code, error, state, error_description, device_id } = queryOrBody;
 
-		if (error) {
-			throw c.redirect(
-				`${defaultErrorURL}?error=${error}&error_description=${error_description}`,
-			);
-		}
-
-		if (!state) {
```
Bekacru, line 52: if there is no state the parse state will fail

frectonz: fixed

frectonz:
We don't really have a way to test an oauth callback endpoint now, but I will figure something out.

PR #4663, opened 2025-09-15
test: fix flaky

himself65:
## Summary by cubic
Stabilizes Prisma adapter tests by isolating each run with its own SQLite DB and making schema pushes synchronous to eliminate race conditions and data leakage.

- **Refactors**
  - Each test uses a copied dev.db (crypto UUID) with datasourceUrl override.
  - getAdapter is now async and returns { adapter, clear } for cleanup.
  - Switched to beforeEach to push schema and reset DB per suite.
  - Replaced async exec with execSync in pushPrismaSchema.
  - runAdapterTest/runNumberIdAdapterTest now non-async wrappers.

- **Bug Fixes**
  - Removed cross-suite file-based state and watchers; no shared DB state.
  - Tests no longer depend on unique email prefixes; fresh DB prevents collisions.
  - Adjusted assertions to avoid fixed counts; created entities per test.
  - Reset sqlite_sequence where required in number-id tests.
  - Dropped the brittle “increment id by 1” assertion that caused flakes.

### DIFF
--- a/packages/better-auth/package.json
+++ b/packages/better-auth/package.json
@@ -28,10 +28,7 @@
     "dev:types": "tsc --project tsconfig.declarations.json --watch",
     "build:types": "tsc --project tsconfig.declarations.json",
     "stub": "unbuild --stub",
-    "pretest": "pnpm prisma:normal:push && pnpm prisma:number-id:push",
     "test": "vitest",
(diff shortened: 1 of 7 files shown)


himself65:
I think this not gonna work. closing

PR #4680, opened 2025-09-15
fix(plugins): asynchronous `init`

LightTab2:
I've run into issue when implementing a custom plugin: `init` cannot be `async`.

I believe there's no harm in allowing that.
    

---

## Summary by cubic
Allow plugin init to be asynchronous, enabling async setup without breaking existing synchronous plugins. runPluginInit is now async and awaited in init; BetterAuthPlugin.init types accept Promise-based returns for context/options (or void).

### DIFF
--- a/packages/better-auth/src/init.test.ts
+++ b/packages/better-auth/src/init.test.ts
@@ -197,4 +197,26 @@ describe("init", async () => {
 		const githubProvider = ctx.socialProviders.find((p) => p.id === "github");
 		expect(githubProvider).toBeDefined();
 	});
+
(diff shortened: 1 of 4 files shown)


himself65 (approved):

LightTab2:
@himself65 After merging 'canary' into this branch, docker containers did not fail this time. Cool test case and cleanup, btw

himself65:
> @himself65 After merging 'canary' into this branch, docker containers did not fail this time. Cool test case and cleanup, btw

This is github issue, you don't need to worry about

PR #4683, opened 2025-09-15
fix(stripe): webhook signature

himself65:
Fixes: https://github.com/simstudioai/sim/pull/1336


himself65:
I'm setting up the e2e test case

himself65:
I think we don't need change anything. It maybe just misconfig

PR #4685, opened 2025-09-15
docs: update siwe.mdx

diogoribeirodev:
docs(siwe): clarify chainId parameter requirements and add chain examples

- Fix misleading "optional" description for chainId parameter
- Use conditional format: "optional for Ethereum mainnet, required for other chains"
- Add chain-specific examples (Polygon, Arbitrum, Base) showing required chainId values
- Add warning callout explaining chainId must match SIWE message Chain ID
- Prevent 401 verification errors from chainId mismatches that confused developers

The chainId parameter was documented as "optional" but actually required for all non-mainnet chains, causing verification failures. Now clearly explains when it's optional vs required.
    

---

## Summary by cubic
Clarifies SIWE chainId docs: optional on Ethereum mainnet (defaults to 1), required on other chains, and must match the Chain ID in the SIWE message. Adds examples for Polygon (137), Arbitrum (42161), and Base (8453), plus a warning to prevent 401 verification errors from mismatches.

### DIFF
--- a/docs/content/docs/plugins/siwe.mdx
+++ b/docs/content/docs/plugins/siwe.mdx
@@ -89,7 +89,7 @@ Before signing a SIWE message, you need to generate a nonce for the wallet addre
 ```ts title="generate-nonce.ts"
 const { data, error } = await authClient.siwe.nonce({
   walletAddress: "0x1234567890abcdef1234567890abcdef12345678",
-  chainId: 1, // optional, defaults to 1 (Ethereum mainnet)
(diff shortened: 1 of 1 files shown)


diogoribeirodev:
In the current implementation, if we dont specify a chain id, it will default to 1, since its an optional field the user won't initally think about using it ( aka me ), and the currently if the chainid is different from the defaultm it returns a 401, saying the nonce is either invalid or expired, nowhere saying that the chainids are misaligned, the docs are technically correct, but are a bit misleading, i think this is way its clearer for the user to know how to use the actual parameter

himself65 (approved):

PR #4703, opened 2025-09-16
fix(api-key): cascade api keys on user deletion

ping-maxwell:
When a user is about to be deleted, you run into an SQL error saying that the user has a foreign key relation to an API key, thus preventing the user from being deleted. This PR adds an `onDelete`  effect to ensure the api keys are deleted as users get deleted.

https://discord.com/channels/1288403910284935179/1417382336315654267


himself65 (approved):
LGTM

PR #4705, opened 2025-09-16
docs: updated autumn plugin page with latest org feature

johnyeocx:
## Summary by cubic
Updated the Autumn plugin docs to cover organization and mixed customer scopes with tabbed setup examples. Also fixed API examples to use featureId and clarified how customers are identified.

- **New Features**
  - Added tabbed examples for User, Organization, User & Organization, and Custom identify flows.
  - Updated callout to explain choosing the customer entity (user, org, both, or custom).
  - Corrected API examples to use featureId in check and track calls.

- **Dependencies**
  - Added: @better-auth/utils@0.3.0, @radix-ui/react-icons, framer-motion, geist, react-use-measure, mini-svg-data-uri.


docs/package.json
himself65, line 1: Why do we need these changes?

PR #4706, opened 2025-09-16
feat: inject db transaction by default

himself65:


himself65:
Closed, see https://github.com/better-auth/better-auth/pull/4711

PR #4713, opened 2025-09-16
fix(adapter): returning null as string for optional id references

jslno:
## Summary by cubic
Prevents optional id fields from being cast to the string "null" in the adapter factory. Now only defined, non-null ids are stringified; null remains null.


himself65 (changes requested):
could you please add a test case?

himself65 (approved):
LGTM

PR #4724, opened 2025-09-17
fix: support for organizations with large member counts

Badbird5907:
Hi,
My app includes an organization with approximately 6,000 members. We've encountered an issue where the `activeOrganizationId` field is set to `null` after calling `getFullOrganization`. More details can be found in the [Discord thread](https://canary.discord.com/channels/1288403910284935179/1417776115069485208).

The current implementation of `getFullOrganization` only loads the first 100 members, checking if the logged-in user is among them. This approach leads to undefined behavior in organizations with more than 100 members. If the member was not found in the 100 members loaded, it would update their session to clear `activeOrganizationId`.

This issue was especially hard to pinpoint as my app did not call `getFullOrganization` directly, but it was indirectly called by other functions (See this [Discord message](https://canary.discord.com/channels/1288403910284935179/1296058482289676320/1417935886267449425)).

I also removed a redundant sanity check that was checked above.

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-org.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-org.ts
@@ -709,7 +709,7 @@ export const getFullOrganization = <O extends OrganizationOptions>(
 			}
 			const isMember = await adapter.checkMembership({
 				userId: session.user.id,
-				organizationId: organizationId,
+				organizationId: organization.id,
 			});
 			if (!isMember) {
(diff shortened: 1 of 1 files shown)


Badbird5907:
I also removed a redundant sanity check that was checked above.

Badbird5907:
I was able to verify this patch works in my app using `pnpm patch`.

Bekacru (approved):

PR #4727, opened 2025-09-17
Fix generic OAuth discovery when providers skip JSON headers

Karavil:
Fixes generic OAuth discovery to handle non-compliant providers like Clever that return 200 status with incorrect Content-Type headers.

**Problem**: Clever's discovery endpoint returns `application/x-www-form-urlencoded` instead of `application/json`, causing betterFetch to return a Blob and breaking the plugin with INVALID_OAUTH_CONFIGURATION.

**Solution**: Replace betterFetch with a robust fetchDiscoveryDocument helper that:
- Uses bare fetch to avoid content-type assumptions
- Validates responses with Zod schema
- Logs RFC 8414 §3.1 warnings but continues with best-effort parsing
- Maintains the same `{ data, error }` API shape for backward compatibility

**Testing**: Added test coverage for both compliant providers and Clever's non-compliant response format using a lightweight discovery server helper.

Compliant providers work exactly as before, while non-compliant ones now succeed without requiring hardcoded endpoints in downstream apps.

### DIFF
--- a/packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
@@ -1,3 +1,6 @@
+import { createServer } from "node:http";
+import { AddressInfo } from "node:net";
+
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
 import { genericOAuth } from ".";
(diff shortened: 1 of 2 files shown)


himself65:

himself65:

PR #4735, opened 2025-09-17
feat: add ai chat assistant for better-auth docs

Kinfe123:


docs/app/api/ai-chat/route.ts
```diff
@@ -0,0 +1,86 @@
+import { NextRequest, NextResponse } from "next/server";
+
+export async function POST(request: NextRequest) {
+	try {
+		const body = await request.json();
```
himself65, line 5: We can use our own `better-fetch`. This is a good chance to dog fooding it

docs/components/markdown-renderer.tsx
himself65, line 1: react-markdown has not been update for 2 years. Let's follow fumadocs impl https://github.com/fuma-nama/fumadocs/blob/ec8250eae29f7f4bc258e0f31249bee6fa5266e7/apps/docs/components/ai/markdown.tsx#L64

docs/components/ai-chat-modal.tsx
```diff
@@ -0,0 +1,389 @@
...
+	timestamp: Date;
+	isStreaming?: boolean;
+}
+
+interface AIChatModalProps {
+	isOpen: boolean;
+	onClose: () => void;
+}
+
+export function AIChatModal({ isOpen, onClose }: AIChatModalProps) {
```
himself65, line 33: I think AIChatModal should use a top-level open state to avoid having two modals in the UI.

We have jotai, right? So let's put isOpen to a jotai atom

docs/components/markdown-renderer.tsx
antonvishal: There's a alternative from vercel too
https://streamdown.ai/

Kinfe123: @himself65 The thing is we are already using it for the other rendering as well like changelogs , blogs or other. I just copied from there. So it is better to do it on follow up pr on replacing it.

docs/components/ai-chat-modal.tsx
Kinfe123: right now we are modal overlap safe on this implementation but you are right on using it.

himself65 (approved):

himself65 (changes requested):
It's not working on preview

Image

Kinfe123:
It is probably the credential issue. May be the env is not propagating as it should I should check on it

himself65 (approved):
I'm gonna merge this to see if it works on canary

PR #4740, opened 2025-09-18
docs: email in passkey is not required

himself65:


Bekacru:
should be against main

PR #4764, opened 2025-09-19
feat(sso): Multi Tenant - use the provider domain for redirect and tokenResponse

KaiSoellch:
# Feature for SSO Plugin
Optional use of provider (authorization) domain for redirect and token response endpoints to support white‑label / custom CNAME auth domains.

Problem Currently the library always builds redirect (authorization) and token/introspection/revocation endpoints from the baseURL. For SaaS tenants using custom CNAME (white‑label) domains this breaks auth flows when the identity provider expects the provider (issuer) domain instead of the tenant vanity domain. This limits multi-tenant + custom domain adoption.

Summary Introduce a configurable option to resolve redirect and token-related endpoints against the provider (issuer) domain rather than the tenant (public) domain.
This allows:
- White‑label tenant domains for end-user UX
- correct redirecting to the provider domain for custom styles and settings

## New Features
Added option: **useProviderDomain** to switch construction of:
- Authorization / redirect URL
- Token endpoint URL

## Fallback
default behavior unchanged when option is absent or false.

## Backward Compatibility
Fully backward compatible: existing integrations continue without changes.
No breaking type changes; option is additive.
    

---

## Summary by cubic
Adds an option to build SSO redirect and token endpoints using the provider’s domain. This fixes auth flows for white‑label/custom CNAME tenants without changing defaults.

- **New Features**
  - Added SSOOptions.useProviderDomain (default false) to resolve the authorization URL, callback URI, and token requests against https://{provider.domain}/api/auth.
  - Default behavior remains unchanged when not set.

### DIFF
--- a/packages/better-auth/src/oauth2/state.ts
+++ b/packages/better-auth/src/oauth2/state.ts
@@ -96,6 +96,28 @@ export async function parseState(c: GenericEndpointContext) {
 	if (!parsedData.errorURL) {
 		parsedData.errorURL = `${c.context.baseURL}/error`;
 	}
+	const stateCookie = c.context.createAuthCookie("state");
+	const stateCookieValue = await c.getSignedCookie(
+		stateCookie.name,
+		c.context.secret,
+	);
+	/**
+	 * This is generally cause security issue and should only be used in
+	 * dev or staging environments. It's currently used by the oauth-proxy
+	 * plugin
+	 */
+	const skipStateCookieCheck = c.context.oauthConfig?.skipStateCookieCheck;
+	if (
+		!skipStateCookieCheck &&
+		(!stateCookieValue || stateCookieValue !== state)
+	) {
+		const errorURL =
+			c.context.options.onAPIError?.errorURL || `${c.context.baseURL}/error`;
+		throw c.redirect(`${errorURL}?error=state_mismatch`);
+	}
+	c.setCookie(stateCookie.name, "", {
+		maxAge: 0,
+	});
 	if (parsedData.expiresAt < Date.now()) {
 		await c.context.internalAdapter.deleteVerificationValue(data.id);
 		const errorURL =


KaiSoellch:
i need to rework some stuff - made some mistakes

FacundoLucci:
Hey, thanks for working on this! 

(I used AI to help me write the comment below.)

We've been running a multi-tenant SaaS with Better Auth 
where organizations have their own subdomains (`cust-org.saas.app`) and some have custom 
domains (`customer.com`). Ran into quite a few challenges getting OAuth to work across 
all of them, so wanted to share what we ended up doing in case any of it's useful.

### What we ran into

**Cookie sharing across subdomains**

When OAuth completed on one subdomain, we had trouble getting the session to be 
recognized on other subdomains. We ended up setting cookies with the parent domain:

setCookie(c, 'better-auth.session-token', token, {
  domain: '.saas.app',
  // ...
})Not sure if this is the "right" way to handle it, but it worked for our subdomain case.

**Custom domains (the tricky part)**

For custom domains like `customer.com`, we couldn't figure out how to share cookies 
with our main domain. We ended up building a token exchange flow:

1. All OAuth callbacks go to our single backend (`backend.saas.app`)
2. After auth, we redirect to the custom domain with a short-lived token in the URL
3. The custom domain frontend exchanges that token for a session cookie

This let us keep a single redirect URI registered with Google. There might be a cleaner 
way to do this that we missed!

**Organization context in sessions**

We wanted sessions to automatically know which organization they belonged to based on 
which domain initiated the OAuth. We ended up manually setting `activeOrganizationId` 
when creating sessions, but it felt like we might have been fighting the library a bit.

**Session lookup after creation**

We had some issues where `auth.api.getSession()` wasn't finding sessions right after 
we created them. Ended up falling back to direct database queries in some places. 
Could have been something wrong with how we were creating sessions though.

---

### Questions I'd love to understand better

1. Does this PR handle cookie domains automatically, or is that something we'd still 
   need to configure separately?

2. For custom domains that aren't subdomains of the base URL - does `useProviderDomain` 
   handle that case, or would we still need a token exchange pattern?

3. Is there a way to automatically populate `activeOrganizationId` on sessions based 
   on the SSO provider's organization?

Really appreciate the work on this - multi-tenant OAuth is surprisingly complex and 
having native support would be amazing. Happy to share more details about our setup 
if any of it would be helpful!

KaiSoellch:
@FacundoLucci thanks for this insights. my approach was deprecated after the 1.4.0 release where the sso plugin got overhauled with domain verification. we migrated to this approach. so i close this pr

PR #4768, opened 2025-09-19
fix: wrap `Math.floor` around the division when calculating TTL

DevDuki:
For more context see this thread in discord:
https://discord.com/channels/1288403910284935179/1418569936204468255


ping-maxwell:
@DevDuki can you run lint:fix?

himself65 (changes requested):
What's the difference after lint?

Issue #4767, opened 2025-09-19
The data in `databaseHooks.session.update.before` is limited compared to the `.create.before` version


SkepticMystic:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Add the following databaseHooks:
```ts
      databaseHooks: {
        session: {
          create: {
            before: async (session, ctx) => {
              console.log(session, ctx, "session.create.before");
            },
          },

          update: {
            before: async (session, ctx) => {
              console.log(session, ctx, "session.update.before");
            },
          },
        },
      },
```
2. Signin to create a new session, view logs
3. Do something to _update_ the current session, like `authClient.organization.setActive`, view logs

### Current vs. Expected behavior

The data available on the `session` object in the `create` hook has far more fields available. The `update` hook only has `activeOrganizationId` for me

### What version of Better Auth are you using?

1.3.12

### System info

```bash
Nothing gets output when I run this...
```

### Which area(s) are affected? (Select all that apply)

Backend

### Additional context

Initial Discord support thread: https://discord.com/channels/1288403910284935179/1415281222304661534

SkepticMystic:
Looking at [this code](https://github.com/better-auth/better-auth/blob/a208c0989455cba0843cd34679618c457fe05c62/packages/better-auth/src/db/internal-adapter.ts#L480), I understand why. It's a patch, not a full update.

The broader problem in my use case is that I need other info about the current session, like userId. But when setting active org, the only info available in the session _patch_ is the new orgId. So I can't tell which user is being update (in the context of the hook).
I would usually use the escape hatch of `ctx` in the hook, but it's `undefined` in this case

SkepticMystic:
Indeed, the org plugin doesn't pass `ctx`: https://github.com/better-auth/better-auth/blob/a208c0989455cba0843cd34679618c457fe05c62/packages/better-auth/src/plugins/organization/adapter.ts#L378

SkepticMystic:
I see two possible solutions for my use case:
- The org plugin can pass `ctx` into the updateSession call
- The internal adapter could pass the sessionToken with the `session` patch, which I could use to lookup the full object: https://github.com/better-auth/better-auth/blob/a208c0989455cba0843cd34679618c457fe05c62/packages/better-auth/src/db/internal-adapter.ts#L480

ping-maxwell:
> The org plugin can pass ctx into the updateSession call

We should do this

ping-maxwell:
[...]

PR #4769, opened 2025-09-19
fix(organization): pass `ctx` to DB hooks

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/4767
    

---

## Summary by cubic
Pass the request context (ctx) to session updates in the organization plugin so DB hooks run with the correct context. Fixes missing context in setActiveOrganization/setActiveTeam flows and closes better-auth/better-auth#4767.

- **Bug Fixes**
  - Adapter: setActiveOrganization and setActiveTeam now accept ctx and pass it to internalAdapter.updateSession.
  - Routes: All call sites updated to forward ctx (create/delete org, set active org/team, invitations, member removal/leave).
  - Ensures DB hooks receive tenant/audit context and execute as expected.

### DIFF
--- a/packages/better-auth/src/plugins/organization/adapter.ts
+++ b/packages/better-auth/src/plugins/organization/adapter.ts
@@ -1,4 +1,4 @@
-import type { Session, User } from "../../types";
+import type { GenericEndpointContext, Session, User } from "../../types";
 import { getDate } from "../../utils/date";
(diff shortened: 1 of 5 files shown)


himself65 (approved):
LGTM, in the next version we should use async local storage to automatically share the ctx

PR #4163, opened 2025-08-22
feat(oauth-provider): an oauth 2.1 compliant plugin

dvanmali:
# oAuth 2.1 Provider

An upgrade to oidc-provider plugin that makes it oauth2.1 compliant and has a configuration that is secure by default.

Plans for the deprecation of oidc-provider plugin due to many inherent flaws in its design. Internally, plugin functions now share logic, providing for better future extensibility if new `code_grants` need to be written or user/client jwt or opaque tokens need to be written. Furthermore, as an oAuth 2.1 provider, it provides logic valid for an MCP server. When using the scope "openid" (optional, enabled by default), the server acts like an OpenId server able to issue id tokens and provides a `/userinfo` endpoint.

**Features**

- OAuth 2.1 by default
    - Properly supports authorization_code, refresh_token, and client_credentials grants
    - PKCE by default (removes `plain` completely)
- Public and confidential client registration
- JWT plugin is required by default, but can be disabled using `disableJWTPlugin` flag
- Access tokens can now be received in JWT verifiable format using the `resource` parameter (ie JWT `aud` field)
- Id tokens are still verifiable by JWKS when using JWT Plugin, or clientSecret if disabled. Fixes issue to prevent public clients when `disableJWTPlugin: true` from obtaining id tokens directly even when they shouldn't be allowed an id token and should use `/userinfo` instead.
- Protects `/userinfo` with scope check
- Separates Refresh Token and Access token on database schema to allow multiple access tokens per refresh and multiple refresh tokens per login session.
- `oauthAccessToken` strictly deals with opaque tokens
    - Opaque tokens are given only when resource parameter (aka audience) is not provided
- Option to Encode and Decode refresh tokens
- `allowDynamicClientRegistration` with `allowUnauthenticatedClientRegistration` flags
- Separation of default expiration times
- Proper creation of public and confidential clients
- Prevents misconfiguration between `.well-known/openid-configuration` endpoint and plugin settings
- `scopeExpirations` to assign scopes specific expiration
- Custom claims through separated functions: `customAccessTokenClaims`, `customIdTokenClaims`, and `customUserInfoClaims`
- Organizational support through `activeOrganizationalId` on a session such as through the `organizational` plugin. Attaches to oAuthClient via `reference_id`.
- Rp-initiated logout
- Account Selection via `prompt=select_account`.
- Account Creation via `prompt=create`.
- Prompt combinations `prompt=select_account+consent` and `prompt=login+consent`
[...]

PR #4791, opened 2025-09-20
feat(oidc): canUserRegisterClient option

TheUntraceable:
## Summary by cubic
Adds canUserRegisterClient to the OIDC provider to control who can register new clients, and enforces authentication for dynamic client registration. Docs updated with examples and guidance.

- **New Features**
  - New OIDC option: canUserRegisterClient(user) => boolean | Promise<boolean> to allow/deny client registration per user.
  - Registration now validates: authenticated session, allowDynamicClientRegistration=true, and optional user permission.

- **Migration**
  - Dynamic client registration now requires an authenticated session; unauthenticated requests return UNAUTHORIZED.
  - If you need role-based control, set canUserRegisterClient; otherwise omit it to allow all authenticated users.

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/types.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/types.ts
@@ -161,7 +161,7 @@ export interface OIDCOptions {
 	 */
 	schema?: InferOptionSchema<typeof schema>;
 	/**
-	 * Function to check if a user is allowed to create an application.
+	 * Function to check if a user is allowed to create a client.
 	 *
 	 * This is called when a user tries to register a new client via the dynamic
 	 * client registration endpoint.


packages/better-auth/src/plugins/oidc-provider/index.ts
```diff
@@ -1301,14 +1301,31 @@ export const oidcProvider = (options: OIDCOptions) => {
 					const session = await getSessionFromCtx(ctx);
 
 					// Check authorization
-					if (!session && !options.allowDynamicClientRegistration) {
```
dvanmali: This line of code allows the creation of public clients using `dynamicClientRegistration`. The structure is intensional even though the `dynamic` might not be the best variable name here.

dvanmali (changes requested):

dvanmali:
Great functionality! Could you add tests please?

packages/better-auth/src/plugins/oidc-provider/index.ts
TheUntraceable: So users should be able to create a client even when unauthenticated?

TheUntraceable:
> Great functionality! Could you add tests please?

Sure

packages/better-auth/src/plugins/oidc-provider/index.ts
dvanmali: According to how it is currently setup that's how this line works. To fully remove the endpoint one should disable the path using `disablePath`.

I currently have an oauth rewrite #4163 which will separate this variable into `allowDynamicRegistration` and `allowUnauthenticatedDynamicRegistration`.

docs/content/docs/plugins/oidc-provider.mdx
```diff
@@ -377,6 +380,21 @@ const auth = betterAuth({
...
 ```
 
+#### Controlling Client Registration
+You can control which users can register new clients by providing a `canUserRegisterClient` function. This function receives the user object and should return `true` if the user is allowed to register new clients, or `false` otherwise.
+
+```ts title="auth.ts"
+const auth = betterAuth({
+    plugins: [oidcProvider({
+        allowDynamicClientRegistration: true,
+        canUserRegisterClient: (user) => {
```
Bekacru, line 390: this should probably be under `disableRegistration` flag which can be a `boolean` or a callback function

Bekacru (changes requested):

TheUntraceable:
Closing this in favour of the new OAuth2.1 plugin which deprecates the OIDC Provider plugin

Issue #4766, opened 2025-09-19
Database hook delete


benjaminpreiss:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

I am building an integration of better-auth into payload. This works by letting better-auth handle the session and syncing the users from better-auth to payload (upon create and delete).

### Describe the solution you'd like

For the sync I am using database hooks, as I thought this to be the most robust approach as it is not bypassable. Turns out that database hooks don't support delete, only create and update?

It would be great to have a delete hook available as well

### Describe alternatives you've considered

I could instead use regular hooks, but I was afraid it might not be as robust as database hooks / that there might be ways that someone interacting with my better-auth instance might bypass the hooks. Curious what you think though.

himself65:
I think the feature makes sense. I'm +1 on this. What do you think @Bekacru

Bekacru:
yeah we should have delete hooks

gregoiregentil:
I need to do the same ideally in the database hook. As a temporary solution, can anyone post a robust hook to catch the delete user scenario?

PR #4792, opened 2025-09-20
feat(db): delete hooks

Kinfe123:
Fixes: https://github.com/better-auth/better-auth/issues/4766
    

## Summary by cubic
Add database delete hooks (before/after) for user, session, account, and verification. All delete paths now run through hooks, allowing deletions to be aborted and letting you run side effects with access to the endpoint context.

- New Features
  - Added databaseHooks.<model>.delete.before/after with entity payload and optional context.
  - before can return false to cancel deletion; after runs post-delete.
  - Docs updated with user/session examples; tests cover success and abort flows.

- Refactors
  - Introduced deleteWithHooks and deleteManyWithHooks and wired them into internalAdapter.
  - Updated delete user/session/verification flows to pass context and trigger hooks.

### DIFF
--- a/packages/better-auth/package.json
+++ b/packages/better-auth/package.json
@@ -1,6 +1,6 @@
 {
   "name": "better-auth",
-  "version": "1.4.0-beta.1",
+  "version": "1.4.0-beta.2",
   "description": "The most comprehensive authentication library for TypeScript.",
   "type": "module",
   "license": "MIT",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@better-auth/cli",
-  "version": "1.4.0-beta.1",
+  "version": "1.4.0-beta.2",
   "description": "The CLI for Better Auth",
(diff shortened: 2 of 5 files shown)


packages/better-auth/src/db/with-hooks.ts
```diff
@@ -160,9 +160,126 @@ export function getWithHooks(
...
+		customDeleteFn?: {
+			fn: (where: Where[]) => void | Promise<any>;
+			executeMainFn?: boolean;
+		},
+		context?: GenericEndpointContext,
+	) {
+		let entityToDelete: T | null = null;
+
+		try {
+			const entities = await (await getCurrentAdapter(adapter)).findMany<T>({
```
TheUntraceable, line 176: I suppose instead of having this ugly syntax, to make it more readable put `getCurrentAdapter(adapter)` into it's own variable for code readability?

maelp:
I guess it would be easy from there to implement delete as a soft-delete right? eg in the before we add a deleted_at field and then refuse to delete ?

PR #4794, opened 2025-09-21
feat(sso): extend ssoProvider schema with customizable fields and add…

KaiSoellch:
…itionalFields support

### DIFF
--- a/packages/better-auth/src/plugins/sso/index.ts
+++ b/packages/better-auth/src/plugins/sso/index.ts
@@ -1,6 +1,7 @@
 import * as z from "zod";
 import { APIError, createAuthEndpoint, sessionMiddleware } from "../../api";
 import type { BetterAuthPlugin, User } from "../../types";
+import type { FieldAttribute, InferAdditionalFieldsFromPluginOptions } from "../../db/field";
 import {
 	createAuthorizationURL,
 	generateState,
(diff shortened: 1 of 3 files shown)


TheUntraceable (approved):

dvanmali (approved):

KaiSoellch:
its really hard to get a successful run against your canary, but i try

packages/better-auth/src/plugins/sso/index.ts
```diff
@@ -1,6 +1,7 @@
 import * as z from "zod";
 import { APIError, createAuthEndpoint, sessionMiddleware } from "../../api";
-import type { BetterAuthPlugin, User } from "../../types";
```
xiaoyu2er: should we make changes to the new package `packages/sso` instead of `packages/better-auth/plugins/sso`?

see https://www.better-auth.com/blog/1-3#sso-plugin

https://discord.com/channels/1288403910284935179/1296058482289676320/1423337098622144562

> what's the difference between better-auth/plugins/sso and better-auth/sso

> The main difference is that the SSO functionality is now provided as a standalone package in the dedicated "@better-auth/sso" module rather than as a submodule under "better-auth/plugins/sso". The dedicated package offers support for both OIDC and SAML 2.0 and is the recommended approach for new projects, while the plugins path was used in earlier or legacy integrations. This change helps separate SSO concerns from core plugins, making it easier to update and maintain its features

KaiSoellch: u r right. i moved it to the new package

PR #4809, opened 2025-09-22
refactor(admin): default value for both limit and offset and always return them 

hieudien14310:
## Summary by cubic
Set default pagination for the admin list users endpoint (limit=10, offset=0) and always include limit/offset in the response. Also improved OpenAPI docs and simplified parameter handling.

- **Refactors**
  - Added defaults: limit=10, offset=0; both accept string or number.
  - Always return offset and limit in success and error responses.
  - Short-circuit with empty result when limit <= 0 (skips DB call).
  - Pass numeric limit/offset to the adapter.
  - Reordered Zod .meta/.optional to ensure OpenAPI descriptions render.


himself65 (changes requested):
Could you please add a test case?

hieudien14310:
> Could you please add a test case?

Done. Please review again

PR #4813, opened 2025-09-22
fix: prevent TypeError when errorWithStack is undefined in to-auth-en…

NalinDalal:
…dpoints
    

---

## Summary by cubic
Fixes a crash in to-auth-endpoints when debug logging tried to read errorWithStack.stack on APIError. We now guard access with optional chaining so TypeError no longer occurs when errorWithStack is missing.

- **Bug Fixes**
  - Safely access errorWithStack?.stack in endpoint execution and before/after hooks.
  - Preserves stack logging when available; otherwise proceeds without a stack. No API or response changes.


himself65:
please run lint

NalinDalal:
bro i ran all 3 cmd, rest if any issue still persist do let know

himself65:
I think this is not gonna happen even it's undefined

PR #4814, opened 2025-09-22
feat: add plugin Interface Enhancements for Payload CMS Integration

naaa760:
### PR Description:
- This PR adds essential plugin interface improvements to support seamless Payload CMS user synchronization with Better Auth. The changes enable plugins to:

**Key Features Added:**
- **Plugin Dependencies**: Plugins can now specify required dependencies using the `requires` field
- **Enhanced Database Hooks**: Added delete hooks for all models (user, session, account, verification) with before/after functionality
- **Lifecycle Management**: Introduced `onInit` hook for post-initialization setup and `onDestroy` for cleanup
- **Plugin-Level Database Hooks**: Plugins can now provide their own database hooks that integrate with the global hook system

**Benefits:**
- Enables Payload CMS to sync users from Better Auth as the single source of truth
- Provides proper cleanup mechanisms for plugin resources
- Supports conditional delete operations with hook validation
- Maintains backward compatibility with existing plugins


benjaminpreiss:
So, turns out that the payload plugin interface already exposes an `init` method: https://github.com/benjaminpreiss/payload-better-auth/blob/763369a487dd3b5a379b0212b3d878b232c75524/apps/payload-and-frontend/src/lib/betterAuthPlugin.ts#L96-L114

It however does not allow for clean up, which would be desirable.
There are these things that I think need to be done here:

- [ ] allow for returning a cleanup method, next to context `return {context: ..., cleanup: () => {}}`
- [ ] There is a type issue, which I had to avoid by casting - is this solvable? https://github.com/benjaminpreiss/payload-better-auth/blob/763369a487dd3b5a379b0212b3d878b232c75524/apps/payload-and-frontend/src/lib/betterAuthPlugin.ts#L109-L111
- [ ] Throughout the whole plugin, the extended context from the `init` return is not reflected in the types, as can be seen by my cast: https://github.com/benjaminpreiss/payload-better-auth/blob/main/apps/payload-and-frontend/src/lib/betterAuthPlugin.ts#L72
- [ ] documentation on the `init` method is missing

Issue #4815, opened 2025-09-22
Inconsistent use of baseURL vs productionURL in oauth-proxy plugin


jens-kuerten:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Configure the oauth-proxy like this:

```
betterAuth({
  baseURL: "https://myapp.com",          // (main) app domain
  trustedOrigins: ["https://myapp.com","https://*.myapp.com"]
  plugins: [
    oAuthProxy({
      productionURL: "https://login.myapp.com",  // OAuth redirect domain
    }),
  ],
});
```

Deploy the app on a VPS and point multiple domains at the app. (myapp.com, login.myapp.com, tenant1.myapp.com, etc..)

- If you log in with Google at tenant1.myapp.com, the user is redirected back to tenant1.myapp.com correctly. ✅

- If you log in at myapp.com, the user ends up at login.myapp.com instead of myapp.com. ❌

### Current vs. Expected behavior

**Current behavior:**
When baseURL and productionURL differ, the oauth-proxy plugin behaves inconsistently in a multi-domain setup.

When a user starts on baseURL (myapp.com), they will be redirect to productionURL (login.myapp.com) after login instead of staying on myapp.com.

If a user starts on any other url (e.g. tenant1.myapp.com) the redirect after login works correctly.

**Expected behavior:**
The user is always redirected back to the domain they started on, even if that domain is baseURL and differs from the productionURL.

### What version of Better Auth are you using?

1.3.9

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#1 SMP Tue Nov 5 00:21:55 UTC 2024",
    "release": "5.15.167.4-microsoft-standard-WSL2",
    "cpuCount": 16,
    "cpuModel": "AMD Ryzen 7 8845HS w/ Radeon 780M Graphics",
    "totalMemory": "46.71 GB",
    "freeMemory": "36.12 GB"
  },
  "node": {
    "version": "v22.15.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "10.15.0"
  },
  "frameworks": [
    {
      "name": "react",
      "version": "^19.1.1"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.16.3"
    },
    {
      "name": "kysely",
      "version": "^0.28.7"
    }
  ],
  "betterAuth": {
[...]

PR #4819, opened 2025-09-22
fix: OAuth Proxy Multi-Domain Redirect Issue

naaa760:
**Description:**
```
- Fix inconsistent URL handling in oauth-proxy plugin for multi-domain setups

**Problem:**
- Users starting OAuth flow on baseURL (myapp.com) were redirected to productionURL (login.myapp.com) after login instead of staying on original domain
- Inconsistent use of baseURL vs productionURL between before/after hooks
- Callback URLs resolved relative to wrong domain in multi-domain environments

**Solution:**
- Updated after hook to use productionURL consistently with before hook  
- Preserve original domain context by passing originalOrigin parameter
- Improve proxy logic for better multi-domain support
- Clean up unnecessary comments for better code readability

**Impact:**
Users now correctly redirected back to their starting domain (myapp.com, tenant1.myapp.com, etc.) after OAuth login, resolving the multi-domain setup issue.
```

**Related Issue:** #4815 (OAuth Proxy Multi-Domain Redirect Bug)

**Files Changed:**
- `packages/better-auth/src/plugins/oauth-proxy/index.ts` - Core fix
- `packages/better-auth/src/plugins/oauth-proxy/oauth-proxy.test.ts` - Test updates

### DIFF
--- a/packages/better-auth/src/plugins/oauth-proxy/index.ts
+++ b/packages/better-auth/src/plugins/oauth-proxy/index.ts
@@ -153,16 +153,20 @@ export const oAuthProxy = (opts?: OAuthProxyOptions) => {
 							}
 							const locationURL = new URL(location);
 							const origin = locationURL.origin;
-							/**
-							 * We don't want to redirect to the proxy URL if the origin is the same
-							 * as the current URL
-							 */
-							if (origin === getOrigin(ctx.context.baseURL)) {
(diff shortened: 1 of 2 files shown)


dvanmali:
I've noticed some inconsistencies between tests and production using `baseURL`.  Usually `ctx.baseURL` not `ctx.options.baseURL` points to options list `baseURL` in tests. I can take a deeper dive into this.

dvanmali:
Fixed with #4825

PR #4822, opened 2025-09-22
feat(client): add dedicated package for vanilla and react

himself65:


packages/better-auth/package.json
```diff
@@ -696,6 +696,9 @@
     }
   },
   "dependencies": {
+    "@better-auth/client-core": "workspace:*",
+    "@better-auth/core": "workspace:*",
+    "@better-auth/react": "workspace:*",
```
benmccann, line 701: As a non-React user, I would prefer having a lighter package without a dependency on code meant for React users. Could React users instead use `@better-auth/react` directly or could it be made an optional peer dependency?

himself65:
Closing as we will do that in another PR.

PR #4826, opened 2025-09-22
chore: use tsconfig monorepo setup

himself65:


dvanmali:
@himself65 `tsconfig.json` needs to be reverted. Duplicates are causing errors on `canary`.

Issue #4699, opened 2025-09-16
Allow to retrieve clientSecret from an async function


belgattitude:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

I couldn't find any way to refresh the clientSecret. 

For example in our entra configuration, we use the clientSecret from our azure keyvault. Those secrets will automatically by rotated (max age 90 ays, rotation 45 days)

```typescript
export const auth = createBetterAuth({
  db: dbBaseAuth,
  socialProviders: {
    microsoft: {
      clientId: serverEnv.AUTH_ENTRA_CLIENT_ID!,
      tenantId: serverEnv.AUTH_ENTRA_CLIENT_TENANT_ID!,
      authority: serverEnv.AUTH_ENTRA_AUTHORITY,
      prompt: 'select_account', 
      // 👉 On first run -> get the secret from the keyvault (async with @azure/keyvault-secrets)
      clientSecret:  await getEntraSecretFromKeyvault(),
    },
  },
});
```

Over a period of time (90days in this example) this secret will be disabled. The only way to get the refreshed one, is to restart the application server.

### Describe the solution you'd like

I'd like the have the possibility to set the clientSecret or to allow the configuration to support an async function that get the token (can be lru-ttl-cached for example)

### Describe alternatives you've considered

For now there's no urgency, cause everytime the app is restarted a new token is retrieved.

himself65:
lazy loading the social providers is a good idea

DustinMoriarty:
This would also prevent having to provide a placeholder secret at build time. I think it would be a good idea for all of the primitive type values in the configuration to support callbacks.

belgattitude:
Still relevant in my humble opinion

Can you reopen ?

Issue #1522, opened 2025-02-21
Apple Authentication client secrets will eventually expire


etler:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

* Set up better auth with [apple authentication](https://www.better-auth.com/docs/authentication/apple).
* Simulate an apple client secret expiring by generating a JWT with a short expiration time
* Attempt to log in with apple social auth via the auth client

### Current vs. Expected behavior

Current: The auth server eventually becomes invalid for apple authentication

Expected: The auth server should not unexpectedly begin failing at a point in the future

### What version of Better Auth are you using?

1.1.16

### Provide environment information

```bash
- Device: Pixel 7 Pro
- OS: Android 15
- App Platform: React Native with Expo managed development build
- Plugin: Expo
- Default Browser: Chrome
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
export const auth = betterAuth({
  appName: "App",
  baseURL: host,
  basePath: "/auth",
  secret: sessionSecret,
  database: new Pool({
    connectionString: dbUrl,
  }),
  logger: {
    disabled: false,
    leve: "debug",
  },
  user: {
    // "user" is a reserved keyword in postgres so we are using plural names for tables
    modelName: "users",
    deleteUser: {
      enabled: true,
      afterDelete: async (user) => {
        await deleteMember(user.email)
      },
    },
  },
  session: {
    modelName: "sessions",
  },
  account: {
    modelName: "accounts",
    accountLinking: {
      enabled: true,
    },
  },
  socialProviders: {
    google: {
      clientId: googleClientId,
      clientSecret: googleClientSecret,
      scope: ["openid", "profile", "email"],
    },
    apple: {
      clientId: appleClientId,
      clientSecret: makeAppleClientSecret({
        clientId: appleClientId,
        teamId: appleTeamId,
        keyId: appleKeyId,
        key: appleKey,
      }),
    },
  },
  plugins: [expo()],
  trustedOrigins: [siteHostname, mobileHostname],
  advanced: {
    cookiePrefix: cookieNamespace,
  },
  secondaryStorage: {
    get: async (key) => cache.get(key),
    set: async (key, value, ttl) => cache.set(key, value, ttl),
    delete: async (key) => cache.delete(key),
  },
})
```

### Additional context

[...]

PR #4829, opened 2025-09-23
feat: awaitable social provider config

dvanmali:
Allows awaitable configuration for social providers. Fetches options at initialization to check for population of specific fields and per request. Useful for clients with expirable secrets such as [Apple](https://developer.apple.com/documentation/accountorganizationaldatasharing/creating-a-client-secret) and other dynamically registered clients that utilize non-zero `client_secret_expires_at` as specified in [RFC 7591](https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.1).

Suggestion:
-  [x] #4699

Fixes:
-  [x] #1522

---
## Summary by cubic
Adds awaitable social provider config so providers and OAuth options can be resolved at runtime. This improves support for rotating/expiring secrets and dynamic clients across all OAuth flows.

- **New Features**
  - SocialProviders entries now accept a config object or () => Promise<config>.
  - ProviderOptions (including clientSecret) is awaitable in authorization URL, authorization code, refresh token, and client credentials flows.
  - Added async request helpers: authorizationCodeRequest, refreshAccessTokenRequest, clientCredentialsTokenRequest.
  - Added getAwaitableValue helper and updated endpoints/plugins to resolve provider configs on demand. Apple provider and telemetry detection updated to use async config.

- **Migration**
  - Prefer the new async request helpers listed above; the existing create* functions remain for backward compatibility but are deprecated.
  - If you access social provider config directly, it may be a function; handle resolving it asynchronously.

<sup>Written for commit 15c5e1ce26ae5d43eef807d07cb00d4e433e3e1b. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/src/oauth2/client-credentials-token.ts
+++ b/packages/core/src/oauth2/client-credentials-token.ts
@@ -3,7 +3,7 @@ import { betterFetch } from "@better-fetch/fetch";
 import type { AwaitableFunction } from "../types";
 import type { OAuth2Tokens, ProviderOptions } from "./oauth-provider";
 
-export async function createClientCredentialsTokenRequest({
+export async function clientCredentialsTokenRequest({
 	options,
 	scope,
 	authentication,
@@ -14,12 +14,34 @@ export async function createClientCredentialsTokenRequest({
 	authentication?: ("basic" | "post") | undefined;
 	resource?: (string | string[]) | undefined;
 }) {
+	options = typeof options === "function" ? await options() : options;
+	return createClientCredentialsTokenRequest({
+		options,
+		scope,
+		authentication,
+		resource,
+	});
+}
+
+/**
+ * @deprecated use async'd clientCredentialsTokenRequest instead
+ */
+export function createClientCredentialsTokenRequest({
+	options,
+	scope,
+	authentication,
+	resource,
+}: {
+	options: ProviderOptions & { clientSecret: string };
+	scope?: string | undefined;
+	authentication?: ("basic" | "post") | undefined;
+	resource?: (string | string[]) | undefined;
+}) {
 	const body = new URLSearchParams();
 	const headers: Record<string, any> = {
 		"content-type": "application/x-www-form-urlencoded",
 		accept: "application/json",
 	};
-	options = typeof options === "function" ? await options() : options;
 
 	body.set("grant_type", "client_credentials");
 	scope && body.set("scope", scope);
@@ -67,7 +89,7 @@ export async function clientCredentialsToken({
 	authentication?: ("basic" | "post") | undefined;
 	resource?: (string | string[]) | undefined;
 }): Promise<OAuth2Tokens> {
-	const { body, headers } = await createClientCredentialsTokenRequest({
+	const { body, headers } = await clientCredentialsTokenRequest({
 		options,
 		scope,
 		authentication,
--- a/packages/core/src/oauth2/index.ts
+++ b/packages/core/src/oauth2/index.ts
@@ -1,5 +1,6 @@
 export {
 	clientCredentialsToken,
+	clientCredentialsTokenRequest,
 	createClientCredentialsTokenRequest,
 } from "./client-credentials-token";
 export { createAuthorizationURL } from "./create-authorization-url";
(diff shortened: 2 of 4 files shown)


dvanmali:
This shall be marked for 1.4.x.

Minor change:
- some exported functions which are now `async` (which may be used in some custom plugins tests)

dvanmali:
@himself65 I like that approach but then it wouldn't be supported by `genericOAuth`. Async plugins might be a difficult feature altogether, unless genericOAuth is different and performs like the following?:

```ts
export const auth = createBetterAuth({
  db: dbBaseAuth,
    plugins: async() => ([ // awaitable option1
        genericOAuth({ 
            config: [ 
                { 
                    providerId: "provider-id",
                    clientId: "test-client-id",  // awaitable option2
                    clientSecret: await getEntraSecretFromKeyvault(), // option2
                    discoveryUrl: "https://auth.example.com/.well-known/openid-configuration", 
                    // ... other config options
                }, 
                // Add more providers as needed
            ] 
        }) 
    ])
});
```

packages/better-auth/src/social-providers/social.test.ts
```diff
@@ -75,7 +75,11 @@ vi.mock("../oauth2", async (importOriginal) => {
 			const { refreshToken, options, tokenEndpoint } = args;
 			expect(refreshToken).toBeDefined();
 			expect(options.clientId).toBe("test-client-id");
-			expect(options.clientSecret).toBe("test-client-secret");
+			expect(
+				typeof options.clientSecret === "string"
+					? options.clientSecret
+					: await options.clientSecret(),
+			).toBe("test-client-secret");
```
himself65: should be a determined type

dvanmali: Like this?

```ts
if (typeof options.clientSecret === "string") {
	expect(options.clientSecret).toBe("test-client-secret");
} else {
	const secret = await options.clientSecret();
	expect(secret).toBe("test-client-secret");
}
```

Bekacru:
Why can't we make generic oauth config async function as well

dvanmali:
> Why can't we make generic oauth config async function as well

That would be fine for generic oauth but the issue for the apple social provider won't be solved since it is top level. Additionally, the apple client secret needs to be provisioned per call since it can expire.

Bekacru:
oh no im saying for both social provider and generic oauth...and we resolve the function per call not on init

Bekacru:
why did you close it @dvanmali? I was saying why not do what @himself65 suggested
```ts
export const auth = createBetterAuth({
  db: dbBaseAuth,
  socialProviders: {
    microsoft: async () => ({
      clientId: serverEnv.AUTH_ENTRA_CLIENT_ID!,
      tenantId: serverEnv.AUTH_ENTRA_CLIENT_TENANT_ID!,
      authority: serverEnv.AUTH_ENTRA_AUTHORITY,
      prompt: 'select_account', 
      // 👉 On first run -> get the secret from the keyvault (async with @azure/keyvault-secrets)
      clientSecret:  await getEntraSecretFromKeyvault(),
    }),
  },
});
```

Bekacru:
it will also allow if people want to change client id dynamically...maybe we should pass `ctx` object to the callback. 


n.b. you can already do this through hooks

dvanmali:
@Bekacru Was going to open a new PR for async config but I'll just reset this one

dvanmali:
@himself65 Updated this to be backwards compatible by marking sync functions as deprecated and making new async functions. Thus no longer breaking.

Alvi24:
will this be merged?

reslear:
It would be helpful to get some like `context` for example, it would be useful to have `headers`: 

```ts
export const auth = createBetterAuth({
  socialProviders: {
    apple: async (ctx) => ({
      appBundleIdentifier: 
      	ctx.headers.get('x-platform') === 'web'
      		? `${env.BUNDLE_ID}.webapp` 
      		: env.BUNDLE_ID ,
      clientId: '',
      clientSecret: '',
    }),
  }
})
```

This would help to obtain accurate dynamic data, for example, for white label applications or applications with multiple domains and a single backend. 

@dvanmali @Bekacru @himself65 What do you think?

dvanmali:
@reslear nice idea! could you open a feature request after this PR?

himself65 (approved):

reslear:
@dvanmali sure https://github.com/better-auth/better-auth/issues/7830

PR #4835, opened 2025-09-23
fix(better-auth): moved email verification check after password check

QuintenStr:
### Clearly describe what changes you made and why
I moved the email verification check to after the password validation check. The issue was that on username sign in, nonetheless if the password was correct, you'd always get email verification error (if the user's email is unverified). This leaks information of the user, and is a possible spam vulnerability for email sending.

### Include any relevant context or background
Just moved the same code a bit lower. Refer to changed file.

### List any breaking changes or deprecations
No breaking changes found. All username plugin tests still run.

### Add screenshots for UI changes
No UI changes.

### Reference related issues or discussions
Not applicable.
    

---

## Summary by cubic
Moved the email verification check to run after password validation in the username sign-in flow. This prevents leaking unverified status on incorrect passwords and avoids sending verification emails unless the password is correct.

- **Bug Fixes**
  - Validate password first; skip email verification when password is wrong.
  - Check email verification only after a valid password and block sign-in if unverified.
  - Send verification email on sign-in (if enabled) only after password is confirmed.


QuintenStr:
> Could you please add one test case for this?

Tests have been added :)

himself65 (approved):

PR #4842, opened 2025-09-23
fix(admin): validate admin role updates against the configured roles to prevent setting a non-existent role

hieudien14310:
## Summary by cubic
Validate admin role updates against the configured roles to prevent setting a non-existent role. If the role is invalid, the request returns BAD_REQUEST.

- **Bug Fixes**
  - Check opts.roles for the provided role before updating.
  - Add ADMIN_ERROR_CODES.YOU_ARE_NOT_ALLOWED_TO_SET_NON_EXISTENT_VALUE.


hieudien14310:
Hi @himself65, I have add a test. Please review again.

PR #4876, opened 2025-09-24
[two-factor] fix: forward invalid errorKey when verifying existing session

hartbit:
I noticed, while debugging two-factor integration inside my app, that when `verifyTwoFactor` is called with an existing session (to confirm enabling two factor), that error messages are not propagated correctly.


hartbit:
This has been fixed recently. The PR is no longer necessary.

PR #4880, opened 2025-09-24
Update installation.mdx

otakutyrant:
use process.env.BETTER_AUTH_URL instead.

I think it works better?
    

---

## Summary by cubic
Use process.env.BETTER_AUTH_URL for the authClient baseURL in the Solid example, replacing the hardcoded localhost value. This makes configuration environment-aware and reduces deployment mistakes.


himself65:
I think no. process.env is not available in cloudflare worker, deno...

otakutyrant:
Holy environement!

PR #4885, opened 2025-09-24
docs: use `createFileRoute` on TanStack API 

byte-264:
Updated the TanStack integration example to use createFileRoute instead of createServerFileRoute and added POST method handling.
    

---

## Summary by cubic
Updated the TanStack Router integration example to use createFileRoute with server.handlers and add POST support, aligning docs with the current API.

- **Refactors**
  - Replace createServerFileRoute with createFileRoute from @tanstack/react-router.
  - Define GET and POST under server.handlers, both delegating to auth.handler.
  - Export Route instead of ServerRoute.

### DIFF
--- a/docs/content/docs/integrations/tanstack.mdx
+++ b/docs/content/docs/integrations/tanstack.mdx
@@ -13,15 +13,19 @@ We need to mount the handler to a TanStack API endpoint/Server Route.
 Create a new file: `/src/routes/api/auth/$.ts`
 
 ```ts title="src/routes/api/auth/$.ts"
-import { auth } from '@/lib/auth' // import your auth instance
-import { createServerFileRoute } from '@tanstack/react-start/server'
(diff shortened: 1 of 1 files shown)


docs/content/docs/integrations/tanstack.mdx
```diff
@@ -13,15 +13,19 @@ We need to mount the handler to a TanStack API endpoint/Server Route.
 Create a new file: `/src/routes/api/auth/$.ts`
 
 ```ts title="src/routes/api/auth/$.ts"
```
himself65, line 15: Maybe it's not in the same file path now?

alexjamesgodfrey: Nope, it is! This works

alexjamesgodfrey (approved):

himself65 (approved):

Issue #4722, opened 2025-09-17
Nuxt SSR: useFetch integration causes session to return undefined after OAuth redirect in v1.3.10/v1.3.11


cwandev:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1、Use Nuxt SSR with better-auth
2、Use const { data: session } = await authClient.useSession(useFetch)
3、Login with GitHub OAuth
4、Session data is available initially
5、Page auto-refreshes after successful login
6、Session becomes undefined (but cookies are still present)

https://github.com/nuxt-one/nuxt-one

### Current vs. Expected behavior

Current:

-  Session data available initially after login
- Page auto-refreshes
- session becomes undefined  (cookies still exist)
- Using authClient.useSession() without useFetch works fine

Expected:

- Session should return user data, not undefined

### What version of Better Auth are you using?

1.3.10

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:29 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6000",
    "release": "24.6.0",
    "cpuCount": 8,
    "cpuModel": "Apple M1 Pro",
    "totalMemory": "16.00 GB",
    "freeMemory": "0.34 GB"
  },
  "node": {
    "version": "v22.17.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.2"
  },
  "frameworks": [
    {
      "name": "vue",
      "version": "^3.5.21"
    },
    {
      "name": "nuxt",
      "version": "^4.1.2"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.16.3"
    },
    {
      "name": "drizzle",
      "version": "^0.44.5"
    }
  ],
  "betterAuth": {
    "version": "^1.3.10",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Client

### Auth config (if applicable)

```typescript

```

### Additional context

Related: Not sure if this is related to #4578

himself65:
does it same when disable SSR?

cwandev:
> does it same when disable SSR?

It works fine when SSR is disabled.

himself65:
Thanks, I will have a try

himself65:
Hello, I've tried and it works well

[...]

PR #4901, opened 2025-09-25
fix(vue): improve baseURL normalization for Nuxt SSR compatibility

Mateleo:
Following the discussion in [#4722](https://github.com/better-auth/better-auth/issues/4722), and after the fix that landed in [#4887](https://github.com/better-auth/better-auth/pull/4887), Nuxt `useSession(useFetch)` started working again **only** when using the default `/api/auth` base path.

However, the fix did not cover cases where a **custom base path** is configured (e.g. `/api/v1/auth`).

Like this :
```ts
  baseURL: "http://localhost:3000/api/v1/auth",
});
```

In these cases, the client still builds an absolute URL for `get-session` (like so `http://localhost:3000/api/v1/auth`), which in Nuxt SSR prevents cookies from being forwarded. The result is that `session` is always `null`, even though the endpoint responds correctly in the browser.

```ts
  const loggedFetch = ((url: string, opts?: any) => {
    console.log("[useSession] fetching:", url);
    return useFetch(url, opts);
  }) as any;

  const { data: session, error } = await authClient.useSession(loggedFetch);
  // return [useSession] fetching: http://localhost:3000/api/v1/auth/get-session
  // After this PR
  // return [useSession] fetching: /api/v1/auth/get-session
```

This PR ensures the client:

* Normalizes the `baseURL` to extract only the pathname.
* Defaults `/` to `/api/auth`.
* Trims trailing slashes.
* Always calls a **relative** endpoint for `…/get-session`, which allows Nuxt SSR to forward cookies properly.

**For Nuxt SSR custom base url**

* **Before #4887:** regression → always `null`.
* **After #4887:** works only for default baseURL.
* **With this PR:** works for both default and custom base paths, without needing app-side workarounds (like wrapping `useFetch` to strip the origin).

    

---

## Summary by cubic
Fixes Nuxt SSR cookie forwarding when using a custom auth baseURL. We now normalize baseURL and call a relative /get-session path so useSession returns the session for both default and custom paths.

- **Bug Fixes**
  - Normalize baseURL to a relative pathname; strip origin if absolute.
  - Ensure leading slash, trim trailing slashes, fallback to /api/auth.
  - Build /get-session from the normalized path to avoid absolute requests in SSR.

### DIFF
--- a/packages/better-auth/src/client/vue/index.ts
+++ b/packages/better-auth/src/client/vue/index.ts
@@ -34,7 +34,10 @@ function normalizeAuthPath(baseURL?: string): string {
 
 	try {
 		// Absolute URL → use only pathname so Nuxt treats it as same-origin and forwards cookies.
-		if (baseURL.toLowerCase().startsWith("http://") || baseURL.toLowerCase().startsWith("https://")) {
+		if (
+			baseURL.toLowerCase().startsWith("http://") ||
+			baseURL.toLowerCase().startsWith("https://")
+		) {
 			const u = new URL(baseURL);
 			return (u.pathname || fallback).replace(/\/+$/, "") || fallback;
 		}


dvanmali:
@himself65

Issue #4774, opened 2025-09-19
[flaky] mongodbAdpater unit test


himself65:
https://github.com/better-auth/better-auth/actions/runs/17867884619/job/50814489729?pr=4768

https://github.com/better-auth/better-auth/actions/workflows/ci.yml?query=is%3Afailure

himself65:
Great thanks to @ping-maxwell

himself65:
Landed in https://github.com/better-auth/better-auth/commit/0a1df9e43fbab858587d657e51d141e1e5fd154c

PR #4921, opened 2025-09-26
test: increase retry time for flaky test

himself65:
Related: https://github.com/better-auth/better-auth/issues/4774
    

## Summary by cubic
Increase suite-level retry to 3 for both Prisma adapter test suites to reduce flaky failures and stabilize CI runs. No runtime or adapter changes; only test configuration was updated.


himself65:
Not working, I guess if you fail once, you will fail forever because of the side effect

PR #4929, opened 2025-09-26
refactor: move parse user input before insert into db

himself65:
I think the timing of checking the schema is weird to me. Trying to move this now to see if it works
    

---

## Summary by cubic
Move user input parsing into the adapter so all user writes are validated before hitting the DB. This simplifies sign-up and update flows and ensures consistent schema handling.

- **Refactors**
  - Parse user input in the adapter for the user model before insert/update.
  - Remove redundant parsing from sign-up and update routes; pass fields through.
  - Make parseOutputData and getAllFields internal.
  - Default parseUserInput to an empty object to avoid undefined cases.


himself65:
It's landed, but we have https://github.com/better-auth/better-auth/commit/115438f7fb608974ee9d67ed3c1a78dd56c5bda0 to fix it

Issue #4903, opened 2025-09-25
Support for Custom Fields in the Account Schema


j-xzy:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

Currently, better-auth works seamlessly with its core schema for the Account model. However, in many real-world applications, developers need to extend the Account schema to store additional, provider-specific, or application-specific data.

For example, one might need to store:
- A team_id when linking a Slack account.
- A tenant_id for multi-tenant applications.
- An status enum (ACTIVE, REVOKED) to manage the state of the connection.
- A bot_id or other metadata returned from the OAuth provider's token response.

### Describe the solution you'd like

```
betterAuth({
	account: {
		additionalFields: {
			clientId: {
				type: 'string',
				required: true,
			},
		},
	},
});
```

### Describe alternatives you've considered

```
betterAuth({
	databaseHooks: {
		account: {
			create: {
				after: async (account) => {
					this.prismaService.account.update({
						data: {
							clientId: config.clientId,
						},
						where: {
							id: account.id,
						},
					});
				},
			},
		},
	},
});
```

dvanmali:
Hi @j-xzy This feature should be available as you described. is something missing from the [docs](https://www.better-auth.com/docs/concepts/database#extending-core-schema)?

j-xzy:
@dvanmali https://www.better-auth.com/docs/concepts/database#extending-core-schema
Schema extension is only available for the User and Session tables, not for the Account table.

AlexErrant:
Can this issue be reopened? I quote from [the above PR](https://github.com/better-auth/better-auth/pull/4935#issuecomment-3372643367): "But in conclusion, this PR, for now, is incomplete."

PR #4935, opened 2025-09-27
feat: additional fields on account

dvanmali:
Closes: #4903
    

---

## Summary by cubic
Added support for account.additionalFields in BetterAuthOptions to let apps store custom Account attributes. Updated the Next.js example docs with an SSO login flow using DummyIDP.

### DIFF
--- a/packages/better-auth/src/db/get-tables.ts
+++ b/packages/better-auth/src/db/get-tables.ts
@@ -231,6 +231,7 @@ export const getAuthTables = (
 					onUpdate: () => new Date(),
 				},
 				...account?.fields,
+				...options.account?.additionalFields,
 			},
 			order: 3,
 		},


JE4GLE:
@himself65 @Bekacru @cubic-dev-ai Is this a schema change for the accounts table? If yes, it needs to be documented here: https://www.better-auth.com/docs/concepts/database#account

himself65:
> @himself65 @Bekacru @cubic-dev-ai Is this a schema change for the accounts table? If yes, it needs to be documented here: https://www.better-auth.com/docs/concepts/database#account

No, this is not. This just allows you to add additional fields. So for the old user, it won't affect anything

himself65:
@dvanmali I think we might need to update the api impl for that. I forgot what I discussed last week with @Bekacru. But in conclusion, this PR, for now, is incomplete.

PR #4950, opened 2025-09-27
fix(client): add lynx client exports

JagritGumber:
Add exports for the lynx client in build config and package.json to enable integration with the Lynx framework.
    

---

## Summary by cubic
Added Lynx client support to better-auth so Lynx apps can use the client directly. The build now includes the Lynx client, with ESM/CJS entry points and TypeScript types exposed.

- **New Features**
  - Build: include ./src/client/lynx/index.ts in build entries.
  - Exports: add better-auth/lynx ESM/CJS entries with type definitions.
  - TypeScript: map lynx types to ./dist/client/lynx/index.d.ts.


JagritGumber:
The issue was that the lynx client wasn't being compiled because it was missing from the build configuration. Here's what was preventing it from compiling:

1. **Missing build entry**: The `build.config.ts` didn't include `"./src/client/lynx/index.ts"` in the `entries` array, so `unbuild` wasn't processing the lynx directory.

2. **Missing package exports**: The `package.json` didn't have an export for `"./lynx"` in the `exports` field.

3. **Missing typesVersions**: The `typesVersions` didn't include `"lynx"` mapping to the built type definitions.

I fixed this by:
- Adding `"./src/client/lynx/index.ts"` to the build entries
- Adding the `"./lynx"` export in `package.json` with proper paths to the built files
- Adding `"lynx": ["./dist/client/lynx/index.d.ts"]` to `typesVersions`

After running the build, the lynx client now compiles successfully and appears in `dist/client/lynx/` with the expected exports (`createAuthClient`, `capitalizeFirstLetter`, `useStore`).

Issue #4472, opened 2025-09-05
Rate limiter keys in secondary storage lack TTL, causing memory growth over time


ianseddon:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Configure better-auth with any secondary storage (e.g., Redis, DynamoDB, etc.)
2. Set rate limiting to use secondary storage:
   ```typescript
   rateLimit: {
     window: 10, // seconds
     max: 100,
     storage: "secondary-storage"
   }
   ```
3. Make requests from different IPs to trigger rate limit key creation.
4. Observe that keys are created in secondary storage but never expire (without additional configuration of eviction policies on the secondary storage)
5. Over time, storage fills with stale rate limit keys

### Current vs. Expected behavior

Current: Rate limit keys are stored indefinitely in secondary storage.
Expected: Stale rate limiter state should automatically expire after the rate limit window has passed.

### What version of Better Auth are you using?

1.3.7

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#1 SMP PREEMPT_DYNAMIC Sun, 06 Jul 2025 11:14:36 +0000",
    "release": "6.15.5-arch1-1",
    "cpuCount": 32,
    "cpuModel": "13th Gen Intel(R) Core(TM) i9-13900K",
    "totalMemory": "62.56 GB",
    "freeMemory": "46.68 GB"
  },
  "node": {
    "version": "v22.14.0",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.2.17"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "^15.3.5"
    },
    {
      "name": "react",
      "version": "^19.0.0"
    }
  ],
  "databases": [
    {
      "name": "@prisma/client",
      "version": "^6.14.0"
    }
  ],
  "betterAuth": {
    "version": "0.1.0",
    "config": {
      "baseURL": "http://localhost:3000",
      "secondaryStorage": {},
      "emailAndPassword": {
        "enabled": false
      },
      "plugins": [
        {
          "name": "next-cookies",
          "config": {
            "id": "next-cookies",
            "hooks": {
              "after": [
                {}
              ]
            }
          }
        },
        {
          "name": "generic-oauth",
          "config": {
            "id": "generic-oauth",
            "endpoints": {},
            "$ERROR_CODES": {
[...]

PR #4961, opened 2025-09-29
fix: ttl for rate limited secondary storage

### LINKED ISSUE #4034: Why few keys in redis has no TTL?

### To Reproduce

Hey im noticing many keys are there in redis which has no TTL, so it stays forever, session information will be expiry after 30 days, 

Im using secondary database as redis by passing 
 secondaryStorage: {
    get: async (key) => {
      const value = await redis.get(key);
      return value ? value : null;
    },
      if (ttl) await redis.set(key, value, ttl);
      else await redis.set(key, value);
    },
    delete: async (key) => {
      await redis.delete(key);
    },
  },

to auth 

Image

Just need to understand how to add ttl to this? and is it bug? 
because of this redis is getting huge amout of data, 
even when this is used?

Should have TTL for all the keys

1.3.1

### System info

```bash
System:
    OS: macOS 15.5
    CPU: (11) arm64 Apple M3 Pro
    Memory: 82.19 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 139.0.7258.128
    Safari: 18.5
```

Backend

```typescript
import dotenv from "dotenv";
dotenv.config();
import { MongoClient } from "mongodb";
import { mongodbAdapter } from "better-auth/adapters/mongodb";
import redis from "@/lib/redis";

const client = new MongoClient(
  (process.env.MONGODB_URI as string) || "mongodb://localhost:27017",
);
const db = client.db();

  database: mongodbAdapter(db),
  socialProviders: {
    google: {
      prompt: "select_account",
    },
    linkedin: {
      clientId: process.env.LINKEDIN_CLIENT_ID as string,
      clientSecret: process.env.LINKEDIN_CLIENT_SECRET as string,
    },
    tiktok: {
      clientId: process.env.TIKTOK_CLIENT_ID as string,
      clientSecret: process.env.TIKTOK_CLIENT_SECRET as string,
      clientKey: process.env.TIKTOK_CLIENT_KEY as string,
    },
    facebook: {
      clientId: process.env.FACEBOOK_CLIENT_ID as string,
      clientSecret: process.env.FACEBOOK_CLIENT_SECRET as string,
    },
    spotify: {
      clientId: process.env.SPOTIFY_CLIENT_ID as string,
      clientSecret: process.env.SPOTIFY_CLIENT_SECRET as string,
    },
  },
  account: {
    accountLinking: {
      enabled: true,
      trustedProviders: ["google", "linkedin", "facebook", "tiktok", "spotify"],
    },
  },
  plugins: [nextCookies(), admin()],
  secondaryStorage: {
    get: async (key) => {
      const value = await redis.get(key);
      return value ? value : null;
    },
      if (ttl) await redis.set(key, value, ttl);
      else await redis.set(key, value);
    },
    delete: async (key) => {
      await redis.delete(key);
    },
  },
});
```

### Additional context

This has been created help thread in Discord as well, here is ref https://discord.com/channels/1288403910284935179/1402531236454338641

dvanmali:
Adds ttl for rate limited secondary storage values.

Fixes:
- #4034
- #4472

Additional PR:
- #3836

---

## Summary by cubic
Adds TTL to rate-limiter entries stored in secondary storage, using the path-specific window so keys expire correctly. Fixes #4034 and #4472.

- **Bug Fixes**
  - Apply TTL when writing to secondaryStorage (uses route window, falls back to global, default 10s).
  - Pass window into storage factory and tidy storage selection.
  - Tests verify TTL=10 and correct count/429 behavior with custom storage.

### DIFF
--- a/packages/better-auth/src/api/rate-limiter/index.ts
+++ b/packages/better-auth/src/api/rate-limiter/index.ts
@@ -99,8 +99,7 @@ export function getRateLimitStorage(
 			},
 			set: async (key: string, value: RateLimit, _update?: boolean) => {
 				const ttl =
-					(rateLimitSettings?.window ?? ctx.options.rateLimit?.window ?? 10) *
-					1000;
+					rateLimitSettings?.window ?? ctx.options.rateLimit?.window ?? 10;
 				await ctx.options.secondaryStorage?.set?.(
 					key,
 					JSON.stringify(value),


packages/better-auth/src/api/rate-limiter/rate-limiter.test.ts
```diff
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi } from "vitest";
 import { getTestInstance } from "../../test-utils/test-instance";
+import type { RateLimit } from "better-auth/types";
```
himself65: import the wrong file path

dvanmali: Fixed 👍

packages/better-auth/src/api/rate-limiter/index.ts
```diff
@@ -81,23 +81,34 @@ function createDBStorage(ctx: AuthContext) {
 }
 
 const memory = new Map<string, RateLimit>();
-export function getRateLimitStorage(ctx: AuthContext) {
+export function getRateLimitStorage(
+	ctx: AuthContext,
+	rateLimitSettings?: {
+		window?: number;
+		max?: number;
```
himself65: `max` is not used at all

packages/better-auth/src/api/rate-limiter/index.ts
```diff
@@ -81,23 +81,34 @@ function createDBStorage(ctx: AuthContext) {
...
 		return {
 			get: async (key: string) => {
 				const data = await ctx.options.secondaryStorage?.get(key);
 				return data ? safeJSONParse<RateLimit>(data) : undefined;
 			},
-			set: async (key: string, value: RateLimit) => {
-				await ctx.options.secondaryStorage?.set?.(key, JSON.stringify(value));
+			set: async (key: string, value: RateLimit, _update?: boolean) => {
+				const ttl =
+					rateLimitSettings?.window ?? ctx.options.rateLimit?.window ?? 10000;
```
himself65: Why does it read rateLimit.window? secondary storage should not related with that

packages/better-auth/src/api/rate-limiter/index.ts
dvanmali: Correct, it is part of the settings, I passed it in case it would need to be used in the future but I'll remove it

packages/better-auth/src/api/rate-limiter/index.ts
dvanmali: Rate limit window determines the retry rate. For example, an endpoint with `window: 10` (10 seconds) and `max: 3` (requests) is as followed:

- Create rate limit object with `count: 1`. TTL in 10 secs.
- At 5 sec, count increments and `lastRequest` updates to present. TTL updates to expire in 10 seconds. (I could do `_update ? undefined : ttl` which would keep the existing TTL but having unset keys was the primary concern of the issue report. To ensure all keys have a TTL, it'd be better to reset and let the update logic (ie. `data` value) handle counting not the secondary-storage expiration).
- At 15 seconds, ttl would expire and future counts would need to set a new key.

dvanmali: _Side note_: the current rate limiting approach is not atomic so `count` isn't very accurate such as with spikey requests. I'd suggest a sliding window approach but this is OOS with the current issue and would require a larger PR change.

himself65 (approved):

grrowl:
`ctx.options.rateLimit?.window ?? 10;` is too low for cloudflare KV, which has a minimum KV TTL of 60s, breaking auth entirely when the default is set.

PR #4974, opened 2025-09-30
fix(oidc): properly enforce consent requirements per OIDC spec

himself65:
## Summary by cubic
Enforces OIDC consent correctly: consent is required unless the client is trusted or the user previously consented, and prompt=consent always forces consent. Authorization now decides when to show consent and only redirects with a code when consent isn’t needed.

- **Bug Fixes**
  - Authorization computes requireConsent: !skipConsent && (!hasAlreadyConsented || prompt==="consent").
  - Removes forced prompt="consent" in cookie-based flow; lets authorize decide per spec.
  - Redirects with code immediately when consent isn’t required; otherwise routes to consent page/cookie flow with state preserved.
  - Updates consent endpoint schema to accept nullish consent_code for cookie-based flows.
  - Expands tests with a consent-flow helper and coverage for trusted vs. untrusted clients and JWT signing paths.


himself65 (commented):
reviewed by the internal group. will merge later

PR #4984, opened 2025-09-30
fix: passkeys plugin response structure

Bartek532:
In multiple places passkeys plugin expect to get a response with key `data`, but we're currently returning the data itself resulting in passkeys plugin not working at all.

Some examples:
- https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/plugins/passkey/client.ts#L111
- https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/plugins/passkey/client.ts#L111
    

---

## Summary by cubic
Standardized passkeys plugin responses to return { data: ... } as the client expects. This fixes broken passkey flows caused by mismatched response shapes.

- **Bug Fixes**
  - Wrapped options responses in { data } for passkey initialization endpoints.
  - Wrapped create-passkey response in { data } to align with client parsing.

### DIFF
--- a/docs/content/docs/concepts/hooks.mdx
+++ b/docs/content/docs/concepts/hooks.mdx
@@ -136,7 +136,7 @@ const hook = createAuthMiddleware(async (ctx) => {
 #### Cookies
 
 - Set cookies: `ctx.setCookies` or `ctx.setSignedCookie`.
-- Get cookies: `ctx.getCookies` or `ctx.getSignedCookies`.
+- Get cookies: `ctx.getCookies` or `ctx.getSignedCookie`.
 
 Example:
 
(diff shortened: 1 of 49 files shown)


Bekacru:
`options.data` is there because `$fetch` helper function by default returns `data` and `error` object. Have you by any chance enabled `throw:true` in your createAuthClient

Bartek532:
> `options.data` is there because `$fetch` helper function by default returns `data` and `error` object. Have you by any chance enabled `throw:true` in your createAuthClient

Hey, got it, yes, I enabled it globally because I want to throw an error and catch it on my own, shouldn't this be reflected in the implementation of the passkeys plugin?

PR #4986, opened 2025-09-30
fix(mongodb-adapter): add where clause support to count method

t-kirsch:
## Problem
   The `count` method in the MongoDB adapter was ignoring the `where` parameter, causing it to count all documents in the collection instead of filtering by the provided conditions.

   This was particularly problematic when checking organization membership limits before accepting invitations - it would count members from **all organizations** instead of just the specific organization, leading to incorrect limit validation.

   ## Solution
   - Added `where` parameter to the `count` method signature
   - Use `convertWhereClause` to build the MongoDB filter properly
   - Matches the implementation pattern of other methods (`findOne`, `findMany`, etc.)

   ## Changes
   - Modified `packages/better-auth/src/adapters/mongodb-adapter/mongodb-adapter.ts`
   - Changed the `count` method to accept and use the `where` parameter

   ## Impact
   This fix ensures that organization membership limits are correctly enforced per organization when:
   - Accepting invitations
   - Adding new members
   - Any other operation that uses `countMembers`
    

---

## Summary by cubic
Fixes the MongoDB adapter’s count method to respect the where clause, so counts are correctly filtered. This prevents overcounting and ensures accurate organization membership limit checks.

- **Bug Fixes**
  - Added where to the count method signature.
  - Built the MongoDB filter with convertWhereClause.
  - Aligned count with findOne/findMany implementation patterns.


t-kirsch:
My apologies, should have thought of that.

I've added two tests to the generic adapter suite for the count method. This covers both the base case and for filtering with where, making sure it works as expected for all adapters.

@himself65

t-kirsch:
It looks like you addressed this issue in parallel with #4844 @ping-maxwell  – thanks for the quick work on it!

You added the where clause in the mongodb-adapter:
https://github.com/better-auth/better-auth/blob/19f39cb400d1b7afe925528bb58bccb8c3f077c3/packages/better-auth/src/adapters/mongodb-adapter/mongodb-adapter.ts#L217-L223

and test cases for the count method in adapters/tests/normal.ts:
https://github.com/better-auth/better-auth/blob/19f39cb400d1b7afe925528bb58bccb8c3f077c3/packages/better-auth/src/adapters/tests/normal.ts#L536-L560

Since this is now resolved, I'll go ahead and close this pull request.

Issue #3088, opened 2025-06-19
sendVerificationEmail error during the email change process


TonybynMp4:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Have better auth setup to use email & password, with sendVerificationEmail & sendChangeEmailVerification.

- Trigger authClient.changeEmail
- Click on the link in the mail sent to the original email
The verification email sent to the new email shouldn't arrive because the sendVerificationEmail function crashes

### Current vs. Expected behavior

```sendVerificationEmail crashes because user is undefined:
 POST /api/auth/change-email 200 in 843ms
# SERVER_ERROR:  TypeError: Cannot read properties of null (reading 'email')
    at Object.sendVerificationEmail (src\lib\auth.ts:100:13)
```

(line 100): `to: user.email,`

Expected: user data should be present in the function

### What version of Better Auth are you using?

1.2.9

### Provide environment information

```bash
- OS: Windows 11
- Browser: Edge (don't judge)
- Node: v23.6.1
- Framework: Next.js (latest)
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
export const auth = betterAuth({
	user: {
		...options.user,
		changeEmail: {
			enabled: true,
			sendChangeEmailVerification: [...]
	},
	database: drizzleAdapter(db, {
		provider: "mysql",
	}),
	emailVerification: {
		sendOnSignUp: true,
		autoSignInAfterVerification: true,
		sendVerificationEmail: async ({ user, url, token }, _) => {
			console.log(user) // => null
		},
	},
	emailAndPassword: {
		enabled: true,
		requireEmailVerification: false,
		sendResetPassword
	},
})
```

Kinfe123:
how are you calling it on the client ?

TonybynMp4:
> how are you calling it on the client ?

```ts
// formData: { newEmail: string }
async function onSubmit(formData: NewUserEmailFormSchema) {
	const { data, error } = await authClient.changeEmail({
		newEmail: formData.newEmail,
		callbackURL: `${window.location.origin}/dashboard`,
	})

        [...]
}
```

TonybynMp4:
Can confirm it still happens (screenshot below from v1.3.18)!
It could be linked to custom session or something maybe? i'm open to share any details the team might need!

[...]

PR #5004, opened 2025-09-30
fix(adapter): use updated field values in WHERE clause during update

QuintenStr:
### Clearly describe what changes you made and why
I modified the logic that builds the `WHERE` clause during updates to ensure that if a field appears in both the `data` payload *and* the existing `where` conditions, the updated value is used for the lookup.

Previously, when updating a record and using fields that were also part of the `WHERE` filter (e.g. updating a unique identifier or key field), the query builder would still use the **old value** — potentially resulting in no record being matched or, worse, unexpected updates.

One known issue occurred with the MySQL-Drizzle adapter, where updatedUser returned null after sending a verification email when a verified user attempted to change their email address.

See: [sendVerificationEmail error during the email change process #3088
](https://github.com/better-auth/better-auth/issues/3088)

### Include any relevant context or background
/packages/better-auth/src/adapters/drizzle-adapter/drizzle-adapter.ts:
Previous behaviour:
```
const clause = convertWhereClause(where, model);

```
New behaviour:
```
// If we're updating a field that's in the where clause, use the new value
const updatedWhere = where.map((w) => {
	// If this field was updated, use the new value for lookup
	if (data[w.field] !== undefined) {
		return { ...w, value: data[w.field] };
	}
	return w;
});
const clause = convertWhereClause(updatedWhere, model);
```

### List any breaking changes or deprecations
No breaking changes found. All tests still run.

### Add screenshots for UI changes
No UI changes.

### Reference related issues or discussions
See: [sendVerificationEmail error during the email change process #3088
](https://github.com/better-auth/better-auth/issues/3088)
    

---

## Summary by cubic
Use updated field values in the WHERE clause during update operations in the Drizzle MySQL adapter. This prevents missed matches when changing key fields and resolves the email change flow bug where updatedUser could be null (#3088).


QuintenStr:
The test failing is in mongodb adapter, not my error. Please recheck

himself65 (changes requested):
test case is required

QuintenStr:
> test case is required

Where are the other tests located? I don't seem to be able to find any tests that use the "withReturning" before, hence why issue #3088 is an issue. I'm unsure where, and how to test this. I've manually built and packaged the changes, then imported the local better-auth package. This fixed the issue in 3088.

ping-maxwell (approved):
This is great!

PR #5013, opened 2025-10-01
fix: optimize ai streaming query

Kinfe123:


docs/components/floating-ai-search.tsx
```diff
@@ -310,77 +311,175 @@ export function AISearchTrigger() {
...
+				const decoder = new TextDecoder();
+				let accumulatedContent = "";
+
+				if (reader) {
+					while (true) {
+						const { done, value } = await reader.read();
+						if (done) break;
+
+						const chunk = decoder.decode(value, { stream: true });
+						accumulatedContent += chunk;
```
himself65, line 413: I think here you can use ReadableStream

himself65: nvm, I think it's not a big deal.

PR #5039, opened 2025-10-02
fix(sso): add foreign key references in schema

xiaoyu2er:
- Added foreign key reference from ssoProvider.organizationId to organization.id with cascade delete
- Added foreign key reference from ssoProvider.userId to user.id with set null on delete

see https://github.com/better-auth/better-auth/issues/4973


xiaoyu2er:
@Bekacru @himself65 could you help review the PR 🙏

himself65 (approved):

opswiz:
Yes. this is needed.

PR #5042, opened 2025-10-02
fix:`signUpEmail` provided dates will cause DB insert failure

ping-maxwell:
When creating an additional field with a type of `date`, then using the `authClient.signUp.email` with that new date field, the endpoint will fail saying it couldn't insert a user. The error will be `value.toIsoString` is not a function.

This is caused by the client sending dates to the server as a string (since it can't send Date type over HTTP). Normally, our endpoints which have Zod validation will automatically convert dates from string back to dates, however since the signUpEmail endpoint doesn't do this it causes bugs when inserting.

This PR fixes this and adds tests.

### DIFF
--- a/packages/better-auth/src/adapters/adapter-factory/index.ts
+++ b/packages/better-auth/src/adapters/adapter-factory/index.ts
@@ -16,7 +16,6 @@ import type {
 } from "./types";
 import { colors } from "../../utils/colors";
 import type { DBFieldAttribute } from "@better-auth/core/db";
-import { parseUserInput } from "../../db";
 export * from "./types";
 
 let debugLogs: { instance: string; args: any[] }[] = [];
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/adapters/adapter-factory/index.ts
```diff
@@ -343,6 +342,18 @@ export const createAdapterFactory =
...
 				let newValue = withApplyDefault(value, fieldAttributes!, action);
 
+				// In some endpoints (like signUpEmail) where there isn't proper Zod validation,
+				// we might recieve a date as a string (this is because of the client converting the Date to a string
+				// when sending to the server). Because of this, we'll convert the string to a Date.
+				if (
+					fieldAttributes &&
+					fieldAttributes.type === "date" &&
+					!(newValue instanceof Date) &&
+					newValue !== null
```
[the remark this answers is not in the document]
ping-maxwell: What are you saying?

packages/better-auth/src/adapters/adapter-factory/index.ts
```diff
@@ -340,6 +339,23 @@ export const createAdapterFactory =
...
+				if (
+					fieldAttributes &&
+					fieldAttributes.type === "date" &&
+					!(value instanceof Date) &&
+					typeof value === "string"
+				) {
+					try {
+						value = new Date(value);
+					} catch {
+						// Do nothing; If failed to convert, then it's not a date.
```
himself65: Should we do context.logger.error ?

packages/better-auth/src/adapters/tests/auth-flow.ts
```diff
@@ -171,3 +169,12 @@ function recoverProcessTZ() {
 		},
 	};
 }
+
+function convertSetCookieToCookie(headers: Headers): Headers {
```
Bekacru: we have a helper function for parsing set cookie headers. You can also use `cookieSetter` helper returned from `getTestInstance`

packages/better-auth/src/adapters/adapter-factory/index.ts
ping-maxwell: Hmm, yeah maybe

ping-maxwell:
@Bekacru ready for re-review.

PR #5054, opened 2025-10-02
docs: add password hashing step on auth0 and supabase migration guides

### LINKED ISSUE #5016: Auth0 migration: built‑in verification for migrated credential hashes ($2a/$2b bcrypt) fails

### To Reproduce

**Problem:** Following the [Auth0 migration guide](https://www.better-auth.com/docs/guides/auth0-migration-guide), we imported users credential with their original password Bcrypt hashes($2b) but better-auth don’t verify out-of-the-box; login returns error  `hex string expected got undefined`. To fix this custom emailAndPassword.password.verify logic is needed, which works but then password created using better-auth failed as it uses scrypt hashing algorithm.

```
  emailAndPassword: {
    enabled: true,
    password: {
      verify: async ({
        hash,
        password,
      }: {
        hash: string
        password: string
      }) => {
        if (!hash) return false
        if (hash.startsWith('$2a$') || hash.startsWith('$2b$')) {
          try {
            return await bcrypt.compare(password, hash)
          } catch {
            return false
          }
        }
        return false
      },
    },
```

Example data:
auth0User.password_hash: $2b$10$35ei8G8tZgZP.qNJi6d6mecPCTqtfIjtmy4P2/JwlFK36xOYHbKHe

Current: 
better-auth don’t verify out-of-the-box auth0 hashed password; On login returns error  `hex string expected got undefined`

Expected:
Better Auth  should verifies migrated Auth0 users without any custom code. It will be helpful if better-auth internally check if hash starts with $2b or $2a then use bcrypt otherwise scrypt.

1.3.4

### System info

```bash
better-auth : 1.3.4
Node: 20
Next: 15
```

Backend, Documentation

```typescript
  emailAndPassword: {  
    enabled: true,
  },
});
```

### Additional context

_No response_

Bekacru:
closes #5016

### DIFF
--- a/docs/content/docs/guides/auth0-migration-guide.mdx
+++ b/docs/content/docs/guides/auth0-migration-guide.mdx
@@ -556,6 +556,39 @@ Important considerations:
 </Callout>
 
 </Step>
+
+<Step>
+ ### Change password hashing algorithm
+
+ By default, Better Auth uses the `scrypt` algorithm to hash passwords. Since Auth0 uses `bcrypt`, you'll need to configure Better Auth to use bcrypt for password verification.
+
+ First, install bcrypt:
+
+ ```bash
+ npm install bcrypt
+ npm install -D @types/bcrypt
+ ```
+
+ Then update your auth configuration:
+
+ ```ts title="auth.ts"
+ import { betterAuth } from "better-auth";
+ import bcrypt from "bcrypt";
+ 
+ export const auth = betterAuth({
+    emailAndPassword: {
+        password: {
+            hash: async (password) => {
+                return await bcrypt.hash(password, 10);
+            },
+            verify: async ({ hash, password }) => {
+                return await bcrypt.compare(password, hash);
+            }
+        }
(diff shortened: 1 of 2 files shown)


docs/content/docs/guides/auth0-migration-guide.mdx
```diff
@@ -556,6 +556,39 @@ Important considerations:
...
 </Step>
+
+<Step>
+ ### Change password hashing algorithm
+
+ By default, Better Auth uses the `scrypt` algorithm to hash passwords. Since Auth0 uses `bcrypt`, you'll need to configure Better Auth to use bcrypt for password verification.
+
+ First, install bcrypt:
+
+ ```bash
```
himself65, line 567:
```package-install

Bekacru: ?

ping-maxwell:
If this is already tested and migration works then we should merge

PR #5057, opened 2025-10-02
feat: implement timing attack protection for authentication endpoints

himself65:


packages/better-auth/src/context/timing.ts
```diff
@@ -0,0 +1,22 @@
...
+ * This is used to avoid timing attacks by making sure that
+ * all endpoints take the same amount of time to execute.
+ *
+ * AuthContext -> [endpoint(string): duration(number)]
+ *
+ * Please note that we assume that AuthContext is a singleton for each auth instance.
+ *
+ * @internal
+ */
+const endpointTimingCache = new WeakMap<AuthContext, Map<string, number>>();
```
himself65, line 15: I think this is not working for serverless case

himself65:
I will do this in another PR

PR #5063, opened 2025-10-03
fix(organization): deleting member from org doesn't delete them from teams

ping-maxwell:
When calling auth.api.removeMember it's expected that they're removed from both the organization they're in, as well as any teams they could be in.

This applies to the `leaveOrganization` endpoint too.

- [x] tests

### DIFF
--- a/packages/better-auth/src/plugins/organization/adapter.ts
+++ b/packages/better-auth/src/plugins/organization/adapter.ts
@@ -301,8 +301,27 @@ export const getOrgAdapter = <O extends OrganizationOptions>(
 			});
 			return member;
 		},
-		deleteMember: async (memberId: string) => {
+		deleteMember: async ({
+			memberId,
+			organizationId,
+			userId: _userId,
+		}: {
+			memberId: string;
+			organizationId: string;
+			userId?: string;
+		}) => {
 			const adapter = await getCurrentAdapter(baseAdapter);
+			let userId: string;
+			if (!_userId) {
+				const member = await adapter.findOne<Member>({
+					model: "member",
+					where: [{ field: "id", value: memberId }],
+				});
+				if (!member) return;
+				userId = member?.userId;
+			} else {
+				userId = _userId;
+			}
 			const member = await adapter.delete<InferMember<O>>({
 				model: "member",
 				where: [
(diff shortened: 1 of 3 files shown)


packages/better-auth/src/plugins/organization/team.test.ts
```diff
@@ -362,7 +363,7 @@ describe("mulit team support", async (it) => {
 				}),
 			],
 			logger: {
-				level: "error",
+				level: "info",
```
himself65: what's the reason? but it's okay

himself65 (approved):
LGTM

packages/better-auth/src/plugins/organization/team.test.ts
ping-maxwell: fixed

ping-maxwell:
@himself65 Are we waiting on @Bekacru to review too to merge?

packages/better-auth/src/plugins/organization/adapter.ts
```diff
@@ -312,6 +331,22 @@ export const getOrgAdapter = <O extends OrganizationOptions>(
...
 				],
 			});
+			// remove member from all teams they're part of
+			if (options?.teams?.enabled) {
+				const teams = await adapter.findMany<Team>({
+					model: "team",
+					where: [{ field: "organizationId", value: organizationId }],
+				});
+				for (const team of teams) {
+					await adapter.deleteMany({
```
Bekacru: we probably want to use `Promise.all` here

packages/better-auth/src/plugins/organization/adapter.ts
```diff
@@ -301,8 +301,27 @@ export const getOrgAdapter = <O extends OrganizationOptions>(
...
+			userId?: string;
+		}) => {
 			const adapter = await getCurrentAdapter(baseAdapter);
+			let userId: string;
+			if (!_userId) {
+				const member = await adapter.findOne<Member>({
+					model: "member",
+					where: [{ field: "id", value: memberId }],
+				});
+				if (!member) return;
```
Bekacru: here you probably want to throw `APIError` instead

Issue #4198, opened 2025-08-24
[OIDC] Kysely + `CamelCasePlugin` choke with `redirectURLs`


julen:
## Background

Hi there 👋, we are using Kysely alongside its [`CamelCasePlugin`](https://kysely-org.github.io/kysely-apidoc/classes/CamelCasePlugin.html) to ensure our DB tables and columns are defined in snake case while the interaction via TS remains in camel case. This is something we do across our entire DB, and it also applies to the tables handled by better-auth.

While using the OIDC plugin, we created a new client and when going through the authorization flow, it'd always fail with a [_"Invalid redirect URI"_ error](https://github.com/better-auth/better-auth/blob/4696b9e7ec02a061af5473a414a99cbdb3bb67f8/packages/better-auth/src/plugins/oidc-provider/authorize.ts/#L121-L123).

## The Issue

Debugging has shown `oauthApplication.redirectURLs` comes as `undefined` from the DB, and this happens because by default the field name cannot be reliably translated both ways: the DB field `redirect_urls` gets mapped to `redirectUrls`, but the code expects a `redirectURLs` column.

## Potential Solutions

The `CamelCasePlugin` can be customized to work around this case:
* Use the `underscoreBetweenUppercaseLetters` option, s.t. `redirectURLs` becomes the `redirect_u_r_l_s` column in the DB. That'd allow the reverse mapping to work. Not good looking, but oh well.
* Customize the `CamelCasePlugin` to special-case this scenario. This is what we are currently doing:
```ts
class BetterAuthCompatibleCamelCasePlugin extends CamelCasePlugin {
	protected override camelCase(str: string): string {
		if (str === "redirect_urls") {
			return "redirectURLs";
		}

		return super.camelCase(str);
	}
}
```

Another possibility would be for better-auth to re-define `redirectURLs` as `redirecUrls` — and more broadly apply this principle to its core + plugin field names — use strict camel case, with no exceptions for acronyms.

## Test Case

I pushed a commit on my fork at https://github.com/julen/better-auth/commit/73a8de84f1d6af8aa7666257ed0ad262f71217f8 that showcases the issue in a test case.

## Is This a Bug?

I'm not sure this can be considered a bug per-se.

[...]

PR #5078, opened 2025-10-04
fix(adapter): Kysely with `CamelCasePlugin` breaks for OIDC.

ping-maxwell:
## TLDR
Kysely using their CamelCasePlugin breaks for OIDC due to one of the fields being defined as `redirectURLs`, which causes an inconsistency when generating the lowercase version.

## Breaking Changes
This does require all OIDC users to run migrations in order to update the field.

closes: https://github.com/better-auth/better-auth/issues/4198
linear: https://linear.app/better-auth/issue/ENG-180/fix-align-oidc-redirect-url-field-with-kysely-camelcase-plugin


FatherTimeSDKP:
gh pr checkout 5078

PR #5085, opened 2025-10-04
feat(admin): add `in` filterOperator to listUsers

jslno:
Blocked by: Bekacru/better-fetch#59

### DIFF
--- a/packages/better-auth/src/plugins/admin/admin.ts
+++ b/packages/better-auth/src/plugins/admin/admin.ts
@@ -640,9 +640,11 @@ export const admin = <O extends AdminOptions>(options?: O) => {
 							})
 							.or(z.number())
 							.or(z.boolean())
+							.or(z.string().array())
+							.or(z.number().array())
 							.optional(),
 						filterOperator: z
-							.enum(["eq", "ne", "lt", "lte", "gt", "gte", "contains"])
+							.enum(["eq", "ne", "lt", "lte", "gt", "gte", "in", "contains"])
 							.meta({
 								description: "The operator to use for the filter",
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/plugins/admin/admin.ts
```diff
@@ -713,11 +715,22 @@ export const admin = <O extends AdminOptions>(options?: O) => {
 						});
 					}
 
-					if (ctx.query?.filterValue) {
+					let filterValue = ctx.query.filterValue as string | number | boolean | undefined;
+					if (ctx.query.filterOperator === "in") {
```
ezeparziale: I suggest adding something like:

```typescript
          if (ctx.query.filterOperator === "in") {
            try {
              if (typeof filterValue === "string") {
                if (filterValue.startsWith("[")) {
                  filterValue = JSON.parse(filterValue)
                } else {
                  filterValue = filterValue.split(",").map((v) => v.trim())
                }
              }
            } catch {
              if (typeof filterValue === "string") {
                filterValue = filterValue.split(",").map((v) => v.trim())
              }
            }
            if (!Array.isArray(filterValue)) {
                message: "Value must be an array",
              })
            }
          } else if (filterValue === "true") {
            filterValue = true
          } else if (filterValue === "false") {
            filterValue = false
          }
```

This would add support for:
- multi-value filters (["admin","user"] or "admin,user")
- boolean filters (true / false)
```
filterField: "banned",
filterValue: false,
filterOperator: "eq"
```

I already have this working in my custom plugin, but it would be great if the official admin plugin supported it as well.

PR #5089, opened 2025-10-04
feat(magic-link): add handler function support for disableSignUp

GoPro16:
## Summary
Allow `disableSignUp` to accept a function that receives the endpoint context and returns a boolean. This enables conditional signup blocking based on request data, application state, or external conditions.

## Changes
- Updated `disableSignUp` option to accept `boolean | ((ctx: GenericEndpointContext) => boolean | Promise<boolean>)`
- Added support for both synchronous and asynchronous handler functions
- Function receives full endpoint context with access to `body`, `context`, and `request`
- Function is evaluated in both the send and verify endpoints
- Only queries the database when signup is actually disabled (efficient)

## Example Usage
```typescript
magicLink({
  sendMagicLink: async ({ email, url, token }) => {
    await sendEmail(email, url);
  },
  disableSignUp: (ctx) => {
    // Only allow signups from specific email domains
    const email = ctx.body.email;
    const allowedDomains = ["company.com", "partner.com"];
    const domain = email.split("@")[1];
    return !allowedDomains.includes(domain);
  }
})
```

Or with async logic:
```typescript
magicLink({
  disableSignUp: async (ctx) => {
    // Check feature flag or application state
    const signupsEnabled = await checkFeatureFlag("magic-link-signups");
    return !signupsEnabled;
  }
})
```

## Tests
- ✅ All existing tests pass
- ✅ Added comprehensive tests for function handler behavior
- ✅ Tests cover sync/async functions, context access, and edge cases
- ✅ Test verifies function is evaluated during verification flow
- ✅ TypeScript compilation successful with proper types

### DIFF
--- a/docs/content/docs/plugins/magic-link.mdx
+++ b/docs/content/docs/plugins/magic-link.mdx
@@ -59,19 +59,19 @@ To sign in with a magic link, you need to call `signIn.magicLink` with the user'
   method="POST"
   requireSession
 >
-  
+
 ```ts
 type signInMagicLink = {
     /**
-     * Email address to send the magic link. 
+     * Email address to send the magic link.
      */
     email: string = "user@email.com"
     /**
-     * User display name. Only used if the user is registering for the first time. 
+     * User display name. Only used if the user is registering for the first time.
      */
     name?: string = "my-name"
     /**
-     * URL to redirect after magic link verification. 
+     * URL to redirect after magic link verification.
      */
     callbackURL?: string = "/dashboard"
     /**
@@ -80,7 +80,7 @@ type signInMagicLink = {
     newUserCallbackURL?: string = "/welcome"
     /**
      * URL to redirect if an error happen on verification
-     * If only callbackURL is provided but without an `errorCallbackURL` then they will be 
+     * If only callbackURL is provided but without an `errorCallbackURL` then they will be
      * redirected to the callbackURL with an `error` query parameter.
      */
     errorCallbackURL?: string = "/error"
@@ -89,7 +89,7 @@ type signInMagicLink = {
 </APIMethod>
 
 <Callout>
-If the user has not signed up, unless `disableSignUp` is set to `true`, the user will be signed up automatically.
+If the user has not signed up, unless `disableSignUp` is set to `true` or a function that returns `true`, the user will be signed up automatically.
 </Callout>
 
 ### Verify Magic Link
@@ -111,11 +111,11 @@ If you want to handle the verification manually, (e.g, if you send the user a di
 ```ts
 type magicLinkVerify = {
     /**
-     * Verification token. 
+     * Verification token.
      */
     token: string = "123456"
     /**
-     * URL to redirect after magic link verification, if not provided will return the session. 
+     * URL to redirect after magic link verification, if not provided will return the session.
      */
     callbackURL?: string = "/dashboard"
 }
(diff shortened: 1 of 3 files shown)


GoPro16 (commented):
Sorry for the spam 😄

packages/better-auth/src/plugins/magic-link/magic-link.test.ts
```diff
@@ -303,3 +303,247 @@ describe("magic link storeToken", async () => {
...
+		expect(capturedEmail).toBe(newEmail);
+	});
+
+	it("should receive request context in disableSignUp function", async () => {
+		const disableSignUpFn = vi.fn((request: Request | undefined) => {
+			// Handle undefined request case
+			if (!request) return false;
+			const referer = request.headers.get("referer");
+			return referer?.includes("blocked-path") ?? false;
+		});
```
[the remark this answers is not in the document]
GoPro16: Thanks for the review\! Just to clarify: "referer" is actually the correct spelling for the HTTP header. It's a famous misspelling from the original HTTP specification that became standardized - the header is officially called "Referer" (not "Referrer"). See: https://en.wikipedia.org/wiki/HTTP_referer

docs/content/docs/plugins/magic-link.mdx
```diff
@@ -134,7 +134,38 @@ and a `request` object as the second parameter.
...
+```ts
+magicLink({
+    sendMagicLink: async ({ email, token, url }, request) => {
+        // send email
+    },
+    disableSignUp: (request) => {
+        if (!request) return false;
+        const referer = request.headers.get("referer");
+        return referer?.includes("/admin") ?? false;
+    }
```
himself65: From logical consideration, this is incorrect. You cannot trust data from http request, so disabling based on referer is insecure

GoPro16: Agreed there, can definitely change the example. The use case we have right now is more of a UX consideration than blocking a potential attacker from spoofing the referer. ie. we want to check if the user is signing up from our sign-up page and not our sign-in page as the same `signIn.magicLink` is used for both when `disableSignUp` is `false`.

There are definitely other workarounds we can add in for now like using the `hooks` but a larger example is the `async` one where we would want to disable signup across the board but then again it also could just be in `hooks`.

packages/better-auth/src/plugins/magic-link/index.ts
```diff
@@ -364,7 +381,17 @@ export const magicLink = (options: MagicLinkopts) => {
...
-						if (!opts.disableSignUp) {
+						// Handle disableSignUp - can be boolean or function
+						let signUpDisabled = false;
+						if (opts.disableSignUp) {
+							if (typeof opts.disableSignUp === "function") {
+								signUpDisabled = await opts.disableSignUp(ctx.request);
+							} else {
+								signUpDisabled = true;
+							}
+						}
```
Bekacru:
```suggestion
						const signUpDisabled = typeof opts.disableSignUp === "function" ? await opts.disableSignUp(ctx.request) : opts.disableSignUp
```

packages/better-auth/src/plugins/magic-link/index.ts
```diff
@@ -29,9 +29,16 @@ interface MagicLinkopts {
...
+	 * This allows you to conditionally disable signups based on request context.
+	 *
+	 * Note: The request may be undefined in some contexts. Make sure to handle this case.
+	 *
 	 * @default false
 	 */
-	disableSignUp?: boolean;
+	disableSignUp?:
+		| boolean
+		| ((request: Request | undefined) => boolean | Promise<boolean>);
```
Bekacru: I think you should pass the `ctx` object instead. That's one of the refactors we're planning to do for other similar callback options as well

`ctx: GenericEndpointContext`

docs/content/docs/plugins/magic-link.mdx
GoPro16: Updated the example to check blocked domain list instead along side the context feedback.
https://github.com/better-auth/better-auth/pull/5089/commits/5a48c8195a4c8dc81db9ae095a29eb828ec588e9

gustavovalverde:
@GoPro16 can you update this PR? Please ping me when you've done it :)

GoPro16:
@gustavovalverde I am actually going to close this PR as the original intention that I had for this was correct but implemented at the wrong level. Reason being that if we are trying to "block" domains typically we would not want to expose as a raw error of "signup is disabled" or "email is blocked" because of an attacker gaining insight easily that the policy exists — and from there enumerating which domains/emails the policy applies to. We also pivoted to using email-otp and handling similar logic in a before hook.

PR #5094, opened 2025-10-05
fix: getAcccessToken refresh should properly refresh when oauth tokens are encrypted 

bsklaroff:
If the encryptOAuthTokens option is set to true, then the token refresh behavior within the getAccessToken route was broken in two ways:
1. The refreshToken needed to be decrypted before calling provider.refreshAccessToken
2. Once (1) was fixed, the new accessToken was returned from provider.refreshAccessToken, and saved to newTokens. But then, the code was calling decryptOAuthTokens on it, even though it's not encrypted, which was causing an error

This PR fixes both issues, and getAccessToken now works property when the encryptOAuthTokens option is set
    

---

## Summary by cubic
Fixes getAccessToken to correctly handle encrypted OAuth tokens during refresh. Prevents errors when encryptOAuthTokens=true and restores working token refresh.

- **Bug Fixes**
  - Decrypt refreshToken before calling provider.refreshAccessToken and save refreshed tokens via setTokenUtil.
  - Do not decrypt the provider’s new accessToken; only decrypt the stored accessToken when no refresh occurs.

### DIFF
--- a/packages/better-auth/src/api/routes/account.ts
+++ b/packages/better-auth/src/api/routes/account.ts
@@ -538,9 +538,11 @@ export const getAccessToken = createAuthEndpoint(
 				accessTokenExpired &&
 				provider.refreshAccessToken
 			) {
-				newTokens = await provider.refreshAccessToken(
-					account.refreshToken as string,
+				const refreshToken = await decryptOAuthToken(
+					account.refreshToken,
+					ctx.context,
 				);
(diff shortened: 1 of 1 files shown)


ElasticBottle:
Ran into exactly this issue, can confirm that running the preview build here fixes the issue

ElasticBottle:
@Bekacru any idea when this will land in main?

Bekacru (approved):

PR #5096, opened 2025-10-05
fix(passkey): atom listeners not working

ping-maxwell:
closes: https://github.com/better-auth/better-auth/issues/858
linear: https://linear.app/better-auth/issue/ENG-17/fix-usesession-does-not-update-after-passkey-authentication

This PR doesn't just fix the passkey bugs with Atom listeners, but also addresses some major flaws in the client atom listeners.

## Fixes
* the client path proxy doesn't find all successful `atomListeners.matcher` cases, instead it only finds 1 matched atomLister, and calls that. This means if multiple atom listeners are trying to watch the same path (e.g. `/sign-out`) only the first plugin which passes the matcher for that path will be called, all other plugins will be ignored.
* While debugging this issue, I noticed the atomListener for the passkey plugin was using the signal `_listPasskeys` instead of `$listPasskeys`. Not sure if that's intentional or not, but I'm pretty sure it's been like this for months now. So everyone's `usePasskeyList` hook only partially work. (I say partially because there is 1 case where we forcefully call the listener, but besides that no cases will call the listPasskeys hook)
* Initially, this PR was made to resolve not updating the session when passkey sign in was successful, however along the way I also added other fixes specific to passkey atoms:
  * fixed `/sign-out` not re-triggering the `listPasskeys` atom. 
  * And lastly - the reason this PR was made in the first place - fixed session not updating when passkey sign in succeeded.
    

## Summary by cubic
Fixes atom listeners for passkey flows so the UI refreshes the passkey list and session state after registration, authentication, update, delete, and sign-out.

- **Bug Fixes**
  - Proxy now triggers all matching atom listeners (not just the first), with guards for missing atoms and a safe async toggle to prevent race conditions.
  - Passkey client corrects signal name to "$listPasskeys", adds "$sessionSignal" listener, includes "/sign-out" and "/passkey/verify-authentication" routes, and notifies both signals after passkey sign-in.


packages/better-auth/src/plugins/passkey/client.ts
```diff
@@ -55,6 +57,8 @@ export const getPasskeyActions = (
 				...options,
 				method: "POST",
 			});
+			$listPasskeys.set(Math.random());
```
himself65, line 60: What's the trick here?

himself65 (changes requested):
Looks good.

packages/better-auth/src/plugins/passkey/client.ts
ping-maxwell: That was there already, but essentially it's a signal, you set it to anything and it will trigger the `useListPasskey` hook

himself65 (approved):

packages/better-auth/src/plugins/passkey/client.ts
himself65: So it's basically a `rerender` call

PR #5101, opened 2025-10-05
feat: add getPlugin helper

jslno:
### Basic Usage

```ts
const adminPlugin = getPlugin<ReturnType<typeof admin<any>>>(
	"admin",
	ctx.context,
);

adminPlugin?.endpoints.listUsers({
	query: {},
	headers,
});
```


himself65 (changes requested):
This is a good idea, but I think we will make this in another way when I refactor things to the core module

PR #5128, opened 2025-10-06
fix(expo): remove local cache

### LINKED ISSUE #2493: Corrupted/malformed headers with Expo + Next.js break session retrieval

### To Reproduce

Find a full reproduction repo with instructions at https://github.com/subvertallchris/BetterAuthExpoNextBroken/

Current: when using Expo + Next.js plugins, it appears that refresh behavior around `cookieCache` is getting a new `better_auth.session_data` value that it holds in memory and seemingly combines with `fetch` (!?) preventing serverside session verification.

Expected: it should not do this.

1.27

```bash
- Mac 14.3.1
- Expo 52.0.46
- iOS simulator 15.2 running iOS 17.2
```

Backend, Client

```typescript
You'll find it in the reproduction.
```

### Additional context

We are using BetterAuth on the web and it is flawless, this problem does not exist anywhere other than iOS. We have not tested this with Android or physical device yet.

Bekacru:
Avoid caching session data locally

closes #2493

### DIFF
--- a/packages/expo/src/client.ts
+++ b/packages/expo/src/client.ts
@@ -75,7 +75,6 @@ interface ExpoClientOptions {
 		getItem: (key: string) => string | null;
 	};
 	storagePrefix?: string;
-	disableCache?: boolean;
 }
 
 interface StoredCookie {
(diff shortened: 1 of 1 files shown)


Bekacru:
the cache is unnecessary in the first place

PR #5136, opened 2025-10-07
fix(oidc): early cleanup prompt cookie

### LINKED ISSUE #4594: OIDC plugin: oidc_login_prompt not cleared after login, causing main app logins to redirect to client

### To Reproduce

1. Start an OIDC login from a client app (RP) so the OP sets `oidc_login_prompt` cookie.
2. Complete login successfully (session is created).
3. Later visit the main app’s `/login` and sign in normally.
4. You’re redirected to the OIDC client app instead of staying in the main app.

**Current:** The `oidc_login_prompt` cookie remains after a successful login, causing subsequent main-app logins to redirect to the OIDC client.

**Expected:** The plugin clears `oidc_login_prompt` upon successful login so normal main-app logins land in the main app; the cookie should only influence an active OIDC flow.

1.3.9

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64"
  },
  "node": {
    "version": "v22.18.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "10.15.1"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "1.3.9",
    "config": null
  }
}
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

Bekacru:
closes #4594

### DIFF
--- a/packages/better-auth/src/plugins/mcp/index.ts
+++ b/packages/better-auth/src/plugins/mcp/index.ts
@@ -141,6 +141,9 @@ export const mcp = (options: MCPOptions) => {
 							"oidc_login_prompt",
 							ctx.context.secret,
 						);
+						ctx.setCookie("oidc_login_prompt", "", {
+							maxAge: 0,
+						});
 						const cookieName = ctx.context.authCookies.sessionToken.name;
 						const parsedSetCookieHeader = parseSetCookieHeader(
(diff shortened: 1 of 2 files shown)


himself65:
See https://github.com/better-auth/better-auth/pull/5912

Issue #4728, opened 2025-09-17
Custom fields with Magic Links/Social Sign Up


JE4GLE:
I have a custom field referredBy with my user, which stores the referral code they used to sign up. With email/password signup, I use it as follows:
```typescript
authClient.signUp.email(
  {
    email: data.username,
    password: data.password,
    referredBy: "my-referral-code"
  }
)
```

Better Auth v1.3.11

I'm unsure on how to populate that field on sign up, when signin up using either **magic links** or **social sign-ups**. I don't know if this isn't implemented yet or if if I'm missing something.

JE4GLE:
@dosu I would relabel it then as a feature request.

Bekacru:
THis should work with additional fields. Can you share you auth config and you auth client config as well?

micnil:
I want this feature to set the username of users when signing up. It makes the system simpler if I can rely on that the username is always set on all users, instead of prompting the user to set it after signing up.

If it makes the feature easier to implement I would propose to **enable this for social signup with idToken**. This is how chess.com seems to handle it for instance:

<img width="2656" height="1730" alt="Image" src="https://github.com/user-attachments/assets/09b307ca-3ff1-44ad-bf5e-de707e702f4a" />

The client api could perhaps be:

```
authClient.signIn.social({
  provider: 'google',
  idToken: {
    token,
    accessToken,
  },
  additionalFields: {
    username: 'foo123', // Example of sending additional fields
  }
});
```

Tranthanh98:
same problem, any update ? I really want to have this feature

kno-raziel:
> I want this feature to set the username of users when signing up. It makes the system simpler if I can rely on that the username is always set on all users, instead of prompting the user to set it after signing up.
> 
> If it makes the feature easier to implement I would propose to **enable this for social signup with idToken**. This is how chess.com seems to handle it for instance:
> 
[...]

PR #5140, opened 2025-10-07
feat(magic-link): require name for signup

### LINKED ISSUE #5139: Add requireNameForSignUp option to Magic Link plugin

Yes. Currently, the Magic Link plugin allows new users to be created without providing a name. In some applications, having a name is required for new accounts, and allowing sign-ups without it can lead to incomplete user profiles or inconsistent data. There is no existing option to enforce this requirement during sign-up, which makes it difficult to ensure that all new users provide a name when registering via Magic Link.

I would like a new configuration option for the Magic Link plugin, called `requireNameForSignUp`. When set to true, any sign-in attempt that triggers the creation of a new user must include a non-empty name field in the request. If the name is missing or empty, the sign-up should fail gracefully with an appropriate error, without exposing whether the email already exists (to prevent user enumeration). This ensures that all new users created via Magic Link have a name.

I have read the recommended solution in [this comment](https://github.com/better-auth/better-auth/issues/4728#issuecomment-3304537808), which suggests handling user creation flows carefully. While this approach works, having native support for requiring the `name` field during Magic Link signup would simplify validation and prevent inconsistent user creation.

### Additional context

_No response_

ouariachi:
**What changes were made and why:**

* Added a new configuration option `requireNameForSignUp` in the Magic Link plugin.
* When enabled, any attempt to create a new user via Magic Link without a non-empty `name` field will fail.
* This ensures new users always have a `name`, improving data consistency and preventing incomplete user records.
* Updated the plugin documentation to include the new requireNameForSignUp option and describe its behavior.

**Relevant context:**

* Currently, Magic Link signup allows creating users without a name.
* The recommended workaround (from [issue comment](https://github.com/better-auth/better-auth/issues/4728#issuecomment-3304537808)) was to collect `name` later, but adding native support avoids extra steps and simplifies the flow.

**Breaking changes / deprecations:**

* No breaking changes.
* Flows without `requireNameForSignUp` enabled remain unchanged.

**Tests / screenshots:**

* Added test to verify that signup fails if `name` is missing when `requireNameForSignUp` is true.

**Related issues / discussions:**

* Closes 
#5139

### DIFF
--- a/packages/better-auth/src/plugins/magic-link/index.ts
+++ b/packages/better-auth/src/plugins/magic-link/index.ts
@@ -59,6 +59,20 @@ interface MagicLinkopts {
 		| "plain"
 		| "hashed"
 		| { type: "custom-hasher"; hash: (token: string) => Promise<string> };
+
+	/**
+	 * Require the "name" field when creating a new user via magic link.
+	 *
+	 * If set to true, sign-in attempts that trigger user creation
+	 * (i.e. when the email is not yet registered) will fail unless
+	 * the request body includes a non-empty `name` value.
+	 *
+	 * If `disableSignUp` is set to true, this option has no effect.
+	 *
+	 * @see {@link MagicLinkopts.disableSignUp}
+	 * @default false
+	 */
+	requireNameForSignUp?: boolean;
 }
 
 export const magicLink = (options: MagicLinkopts) => {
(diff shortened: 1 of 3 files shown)


packages/better-auth/src/plugins/magic-link/index.ts
```diff
@@ -365,6 +379,15 @@ export const magicLink = (options: MagicLinkopts) => {
 
 					if (!user) {
 						if (!opts.disableSignUp) {
+							if (
+								opts.requireNameForSignUp &&
+								(typeof name !== "string" || name.trim() === "")
+							) {
+								throw ctx.redirect(
+									`${errorCallbackURL}?error=NAME_REQUIRED_FOR_SIGNUP`,
```
ping-maxwell: make the error code lowercase to keep consistency with others

ping-maxwell (changes requested):

ping-maxwell:
I think you forgot to update your tests to use the lowercase error name, yet CI didn't fail. Is your test not working?

ouariachi:
> I think you forgot to update your tests to use the lowercase error name, yet CI didn't fail. Is your test not working?

Yes, I was checking the error code in the signIn response when it should be in the verify response.

bytaesu:
Hi @ouariachi,

We're considering making the name field optional in the next major release, and will work on it in the PR below 🙂

- https://github.com/better-auth/better-auth/pull/7165

PR #5154, opened 2025-10-07
docs: calls out changes to cookie name when the cookie is secure with code sample

j23schoen:
I made a few tweaks to the docs around cookies and the name of the session cookie. My team shipped something that relied on the cookie name being what it is in local development, only to find that in prod the cookie is secure and prefixed with `__Secure-`.

I updated the docs in two spots and called this out. I did some local testing to verify the effects that updating the cookie prefix and cookie name had, but the `__Secure-` persists as long as `useSecureCookies` is true or you meet the other criteria mentioned for secure cookies.

### DIFF
--- a/docs/content/docs/concepts/cookies.mdx
+++ b/docs/content/docs/concepts/cookies.mdx
@@ -94,3 +94,18 @@ export const auth = betterAuth({
     }
 })
 ```
+
+*Note: The name of your cookie will automatically be prefixed with `__Secure-`. Even if you specify your own prefix. For example:*
+
+```ts title="cookie.ts"
+import { betterAuth } from "better-auth"
+
+export const auth = betterAuth({
+    advanced: {
+        cookiePrefix: 'make-believe'
+        useSecureCookies: true
+    }
+})
+
+# produces a cookie with this name: __Secure-make-believe.session_token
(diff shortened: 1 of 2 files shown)


docs/content/docs/concepts/cookies.mdx
```diff
@@ -93,4 +93,19 @@ export const auth = betterAuth({
         useSecureCookies: true
     }
 })
+```
+
+*Note: The name of your cookie will automatically be prefixed with `__Secure-`. Even if you specify your own prefix. For example:*
```
himself65: use callout?

j23schoen: sweet will do, didn't know about that

Bekacru:
good callout but this make it sound like we're doing arbitrarily, it's a browser requirement for secure cookies to have __secure prefix

j23schoen:
> good callout but this make it sound like we're doing arbitrarily, it's a browser requirement for secure cookies to have __secure prefix

ah, TIL. I can tweak the wording a bit if you think it'd still be valuable to call it out

himself65:
__Secure is a browser forced prefix, it's not related to better auth

Issue #5111, opened 2025-10-06
Release 1.3.25 Breaking Changes


l0gicgate:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Test suite with Jest started failing after updating to 1.3.25:

```
  ● Test suite failed to run

    TypeError: A dynamic import callback was invoked without --experimental-vm-modules

      at createTelemetry (node_modules/better-auth/dist/shared/better-auth.Cn-kIXzo.cjs:34:5)
      at init (node_modules/better-auth/dist/shared/better-auth.Cn-kIXzo.cjs:97:29)
```

This PR is what broke things:
https://github.com/better-auth/better-auth/pull/5007

### Current vs. Expected behavior

It should not break.

Reverting to 1.3.24 everything works fine.

### What version of Better Auth are you using?

1.3.26

### System info

```bash
Nothing comes out of that command. It's broken.
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  telemetry: { enabled: false },
});
```

jonaspm:
I hit the same issue, the breaking change does not work in Cloudflare Workers, had to rollback to v1.3.24

himself65:
We have removed the dynamic import in 1.3.27. 

https://github.com/better-auth/better-auth/commit/e704398a250b76bf781d8f4232b6b63c19b375b4
https://github.com/better-auth/better-auth/commit/721326c3b3d3ab919564141ea862e1231c411278

PR #5161, opened 2025-10-08
fix: telemetry lazy loading issues

Kinfe123:


himself65:
https://github.com/better-auth/better-auth/issues/5111#issuecomment-3382481614

Now I think dynamic import will make things complex. We should deprotrize this task since it's not a super big issue. Let's use conditional exports to ignore the serverless runtime.

Kinfe123:
But still the dynamic import something that we are still using to solve those issue and specially when it comes to lazily loading still have an effect.

Issue #5045, opened 2025-10-02
Support Linear app-actor OAuth flow (actor=app)


codybrouwers:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

The Better Auth Linear provider currently only supports user-centric OAuth flows.

Linear also supports **app-actor authorization** (`actor=app` in the authorization URL), which mints tokens that act as the application itself (not a user).  
This flow is required for service accounts, agents, and automations, but the existing better-auth provider does not currently expose a way to request it.

Without support for `actor=app`:
- All tokens are user-scoped.
- Integrations cannot perform mutations (issue creation, comments, status changes) as the app.  
- Many automation and agent use cases are blocked.  

References:  
- [Linear OAuth Actor Authorization](https://linear.app/developers/oauth-actor-authorization)  
- [Linear provider source](https://github.com/better-auth/better-auth/blob/778bf4043da54f03ac8c696cf8ae2549920ec78e/packages/better-auth/src/social-providers/linear.ts)  
- [Authorization URL builder](https://github.com/better-auth/better-auth/blob/778bf4043da54f03ac8c696cf8ae2549920ec78e/packages/better-auth/src/oauth2/create-authorization-url.ts)  

### Describe the solution you'd like

1. **Authorization param support**  
   - Option A: Add a generic `authorizationParams?: Record<string,string>` forwarded to `createAuthorizationURL`.  
   - Option B: Add a Linear-specific option (e.g. `actor?: 'app' | 'user'`).  

2. **Support for non-user “app actor” flows**  
   - Make the provider resilient to tokens without a user profile.  
   - Allow skipping user linking and return raw OAuth credentials.  
   - Alternatively, expose a variant of the provider that does not assume user identity.  

3. **Documentation**  
   - Explain how to enable `actor=app`.  
   - Clarify the implications (no user identity).  
   - Provide example usage for service accounts and automations.  

### Describe alternatives you've considered

Currently I have written a custom OAuth2 provider with Better Auth primitives that enables me to do this but I would love to take advantage of Better Auth instead.

PR #5164, opened 2025-10-08
feat(social-provider): linear app actor oauth support

Kinfe123:
## Summary by cubic
Adds support for Linear OAuth “actor” to enable app-level tokens for service accounts and automation. Handles tokens without a user profile by returning null instead of creating/linking a user.

- **New Features**
  - Add actor option to LinearOptions ("app" | "user").
  - Include actor in the authorization URL when provided.
  - Return null when app-actor tokens lack a viewer to prevent user creation/linking.
  - Backward compatible; default behavior unchanged.

Addresses Linear issue #5045.

### DIFF
--- a/packages/core/src/social-providers/linear.ts
+++ b/packages/core/src/social-providers/linear.ts
@@ -50,10 +50,7 @@ export const linear = (options: LinearOptions) => {
 				state,
 				redirectURI,
 				loginHint,
-				additionalParams:
-					Object.keys(additionalParams).length > 0
-						? additionalParams
-						: undefined,
+				additionalParams: additionalParams,
 			});
 		},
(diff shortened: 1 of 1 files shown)


packages/core/src/social-providers/linear.ts
```diff
@@ -96,7 +104,12 @@ export const linear = (options: LinearOptions) => {
 					}),
 				},
 			);
+
+			// Handle app-actor tokens that may not have a user profile
 			if (error || !profile?.data?.viewer) {
+				// For app-actor flows, we might not have a user profile
```
Bekacru, line 110: This means the oauth flow will fail

Kinfe123: It is up on their config on setting that up.

Issue #3236, opened 2025-07-01
Bad Request when upgrading subscription with identical plan/seats from monthly to yearly


MikeCodeur:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

The `authClient.subscription.upgrade()` function allows upgrading between monthly and yearly billing modes, and changing the number of seats without issues. However, when attempting to upgrade a subscription with the same plan and same number of seats but changing from monthly to yearly (or vice versa), the API returns a Bad Request error.

## Steps to Reproduce

1. **Create a monthly subscription**
   - Create a Pro subscription with 5 seats in monthly mode
   - Verify the subscription is active

2. **Attempt to upgrade to yearly with same plan/seats**
   - Use `authClient.subscription.upgrade()` with:
     - Same plan (Pro)
     - Same number of seats (5)
     - `annual: true` (changing from monthly to yearly)
   - Observe that the API returns a Bad Request error

3. **Verify other upgrade scenarios work**
   - Change seats only (monthly, 5 → 10 seats): Works
   - Change plan only (Pro → Enterprise, monthly): Works
   - Change billing mode only (monthly → yearly, different seats): Works

## Expected Behavior

- The upgrade should succeed when changing only the billing mode (monthly ↔ yearly)
- The API should accept the request even if plan and seats remain identical
- The subscription should be updated with the new billing cycle

## Actual Behavior

- The API returns a Bad Request error
- The upgrade fails when plan and seats are identical but billing mode changes
- Other upgrade scenarios (different plans, different seats) work correctly

### What version of Better Auth are you using?

1.2.12

### Provide environment information

```bash
-WSL2
```

### Which area(s) are affected? (Select all that apply)

Client

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});
```

PR #5167, opened 2025-10-08
fix(stripe): allow subscription upgrade between different intervals for a plan

### LINKED ISSUE #3393: [Stripe] User can't update same product subscription

## How to reproduce 

- create on stripe.com a product with monthly and yearly payment
- init stripe plugin with this basic type of plan: 
```
const stripePlan = [
  {   
      name: "my-product',
      priceId: "monthy_priceId",
      annualDiscountPriceId: "yearly_priceId",
      ....
  }
]
```
- purchase the plan on monthly
```
await authClient.subscription.upgrade({
          plan,
          referenceId: subscription.referenceId,
          annual: false, 
          ...
        });
```
- update the plan to yearly
```
await authClient.subscription.upgrade({
          plan,
          subscriptionId: subscription.stripeSubscriptionId,
          annual: true,
          ...
        });
```
- you should get an error 400 with the following code: "YOURE_ALREADY_SUBSCRIBED_TO_THIS_PLAN"  

### Current

We can't have a single stripe product with monthly and yearly payment under it. Workaround is to create one product per type of payment (ex: my_product_monthly, my_product_yearly)

### Expected

It should be possible to change the payment price under the same plan name

1.2.12

```bash
- Os: MacOs
```

Package

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

Issue coming from : 
https://github.com/better-auth/better-auth/blob/0a8c8d2bf4d06548f5efa1cd4e1504bc83ad533b/packages/stripe/src/index.ts#L309 
which doesn't compare the price_idk

Maybe we should store the price_id the user subscribed or an isYearlyPayment boolean

andrewdoro:
This PR enables upgrades for subscriptions on the same plan between different billing intervals. (annually/monthly)

- Deduplicates priceId checks for lookup keys and moves them before `ALREADY_SUBSCRIBED_PLAN` check
- Implements checks for `priceId` by using the active priceId from the current active stripe subscription and the future priceId of the upgrade.
- Adds test for switching to annually for same plan

fixes #3393 #3236
    

---

## Summary by cubic
Allows upgrading a subscription between monthly and annual intervals on the same plan. Prevents false “already subscribed” errors by comparing Stripe price IDs correctly.

- **Bug Fixes**
  - Resolve price ID (including lookup keys) before the already-subscribed check and deduplicate logic.
  - Compare the active Stripe subscription’s price ID to the target price ID to allow interval switches.
  - Add test covering monthly-to-annual upgrade on the same plan.

### DIFF
--- a/packages/stripe/src/index.ts
+++ b/packages/stripe/src/index.ts
@@ -379,7 +379,7 @@ export const stripe = <O extends StripeOptions>(options: O) => {
 				);
 
 				// Resolve price ID if using lookup keys
-				let priceIdToUse: string | undefined = undefined;
+				let priceIdToUse: string | undefined;
 				if (ctx.body.annual) {
 					priceIdToUse = plan.annualDiscountPriceId;
 					if (!priceIdToUse && plan.annualDiscountLookupKey) {
--- a/packages/stripe/src/stripe.test.ts
+++ b/packages/stripe/src/stripe.test.ts
@@ -902,7 +902,6 @@ describe("stripe", async () => {
 		);
 
 		const headers = new Headers();
-
 		await authClient.signIn.email(
 			{
 				...testUser,


andrewdoro:
hey @himself65 lmk if you need any new tests for this

we are currently updating our @anaralabs stripe subscription handling to the better auth one

slig:
+1. This is a must have for me as well, and I believe to a lot of indie devs that have just one plan with two prices (monthly/yearly).

slig:
Hi @himself65, could you please take a look at this again? Thank you!

Hi @andrewdoro , did you manage to work around this for your project? Thanks!

andrewdoro:
hey @slig I've used my fork to build the a local better auth and then patched the dep with pnpm

bytaesu:
Hi @andrewdoro,

Thanks for you work 🙂

In the meantime, the Stripe plugin has been improved a lot, so there are many conflicts.. I’ll handle this!

PR #5173, opened 2025-10-08
docs: clarify generateId behavior and mixed ID types

### LINKED ISSUE #5081: Documentation: Clarify generateId behavior and mixed ID type handling
## Summary

The current documentation for `advanced.database.generateId` is unclear about how to handle mixed ID types (e.g., integer IDs for users, UUID strings for other tables). This led to confusion during implementation.

### 1. Returning `false` from generateId doesn't work as documented

The documentation suggests you can return `false` from the `generateId` function to let the database handle ID generation:

```ts
generateId: (options) => {
  if (options.model === 'user') {
    return false; // ❌ This tries to insert "false" as a string!
  }
  return "my-id";
}
```

**Error encountered:** `error: invalid input syntax for type integer: "false"`

**Solution:** Return `undefined` instead:

```ts
generateId: (options) => {
  if (options.model === 'user' || options.model === 'users') {
    return undefined as any; // ✅ Let database generate the ID
  }
  return crypto.randomUUID();
}
```

Looking at the source code (`better-auth.BrEH5bcK.mjs`), the check is:
```js
if (generateId$1 === false || useNumberId) return void 0;
```

This checks if the *config value* is `false`, not if the *function returns* `false`.

### 2. Mixed ID Types Not Documented

**Scenario:** Users table with PostgreSQL `serial` (integer) IDs, but session/account/verification tables with `text` (UUID) IDs.

The documentation doesn't explain:
- That `useNumberId` is a global setting affecting all tables
- How to handle mixed ID types across different tables
- That you should NOT use `useNumberId` when mixing types

### 1. Update `advanced.database.generateId` section

```ts
database: {
  // Option 1: Disable ID generation for ALL tables
  generateId: false,
  
  // Option 2: Custom ID generation per model
  generateId: (options) => {
    // Let database auto-generate (for serial, auto_increment, etc.)
    if (options.model === 'user' || options.model === 'users') {
      return undefined as any;
    }
    
    // Generate UUIDs for other tables
    return crypto.randomUUID();
  },
}
```

**Important:** To let the database generate IDs, return `undefined` from the function. Setting `generateId: false` disables generation for ALL tables.

### 2. Add Mixed ID Types Example

```ts
// Example: Integer IDs for users, UUID strings for other tables
  user: {
    modelName: 'users', // PostgreSQL: id serial primary key
  },
  session: {
    modelName: 'session', // PostgreSQL: id text primary key
  },
  advanced: {
    database: {
      // Do NOT set useNumberId - it's global and affects all tables
      generateId: (options) => {
        if (options.model === 'user' || options.model === 'users') {
          return undefined as any; // Let PostgreSQL serial generate it
        }
        return crypto.randomUUID(); // UUIDs for session, account, verification
      },
    },
  },
})
```

### 3. Document `additionalFields.returned` property

```ts
user: {
  additionalFields: {
    firstName: {
      type: 'string',
      fieldName: 'first_name',
      returned: true, // ✅ Required to include field in SELECT queries
    },
  },
}
```

Without `returned: true`, fields are stored but not retrieved from the database.

### 4. Clarify Database Hooks Naming

```ts
databaseHooks: {
  user: {
    create: {
      before: async (user) => {
        // ✅ Use JavaScript property names, not database column names
        return {
          data: {
            ...user,
            firstName: 'John', // Better Auth maps this to first_name
          },
        };
      },
    },
  },
}
```

## Additional Context

- Database: PostgreSQL with Kysely adapter
- Use case: Migrating from Clerk to better-auth with existing schema
- Constraint: Cannot change users table from integer to text IDs (many dependencies)

Would be happy to submit a PR with these documentation improvements if that would be helpful

ahmed-abdat:
Clarifies the documentation for `generateId` configuration and mixed ID type handling, addressing confusion around when to use `generateId: false` vs returning `false`/`undefined` from a function.

Closes #5081
    

---

## Summary by cubic
Clarifies how generateId works and how to mix ID types safely, with examples and warnings. Documents additionalFields.returned and fieldName, and database hooks property naming, closing #5081.

- **Migration**
  - Set generateId: false to let the database generate IDs for all tables.
  - For mixed ID types, use a generateId function; return false/undefined for DB-generated models and generate UUIDs for others. Do not use useNumberId.
  - If a field uses a different fieldName than its property, set returned: true to include it in SELECT queries.
  - In databaseHooks, use JavaScript property names (e.g., firstName), not database column names.

### DIFF
--- a/docs/content/docs/concepts/database.mdx
+++ b/docs/content/docs/concepts/database.mdx
@@ -473,9 +473,9 @@ export const auth = betterAuth({
 
 Better Auth by default will generate unique IDs for users, sessions, and other entities. You can customize ID generation behavior using the `advanced.database.generateId` option or `advanced.database.useNumberId`.
 
-#### Option 1: Let Database Generate IDs
+#### Option 1: Auto-Increment Numeric IDs
 
-Setting `generateId` to `false` disables ID generation globally, letting your database handle all ID generation:
+For auto-incrementing numeric IDs (serial, auto_increment) across all tables, use `useNumberId`:
 
 ```ts title="auth.ts"
 import { betterAuth } from "better-auth";
@@ -485,19 +485,37 @@ export const auth = betterAuth({
   database: db,
   advanced: {
     database: {
-      generateId: false, // Database generates all IDs
+      useNumberId: true,
     },
   },
 });
 ```
 
 <Callout type="info">
-  **Note**: For auto-incrementing numeric IDs (serial, auto_increment), use `useNumberId: true` instead (see the Numeric IDs section below).
+  **How it works:** `useNumberId` automatically disables Better Auth ID generation and configures the system to expect numeric IDs from your database. The Better Auth CLI will generate schema with numeric ID fields with auto-incrementing attributes (serial, auto_increment, etc.). IDs are returned as strings in Better Auth responses.
 </Callout>
 
-#### Option 2: Custom ID Generation Function
+#### Option 2: Database-Generated IDs
 
-Use a function to generate IDs. You can return `false` or `undefined` from the function to let the database generate the ID for specific models:
+If your database generates IDs using functions (like PostgreSQL `gen_random_uuid()`) or other mechanisms, set `generateId` to `false`:
+
+```ts title="auth.ts"
+import { betterAuth } from "better-auth";
+import { db } from "./db";
+
+export const auth = betterAuth({
+  database: db,
+  advanced: {
+    database: {
+      generateId: false,
+    },
+  },
+});
+```
+
+#### Option 3: Custom ID Generation Function
+
+Use a function to generate IDs with custom logic. You can return `false` or `undefined` from the function to let the database generate the ID for specific models:
 
 ```ts title="auth.ts"
 import { betterAuth } from "better-auth";
@@ -524,7 +542,7 @@ export const auth = betterAuth({
   **Important**: Returning `false` or `undefined` from the `generateId` function lets the database handle ID generation for that specific model. Setting `generateId: false` (without a function) disables ID generation for **all** tables.
 </Callout>
 
-#### Option 3: Consistent Custom ID Generator
+#### Option 4: Consistent Custom ID Generator
 
 Generate the same type of ID for all tables:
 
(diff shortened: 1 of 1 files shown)


docs/content/docs/concepts/database.mdx
```diff
@@ -388,6 +388,8 @@ To add custom fields, use the `additionalFields` property in the `user` or `sess
 - `required`: A boolean indicating if the field is mandatory.
 - `defaultValue`: The default value for the field (note: this only applies in the JavaScript layer; in the database, the field will be optional).
 - `input`: This determines whether a value can be provided when creating a new record (default: `true`). If there are additional fields, like `role`, that should not be provided by the user during signup, you can set this to `false`.
+- `returned`: Controls whether the field is included in SELECT queries (default: `true`). Set to `false` if you want the field stored in the database but not retrieved in read operations.
```
ping-maxwell: we don't properly implement this as a standard across everything in BA for now, we do intend to in the future but for now we shouldn't have this in the docs as it could mislead people

docs/content/docs/concepts/database.mdx
```diff
@@ -408,11 +410,23 @@ export const auth = betterAuth({
         required: false,
         defaultValue: "en",
       },
+      firstName: {
+        type: "string",
+        fieldName: "first_name", // Database column name
+      },
+      internalNotes: {
+        type: "string",
+        returned: false, // Stored but not retrieved in normal queries
```
ping-maxwell: same here, please remove. We should add this when proper support lands.

docs/content/docs/concepts/database.mdx
```diff
@@ -408,11 +410,23 @@ export const auth = betterAuth({
...
+      },
     },
   },
 });
 ```
 
+<Callout type="info">
+  **Field Retrieval**: The `returned` property controls whether a field appears in query results (default: `true`). Set `returned: false` only when you want to store a field in the database but exclude it from being retrieved in normal operations (e.g., internal metadata or audit fields).
```
ping-maxwell: this too

docs/content/docs/concepts/database.mdx
```diff
@@ -471,11 +485,11 @@ export const auth = betterAuth({
...
 ### ID Generation
 
-Better Auth by default will generate unique IDs for users, sessions, and other entities. If you want to customize how IDs are generated, you can configure this in the `advanced.database.generateId` option in your auth config.
+Better Auth by default will generate unique IDs for users, sessions, and other entities. You can customize ID generation behavior using the `advanced.database.generateId` option.
 
-You can also disable ID generation by setting the `advanced.database.generateId` option to `false`. This will assume your database will generate the ID automatically.
+#### Option 1: Disable ID Generation for All Tables
 
-**Example: Automatic Database IDs**
+Setting `generateId` to `false` disables ID generation globally, letting your database handle all ID generation (useful for auto-increment/serial columns):
```
ping-maxwell: you don't actually have to disable `generateId` in order for auto-increment/serial to work, instead we recommend using `database.useNumberId` for that case (and that would automatically disable BA id generation)

ahmed-abdat:
@ping-maxwell Thanks for the feedback! I've removed all documentation for the unimplemented properties (fieldName, returned) and updated the ID
  generation section to recommend useNumberId for auto-increment scenarios. The docs now only cover fully implemented features.

PR #5174, opened 2025-10-08
fix(two-factor): return parsed array in viewBackupCodes

### LINKED ISSUE #5168: after updating backup codes using generateBackupCodes it gives json parse error

### To Reproduce

enable 2fa on account
generate new backup codes
now when u call const data = await auth.api.viewBackupCodes({
    body: {
      userId: session.user.id,
    },
  });
u will get json parse error always 

it should give backup codes as it was giving before updating backup codes but giving error instead  POST /api/auth/two-factor/generate-backup-codes 200 in 5423ms
 ⨯ SyntaxError: Unexpected number in JSON at position 1 (line 1 column 2)
    at JSON.parse (<anonymous>)
    at async page (src\app\(settings)\settings\security\backup-codes\page.tsx:33:15)
  31 |   }
  32 |
> 33 |   const data = await auth.api.viewBackupCodes({
     |               ^
  34 |     body: {
  35 |       userId: session.user.id,
  36 |     }, {
  digest: '73072180'

1.3.24

### System info

```bash
"system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Home Single Language",
    "release": "10.0.22635",
    "cpuCount": 4,
    "cpuModel": "AMD Ryzen 3 3250U with Radeon Graphics         ",
    "totalMemory": "5.95 GB",
    "freeMemory": "1.43 GB"
  },
  "node": {
    "version": "v24.4.1",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.2.18"
  },
```

Client

```typescript

```

### Additional context

_No response_

ahmed-abdat:
Fixes #5168

### Problem
When calling `auth.api.viewBackupCodes()` after generating new backup codes, users get a JSON parse error:
```
SyntaxError: Unexpected number in JSON at position 1
```

The `viewBackupCodes` endpoint was missing the `decryptBackupCodes()` call, causing it to fail when storage encryption is enabled.

### Solution
Added proper dual-layer decryption to handle all storage configurations:
- `decryptBackupCodes()` removes optional storage encryption layer
- `getBackupCodes()` removes internal encryption and parses JSON

This matches the pattern used in the `verifyBackupCode` endpoint.

### Tests
- [x] All 17 two-factor tests pass
- [x] Handles plain/undefined storage mode
- [x] Handles encrypted storage mode  
- [x] Handles custom encryption functions


ahmed-abdat:
done i have add viewBackupCodes regression tests

PR #5180, opened 2025-10-09
fix: Argument `where` of type TwoFactorWhereUniqueInput needs at least one of `id` arguments

### LINKED ISSUE #2474: Prisma invalid invocation when trying to find/update/delete one entry without a unique property

### To Reproduce

1. Setup better auth with prisma and the `twoFactor()` plugin
2. Generate the schema with `npx @better-auth/cli generate`
3. Migrate the new schema with `prisma migrate dev`
4. Generate the new prisma client with `prisma generate`
5. Enable 2FA on a user via BA
6. Disable 2FA on a user via BA
7. Check logs on the server

When BA tries to delete a twoFactor entity from the `twoFactor` table, it tries to access it via the `userId` property. For some reason prisma does not like it and throws this error:
```
prisma:error 
Invalid `prisma.twoFactor.delete()` invocation:

{
  where: {
    userId: "as34ViBwpxNeAgDdZMj4OnvkuPEWkwtT",
?   id?: String,
?   AND?: TwoFactorWhereInput | TwoFactorWhereInput[],
?   OR?: TwoFactorWhereInput[],
?   NOT?: TwoFactorWhereInput | TwoFactorWhereInput[],
?   secret?: StringFilter | String,
?   backupCodes?: StringFilter | String,
?   user?: UserScalarRelationFilter | UserWhereInput
  }
}

Argument `where` of type TwoFactorWhereUniqueInput needs at least one of `id` arguments. Available options are marked with ?.
```

I am a bit confused, because according to the generated schema a user can have multiple twoFactors, which is probably why prisma is complaining here. Perhaps BA has to use the `deleteMany()` function when calling prisma, instead of just `delete()`.

1.2.7

```bash
- OS: MacOS Sequoia 15.1
- Browser: Arc
```

Backend

```typescript
	// Used as issuer for 2FA
	appName: 'Blah',

	// Session
	session: {
		expiresIn: SESSION_EXPIRES_IN
	},

	// Database
		provider: 'sqlite'
	}),

	// Cookies
	advanced: {
		cookiePrefix: 'recipe'
	},

	// Email/Password
	emailAndPassword: {
		enabled: true,
		sendResetPassword: async ({ user, url }, request) => {
			const lng = getLanguageFromRequestHeaders(request);
			await EmailService.sendEmailRequest('RESET_PASSWORD', lng, {
				to: user.email,
				href: encodeURIComponent(url)
			});
		}
	},
	emailVerification: {
		sendOnSignUp: true,
			const lng = getLanguageFromRequestHeaders(request);

			const newUrl = new URL(url);
			newUrl.searchParams.set('callbackURL', route('/(protected)'));

			await EmailService.sendEmailRequest('VERIFY_EMAIL', lng, {
				to: user.email,
				href: encodeURIComponent(newUrl.toString())
			});
		}
	},

	// User
	user: {
		additionalFields: {
			hasLifetimeAccess: {
				type: "boolean",
				required: false,
				defaultValue: false,
				input: true,
			}
		},
		changeEmail: {
			enabled: true,
			sendChangeEmailVerification: async ({ user, url }, request) => {
				const lng = getLanguageFromRequestHeaders(request);

				const newUrl = new URL(url);
				newUrl.searchParams.set('callbackURL', route('/(protected)'));

				await EmailService.sendEmailRequest('CONFIRM_CHANGE_EMAIL_REQUEST', lng, {
					to: user.email,
					href: encodeURIComponent(newUrl.toString())
				});
			}
		},
		deleteUser: {
			enabled: true,
			sendDeleteAccountVerification: async ({ user, url }, request) => {
				const lng = getLanguageFromRequestHeaders(request);

				const newUrl = new URL(url);
				newUrl.searchParams.set('callbackURL', route('/(protected)'));

				await EmailService.sendEmailRequest('CONFIRM_DELETE_ACCOUNT_REQUEST', lng, {
					to: user.email,
					href: encodeURIComponent(newUrl.toString())
				});
			}
		}
	},

	// Socials
	socialProviders: {
		google: {
			clientId: GOOGLE_CLIENT_ID,
			clientSecret: GOOGLE_CLIENT_SECRET
		},
		github: {
			clientId: GITHUB_CLIENT_ID,
			clientSecret: GITHUB_CLIENT_SECRET
		}
	},

	// Plugins
	plugins: [
		twoFactor(),
		passkey(),
		stripePlugin
	],

	// Redis
	secondaryStorage: {
		get: async (key) => {
			const value = await redis.get(REDIS_PREFIX + key);
			return value ? value : null;
		},
			if (ttl) await redis.set(REDIS_PREFIX + key, value, { EX: ttl });
			else await redis.set(REDIS_PREFIX + key, value);
		},
		delete: async (key) => {
			await redis.del(REDIS_PREFIX + key);
		}
	},

	// Hooks
	hooks: {
			if (ctx.path === '/get-session') {
				const sessionToken = await ctx.getSignedCookie(COOKIES.SESSION, BETTER_AUTH_SECRET);

				if (!sessionToken) {
					return;
				}

				await refreshUserInSessionIfNecessary(sessionToken);
			}
		})
	}
});
```

### Additional context

This error actually happens in quite a few places, where BA tries to access one single thing via a non-unique property. I sometimes saw it popup too when using the Stripe plugin, but don't remember anymore in which scenario.

AlexStrNik:
Fixes #2474
    

---

## Summary by cubic
Switches two-factor backup code updates to use adapter.updateMany, removing the unique ID requirement and preventing the “TwoFactorWhereUniqueInput needs at least one of `id` arguments” error.
This ensures encrypted backup codes are stored without failing when a unique identifier isn’t available.

### DIFF
--- a/packages/better-auth/src/plugins/two-factor/backup-codes/index.ts
+++ b/packages/better-auth/src/plugins/two-factor/backup-codes/index.ts
@@ -445,7 +445,7 @@ export const backupCode2fa = (options?: BackupCodeOptions) => {
 						ctx,
 						backupCodes.encryptedBackupCodes,
 					);
-					await ctx.context.adapter.update({
+					await ctx.context.adapter.updateMany({
 						model: twoFactorTable,
 						update: {
 							backupCodes: storedBackupCodes,


AlexStrNik:
> could you please update test case?

Sorry, at first I thought I broke a test. Did you mean adding a new test that covers backup codes generation with the Prisma adapter? If so, where would it be better to place it `in packages/better-auth/src/adapters/prisma-adapter/test` or in `packages/better-auth/src/plugins/two-factor/two-factor.test.ts`?

AlexStrNik:
The issue mainly affects Prisma users (though maybe not exclusively) because of the generated schema. The two-factor auth plugin expects a one-to-one relation between `User` and `TwoFactor`, but since the `userId` field isn’t marked as `@id` or `@unique` in the generated Prisma schema, Prisma (or the generator) adds another `id` field to `TwoFactor` and treats the `User` to `TwoFactor` relation as one-to-many.

The reason it only breaks on update is that `findOne` from the internal adapter actually maps to `prisma.findFirst` (which might not be technically correct, but understandable), and `prisma.findFirst` doesn’t require unique constraints in `where` clauses, it just returns the first element in the array. However, `update` maps to `prisma.update` which should update only one element, and since `userId` isn’t a unique constraint in the Prisma schema, it fails.

The proper fix would be to change how schemas are generated, but that would be a breaking change, introducing unwanted migrations and potentially affecting other parts of the codebase. The current fix is simple, doesn’t affect other adapters, and can safely be released as a patch or minor update.

PR #5185, opened 2025-10-09
docs: update line social provider example

sacrosanctic:
The comments are inlined and adjusted so they can be used when uncommented.
    

---

## Summary by cubic
Clarified LINE provider docs by updating inline comments to show how to set a custom redirectURI and pass scopes. The example lines are now ready to uncomment and use directly.


sacrosanctic:
I just realized adding `scope` does not overwrite the default values. But I see this is already reviewed so I don't want to change it without asking. Can I add another line to refer to `disableDefaultScope`.

```ts
// scope: ["custom"], // uncomment to append scope
// disableDdefaultScope: true, // uncomment to overwrite default scope which includes `openid`, `profile`, and `email`
```

Bekacru:
@sacrosanctic you should change it you're right

sacrosanctic:
done

sacrosanctic:
just giving this a ping

PR #5192, opened 2025-10-09
Update list organizations with skip and offset query

notaphplover:
## Changed
- Updated `GET /organization/list` with optional `limit` and `offset` query parameters.
- Updated tests accordingly.
- Updated docs accordingly.

### Context
After reading [this comment](https://github.com/better-auth/better-auth/issues/4586#issuecomment-3287762096), I wanted to propose this PR to provide query pagination options the same way `GET /organization/list-members` includes limit and offset query params.

### Motivation
I want to have paginated queries to avoid performance bottlenecks when a large amount of organizations are about to be retrieved. Refer to #4586 for more information.

I'm also considering `better-auth` as one of the possible integrations for the future of the [inversify HTTP tools](https://inversify.io/http/). I do like this library and I would love to give my 5 cents to make it better.
    

---

## Summary by cubic
Adds pagination to GET /organization/list with optional limit and offset to fetch organizations in pages. Mirrors /organization/list-members and helps avoid heavy responses for large org sets (addresses #4586).

- **New Features**
  - Accept limit and offset query params (string or number) on /organization/list.
  - Validate and pass params to adapter; adapter applies limit/offset in queries.
  - Added tests for limit/offset paging and updated docs with param details.

### DIFF
--- a/docs/content/docs/plugins/organization.mdx
+++ b/docs/content/docs/plugins/organization.mdx
@@ -587,6 +587,14 @@ Or alternatively, you can call `organization.list` if you don't want to use a ho
 <APIMethod path="/organization/list" method="GET" requireSession>
   ```ts
   type listOrganizations = {
+    /**
+     * The maximum number of organizations to return.
+     */
+    limit?: number
+    /**
+     * The number of organizations to skip. Default is 0.  
+     */
+    offset?: number = 0
   }
   ```
 </APIMethod>


docs/content/docs/plugins/organization.mdx
```diff
@@ -587,6 +587,14 @@ Or alternatively, you can call `organization.list` if you don't want to use a ho
...
   ```ts
   type listOrganizations = {
+    /**
+     * The maximum number of organizations to return.
+     */
+    limit?: number
+    /**
+     * The number of organizations to skip. Default is 0.  
+     */
+    offset?: number = 0
```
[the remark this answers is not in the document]
notaphplover: Fair enough. I am trying to keep the same style used to document other blocks though

notaphplover:
I see there's a single test failing on CI. I don't see how could it possibly be related with my changes, any insights would be very much appreciated :slightly_smiling_face:

packages/better-auth/src/plugins/organization/routes/crud-org.ts
```diff
@@ -902,6 +920,10 @@ export const listOrganizations = <O extends OrganizationOptions>(options: O) =>
 			const adapter = getOrgAdapter<O>(ctx.context, options);
 			const organizations = await adapter.listOrganizations(
 				ctx.context.session.user.id,
+				{
+					limit: ctx.query?.limit ? Number(ctx.query.limit) : undefined,
+					offset: ctx.query?.offset ? Number(ctx.query.offset) : undefined,
+				},
 			);
 			return ctx.json(organizations);
```
himself65, line 928: I think the response should also include `total` and `offset`

notaphplover: Hey, thanks for your feedback. We could go that way, but then it would be a breaking change in the API, right? I mean, the API returns an array of organizations. Returning `total` and `organizations` would require returning an object if I'm not missing something. My cup of tea would be avoid breaking. As an alternative, maybe we could:

1. Provide `skip` and `offset` query params in this PR.
2. Create an issue to update the endpoint  (I think offset is sort of redundant, `listMembers` endpoint only return `total` as well).
3. When you plan the 2.0 iteration, submit a PR to complete the previous issue.

packages/better-auth/src/plugins/organization/routes/crud-org.ts
```diff
@@ -876,8 +876,25 @@ export const listOrganizations = <O extends OrganizationOptions>(options: O) =>
 		"/organization/list",
 		{
 			method: "GET",
+			query: z
+				.object({
+					limit: z
+						.string()
```
himself65, line 882: use coerce

himself65 (changes requested):

PR #5198, opened 2025-10-09
fix: import `node:async_hooks` directly

himself65:
This will fix nextjs build issue
    

---

## Summary by cubic
Fix Next.js build errors by using a static import specifier for node:async_hooks in the AsyncLocalStorage dynamic import, replacing the variable-based path. Behavior stays the same while allowing bundlers to statically analyze the import.


landoncolburn:
This is causing issues in the client of some plugins

https://github.com/better-auth/better-auth/issues/5539

PR #5202, opened 2025-10-10
fix(two-factor): backup codes shouldn't be encrypted twice

Bekacru:
## Summary by cubic
Fixes double encryption of two-factor backup codes and ensures verification and viewing return parsed arrays instead of encrypted strings. Encrypted storage is now the default to prevent misconfiguration.

- **Bug Fixes**
  - Centralized encryption/decryption in generateBackupCodes/getBackupCodes with safeJSONParse.
  - Defaulted backupCodeOptions to storeBackupCodes: "encrypted" and removed redundant helpers.
  - verifyBackupCode and viewBackupCodes now pass options correctly and return string[].
  - Updated tests to validate array output and use the verifyBackupCode response.

### DIFF
--- a/packages/better-auth/src/plugins/two-factor/backup-codes/index.ts
+++ b/packages/better-auth/src/plugins/two-factor/backup-codes/index.ts
@@ -482,7 +482,7 @@ export const backupCode2fa = (opts: BackupCodeOptions) => {
 					});
 					if (!twoFactor) {
 						throw new APIError("BAD_REQUEST", {
-							message: "Backup codes aren't enabled",
+							message: TWO_FACTOR_ERROR_CODES.BACKUP_CODES_NOT_ENABLED,
 						});
 					}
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/two-factor/backup-codes/index.ts
```diff
@@ -506,22 +491,20 @@ export const backupCode2fa = (options?: BackupCodeOptions) => {
...
-						decryptedBackupCodes,
 						ctx.context.secret,
+						opts,
 					);
-					if (!backupCodes) {
+
+					if (!decryptedBackupCodes) {
 						throw new APIError("BAD_REQUEST", {
-							message: TWO_FACTOR_ERROR_CODES.BACKUP_CODES_NOT_ENABLED,
+							message: "Backup codes are not valid",
```
himself65: use `TWO_FACTOR_ERROR_CODES `

PR #5206, opened 2025-10-10
docs(stripe): add warning about signature verification for webhooks

ouariachi:
This PR adds a warning callout in the Stripe plugin documentation to clarify a common cause of webhook signature verification errors.

The note explains that parsing or modifying the request body (e.g., with express.json()) before passing it to Better Auth breaks Stripe’s signature validation. It instructs developers to use the raw request body when handling webhooks.

image

    

---

## Summary by cubic
Add a warning callout to the Stripe plugin docs about webhook signature verification. It explains that parsing the request body (e.g., express.json()) before Better Auth breaks Stripe’s check and tells you to use the raw body.

Also removes a redundant line in the local development section.

### DIFF
--- a/docs/content/docs/plugins/stripe.mdx
+++ b/docs/content/docs/plugins/stripe.mdx
@@ -857,6 +857,10 @@ If webhooks aren't being processed correctly:
 3. Ensure you've selected all the necessary events in the Stripe dashboard
 4. Check your server logs for any errors during webhook processing
 
+<Callout type="warning">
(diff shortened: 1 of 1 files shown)


Bekacru:
If you parse the body earlier, Better Auth won't work for all endpoints. It's not specific to the stripe plugin

ouariachi:
> If you parse the body earlier, Better Auth won't work for all endpoints. It's not specific to the stripe plugin

That doesn’t seem to be the case for me, all my endpoints work correctly, except stripe

ping-maxwell:
We do mention in our express docs not to use `express.json()`.

image

Other users shouldn't run into this issue unless they failed to properly go through our express docs in the first place 🤔

Issue #5144, opened 2025-10-07
RevokeSession with secondary storage


Girbi:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Enable secondary storage and set auth with the following settings for sessions:
`session: {
    modelName: "sessions",
    storeSessionInDatabase: true,
    preserveSessionInDatabase: true,
    cookieCache: {
      enabled: true,
      maxAge: 5 * 60, // 5 minutes
    },
  },`
2. Calling revokeSession only deletes secondary storage data, does not update expiredAt in the database, because the internalAdapter just finds and deletes any sessions, it doesnt consider the database session.

### Current vs. Expected behavior

Current: revokeSession just deletes secondaryStorage and doesn't consider the database session.

Expected: revokeSession to update session inside database also, not just delete it from secondary storage, because i passed `preserveSessionInDatabase: true` and this means the session cannot be deleted from the database.

### What version of Better Auth are you using?

1.3.27

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#1 SMP PREEMPT_DYNAMIC Thu, 02 Oct 2025 19:26:36 +0000",
    "release": "6.16.10-arch1-1",
    "cpuCount": 12,
    "cpuModel": "AMD Ryzen 5 3600X 6-Core Processor",
    "totalMemory": "31.26 GB",
    "freeMemory": "5.01 GB"
  },
  "node": {
    "version": "v22.14.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "10.15.0"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "1.3.27",
    "config": {
      "appName": "App",
      "secret": "[REDACTED]",
      "advanced": {
        "cookiePrefix": "app"
      },
      "session": {
        "modelName": "sessions",
        "storeSessionInDatabase": true,
        "preserveSessionInDatabase": true,
        "cookieCache": {
          "enabled": true,
          "maxAge": 300
        }
      },
      "trustedOrigins": [
        "http://localhost:5173"
      ],
      "baseURL": "http://localhost:3333",
      "database": {
        "db": {},
        "type": "postgres",
        "casing": "snake"
      },
      "secondaryStorage": {},
      "rateLimit": {
        "enabled": true,
        "window": 10,
        "max": 100,
        "storage": "secondary-storage"
      },
      "emailAndPassword": {
[...]

PR #5213, opened 2025-10-10
fix(session): update expiresAt in database when preserveSessionInDatabase is enabled

notsliver:
Fixed [Issue #5144](https://github.com/better-auth/better-auth/issues/5144)
    

---

## Summary by cubic
Fixes session revocation when preserveSessionInDatabase is enabled: we now update expiresAt to epoch in the database instead of deleting the row, while still clearing secondary storage. This keeps historical session records without leaving active tokens.

- **Bug Fixes**
  - Set expiresAt = new Date(0) for revokeSession and revokeSessions when preserveSessionInDatabase: true; otherwise behavior is unchanged.
  - Clear secondary storage entries and add tests to verify DB expiry updates and cache removal.


Girbi:
Sorry why wasn't this continued? Is it fixed somehow?

Issue #1286, opened 2025-01-27
`emailOtp.verifyEmail` does not revalidate the user session


itsyoboieltr:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

I was told in https://discord.com/channels/1288403910284935179/1333163775880335502/1333163775880335502 to create an issue for this.

When a user is logged in, and verifies their email using `emailOtp.verifyEmail`, the `emailVerified` field is not updated on the client if there is a cookie cache. Once the cookie cache expires, and the page is manually refreshed, only then will the `emailVerified` field return true. This cookie cache should automatically be revalidated (similar to how it is done when signing in, or signing out), so that the `useSession` hook always returns the correct value.

### Current vs. Expected behavior

Current behaviour:

1. user logs in (`emailVerified: false`)
2. `useSession` returns `emailVerified: false`
3. user asks for OTP using `emailOtp.sendVerificationOtp`
4. user enters OTP using `emailOtp.verifyEmail` to verify their email.
5. After `emailOtp.verifyEmail` succeeds, `useSession` still returns `emailVerified: false` (the issue is that this value does not get updated to true)

Expected behaviour:

1. user logs in (`emailVerified: false`)
2. `useSession` returns `emailVerified: false`
3. user asks for OTP using `emailOtp.sendVerificationOtp`
4. user enters OTP using `emailOtp.verifyEmail` to verify their email.
5. After `emailOtp.verifyEmail` succeeds, `useSession` returns `emailVerified: true` (cookie cache gets revalidated)

### What version of Better Auth are you using?

1.1.14

### Provide environment information

```bash
- OS: macOS
- Browser: Chrome
```

### Which area(s) are affected? (Select all that apply)

Backend, Client

dassaswat:
Is the value getting updated in the Database ?

itsyoboieltr:
@dassaswat in the database it is updated, yes.

IronHeartDan:
+1

samir1498:
is there a proper fix for this issue?

peyronoscar:
I had the same issue with the session cookie cache when using email verification on email sign in. This solved it for me:

```
const auth = betterAuth({
    emailVerification: {
        autoSignInAfterVerification: true
    }
})
```

[Docs](https://www.better-auth.com/docs/concepts/email#auto-signin-after-verification)

itsyoboieltr:
[...]

PR #5219, opened 2025-10-10
fix(email-otp): email-verification doesn't trigger session signal

ping-maxwell:
closes: https://github.com/better-auth/better-auth/issues/1286
linear: https://linear.app/better-auth/issue/ENG-41/fix-revalidate-session-after-emailotpverifyemail
    

---

## Summary by cubic
Email OTP verification now triggers the session signal so the client revalidates the session right after /email-otp/verify-email. Aligns with Linear ENG-41 to prevent stale auth state.

- **Bug Fixes**
  - Added atom listener for "/email-otp/verify-email" to emit "$sessionSignal" and refresh the session.

### DIFF
--- a/packages/better-auth/src/plugins/jwt/sign.ts
+++ b/packages/better-auth/src/plugins/jwt/sign.ts
@@ -66,7 +66,7 @@ export async function signJWT(
 				data: JSON.parse(key.privateKey),
 			}).catch(() => {
 				throw new BetterAuthError(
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/email-otp/index.ts
```diff
@@ -725,7 +725,24 @@ export const emailOTP = (options: EmailOTPOptions) => {
 							},
 						});
 					}
-
+					const currentSession = await getSessionFromCtx(ctx);
+					if (currentSession) {
```
Bekacru:
```suggestion
					if (currentSession && updatedUser.emailVerified) {
```

Issue #4357, opened 2025-09-02
MagicLink unnecessary Pending state of authClient.useSession()


ArtGurianov:
### Is this suited for github?

- Yes, this is suited for github

### To Reproduce

1. Implement basic SignIn form with some state for showing results using Magic Link (or any multi-step provider)
2. Wrap all different options of your SignIn client pages with SignedOutGuard at layout level making sure that only unauthorized users can access those pages
3. Try to Sign In with Magic Link

### Current vs. Expected behavior

Current Behavior:
- Sending magic link email unnecessarily turns session state into Pending, leading to Guard triggers in my case. It might cause other unexpected triggers depending on business logic - the most important is that at this step it doesn't have anything to do with session YET.
- Current implementation of BetterAuth plugins treats any endpoint's hit as a potential session mutating, hence switching the state into pending. This is exactly the case for 1-step auth flows, but not for multi-step flows.

Expecting Behavior:
- Expecting to switch session state only at steps where session is involved.

### What version of Better Auth are you using?

latest

### System info

```bash
React + Next.js App Router
```

### Which area(s) are affected? (Select all that apply)

Client

### Auth config (if applicable)

```typescript

```

### Additional context

One possible solution would be to pass an extra options at authClient's plugin config, letting developer chose the exact auth route or number of routes, where session mutation is expected.

mawallace:
This is still happening for me at 1.4.19 — should this be reopened?

PR #5221, opened 2025-10-10
fix(magic-link): dont trigger session refresh on sign-in

ping-maxwell:
the magic-link sign-in endpoint doesn't actually sign you in, it's there to send an email for the end user to sign-in via the URL contained in said email.

we automatically trigger a session refresh when the path starts with `/sign-in`, however in this case since the session doesn't actually refresh it doesn't make sense to trigger all useSession hooks to go into `isPending`.

closes: https://github.com/better-auth/better-auth/issues/4357
linear: https://linear.app/better-auth/issue/ENG-283/fix-magic-link-shouldnt-refresh-usesession-during-sign-in-endpoint
    

---
## Summary by cubic
Stop triggering session refresh on the /sign-in/magic-link endpoint so useSession doesn’t go into isPending when we’re only sending the magic link email (ENG-283). Move session refresh signals to plugin-specific sign-in endpoints for accurate updates.

- **Bug Fixes**
  - Narrowed $sessionSignal matcher in core to exact paths (e.g., /sign-in/email, /sign-up/email), excluding /sign-in/magic-link.
  - Added plugin atomListeners to trigger $sessionSignal on real sign-in flows: /sign-in/anonymous, /sign-in/email-otp, /sign-in/phone-number, /sign-in/username.

<sup>Written for commit fe39f86ec5c4fdcf54d1d46675994306dc86b1bf. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/__snapshots__/init.test.ts.snap
+++ b/packages/better-auth/src/__snapshots__/init.test.ts.snap
@@ -220,6 +220,7 @@ exports[`init > should match config 1`] = `
         },
         "userId": {
           "fieldName": "userId",
+          "index": true,
           "references": {
             "field": "id",
             "model": "user",
@@ -269,6 +270,7 @@ exports[`init > should match config 1`] = `
         },
         "userId": {
           "fieldName": "userId",
+          "index": true,
           "references": {
             "field": "id",
             "model": "user",
@@ -340,6 +342,7 @@ exports[`init > should match config 1`] = `
         },
         "identifier": {
           "fieldName": "identifier",
+          "index": true,
           "required": true,
           "type": "string",
         },
--- a/packages/better-auth/src/adapters/drizzle-adapter/test/adapter.drizzle.mysql.test.ts
+++ b/packages/better-auth/src/adapters/drizzle-adapter/test/adapter.drizzle.mysql.test.ts
@@ -7,7 +7,6 @@ import {
 	authFlowTestSuite,
 	normalTestSuite,
 	numberIdTestSuite,
-	performanceTestSuite,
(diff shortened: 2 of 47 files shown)


packages/better-auth/src/client/config.ts
```diff
@@ -72,14 +72,17 @@ export const getClientConfig = (
 		{
 			signal: "$sessionSignal",
 			matcher(path) {
-				return (
+				const matchesCommonPaths =
 					path === "/sign-out" ||
 					path === "/update-user" ||
 					path.startsWith("/sign-in") ||
```
Bekacru:
```suggestion
					path === "/sign-in/email"
```

packages/better-auth/src/client/config.ts
```diff
@@ -73,16 +73,22 @@ export const getClientConfig = (
...
 					path === "/update-user" ||
-					path.startsWith("/sign-in") ||
-					path.startsWith("/sign-up") ||
+					path === "/sign-up/email" ||
+					path === "/sign-in/email" ||
+					path === "/sign-in/username" ||
+					path === "/sign-in/anonymous" ||
+					path === "/sign-in/phone-number" ||
+					path === "/sign-in/email-otp" ||
+					path === "/sign-in/passkey" ||
```
Bekacru: this should be defined by each plugin instead

Bekacru (changes requested):

packages/better-auth/src/client/config.ts
ping-maxwell: done!

Bekacru (approved):

PR #5224, opened 2025-10-10
feat(oidc-provider): add verify option for client secret hashing

DarthGigi:
### Problem

When using the OIDC provider plugin with non-deterministic hashing algorithms like `Argon2` or `bcrypt` for storing client secrets, authentication fails with `invalid client_secret` errors.

The current client secret verification logic re-hashes the incoming secret and compares it to the stored hash:
```typescript
const hashedClientSecret = await hash(clientSecret);
if (hashedClientSecret !== storedClientSecret) {
  return false;
}
```

This approach only works for deterministic hashing algorithms (e.g., SHA-256) where the same input always produces the same output. However, algorithms like Argon2 and bcrypt generate different hashes each time because they use random salts, making direct comparison impossible.

### Solution

Add an optional `verify` callback to the `storeClientSecret` configuration, following the same pattern used for password hashing in Better Auth:

```typescript
storeClientSecret: {
  hash(clientSecret) {
    return argon2.hash(clientSecret);
  },
  verify(data) {
    return argon2.verify(data.hash, data.clientSecret);
  }
}
```

### Changes

- **`types.ts`**: Added an optional `verify` function to the `storeClientSecret` hash configuration
- **`index.ts`**: Updated `verifyStoredClientSecret` to use the custom verify callback if provided

### Backwards Compatibility

The `verify` option is fully optional, so existing code using deterministic hashing (SHA-256, etc.) continues to work without any changes.

```typescript
// Existing code - still works
storeClientSecret: {
  hash(clientSecret) {
    return sha256(clientSecret)
  }
}

// New code - with verify
storeClientSecret: {
  hash(clientSecret) {
    return argon2.hash(clientSecret);
  },
  verify(data) {
    return argon2.verify(data.hash, data.clientSecret);
  }
}
```

### Related

This follows the established pattern from Better Auth's password verification system, which already supports custom verify functions for the same reason.

    

---

## Summary by cubic
Adds an optional verify callback for OIDC client secret hashing to support non-deterministic algorithms (Argon2, bcrypt) and fix invalid client_secret errors. Existing deterministic hashing continues to work.

- **New Features**
  - storeClientSecret now supports verify({ hash, clientSecret }): Promise<boolean>.
  - Verification uses this callback when provided, enabling Argon2/bcrypt without re-hashing.

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/index.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/index.ts
@@ -210,6 +210,12 @@ export const oidcProvider = (options: OIDCOptions) => {
 			typeof opts.storeClientSecret === "object" &&
 			"hash" in opts.storeClientSecret
 		) {
+			if (typeof opts.storeClientSecret.verify === "function") {
+				return await opts.storeClientSecret.verify({
+					hash: storedClientSecret,
+					clientSecret,
+				});
+			}
 			const hashedClientSecret =
 				await opts.storeClientSecret.hash(clientSecret);
 			return hashedClientSecret === storedClientSecret;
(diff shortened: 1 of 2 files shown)


dvanmali:
Hi all, we released the new [OAuth Provider Plugin](https://www.better-auth.com/docs/plugins/oauth-provider) which provides the `verify` option under `storeClientSecret`. Feel free to let us know how it works :)

Issue #406, opened 2024-11-03
Error “Response body object should not be disturbed or locked” when sending JSON requests with better-auth in NestJS


wh5938316:
**Describe the bug**
I’m encountering an issue when using better-auth with drizzleAdapter in a NestJS application. When I send requests with JSON bodies to any auth-related route, I receive the following error:

```
ERROR [ExceptionsHandler] Request with GET/HEAD method cannot have body.
TypeError: Request with GET/HEAD method cannot have body.
    at new Request (node:internal/deps/undici/undici:6066:17)
    at getRequest (/Users/hao/workspace/bizniff-front/node_modules/.pnpm/better-call@0.2.14-beta.1/node_modules/better-call/src/adapter/request.ts:98:9)
    at <anonymous> (/Users/hao/workspace/bizniff-front/node_modules/.pnpm/better-call@0.2.14-beta.1/node_modules/better-call/src/adapter/node.ts:10:34)
    at toNestJsController (/Users/hao/workspace/bizniff-front/apps/server/dist/auth/better-auth.js:89:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async AuthController.handleAuth (/Users/hao/workspace/bizniff-front/apps/server/dist/auth/auth.controller.js:25:20)
    at async /Users/hao/workspace/bizniff-front/node_modules/.pnpm/@nestjs+core@10.4.1_@nestjs+common@10.4.1_reflect-metadata@0.1.14_rxjs@7.8.1__@nestjs+platfor_fcwotrhkfhfbsqcmjq6msnog44/node_modules/@nestjs/core/router/router-execution-context.js:46:28
    at async /Users/hao/workspace/bizniff-front/node_modules/.pnpm/@nestjs+core@10.4.1_@nestjs+common@10.4.1_reflect-metadata@0.1.14_rxjs@7.8.1__@nestjs+platfor_fcwotrhkfhfbsqcmjq6msnog44/node_modules/@nestjs/core/router/router-proxy.js:9:17
```

It seems that the issue originates from better-call, a dependency of better-auth. When better-auth processes JSON requests, it appears to encounter an issue.

**To Reproduce**

1.	Set up better-auth with drizzleAdapter in a NestJS application.
2.	Send a JSON request body to an auth-related route.

wh5938316:
[...]

PR #5245, opened 2025-10-12
docs: update `nestjs.mdx`

i-shm:
- https://github.com/better-auth/better-auth/issues/406
- https://github.com/ThallesP/nestjs-better-auth

I believe the comments here should be synchronized with nestjs-better-auth README, otherwise users cannot determine whether globally disabling `bodyParser` will affect other modules.
    

---

## Summary by cubic
Updated the NestJS integration docs to clarify that setting bodyParser: false is safe, since Better Auth automatically re-adds the default body parsers and other modules won’t be affected.


himself65:

To update your PR:
```bash
git checkout your-branch
```

PR #5246, opened 2025-10-12
feat: add capjs captcha plugin

chris-windsor:
Add support for [cap.js](https://capjs.js.org/) captcha plugin.

[demo implementation repo](https://github.com/chris-windsor/better-auth-capjs-demo)

My only concern is what would be the preferred way to update typing so that `siteVerifyURLOverride` is required for this provider or make a new config option that is `siteVerifyURL` and ignore `siteVerifyURLOverride`.
    

---

## Summary by cubic
Added Cap (cap.js) as a new captcha provider to the Better Auth captcha plugin with a server-side verify handler and docs updates.

- **New Features**
  - New provider: "cap" with a verify handler that POSTs JSON { secret, response, remoteip }.
  - Wired provider into constants, types, and plugin selection; added tests for success, 403, and 500 flows.
  - Docs updated to list Cap and link to client integration guidance.

- **Migration**
  - For provider "cap", set siteVerifyURLOverride to your Cap siteverify endpoint (no default URL is provided).

### DIFF
--- a/packages/better-auth/src/plugins/captcha/constants.ts
+++ b/packages/better-auth/src/plugins/captcha/constants.ts
@@ -11,6 +11,7 @@ export const Providers = {
 	GOOGLE_RECAPTCHA: "google-recaptcha",
 	HCAPTCHA: "hcaptcha",
 	CAPTCHAFOX: "captchafox",
+	CAP: "cap",
 } as const;
 
 export const siteVerifyMap: Record<Provider, string> = {
@@ -20,4 +21,5 @@ export const siteVerifyMap: Record<Provider, string> = {
 		"https://www.google.com/recaptcha/api/siteverify",
 	[Providers.HCAPTCHA]: "https://api.hcaptcha.com/siteverify",
 	[Providers.CAPTCHAFOX]: "https://api.captchafox.com/siteverify",
+	[Providers.CAP]: "",
 };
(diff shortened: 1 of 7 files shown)


packages/better-auth/src/plugins/captcha/constants.ts
```diff
@@ -20,4 +21,5 @@ export const siteVerifyMap: Record<Provider, string> = {
 		"https://www.google.com/recaptcha/api/siteverify",
 	[Providers.HCAPTCHA]: "https://api.hcaptcha.com/siteverify",
 	[Providers.CAPTCHAFOX]: "https://api.captchafox.com/siteverify",
+	[Providers.CAP]: "",
```
himself65, line 24: why this is empty?

himself65 (changes requested):

packages/better-auth/src/plugins/captcha/constants.ts
chris-windsor: There is no public endpoint for cap.js. It is a self hosted OSS solution. I was unsure what the preferred changes would be since `siteVerifyURLOverride` would need to be required for cap.js but none of the others and since this Map expects a cap.js entry to be defined

The-Best-Codes (commented):
+1 to this PR, I would love to see it merged!

PR #5255, opened 2025-10-12
refactor(email-otp): deprecate forget-password in favor of request-password-reset

ahmed-m-abbass:
---

## Summary by cubic
Replaced the “forget-password” flow with “request-password-reset” in the email OTP plugin. Adds a new endpoint and keeps legacy behavior working while improving verification lookup.

- **Refactors**
  - Added POST `/request-password-reset/email-otp` with `auth.api.requestPasswordResetEmailOTP` and `authClient.requestPasswordReset.emailOtp`.
  - Deprecated `forget-password` endpoint and type; legacy endpoint retained with warnings.
  - Unified OTP type handling via `EmailOTPType`; verification checks both new and legacy identifiers.

- **Migration**
  - Switch OTP type to `"request-password-reset"`.
  - Use `auth.api.requestPasswordResetEmailOTP` and `authClient.requestPasswordReset.emailOtp`.
  - Existing `forget-password` calls still work for now but will be removed in the next major.

### DIFF
--- a/packages/better-auth/src/plugins/email-otp/index.ts
+++ b/packages/better-auth/src/plugins/email-otp/index.ts
@@ -998,7 +998,9 @@ export const emailOTP = (options: EmailOTPOptions) => {
 					await ctx.context.internalAdapter.createVerificationValue(
 						{
 							value: `${storedOTP}:0`,
-							identifier: `forget-password-otp-${email}`,
+							// use the same identifier as request-password-reset
+							// to avoid performing multiple database queries
+							identifier: `request-password-reset-otp-${email}`,
 							expiresAt: getDate(opts.expiresIn, "sec"),
 						},
 						ctx,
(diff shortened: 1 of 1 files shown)


ahmed-m-abbass:
@Bekacru , @himself65 I’ve been looking into how `overrideDefaultEmailVerification?: boolean` and it seems it doesn’t fully inherit all default `emailVerification` options.

We could consider exposing a configuration option in the plugin that mirrors the default options, for example:
```ts
overrideDefaultEmailVerification: true,
emailVerification: {
	sendOnSignIn: true,
	sendOnSignUp: true,
},
```
instead of `sendVerificationOnSignUp?: boolean;` only and when `overrideDefaultEmailVerification: true,` this option ignored or inherited from the `emailVerification` base config. this way if we want to support both link and OTP they still separated and same for `onPasswordReset()`

I was considering adding support for handling both link and OTP verification within a single `sendVerificationEmail()` handler, though I haven’t quite figured out the ideal structure yet.

also we could enhance the sign-in and sign-up endpoint responses to include: `{ emailOTPRedirect = true }` and maybe global config to handle redirection similar to twoFactor plugin.

what do you think about this? does it make sense to continue implementing it, or would you suggest a different approach?

PR #5259, opened 2025-10-12
feat(phone): add `customOTP` option

Manokii:
### Feature
Custom OTP for testing/bypass purposes

### Why
I've tried to implement a custom otp for specific phoneNumbers in `sendOTP()` but the problem was if I create a new row in verification table, the original verification code generated before sendOTP() is called won't get deleted

### Changes
- added a `customOTP` option to `phoneNumber()` plugin
- added test cases for `customOTP`
- updated `phone-number.mdx` with custom otp usage 

### Question:
- should this apply to `requestPasswordResetPhoneNumber` & `forgetPasswordPhoneNumber`? 
    

---

## Summary by cubic
Adds a customOTP option to the phoneNumber plugin to use predefined OTPs for specific numbers and optionally skip sending SMS. Ensures the stored verification code matches the custom code by generating it before creating the verification record.

- **New Features**
  - customOTP.phoneNumbers: record or async function mapping phone numbers to OTP codes.
  - customOTP.skipSendOTP: skips sendOTP when a custom code is used.
  - Added unit tests and updated docs with usage examples.

### DIFF
--- a/docs/content/docs/plugins/phone-number.mdx
+++ b/docs/content/docs/plugins/phone-number.mdx
@@ -261,25 +261,37 @@ export const auth = betterAuth({
     - `skipSendOTP`: A boolean to skip sending the OTP code for phone numbers listed in `customOTP.phoneNumbers`. Default is `false`. 
 
 ```ts
+// Example usage of custom OTPs using a static record
 export const auth = betterAuth({
     plugins: [
         phoneNumber({
-            // ...opts
             sendOTP: ({ phoneNumber, code }, request) => {
                 // Implement sending OTP code via SMS
             },
             customOTP: {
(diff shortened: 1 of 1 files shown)


jspr-vc:
Searched through open issues, it seems like no one has mentioned custom OTPs yet

Also just to mention, I'm basing this on supabase test numbers where they allow to input `<number>=<otp>` separated by commas 

It's located here (at the bottom:
https://supabase.com/dashboard/project/_YOUR_PROJECT_ID_/auth/providers?provider=Phone

Bekacru:
You should use custom verifier instead for testing purposes but we probably should add a guide on how to do that in the docs

PR #5276, opened 2025-10-13
docs: bump to next.js 16

Kinfe123:
changes helps on bring the prebundling (cold-start graph built) of turbopack from 1050ms to 600ms, also the compiling got ~1.5x faster and deprecated some flag since turbopack is an opt-out. and much more improvement on DX specially both on dev and prod

prod build from an avg of 2min 40sec to 53sec ( ~1min )
Screenshot 2025-10-13 at 5 06 08 PM

---

## Summary by cubic
Upgraded the docs app to Next.js 16 (beta) to improve dev build speed. Cold starts drop ~1050ms→~600ms and compiles are ~1.5x faster; plus minor TS JSX and proxy export updates.

- **Performance**
  - Enabled Turbopack filesystem cache for dev in next.config.js.
  - Removed the --turbopack flag from the dev script (use plain "next dev").

- **Dependencies**
  - Bumped next to 16.0.0-beta.0.
  - Pinned react and react-dom to 19.2.0; pinned @types via pnpm overrides.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -103,17 +103,11 @@
     "@tailwindcss/postcss": "^4.1.13",
     "@types/jsrsasign": "^10.5.15",
     "@types/mdx": "^2.0.13",
-    "@types/react": "19.2.2",
-    "@types/react-dom": "19.2.1",
+    "@types/react": "^19.2.2",
+    "@types/react-dom": "^19.2.1",
     "mini-svg-data-uri": "^1.4.4",
     "postcss": "^8.5.6",
     "tailwindcss": "^4.1.13",
     "typescript": "^5.9.2"
-  },
-  "pnpm": {
-    "overrides": {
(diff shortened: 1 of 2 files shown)


docs/package.json
```diff
@@ -103,11 +103,17 @@
...
     "mini-svg-data-uri": "^1.4.4",
     "postcss": "^8.5.6",
     "tailwindcss": "^4.1.13",
     "typescript": "^5.9.2"
+  },
+  "pnpm": {
+    "overrides": {
+      "@types/react": "19.2.2",
+      "@types/react-dom": "19.2.1"
+    }
```
himself65: I think this is not working if you put it in submodule

Kinfe123: this is what is recommended on their codegen and they didnt specify why. probably some internal backward compat or sth

Issue #5277, opened 2025-10-13
hard to debug oauth provider errors during getUserInfo


ping-maxwell:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Making this issue on behalf of a Discord user for the purpose of tracking.
discord message: https://discord.com/channels/1288403910284935179/1288403910284935182/1427311748918284442

---

### User message:

hello everyone!
Today a user reported an error trying to login with GitHub, after a bit of digging we figured it was just a blip from GH itself, however we didn't get any context from bettar-auth as getUserInfo simply returns null in case of error when retrieving a user profile fails: https://github.com/better-auth/better-auth/blob/b896e83631f5b633585b993cf9e531f33dc91402/packages/core/src/social-providers/github.ts#L116, which then renders the default error page with unable_to_get_user_info.  should any error there be returned as well to get logged in https://github.com/better-auth/better-auth/blob/b896e83631f5b633585b993cf9e531f33dc91402/packages/better-auth/src/api/routes/callback.ts#L116 and possibly passed down to the error page?

<img width="607" height="549" alt="Image" src="https://github.com/user-attachments/assets/49ef5c8b-7cf1-44fd-a1fa-bb90ee8b4641" />

### Current vs. Expected behavior

Improve logging.

### What version of Better Auth are you using?

any

### System info

```bash
.
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});
```

ping-maxwell:
Made this issue for tracking purposes.
Linear: https://linear.app/better-auth/issue/ENG-338/chore-improve-debugging-experience-during-oauth-unable-to-get-user

PR #5278, opened 2025-10-13
chore: log getUserInfo errors to console

ping-maxwell:
This PR will log any thrown errors from any getUserInfo returned errors from oauth providers. It also passes the error message to the /api/auth/error as a error description.

issue: https://github.com/better-auth/better-auth/issues/5277

linear: https://linear.app/better-auth/issue/ENG-338/chore-improve-debugging-experience-during-oauth-unable-to-get-user
    

---
## Summary by cubic
Log OAuth getUserInfo errors and forward the message to /api/auth/error for easier debugging. Addresses ENG-338 and improves unable_to_get_user_info handling, including logging when userInfo is missing without an error.

<sup>Written for commit aec18c03e62b24b075aa5995119222a187462827. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/authentication/google.mdx
+++ b/docs/content/docs/authentication/google.mdx
@@ -9,6 +9,22 @@ description: Google provider setup and usage.
         To use Google as a social provider, you need to get your Google credentials. You can get them by creating a new project in the [Google Cloud Console](https://console.cloud.google.com/apis/dashboard).
 
         In the Google Cloud Console > Credentials > Authorized redirect URIs, make sure to set the redirect URL to `http://localhost:3000/api/auth/callback/google` for local development. For production, make sure to set the redirect URL as your application domain, e.g. `https://example.com/api/auth/callback/google`. If you change the base path of the auth routes, you should update the redirect URL accordingly.
+
+        <Callout type="info">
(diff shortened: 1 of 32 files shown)


Elfo404:
Won't this work only if the fetch itself fails? i.e. any error response (let's say a valid 403) won't throw. 
Shouldn't it instead log the whole error returned from the better-fetch call inside (which probably requires some breaking changes)

jacogasp:
Hello,

I post here my question in the discussion section https://github.com/better-auth/better-auth/discussions/5275

I think that this issue/pr answers my question: looking at the original code (before this PR) I see that any error in the `getUserInfo` function is not caught, thus, at present, the only way to handle errors is to put a try/catch inside the function and return null if an error occurs. This is not ideal in my use case, since I would like run specific actions for specific errors (e.g, redirect to a proper page if a 403 forbidden error occurs), but at least I should be able to use it as workaround.

It would be nice to have a general mechanism to handle errors in the `getUserInfo` function.

Bekacru:
is this still draft? @ping-maxwell

ping-maxwell:
yes, more to do still

himself65:

himself65:

PR #5280, opened 2025-10-13
feat(session): add option to mark sessions as inactive instead of deleting

### LINKED ISSUE #4518: Add configuration option to mark sessions as inactive instead of deleting them

Currently, when signing out via the default methods, the session entry in the database gets deleted. This makes it hard to keep a log of the sessions that have existed in the application.

I would like for a configuration option to exist that allowed an `active` boolean column to exist in the session table, with the optional ability to make it a timestamp value instead, such as `invalidatedAt`.

This would look something like this:
```typescript
const auth = betterAuth({
  // database configuration
  session: {
    deleteSessionOnSignOut: {
      enabled: false // by default, true
      timestamp: false // whether or not the field is a boolean or a timestamp
    }
  }
})
```
This would then add the field in the generated table schema, and when signing out, for example via:
```typescript
authClient.signOut();
```
Then the session would be kept in the database, but marked as inactive.

### Additional context

_No response_

Blackmamoth:
### Summary
Adds a configuration option to mark sessions as inactive (`isActive`) or timestamped (`invalidatedAt`) instead of deleting them on sign-out, allowing session history to be preserved.

### Changes
- Added `deleteSessionOnSignOut` option under `session`  
- Updated session schema to include `isActive` or `invalidatedAt` based on config  
- Modified sign-out logic to update the field instead of deleting when enabled  
- Added tests and updated documentation  

### Context
This change addresses the open issue where signing out or revoking a session deletes session records entirely, making it hard to track session history. The new option provides a soft-delete alternative.

Closes #4518

### Breaking Changes
- Non-breaking by default  
- Enabling the feature requires adding `isActive` or `invalidatedAt` to the session table  

### Tests
- Verified schema generation and sign-out behavior  
- Default deletion still works  
- All tests passing

    

---
## Summary by cubic
Adds a session option to soft-delete on sign-out by marking sessions inactive or timestamping invalidation instead of deleting. Preserves session history while keeping inactive sessions out of lookups.

- **New Features**
  - Added session.deleteSessionOnSignOut with enabled (default: true) and timestamp (default: false).
  - When enabled=false, createSession sets isActive=true or invalidatedAt=null; deleteSession/deleteSessions flip isActive=false or set invalidatedAt=now.
  - findSession, listSessions, and findSessions skip inactive/invalidated sessions.
  - Conditionally add isActive or invalidatedAt to the session schema when enabled=false.
  - Updated docs and tests.

- **Migration**
  - If using soft delete (enabled=false), add isActive (boolean) or invalidatedAt (date) to the session table. No changes needed if you keep default deletion.

<sup>Written for commit 25f2176f8c3fb78c36256b37729cd9157e11a9bc. Summary will update automatically on new commits.</sup>


ping-maxwell:
Closing as stale.

PR #5298, opened 2025-10-14
fix: narrow type for publicKey to use Uint8Array for decoding

Kinfe123:
after migrating to tsdown which uses rolldown it is casuing a ts type issue which is good since the previous which is unbuild uses rollup which also uses esbuild for transpilation and does not do strictly check for dts generated and type match. so ig from now we can narrow the types like this one to add new Unit8Array since the decode return is - Uint8Array<ArrayBufferLike> which is not the same with Uint8Array<ArrayBuffer> that webauthn is expecting so there is a typecheck issue and the ci is also failing because of this. this also a runtime agnostic. it can run on cf , deno and other runtimes and the api is already supported


himself65:
Closing as we discussed this, you mistakenly downgraded some deps

https://github.com/MasterKale/SimpleWebAuthn/releases/tag/v13.2.0

PR #5311, opened 2025-10-14
fix(drizzle-adapter): handle all operators in multiple `where` conditions

Kinfe123:
## Summary by cubic
Fixes the Drizzle adapter to correctly handle multiple where conditions with all supported operators in both AND and OR clauses. Adds validation and tests to ensure reliable filtering.

- **Bug Fixes**
  - Map operators to Drizzle predicates: eq, ne, lt, lte, gt, gte, contains, starts_with, ends_with, in, not_in (AND/OR groups).
  - Validate that in/not_in values are arrays; throw BetterAuthError otherwise.
  - Add findMany tests covering mixed operators and connectors.


imornar:
When this fix will be released?

Issue #5243, opened 2025-10-12
OAuth providers fail with `state_mismatch` error when using POST callbacks


echeese:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Configure Better Auth with Apple as a social provider
2. Attempt to sign in with Apple
3. Complete authentication on Apple's page
4. Observe `state_mismatch` error after redirect back to application
5. Check browser DevTools → Network tab → the callback POST request does not include the state cookie

### Current vs. Expected behavior

**Current behavior:**
When using Apple Sign-In (or other OAuth providers that use POST-based callbacks), the authentication fails with a `state_mismatch` error. The state cookie is set with `SameSite=Lax` attribute during the OAuth flow, but when Apple redirects back via a POST request, the browser does not send the state cookie (per [SameSite=Lax specification](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value)). Without the state cookie, the server cannot validate the OAuth state parameter and rejects the request.

**Expected behavior:**
Apple Sign-In (and other POST-based OAuth providers) should complete successfully and authenticate the user.

### What version of Better Auth are you using?

1.3.27

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#1 SMP PREEMPT_DYNAMIC Mon, 06 Oct 2025 18:48:29 +0000",
    "release": "6.17.1-arch1-1",
    "cpuCount": 16,
    "cpuModel": "AMD Ryzen 7 5800X 8-Core Processor",
    "totalMemory": "62.72 GB",
    "freeMemory": "43.36 GB"
  },
  "node": {
    "version": "v22.18.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.3"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.0.0-beta.0"
    },
    {
      "name": "react",
      "version": "19.2.0"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.16.3"
    }
  ],
  "betterAuth": {
    "version": "^1.3.27",
    "config": {
      "database": {
        "_events": {},
        "_eventsCount": 0,
        "options": {
          "max": 10,
          "min": 0,
          "maxUses": null,
          "allowExitOnIdle": false,
          "maxLifetimeSeconds": 0,
          "idleTimeoutMillis": 10000
        },
[...]

PR #5322, opened 2025-10-15
fix(oauth): Certain oauth providers won't set `state` cookie

ping-maxwell:
todo: discuss with Beka about this.

The issue arises during the OAuth flow when using Apple Sign-In or similar providers that rely on `POST` callbacks. The state parameter validation fails because the state cookie is not included in the callback request. A workaround suggests using `SameSite=None` and `secure=true` for cookies, but this may compromise security for other session cookies.

issue: https://github.com/better-auth/better-auth/issues/5243

linear: https://linear.app/better-auth/issue/ENG-307/oauth-providers-fail-with-state-mismatch-error-when-using-post
    

---

## Summary by cubic
Set the OAuth state cookie to SameSite=None and Secure so it’s sent on POST callbacks (e.g., Apple Sign-In), fixing state mismatch errors. This change is limited to the state cookie only.

- **Bug Fixes**
  - Create the state cookie with sameSite: "none" and secure: true to ensure it’s included in POST callback requests.
  - Scoped to the state cookie only; other session cookies are unchanged. Addresses Linear ENG-307.

### DIFF
--- a/docs/content/docs/concepts/database.mdx
+++ b/docs/content/docs/concepts/database.mdx
@@ -471,11 +471,11 @@ export const auth = betterAuth({
 
 ### ID Generation
 
-Better Auth by default will generate unique IDs for users, sessions, and other entities. If you want to customize how IDs are generated, you can configure this in the `advanced.database.generateId` option in your auth config.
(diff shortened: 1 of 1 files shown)


Bekacru:
this woudln't solve the issue

Issue #5326, opened 2025-10-15
Add twoFactorPage option to twoFactorClient plugin


wuzgood98:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

The docs for basic usage of Two Factor client plugin includes `twoFactorPage` option, a page to redirect if a user needs to verify their two factor. 
It does not exist as an option currently so I figured why not add its implementation rather than remove it from the docs.

<img width="3680" height="1540" alt="Image" src="https://github.com/user-attachments/assets/fe44c5b3-b3d0-4992-9d7a-8383378a33bf" />

### Current vs. Expected behavior

Following the previous section, I expected `twoFactorClient` to have `twoFactorPage` as an option, but it did not.

### What version of Better Auth are you using?

1.3.27

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Pro",
    "release": "10.0.26120",
    "cpuCount": 4,
    "cpuModel": "Intel(R) Core(TM) i7-6500U CPU @ 2.50GHz",
    "totalMemory": "19.86 GB",
    "freeMemory": "4.34 GB"
  },
  "node": {
    "version": "v22.18.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.1"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "15.5.4"
    },
    {
      "name": "react",
      "version": "^19.2.0"
    }
  ],
  "databases": [
    {
      "name": "@prisma/client",
      "version": "^6.16.2"
    }
  ],
  "betterAuth": {
    "version": "^1.3.27",
    "config": {
      "appName": "GRANTED",
      "user": {
        "changeEmail": {
          "enabled": true
        },
        "additionalFields": {
          "onboardingStartedAt": {
            "type": "string",
            "input": false
          }
        }
      },
      "emailAndPassword": {
        "enabled": true,
        "requireEmailVerification": true,
        "password": {}
      },
      "emailVerification": {
        "sendOnSignUp": true,
        "autoSignInAfterVerification": true
      },
      "socialProviders": {
        "google": {
          "clientId": "[REDACTED]",
          "clientSecret": "[REDACTED]"
        }
      },
      "session": {
        "fields": {
          "expiresAt": "expires",
          "token": "[REDACTED]"
        }
      },
      "account": {
        "fields": {
          "accountId": "providerAccountId",
[...]

PR #5329, opened 2025-10-15
feat(two-factor): add `twoFactorPage` in config

wuzgood98:
**What changes were made and why:**

- Add the TypeScript options interface for `twoFactorClient` to accept `twoFactorPage`.
- Add the logic to handle redirecting users to the specified page when 2FA verification is required and `twoFactorPage` is provided.
- This ensures users do not encounter typescript errors when they use the example from the docs on the [basic usage](https://www.better-auth.com/docs/basic-usage#client-configuration) of Two Factor.
- Users can either use `twoFactorPage` or `onTwoFactorRedirect` to redirect if a user needs to verify their two factor.

**Relevant context:**
- Currently `twoFactorClient` does not accept `twoFactorPage` as an option.
- The recommended option was to use `onTwoFactorRedirect` which was not used in the docs on its basic usage.

**Breaking changes / depreciations:**
- No breaking changes.

**Screenshots for UI changes:**
- No UI changes, but this snippet works now:
Better-Auth-Two-Factor-client-plugin-basic-usage

**Related issues / discussions:**
- Closes: #5326
    

---
## Summary by cubic
Added a twoFactorPage option to twoFactorClient to redirect users to a 2FA page when verification is required, with onTwoFactorRedirect prioritized and twoFactorPage used as a full-page reload fallback. Docs add a usage example and a warning about the reload.

<sup>Written for commit 4e4b8753156865b1ac8982bd1b1786b8fbee83c3. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/two-factor/client.ts
+++ b/packages/better-auth/src/plugins/two-factor/client.ts
@@ -21,7 +21,7 @@ export const twoFactorClient = (
 				onTwoFactorRedirect?: () => void | Promise<void>;
 		  }
 		| undefined,
-): BetterAuthClientPlugin => {
+) => {
 	return {
 		id: "two-factor",
 		$InferServerPlugin: {} as ReturnType<typeof twoFa>,
@@ -48,15 +48,14 @@ export const twoFactorClient = (
 				hooks: {
 					async onSuccess(context) {
 						if (context.data?.twoFactorRedirect) {
-							if (options?.twoFactorPage) {
-								if (typeof window !== "undefined" && window.location) {
-									window.location.href = options.twoFactorPage;
-									return;
-								}
-							}
-
 							if (options?.onTwoFactorRedirect) {
 								await options.onTwoFactorRedirect();
+								return;
+							}
+
+							// fallback for when `onTwoFactorRedirect` is not used and only `twoFactorPage` is provided
+							if (options?.twoFactorPage && typeof window !== "undefined") {
+								window.location.href = options.twoFactorPage;
 							}
 						}
 					},


packages/better-auth/src/plugins/two-factor/client.ts
```diff
@@ -30,6 +35,13 @@ export const twoFactorClient = (options?: {
 				hooks: {
 					async onSuccess(context) {
 						if (context.data?.twoFactorRedirect) {
+							if (options?.twoFactorPage) {
+								if (typeof window !== "undefined" && window.location) {
+									window.location.href = options.twoFactorPage;
```
hades200082: One issue here is that it causes a full page reload. In frameworks like NextJS for instance, it may be more desirable to use Next's own router.push() instead.

The current method of setting onSuccess at the point of calling signIn on the client supports this.

wuzgood98: Thanks for your suggestion! I understand your concern about full page reload. However, I have decided to keep the implementation as a fallback rather than use Next.js's `router.push()` for one reason:

-  Since Better-Auth is framework-agnostic and supports many environments, hardcoding Next.js's router will create a dependency on a specific framework, which would either require duplicate logic for each framework or force all users to implement framework-specific workarounds. 

**My Implementation:**

- `onTwoFactorRedirect` takes priority (so framework-specific routing works perfectly)
- `twoFactorPage` serves as a simple fallback with a clear `@warning` that it causes a full page reload
This gives users flexibility while maintaining the library's framework-agnostic design

Users who want smooth client-side navigation get it through the callback, while those who need a quick fallback have `twoFactorPage`.

Bekacru (approved):

kvanbere:
I think this should be consistent between the `Server` API right? The server API should return a redirect to the two factor page. Also, you may want to redirect the user to a different page depending if they are using `OTP` or `TOTP`.

Issue #5331, opened 2025-10-15
Error generating backup codes


hades200082:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Setup better-auth in a NextJS application using Prisma with a Postgres (I have it running in docker locally)
2. Configure auth.ts as follows:
  ```ts
  export const auth = betterAuth({
  appName: "wahtever",
  database: prismaAdapter(prisma, { provider: "postgresql" }),
  plugins: [
    username(),
    twoFactor({
      totpOptions: {
        backupCodes: {
          storeBackupCodes: "encrypted"
        }
      },
      backupCodeOptions: {
        storeBackupCodes: "encrypted"
      },
    }),
    admin(),
    emailOTP({
      async sendVerificationOTP({ email, otp, type }) {
        console.log(`Sending OTP ${otp} to ${email} for ${type}`)
      },
    })
  ],
  emailAndPassword: {
    enabled: true,
    sendResetPassword: async ({ user, url, token }, req) => {
      console.log("Reset password email sending", url, token)
    },
  },
})
  ```
3. Create the auth-client as follows:
  ```ts
  const authClient = createAuthClient({
    plugins: [
        usernameClient(),
        twoFactorClient(),
        adminClient(),
        emailOTPClient()]
})
  ```
4. Generate the prisma schema and create a migration
5. Create a form that calls `authClient.twoFactor.generateBackupCodes()`  passing in the password

```ts
const { data, error } = await authClient.twoFactor.generateBackupCodes({
                password, 
            })
```



### Current vs. Expected behavior

In my setup, this results in a 500 error on the client, and the following error logged in the NextJS dev console:

```
# SERVER_ERROR:  Error [PrismaClientValidationError]: 
Invalid `prisma.twoFactor.update()` invocation:

{
  where: {
    userId: "QsVPqNBcs1Ock9S4eCIBhriWGDpGm43f",
?   id?: String,
?   AND?: TwoFactorWhereInput | TwoFactorWhereInput[],
?   OR?: TwoFactorWhereInput[],
?   NOT?: TwoFactorWhereInput | TwoFactorWhereInput[],
?   secret?: StringFilter | String,
?   backupCodes?: StringFilter | String,
?   user?: UserScalarRelationFilter | UserWhereInput
  },
  data: {
[...]

PR #5347, opened 2025-10-16
fix(two-factor): prisma issue

okisdev:
This PR fixes a critical issue where Prisma adapter operations were failing when updating two-factor authentication records.

fix https://github.com/better-auth/better-auth/issues/5331
    

---
## Summary by cubic
Fixes Prisma adapter failures in two-factor flows by updating records using the unique two-factor id instead of userId. This makes backup code generation, viewing, and verification reliable.

- **Bug Fixes**
  - Replace updateMany (userId) with update (id) for backup code updates.
  - Fetch twoFactor by userId to get id; throw TWO_FACTOR_NOT_ENABLED when missing.
  - Add id to TwoFactorTable to support unique updates.
  - Expand tests to cover repeated regeneration, removal of used codes, and parsed viewing.

<sup>Written for commit da42d0a2c4e59f674b47ec96d2aef25e6b6d66f9. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/two-factor/backup-codes/index.ts
+++ b/packages/better-auth/src/plugins/two-factor/backup-codes/index.ts
@@ -457,9 +457,10 @@ export const backupCode2fa = (opts: BackupCodeOptions) => {
 					});
 
 					if (!twoFactor) {
-						throw new APIError("BAD_REQUEST", {
-							message: TWO_FACTOR_ERROR_CODES.TWO_FACTOR_NOT_ENABLED,
-						});
+						throw APIError.from(
+							"BAD_REQUEST",
+							TWO_FACTOR_ERROR_CODES.TWO_FACTOR_NOT_ENABLED,
+						);
 					}
 
 					const backupCodes = await generateBackupCodes(


hades200082:
Would it not be simpler to just make the two-factor table's `id` column contain the user ID instead of needing to have a completely new generated Id and then have to make an additional query to get the custom ID?

Or, make the `userId` column the primary key?

packages/better-auth/src/plugins/two-factor/backup-codes/index.ts
```diff
@@ -419,19 +420,39 @@ export const backupCode2fa = (opts: BackupCodeOptions) => {
...
+						});
+					}
+
 					const backupCodes = await generateBackupCodes(
 						ctx.context.secret,
 						opts,
 					);
+
+					// Use the id to update the record
 					await ctx.context.adapter.updateMany({
```
Bekacru: if we update with id we don't need update many

packages/better-auth/src/plugins/two-factor/backup-codes/index.ts
```diff
@@ -322,15 +322,16 @@ export const backupCode2fa = (opts: BackupCodeOptions) => {
 						data: JSON.stringify(validate.updated),
 					});
 
+					// Use the id to update the record
 					await ctx.context.adapter.updateMany({
```
Bekacru: if we update with `id` we don't need update many

packages/better-auth/src/plugins/two-factor/types.ts
```diff
@@ -48,6 +48,7 @@ export interface TwoFactorProvider {
 }
 
 export interface TwoFactorTable {
+	id: string; // Primary key of the two factor table
```
Bekacru: lets remove the comment

Bekacru (changes requested):

okisdev:
@Bekacru comments are fixed. thx

packages/better-auth/src/plugins/two-factor/backup-codes/index.ts
```diff
@@ -322,15 +322,16 @@ export const backupCode2fa = (opts: BackupCodeOptions) => {
 						data: JSON.stringify(validate.updated),
 					});
 
-					await ctx.context.adapter.updateMany({
+					// Use the id to update the record
```
Bekacru:
```suggestion
```

PR #5349, opened 2025-10-16
feat(sso): add option to customize db schema

XiNiHa:
This PR adds the `schema` option to the SSO plugin, used for customizing the database schema used by the plugin.
The documentation is also updated.
    

---

## Summary by cubic
Adds a modelName option to the SSO plugin so you can rename the ssoProvider database model. This helps avoid naming conflicts and match existing schemas.

- New Features
  - SSOOptions.modelName lets you set the model name (default: "ssoProvider").
  - Docs updated with the new option and defaults.

### DIFF
--- a/docs/content/docs/plugins/sso.mdx
+++ b/docs/content/docs/plugins/sso.mdx
@@ -746,6 +746,8 @@ For a detailed guide on setting up SAML SSO with examples for Okta and testing w
 
 **trustEmailVerified**: Trust the email verified flag from the provider.
 
+**schema**: Custom database schema options for the SSO plugin.
+
 <TypeTable
   type={{
     provisionUser: {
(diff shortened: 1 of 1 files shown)


packages/sso/src/index.ts
```diff
@@ -153,6 +153,12 @@ export interface SSOProvider {
 }
 
 export interface SSOOptions {
+	/**
+	 * The model name for the SSO provider.
+	 *
+	 * @default "ssoProvider"
+	 */
+	modelName?: string;
```
Bekacru: we should instead have `schema` option similar to other plugins to customize both model name and other fields

XiNiHa: That looks neat!

XiNiHa:
Looks like it's resolved by #6047

PR #5350, opened 2025-10-16
docs: add note about redis seconds

reslear:
Added a clarification note in the Secondary Storage documentation about TTL (time-to-live) units.
Better Auth uses seconds for TTL values, while some external storages (like Redis or memory-based stores) may expect milliseconds.
    

---

## Summary by cubic
Clarified TTL units in the Secondary Storage docs: Better Auth uses seconds for set() TTL, while some stores expect milliseconds. Added a note and updated the Redis example comment to show how to convert when needed (ttl * 1000).

### DIFF
--- a/docs/content/docs/concepts/database.mdx
+++ b/docs/content/docs/concepts/database.mdx
@@ -64,8 +64,10 @@ betterAuth({
 });
 ```
 
-Note: Better Auth uses **seconds** for the TTL value in `set()`.
+<Callout type="warn">
+Better Auth uses **seconds** for the TTL value in `set()`.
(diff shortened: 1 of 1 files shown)


docs/content/docs/concepts/database.mdx
```diff
@@ -64,6 +64,9 @@ betterAuth({
 });
 ```
 
+Note: Better Auth uses **seconds** for the TTL value in `set()`.
```
Bekacru: can you change this to use `Callout` component instead?

reslear: done

PR #5357, opened 2025-10-16
feat: auto-index CLI

ping-maxwell:
## Feature
Adds `index` support for field attributes, which allows the BA CLI to generate indexes in the schema.

Also fixes:
* kysely `mysql` doesn't apply `varchar` when field is `sortable: true`.
* Drizzle `mysql` doesn't apply `varchar` when field is `sortable: true`.
* Removed adapter performance tests since they don't test anything.
    

---
## Summary by cubic
Adds auto-indexing to the CLI and schemas so common lookup fields automatically generate indexes in Drizzle, Kysely, and Prisma. This improves query performance with no breaking changes.

- **New Features**
  - Drizzle generator emits index() with proper names and imports.
  - Kysely migrations create indexes for indexed fields on table create/alter.
  - Prisma generator adds @@index for non-unique indexed fields (one per field).
  - String fields marked sortable or index now use varchar so indexes apply correctly (Drizzle, Kysely MySQL).
  - Core and plugin schemas mark key fields as indexed (e.g., session.userId, account.userId, verification.identifier, passkey.userId/credentialID, twoFactor.secret/userId, OIDC, SIWE.userId, organization relations, API key).
  - Added index?: boolean to DBFieldAttributeConfig; CLI now uses DBAdapter and depends on @better-auth/core.

- **Migration**
  - Regenerate your schema with the CLI and run migrations to add the new indexes.

<sup>Written for commit 5b5f353996c2cff7f048e552956457b0e6a72b41. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/client/query.ts
+++ b/packages/better-auth/src/client/query.ts
@@ -7,7 +7,7 @@ import { atom, onMount, type PreinitializedWritableAtom } from "nanostores";
 import type { SessionQueryParams } from "./types";
 
 // SSR detection
-const isServer = typeof window === "undefined";
+const isServer = () => typeof window === "undefined";
 
 export const useAuthQuery = <T>(
 	initializedAtom:
@@ -113,7 +113,7 @@ export const useAuthQuery = <T>(
 
 	for (const initAtom of initializedAtom) {
 		initAtom.subscribe(() => {
-			if (isServer) {
+			if (isServer()) {
 				// On server, don't trigger fetch
 				return;
 			}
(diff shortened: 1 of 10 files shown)


packages/cli/test/__snapshots__/schema-mongodb.prisma
```diff
@@ -80,5 +83,7 @@ model TwoFactor {
   userId      String
   user        User   @relation(fields: [userId], references: [id], onDelete: Cascade)
 
+  @@index([secret])
```
Bekacru, line 86: I think we want this to be in one index `@index([userId, secret])`

Bekacru (approved):

PR #5362, opened 2025-10-16
feat(open-api): add the option for a custom base url

rodsnts:
**Add baseUrl option to OpenAPI plugin**

  **What & Why**

  Added a baseUrl config option to the OpenAPI plugin. This fixes an issue where the OpenAPI docs would point to the wrong URL
  when your auth backend runs on a different domain/port than your frontend.

  Before this change, if your backend was on localhost:3001 but your frontend on localhost:3000, the "Try it out" feature would
  send requests to the frontend URL instead of the backend.

  **Changes**

-   Added baseUrl option to OpenAPIOptions
-   Updated the generator to use the custom base URL when provided
-   Falls back to existing behavior (BETTER_AUTH_URL or configured base URL) if not set
-   Added tests and doc
    

---

## Summary by cubic
Adds a baseUrl option to the OpenAPI plugin so the generated schema and “Try it out” requests point to the correct auth backend, even when it runs on a different domain or port.

- **New Features**
  - Added baseUrl to OpenAPIOptions; used as the OpenAPI server URL.
  - Updated generator and plugin to honor baseUrl, with fallback to BETTER_AUTH_URL or ctx.baseURL.
  - Added tests for custom baseUrl and fallback behavior.
  - Updated docs with a simple usage example.

### DIFF
--- a/docs/content/docs/plugins/open-api.mdx
+++ b/docs/content/docs/plugins/open-api.mdx
@@ -74,7 +74,26 @@ app.get("/docs", Scalar({
 }));
 ```
 
+### Using a different base URL for OpenAPI reference
+
+If your authentication backend is hosted on a different domain or port than your frontend, you can specify a different `baseUrl` for the OpenAPI reference. This ensures that the API requests made from the OpenAPI Scalar UI are directed to the correct server:
+
+```ts
+import { betterAuth } from "better-auth"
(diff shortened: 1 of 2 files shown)


Bekacru:
but `ctx.baseURL` is always your backend url regardless of your frontend

PR #5363, opened 2025-10-16
test: remove `runWithEndpointContext`

himself65:
## Summary by cubic
Removed runWithEndpointContext from tests and set up AsyncLocalStorage-based context in test utils. This simplifies tests and aligns them with the updated core context handling.

- **Refactors**
  - Replaced runWithEndpointContext with direct internalAdapter/API calls across session, internal adapter, generic OAuth, social, and Stripe tests.
  - Test utils now enter the current auth context via getCurrentAuthContextAsyncLocalStorage when creating a test instance.
  - Kept existing assertions (headers, TTLs, session/account ops) working without the wrapper.


himself65:
hard to fix. Will do this later

PR #5385, opened 2025-10-17
Feat: add pattern option to validate password

hieudien14310:
Currently, I see that password are only being validated based on min and max length. That is not really comprehensive. So I added a feature to validate password using a pattern to make it stronger and more customizable.

I put  the pattern configuration inside `emailAndPassword`. Once this pattern is set, it will affect all apis related to the password.
If you guys have any feedback, please comment. So I will fix it accordingly.
    

---

## Summary by cubic
Add an optional regex pattern to enforce password strength across all email/password flows. This keeps passwords consistent and stronger in sign-up, reset, change, and admin actions.

- New Features
  - emailAndPassword.pattern (RegExp, default: null) to validate passwords.
  - Enforced in sign-up, changePassword, setPassword, resetPassword, and admin setUserPassword; returns 400 BAD_REQUEST with code PASSWORD_NOT_VALID_PATTERN.
  - Added error code, logging, types, docs, and tests.


R5dan:
@hieudien14310 Personally, I would prefer to have zod validation (or standard schema, but I don't think better auth supports that currently) to validate the password. You would still be able to have custom validation (including with regex, as it has a `z.string().regex()`) This would then allow you to merge all the validators into one as well

hieudien14310:
> @hieudien14310 Personally, I would prefer to have zod validation (or standard schema, but I don't think better auth supports that currently) to validate the password. You would still be able to have custom validation (including with regex, as it has a `z.string().regex()`) This would then allow you to merge all the validators into one as well

Thanks for suggesting using `z.string().regex()`. I agree that using zod for validation will help us avoid the `lastIndex` bug that the AI detected.

However, using of a standard schema for password validation, i think that isnot necessary.  Sometimes, people prefer to set simple password or have standards based on their project. 

So, I think it make more sense to let them configure the pattern themsalves

hieudien14310:
Hi @Bekacru , @himself65 , @ping-maxwell. Could you guys give me your thoughts on this feature

R5dan:
> However, using of a standard schema for password validation, i think that isnot necessary. Sometimes, people prefer to set simple password or have standards based on their project.
> 
> So, I think it make more sense to let them configure the pattern themsalves

Sorry, I think you misunderstood me. I meant having the user input a zod schema which is then used for validation (replace all the `minPasswordLength` `maxPasswordLength` `pattern` with just a `schema` option)

R5dan:
> However, using of a standard schema for password validation, i think that isnot necessary. Sometimes, people prefer to set simple password or have standards based on their project.

I have just realized what you meant by this. I meant [the common interface for TypeScript validation libraries](https://standardschema.dev/) which is [supported by lots of validation libaries](https://standardschema.dev#what-schema-libraries-implement-the-spec). But like I said, better auth currently doesn't support it (I might make a pr for it soon)

ping-maxwell:
> Hi @Bekacru , @himself65 , @ping-maxwell. Could you guys give me your thoughts on this feature

I'd normally recommend hooks for password validation, but there's a case for adding this to core auth config. It lets other plugins (like the admin plugin's set user password endpoint) consistently validate against the same rules, rather than devs having to duplicate validation logic in hooks to check against multiple endpoints with potentially different body parameters.

Also makes writing custom password validation easier for our end users.

I think this is good.

packages/core/src/error/codes.ts
```diff
@@ -18,6 +18,7 @@ export const BASE_ERROR_CODES = defineErrorCodes({
 	EMAIL_NOT_VERIFIED: "Email not verified",
 	PASSWORD_TOO_SHORT: "Password too short",
 	PASSWORD_TOO_LONG: "Password too long",
+	PASSWORD_NOT_VALID_PATTERN: "Password not valid pattern",
```
Bekacru:
```suggestion
	INVALID_PASWORD: "Invalid Password",
```

himself65:

Issue #4744, opened 2025-09-18
[BUG] better-auth/react with @better-auth/expo/client useSession infinite refetch on multiple Set-Cookie headers


ArvidAnderson:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Setup better auth with Expo and a Hono backend.
2. Make your backend return its own Set-Cookie header.
3. Use the useSession hook from the auth client
4. View network requests and see infinite refetching of useSession in dev and in prod until rate limit is hit

### Current vs. Expected behavior

Following the steps in the previous section i expect the better-auth/react useSession with the @better-auth/expo/client plugin hook to properly handle multiple Set-Cookie headers according to [RFC 6265 (Section 4.1, HTTP State Management Mechanism)](https://datatracker.ietf.org/doc/html/rfc6265)

I also noted that it seems like if useSession response includes any other set cookie header than the better auth one, it will get into this refetching loop.

### What version of Better Auth are you using?

1.3.11

### System info

```bash
{
    "system": {
      "platform": "darwin",
      "arch": "arm64",
      "version": "Darwin Kernel Version 25.0.0: Mon Aug 25 21:17:56 PDT 2025; root:xnu-12377.1.9~3/RELEASE_ARM64_T6030",
      "release": "25.0.0",
      "cpuCount": 11,
      "cpuModel": "Apple M3 Pro",
      "totalMemory": "36.00 GB",
      "freeMemory": "1.93 GB"
    },
    "node": {
      "version": "v22.13.1",
      "env": "development"
    },
    "packageManager": {
      "name": "bun",
      "version": "1.2.21"
    },
    "frameworks": null,
    "databases": null,
    "betterAuth": {
      "version": "^1.3.11",
      "config": {
        "appName": "redacted",
        "basePath": "/api/auth",
        "secret": "",
        "trustedOrigins": [
          "redacted://"
        ],
        "emailAndPassword": {
          "enabled": false
        },
        "plugins": [
          {
            "name": "phone-number",
            "config": {
              "id": "phone-number",
              "endpoints": {},
              "schema": {
                "user": {
                  "fields": {
                    "phoneNumber": {
                      "type": "string",
                      "required": false,
                      "unique": true,
                      "sortable": true,
                      "returned": true
                    },
                    "phoneNumberVerified": {
                      "type": "boolean",
                      "required": false,
                      "returned": true,
                      "input": false
                    }
                  }
                }
              },
              "rateLimit": [
                {
[...]

PR #5393, opened 2025-10-17
fix(expo): set-header retrigger `$sessionSignal`

himself65:
Fixes: https://github.com/better-auth/better-auth/issues/4744
    

---

## Summary by cubic
Prevented infinite session refetch loops in the Expo client by only retriggering $sessionSignal when the Set-Cookie header contains better-auth cookies. Fixes better-auth/better-auth#4744.

- **Bug Fixes**
  - Added hasBetterAuthCookies(setCookieHeader, cookiePrefix) to detect better-auth cookies, including __Secure- prefixed names.
  - Only update stored cookies and notify $sessionSignal when detection passes; respects custom storagePrefix.
  - Added tests for better-auth, secure, non-auth (e.g., __cf_bm), mixed, and custom prefix scenarios.

### DIFF
--- a/packages/expo/src/client.ts
+++ b/packages/expo/src/client.ts
@@ -151,7 +151,10 @@ function getOrigin(scheme: string) {
  * @param newCookie - New cookie JSON string
  * @returns true if session cookies have changed, false otherwise
  */
-function hasSessionCookieChanged(prevCookie: string | null, newCookie: string): boolean {
+function hasSessionCookieChanged(
+	prevCookie: string | null,
+	newCookie: string,
+): boolean {
 	if (!prevCookie) return true;
 
 	try {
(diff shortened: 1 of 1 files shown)


himself65:
Double check with @zouhangwithsweet. Seems like the current fix doesn't work? huh?

himself65:
Double checked. This PR is working. will release a release soon

himself65:
Thanks @zouhangwithsweet giving me code to reproduce

spread-admin:
Hi @himself65 - This PR seems to break using the passkey plugin with the expo plugin.
The passkey plugin relies on `webAuthnChallengeCookie` to pass and then be verified here  https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/passkey/index.ts#L530 however, with this change the cookie is ignored and not set.
Would appreciate it if you could reconsider how this is implemented to allow other cookies to be set, even if they dont need to trigger $sessionSignal

PR #5403, opened 2025-10-18
fix(org): removed noisy `logger.info` triggered by `useActiveOrganization`

Konixy:
In [`crud-org.ts`](https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/plugins/organization/routes/crud-org.ts#L711), there is a `logger.info` call that pollutes the server console when the user is not using an active organization and `useActiveOrganization` is called, which is very annoying.
    

---

## Summary by cubic
Removed a noisy server log in the organization route that fired when useActiveOrganization ran without an active org, stopping console spam. Also added an Expo demo and refreshed docs and CI to improve developer experience.

- **Bug Fixes**
  - Removed the logger.info call in the organization CRUD route to prevent unnecessary console output.
  - Cleaned up Tailwind v4 class usage in the Next.js demo.

- **New Features**
  - Added an Expo demo using @better-auth/expo with API Router (sign-in, sign-up, dashboard).
  - Introduced docs AI chat modal and API routes for chat, analytics, feedback, and support, plus an anchor-scroll fix.
  - Updated Next.js demo with Tailwind v4 styles and SSO plugin support.
  - CI/tooling updates: pinned actions/setup-node, updated cache action, removed redundant branch protection workflows, added a noDebugger lint rule, and expanded docker-compose with additional Postgres/MySQL services.

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-org.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-org.ts
@@ -694,7 +694,6 @@ export const getFullOrganization = <O extends OrganizationOptions>(
 				session.session.activeOrganizationId;
 			// return null if no organization is found to avoid erroring since this is a usual scenario
 			if (!organizationId) {
-				ctx.context.logger.info("No active organization found, returning null");
 				return ctx.json(null, {
 					status: 200,
 				});


Konixy:
oups

PR #5406, opened 2025-10-18
docs: clarify relationship between adminRoles and custom access control

### LINKED ISSUE #5380: Docs: Clarify relationship between adminRoles and custom access control
## Issue Description

The admin plugin documentation needs clarification on when `adminRoles` is required vs when custom access control is sufficient.

## Current Documentation Gap

The docs contain what appears to be conflicting guidance:

1. **In the Options section**, there's a warning:
   > "Any role that isn't in the `adminRoles` list, even if they have the permission, will not be considered an admin."

2. **In the Access Control section**, it shows how to define custom roles with specific permissions using `ac` and `roles`.

3. **The docs don't explicitly state** whether `adminRoles` is still required when using custom access control.

## Real-World Scenario

I want my `manager` role to have impersonation permissions. Based on the docs, I can:

```ts
// permissions.ts
export const manager = ac.newRole({
  ...adminAc.statements, // Gives all admin permissions including impersonate
});

// auth.ts
admin({
  ac,
  roles: { manager, admin, user }
})
```

**Question:** Do I still need `adminRoles: ['admin', 'manager']` here, or does the custom access control replace it?

## Suggested Solution

Add a subsection in the Access Control documentation that explicitly states:

### "Using Custom Access Control Without adminRoles"

When you define custom roles with access control:
- The `adminRoles` option [is/is not] required
- The custom role definitions in `roles` [do/do not] replace the default role checking
- [Provide clear example of the recommended approach]

OR

### "Combining adminRoles with Custom Access Control"

When using both options together:
- `adminRoles` determines which roles can access admin endpoints
- Custom access control defines granular permissions within those admin roles
- [Provide clear example]

## Impact

Without this clarity, developers may:
- Add redundant configuration
- Accidentally block intended admin access
- Not understand which approach to use for their use case

GautamBytes:
Fixes #5380 

## Summary
Clarifies the documentation about when `adminRoles` is required vs when custom access control is sufficient.

## Changes
- Updated "Admin Roles" section heading to specify it applies when **not** using custom access control
- Added Note explaining `adminRoles` is not required when using custom AC (via `ac` and `roles`)
- Updated warning callout to clarify it only applies without custom AC
- Added new subsection "Using Custom Access Control Without adminRoles" with working example

## Context
The previous documentation had ambiguous guidance that caused confusion (reported in #5380). User testing confirmed that `adminRoles` is NOT required when using custom access control - the custom role definitions determine permissions directly.

## Type of Change
- [x] Documentation update
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Testing
- [x] Documentation builds successfully locally
- [x] All sections render correctly
- [x] Code examples are syntactically correct
- [x] Navigation links work properly
    

---

## Summary by cubic
Clarified admin plugin docs: adminRoles are only needed when not using custom access control; custom roles defined via ac and roles control permissions.

Updated the Admin Roles section, added a scoped note and warning, and included a working example of custom AC without adminRoles.


docs/content/docs/plugins/admin.mdx
```diff
@@ -754,17 +783,21 @@ admin({
...
-The roles that are considered admin roles. Defaults to `["admin"]`.
+The roles that are considered admin roles when **not** using custom access control. Defaults to `["admin"]`.
 
 ```ts title="auth.ts"
 admin({
   adminRoles: ["admin", "superadmin"],
 });
 ```
 
+**Note:** The `adminRoles` option is **not required** when using custom access control (via `ac` and `roles`). When you define custom roles with specific permissions, those roles will have exactly the permissions you grant them through the access control system.
```
Bekacru: Let's have this in `Callback` component

docs/content/docs/plugins/admin.mdx
```diff
@@ -754,17 +783,21 @@ admin({
...
+This option only applies when you are **not** using custom access control. In that case, any role that isn't in the `adminRoles` list will not be able to perform admin operations.
+
 <Callout type="warning">
-  Any role that isn't in the `adminRoles` list, even if they have the permission,
-  will not be considered an admin.
+  When **not** using custom access control, any role that isn't in the `adminRoles` list
```
Bekacru: Merge this callout with the above `note`

docs/content/docs/plugins/admin.mdx
```diff
@@ -582,6 +582,35 @@ The plugin provides an easy way to define your own set of permissions for each r
...
+        }
+        // adminRoles is NOT needed here
+      }),
+    ],
+  });
+```
+
+In this configuration, if your `manager` role includes admin permissions (like `impersonate`, `ban`, etc.), users with the `manager` role can perform those operations without being in an `adminRoles` list.
+
+
```
Bekacru: We don’t need this information to take up this much space. The callouts below should be enough, so this section can be removed

GautamBytes: @Bekacru did the changes . can you review it again?

PR #5416, opened 2025-10-19
feat(bench): add benchmarks for cold-start and requests/second

### LINKED ISSUE #5415: Add benchmarks to track performance

Better Auth is a feature-rich framework that works out-of-the-box in many environments. Better Auth does not yet have any way to benchmark performance such as cold start or requests/second. It's important to have benchmarks to help developers understand the impact of their latest commits, help integrators determine which endpoints are faster or slower than others, and help the community compare the performance of open source projects.

A good first step is to measure cold-start performance and requests/second (rps). Because each endpoint is different, hooks can be used to perform any setup or teardown (e.g. signing up before benching the /sign-in/email endpoint). The solution shouldn't measure unnecessary overhead like `betterAuth()`, `getTestInstance()`, or network overhead in the case of rps, but should measure any deferred loading before the first request returns a response in the case of cold starts (like `await $context`). In the future an ideal solution would measure concurrent rps using tasks, but for now sequential rps (awaiting the current response before sending the next request) works fine.

There aren't many good options. You could measure the time it takes for tests to complete. But users shouldn't have to measure in edge environments where they're charged per request.

### Additional context

I have a working implementation I'll submit in a pull request if this is something worth pursuing.

igoforth:
This pull request adds to the better-auth package, a new file `rps.bench.ts` in the `src/` directory, a new `bench` npm script, and tinybench, a new dev dependency. `rps.bench.ts` adds two benchmarks, a cold-start benchmark and a requests/second (rps) benchmark.

- Added tinybench because the experimental vitest benchmarking tool (based on tinybench) doesn't support the hooks beforeAll, beforeEach, afterEach, and afterAll. These hooks were also important to make the benchmark tool easily extensible to additional endpoints. I still use the vitest bench function to initialize the tests for debugging and so the `vitest bench` command works.
- Added a debug flag for the rps benchmark to make sure I was receiving valid responses from the endpoints.
- Added types to assist in adding new endpoints and hooks.

No breaking changes or deprecations.

Sample result (AMD Ryzen 7 5800X):
image

Fixes #5415 
    

---

## Summary by cubic
Adds a benchmark harness for better-auth to measure cold-start time and requests/second. Includes a simple script to run the suite locally. No breaking changes.

- **New Features**
  - rps.bench.ts adds cold-start and RPS benchmarks using tinybench.
  - Extensible endpoint config with setup/before/after/teardown hooks and types; includes a ready-to-run auth suite (sign up/in, sessions, user ops).
  - Debug mode to validate responses and print results locally.

- **Dependencies**
  - Adds tinybench ^5.0.1.
  - New npm script: bench → vitest bench.


himself65 (commented):
LGTM.

nit: could we add this to CI?

igoforth:
> LGTM.
> 
> nit: could we add this to CI?

Let me see what I can do

igoforth:
I had some fun refactoring. An inheritance pattern made more sense for the benchmark classes, now `EndpointConfig` is more explicit. Added a light `isCI` to the core package, which was also in std-env like the rest of the env impl. 

New CI job calls bench on all implementing subpackages in `packages/`. If in a CI PR, a markdown table is built in `packages/better-auth/logs/benchmark.md`.  `logs` is already in `.gitignore` so it made sense. The job reads all `benchmark.md` from `packages/*/logs/` and posts a comment on the PR in a format identical to `pkg-pr-new`. I thought about doing some fancy regression detection by comparing PR benchmarks with upstream, but this should be a good start.

@himself65 New CI script isn't passing at the moment probably due to repository-level permissions? Might need some tweaking on your end. I might need to revert the commit that added permissions to the bench job if that's the case.

Can be tested with `act pull_request -j bench -s GITHUB_TOKEN=<a_valid_pat>`. It won't do anything in an act environment due to the conditional.
Screenshot from 2025-10-24 21-47-00

PR #5422, opened 2025-10-20
feat: validate reset password token function

jer-tan:
Hello! I noticed that when users land on a password reset page, they don't know if their token is valid or expired until they actually try to submit the form. This can be frustrating for user, imagine filling out new password only to find out the link expired.

### What's new

This PR adds a new `validateResetPasswordToken` method that lets you check if a reset token is still valid without consuming it. 

**Client usage:**
```ts
const validation = await client.validateResetPasswordToken({ token });

if (validation.data?.valid) {
  // Show the reset form
} else {
  // Show error: token expired or invalid
}
```

### Changes made

- **Server**: Added `/validate-reset-password-token` POST endpoint that checks token validity and expiration
- **Client**: The endpoint is automatically available as `client.validateResetPasswordToken()`
- **Tests**: Added 3 test cases covering valid, invalid, and expired tokens
- **Docs**: Updated the email-password authentication docs with usage examples including a React component

The validation returns:
```ts
{
  valid: boolean,
  message: string  // "Token is valid" | "Invalid token" | "Token has expired"
}
```

### Testing

All tests pass including the 3 new validation tests:
- Valid token returns `{ valid: true }`
- Invalid token returns `{ valid: false }`  
- Expired token returns `{ valid: false }`
    

---
## Summary by cubic
Add a password reset token validator so users can see if their link is valid or expired before filling the form. Docs and OpenAPI updated.

- **New Features**
  - Server: POST /validate-reset-password-token checks validity and expiry without consuming the token.
  - Client: `validateResetPasswordToken({ token })` returns `{ valid, message }` to gate the reset form.
  - Tests & docs: 3 tests (valid/invalid/expired) and updated docs with usage examples (incl. a React snippet).
  - OpenAPI: Endpoint added to the generated schema.

<sup>Written for commit cd774b9e9e22a73c40169f1a27e52de45d9b35a7. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/validate-reset-password-token.md
+++ b/.changeset/validate-reset-password-token.md
@@ -1,0 +1,5 @@
+---
+"better-auth": minor
+---
+
+Add `validateResetPasswordToken` endpoint to check if a password reset token is valid without consuming it
--- a/packages/better-auth/src/api/routes/password.ts
+++ b/packages/better-auth/src/api/routes/password.ts
@@ -226,6 +226,75 @@ export const requestPasswordResetCallback = createAuthEndpoint(
 	},
 );
 
+export const validatePasswordResetToken = createAuthEndpoint(
+	"/validate-reset-password-token",
+	{
+		method: "POST",
+		body: z.object({
+			token: z.string().meta({
+				description: "The token to validate",
+			}),
+		}),
+		metadata: {
+			openapi: {
+				description: "Validate a password reset token without consuming it",
+				responses: {
+					"200": {
+						description: "Token validation result",
+						content: {
+							"application/json": {
+								schema: {
+									type: "object",
+									properties: {
+										valid: {
+											type: "boolean",
+										},
+										message: {
+											type: "string",
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	},
+	async (ctx) => {
+		const { token } = ctx.body;
+		if (!token) {
+			return ctx.json({
+				valid: false,
+				message: "Token is required",
+			});
+		}
+
+		const id = `reset-password:${token}`;
+
+		const verification =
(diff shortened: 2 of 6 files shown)


jer-tan:
Hello @Bekacru, this is a simple helper function to validate the password reset token, I find it helpful in my project, would be great if you can merge this PR :D

Fyrlex:
Update? It looks like it has been a while.

gleandroj:
+1

jer-tan:
@himself65 This is ready to merge and all tests are passing :D

packages/better-auth/src/api/routes/password.ts
```diff
@@ -226,6 +226,75 @@ export const requestPasswordResetCallback = createAuthEndpoint(
...
+				message: "Invalid token",
+			});
+		}
+
+		if (verification.expiresAt < new Date()) {
+			return ctx.json({
+				valid: false,
+				message: "Token has expired",
+			});
+		}
```
ping-maxwell, line 289: While the token itself is randomly generated (24 chars), the differentiation between "Invalid token" and "Token has expired" messages reveals timing information about token creation which can be used to enumerate whether a user has requested a password reset (and therefore exists).

I recommend keeping the same error message of "Invalid token"

.changeset/validate-reset-password-token.md
```diff
@@ -0,0 +1,5 @@
+---
+"@better-auth/better-auth": minor
```
ping-maxwell: this should be `better-auth` not `@better-auth/better-auth`.

ping-maxwell (changes requested):
we need to add default rate-limits to this new endpoint also.

ping-maxwell (approved):

jer-tan:
Any update on this?

PR #5424, opened 2025-10-20
fix(organization): refetch organization list on membership changes

aryasaatvik:
## Summary

Fixes issue where `useListOrganizations()` hook does not automatically refetch after organization membership changes, causing stale data in UI components like workspace switchers.

## Problem

When a user accepts an organization invitation via `organization.acceptInvitation()`:
- ✅ `useActiveOrganization()` refetches automatically
- ❌ `useListOrganizations()` does NOT refetch - shows stale data until manual browser refresh

## Root Cause

The `atomListeners` configuration only triggered `$listOrg` for create/delete/update operations, but NOT for membership changes like accepting invitations, being removed, or leaving organizations.

## Solution

Updated the `atomListeners` matcher in `packages/better-auth/src/plugins/organization/client.ts` to trigger `$listOrg` signal for:
- `/organization/accept-invitation` - user joins organization
- `/organization/remove-member` - user removed from organization
- `/organization/leave` - user leaves organization

## Testing

After this fix, components using `useListOrganizations()` will automatically display updated organization lists when membership changes occur, without requiring manual refresh or workarounds.

## Related

Removes the need for manual workarounds like:
```typescript
authClient.$store.notify("$listOrg");
```


ping-maxwell (changes requested):
missing changeset.

PR #5431, opened 2025-10-20
feat(sso): add option to encrypt oidc client secret at rest

### LINKED ISSUE #4291: Encrypt OIDC Config at Rest

The OIDC config used by the SSO plugin is stored in the database as plain-text; this includes the `clientSecret` field which is sensitive information.

This config (or at least the `clientSecret`) should be encrypted at rest using the `BETTER_AUTH_SECRET` to avoid exposure of sensitive information.

N/A

### Additional context

_No response_

tnkuehne:
Closes #4291 

Currently the client secret of any registered oidc provider is saved as plain text in the database.

This PR changes this by encrypting the client secret with BETTER_AUTH_SECRET at rest.

    

---
## Summary by cubic
Adds a configurable way to store OIDC client secrets in the SSO plugin. Secrets can be kept plain, encrypted with BETTER_AUTH_SECRET, or encrypted via a custom encryptor; they’re decrypted on read and during sign‑in.

- **New Features**
  - Added sso({ storeSecretAs }) with "plain" (default), "encrypted", or custom { encrypt, decrypt }.
  - Encrypt on write and return decrypted clientSecret via API; decryption failures return a clear API error.
  - Updated docs and added tests for encrypted and custom encryptor flows.

- **Migration**
  - Migrate existing plaintext secrets before switching storeSecretAs to "encrypted" or a custom encryptor.
  - Restrict access to endpoints that return decrypted clientSecret.

<sup>Written for commit ac5b45d6e18cb579c0f1103e58c79ce2f390acab. Summary will update on new commits.</sup>


packages/sso/src/index.ts
```diff
@@ -252,6 +253,11 @@ export interface SSOOptions {
 	 * @default false
 	 */
 	trustEmailVerified?: boolean;
+	/**
+	 * Enables encryption at rest for sensitive fields (e.g. OIDC clientSecret).
+	 * When true, values are encrypted with BETTER_AUTH_SECRET before storing.
+	 */
+	encryptionEnabled?: boolean;
```
Bekacru: We should probably follow the pattern used in other plugins like Email OTP and support `"hashed"`, `"plain"`, `"encrypted"`, as well as custom hash or custom encryption methods.
Let’s also rename the option to `storeSecretAs`.

Reference: [email-otp](https://github.com/better-auth/better-auth/blob/8bbe7d5d17f34b7a7af44d4ad0653b5f12e37fcc/packages/better-auth/src/plugins/email-otp/index.ts#L78-L79)](https://github.com/better-auth/better-auth/blob/8bbe7d5d17f34b7a7af44d4ad0653b5f12e37fcc/packages/better-auth/src/plugins/email-otp/index.ts#L78-L79)

Bekacru (changes requested):

packages/sso/src/index.ts
tnkuehne: Good idea. Implemented it, but left out `"hashed"` as it doesn't make sense in this context.

isaacwasserman:
Any update on this? This encryption really feels like a non-negotiable for compliance.

packages/sso/src/index.ts
```diff
@@ -86,180 +39,2349 @@ const fastValidator = {
...
+						},
+					});
+
+					return ctx.json({
+						...provider,
+						oidcConfig: await (async () => {
+							const cfg = JSON.parse(
+								provider.oidcConfig as unknown as string,
+							) as OIDCConfig;
+							if (cfg?.clientSecret) {
```
Paola3stefania, line 924: This decrypt block is duplicated 3 times in this file ...

It would be a great idea if we extract it into a helper like:

``` typescript
 async function decryptClientSecret(
   cfg: OIDCConfig,
   secret: string,
   storeSecretAs: SSOOptions["storeSecretAs"],
 ): Promise<OIDCConfig> {
   if (!cfg?.clientSecret || !storeSecretAs || storeSecretAs === "plain") return cfg;
   if (typeof storeSecretAs === "object" && "decrypt" in storeSecretAs) {
     cfg.clientSecret = await storeSecretAs.decrypt(cfg.clientSecret);
   } else if (storeSecretAs === "encrypted") {
     try {
       cfg.clientSecret = await symmetricDecrypt({ key: secret, data: cfg.clientSecret });
     } catch {
       throw new APIError("INTERNAL_SERVER_ERROR", {
         message: "Failed to decrypt client secret. Ensure BETTER_AUTH_SECRET matches the key used during encryption.",
       });
     }
   }
   return cfg;
 }

```

packages/sso/src/index.ts
```diff
@@ -86,180 +39,2349 @@ const fastValidator = {
...
+					if (existingProvider) {
+						ctx.context.logger.info(
+							`SSO provider creation attempt with existing providerId: ${body.providerId}`,
+						);
+						throw new APIError("UNPROCESSABLE_ENTITY", {
+							message: "SSO provider with this providerId already exists",
+						});
+					}
+
+					const clientSecretToStore = body.oidcConfig?.clientSecret
```
Paola3stefania, line 853: We should extract the encrypt logic into a shared helper ( **encryptClientSecret**) and call it in both `registerSSOProvider` and `updateSSOProvider`

packages/sso/src/oidc.test.ts
```diff
@@ -568,3 +568,138 @@ describe("provisioning", async (ctx) => {
...
+		plugins: [ssoClient()],
+		baseURL: "http://localhost:3000",
+		fetchOptions: {
+			customFetchImpl,
+		},
+	});
+
+	beforeAll(async () => {
+		await server.issuer.keys.generate("RS256");
+		server.issuer.on;
```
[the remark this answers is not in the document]
Paola3stefania: could we please address this_ @tnkuehne

packages/sso/src/oidc.test.ts
```diff
@@ -571,3 +571,281 @@ describe("provisioning", async (ctx) => {
...
+		token.payload.email_verified = true;
+		token.payload.name = "Test User";
+		token.payload.picture = "https://test.com/picture.png";
+	});
+
+	async function simulateOAuthFlow(
+		authUrl: string,
+		headers: Headers,
+		fetchImpl?: (...args: any) => any,
+	) {
```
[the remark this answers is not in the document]
Paola3stefania: also please could we address this :) @tnkuehne

Paola3stefania (changes requested):
Great work @tnkuehne 

updateSSOProvider doesn't apply encryption when a provider's clientSecret is updated...  (see https://github.com/better-auth/better-auth/pull/5431/changes#r2814377612) 

I will create a follow up PR with a migration helper/CLI command that make adoption of thus safer

Daymannovaes:
This PR is really great. I wonder if would be beneficial if instead we just use a generic field-level encryption-at-rest, something like this: https://github.com/better-auth/better-auth/pull/8337

isaacwasserman:
@Paola3stefania, the original contributor seems to have abandoned this PR. I've addressed your comments in a new PR here https://github.com/better-auth/better-auth/pull/9878

andrew-rheingans:
Definitely would love to see this feature land! Any update?

PR #5439, opened 2025-10-20
fix(organization): allow owners/admins to view invitations

### LINKED ISSUE #4272: `getInvitation` fails when called by an org owner/admin with `YOU_ARE_NOT_THE_RECIPIENT_OF_THE_INVITATION`

### To Reproduce

1. Create an invitation
2. Use any authenticated `owner` or `admin` user and try to execute `auth.api.getInvitation(...)`
3. better-auth [throws an error](https://github.com/better-auth/better-auth/blob/878de4e524780142546f950ffc79d9801e8af17b/packages/better-auth/src/plugins/organization/routes/crud-invites.ts#L863) if the user accessing the invitation is not the invitee:
```
throw new APIError("FORBIDDEN", {
    message:
        ORGANIZATION_ERROR_CODES.YOU_ARE_NOT_THE_RECIPIENT_OF_THE_INVITATION,
});
```

I expect better-auth to return the invitation.
The [docs](https://www.better-auth.com/docs/plugins/organization#permissions) don't mention only invitees can get their own invitations.
The API already returns all invitations when calling `listInvitations`. `getInvitation` should also work.

1.3.7

### System info

```bash
System:
    OS: macOS 15.6.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 351.61 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Brave Browser: 101.1.38.109
    Chrome: 139.0.7258.154
    Safari: 18.6
```

Package

```typescript
const auth = betterAuth({
  plugins: [
    organization({
      ac,
      roles: {
        owner,
        admin,
        member,
      },
    }),
  ],
});
```

### Additional context

I was surprised by this exception when implementing re-sending invitations.
The client sent the `invitationId` back to server and the backend used the `getInvitation` method but got an exception.
To fix this I have to call `listInvitations` + `find(...)`, which is not ideal, or just drop down to accessing the DB directly, which kinda defeats the purpose of having this API which already checks permissions.

All other available methods in the organization plugin properly check permissions. The `getInvitation` method should do the same.

ahmed-abdat:
Fixes #4272

### Problem
The `getInvitation` endpoint only allowed invitation recipients to view invitations. Organization owners and admins should also be able to view invitations in their organization for management purposes.

### Solution
Added permission check using `hasPermission` with `invitation: ["cancel"]` permission, which is granted to owners and admins by default. This follows the same pattern used in the `cancelInvitation` endpoint.

### Tests
- [x] Recipients can view their invitations
- [x] Owners can view any invitation in their org
- [x] Admins can view any invitation in their org
- [x] Regular members cannot view invitations they're not recipient of
- [x] All organization tests pass
- [x] Biome lint check passes
    

---

## Summary by cubic
Allow org owners and admins to view invitations via getInvitation, matching cancelInvitation permissions. Regular members still can’t view others’ invites.

- **Bug Fixes**
  - Added hasPermission check (invitation: ["cancel"]) for non-recipient access; owners/admins allowed by default.
  - Return 403 for non-members and members without this permission, with YOU_ARE_NOT_THE_RECIPIENT_OF_THE_INVITATION.
  - Added tests for recipient, owner, admin, and regular member cases.

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-invites.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-invites.ts
@@ -957,11 +957,40 @@ export const getInvitation = <O extends OrganizationOptions>(options: O) =>
 					message: "Invitation not found!",
 				});
 			}
-			if (invitation.email.toLowerCase() !== session.user.email.toLowerCase()) {
-				throw new APIError("FORBIDDEN", {
-					message:
-						ORGANIZATION_ERROR_CODES.YOU_ARE_NOT_THE_RECIPIENT_OF_THE_INVITATION,
+			const isRecipient =
+				invitation.email.toLowerCase() === session.user.email.toLowerCase();
+
+			if (!isRecipient) {
+				const member = await adapter.findMemberByOrgId({
+					userId: session.user.id,
+					organizationId: invitation.organizationId,
 				});
+
+				if (!member) {
+					throw new APIError("FORBIDDEN", {
+						message:
+							ORGANIZATION_ERROR_CODES.YOU_ARE_NOT_THE_RECIPIENT_OF_THE_INVITATION,
+					});
+				}
+
+				const canViewInvitation = await hasPermission(
+					{
+						role: member.role,
+						options: ctx.context.orgOptions,
+						permissions: {
+							invitation: ["cancel"],
+						},
+						organizationId: invitation.organizationId,
+					},
+					ctx,
+				);
+
+				if (!canViewInvitation) {
+					throw new APIError("FORBIDDEN", {
+						message:
+							ORGANIZATION_ERROR_CODES.YOU_ARE_NOT_THE_RECIPIENT_OF_THE_INVITATION,
+					});
+				}
 			}
 			const organization = await adapter.findOrganizationById(
 				invitation.organizationId,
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/plugins/organization/routes/crud-invites.ts
```diff
@@ -957,11 +957,40 @@ export const getInvitation = <O extends OrganizationOptions>(options: O) =>
...
+							ORGANIZATION_ERROR_CODES.YOU_ARE_NOT_THE_RECIPIENT_OF_THE_INVITATION,
+					});
+				}
+
+				const canViewInvitation = await hasPermission(
+					{
+						role: member.role,
+						options: ctx.context.orgOptions,
+						permissions: {
+							invitation: ["cancel"],
```
himself65, line 981: User could define custom invitation permission here

ahmed-abdat:
@himself65 Good catch! You're right about the hardcoded permission.

I noticed `listInvitations` actually has no permission check at all - any member can see all invitations. So we have an inconsistency:
- Regular members can view all invitations via `listInvitations` 
- But can't view individual invitations via `getInvitation` (blocked by the "cancel" check)

Two options:

**Option A**: Remove the permission check here to match `listInvitations` behavior (any member can view)

**Option B**: Add "read" to the default invitation permissions and check for that instead

I'd lean toward Option A since it's consistent with how `getFullOrganization` works (membership-based access), but happy to go either way. What do you think?

PR #5471, opened 2025-10-21
[rolldown bug]: Error: Panic in async function

himself65:
## Summary by cubic
Lazy-loads database adapters so we only import Kysely/memory backends when needed, reducing startup cost and bundle size. Also simplifies init imports.

- **Refactors**
  - Switched to dynamic imports in getAdapter for memoryAdapter, createKyselyAdapter, and kyselyAdapter.
  - Kept MemoryDB as a type-only import.
  - Updated init imports: use ./crypto for password helpers and ./db for getAdapter.


bytaesu:
Oh I kept getting that error too when trying dynamic import, so I split it into a separate file. Hope this gets fixed 👀

PR #5481, opened 2025-10-22
Remove server route handler instructions from TanStack Start Integration Guide

max-programming:
The "server route handler" (i.e. `/src/server.ts` file) is no longer needed in the recent TanStack Start versions so it's redundant to keep it in this guide


max-programming:
Ok this turned into a disaster
Will open a new one

Issue #5467, opened 2025-10-21
Oauth Error callback URL discrepencies


ZiggyQubert:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Specify an errorCallbackURL URL when calling the google social provider, ensure that skipStateCookieCheck is false, code takes you to the default error page rather than using the calculated error url

```
if (!skipStateCookieCheck && (!stateCookieValue || stateCookieValue !== state)) {
    const errorURL = c.context.options.onAPIError?.errorURL || `${c.context.baseURL}/error`;
    throw c.redirect(`${errorURL}?error=state_mismatch`);
  }
```
Note that its using c.context.options.onAPIError rather than the parsedData.errorURL

### Current vs. Expected behavior

I expect to be taken to the specified error page rather than the default error page

### What version of Better Auth are you using?

1.3.28

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Home",
    "release": "10.0.26100",
    "cpuCount": 16,
    "cpuModel": "Intel(R) Core(TM) Ultra 9 285H",
    "totalMemory": "31.50 GB",
    "freeMemory": "9.94 GB"
  },
  "node": {
    "version": "v22.15.1",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.2"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "15.5.3"
    },
    {
      "name": "react",
      "version": "^19.0.0"
    }
  ],
  "databases": [
    {
      "name": "@prisma/client",
      "version": "^6.9.0"
    }
  ],
  "betterAuth": {
    "version": "^1.3.28",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

note that this directly effects expo where you need to be taken to a native page, rather than a backend page

ZiggyQubert:
@dosu true however previous to using the error the OAuth state has already ben retrieved and the errorURL parsed out and defaulted if necessary, its just not being used after its ben calculated

ping-maxwell:
Hello can you confirm this is still an issue on the latest version of Better Auth?

arunanshub:
[...]

PR #5521, opened 2025-10-23
PR: Fix — OAuth per-flow errorCallbackURL on state-mismatch (#5467)

Adityakk9031:
## Summary

This PR addresses issue **#5467 — OAuth Error callback URL discrepancies**. When an OAuth state cookie check failed (state cookie missing or mismatch), the code always fell back to the global `onAPIError.errorURL` (or the default backend error page) and ignored the per-flow `errorCallbackURL` that is stored in the OAuth state. This prevented client/native flows (e.g., Expo deep links) from being redirected to their per-flow error handlers.

This change implements an **opt-in, safe mechanism** to allow using the parsed per-flow `errorCallbackURL` on state-mismatch scenarios **only when the origin of that per-flow URL is explicitly trusted** via a new `trustedErrorRedirectOrigins` option.

---

## Files changed (high-level)

1. **`packages/core/src/init-options.ts`** (types)

   * Add `trustedErrorRedirectOrigins?: string[]` to `BetterAuthOptions` so that the option is typed and documented.

2. **`packages/better-auth/src/init.ts`** (runtime initialization)

   * Merge a default value into runtime options: `trustedErrorRedirectOrigins: options.trustedErrorRedirectOrigins ?? []` so the option is always defined at runtime.

3. **`packages/better-auth/src/oauth2/state.ts`** (state parsing / error redirect)

   * Update `parseState` to prefer the per-flow `errorURL` on a state-cookie mismatch *only if* that URL's origin is present in `ctx.options.trustedErrorRedirectOrigins` (opt-in). Otherwise it falls back to the existing global error URL.

---

## Motivation & security rationale

* **Problem:** per-flow `errorCallbackURL` is stored inside the OAuth `state` verification record. When the state cookie mismatch occurs, the code previously did not use the parsed per-flow `errorCallbackURL` because the cookie mismatch made the state appear untrusted.

* **Risk:** blindly trusting the `state` payload when cookies are missing opens the door to open-redirect and CSRF-like misuse.

* **Solution:** make per-flow redirects available only when the developer explicitly opts-in via a whitelist of trusted origins (`trustedErrorRedirectOrigins`). By default this list is empty, preserving the existing secure behavior.

This keeps the security posture conservative (opt-in) while giving native clients (Expo / deep links) a way to receive per-flow redirects for state-mismatch errors.

---

### `init-options.ts`

* Added to `BetterAuthOptions`:

```ts
/**
 * Optional whitelist of origins that are allowed as per-flow error redirect targets.
 * Example: ["https://app.example.com", "myapp://host"]
 * If omitted or empty, parsed per-flow errorCallbackURL will NOT be used on state-mismatch.
 */
trustedErrorRedirectOrigins?: string[];
```

### `init.ts`

* Ensure runtime `options` always has the field defined (defaults to `[]`):

```ts
options = {
  ...options,
  // existing defaults ...
  trustedErrorRedirectOrigins: options.trustedErrorRedirectOrigins ?? [],
};
```

### `oauth2/state.ts` (`parseState` change)

* Current behavior (before): on state-cookie mismatch the code used `c.context.options.onAPIError?.errorURL || default`.
* New behavior: compute `redirectTo` as follows:

  1. `defaultErrorURL = c.context.options.onAPIError?.errorURL || `${c.context.baseURL}/error``
  2. If `parsedData.errorURL` exists **and** `ctx.options.trustedErrorRedirectOrigins` is non-empty, try to parse `parsedData.errorURL` (using `new URL(parsedData.errorURL, c.context.baseURL)` to support relative URLs).
  3. If parsing succeeds and `url.origin` is present in `ctx.options.trustedErrorRedirectOrigins`, use `parsedData.errorURL` as the redirect target.
  4. Otherwise fall back to `defaultErrorURL`.

This logic ensures developers must explicitly list client/native deep-link origins in order to enable per-flow redirects in state-mismatch scenarios.

---

## Example configuration (docs / README snippet)

```ts

  // ...other options
  onAPIError: { errorURL: "https://example.com/error" },
  // Opt-in whitelist for per-flow error redirect targets
  trustedErrorRedirectOrigins: [
    "https://app.example.com",
    "myapp://host",
  ],
});
```

* For Expo/native error deep links, add the app deep link origin (e.g., `myapp://host`).
* **Warning**: Only list origins you control. This is opt-in and must be populated by the developer to allow per-flow error redirects on state-mismatch.

---

## Tests added / manual validation

1. **Unit tests** (suggested):

   * `parseState` when:

     * a) `parsedData.errorURL` is present and `trustedErrorRedirectOrigins` includes its origin ⇒ redirected to `parsedData.errorURL` on state-cookie mismatch
     * b) `parsedData.errorURL` present but `trustedErrorRedirectOrigins` is empty ⇒ redirected to global error URL * c) `parsedData.errorURL` present but origin not in whitelist ⇒ global error URL * d) `parsedData.errorURL` is relative (e.g., `/err`) and baseURL origin is in whitelist ⇒ redirect to resolved URL
   * Negative tests for malformed URLs in `parsedData.errorURL` (should fall back to global error URL)

2. **Manual reproduction (QA)**

   * Reproduce original issue steps (from #5467):

     * Start OAuth flow with an `errorCallbackURL` set to `myapp://host/err` and state stored in DB.
     * Simulate missing/mismatched state cookie on callback. * With `trustedErrorRedirectOrigins: []` (default) — expect redirect to global backend error page. * With `trustedErrorRedirectOrigins: ["myapp://host"]` — expect redirect to `myapp://host/err?error=state_mismatch`.

---

## Docs

* Add `trustedErrorRedirectOrigins` to BetterAuth options docs, explaining purpose, examples, and security warning.
* Add migration note: this option is **opt-in** (default `[]`) — existing behavior unchanged.

---

## Security & backward compatibility

* Default behavior unchanged (secure): if `trustedErrorRedirectOrigins` is empty or not set, **per-flow** `errorCallbackURL` will **not** be used for state-mismatch redirections.
* Developers must explicitly opt-in by providing a list of origins they control.

---

## Release notes

* Feature: allow per-flow OAuth `errorCallbackURL` to be used on state-cookie mismatch **if** the target origin is whitelisted via `trustedErrorRedirectOrigins`.
* Security: default is secure (empty whitelist).

---

## Checklist

* [ ] Type definitions updated (`init-options.ts`)
* [ ] Runtime default merged (`init.ts`)
* [ ] `parseState` modified to check whitelist before honoring per-flow `errorCallbackURL` (`state.ts`)
* [ ] Unit tests added/updated
* [ ] Docs entry added
* [ ] Changelog entry added

---

## Reviewer notes

* This change introduces a new opt-in surface (`trustedErrorRedirectOrigins`) — please carefully review the whitelist check in `parseState`.
* Suggested reviewers: security-focused maintainers and OAuth plugin owners.

---


gustavovalverde:
Superseded by #9789, which fixes #5467 with a narrower approach. Thanks for surfacing the right direction (recover the per-flow error URL from the parsed state instead of always using the default).

The new PR drops the `trustedErrorRedirectOrigins` allowlist this PR proposed. It turned out to be unnecessary: `errorCallbackURL` is already validated against `trustedOrigins` at sign-in (`origin-check.ts` rejects an untrusted value with `INVALID_ERROR_CALLBACK_URL` before any state is created), so a URL that reached the parsed state was provably trusted when stored. A second allowlist would duplicate `trustedOrigins`.

This branch also predates the split of `parseState` into `parseGenericState` plus the `StateError` class, so the diff no longer applies to the current code. Closing in favor of #9789.

PR #5553, opened 2025-10-24
fix: call db hooks when calling `deleteUser`

ping-maxwell:
This change now calls all 3 DB hooks when calling deleteUser:
* session
* account
* user

---
## Summary by cubic
Fixed deleteUser to call DB hooks for sessions, accounts, and user records so hooks run and data stays consistent.

- **Bug Fixes**
  - Switched account deletion to use deleteManyWithHooks in the internal adapter.
  - Removed direct deleteAccounts call in the route; the adapter now handles account deletion and hooks.

### DIFF
--- a/docs/components/builder/code-tabs/index.tsx
+++ b/docs/components/builder/code-tabs/index.tsx
@@ -102,7 +102,7 @@ ${
 		initialFiles.push({
 			id: "4",
 			name: "sign-up.tsx",
-			content: signUpString,
+			content: signUpString(options),
 		});
 	}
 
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/api/routes/update-user.ts
```diff
@@ -526,7 +526,6 @@ export const deleteUser = createAuthEndpoint(
 		}
 		await ctx.context.internalAdapter.deleteUser(session.user.id);
 		await ctx.context.internalAdapter.deleteSessions(session.user.id);
-		await ctx.context.internalAdapter.deleteAccounts(session.user.id);
```
ping-maxwell, line 529: removed this here because `deleteUser` internal adapter already handles this

PR #5569, opened 2025-10-25
Fix(admin-user): Add min/max password length validation

hieudien14310:
## Summary by cubic
Enforces min/max password length when admins create users, preventing weak or overly long passwords. Adds a reusable length check and returns clear error codes.

- **Bug Fixes**
  - Validate password length in admin.createUser using checkPasswordLength.
  - Expose checkPasswordLength on ctx.context.password; respects configured minPasswordLength and maxPasswordLength (max defaults to 128).
  - Return BAD_REQUEST with PASSWORD_TOO_SHORT or PASSWORD_TOO_LONG on violations.
  - Update admin tests to use compliant passwords.
  - Add types for checkPasswordLength and wire it in init.


hieudien14310:
I didn't add a test case because simply changing the password = 'test' to 'testuser123', already satisfies the min and max password validation.

hieudien14310:
If this PR is approved, I will refactor all places where password length is currently checked (min and max) to use the `checkPasswordLength` function.
To avoid code dup.

PR #5570, opened 2025-10-25
feat: replace zod type with standard schema

R5dan:
Adds standard schema to better auth

---
## Summary by cubic
Introduce a vendor-agnostic Standard Schema v1 for field validation. This removes the Zod dependency in DB validators and standardizes validate calls across core and better-auth.

- **Refactors**
  - Added StandardSchemaV1 interface (core/types/standard-schema.ts).
  - Replaced ZodType with StandardSchemaV1 for DBFieldAttributeConfig.validator input/output.
  - Updated better-auth parseInputData to call schema["~standard"].validate(...). 

- **Migration**
  - Update any custom DB field validators to implement Standard Schema v1 or wrap existing Zod schemas.


R5dan:
@cubic-dev-ai Can you find any other places where there is a ZodType being used that should be replaced by StandardSchema. Ignore any situations where zod is being used as the validation library to provide the validation, but flag any situations where the user is being forced to use zod

R5dan:
@himself65 @Bekacru @ping-maxwell Do any of you have any feedback on this? Can it be reviewed and merged

ping-maxwell:
hmm I think this is a fine change.
Also I'm pretty sure we've already exported our own standard schema from better-call or somewhere already, so let's use that instead of having another standard schema code in the codebase

R5dan:
> Also I'm pretty sure we've already exported our own standard schema from better-call or somewhere already, so let's use that instead of having another standard schema code in the codebase

@ping-maxwell yes, `better-call` or `better-fetch` export standard schema, however personally I think this should have its own type as they are not "standard schema packages" but instead they both have there own functionality, so it feels weird to me to use them for it, however it is upto you

R5dan:
@himself65 can I ask what happened here? Did you rewrite to make the code cleaner?

PR #5571, opened 2025-10-25
Validator

R5dan:
Requires #5570 

Replaces manual validation by fields like `minPasswordLength` with a `passwordSchema` field that takes a standard schema type for validation. Does not replace the `validatePasswordFunction` (and equivalent) fields.

---
## Summary by cubic
Switch password and username validation to a standard schema-based approach, replacing min/max length checks with a configurable validator. Adds a StandardSchemaV1 interface and updates routes, context, and plugins to use it while keeping custom password validation functions.

- **Refactors**
  - Added StandardSchemaV1 and replaced min/max password config with password.validator in context (default: z.string().min(8).max(128)).
  - Updated sign-up, reset-password, and change/set password endpoints to use schema validation with async support.
  - Migrated username plugin to usernameSchema, removed min/max options, and updated error codes to USERNAME_SCHEMA_VALIDATION_FAILED.
  - Switched DB field validators and parseInputData to use the standard schema validate API.
  - Adjusted telemetry to report schema-based configuration.

- **Migration**
  - Replace emailAndPassword.minPasswordLength/maxPasswordLength with emailAndPassword.validator (StandardSchemaV1).
  - Update any usage of ctx.context.password.config to ctx.context.password.validator.
  - For the username plugin, replace minUsernameLength/maxUsernameLength with usernameSchema and handle the new error code.

### DIFF
--- a/packages/better-auth/src/api/routes/reset-password.ts
+++ b/packages/better-auth/src/api/routes/reset-password.ts
@@ -6,6 +6,8 @@ import { generateId } from "../../utils";
 import { BASE_ERROR_CODES } from "@better-auth/core/error";
 import { originCheck } from "../middlewares";
 import type { AuthContext } from "@better-auth/core";
+import { isPromise } from "../../utils/is-promise";
+import { ERROR_CODES } from "../../plugins";
 
 function redirectError(
 	ctx: AuthContext,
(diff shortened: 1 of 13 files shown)


R5dan:
Standard schema has now been merged via #5627 and #5629, so this pr can be

R5dan:
Standard schema has been merged via #5627 and #5629 and so this pr can be opened

PR #5579, opened 2025-10-25
feat: use separete client SDKs

R5dan:
This pr adds sdks, separating the  packages into multiple small ones.

Adds a main `@better-auth/client` package which contains the utils from  `better-auth/client`, and all the client plugins. Also adds a separate package for each framework, including vanilla (`@better-auth/js`) 

---
## Summary by cubic
Split the client into standalone SDKs. Added @better-auth/client for shared utilities and plugins, plus framework-specific packages for JS, React, Vue, Svelte, Solid, and Lynx to simplify imports and reduce coupling.

- **New Features**
  - New @better-auth/client with config, proxy, types, and plugin exports.
  - New framework SDKs: @better-auth/js, @better-auth/react, @better-auth/vue, @better-auth/svelte, @better-auth/solid, @better-auth/lynx.
  - Removed client exports from better-auth; builds updated per package.

- **Migration**
  - better-auth/client -> @better-auth/js
  - better-auth/client/react|vue|svelte|solid|lynx -> @better-auth/react|vue|svelte|solid|lynx
  - better-auth/client/plugins -> @better-auth/client/plugins
  - better-auth/client/{config,proxy,types} -> @better-auth/client/{config,proxy,types}


packages/js/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@better-auth/js",
```
himself65: vanilla

himself65 (commented):
This looks good, should we put all packages under packages/clients/*?

packages/js/package.json
R5dan: Ok, I will change that

R5dan:
@himself65 done

R5dan:
@himself65 do i need to do anything more? or can this be merged?

himself65:
I think overall is good. But I need double check later

himself65 (changes requested):
IMO, this is a too-breaking change which requires the user to download the client package manually. RN, we still hope the user could use the client SDK directly.

But we could still seperate into different package and re-export from root package

R5dan:
@himself Do you want me to export them all from `better-auth/client/*` but deprecate them for removal in future?

gustavovalverde:
@R5dan if you think this is still needed, please open an issue or discussion instead to agree

PR #5599, opened 2025-10-26
Implement beforeLeave and afterLeave hooks for organization

### LINKED ISSUE #5435: Before/After hooks for leaveOrganization

I'm implementing organizations in my app and noticed that there are no hooks for leaving the organization. Thought that maybe hooks for removeMember will fire, but sadly that is not the case. 

It would be great to have a possibility to cleanup some user data using before/after hooks for leaveOrganization same way as it works for removeMember.

Call cleanup logic where api.leaveOrganization is called. But it leads to logic fragmentation, as some lives in organizations config and some is server logic folder.

### Additional context

_No response_

GautamBytes:
Closes #5435

This PR introduces two new lifecycle hooks to the organization plugin: `beforeLeaveOrganization` and `afterLeaveOrganization`.

This allows developers to run custom logic, such as cleaning up user-related data or sending notifications, when a user voluntarily leaves an organization.

#### Implementation Details
* **`crud-members.ts`**: The `leaveOrganization` endpoint now wraps the deletion logic in a `try...catch` block.
    * `beforeLeaveOrganization` is called *before* the member is deleted. If this hook throws an error, the entire operation is aborted, and the user is not removed.
    * `afterLeaveOrganization` is called *after* the member has been successfully deleted.
* **`organization/types.ts`**: Added the new hook definitions to the `OrganizationHooks` interface.
* **`organization.mdx`**: Added documentation and example usage for the new hooks.
* **`crud-members.test.ts`**: Added new tests to verify:
    1.  Both hooks are called in the correct order on a successful leave.
    2.  If `beforeLeaveOrganization` throws an error, the member is **not** deleted, and `afterLeaveOrganization` is **not** called.

---
## Summary by cubic
Make leaveOrganization run the existing beforeRemoveMember and afterRemoveMember hooks, and pass ctx so you can tell self-leave vs admin removal. Leaving is blocked if beforeRemoveMember throws; afterRemoveMember runs only on success.

- **New Features**
  - Hooks now receive ctx in OrganizationHooks for beforeRemoveMember and afterRemoveMember.
  - leaveOrganization calls these hooks with member, user, organization; aborts on error from beforeRemoveMember.
  - Updated docs and tests for success, order, failure, and context differentiation.

<sup>Written for commit 23999646110f5424124909344b122103814ff8cc. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/organization.mdx
+++ b/docs/content/docs/plugins/organization.mdx
@@ -245,7 +245,7 @@ export const auth = betterAuth({
           // This hook is called by BOTH `removeMember` (admin action)
           // and `leaveOrganization` (self-leave action)
           // You can differentiate them by checking the request URL in the context:
-          if (ctx.request.url.includes("/organization/leave")) {
+          if (ctx.request?.url.includes("/organization/leave")) {
             console.log("User is leaving the organization themselves.");
           } else {
             console.log("User is being removed by an admin.");
--- a/docs/content/docs/plugins/organization.mdx
+++ b/docs/content/docs/plugins/organization.mdx
@@ -261,10 +261,6 @@ export const auth = betterAuth({
           await logMemberRemoval(user.id, organization.id);
         },
 
-        <Callout type="info">
-          Throwing an error in `beforeRemoveMember` will prevent the operation in both cases (admin removal and self-leave).
-        </Callout>
-
         // Before updating a member's role
         beforeUpdateMemberRole: async ({
           member,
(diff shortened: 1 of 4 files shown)


GautamBytes:
@ping-maxwell its ready for review!

Bekacru:
I think we should trigger the remove member hook instead of adding leave hook. Leave and remove are essentially the same concept, the only difference is who’s initiating the action. Since we pass `ctx`, people can still differentiate between the leave and remove endpoints if they want to implement different logic for each.

GautamBytes:
@Bekacru can u review it again , messed up with conflicts but its fixed now along with ur suggestions , so you can safely ignore bot comments

PR #5608, opened 2025-10-27
docs: provide og schema defaults

ibxbit:
## Summary by cubic
Added defaults to the OG image API schema to prevent Zod errors during prerender when query params are missing. This makes OG image generation reliable.

- **Bug Fixes**
  - Set zod defaults: heading "Better Auth Documentation", mode "dark", type "documentation".


ibxbit (commented):
provide og schema defaults

PR #5610, opened 2025-10-27
refactor(oidc-provider): Remove dependency on request object

momoneko:
With the current setup, in order to call the oidc provider server side, I need to build the request object and pass it directly, which is inconsistent with the main api calls, such as sign-in or get-session.

E.g.
```typescript
    /* req is the request object passed from Nest */
    const request = {
      headers: fromNodeHeaders(req.headers),
      url: req.url,
    } as Request;
    const response = await this.auth.api.oAuth2authorize({
      request: request,
      asResponse: true,
    });
```

This PR removes the requirement to pass in requests and gets the data directly from `ctx`, allowing code such as 
```typescript
    const response = await this.auth.api.oAuth2authorize({
       headers: fromNodeHeaders(req.headers),
       query: req.query,       
       asResponse: true,
    });
```

---
## Summary by cubic
Removes the OIDC provider’s dependency on the Request object by reading headers and query from ctx, making server-side calls consistent with sign-in and get-session.

- **Refactors**
  - Read authorization, sec-fetch-mode, and x-forwarded-for from ctx.headers.
  - Build login redirects from URLSearchParams(ctx.query) instead of parsing ctx.request.url.
  - Remove UNAUTHORIZED error when ctx.request is missing.

- **Migration**
  - Stop passing a Request object to oAuth2authorize. Pass headers and query instead.

<sup>Written for commit a366bf3af5568bf0ae58d9c5571be62634e126fc. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/authorize.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/authorize.ts
@@ -67,7 +67,7 @@ export async function authorize(
 			},
 		);
 		const queryParams = new URLSearchParams(ctx.query);
-		return handleRedirect(`${options.loginPage}?${queryParams.toString}`);
+		return handleRedirect(`${options.loginPage}?${queryParams.toString()}`);
 	}
 
 	const query = ctx.query as AuthorizationQuery;


gustavovalverde:
#4163 is the primary resolution of this issue. The new `@better-auth/oauth-provider` package replaces the plugin entirely with a better architecture.

Because the old oidc-provider plugin is deprecated and will be removed, modifying it is no longer worthwhile.

Issue #5600, opened 2025-10-26
Session is null


timothygachengo:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Login with any sign-in method (email, google, e.t.c) when using expo.
2. Try and use `authClient.getSession()`
 3. `authClient.getSession()` returns `null`

### Current vs. Expected behavior

`authClient.getSession()` should return the user data.

### What version of Better Auth are you using?

1.3.31

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.0.0: Wed Sep 17 21:35:32 PDT 2025; root:xnu-12377.1.9~141/RELEASE_ARM64_T6020",
    "release": "25.0.0",
    "cpuCount": 12,
    "cpuModel": "Apple M2 Pro",
    "totalMemory": "16.00 GB",
    "freeMemory": "0.16 GB"
  },
  "node": {
    "version": "v22.20.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.3"
  },
  "frameworks": [
    {
      "name": "react",
      "version": "19.0.0"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "1.3.31",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Backend, Client

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});
```

Mokhtar25:
facing the same issue, workaround to downgrade to 1.3.27

lucasoliveirabr:
I'm facing the same issue and downgrading back to v.1.3.27 is really a workaround for now. Thanks @Mokhtar25.

Mokhtar25:
after some further investigating, turns out that they added cookiePrefix on the client in that update. so if the prefix missmatch on the client and the server you get that error. what to do is to make sure the cookie prefix is the same on client and server after that update.

lucasoliveirabr:
Adding the cookiePrefix to expoClient in authClient solved the issue.
Thanks again @Mokhtar25!

ping-maxwell:
[...]

PR #5646, opened 2025-10-28
feat(expo): error handling when not set

okisdev:
This PR adds error handling when `expo()` is not set in `auth.ts` on server side.

Related issue #5600.

---
## Summary by cubic
Show a helpful error in native Expo apps when the server isn’t configured with `expo()`. Prevents silent failures by logging guidance during auth calls.

- **Bug Fixes**
  - Added `onError` in the Expo client to log a clear configuration error (with docs link) when a 403 occurs on `/sign-in`, `/sign-up`, or `/get-session` and the request has the `expo-origin` header. Runs only on native (not web).
  - Added a test that verifies the console error includes guidance to add `expo()` on the server.

<sup>Written for commit 38b7466170bc9cd2a58489da0b6ff195f3297e3e. Summary will update on new commits.</sup>

### DIFF
--- a/packages/expo/src/client.ts
+++ b/packages/expo/src/client.ts
@@ -317,6 +317,28 @@ export const expoClient = (opts: ExpoClientOptions) => {
 				id: "expo",
 				name: "Expo",
 				hooks: {
+					async onError(context) {
+						if (isWeb) return;
+
+						const { error, request } = context;
+						const url = String(request.url);
+
+						if (error.status === 403) {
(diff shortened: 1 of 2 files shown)


codewithbharat:
Hit this same gap testing against current main — if the server-side expo() plugin isn't wired up, requests just fail with a bare 403 and nothing points at the cause. Checked client.ts and it's still not handled there, so this is still applicable ten months later. Would be nice to get unstuck if anyone has review bandwidth.

Issue #5657, opened 2025-10-29
No way to get BIGINT columns for IDs?


aybabtme:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

When requesting numeric IDs, the schema generate uses 32 bits integer columns instead of 64 bits columns.

```
advanced: {
  database: {
    useNumberId: true,
  },
},
```

### Current vs. Expected behavior

It should generate 64 bits integer columns by default, or at least provide the ability to use 64 bits columns.

While some columns like `organization.id` and `user.id` might ostensibly make sense as 32bits integers (if one expects a small number of users compared to the world's population), columns like `session.id` should absolutely be 64 bits to avoid overflow. This is an issue that many companies faced in the past as sessions can increment quickly beyond the range representable by 32 bits integers.

## Why UUIDs are not a solution.

While UUIDs sidestep this problem completely, they come with their own set of massive problems. Most databases perform poorly with random IDs as primary keys, and so performance sensitive companies tend to stick to integer IDs. Random strings as ID cause index fragmentation, poor cache locality and cache thrashing, which kill database performance. Also UUIDs compress poorly and have larger column width, leading to worse join performance and disk I/O.

## Why ULIDs are not a solution

ULIDs help with the caching/index downsides of UUIDs but have poor native support by database engines and also compress poorly compared to integer sequences. They _may_ be a stop gap solution that's *less bad than 32 bits integers or UUIDs* but really, 64 bits integers have been used in production at scale by large tech companies for decades now, it's kind of expected as an option.

### What version of Better Auth are you using?

1.3.33

### System info

```bash
"system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.0.0: Wed Sep 17 21:41:45 PDT 2025; root:xnu-12377.1.9~141/RELEASE_ARM64_T6000",
    "release": "25.0.0",
    "cpuCount": 10,
    "cpuModel": "Apple M1 Pro",
    "totalMemory": "16.00 GB",
    "freeMemory": "0.11 GB"
  },
  "node": {
    "version": "v23.3.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.0"
  },
  "frameworks": null,
  "databases": [
    {
      "name": "mysql2",
      "version": "^3.11.0"
    }
  ],
  "betterAuth": {
    "version": "^1.3.33",
    "config": null
  }
}
```

[...]

PR #5661, opened 2025-10-29
fix(cli): use 64 bit ints for numeric IDs for mysql/mssql

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/5657

Should generate 64bit for ids with `useNumberId`. We don't do this for mysql & mssql on kysely, Prisma & Drizzle. This PR resolves such for the CLI generator & handles IDs as `BigInt` in the adapter factory.

---
## Summary by cubic
Switch numeric IDs to 64-bit (BIGINT) for MySQL and MSSQL when useNumberId is enabled, and handle IDs as BigInt throughout the adapter. This prevents truncation and keeps the CLI generator and runtime consistent.

- **Bug Fixes**
  - Generate BIGINT columns for numeric IDs in MySQL/MSSQL migrations.
  - Convert ID values to BigInt in the adapter factory for reads/writes.
  - Allow BigInt and BigInt[] in query Where values.

- **Migration**
  - If you previously created INTEGER ID columns on MySQL/MSSQL with useNumberId, alter them to BIGINT.
  - Update application code to use BigInt for IDs (not Number), including query filters and payloads.

### DIFF
--- a/packages/cli/src/generators/drizzle.ts
+++ b/packages/cli/src/generators/drizzle.ts
@@ -56,7 +56,7 @@ export const generateDrizzleSchema: SchemaGenerator = async ({
 					if (databaseType === "pg") {
 						return `integer('${name}')`;
 					} else if (databaseType === "mysql") {
-						return `bigint('${name}')`;
+						return `bigint('${name}', { mode: 'bigint' })`;
 					} else {
 						// using sqlite
 						return `integer('${name}')`;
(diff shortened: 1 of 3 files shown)


LKNSI:
Hey! Have been following along on this. Implementing [Snowflakes](https://en.wikipedia.org/wiki/Snowflake_ID) was something we were excited to do, but this seems to be still blocking us. We're running Postgres + Prisma - and would love, some more love here if possible; we've tried writing adapters - but they flake and kinda overlap with our existing prisma models/controllers. 

(Happy Holidays! <3 btw!)

PR #5668, opened 2025-10-30
fix(slack): scopes should use comma delimiter

Bekacru:
## Summary by cubic
Fix Slack OAuth by using comma-separated scopes in the authorize URL. This prevents invalid_scope errors and ensures permissions are requested correctly.


himself65:
Reverted

PR #5679, opened 2025-10-30
fix(admin): implement missing adminRoles permission bypass logic

jero237:
## Description

This PR fixes the `adminRoles` option in the admin plugin which was documented but not functioning. The option was defined in the types and set with a default value of `["admin"]`, but the actual permission bypass logic was never implemented in `has-permission.ts`.

## Changes

- Implemented missing permission bypass logic in `has-permission.ts` for the existing `adminRoles` option
- Added comprehensive test coverage (10 new tests covering multiple scenarios)

## Problem

The `adminRoles` option was documented in the TypeScript types with clear JSDoc comments stating:

> "Roles that are considered admin roles. Any user role that isn't in this list, even if they have the permission, will not be considered an admin."

However, this functionality was never actually implemented. Users setting `adminRoles` would find it had no effect, as the permission check logic in `has-permission.ts` never checked this option.

## Solution

Added the missing implementation in `has-permission.ts` that:
- Checks if a user's role matches any role in the `adminRoles` array
- Automatically grants all permissions to users with admin roles
- Supports both single string and array formats as documented

## Testing

- ✅ All existing tests pass (62/62)
- ✅ Added test suite for `adminRoles` with array of roles
- ✅ Added test suite for `adminRoles` with single string  
- ✅ Tests cover userId and role parameter scenarios
- ✅ Tests verify multi-role users work correctly
- ✅ Code is formatted with BiomeJS

## Example Usage (as originally documented)

```typescript
admin({
  adminRoles: ["superadmin", "moderator"], // or single string: "admin"
  // ... other options
})
```

## Breaking Changes

None - this implements existing documented functionality that was previously non-functional.

## Related Issues

---
## Summary by cubic
Fixes the adminRoles option in the admin plugin so listed roles bypass all permission checks. Supports single or multiple roles and works with both userId and role inputs.

- **Bug Fixes**
  - Implemented adminRoles bypass logic in permission checks.
  - Supports string or array values and multi-role users (any match grants full access).
  - Added tests for array vs. string, userId vs. role, and multi-role scenarios.
  - No breaking changes; matches documented behavior.

<sup>Written for commit cb54ff1. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/admin/admin.test.ts
+++ b/packages/better-auth/src/plugins/admin/admin.test.ts
@@ -1257,3 +1257,406 @@ describe("access control", async (it) => {
 		);
 	});
 });
+
+describe("adminRoles option", async (it) => {
+	const ac = createAccessControl({
+		user: ["create", "read", "update", "delete"],
+		order: ["create", "read", "update", "delete"],
+	});
+
+	const superAdminAc = ac.newRole({
+		user: ["read"],
+		order: ["read"],
+	});
+	const moderatorAc = ac.newRole({
+		user: ["read", "update"],
+		order: ["read"],
+	});
+	const userAc = ac.newRole({
+		user: ["read"],
+		order: ["read"],
+	});
+
+	const { signInWithTestUser, signInWithUser, auth, customFetchImpl } =
+		await getTestInstance(
+			{
+				plugins: [
+					admin({
+						ac,
+						roles: {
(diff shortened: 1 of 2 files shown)


barryShevlin:
Looking at the documentation, I don't think the `adminRoles` option was intended to bypass all permissions like this. This seems like it would implicitly enable the `impersonate-admins` permission, for example.

PR #5681, opened 2025-10-30
fix(generic-oauth): async `mapProfileToUser` breaks

ping-maxwell:
After debugging for an unacceptable amount of time, I found out that our generic oauth plugin with `mapProfileToUser` breaks when users apply `async` to their function despite our function signature supporting this. It breaks because we didn't use `await`...

Thanks so much @michidk for patiently debugging alongside me!

closes https://github.com/better-auth/better-auth/pull/5681
linear https://linear.app/better-auth/issue/ENG-547/fix-generic-oauth-breaks-entirely-with-async-mapprofiletouser

---
## Summary by cubic
Fixes the generic OAuth plugin to correctly handle async mapProfileToUser by awaiting its result when constructing the user object. Prevents auth flow from breaking when mapProfileToUser returns a Promise.

- **Bug Fixes**
  - Await mapProfileToUser(userInfo) so async implementations are supported.
  - Ensures mapped fields merge correctly into the user payload.

<sup>Written for commit 3c2959e. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/generic-oauth/index.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/index.ts
@@ -329,7 +329,7 @@ export const genericOAuth = (options: GenericOAuthOptions) => {
 								emailVerified: userInfo?.emailVerified,
 								image: userInfo?.image,
 								name: userInfo?.name,
-								...c.mapProfileToUser?.(userInfo),
+								...(await c.mapProfileToUser?.(userInfo)),
 							},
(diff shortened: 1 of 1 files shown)


ping-maxwell:
closing mine in favour of https://github.com/better-auth/better-auth/pull/5479/files#diff-9953e0ed52ef7707c7cd72e0b1a3091e5e02d9e4a5d9d935418d2ffed8ed5b74

PR #5684, opened 2025-10-30
fix(jwt): add mutex lock to JWT

### LINKED ISSUE #5663: Prevent duplicate JWKs caused by race conditions during creation and signing

### To Reproduce

```bash
curl https://better-auth-endpoint/jwks & \
> curl https://better-auth-endpoint/jwks & \
> wait
```

With enough luck, you will get into a data race; otherwise, keep trying until you get two entries in the database 😃 

The JWK creation logic in the JWT plugin appears to be prone to data races. The following code, part of the handler for the [/jwks](https://github.com/better-auth/better-auth/blob/cd567659b8b6c78bf90843fd7218446a4ca6aa9b/packages/better-auth/src/plugins/jwt/index.ts#L42) endpoint, illustrates the issue:

https://github.com/better-auth/better-auth/blob/cd567659b8b6c78bf90843fd7218446a4ca6aa9b/packages/better-auth/src/plugins/jwt/index.ts#L134-L141

The current implementation first queries all keys in the database, and if none are found, it proceeds to create a new one. This introduces a race condition: two (or more) concurrent requests may both determine that no keys exist before either has inserted one, resulting in multiple keys being created.
While this doesn’t cause a critical failure (you just end up with two keys), it is still an unintended and inefficient behavior that can be improved.

A similar race condition can occur when signing a token:

https://github.com/better-auth/better-auth/blob/cd567659b8b6c78bf90843fd7218446a4ca6aa9b/packages/better-auth/src/plugins/jwt/sign.ts#L54-L61

JWKs are created lazily (just like the `/jwks` endpoint) if a token is requested but no JWK exists, a new one is generated. However, multiple concurrent requests might all evaluate the last known key as undefined, again leading to duplicate key creation.

Addressing this issue is non-trivial and will likely depend on the database dialect. Ideally, the existence check should be optimized to query specifically for one valid key rather than fetching all keys (getting all keys in the database with all fields is not very optimized just to check if one key exists). The operation should also be wrapped in a transaction or similar concurrency-safe mechanism to ensure that only one JWK can be inserted at a time. Depending on the database, this could be achieved with a row-level or table-level lock, or by leveraging unique constraints and catching duplicate-insert errors to enforce atomicity without locking the entire table.

1.3.34

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.6.0: Mon Aug 11 21:15:09 PDT 2025; root:xnu-11417.140.69.701.11~1/RELEASE_ARM64_T6041",
    "release": "24.6.0",
    "cpuCount": 16,
    "cpuModel": "Apple M4 Max",
    "totalMemory": "64.00 GB",
    "freeMemory": "0.16 GB"
  },
  "node": {
    "version": "v20.15.1",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.8.2"
  },
  "frameworks": [
    {
      "name": "react",
      "version": "^19.1.0"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null
  }
}
```

Backend

```typescript

```

### Additional context

_No response_

okisdev:
This PR adds a mutex lock to ensure concurrency safety for JWT.

This PR fixes #5663.

---
## Summary by cubic
Adds application-level concurrency control for JWK creation to prevent duplicate keys when JWKS and token signing are called in parallel. Fixes #5663.

- **Bug Fixes**
  - Serialize JWK creation with an app-level mutex.
  - Added adapter.getOrCreateLatestKey and use it in JWKS and signJWT paths.
  - Ensures only one latest key exists under concurrent requests.

- **Refactors**
  - Split key generation into generateJwkData; createJwk now delegates to it.
  - Updated jwt/index and sign.ts to use the new adapter flow.
  - Added tests for concurrent JWKS, token signing, signJWT, and mixed scenarios.

<sup>Written for commit 5e3eb75a811e839d30416bc8ab36771281307078. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/jwt/jwt.test.ts
+++ b/packages/better-auth/src/plugins/jwt/jwt.test.ts
@@ -906,29 +906,6 @@ describe("jwt - race condition prevention", async () => {
 });
 
 describe("jwt - edge cases and error handling", async () => {
-	it("should handle database query errors gracefully", async () => {
-		const { auth } = await getTestInstance({
-			plugins: [jwt()],
-			logger: {
-				level: "error",
-			},
-		});
-
-		const token1 = await auth.api.signJWT({
-			body: {
-				payload: {
-					sub: "123",
-					exp: Math.floor(Date.now() / 1000) + 600,
-					iat: Math.floor(Date.now() / 1000),
-				},
-			},
-		});
-		expect(token1?.token).toBeDefined();
-
-		const jwks = await auth.api.getJwks();
-		expect(jwks.keys.length).toBe(1);
-	});
-
 	it("should return existing key on subsequent calls", async () => {
 		const { auth } = await getTestInstance({
 			plugins: [jwt()],
@@ -948,5 +925,56 @@ describe("jwt - edge cases and error handling", async () => {
 		expect(jwks1.keys.length).toBe(1);
 		expect(jwks2.keys.length).toBe(1);
 		expect(jwks3.keys.length).toBe(1);
+	});
+
+	it("should handle race condition when key is created between query and create", async () => {
+		const { auth } = await getTestInstance({
+			plugins: [jwt()],
+			logger: {
+				level: "error",
+			},
+		});
+
+		const jwks1 = await auth.api.getJwks();
+		const firstKeyId = jwks1.keys[0]?.kid;
+		expect(firstKeyId).toBeDefined();
+
+		const token1 = await auth.api.signJWT({
+			body: {
+				payload: {
+					sub: "123",
+					exp: Math.floor(Date.now() / 1000) + 600,
+					iat: Math.floor(Date.now() / 1000),
+				},
+			},
+		});
+		expect(token1?.token).toBeDefined();
+
+		const jwks2 = await auth.api.getJwks();
+		expect(jwks2.keys[0]?.kid).toBe(firstKeyId);
+		expect(jwks2.keys.length).toBe(1);
+	});
+
+	it("should query for existing key before attempting to create", async () => {
+		const { auth } = await getTestInstance({
+			plugins: [jwt()],
+			logger: {
+				level: "error",
+			},
+		});
+
+		const jwks1 = await auth.api.getJwks();
+		const firstKeyId = jwks1.keys[0]?.kid;
+		expect(firstKeyId).toBeDefined();
+
+		for (let i = 0; i < 5; i++) {
+			const jwks = await auth.api.getJwks();
+			expect(jwks.keys[0]?.kid).toBe(firstKeyId);
+			expect(jwks.keys.length).toBe(1);
+		}
+
+		const finalJwks = await auth.api.getJwks();
+		expect(finalJwks.keys.length).toBe(1);
+		expect(finalJwks.keys[0]?.kid).toBe(firstKeyId);
 	});
 });


packages/better-auth/src/plugins/jwt/adapter.ts
```diff
@@ -2,6 +2,42 @@ import type { BetterAuthOptions } from "@better-auth/core";
 import type { DBAdapter } from "@better-auth/core/db/adapter";
 import type { Jwk } from "./types";
 
+class JwkCreationMutex {
```
himself65, line 5: This does not seem correct. Imaging you have multiple server or serverless environment

okisdev: ohh yes, indeed this issue still exists and will be triggered in a multi-server environment. let me investigate further. thx for rising up.

luist18: To achieve this, you'll need a service that holds the lock for this data race. You can leverage row level or table level locks, or some database constraints as explained in the issue description

LightTab2:
We could also think about creating a key on the plugin init, if the `jwks` table is empty.

But unfortunately the table might not be created at that time and we cannot force it. It might not be a dead end though, idk.

okisdev:
> We could also think about creating a key on the plugin init, if the `jwks` table is empty.
> 
> But unfortunately the table might not be created at that time and we cannot force it. It might not be a dead end though, idk.


emm... but actually when the table may not yet exist, users need to manually perform migrations. when during each cold start, the serverless environment initializes, which leads to waste. we cannot even guarantee that it is created only once.

packages/better-auth/src/plugins/jwt/adapter.ts
okisdev: > To achieve this, you'll need a service that holds the lock for this data race. You can leverage row level or table level locks, or some database constraints as explained in the issue description

thanks for this suggestion, still hope to implement it without affecting or introducing new thing.

okisdev:
@himself65 can review now, please have a look whether see if can match the design pattern of better-auth. thx.

packages/better-auth/src/plugins/jwt/adapter.ts
```diff
@@ -32,5 +68,61 @@ export const getJwksAdapter = (adapter: DBAdapter<BetterAuthOptions>) => {
 
 			return jwk;
 		},
+		getOrCreateLatestKey: async (
+			createKeyFn: () => Promise<Omit<Jwk, "id">>,
+		): Promise<Jwk> => {
+			return jwkMutex.runExclusive(async () => {
```
luist18, line 74: Since this uses a local mutex what prevents two serverless requests coming in at the same time? The mutex will not work in that case

LightTab2:
> > We could also think about creating a key on the plugin init, if the `jwks` table is empty.
> > But unfortunately the table might not be created at that time and we cannot force it. It might not be a dead end though, idk.
> 
> emm... but actually when the table may not yet exist, users need to manually perform migrations. when during each cold start, the serverless environment initializes, which leads to waste. we cannot even guarantee that it is created only once.

Oh yeah, the thing I experienced is because test instances are running the migrations after the auth is initalized, which is a wrong order (I assume). I'll post my fix to this issue as a separate PR then.

packages/better-auth/src/plugins/jwt/adapter.ts
okisdev: > Since this uses a local mutex what prevents two serverless requests coming in at the same time? The mutex will not work in that case

yes after ur comment and i checked. this is absolutely not solve the issue at all. i am sorry about that. however, the best approach might be to use database unique constraints. i am not able to go further with this as adding the unique constraints will breaks the current experience. #5723 is a good trial on this. i will leave this be solved to better-auth team. pr closed now.

Issue #4830, opened 2025-09-23
Admin api.listUser wrong type return


TheDevick:
### Is this suited for github?

- [ ] Yes, this is suited for github

### To Reproduce

```ts
import { auth } from '@/lib/auth'
import { authClient } from '@/lib/auth/client'
import { headers } from 'next/headers'

async function fetchUsers(
  page: number,
  perPage: number,
): Promise<(typeof authClient.$Infer.Session.user)[]> {
  const result = await auth.api.listUsers({
    headers: await headers(),
    query: {
      offset: (page - 1) * perPage,
      limit: perPage,
    },
  })

  return result.users
}
```
```
Type 'never[] | UserWithRole[]' is not assignable to type '{ id: string; createdAt: Date; updatedAt: Date; email: string; emailVerified: boolean; name: string; image?: string | null | undefined; banned: boolean | null | undefined; role?: string | ... 1 more ... | undefined; banReason?: string | ... 1 more ... | undefined; banExpires?: Date | ... 1 more ... | undefined; }[]'.
  Type 'UserWithRole[]' is not assignable to type '{ id: string; createdAt: Date; updatedAt: Date; email: string; emailVerified: boolean; name: string; image?: string | null | undefined; banned: boolean | null | undefined; role?: string | ... 1 more ... | undefined; banReason?: string | ... 1 more ... | undefined; banExpires?: Date | ... 1 more ... | undefined; }[]'.
    Type 'UserWithRole' is not assignable to type '{ id: string; createdAt: Date; updatedAt: Date; email: string; emailVerified: boolean; name: string; image?: string | null | undefined; banned: boolean | null | undefined; role?: string | ... 1 more ... | undefined; banReason?: string | ... 1 more ... | undefined; banExpires?: Date | ... 1 more ... | undefined; }'.
      Property 'banned' is optional in type 'UserWithRole' but required in type '{ id: string; createdAt: Date; updatedAt: Date; email: string; emailVerified: boolean; name: string; image?: string | null | undefined; banned: boolean | null | undefined; role?: string | ... 1 more ... | undefined; banReason?: string | ... 1 more ... | undefined; banExpires?: Date | ... 1 more ... | undefined; }'.ts(2322)
No quick fixes available
```

## The problem

When listing the users, the return type have an `users` property, which is an array of users, which sets the property `banned` as optional.

The problem is: The return type from the `auth.api.listUsers()` is `UserWithRole`, which sets the property `banned` as optional.

[...]

PR #5701, opened 2025-10-31
fix: type mismatch for 'banned' on UserWithRole

GautamBytes:
**Resolves: #4830**

### Description

This PR fixes a TypeScript type mismatch between the `admin` plugin's `UserWithRole` type and the client-side inferred type `authClient.$Infer.Session.user`.

### The Problem

The `UserWithRole` interface in `packages/better-auth/src/plugins/admin/types.ts` incorrectly defined the `banned` property as optional:
`banned?: (boolean | null) | undefined;`

This conflicts with the client-side type inference, which correctly expects the field to always be present.

### The Fix

This is a pure type-definition fix. The database adapters (like the Drizzle adapter) use `db.select()` (which is `SELECT *`), guaranteeing the `banned` field will always be returned as `true`, `false`, or `null` - but **never** `undefined`.

This PR corrects the type definition in `types.ts` to `banned: boolean | null;` to match the actual data returned by the adapter, resolving the type error. No runtime changes are necessary.

---
## Summary by cubic
Make UserWithRole.banned a required boolean | null to match the client session user type and prevent TypeScript errors in the admin plugin. No runtime changes; this aligns with DB adapters that always return the field.

<sup>Written for commit b687fcd341a8d0669d6412f3e98392ca1bc97566. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/admin/types.ts
+++ b/packages/better-auth/src/plugins/admin/types.ts
@@ -4,7 +4,7 @@ import { type AdminSchema } from "./schema";
 
 export interface UserWithRole extends User {
 	role?: string | undefined;
-	banned?: (boolean | null) | undefined;
+	banned: boolean | null;
 	banReason?: (string | null) | undefined;
 	banExpires?: (Date | null) | undefined;
 }


himself65 (approved):
Thanks for the fixing

PR #5710, opened 2025-11-01
Update next-auth-migration-guide.mdx

kristianka:
The migration docs from Next-auth / Auth.js is missing a step: renaming a table. Without this, the following error shows in the console:

```bash
2025-11-01T01:41:38.007Z ERROR [Better Auth]: Model verification does not exist in the database. If you haven't generated the Prisma client, you need to run 'npx prisma generate'
# SERVER_ERROR:  [Error [BetterAuthError]: Model verification does not exist in the database. If you haven't generated the Prisma client, you need to run 'npx prisma generate'] {
  cause: undefined
}
```

---
## Summary by cubic
Added the missing migration step to rename the verification table from VerificationToken to Verification so Better Auth recognizes the model and avoids the “Model verification does not exist” error.

- **Migration**
  - Rename the verification schema/table from VerificationToken to Verification and match ORM casing (verification or Verification).

<sup>Written for commit 178e630bc04442eb8050487d9c4e7714f65e7cf8. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/guides/next-auth-migration-guide.mdx
+++ b/docs/content/docs/guides/next-auth-migration-guide.mdx
@@ -48,6 +48,12 @@ export const auth = betterAuth({
 
 Make sure to have `createdAt` and `updatedAt` fields on your session schema.
 
+
+#### Verification Schema
+
+- Rename the schema name from `VerificationToken` to `Verification`.
+
+
 #### Account Schema
 
 Map these fields in the account schema:
@@ -78,6 +84,9 @@ export const auth = betterAuth({
     },
 });
 ```
+
+
+
 
 **Note:** If you use ORM adapters, you can map these fields in your schema file.
 
--- a/docs/content/docs/guides/next-auth-migration-guide.mdx
+++ b/docs/content/docs/guides/next-auth-migration-guide.mdx
@@ -85,9 +85,6 @@ export const auth = betterAuth({
 });
 ```
 
-
-
-
 **Note:** If you use ORM adapters, you can map these fields in your schema file.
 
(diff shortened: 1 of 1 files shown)


docs/content/docs/guides/next-auth-migration-guide.mdx
```diff
@@ -79,6 +85,9 @@ export const auth = betterAuth({
 });
 ```
 
+
+
+
```
Bekacru:
```suggestion
```

docs/content/docs/guides/next-auth-migration-guide.mdx
```diff
@@ -48,6 +48,12 @@ export const auth = betterAuth({
 
 Make sure to have `createdAt` and `updatedAt` fields on your session schema.
 
+
+#### Verification Schema
+
+- Rename the schema name from `VerificationToken` to `Verification`.
```
Bekacru: I think it should be `verification` or at least mention that it might be `verification`, since we use lowercase for model names by default. But for specific orms like Prisma it shouldn't matter

Bekacru (changes requested):

docs/content/docs/guides/next-auth-migration-guide.mdx
kristianka: Updated, sorry for the late reply!

gustavovalverde:
Closing in favor of #6076 (merged). #6076 rewrote the Auth.js migration guide in 696 lines; this PR's content (renaming `VerificationToken` to `Verification`) is a strict subset of what landed.

PR #5720, opened 2025-11-01
feat(api-key): add create, update and delete hooks

### LINKED ISSUE #5694: Add hooks for API keys

API Keys support metadata. But there are no hooks in the config options. As such, metadata becomes more-or-less useless unless we define our own custom endpoints for API key management. 

Typical cases for metadata would include things like the `organizationId` but since we have no hooks available, we can never verify that metadata used upon creation of an API key. Upon creation the user may use any organizationId to create their API key.

On my system, I need to have access to this object before the key is created to potentially reject the request. This is true for all meta-information one could possibly store on an API key.

I'm proposing adding hooks for API keys. 

None

### Additional context

_No response_

Blackmamoth:
## Description
Adds lifecycle hooks (create, update, delete) for the API Key plugin.

Closes #5694 

## Changes
- Added before/after hooks for create, update, and delete operations
- Added tests for all hook scenarios
- Updated documentation with hook examples

---
## Summary by cubic
Adds before/after lifecycle hooks to the API Key plugin for create, update, and delete so you can validate, modify, or block these actions. Also updates docs, tests, and adds a new error code for hook failures.

- **New Features**
  - Hooks: before can return false to cancel, or { data } to modify; hooks receive parsed metadata/permissions and values are serialized automatically; after runs with the created/updated/deleted record.
  - Errors: adds BEFORE_HOOK_FAILED; returns BAD_REQUEST when a before hook cancels create, update, or delete.
  - Docs and tests: extensive examples and coverage for all hook paths.

<sup>Written for commit 4c02248a63ccf9ebca76adc25bf5f14a8fcd8da8. Summary will update automatically on new commits.</sup>


ping-maxwell:
Hey this is great, but I'm going to have to close this as not-planned as we don't want to introduce additional/custom hook configs if normal better-auth hooks can already achieve this.

Blackmamoth:
understood, thanks!

Issue #3875, opened 2025-08-08
Do not unconditionally inject `?error` into error redirect URL; allow custom error URL builder


danielo515:
## Summary
When Better Auth redirects to an error page, it unconditionally appends an `error` search parameter to the provided error URL. This causes conflicts when the application already includes its own search params (e.g. a `title` or `message` param) or when a different param name/format is desired.

As a result, the final URL ends up with unexpected params or key collisions (e.g. `...?title=Invalid+invite?error=Some+message`), which can override or conflict with the app’s intended messaging / UX.

## Steps to Reproduce
1. Configure Better Auth to redirect to a custom error page with an existing query param, for example:
   - Error URL: `/error?title=Invalid%20invite`
2. Trigger an auth error that causes Better Auth to redirect to the error URL.
3. Observe that Better Auth appends `?error=<message>`, even if the URL already contains search params.

## Actual Behavior
The library appends an `error` query parameter unconditionally to the error URL, leading to duplicate/conflicting search params.

## Expected Behavior
- Either check the URL’s existing search params and merge safely without overriding (and ideally without enforcing the specific `error` key name),
- Or provide a way to fully customize the error URL construction so apps can decide how to represent the error information.

## Proposed Solutions
1. Guard/merge logic:
   - If the provided error URL already contains a query string, merge the new params without overwriting existing keys.
   - Only add the `error` param if it does not already exist.
   - Consider allowing a custom key name so consumers aren’t forced to use `error`.

2. Extensibility hook (preferred):
   - Add a configuration option like `errorUrlBuilder: (error: ErrorLike, baseUrl: URL | string) => string`.
   - Better Auth would call this function with the error and the base error URL, letting the app return the final redirect URL.
   - This avoids library-level assumptions and enables full control (i18n, structured JSON in a param, fragments, etc.).

If a configuration already exists that covers this use case, documentation/examples would be appreciated.

Thanks for considering this!

danielo515:
This issue is still relevant. I would love to hear something from the team on what they think at least

ping-maxwell:
> This issue is still relevant. I would love to hear something from the team on what they think at least

Sounds good.

danielo515:
The summary is correct, and this is indeed still a problem.

danielo515:
This is an issue for extending and properly using the library. Can you
please do a proper implementation?

---
https://danielorodriguez.com

PR #5721, opened 2025-11-01
feat: support callback function in onAPIError's errorURL

### LINKED ISSUE #4566: Support callback in onAPIError's errorURL

Currently, `onAPIError` supports a string property called `errorURL`.

The path is returned when receiving an error in the authentication, and automatically appends some search parameters like `error` and `error_description`.

It would be useful to support both string option as well as a customizable `callback`:

```js
// errorURL: "/error", // <--- current usage
errorURL: ({ error, error_description }) => {
    if (error === "access_denied") 
        return "/denied";
    else 
        return "/error";
}
```

redirecting in another page/route based on the **error query param**, but it's not optimal

### Additional context

_No response_

GautamBytes:
Closes #4566

This PR enhances error handling by allowing onAPIError.errorURL to be a function, not just a static string. This enables developers to dynamically generate redirect URLs based on the specific error.

The new function signature is: ({ error, error_description }) => string | Promise<string>

All error-handling logic across OAuth, SSO, and state validation has been updated to check if errorURL is a function and await its result if it is.

This change is fully backward compatible with existing string-based configurations.

---
## Summary by cubic
Adds support for a callback in onAPIError.errorURL so apps can redirect to dynamic error pages per error. String configs still work; we append error and error_description to the chosen URL.

- **New Features**
  - errorURL can be a function: ({ error, error_description }) => string | Promise<string>.
  - Applied across OAuth, SSO, admin, proxy, and OIDC flows; awaits async URLs.
  - Per-request state.errorURL (string) takes priority over global config; otherwise falls back to baseURL/error.
  - Consistent query param building for error and error_description.
  - Telemetry now indicates if errorURL is a function or present.

- **Refactors**
  - Centralized error redirect handling via ctx.handleErrorRedirect, with optional overrideErrorURL and default path.

<sup>Written for commit 9a4f30160be68c1b38d401bddc1f05419693c0aa. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/jwt.mdx
+++ b/docs/content/docs/plugins/jwt.mdx
@@ -338,6 +338,12 @@ Table Name: `jwks`
       type: "Date", 
       description: "Timestamp of when the web key was created" 
     },
+    { 
+      name: "expiresAt", 
+      type: "Date", 
+      description: "Timestamp of when the web key expires",
+      isOptional: true
+    },
   ]}
   />
 
@@ -403,6 +409,24 @@ jwt({
   }
 })
 ```
+
+### Key Rotation
+
+You can enable key rotation by setting the `rotationInterval` option. This will automatically rotate the key pair at the specified interval.
+
+The default value is `undefined` (disabled).
+
+```ts title="auth.ts"
+jwt({
+  jwks: {
+    rotationInterval: 60 * 60 * 24 * 30, // 30 days
+    gracePeriod: 60 * 60 * 24 * 30 // 30 days
+  }
+})
+```
+
+- `rotationInterval`: The interval in seconds to rotate the key pair.
+- `gracePeriod`: The period in seconds to keep the old key pair valid after rotation. This is useful to allow clients to verify tokens signed by the old key pair. The default value is 30 days.
 
 ### Modify JWT payload
 
--- a/packages/better-auth/src/plugins/jwt/adapter.ts
+++ b/packages/better-auth/src/plugins/jwt/adapter.ts
@@ -19,19 +19,12 @@ export const getJwksAdapter = (
 			});
 		},
 		getLatestKey: async (ctx: GenericEndpointContext) => {
-			if (options?.adapter?.getLatestKey) {
-				return await options.adapter.getLatestKey(ctx);
+			if (options?.adapter?.getJwks) {
+				const keys = await options.adapter.getJwks(ctx);
+				return keys?.sort(
+					(a, b) => b.createdAt.getTime() - a.createdAt.getTime(),
+				)[0];
 			}
-			const key = await adapter.findMany<Jwk>({
-				model: "jwks",
-				sortBy: {
-					field: "createdAt",
-					direction: "desc",
-				},
-				limit: 1,
-			});
-
-			return key[0];
 		},
 		createJwk: async (ctx: GenericEndpointContext, webKey: Omit<Jwk, "id">) => {
 			if (options?.adapter?.createJwk) {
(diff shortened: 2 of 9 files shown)


GautamBytes:
@ping-maxwell @Bekacru its ready for review!!

ping-maxwell:
This is great!

ping-maxwell:
Though I do wonder if it makes sense for us to introduce a `ctx.context.redirectToErrorPage` or some function to handle a lot of the logic of this, just to make it easier to redirect to the error page - since I do see a lot of duplicated logic, also in the future when people implement error page redirects it's possible they overlook that it could be a callback function or some other issue.

cc @Bekacru wdyt?

Bekacru:
Good idea. @GautamBytes interested on adding this?

GautamBytes:
> Good idea. @GautamBytes interested on adding this?

I'm happy to tackle that. If this PR is good to go, I can open a new, separate PR for that refactor right after this one gets merged.

ping-maxwell:
Why not apply it on this PR?

GautamBytes:
> Why not apply it on this PR?

Ah, okay, no problem! I wasn't sure if you'd prefer to keep the PRs separate, but I'm happy to add the refactor here.

GautamBytes:
@ping-maxwell ready for review!!

ping-maxwell:
Hmm, sorry to be such a bother, but instead of the redirect helper function, I meant for it to be built right into the `context`.

Also would help with things in the future like [this feature](https://github.com/better-auth/better-auth/issues/3875).

ping-maxwell (approved):

ping-maxwell:
Don't worry I've moved it to ctx!

packages/better-auth/src/context/create-context.ts
```diff
@@ -216,6 +216,37 @@ export async function createAuthContext(
 				: isTest()
 					? true
 					: false,
+		handleErrorRedirect: async (
```
himself65, line 214: I think we can handle this in to auth endpoint internally instead of adding more API in the context

himself65 (changes requested):

packages/better-auth/src/context/create-context.ts
ping-maxwell: We already do this, but it's too much duplicated code that breaks the moment something changes regarding error url, so I do think a helper is better.
Refer here: https://github.com/better-auth/better-auth/pull/5721#issuecomment-3487388530

ping-maxwell: In the future, new features like [this](https://github.com/better-auth/better-auth/issues/3875) can be done easily with this.

GautamBytes:
> Don't worry I've moved it to ctx!

Apologies you had to step in , I'll be sure to keep that pattern in mind for future contributions 🙌

ping-maxwell:
> Apologies you had to step in , 

Don't stress! Wanted to move this forward as we want this PR for 1.4 👀

> I'll be sure to keep that pattern in mind for future contributions

Using helpers is actually better, we want to avoid using ctx or introducing config options, just in this case I think it makes a lot of sense.

ajaykarthikr:
Hey @ping-maxwell , I see this PR has been pending for few weeks. I would be happy to contribute if there are bandwidths constraints.

asionesjia:
Hey, @ping-maxwell . Is there any latest development on this PR? I've been following it closely. My own version uses helper functions to solve the hard-coded ?error issue, but I think this PR is a better solution.

PR #5725, opened 2025-11-02
feat(stripe): allow any scheme

hyoban:
This allows user to return from the stripe to the Electron app.

---
## Summary by cubic
Support custom URL schemes in Stripe redirects so users can return to the Electron app after checkout. getUrl now treats any scheme (e.g., myapp://) as absolute and no longer prefixes baseURL when a scheme is present.

<sup>Written for commit 4551df4ff0a000fbc68ace4dbcfabb681bd3ebd6. Summary will update automatically on new commits.</sup>


himself65 (approved):

himself65:
LGTM

PR #5736, opened 2025-11-03
fix(passkey): ensure addPasskey returns passkey data instead of undefined

mburumaxwell:
## Description

Fixes an issue where `client.passkey.addPasskey()` was returning `undefined` instead of the passkey data when registration succeeded.

## Changes

- Added missing `return verified;` statement in the `registerPasskey` function's success path
- The function now correctly returns the passkey data from the server response
- No breaking changes to the API contract

## Problem

The `addPasskey` client method was missing a return statement after successful passkey registration. This caused the function to return `undefined` instead of the expected passkey data, making it impossible for consumers to access information about the newly created passkey.

---
## Summary by cubic
Ensures client.passkey.addPasskey returns the passkey data on successful registration instead of undefined.

- **Bug Fixes**
  - Added missing return after list refresh so the verified passkey is returned.
  - Updated verify-registration fetch to use Passkey as the response type.
  - No breaking changes to the API.

<sup>Written for commit 8ccfc36cca05ee3e61183ed0cb64863c65b6f91b. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/passkey/src/client.ts
+++ b/packages/passkey/src/client.ts
@@ -128,9 +128,7 @@ export const getPasskeyActions = (
 				optionsJSON: options.data,
 				useAutoRegister: opts?.useAutoRegister,
 			});
-			const verified = await $fetch<{
-				passkey: Passkey;
-			}>("/passkey/verify-registration", {
+			const verified = await $fetch<Passkey>("/passkey/verify-registration", {
 				...opts?.fetchOptions,
 				...fetchOpts,
 				body: {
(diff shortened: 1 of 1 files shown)


mburumaxwell:
Rebased on #5769 by @bytaesu 
Ready for review/approval

bytaesu (approved):
Thanks, this looks good!

demo/nextjs/app/dashboard/user-card.tsx
```diff
@@ -836,12 +836,12 @@ function ListPasskeys() {
 			return;
 		}
 		setIsLoading(true);
-		const res = await client.passkey.addPasskey({
+		const { error } = await client.passkey.addPasskey({
```
Bekacru: if the auth client `throw` is set to `true`, `error` will be undefined. Let's revert this change

mburumaxwell: Apologies, I was under the impression that when it is set to throw, the method would not return but since I have not used that before, I wasn't sure.
I have reverted that section.

himself65 (approved):

PR #5739, opened 2025-11-03
Added Two-Factor option to skip TOTP code creation

jackh-sh:
I added an option to the API to skip TOTP creation so that you can enable two factor auth for the OTP method.

---
## Summary by cubic
Add a new twoFactorMethod option to enable 2FA using OTP without generating TOTP codes or QR data. Default remains TOTP, so existing integrations keep working.

- **New Features**
  - API: twoFactor.enable accepts twoFactorMethod: "otp" | "totp" (default "totp").
  - "otp" skips TOTP setup and verification-on-enable, marks the user 2FA-enabled, and returns { twoFactor: true }.
  - "totp" continues to return totpURI and backupCodes as before.
  - Docs updated to show the new parameter.
  - Test added for the OTP path to ensure no TOTP artifacts are created.

<sup>Written for commit a5601c39d583b366bc3cd90cd1e64c995fc2d160. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/2fa.mdx
+++ b/docs/content/docs/plugins/2fa.mdx
@@ -92,7 +92,11 @@ type enableTwoFactor = {
     /**
      * The user's password
      */
-    password: string = "secure-password"
+    password: string = "secure-password",
+    /**
+     * Whether to enable the OTP method or TOTP method (uses the TOTP method by default)
+     */
+    twoFactorMethod?: "otp" | "totp" = "totp",
     /**
      * An optional custom issuer for the TOTP URI. Defaults to app-name defined in your auth config.
      */
--- a/packages/better-auth/src/plugins/two-factor/index.ts
+++ b/packages/better-auth/src/plugins/two-factor/index.ts
@@ -69,6 +69,7 @@ export const twoFactor = (options?: TwoFactorOptions | undefined) => {
 						password: z.string().meta({
 							description: "User password",
 						}),
+						twoFactorMethod: z.enum(["otp", "totp"]).default("totp"), // Default means it doesn't break any exsisting implementations
 						issuer: z
 							.string()
 							.meta({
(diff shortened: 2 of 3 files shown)


hades200082:
Some questions for you to consider:

1. Can I have both OTP and TOTP enabled in my app? e.g. I'd like OTP enabled by default for all users, with users able to configure TOTP if they choose.
2. What happens if the user later wants to enable TOTP?
3. Can OTP be used as a backup if the user doesn't have their TOTP device with them?

kvanbere:
Hey @jackh-sh @hades200082, sorry to butt in, but I have been integrating around this.

I've found that it's [very difficult/basically impossible](https://github.com/better-auth/better-auth/discussions/8411) to detect whether the user is using OTP or TOTP during login and redirect them to the right place.

The key use cases are:
1. If either method is available, but not the other, we might want to redirect to the proper place such as `/mfa/otp` or `/mfa/totp` so the user sees the right UX
2. If both are available then we might want to give the user a link to use `OTP` instead of `TOTP` if they lost their code generator for some reason

**Suggestion 1**

If we could please return `twoFactorOtpRedirect` and `twoFactorTotpRedirect` respectively if each method is available (in addition to `twoFactorRedirect`), then we can decide in the application which method to use.

**Suggestion 2**

Alternatively, perhaps if we could configure `otpRedirectPage` and `totpRedirectPage` in the plugin configuration, then the plugin could just send back the redirect and the user doesn't need to try and figure it out. Long term, this might be the more scalable solution for example if more two factor methods were added (you never know).

Jordan-Hall:
I'm having this nightmare too. Ideally it would support an array of options the users enabled and then say whats enabled and whats preffered so we can redirect to preferred and list switch to for the other options

kvanbere:
I found a total blocker in the current `twoFactor` (it skips `TOTP` verification flow after adding `OTP` to any user) so I opened https://github.com/better-auth/better-auth/issues/8627 which is semi-related to this PR. This PR probably surfaces the same bug.

gustavovalverde:
Superseded by #9057, which ships in v1.7.0-beta.1.

PR #5742, opened 2025-11-03
feat: add request state

himself65:
## Summary by cubic
Adds request-scoped state with AsyncLocalStorage to safely share schema-bound data across async boundaries, and runs auth endpoints within it or reuses an existing one. Enables validated OAuth additionalData to be passed through the flow and exposed to hooks via request state (not persisted).

- **New Features**
  - RequestState using WeakMap<StandardSchemaV1, any>.
  - runWithRequestState(), getCurrentRequestState() (throws), hasRequestState(), getRequestStateAsyncLocalStorage().
  - defineRequestState(schema) for typed get/set.
  - Auth endpoints ensure a per-call request state and reuse when available.
  - OAuth middleware exposes getOauthState; set is internal to the flow.
  - OAuth: advanced.oauthConfig.additionalData { enabled?, schema?: StandardSchemaV1 } to validate and store data in request state.
  - signInSocial, generic OAuth, and linkSocialAccount accept additionalData; forwarded and available in hooks (not stored in DB).

<sup>Written for commit ae7fccda44f2e8379f5c3e11cbd5779e5a39e783. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/call.test.ts
+++ b/packages/better-auth/src/call.test.ts
@@ -219,6 +219,7 @@ describe("call", async () => {
 			],
 		},
 	} satisfies BetterAuthPlugin;
+	let latestOauthStore: Record<string, any> | undefined = {};
 	const options = {
 		baseURL: "http://localhost:3000",
 		plugins: [testPlugin, testPlugin2, bearer()],
@@ -264,7 +265,7 @@ describe("call", async () => {
 				}
 				if (ctx.path === "/callback/:id" && ctx.params.id === "google") {
 					const store = await getOauthState();
-					console.log("store", store);
+					latestOauthStore = store;
 				}
 			}),
 		},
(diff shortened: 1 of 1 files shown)


docs/content/docs/concepts/oauth.mdx
```diff
@@ -132,6 +132,157 @@ const requestAdditionalScopes = async () => {
...
+await auth.api.signInSocial({
+  body: {
+    provider: "google",
+    additionalData: {
+      referralCode: "ABC123",
+      source: "admin-panel",
+    },
+  },
+});
+```
```
Bekacru: we should use the <Endpoint.. component instead to show examples for both client and server

packages/better-auth/src/plugins/generic-oauth/index.ts
```diff
@@ -499,7 +503,11 @@ export const genericOAuth = (options: GenericOAuthOptions) => {
 							? authorizationUrlParams(ctx)
 							: authorizationUrlParams;
 
-					const { state, codeVerifier } = await generateState(ctx);
+					const { state, codeVerifier } = await generateState(
+						ctx,
+						undefined,
+						ctx.body.additionalData,
```
Bekacru: why don't we do schema validatino for additinoal data here?

PR #5776, opened 2025-11-05
chore: re-export better-call types

jslno:
`The inferred type of 'auth' cannot be named without a reference to '.pnpm/better-call@1.0.24/node_modules/better-call'. This is likely not portable. A type annotation is necessary.`

---
## Summary by cubic
Re-exported better-call types and StandardSchemaV1 from better-auth to fix non-portable type references. TypeScript can now name the auth type without referencing the better-call package path, avoiding pnpm node_modules leaks.

<sup>Written for commit 326765cdd32ba4dc8000f7a3b94055cbf5fb3fd9. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/index.ts
+++ b/packages/better-auth/src/index.ts
@@ -1,14 +1,14 @@
 //#region Re-exports necessaries from core module
 
-export * from "@better-auth/core";
-//#endregion
-export type * from "better-call";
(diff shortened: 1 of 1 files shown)


himself65 (approved):
Thanks

PR #5793, opened 2025-11-06
feat: awaitable custom transforms

dvanmali:
Allows `customTransformInput` and `customTransformOutput` to be await-able. Required for specific custom computed fields.

---
## Summary by cubic
Make customTransformInput and customTransformOutput awaitable to support async computed fields while keeping sync behavior intact.

- **New Features**
  - Adapter factory now awaits customTransformInput/output.
  - Added Awaitable type and updated adapter types to return Awaitable<any>.
  - Enables async computed fields (e.g., external lookups) without breaking existing sync transforms.

<sup>Written for commit 17134792e504908ede0ae14155f852b057c7c739. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/src/db/adapter/index.ts
+++ b/packages/core/src/db/adapter/index.ts
@@ -1,4 +1,4 @@
-import type { BetterAuthOptions } from "../../types";
+import type { Awaitable, BetterAuthOptions } from "../../types";
 import type { BetterAuthDBSchema, DBFieldAttribute } from "../type";
 
 export type DBAdapterDebugLogOption =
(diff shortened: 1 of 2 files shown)


packages/core/src/db/adapter/index.ts
```diff
@@ -236,7 +236,7 @@ export interface DBAdapterFactoryConfig<
 				 * The options of the user's Better-Auth instance.
 				 */
 				options: Options;
-		  }) => any)
+		  }) => any | Promise<any>)
```
himself65: Awaitable

dvanmali: Awaitable doesn't exist within core package

dvanmali:
Alternative found via `fieldAttribute` `transform`.

PR #5806, opened 2025-11-06
feat: add support for custom response status codes

jonathansamines:
**What is changing?**
Updates `better-auth` to expose the same functionality as in https://github.com/Bekacru/better-call/pull/69 (custom response status codes via `ctx.setStatus()`).

---
## Summary by cubic
Adds custom response status code support in better-auth via ctx.setStatus. Endpoints can now return a Response with the chosen status or include the status alongside the payload.

- **New Features**
  - Added returnStatus option to toAuthEndpoints to include status with response and headers.
  - asResponse now respects custom status when building a Response.
  - Propagates status from thrown Response-like errors via statusCode and headers.
  - Tests cover both asResponse and returnStatus flows with a 201 status.

- **Migration**
  - No changes required. Use c.setStatus(...) and opt in with { asResponse: true } or { returnStatus: true }.
  - Defaults remain unchanged; session route sets returnStatus to false.

<sup>Written for commit f5a925d8246646568497b562c0ca1e8fc592ce03. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/to-auth-endpoints.test.ts
+++ b/packages/better-auth/src/api/to-auth-endpoints.test.ts
@@ -9,40 +9,6 @@ import { init } from "../init";
 import { getTestInstance } from "../test-utils/test-instance";
 import { toAuthEndpoints } from "./to-auth-endpoints";
 
-describe("custom response code", () => {
-	const endpoints = {
-		responseWithStatus: createAuthEndpoint(
-			"/response-with-status",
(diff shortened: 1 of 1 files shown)


himself65 (commented):
I think we miss the strict type here, maybe we should patch the better-call first?

Bekacru:
@himself65 what strict type?

PR #5808, opened 2025-11-06
feat: allow passing overrides to getMigration

ping-maxwell:
Example
```ts
import { getMigrations } from "better-auth/db";
import type { BetterAuthOptions } from "@better-auth/core";
import type { FieldTypeOverride } from "better-auth/db/get-migration";

const config: BetterAuthOptions = {
  database: yourPostgresDatabase, // Your PostgreSQL database connection
  emailAndPassword: {
    enabled: true,
  },
  // ... other config
};

const fieldTypeOverride: FieldTypeOverride = (
  field,
  fieldName,
  tableName,
  dbType,
  config,
  defaultType,
) => {
  // Override all id fields to UUID for PostgreSQL
  if (dbType === "postgres" && fieldName === "id") {
    return { type: "uuid" };
  }

  // Optionally, also override foreign keys that reference id fields
  if (dbType === "postgres" && field.references?.field === "id") {
    return { type: "uuid" };
  }

  return undefined; // Use defaults for everything else
};

const { runMigrations, compileMigrations } = await getMigrations(
  config,
  fieldTypeOverride,
);

// To see the SQL that will be generated:
const sql = await compileMigrations();
console.log(sql);

// To actually run the migrations:
await runMigrations();
```

---
## Summary by cubic
Adds an override callback to getMigrations so you can customize column SQL types and attributes per field/table/database without changing defaults. This makes migrations more flexible across Postgres, MySQL, SQLite, and MSSQL.

- **New Features**
  - getMigrations now accepts an optional fieldTypeOverride callback.
  - Callback can override type (string or RawBuilder), required, and unique.
  - Receives field, fieldName, tableName, dbType, config, and defaultType for context.
  - Overrides apply to both new tables and ALTER TABLE column additions.
  - Primary key id columns now respect overrides when creating tables.
  - Extensive tests cover ids, foreign keys, booleans, json, arrays, and constraints.

- **Migration**
  - No action needed; behavior is unchanged if no override is provided.
  - If using overrides, import FieldTypeOverride from better-auth/db/get-migration.

<sup>Written for commit eb43727cce00fc11c924bae579d7b32cc3bce9cf. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/components/docs/layout/toc-thumb.tsx
+++ b/docs/components/docs/layout/toc-thumb.tsx
@@ -66,6 +66,11 @@ export function TocThumb({
 	useOnChange(active, () => {
 		if (!containerRef.current || !thumbRef.current) return;
 
+		// Skip updates during anchor scroll animation to prevent flicker
+		if (document.documentElement.hasAttribute("data-anchor-scrolling")) {
+			return;
+		}
+
 		update(thumbRef.current, calc(containerRef.current, active));
 	});
 
--- a/docs/components/docs/layout/toc.tsx
+++ b/docs/components/docs/layout/toc.tsx
@@ -180,6 +180,7 @@ export function TOCItems({ items }: { items: TOCItemType[] }) {
 					<TocThumb
 						containerRef={containerRef}
 						className="mt-(--fd-top) h-(--fd-height) bg-fd-primary transition-all"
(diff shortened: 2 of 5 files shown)


Bekacru:
We should document this API

PR #5809, opened 2025-11-06
feat: add support for uuids

Bekacru:
## Summary by cubic
Add first-class UUID support for IDs and standardize ID generation via advanced.database.generateId. Deprecates useNumberId and updates migrations/adapter logic to handle "uuid" and "serial" across Postgres, MySQL, and MSSQL.

- **New Features**
  - generateId now accepts "uuid" and "serial".
  - Migrations create UUID columns and defaults:
    - Postgres: uuid type with default gen_random_uuid().
    - MySQL/MSSQL: uuid type with default uuid().
  - Validation updated: requesting numeric IDs ("serial"/useNumberId) throws if the adapter doesn’t support them.
  - Deprecation warning added for useNumberId; prefer generateId: "serial".

- **Migration**
  - Replace useNumberId with generateId: "serial".
  - To enable UUIDs, set generateId: "uuid".
  - Postgres requires the pgcrypto extension for gen_random_uuid().

<sup>Written for commit cc50d5f28f7c85d63ea28f8b167bfcf3429bc178. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/db/get-migration-schema.test.ts
+++ b/packages/better-auth/src/db/get-migration-schema.test.ts
@@ -191,7 +191,7 @@ describe.runIf(isPostgresAvailable)(
 		});
 
 		beforeAll(async () => {
-			await schemaPool.query(`DROP SCHEMA IF EXISTS ${schema} CASCADE`);
+			await pool.query(`DROP SCHEMA IF EXISTS ${schema} CASCADE`);
 			await schemaPool.query(`CREATE SCHEMA ${schema}`);
 		});
 


luist18:
Small correction from what's written in the description

> Postgres requires the pgcrypto extension for gen_random_uuid().

Since pg13 this function is now native to Postgres https://www.postgresql.org/docs/current/functions-uuid.html. I wonder how many people still run < pg13 😄

grant0417:
Also support for uuidv7 would be nice, a lot of benefits from adopting it.

songkeys:
> Also support for uuidv7 would be nice, a lot of benefits from adopting it.

we can definitely generate the uuid sql migration schema for pg and then manually change the default field to uuidv7, right?

PR #5810, opened 2025-11-06
fix(email-verification): set emailVerified to true upon email verification

Ipriyankrajai:
# Fix: Remove redundant email verification when changing email

## Problem

When a user with a verified email changes their email address, the system was sending **two verification emails**:

1. **First email** (from `changeEmail` endpoint): Sent via `sendChangeEmailVerification` callback to the **new email address** to confirm the email change request
2. **Second email** (from `verifyEmail` endpoint): After clicking the first verification link, a second verification email was automatically sent to the **new email address** again

This created an unnecessary extra step in the email change flow - the user had to verify their new email twice.

## Solution

When a user verifies an email change request (by clicking the link sent to their new email), we now:

- ✅ Update the email address to the new email
- ✅ Set `emailVerified: true` immediately
- ❌ **Remove** the redundant second verification email

## Why This Makes Sense

The verification flow already ensures the user owns the new email:

1. The user must be **logged in** (session required) to change their email
2. The verification email is sent to the **new email address** via `sendChangeEmailVerification` (lines 793-801 in `update-user.ts`)
3. Clicking the verification link sent to the new email confirms the user has access to that email address

Since the user has already verified their ownership of the new email by clicking the verification link sent to it, there's no need for an additional verification email. The act of verifying the change request **is** sufficient verification of the new email. This matches the pattern used for unverified email changes (lines 744-767 in `update-user.ts`), where a single verification email to the new address is sufficient.

## Changes

- Updated `verifyEmail` endpoint to set `emailVerified: true` when processing email changes (when `updateTo` is present in the token)
- Removed the redundant token creation and second verification email sending logic

## Related Code

- `packages/better-auth/src/api/routes/email-verification.ts` - Removed redundant verification email logic
- `packages/better-auth/src/api/routes/update-user.ts` - Already sends the initial change email verification (lines 751-760)

---
## Summary by cubic
Stop sending a second verification email after users confirm an email change. When the user clicks the change-email verification link, we update their email and mark it verified immediately.

- **Bug Fixes**
  - Set emailVerified: true in verifyEmail when processing email changes.
  - Removed token generation and follow-up verification email to the new address.
  - Updated session cookie to reflect the new email with emailVerified: true.

<sup>Written for commit ddfe08660efed9767700bf87df415c2da6e1a1ed. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/email-verification.test.ts
+++ b/packages/better-auth/src/api/routes/email-verification.test.ts
@@ -357,7 +357,7 @@ describe("Email Verification Secondary Storage", async () => {
 				},
 			});
 			expect(session.data?.user.email).toBe("new@email.com");
-			expect(session.data?.user.emailVerified).toBe(false);
+			expect(session.data?.user.emailVerified).toBe(true);
 		});
 	});
 
--- a/packages/better-auth/src/api/routes/email-verification.ts
+++ b/packages/better-auth/src/api/routes/email-verification.ts
@@ -354,30 +354,12 @@ export const verifyEmail = createAuthEndpoint(
 				},
 			);
 
-			const newToken = await createEmailVerificationToken(
-				ctx.context.secret,
-				parsed.updateTo,
-			);
-
-			//send verification email to the new email
-			const updateCallbackURL = ctx.query.callbackURL
-				? encodeURIComponent(ctx.query.callbackURL)
-				: encodeURIComponent("/");
-			await ctx.context.options.emailVerification?.sendVerificationEmail?.(
-				{
-					user: updatedUser,
-					url: `${ctx.context.baseURL}/verify-email?token=${newToken}&callbackURL=${updateCallbackURL}`,
-					token: newToken,
-				},
-				ctx.request,
-			);
-
 			await setSessionCookie(ctx, {
 				session: session.session,
 				user: {
 					...session.user,
 					email: parsed.updateTo,
-					emailVerified: false,
+					emailVerified: true,
 				},
 			});
 
--- a/packages/better-auth/src/api/routes/email-verification.ts
+++ b/packages/better-auth/src/api/routes/email-verification.ts
@@ -350,7 +350,7 @@ export const verifyEmail = createAuthEndpoint(
 				parsed.email,
 				{
 					email: parsed.updateTo,
-					emailVerified: false,
+					emailVerified: true,
 				},
 			);
 


Ipriyankrajai:
Hey @Bekacru, 

I was migrating from Auth.js to Better Auth and noticed that when users change their email, they receive two verification emails to the new address - one from `changeEmail` and another after clicking the verification link. This seemed redundant since clicking the link already confirms access to the new email.

I've created a PR to remove the redundant second verification email and mark the new email as verified immediately after the user clicks the verification link.

Thanks for building Better Auth! 🚀

Ipriyankrajai:
@bytaesu Can you review it?

bytaesu:
Hello @Ipriyankrajai,

Sure! I'll check it out soon after I finish my current task 😁 Noted!

tomdenboon:
@Ipriyankrajai this is not correct, the first email should go to the email address that is already associated with the account. This is required for extra security, and to make sure that the user that is actually the owner of the email approves the change. Otherwise the security of change email is very lackluster. https://www.better-auth.com/docs/concepts/users-accounts#change-email

Ipriyankrajai:
@tomdenboon The current flow sends the verification email to the new address the user requests.
Note: The user must first verify their old email. Once the old email is verified, they can update it, and the verification link is sent to the new email.

bytaesu:
Hello @Ipriyankrajai,

The first and second emails serve different verification purposes, so this change doesn’t seem appropriate.

Ipriyankrajai:
Hey @bytaesu,
I understand your point. What I am trying to clarify is this:
The first email is sent to the new address so that we can verify user has access to new email address.
The second email is sent to verify new email address.
My question is whether we can combine these steps. Since the first email is already sent to the new address, can we verify the new email at the same time the user clicks the link? That way both actions happen through a single verification flow.

tomdenboon:
> Hey @bytaesu, I understand your point. What I am trying to clarify is this: The first email is sent to the new address so that we can verify user has access to new email address. The second email is sent to verify new email address. My question is whether we can combine these steps. Since the first email is already sent to the new address, can we verify the new email at the same time the user clicks the link? That way both actions happen through a single verification flow.

You'll have to update your code to send the first email to the old address. This way malicious users that acquire a session from the user can't hijack the account. If the user is unverified, this first email step is skipped.

@bytaesu since you are already in the context, could you check if my [PR](https://github.com/better-auth/better-auth/pull/5980) makes any sense?

ixartz:
@bytaesu We can replace the first email by entering the current password. So instead of send approval link to the current email using user.changeEmail.sendChangeEmailVerification, it would be great if we can replace by password. Like changing password, the user needs to enter his current password.

ixartz:
@bytaesu Especially, if the user remember the password but lost access to his old email. He wouldn't be able to change the address email.

gustavovalverde:
Closing in favor of #6088 (merged) and #6106 (merged). #6088 introduced the current `requestType` change-email architecture and explicitly sets `emailVerified: true` when processing `change-email-verification` tokens; #6106 then patched the session requirement bug. The `emailVerified` assertion is already covered in `email-verification.test.ts:581` on main.

PR #5487, opened 2025-10-22
Add cancel_at field to subscription table

GautamBytes:
Fixes #5388 #5535

## What this PR does

This PR resolves an issue where the `customer.subscription.updated` webhook handler was not saving the `cancel_at` timestamp, causing the local database to be out of sync with Stripe.

This is particularly important for subscriptions canceled via the billing portal that have a specific cancellation time (`cancel_at`) rather than just at the period end (`cancel_at_period_end`).

## Changes

- **Schema:** Added a `cancelAt` (type: `date`, optional) field to the `subscription` schema in `packages/stripe/src/schema.ts`.
- **Types:** Updated the `Subscription` interface in `packages/stripe/src/types.ts` with the new `cancelAt?: Date` property.
- **Logic (Webhook):** Modified `onSubscriptionUpdated` in `packages/stripe/src/hooks.ts` to read `subscriptionUpdated.cancel_at` and save it to the database.
- **Logic (Restore):** Updated `restoreSubscription` in `packages/stripe/src/index.ts` to set `cancelAt: null` for data integrity when a subscription is restored.
- **Docs:** Added the new `cancelAt` field to the `subscription` schema documentation in `docs/content/docs/plugins/stripe.mdx`.
- **Tests:**
    - Added a new test to `packages/stripe/src/stripe.test.ts` to simulate a `customer.subscription.updated` event with `cancel_at` and assert it's saved correctly.
    - Updated an existing test snapshot to include the new `cancelAt` field in its default state.

    

---
## Summary by cubic
Add cancelAt to subscriptions and persist Stripe’s cancel_at in the webhook so scheduled cancellations are saved correctly. Prevents database drift when cancellations are set via the billing portal.

- **Bug Fixes**
  - Schema/types: added cancelAt?: Date to subscription.
  - Webhook: read cancel_at and save as Date.
  - Restore: set cancelAt to null on subscription restore.
  - Docs: documented cancelAt in the subscription schema.
  - Tests: added coverage for customer.subscription.updated with cancel_at; updated snapshot.

<sup>Written for commit 1b71302. Summary will update automatically on new commits.</sup>

PR #5818, opened 2025-11-07
fix(stripe): handle billing portal cancellation with cancel_at field

bytaesu:
Source: https://docs.stripe.com/billing/subscriptions/cancel

Closes https://github.com/better-auth/better-auth/issues/5780

---
## Summary by cubic
Handle Stripe Billing Portal cancellations that use the cancel_at field. We now treat either cancel_at or cancel_at_period_end as scheduled cancel, persist cancelAtPeriodEnd, and fire onSubscriptionCancel.

- **Bug Fixes**
  - In onSubscriptionUpdated, set cancelAtPeriodEnd if cancel_at_period_end is true or cancel_at is present.
  - Trigger onSubscriptionCancel when an active subscription is newly marked as canceled via either field.
  - Added tests for cancel_at, cancel_at_period_end, and both fields to prevent regressions.

<sup>Written for commit f42f5a01e46d89890709cb4eff7aef301ec2c35c. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/stripe/src/stripe.test.ts
+++ b/packages/stripe/src/stripe.test.ts
@@ -2176,4 +2176,413 @@ describe("stripe", async () => {
 			expect(data).toEqual({ success: true });
 		});
 	});
+
+	describe("Billing Portal Cancellation (cancel_at field)", () => {
+		it("should detect cancellation via cancel_at field from Billing Portal", async () => {
+			const { id: testReferenceId } = await ctx.adapter.create({
+				model: "user",
+				data: {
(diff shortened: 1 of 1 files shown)


bytaesu:
> [!NOTE] 
> Applying "PR: https://github.com/better-auth/better-auth/pull/5487" seems more appropriate. However, since it adds a new field to the schema, it introduces a minor breaking change. While this PR fixes the issue, **_we should consider whether to adopt the other PR_**.

bytaesu:
-> opening new pr

PR #5825, opened 2025-11-07
fix(anonymous): prevent Convex cleanup from deleting fresh sessions

### LINKED ISSUE #5824: Anonymous not working with Convex

## Summary
The Better Auth anonymous plugin has a critical bug where it deletes newly created anonymous users immediately after creation when used with the Convex adapter. This makes anonymous authentication completely non-functional.

## Expected Behavior
1. "Login Anonymously"
2. Better Auth creates a new anonymous user with email like temp-xyz@localhost:3000
3. Session is created and stored
4. User remains authenticated and can use the app

## Actual Behavior
1. "Login Anonymously"
2. Better Auth creates a new anonymous user ✅
3. Better Auth creates a session ✅
4. Better Auth immediately deletes the user ❌
5. Session lookup fails because user was deleted ❌
6. User appears unauthenticated ❌

## Root Cause Analysis
The anonymous plugin has built-in cleanup logic that:
1. Checks if the incoming request has an existing anonymous JWT cookie
2. If yes, deletes the old anonymous user to prevent accumulation
3. Creates a new anonymous user

**The bug**: This cleanup logic runs **DURING** the sign-in request, not before. The sequence is:

```bash
1. Create new user (id: m5767mek...)
2. Create new session (token: hI4IfkTS...)
3. Generate JWT and prepare to send it
4. Detect "old" JWT in request cookies (from milliseconds ago!)
5. Delete the user that was JUST created
6. Send response with JWT for deleted user
7. Browser receives JWT but user is gone
```

1.3.27

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:34 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T8103",
    "release": "24.6.0",
    "cpuCount": 8,
    "cpuModel": "Apple M1",
    "totalMemory": "16.00 GB",
    "freeMemory": "0.82 GB"
  },
  "node": {
    "version": "v22.19.0",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.1"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.0.1"
    },
    {
      "name": "react",
      "version": "19.2.0"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "1.3.27",
    "config": null
  }
}
```

Backend

```typescript
import { createClient, type GenericCtx } from "@convex-dev/better-auth";
import { convex } from "@convex-dev/better-auth/plugins";
import { anonymous } from "better-auth/plugins";
import { v } from "convex/values";
import { components } from "./_generated/api";
import type { DataModel } from "./_generated/dataModel";
import { query } from "./_generated/server";

const siteUrl = process.env.SITE_URL
  ? process.env.SITE_URL
  : "http://localhost:3000";

// The component client has methods needed for integrating Convex with Better Auth,
// as well as helper methods for general use.
export const authComponent = createClient<DataModel>(components.betterAuth);

export const createAuth = (
  ctx: GenericCtx<DataModel>,
  { optionsOnly } = { optionsOnly: false }
) => {
  return betterAuth({
    // disable logging when createAuth is called just to generate options.
    // this is not required, but there's a lot of noise in logs without it.
    logger: {
      disabled: optionsOnly,
    },
    baseURL: siteUrl,
    database: authComponent.adapter(ctx),
    // Configure simple, non-verified email/password to get started
    emailAndPassword: {
      enabled: true,
      requireEmailVerification: false,
    },
    plugins: [
      // The Convex plugin is required for Convex compatibility
      convex(),
      anonymous({
        disableDeleteAnonymousUser: true,
          // Need to delete the anonymous user after linking
          // There is a bug with the anonymous user
          // biome-ignore lint/suspicious/noConsole: Temporary
          await console.log({ anonymousUser, newUser });
        },
      }),
    ],
  });
};

// Example function for getting the current user
// Feel free to edit, omit, etc.
export const getCurrentUser = query({
  args: {},
  returns: v.union(v.any(), v.null()),
  handler: async (ctx) => authComponent.getAuthUser(ctx),
});


This is the contents of the `auth.ts` inside the convex folder
```

### Additional context

_No response_

rodrigosdev:
# Summary

- Guard the anonymous post-auth middleware so it only deletes the previous anonymous account when linking to a new, non-anonymous user.This keeps newly created Convex anonymous users alive and fixes #5824.
- Leave the onLinkAccount callback behavior unchanged, ensuring integrators can still migrate data before cleanup runs.
- Add regression tests that mock the middleware context for both the Convex-style “same anonymous session” flow (no deletion occurs) and the standard linking flow (cleanup still runs).

# Testing
```
pnpm --filter @better-auth/core build
pnpm --filter @better-auth/telemetry build
pnpm --filter better-auth exec vitest run src/plugins/anonymous/anon.test.ts
```

# Breaking Changes
None.

# Checklist
[x] Tests added/updated for the new behavior
[x] All targeted tests passing locally
[x] References the reported bug better-auth/better-auth#5824

---
## Summary by cubic
Prevents anonymous cleanup from deleting a newly created anonymous session when linking. Keeps Convex anonymous sessions intact until a non-anonymous account is linked; fixes #5824.

- **Bug Fixes**
  - Guard post-auth to skip deletion when the new session user is the same or still anonymous; only delete the previous anonymous user when linking to a non-anonymous account.
  - Keep onLinkAccount callback behavior unchanged so data migration still runs before cleanup.
  - Add regression tests for both same-session anonymous and standard linking flows.

<sup>Written for commit 163217199fa29751cca6d85e58c620a7e33edfaa. Summary will update on new commits.</sup>


packages/better-auth/src/plugins/anonymous/anon.test.ts
```diff
@@ -254,4 +256,111 @@ describe("anonymous", async () => {
 			"Anonymous users cannot sign in again anonymously",
 		);
 	});
+
+	describe("anonymous cleanup safeguards", () => {
```
Bekacru: please use `getTestInstance` for writing tests instead of mocking api module

RhysSullivan:
Nice find on this bug + fix, ran into this one as well

ping-maxwell (approved):
lgtm

PR #5834, opened 2025-11-07
fix(stripe): allow custom metadata in subscription_data without overwriting internal fields

bytaesu:
Allow passing metadata to Stripe subscription via `subscription_data` while preserving internal fields

Closes https://github.com/better-auth/better-auth/issues/4104

---
## Summary by cubic
Allows custom metadata in Stripe checkout sessions via subscription_data and metadata, merging safely with internal fields and trial settings. Also passes through safe checkout session params while blocking overrides of protected fields.

- **Bug Fixes**
  - Merge subscription_data with freeTrial to keep trial_period_days and preserve subscription_data.metadata.
  - Merge custom metadata with internal keys (userId, subscriptionId, referenceId); internal keys win.
  - Pass through non-protected params from getCheckoutSessionParams (e.g., allow_promotion_codes, tax_id_collection, custom_text) and options (e.g., idempotencyKey); block overrides of customer, customer_email, line_items, mode, client_reference_id, success_url, cancel_url.
  - Added tests for metadata merging, safe param passthrough, and idempotencyKey.

<sup>Written for commit ef666b830adecde4038bfd417ff82ef72e13282c. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/stripe/src/index.ts
+++ b/packages/stripe/src/index.ts
@@ -575,13 +575,24 @@ export const stripe = <O extends StripeOptions>(options: O) => {
 					}
 				}
 
-				// Extract only the fields we explicitly want to use
-				const { subscription_data: subscriptionData, metadata } =
-					params?.params || {};
+				// Exclude fields from restParams to prevent override
+				const {
+					subscription_data: subscriptionData,
+					metadata,
+					customer,
+					customer_email,
+					line_items,
+					mode,
+					client_reference_id,
+					success_url,
+					cancel_url,
+					...restParams
+				} = params?.params || {};
 
 				const checkoutSession = await client.checkout.sessions
 					.create(
 						{
+							...restParams,
 							...(customerId
 								? {
 										customer: customerId,


packages/stripe/src/stripe.test.ts
```diff
@@ -2176,4 +2176,218 @@ describe("stripe", async () => {
...
+					email: "sub-metadata-test@email.com",
+					password: "password",
+					name: "Subscription Metadata Test",
+				},
+				{ throw: true },
+			);
+			const userId = userRes.user.id;
+
+			const customCtx = await customAuth.$context;
+			await customCtx.adapter.create({
```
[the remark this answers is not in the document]
bytaesu: It’s not used here

packages/stripe/src/index.ts
```diff
@@ -603,18 +619,18 @@ export const stripe = <O extends StripeOptions>(options: O) => {
...
+								...freeTrial,
+							}),
 							metadata: {
+								...metadata,
 								userId: user.id,
 								subscriptionId: subscription.id,
 								referenceId,
-								...params?.params?.metadata,
 							},
+							...restParams,
```
Bekacru: there could be senstive fields that'll be set by the user here beyond the ones you destructured. We sohuld avoid passing the whole object as is. Remove it

Bekacru (changes requested):

packages/stripe/src/index.ts
bytaesu: okay👌

bytaesu: wouldn’t this prevent users from customizing like this?

cc @Bekacru

https://www.better-auth.com/docs/plugins/stripe#custom-checkout-session-parameters

bytaesu: The reason I destructured is to allow customization while preventing the internal logic from being overwritten 🤔

PR #5838, opened 2025-11-07
feat(username): add `UserWithUsername` type

jslno:
## Summary by cubic
Add a UserWithUsername type to the username plugin, extending User with username and displayUsername fields. This improves type safety and clarifies the expected user shape when working with usernames.

<sup>Written for commit dd64155def2391e059512259b091dc4922c06371. Summary will update automatically on new commits.</sup>


himself65 (changes requested):
I think we should have an InferUser type

PR #5842, opened 2025-11-07
feat(sso): add optional support for domain verification

jonathansamines:
## Summary by cubic
Adds optional domain verification to the SSO plugin and client. Verifies provider domains via DNS TXT tokens and automatically trusts verified domains.

- **New Features**
  - POST /sso/request-domain-verification to submit a provider for verification.
  - Issues verification tokens and checks DNS TXT records.
  - Supports configurable DNS TXT record prefix.
  - OIDC/SAML routes enforce verification status and provider ownership.
  - Endpoints register only when the feature is enabled.

- **Migration**
  - If enabled, run: npx @better-auth/cli migrate and npx @better-auth/cli generate to apply schema changes.

<sup>Written for commit b1dda426f94ad55016a96d2f91a67d739dfe6117. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/sso.mdx
+++ b/docs/content/docs/plugins/sso.mdx
@@ -704,7 +704,7 @@ mapping: {
 }
 ```
 
-### Domain verification
+## Domain verification
 
 You can automatically trust a new domain by enabling the domain verification feature:
 
@@ -753,9 +753,9 @@ Once enabled, make sure you migrate the database schema (again).
     </Tab>
 </Tabs>
 
-See the [Schema](#schema) section to add the fields manually.
+See the [Schema](#schema-for-domain-verification) section to add the fields manually.
 
-#### Verify your domain
+### Verify your domain
 
 When domain verification is enabled, every new SSO provider domain will be marked as unverified at first.
 Upon SSO provider registration, a **verification token** will be issued to the provider (it will be returned as part of the response),
@@ -768,7 +768,7 @@ which you need to prove that you own it. To do this, you'll need to add a `TXT` 
 3.  **Save the record and wait for it to propagate.** This can take up to 48 hours, but it's usually much faster.
 4.  **Once the record has propagated**, you can submit a validation request
 
-#### Domain validation request
+### Domain validation request
 
 Once you have configured your domain, you can use your `auth` instance to submit a validation request.
 This request will either result in a rejection (could not prove your ownership over the domain)
@@ -795,16 +795,13 @@ const res = await auth.api.verifyDomain({
     </Tab>
 </Tabs>
 
-#### Creating a new verification token
+### Creating a new verification token
 
 Every domain verification token will have a default expiry of 1 week since the moment it was issued
 or the moment when the SSO provider was registered.
 
 After that time, the token will expire and cannot longer be used. When that happens,
 you can create a new verification token:
-
-```ts title="create-domain-verification-token.ts"
-```
 
 <Tabs items={["client", "server"]}>
     <Tab value="client">


packages/sso/src/routes/sso.ts
```diff
@@ -0,0 +1,2079 @@
...
+					responses: {
+						"302": {
+							description: "Redirects to the callback URL",
+						},
+					},
+				},
+			},
+		},
+		async (ctx) => {
+			const { code, state, error, error_description } = ctx.query;
```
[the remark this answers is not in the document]
jonathansamines: This is existing code, just moved around.

packages/sso/src/routes/sso.ts
```diff
@@ -0,0 +1,2079 @@
...
+					| "client_secret_basic"
+					| "client_secret_post";
+			}>(config.discoveryEndpoint);
+
+			if (discovery.data) {
+				config = {
+					tokenEndpoint: discovery.data.token_endpoint,
+					tokenEndpointAuthentication:
+						discovery.data.token_endpoint_auth_method,
+					userInfoEndpoint: discovery.data.userinfo_endpoint,
```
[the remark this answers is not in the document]
jonathansamines: This is existing code, just moved around.

packages/sso/src/routes/sso.ts
```diff
@@ -0,0 +1,2079 @@
...
+						: undefined,
+					scopes: ctx.body.scopes ||
+						provider.oidcConfig.scopes || [
+							"openid",
+							"email",
+							"profile",
+							"offline_access",
+						],
+					loginHint: ctx.body.loginHint || email,
+					authorizationEndpoint: provider.oidcConfig.authorizationEndpoint!,
```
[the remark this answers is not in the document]
jonathansamines: This is existing code, just moved around.

packages/sso/src/domain-verification.test.ts
```diff
@@ -0,0 +1,343 @@
...
+			expect(await response.json()).toEqual({ message: "Provider not found" });
+		});
+
+		it("should return conflict if there is an active verification token", async () => {
+			const { auth, getAuthHeaders, registerSSOProvider } = createTestAuth();
+			const headers = await getAuthHeaders();
+			const provider = await registerSSOProvider(headers);
+
+			vi.useFakeTimers();
+			vi.advanceTimersByTime(Date.now() - 100);
```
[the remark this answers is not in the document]
jonathansamines: This should be fine and it is working as intended

docs/content/docs/plugins/sso.mdx
```diff
@@ -704,6 +704,93 @@ mapping: {
...
+
+#### Verify your domain
+
+When domain verification is enabled, every new SSO provider domain will be marked as unverified at first.
+Upon registration, a **verification token** will be issued to the provider,
+which you need to verify it to prove that you own it. To do this, you'll need to add a `TXT` record to your domain's DNS settings.
+
+1.  **You will be provided with a verification token.** This token is a unique string that you'll need to add to your DNS records.
+2.  **Create a new `TXT` record in your DNS provider's dashboard.**
+    *   **Host:** `ba-domain-verification-{your-provider-id}`
```
Bekacru: can we make the host configurable?

docs/content/docs/plugins/sso.mdx
```diff
@@ -704,6 +704,93 @@ mapping: {
...
+    ```
+    </Tab>
+</Tabs>
+
+See the [Schema](#schema) section to add the fields manually.
+
+#### Verify your domain
+
+When domain verification is enabled, every new SSO provider domain will be marked as unverified at first.
+Upon registration, a **verification token** will be issued to the provider,
```
Bekacru: will I get the verification token as a response of `registerProvider` or is there additional step? not clear (i've not looked at the code yet so this is just docs comment)

docs/content/docs/plugins/sso.mdx
```diff
@@ -704,6 +704,93 @@ mapping: {
...
+4.  **Once the record has propagated**, you can submit a validation request
+
+#### Domain validation request
+
+Once you have configured your domain, you can use your `auth` instance to submit a validation request.
+This request will either result in a rejection (could not prove your ownership over the domain)
+or if the verification is successful, your SSO provider domain will be marked as verified.
+
+```ts title="domain-validation-request.ts"
+const response = await auth.api.verifyDomain({
```
Bekacru: can I not do this from the client? - We should have example for both client and server api usage like other plugin examples

docs/content/docs/plugins/sso.mdx
```diff
@@ -704,6 +704,93 @@ mapping: {
...
+#### Creating a new verification token
+
+Every domain verification token will have a default expiry of 1 week since the moment it was issued
+or the moment when the SSO provider was registered.
+
+After that time, the token will expire and cannot longer be used. When that happens,
+you can create a new verification token:
+
+```ts title="create-domain-verification-token.ts"
+const response = await auth.api.submitDomainVerification({
```
Bekacru: I think the api should be named `requestDomainVerification`

docs/content/docs/plugins/sso.mdx
```diff
@@ -704,6 +704,93 @@ mapping: {
...
+    body: {
+        providerId: "acme-corp",
+    },
+    headers,
+});
+
+console.log("Token: ", response.domainVerificationToken);
+```
+
+#### Domain verification endpoints
```
Bekacru: we should remove this section. see how we define endpoint examples in other plugins docs

docs/content/docs/plugins/sso.mdx
```diff
@@ -727,6 +814,7 @@ The plugin requires additional fields in the `ssoProvider` table to store the pr
         { name: "userId", type: "string", description: "The user ID", isRequired: true, references: { model: "user", field: "id" } },
         { name: "providerId", type: "string", description: "The provider ID. Used to identify a provider and to generate a redirect URL.", isRequired: true, isUnique: true },
         { name: "organizationId", type: "string", description: "The organization Id. If provider is linked to an organization.", isRequired: false },
+        { name: "domainVerified", type: "boolean", description: "A flag indicating whether the provider domain has been verified. Only available when domain verification is enabled", isRequired: false },
```
Bekacru: if this field is only added after I enabled domain verification, it should be separated out with a new title instead of adding to the main schema (haven't looked at the code yet so if it's by default required weather you enable domain verification or not this doc is correct but we shouldn't be doing that)

docs/content/docs/plugins/sso.mdx
```diff
@@ -790,6 +878,23 @@ If you want to allow account linking for specific trusted providers, enable the
...
+    domainVerification: {
+        description: "Configure the domain verification feature",
+        type: "object",
+        properties: {
+            enabled: {
+                description: "Enables or disables the domain verification feature",
+                type: "boolean",
+                required: false
+            },
+            verificationTokenPrefix: {
```
Bekacru: this should probably be named `tokenPrefix`

docs/content/docs/plugins/sso.mdx
```diff
@@ -790,6 +878,23 @@ If you want to allow account linking for specific trusted providers, enable the
...
+                description: "Prefix to append to the verification token identifier",
+                type: "string",
+                required: false,
+                default: "ba-domain-verification-"
```
Bekacru: let's do default `better-auth-sso` or something

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,196 @@
+import dns from "node:dns/promises";
+import type { Verification } from "better-auth";
+import { createAuthEndpoint, sessionMiddleware } from "better-auth/api";
+import { randomBytes } from "crypto";
+import * as z from "zod/v4";
+import type { SSOOptions, SSOProvider } from "../types";
+
+export const submitDomainVerification = (options: SSOOptions) => {
+	return createAuthEndpoint(
+		"/sso/domain-verification",
```
Bekacru: this path determines the client method. this converts to `authClient.sso.domainVerification` it should be `/sso/request-domain-verification`

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,196 @@
+import dns from "node:dns/promises";
```
Bekacru: we probably shouldn't import anything from `node:` directly as it breaks support for next js middleware...nextjs middleware doesn't work if you're using `node:` prefixed imports

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,196 @@
...
+import type { Verification } from "better-auth";
+import { createAuthEndpoint, sessionMiddleware } from "better-auth/api";
+import { randomBytes } from "crypto";
```
Bekacru: instead of crypto - we sohuld probably utilize web apis. But I actually would suggest implementing something within better-auth-utils (https://github.com/better-auth/utils) or using the existing utilities

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,196 @@
...
+		},
+		async (ctx) => {
+			const body = ctx.body;
+			const provider = await ctx.context.adapter.findOne<SSOProvider>({
+				model: "ssoProvider",
+				where: [{ field: "providerId", value: body.providerId }],
+			});
+
+			if (!provider) {
+				return Response.json(
```
Bekacru: you shouldn't return a response obejct. see other endpoints. we should be throwing error here. `throw new APIError("NOT_FOUND")`

a return will change the response type and it'll not be properly inferred on ts

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,196 @@
...
+					{ message: "Provider not found" },
+					{ status: 404 },
+				);
+			}
+
+			if (provider.domainVerified) {
+				return Response.json(
+					{ message: "Domain has already been verified" },
+					{ status: 409 },
```
Bekacru: same here

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,196 @@
...
+						field: "identifier",
+						value: options.domainVerification?.verificationTokenPrefix
+							? `${options.domainVerification?.verificationTokenPrefix}-${provider.providerId}`
+							: `ba-domain-verification-${provider.providerId}`,
+					},
+					{ field: "expiresAt", value: new Date(), operator: "lt" },
+				],
+			});
+
+			if (verification) {
```
Bekacru, line 98: if its valid we probably should generate new one and show them that instead of returning there is an existing one...and again resposne type doesn't work

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,196 @@
...
+				return Response.json(
+					{ message: "Current verification token is still valid" },
+					{ status: 409 },
+				);
+			}
+
+			let domainVerificationToken: string | undefined;
+
+			if (options?.domainVerification?.enabled) {
+				domainVerificationToken = randomBytes(24).toString("hex");
```
Bekacru: use better-auth-utils

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,196 @@
...
+						updatedAt: new Date(),
+						value: domainVerificationToken,
+						expiresAt: new Date(Date.now() + 3600 * 24 * 7 * 1000), // 1 week
+					},
+				});
+			}
+
+			return Response.json(
+				{
+					domainVerificationToken,
```
Bekacru, line 126: We haven't made any check if the person who is asking for the verification token has ownership over the provider...if provider has organizationId we should check if the user is a member of that org. If not, we should check if the provider.userId matches the userId

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,196 @@
...
+				},
+				{ status: 201 },
+			);
+		},
+	);
+};
+
+export const verifyDomain = (options: SSOOptions) => {
+	return createAuthEndpoint(
+		"/sso/domain-verification/verify",
```
Bekacru: we generally avoid 3 paths for an endpoint. since this determines the client action. should benamed `/sso/verify-domain` or something

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,196 @@
...
+		},
+		async (ctx) => {
+			const body = ctx.body;
+			const provider = await ctx.context.adapter.findOne<SSOProvider>({
+				model: "ssoProvider",
+				where: [{ field: "providerId", value: body.providerId }],
+			});
+
+			if (!provider) {
+				return Response.json(
```
Bekacru: again remove Response

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,196 @@
...
+					error,
+				);
+				records = [];
+			}
+
+			const record = records.find((record) =>
+				record.includes(`${verification.identifier}=${verification.value}`),
+			);
+			if (!record) {
+				return Response.json(
```
Bekacru: again response

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,196 @@
...
+					{ message: "Unable to verify domain ownership. Try again later" },
+					{ status: 404 },
+				);
+			}
+
+			await ctx.context.adapter.update<SSOProvider>({
```
Bekacru: here also we haven't made any check if the user who is verifying has authorization over the provider. check userid or org membership

packages/sso/src/routes/sso.ts
```diff
@@ -0,0 +1,2081 @@
...
+import type { FlowResult } from "samlify/types/src/flow";
+import * as z from "zod/v4";
+import type { OIDCConfig, SAMLConfig, SSOOptions, SSOProvider } from "../types";
+
+/**
+ * Safely parses a value that might be a JSON string or already a parsed object
+ * This handles cases where ORMs like Drizzle might return already parsed objects
+ * instead of JSON strings from TEXT/JSON columns
+ */
+function safeJsonParse<T>(value: string | T | null | undefined): T | null {
```
Bekacru, line 34: we already have a helper for this in the `core`

packages/sso/src/routes/sso.ts
```diff
@@ -0,0 +1,2081 @@
...
+
+export const spMetadata = () => {
+	return createAuthEndpoint(
+		"/sso/saml2/sp/metadata",
+		{
+			method: "GET",
+			query: z.object({
+				providerId: z.string(),
+				format: z.enum(["xml", "json"]).default("xml"),
+			}),
```
Bekacru, line 65: I think we should require auth for this...according to my research this isnt meant to be a public enpdoint. So we should require user session and also check if the user is authorized to fetch the sp metadata

Bekacru: I know this isn't part of the PR it's just moved here...just wanted to point out

packages/sso/src/routes/sso.ts
```diff
@@ -0,0 +1,2081 @@
...
+							: `ba-domain-verification-${provider.providerId}`,
+						createdAt: new Date(),
+						updatedAt: new Date(),
+						value: domainVerificationToken,
+						expiresAt: new Date(Date.now() + 3600 * 24 * 7 * 1000), // 1 week
+					},
+				});
+			}
+
+			return ctx.json({
```
Bekacru, line 684: the type here won't properly be inferred

Bekacru: `domainVerified` and `domainVerificationToken` should be inferred properly based on if the option is enabled or not

Bekacru: if it's enabled they'll always be defined, if not the keys dont even exist

packages/sso/src/routes/sso.ts
```diff
@@ -0,0 +1,2081 @@
...
+						provider.oidcConfig.scopes || [
+							"openid",
+							"email",
+							"profile",
+							"offline_access",
+						],
+					loginHint: ctx.body.loginHint || email,
+					authorizationEndpoint: provider.oidcConfig.authorizationEndpoint!,
+				});
+				return ctx.json({
```
Bekacru, line 1008: before returning here we should check if the domain is verified or not no?

packages/sso/src/routes/sso.ts
```diff
@@ -0,0 +1,2081 @@
...
+				const loginRequest = sp.createLoginRequest(
+					idp,
+					"redirect",
+				) as BindingContext & { entityEndpoint: string; type: string };
+				if (!loginRequest) {
+					throw new APIError("BAD_REQUEST", {
+						message: "Invalid SAML request",
+					});
+				}
+				return ctx.json({
```
Bekacru, line 1046: here as well, before returinng we should check if the domain is verified or not

packages/sso/src/routes/sso.ts
```diff
@@ -0,0 +1,2081 @@
...
+			}
+
+			if (!userInfo.email || !userInfo.id) {
+				throw ctx.redirect(
+					`${
+						errorURL || callbackURL
+					}/error?error=invalid_provider&error_description=missing_user_info`,
+				);
+			}
+			const linked = await handleOAuthUserInfo(ctx, {
```
Bekacru, line 1328: before account linking or doing any of the process we should be checking if domain is verified here as well

packages/sso/src/routes/sso.ts
```diff
@@ -0,0 +1,2081 @@
...
+							samlConfig: res.samlConfig
+								? safeJsonParse<SAMLConfig>(
+										res.samlConfig as unknown as string,
+									) || undefined
+								: undefined,
+						};
+					});
+			}
+
+			if (!provider) {
```
Bekacru, line 1489: we should check if domain is verified

packages/sso/src/routes/sso.ts
```diff
@@ -0,0 +1,2081 @@
...
+							samlConfig: res.samlConfig
+								? safeJsonParse<SAMLConfig>(
+										res.samlConfig as unknown as string,
+									) || undefined
+								: undefined,
+						};
+					});
+			}
+
+			if (!provider?.samlConfig) {
```
Bekacru, line 1844: we should check if domain is verified

packages/sso/src/index.ts
```diff
@@ -37,2231 +29,67 @@ const fastValidator = {
...
-					let session: Session =
-						await ctx.context.internalAdapter.createSession(user.id);
-					await setSessionCookie(ctx, { session, user });
-
-					const callbackUrl =
-						RelayState || parsedSamlConfig.callbackUrl || ctx.context.baseURL;
-					throw ctx.redirect(callbackUrl);
-				},
-			),
+			...(endpoints as SSOEndpoints),
```
Bekacru: function overloading should work for the type inference but I think you could also just infer the type only for the endpoints...not sure which one is the best...ignore this for now

Bekacru:
Also client plugin is missing

docs/content/docs/plugins/sso.mdx
jonathansamines: It is already configurable via the `domainVerification.verificationTokenPrefix` option. Please let me know if that is what you had in mind.

docs/content/docs/plugins/sso.mdx
jonathansamines: Yeah, that'll be the case, wasn't exactly sure to what extent we do document API responses, but can definitely do that.

docs/content/docs/plugins/sso.mdx
jonathansamines: Yeah, there is no client plugin for now, let me work on that

docs/content/docs/plugins/sso.mdx
jonathansamines: Yeah, this is conditionally added only when domain verification is enabled. I'll create a separate section for this field.

packages/sso/src/routes/domain-verification.ts
jonathansamines: I was just wondering about that, now I know. Let me change all errors to `APIError`

packages/sso/src/routes/sso.ts
jonathansamines: Everything on this file was already there, I just moved everything to a separate file to aid with type definitions. I'll defer all comments related to this file for now, and create follow-up PRs for them.

packages/sso/src/routes/domain-verification.ts
jonathansamines: That's interesting, let me fix this.

PR #5857, opened 2025-11-08
feat: add typed event emitter

jslno:
Related: #5673 

Introduces a built in event emitter.
Unlike hooks, it allows multiple independent listeners to subscribe to the same event from anywhere in userland.

---
## Summary by cubic
Adds a built-in, typed event emitter to BetterAuth and emits default auth lifecycle events (signIn, signOut, signUp, verifyEmail, passwordReset). Apps and plugins can subscribe from anywhere with strong types.

- **New Features**
  - Added EventEmitter/EventMixin with default events and plugin-extensible types.
  - Exposed auth.event and wired the emitter into AuthContext/init.
  - Emitting events in core routes: sign-in, sign-out, sign-up, email verification, and password reset.

<sup>Written for commit e3602b4be79f78bddf458820a9d92e5a194b10da. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/__snapshots__/init.test.ts.snap
+++ b/packages/better-auth/src/__snapshots__/init.test.ts.snap
@@ -67,6 +67,11 @@ exports[`init > should match config 1`] = `
   },
   "baseURL": "http://localhost:3000/api/auth",
   "createAuthCookie": [Function],
+  "event": {
+    "emit": [Function],
+    "events": {},
+    "on": [Function],
+  },
   "generateId": [Function],
   "internalAdapter": {
     "countTotalUsers": [Function],


jslno:
My thought behind this was to allow listening to more high-level events, making things like audit logging easier, since right now it's a bit impractical to add sign-in hooks (particularly distinguishing sign-in from sign-up inside oauth callback). I wasn't happy with it either, so I left it as draft for now. I fully agree it violates the design pattern, and creating a second abstraction layer might confuse users on when to use what. I should've spent a little more time to choose a better approach from the beginning.

However, adding more high-level hooks could still be practical, allowing users to more easily listen to common events without touching specific endpoints. A more applicable approach could be to just add callbacks like `account.afterSignIn` to options for events that are not bound to a single endpoint (otherwise just use hooks). This would achieve the above while staying consistent with the current design pattern (e.g. `user.deleteUser.beforeDelete`).

Let me know if that's something you'd consider, otherwise I'll just close it

PR #5869, opened 2025-11-09
feat: allow Zod schema for additionalFields and inferAdditionalFields

### LINKED ISSUE #5854: Allow zod schema for `inferAdditionalFields`

Currently, we either have to have a monorepo, or type the better-auth function manuallly in all clients.

It would be really interesting to allow the use of a zod schema for this function. Better-auth uses zod internally. Is there any reason this is not possible? Thank you! 🙏 

Using another library? 🤔

### Additional context

_No response_

GautamBytes:
Closes #5854

This PR adds support for using a `z.ZodObject` to define `additionalFields`, both in the main `BetterAuthOptions` on the server and with the `inferAdditionalFields` helper on the client.

This solves the problem of needing to manually type the `DBFieldAttribute` map on the client, which was cumbersome and error-prone.

#### Changes
* **`from-zod.ts` (new):** Adds a runtime converter `fromZodSchema` that maps a `z.ZodObject` to the internal `DBFieldAttribute` format.
* **`from-zod-types.ts` (new):** Adds a compile-time type converter `ZodSchemaToDBFields` to provide full type-safety and inference.
* **`get-tables.ts` & `schema.ts`:** Updated to use the `fromZodSchema` converter at runtime when a Zod schema is provided in the options.
* **`client.ts`:** Updated `inferAdditionalFields` to use the `ZodSchemaToDBFields` type for compile-time inference.
* **`field.ts` & `index.ts`:** Updated to export the new utilities and integrate the types.

---
## Summary by cubic
Adds first-class Zod support for additionalFields on server and client, with runtime conversion and full type inference. This removes manual DBFieldAttribute maps and improves input validation.

- **New Features**
  - Accept z.ZodObject for user/session/account additionalFields; convert at runtime via fromZodSchema.
  - inferAdditionalFields now infers types from Zod using ZodSchemaToDBFields.
  - InferFieldsFromOptions updated to support Zod schemas.
  - Export fromZodSchema and ZodSchemaToDBFields from the db package.

- **Bug Fixes**
  - Input parsing detects Zod validators and returns Zod error details.
  - Default values are executed when functions, not stored as functions.

<sup>Written for commit 5eb3669ea5a87dfd433a2a1a00b9d9bf87a4500f. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/db/from-zod-types.ts
+++ b/packages/better-auth/src/db/from-zod-types.ts
@@ -1,6 +1,9 @@
 import type { DBFieldAttribute } from "@better-auth/core/db";
 import type { ZodTypeAny, z } from "zod";
 
+/**
+ * Maps a single Zod type (z.ZodString, etc.) to a DBFieldType string ("string", etc.)
+ */
 type MapZodType<T extends ZodTypeAny> = T extends z.ZodString
 	? "string"
 	: T extends z.ZodNumber
--- a/packages/better-auth/src/db/from-zod.ts
+++ b/packages/better-auth/src/db/from-zod.ts
@@ -8,20 +8,27 @@ export function isZodObject(value: unknown): value is z.ZodObject<any> {
 	}
 	try {
 		return value instanceof z.ZodObject;
-	} catch {
+	} catch (err) {
+		console.error("Error in isZodObject check:", err);
 		return false;
 	}
 }
 
 export function isZodType(value: unknown): value is z.ZodTypeAny {
 	if (!!value && typeof value === "object") {
-		const typeName = (value as any)?._def?.typeName;
-		if (typeof typeName === "string") return true;
-		if (typeof (value as any).parse === "function") return true;
+		const def = (value as any)?._def;
+		if (
+			def &&
+			typeof def.typeName === "string" &&
+			typeof (value as any).parse === "function"
+		) {
+			return true;
+		}
 	}
 	try {
 		return value instanceof z.ZodType;
-	} catch {
+	} catch (err) {
+		console.error("Error in isZodType check:", err);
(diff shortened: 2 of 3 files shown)


himself65 (changes requested):
we should support general `StandardSchema` here

GautamBytes:
Hey @himself65 

I've been digging into the StandardSchema support for inferAdditionalFields like you suggested and hit a design question.

It looks like the StandardSchema interface is perfect for validation (with the validate() function), but I can't find a way to get the schema's shape (introspection) from it. I need that shape info to automatically generate the DB fields.

Zod schemas work because I can inspect their shape, but a generic StandardSchema seems like a black box.

What's the best way to handle this?

- Should I just make this a "Zod-only" feature for now? (i.e., I'll check if it's StandardSchema and also isZodObject).
- Or should we ask users to provide both the DBFieldAttribute object (for the shape) and a StandardSchema (for validation)?
- Or am I just totally missing an easy way to get the shape from a StandardSchema?

Just want to make sure I'm building this right. Thanks!

taylorfsteele:
> we should support general `StandardSchema` here

Gonna +1 the request for standard schema here. I just tried to add types to an additional field with Valibot:

```
import * as v from "valibot";
// ..
user: {
    additionalFields: {
      preferences: {
        type: "string",
        validator: {
          input: v.string() // Type error
        }
      },
    },
  },
```

```
Type 'StringSchema<undefined>' is missing the following properties from type 'ZodType<unknown, unknown, $ZodTypeInternals<unknown, unknown>>': def, _def, _output, _input, and 38 more.
```

Standard schema support here would be amazing. Happy to break this out into it's own issue if that's easier to track.

PR #5880, opened 2025-11-10
fix(adapter): default value in prisma schema

rovertrack:
## Summary by cubic
Fix default value generation in Prisma schema so fields get valid defaults across providers. For MySQL, avoid invalid defaults on TEXT and support long string defaults via dbgenerated.

- **Bug Fixes**
  - Apply default(...) for string, boolean, and number fields; for MySQL strings, only when the value is <255 chars.
  - For MySQL strings: use @db.Text only when there’s no default or the default is >255 chars; for long defaults, set default(dbgenerated(...)).
  - Added snapshot tests for PostgreSQL and MySQL; updated MySQL custom snapshots for role and status defaults.

<sup>Written for commit 4c931cb5454cd375aa126bda127fff40ec2e7b69. Summary will update automatically on new commits.</sup>


himself65:
Closes https://github.com/better-auth/better-auth/pull/5880

PR #5885, opened 2025-11-10
type ( refactor )

darad25:
i created an ensure email verification helper, it ensure email verification is enabled in options and return the object, it helps make the send verification email fn function simpler and more efficient, i feel using helpers gives code a better structure.

---
## Summary by cubic
Centralized email verification checks and URL building, added secure OAuth refresh token handling, and cleared stale session data on expiry. Improved callback URL handling across reset-password, sign-in, sign-up, change-email, and delete-user, plus hardened OAuth state parsing and fixed update hooks.

- **Refactors**
  - Email verification: applied ensureEmailVerificationEnabled across routes, switched to zod/v4, and centralized buildVerificationUrl.
  - OAuth/linking/cookies/migrations: decrypt refreshToken before refresh; clear session-data cookie on missing/expired session; include callback URLs in OAuth state and optionally store account cookie after linking; normalize cross-subdomain cookie domain; chunk session/account cookies by byte length; arrays stored as JSON and avoid NOT NULL without defaults; with-hooks: pass actualData to update.before and reuse adapter instance.
  - Sign-up/update-user: explicit schema and originCheck-decode callbackURL; lowercase email for lookup/create/token; block restricted fields and validate name/image.

- **Bug Fixes**
  - Delete-user: fix session freshness check to prevent false “expired” errors.
  - Cookies: split combined Set-Cookie when Expires contains commas.
  - Internal adapter: correct session TTL math and avoid premature loop exits.
  - OAuth state: fix cookie maxAge units and handle bad JSON with safe redirect.

<sup>Written for commit 0841e67540890fac404aaeaf91dc2608d2d4f0f0. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/admin/routes.ts
+++ b/packages/better-auth/src/plugins/admin/routes.ts
@@ -346,6 +346,21 @@ export const createUser = <O extends AdminOptions>(opts: O) =>
 				throw APIError.from("BAD_REQUEST", BASE_ERROR_CODES.INVALID_EMAIL);
 			}
 
+			// Validate provided role(s) against allowed role list if present
+			if (ctx.body.role && opts.roles) {
+				const inputRoles = Array.isArray(ctx.body.role)
+					? ctx.body.role
+					: [ctx.body.role];
+				for (const role of inputRoles) {
+					if (!opts.roles[role as keyof typeof opts.roles]) {
+						throw APIError.from(
+							"BAD_REQUEST",
+							ADMIN_ERROR_CODES.YOU_ARE_NOT_ALLOWED_TO_SET_NON_EXISTENT_VALUE,
+						);
+					}
+				}
+			}
+
 			const existUser =
 				await ctx.context.internalAdapter.findUserByEmail(email);
 			if (existUser) {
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/api/routes/email-verification.ts
darad25, line 1: I made use of the ensureEmailVerificationEnabled(ctx) helper in sendVerificationEmail, also in the verifyEmail i made use of the ensureEmailVerificationEnabled(ctx) helper so all endpoint code are consistent.

darad25 (commented):
I implemented an add origin validation middleware with decoding included callback URLs in OAuth state, fixed ID token not supported error to BAD_REQUEST, removed duplicate operationIds and standardized the parseUserOutput import.

darad25 (commented):
I added stricter request validation with passthrough for extra fields, validate and decode callbackURL via the origin check middleware, de-duplicate the operationId, standardize parseUserOutput import and normalize email casing before lookups and creation.

himself65:
Please use a meaningful PR title

PR #5887, opened 2025-11-10
fix(adpaters/drizzle-adapter) use primary db connection for transactions

JacobSchroder:
When using drizzle with [read replicas](https://orm.drizzle.team/docs/read-replicas), transactions that have both read and write operations seem to use both connections (primary and read-only), resulting in the writes not being reflected on to the read-only instances (before the transaction is committed).

For instance, when using the emailOTP plugin, and users signed up using this method the following would happen within the same drizzle transaction:

1. The user would be created (primary/writer instance)
2. Read the newly created user (read replica) > `TypeError: null is not an object (evaluating 'user.banned')`

I did a workaround in my application using better-auth and the drizzle-adapter, by passing the primary db instance to the drizzleAdapter `drizzleAdapter(db.$primary, ...)`, which fixed the issue, but completely disables using read replicas with better-auth.

This PR should solve the issue, while still allowing for the use of read replicas (when transactions are not used, internally)

---
## Summary by cubic
Run transactions on the primary DB connection in the Drizzle adapter (falling back to the default when no primary exists). This avoids mixed read/write transactions hitting read replicas and missing uncommitted writes (e.g., email OTP user reads returning null).

<sup>Written for commit 6146dbda001744ceac88ee5297bc905c5ded8f68. Summary will update automatically on new commits.</sup>


JacobSchroder:
Drizzle-orm always handles this internally
https://github.com/drizzle-team/drizzle-orm/blob/ad4ddd444d066b339ffd5765cb6ec3bf49380189/drizzle-orm/src/pg-core/db.ts#L671

PR #5888, opened 2025-11-10
feat: support pass raw function as middleware

himself65:
## Summary by cubic
Support passing raw functions as middleware in plugin hooks, reducing boilerplate and making plugins easier to write. Raw functions are auto-wrapped with createAuthMiddleware and a default matcher.

- **New Features**
  - Plugin hooks (before/after) accept either {matcher, handler} or a raw middleware function.
  - Raw functions are wrapped with createAuthMiddleware and use a matcher that always returns true.

- **Refactors**
  - optionsMiddleware is no longer re-exported from better-auth.
  - Removed the unused shimEndpoint helper.

<sup>Written for commit f053400d8637916be15383e91d90896d0c39f0d9. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/src/types/plugin.ts
+++ b/packages/core/src/types/plugin.ts
@@ -36,7 +36,9 @@ export type PluginHook =
 			matcher: (context: HookEndpointContext) => boolean;
 			handler: AuthMiddleware;
 	  }
-	| ((context: MiddlewareContext<MiddlewareOptions, any>) => Promise<unknown>);
+	| ((
+			context: MiddlewareContext<MiddlewareOptions, AuthContext>,
+	  ) => Promise<unknown>);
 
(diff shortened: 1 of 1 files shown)


packages/core/src/types/plugin.ts
```diff
@@ -29,6 +31,13 @@ export type HookEndpointContext = Partial<
 	headers?: Headers | undefined;
 };
 
+export type PluginHook =
+	| {
+			matcher: (context: HookEndpointContext) => boolean;
+			handler: AuthMiddleware;
+	  }
+	| ((context: MiddlewareContext<MiddlewareOptions, any>) => Promise<unknown>);
```
himself65: type is missing

PR #5894, opened 2025-11-11
fix(oauth2): fix user data not reflecting provider updates with overrideUserInfo

dandamian:
When `overrideUserInfo` is enabled, `updateUser()` was called but its return value wasn't captured, causing sessions to be created with stale user data instead of the updated information from the OAuth provider.

---
## Summary by cubic
Fixed OAuth sessions using stale user data when overrideUserInfo is enabled. We now assign the result of updateUser to the user object so sessions reflect provider updates on sign-in and account linking.

<sup>Written for commit ac9664c3b26449d2da631327b5ff8aeabe16e7ed. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/oauth2/link-account.test.ts
+++ b/packages/better-auth/src/oauth2/link-account.test.ts
@@ -232,7 +232,6 @@ describe("oauth2 - email verification on link", async () => {
 });
 
 describe("oauth2 - override user info on sign-in", async () => {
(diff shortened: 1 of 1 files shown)


dandamian:
Hey @himself65 thanks for looking into it! The only way I could reproduce this issue was by enabling cookie cache and simulating DB latency on updateUser.

Issue #5543, opened 2025-10-24
Domain Verification for SSO


tnkuehne:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

When implementing self-serve SSO for customers, currently, the only safe way is to manually approve the domains, as there is a serious risk when doing the email matching via domain.

### Describe the solution you'd like

A plugin that does domain verification via TXT DNS records by creating a token for a domain, saving it to the database, and then, in a verification step, checks if the user added the token in a TXT entry and saves the domain as verified in the database.

### Describe alternatives you've considered

- manual approval
- implement outside of better-auth

### Additional context

I have a working prototype of the plugin. If there is interest, I am happy to share, publish to npm, or contribute to better-auth.

TimFL:
Maybe this PR could be enhanced, it seems to be doing something similar (although it's mainly geared at having custom domains for organizations): https://github.com/better-auth/better-auth/pull/3688/files#diff-2200df8a5fb2f75b86d45fcd3a4e80760e0f837fce02486e2006a67d7b615f16

Bekacru:
This is something we're interested on if anyone want to work on it!

PR #5910, opened 2025-11-11
feat(sso): add domain verification for SSO providers

jonathansamines:
Closes: #5543 

---
## Summary by cubic
Adds domain verification for SSO providers. Domains must prove ownership via a DNS TXT token before sign-in is allowed.

- **New Features**
  - Added POST /sso/request-domain-verification and POST /sso/verify-domain with DNS TXT validation and ownership checks (configurable tokenPrefix).
  - Provider registration now issues and returns a 1-week verification token stored in the Verification table.
  - Added domainVerified to the ssoProvider schema and enforce verification across sign-in, callback, and ACS flows.
  - Client plugin now supports domainVerification.enabled and infers server types accordingly.
  - Updated docs with setup and DNS steps; added comprehensive tests.
  - request-domain-verification returns the existing active token when available.

- **Migration**
  - Enable domainVerification in both server (sso) and client (ssoClient).
  - Run npx @better-auth/cli migrate and npx @better-auth/cli generate.
  - Add a TXT record: host better-auth-token-{providerId} (or {tokenPrefix}-{providerId}), value = issued token.
  - Submit verification via auth.api.verifyDomain; request a token via auth.api.requestDomainVerification.
  - Default SSO providers are treated as verified.

<sup>Written for commit ec20fc3e8b0753a80a9468510bb6526d224a0bf7. Summary will update automatically on new commits.</sup>


docs/content/docs/plugins/sso.mdx
```diff
@@ -704,6 +704,128 @@ mapping: {
...
+
+### Verify your domain
+
+When domain verification is enabled, every new SSO provider domain will be marked as unverified at first.
+Upon SSO provider registration, a **verification token** will be issued to the provider (it will be returned as part of the response),
+which you need to prove that you own it. To do this, you'll need to add a `TXT` record to your domain's DNS settings.
+
+1.  **You will be provided with a verification token.** This token is a unique string that you'll need to add to your DNS records.
+2.  **Create a new `TXT` record in your DNS provider's dashboard.**
+    *   **Host:** `better-auth-token-{your-provider-id}`
```
Bekacru: since host could be different based on the config we should probably mention it here

Bekacru: also can you use the `<Step>` component here instead

docs/content/docs/plugins/sso.mdx
```diff
@@ -704,6 +704,128 @@ mapping: {
...
+3.  **Save the record and wait for it to propagate.** This can take up to 48 hours, but it's usually much faster.
+4.  **Once the record has propagated**, you can submit a validation request
+
+### Domain validation request
+
+Once you have configured your domain, you can use your `auth` instance to submit a validation request.
+This request will either result in a rejection (could not prove your ownership over the domain)
+or if the verification is successful, your SSO provider domain will be marked as verified.
+
+<Tabs items={["client", "server"]}>
```
Bekacru: don't use tab...see the docs here as an example https://github.com/better-auth/better-auth/blob/d34877ade1ca64d6e5892284c5812b319647f719/docs/content/docs/plugins/organization.mdx#L133

docs/content/docs/plugins/sso.mdx
```diff
@@ -726,7 +848,17 @@ The plugin requires additional fields in the `ssoProvider` table to store the pr
         { name: "samlConfig", type: "string", description: "The SAML configuration (JSON string)", isRequired: false },
         { name: "userId", type: "string", description: "The user ID", isRequired: true, references: { model: "user", field: "id" } },
         { name: "providerId", type: "string", description: "The provider ID. Used to identify a provider and to generate a redirect URL.", isRequired: true, isUnique: true },
-        { name: "organizationId", type: "string", description: "The organization Id. If provider is linked to an organization.", isRequired: false },
+        { name: "organizationId", type: "string", description: "The organization Id. If provider is linked to an organization.", isRequired: false }
+    ]}
+/>
+
+### Schema for domain verification
```
Bekacru: instead of a title 

`If you have enabled domain verification:`

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,269 @@
+import type { Verification } from "better-auth";
+import {
+	APIError,
+	createAuthEndpoint,
+	sessionMiddleware,
+} from "better-auth/api";
+import { generateRandomString } from "better-auth/crypto";
+import dns from "dns/promises";
```
himself65: This imports DNS from `node:`, which will not be supported in Cloudflare and Edge Runtime. 
We can change it to dynamic lazy import

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,269 @@
...
+			if (!provider) {
+				throw new APIError("NOT_FOUND", {
+					message: "Provider not found",
+					code: "PROVIDER_NOT_FOUND",
+				});
+			}
+
+			const userId = ctx.context.session.user.id;
+
+			let isOrgMember: boolean = true;
```
Bekacru: no need to do `boolean` true already infers boolean

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,269 @@
...
+					where: [
+						{ field: "userId", value: userId },
+						{ field: "organizationId", value: provider.organizationId },
+					],
+				});
+
+				isOrgMember = membershipsCount > 0;
+			}
+
+			if (!(provider.userId === userId && isOrgMember)) {
```
Bekacru:
```suggestion
			if (provider.userId !== userId ||  !isOrgMember) {
```

Bekacru: just to make it clearer

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,269 @@
...
+						field: "identifier",
+						value: options.domainVerification?.tokenPrefix
+							? `${options.domainVerification?.tokenPrefix}-${provider.providerId}`
+							: `better-auth-token-${provider.providerId}`,
+					},
+					{ field: "expiresAt", value: new Date(), operator: "lt" },
+				],
+			});
+
+			if (verification) {
```
Bekacru: I don't think we want to throw an error if it's still valid. Either we need to pass the token back or generate a new one

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,269 @@
...
+				throw new APIError("CONFLICT", {
+					message: "Current verification token is still valid",
+					code: "TOKEN_FOUND",
+				});
+			}
+
+			let domainVerificationToken: string | undefined;
+
+			if (options?.domainVerification?.enabled) {
```
Bekacru: shouldn't we be check if it's enabled top of the endpoint before doing anything?

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,269 @@
...
+							: `better-auth-token-${provider.providerId}`,
+						createdAt: new Date(),
+						updatedAt: new Date(),
+						value: domainVerificationToken,
+						expiresAt: new Date(Date.now() + 3600 * 24 * 7 * 1000), // 1 week
+					},
+				});
+			}
+
+			return Response.json(
```
Bekacru: why are we still using `Response` here?

Bekacru: this will not be inferred properly

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,269 @@
...
+					where: [
+						{ field: "userId", value: userId },
+						{ field: "organizationId", value: provider.organizationId },
+					],
+				});
+
+				isOrgMember = membershipsCount > 0;
+			}
+
+			if (!(provider.userId === userId && isOrgMember)) {
```
Bekacru: change this similar to the above suggestion

docs/content/docs/plugins/sso.mdx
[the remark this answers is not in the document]
jonathansamines: Both the implementation and docs are correct.

packages/sso/src/types.ts
```diff
@@ -205,4 +212,23 @@ export interface SSOOptions {
...
+	 *
+	 * When this option is enabled, new SSO providers will require the associated domain to be verified by the owner
+	 * prior to allowing sign-ins.
+	 */
+	domainVerification?: {
+		/**
+		 * Enables or disables the domain verification feature
+		 */
+		enabled?: boolean;
+
```
Bekacru: remove empty space
```suggestion
```

packages/sso/src/routes/domain-verification.ts
```diff
@@ -0,0 +1,270 @@
...
+
+			await ctx.context.adapter.update<SSOProvider<SSOOptions>>({
+				model: "ssoProvider",
+				where: [{ field: "providerId", value: provider.providerId }],
+				update: {
+					domainVerified: true,
+				},
+			});
+
+			return new Response(null, { status: 204 });
```
Bekacru: not sure why we're returning Response object here?

packages/sso/src/routes/domain-verification.ts
jonathansamines: I think you are right here, generally we don't need this check here as these endpoints are conditionally enabled already.

packages/sso/src/routes/sso.ts
```diff
@@ -839,7 +884,7 @@ export const signInSSO = (options?: SSOOptions) => {
 						return res.id;
 					});
 			}
-			let provider: SSOProvider | null = null;
+			let provider: SSOProvider<SSOOptions> | null = null;
```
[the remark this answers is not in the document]
jonathansamines: Not introduced here, so will defer for later

packages/sso/src/routes/domain-verification.ts
jonathansamines: Oh, I think it got this wrong, I thought the problem was the specific use of "node:" prefix at first, but sounds like we want to avoid unconditional use of node APIs altogether. Let me fix this.

PR #5912, opened 2025-11-11
fix(oidc-provider): `oidc_login_prompt` not cleared after login

himself65:
Closes https://github.com/better-auth/better-auth/issues/4594

---
## Summary by cubic
Fixes the OIDC flow so the oidc_login_prompt cookie is cleared after login, preventing repeated consent prompts. The cookie is now removed once a valid session is detected.

- **Bug Fixes**
  - Validate session with getSessionFromCtx and ctx.context.newSession instead of parsing Set-Cookie headers.
  - Clear oidc_login_prompt when a session exists in both mcp and oidc-provider plugins via post-login middleware.
  - Added a unit test that verifies the cookie is cleared after sign-up/login.

<sup>Written for commit c173710244cda86aa36301a44e809007d1c57279. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/index.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/index.ts
@@ -251,6 +251,24 @@ export const oidcProvider = (options: OIDCOptions) => {
 			],
 			after: [
 				{
+					matcher() {
+						return true;
+					},
+					handler: createAuthMiddleware(async (ctx) => {
+						// clean up the prompt cookie after a successful login
+						const hasNewSession = !!ctx.context.newSession;
+						const cookie = await ctx.getSignedCookie(
+							"oidc_login_prompt",
+							ctx.context.secret,
+						);
+						if (hasNewSession && cookie) {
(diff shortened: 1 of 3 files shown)


packages/better-auth/src/plugins/oidc-provider/index.ts
```diff
@@ -259,27 +258,19 @@ export const oidcProvider = (options: OIDCOptions) => {
...
-						const sessionToken = sessionCookie?.split(".")[0]!;
-						if (!sessionToken) {
+						if (!cookie) {
 							return;
 						}
-						const session =
-							await ctx.context.internalAdapter.findSession(sessionToken);
+
+						const session = await getSessionFromCtx(ctx);
 						if (!session) {
```
Bekacru: This wouldn't check if we're about to set session...it's not about getting current session but the session we're about to set...you can rather check `ctx.context.newSession` instead

Bekacru: But sometimes it may not be set that's why we're getting the set cookie header and fetching the session

PR #5913, opened 2025-11-11
fix(sso): use the internalAdapter for user queries to avoid skipping database hooks

hartbit:
I noticed my user database hook wasn't being called when using the SSO plugin. I saw it came from the fact that it uses the adapter directly instead of the internal adapter when creating users and account. This fixes that.

---
## Summary by cubic
Switch SSO user/account creation to internalAdapter so user database hooks and side effects run during SAML logins. Aligns SSO behavior with standard auth with no API changes.

- **Bug Fixes**
  - Use internalAdapter.createUser/createAccount in callbackSSOSAML and acsEndpoint.
  - Keep existing fields and trustEmailVerified logic; timestamps handled by internal adapter.

<sup>Written for commit 2de4e0a8f2e552b2c99d38bca3da17f460ffaec2. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/sso/src/routes/sso.ts
+++ b/packages/sso/src/routes/sso.ts
@@ -1584,15 +1584,10 @@ export const callbackSSOSAML = (options?: SSOOptions) => {
 			if (existingUser) {
 				user = existingUser;
 			} else {
-				user = await ctx.context.adapter.create({
-					model: "user",
-					data: {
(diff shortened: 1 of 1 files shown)


hartbit:
@Bekacru I'd love this to be merged into the `main` branch. Should I open a separate PR for that ?

PR #5916, opened 2025-11-11
chore: support stateless without any configuration

Bekacru:
## Summary by cubic
Enable stateless session mode automatically when no database is configured, with secure cookie/JWE and OAuth state defaults. Updates the demo and docs to support zero-config stateless usage.

- **New Features**
  - Auto-enable stateless when `database` is not provided.
  - Default `session.cookieCache` to enabled, JWE, and refresh.
  - Default `oauthConfig.storeStateStrategy` to `cookie`.
  - Demo switches to `better-auth/minimal` and removes extra config.
  - Removed the "no database" memory adapter warning.

- **Docs**
  - Updated stateless guide: zero-config behavior, manual setup, and session invalidation via `cookieCache.version`.
  - Added “Signup disabled” error page with guidance to use database hooks in stateless mode.
  - Linked the new error in the errors index.

<sup>Written for commit 7e5ff951e9d18c2c66a75ed92de722569aa0aefc. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/guides/optimizing-for-performance.mdx
+++ b/docs/content/docs/guides/optimizing-for-performance.mdx
@@ -207,5 +207,5 @@ Simply import from `better-auth/minimal` instead of `better-auth`:
 <Callout type="warning">
   **Limitations:**
   - Direct database connections are not supported (you must use an adapter)
-  - Migrations are not supported (use `better-auth` if you need migration support)
+  - Built-in migrations are not supported. Use external migration tools (or use `better-auth` if you need built-in migration support)
 </Callout>
--- a/packages/better-auth/src/api/middlewares/oauth.ts
+++ b/packages/better-auth/src/api/middlewares/oauth.ts
@@ -20,6 +20,6 @@ const {
 	 * @internal This is unsafe to be used directly. Use setOAuthState instead.
 	 */
 	set: setOAuthState,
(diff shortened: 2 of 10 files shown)


packages/better-auth/src/db/adapter-base.ts
```diff
@@ -18,9 +18,6 @@ export async function getBaseAdapter(
 			acc[key] = [];
 			return acc;
 		}, {});
-		logger.warn(
```
himself65, line 21: We overlook the fact that users are using a memory database. probably a hidden issue in the future

packages/better-auth/src/context/create-context.ts
```diff
@@ -30,6 +30,28 @@ export async function createAuthContext(
 	options: BetterAuthOptions,
 	getDatabaseType: (database: BetterAuthOptions["database"]) => string,
 ): Promise<AuthContext> {
+	//set default options for stateless mode
+	if (!options.database) {
```
himself65, line 35: nit: what if we use defu here?

packages/better-auth/src/db/adapter-base.ts
Bekacru: I think it's fine since we now enable cookie cache by default

PR #5917, opened 2025-11-11
feat(cli): use vite to load auth config when we detect a vite project

shreyassanthu77:
When using frameworks that build on top of vite, some imports may not be resolved correctly when using jiti (like path aliases, virtual modules etc)

so this pr uses vite to try and resolve a config file and if one exists we use that with all the plugins enabled to load the auth config file so that will make sure all the imports are resolved properly and any vite specific logic need not be stubbed out just to use the better auth cli

## Notes
- splits out the config file path resolution logic from the actual loading of the file
- added vite as a dev dependency for type checking
- loads the actual vite module from the user's project to load the config
- falls back to jiti if vite is not found, package json is not found or if vite could not load a vite config 

## Few Considerations
- the change currently expects a few APIs from node and vite to be available for it to work
    1. we expect the runnerImport API to be available in vite but may not be available in older vite versions, needs better version resolution which currently is not implemented
    1. the runnerImport API used to load the config is marked experimental tho the api didn't change since 6.2
- some frameworks like astro and solid start wrap vite so we might need special handling for them (or leave it as is and just let it fallback to jiti?)
- no tests have been implemented for the new vite specific code

---
## Summary by cubic
Use Vite v6–v7 to load the auth config in Vite projects so aliases and plugin-based imports resolve correctly, with a safe fallback to c12/Jiti when Vite isn’t available or fails.

- **New Features**
  - Auto-detect Vite and load the auth config with the project’s Vite (plugins on).
  - Resolve TS path aliases, virtual modules, and framework imports.
  - Fallback to c12/Jiti if Vite is missing, runnerImport is unavailable, or Vite load fails.

- **Refactors**
  - Split config path resolution from loading; search paths relative to cwd.
  - Stronger validation and errors: try/catch around Vite load, export checks after both loaders, unified “server-only” guidance, and deduped messages.
  - Add Vite dev dep and optional peer; load user’s Vite at runtime via package.json resolution.
  - Tests cover Vite alias resolution and fallback to c12/Jiti when no Vite config is present.

<sup>Written for commit e8d02516309694674a3357db59f5b557eb20cbfc. Summary will update on new commits.</sup>


himself65 (commented):
I think the idea is cool. Probably, we can remove Jiti? @Bekacru

shreyassanthu77:
not everything uses vite though

- for example astro and solid start wrap vite and don't directly expose a way to load the vite config
- some projects might be using turbo

jiti acts as a nice fallback to support cases where we don't really support the specific env the user has

but yeah probably a good idea to do some framework specific config resolution for vite based ones (should be pretty easy to do i'd assume but it then becomes a maintenance decision to support multiple frameworks and/or versions etc)

packages/cli/src/utils/get-config.ts
```diff
@@ -162,6 +166,105 @@
...
+			}
+			logger.error(
+				`[#better-auth]: Couldn't read your auth config in ${configPath}. Make sure to default export your auth instance or to export as a variable named auth.`,
+			);
+			process.exit(1);
+		}
+	}
+
+	for (const possiblePath of possiblePaths) {
+		if (existsSync(possiblePath)) {
```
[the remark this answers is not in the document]
shreyassanthu77: seems like a good fix probably should just apply it? @himself65 

the cli is expected to run in the project dir anyway but just a good thing for the sake of consistency imo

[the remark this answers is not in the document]
himself65: y

shreyassanthu77:
I think we can drop the special case svelte kit handling

been using basically this same code (well with an older better auth version but still) for about ~~a year~~ 6 months :) and seems to be working well

himself65:
> I think we can drop the special case svelte kit handling
> 
> been using basically this same code (well with an older better auth version but still) for about ~a year~ 6 months :) and seems to be working well

What's that?

shreyassanthu77:
I just mean the current version has special handling for svelte specific virtual modules but since the vite implementation already takes care of that we can remove the sveltekit specific stuff from the code.

if you are asking about the second half of my comment, I just meant I was using this vite specific code from my fork for last few months and it's working fine

shreyassanthu77:
image
on line 124

himself65:
> I just mean the current version has special handling for svelte specific virtual modules but since the vite implementation already takes care of that we can remove the sveltekit specific stuff from the code.
> 
> if you are asking about the second half of my comment, I just meant I was using this vite specific code from my fork for last few months and it's working fine

Oh yeah, I think we can remove that path hack if we are using vite

shreyassanthu77:
uhmm can you do it please 😅 I have a lot of stuff to do this week

or merge this and I'll make a new pr next week when I'm free that is also fine

packages/cli/src/utils/get-config.ts
```diff
@@ -162,6 +165,111 @@ const isDefaultExport = (
...
+	logger.error(
+		"Couldn't find a configuration file. Add a `auth.ts` file to your project or pass the path to the configuration file using the `--config` flag.",
+	);
+	return null;
+}
+
+async function loadConfigWithVite(cwd: string, resolvedConfigPath: string) {
+	const packageJsonPath = path.join(cwd, "package.json");
+	if (existsSync(packageJsonPath)) {
+		const require = createRequire(packageJsonPath);
```
himself65, line 246: What's the reason we don't use import('vite') here?

himself65:
I just realized that this might affect https://github.com/better-auth/better-auth/pull/6120, https://github.com/better-auth/better-auth/pull/6113, since we use vitest for testing

packages/cli/src/utils/get-config.ts
shreyassanthu77: From what I understand dynamic import will check from the dir where the cli is installed in so if say for example you installed better auth cli globally it might find a different vite version that the user has globally installed

could be wrong though

shreyassanthu77:
> I just realized that this might affect https://github.com/better-auth/better-auth/pull/6120, https://github.com/better-auth/better-auth/pull/6113, since we use vitest for testing

why is that?

shreyassanthu77:
Heyy I'm back and ready to work on anything to make it easy for y'all to merge

packages/cli/src/utils/get-config.ts
shreyassanthu77: @himself65 so I did some digging, you can technically just point to the vite main file directly with a dynamic import but we have to resolve the vite main file which is non trivial cuz we have to locate the vite package which may or may not be specified in the current project package json so we have to treverse the file system and find where vite is and then we have to look for exports in vite's package json then get to vite

but then if the user is using say deno (I am that user :)) there's no node_modules dir sometimes which is another giant rabbit hole if we want to support that

createRequire does all that for us and is usually implemented in native code so it will be much faster and more reliable anyway

shreyassanthu77:
> I just realized that this might affect #6120, #6113, since we use vitest for testing

looked at the prs and i didn't get how this would affect them? we don't even need vite as a dependecy strictly speaking, we load it from the user's codebase anyway. i just added vite so we can have type checking for the vite apis used but we can just shim them if needed not a big deal

unless i misunderstood what you mean (which could very well be the case) I don't see how this pr would affect it

himself65 (commented):
I think we need to add vite to peer deps

shreyassanthu77:
@himself65 regarding #6373 you want me to try the esbuild thing?

Also recently started an astro project and ran into this issue myself because my pr only works with a pure vite projects lol. I looked into it and astro wraps vite with their own cli without a real public api... which is kinda sad

shreyassanthu77:
heyyy @himself65 sorry for the ping but is there anything i can do to get this pr merged? happy to fix anything or change things

Bekacru:
@shreyassanthu77 can we get ci and merge conflict fixed?

shreyassanthu77:
> @shreyassanthu77 can we get ci and merge conflict fixed?

Sure gimme a couple hours

shreyassanthu77:
@Bekacru rebased and fixed all the issues and also added tests

gustavovalverde:
Closing in favor of #8520, #9032, and #9477 (all merged). The Vite-loader path here depends on the experimental `vite.runnerImport` API; the merged trio chose a more conservative approach (`get-tsconfig` + Babel AST + c12 v4 `resolveModule`). SvelteKit and Cloudflare virtual-module cases are partially handled by `addSvelteKitEnvModules` and `addCloudflareModules` stubs already on main.

Issue #4913, opened 2025-09-25
Anonymous user plugin failing due to Stripe plugin


focux:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

When the anonymous user plugin is enabled, the signup flow attempts to create a Stripe customer while the user is still anonymous. Stripe rejects the request due to an invalid email format generated for the temporary anonymous user, causing customer creation to fail.
```
Error
Error: Invalid email address: temp-tzhd5yviw70h7avlkdyutu6xno2kqisi@http://localhost:3000
```

## Steps to Reproduce

 1. Enable anonymous and stripe plugin.
 2. Start session as an anonymous user.

### Current vs. Expected behavior

I expect the customer no to be created for anonymous users

### What version of Better Auth are you using?

1.3.7

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.1.0: Fri Sep 19 19:14:26 PDT 2025; root:xnu-12377.40.77.505.1~4/RELEASE_ARM64_T6000",
    "release": "25.1.0",
    "cpuCount": 10,
    "cpuModel": "Apple M1 Max",
    "totalMemory": "32.00 GB",
    "freeMemory": "0.11 GB"
  },
  "node": {
    "version": "v22.16.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.2"
  },
  "frameworks": null,
  "databases": [
    {
      "name": "postgres",
      "version": "^3.4.7"
    },
    {
      "name": "drizzle",
      "version": "1.0.0-beta.1-84d9a79"
    }
  ],
  "betterAuth": {
    "version": "^1.3.7",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
 plugins: [
    anonymousClient(),
    stripeClient({
      subscription: true,
    }),
  ],
});
```

antonvishal:
Hey @focux my use case is that I wanted to allow anonymous users to purchase in my marketplace like a one time thing, they dont have to authenticate. So what would you recommend, This PR contradicts that right?

Bekacru:
The user should still be able to checkout with stripe even if they are anonymous. You can change the `emailDomainName` in your anonymous plugin config to fix this

PR #5922, opened 2025-11-12
feat(anonymous): allow customizing random email generator

bytaesu:
This PR includes the following changes:

- Refactor anon plugin file structure
- Change default email format to fix emails containing the protocol
- Add `generateRandomEmail ` option to use a custom email generation function
- Update related documentation

Related https://github.com/better-auth/better-auth/issues/4913

---
## Summary by cubic
Update the anonymous plugin to fix invalid default emails and add a custom email generator. Default emails are now temp@{id}.com, and you can use generateRandomEmail to define your own format.

- **New Features**
  - Added generateRandomEmail option (sync or async). Returns an error if the email is invalid.
  - Updated defaults: temp@{id}.com by default; with emailDomainName, emails are temp-{id}@<domain>.

- **Migration**
  - If you relied on baseURL-derived emails, set emailDomainName (e.g., "example.com") to keep that pattern.
  - When using generateRandomEmail, ensure it returns a unique, valid email; emailDomainName is ignored.

<sup>Written for commit 7f90ce56097e7954e9e87cb6d417b94198d3ac42. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/anonymous/anon.test.ts
+++ b/packages/better-auth/src/plugins/anonymous/anon.test.ts
@@ -10,7 +10,6 @@ import {
 	it,
 	vi,
 } from "vitest";
-import { createAuthClient } from "../../client";
 import { signJWT } from "../../crypto";
 import { getTestInstance } from "../../test-utils/test-instance";
 import { DEFAULT_SECRET } from "../../utils/constants";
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/anonymous/anon.test.ts
```diff
@@ -178,6 +178,124 @@ describe("anonymous", async () => {
 		expect(res.data?.user.name).toBe("i-am-anonymous");
 	});
 
+	it("should work with generateRandomEmail", async () => {
+		const testHeaders = new Headers();
+		const { customFetchImpl, sessionSetter } = await getTestInstance({
```
himself65: there is auth and client in the return value

himself65 (changes requested):

packages/better-auth/src/plugins/anonymous/anon.test.ts
bytaesu: this is cool, thanks!

PR #5933, opened 2025-11-12
feat(email-otp): allow signup with additional fields

### LINKED ISSUE #1557: Ability to set "name" and any "additionalFields" when signing-up through authClient.signIn.emailOtp

_No response_

Currently `authClient.signIn.emailOtp` can only handle `email`, and `otp` in the object. 

`authClient.signIn.emailOtp` could either inject in the same payload the additionalFields and "name" which is a already existing field in the Users table, or add a extra object in the function to carry those values.

This allows to not call `authClient.updateUser` right after OTP verification.

### Additional context

_No response_

jonathansamines:
---
## Summary by cubic
Enable signup via Email OTP with extra user fields during auto-registration using a new signupData object. The endpoint supports core properties (e.g., name, image) and configured additionalFields, with validation, docs, and tests updated.

- **New Features**
  - /sign-in/email-otp accepts signupData with name, image, and any configured additionalFields during auto-signup.
  - OpenAPI documents signupData; request body includes email, otp, and signupData.
  - Validates additional fields via parseUserInput.
  - Added tests for core properties and an additional field; docs show the new usage.

<sup>Written for commit 24d91deb52ee78b54f86a77a6d1e636cf611fb59. Summary will update automatically on new commits.</sup>

Closes #1557


packages/better-auth/src/plugins/email-otp/index.ts
```diff
@@ -775,18 +776,46 @@ export const emailOTP = (options: EmailOTPOptions) => {
...
-						}),
-					}),
+						z.any(),
+					),
 					metadata: {
+						$Infer: {
+							body: {} as {
+								name?: string;
+								email: string;
+								otp: string;
```
Bekacru: how would they pass additional fields beyond `name`?

jonathansamines: To be honest, this is where I wanted to do some inference magic to derive all other additional fields but couldn't figure out how without asking them to declare additional fields for this plugin as well.

Right now, they would need to ignore any ts error or I could update this annotation to allow any `Record<string, any>`.

Bekacru: >  need to ignore any ts error

this is a really bad expectation 👀

Bekacru: instead of taking a `name`, you can take a `data` object that includes core properties like `name`, `image`...plus a `Record<string, any>` for any additional fields

jonathansamines: Got it, this is the kind of feedback I was looking for, it looks like we support a combination of pattern for additional fields (e.g additional fields combined or separate from the core properties) for various methods, and wasn't sure which one to follow. Will implement your suggestion.

packages/better-auth/src/plugins/email-otp/email-otp.test.ts
```diff
@@ -97,6 +106,65 @@ describe("email-otp", async () => {
...
+			type: "sign-in",
+		});
+		const headers = new Headers();
+		const newUser = await client.signIn.emailOtp(
+			{
+				email: testUser3.email,
+				otp,
+				name: "the user",
+				// @ts-expect-error testing additional fields
+				firstName: "First",
```
himself65: why this is ts error

jonathansamines: Should be fixed now that signup properties are hold in a separate property.

docs/content/docs/plugins/email-otp.mdx
```diff
@@ -127,13 +127,18 @@ type signInEmailOTP = {
     /**
      * OTP sent to the email. 
      */
-    otp: string = "123456"
+    otp: string = "123456",
+    /**
+     * Additional signup properties
+    */
+    signupData?: Record<string, any> = { name: "Daniel", image: "https://user-profiles.com/users/daniel" }
```
himself65, line 134: Can we use additionalData?

himself65 (changes requested):

docs/content/docs/plugins/email-otp.mdx
jonathansamines: Could be, but I have two arguments in favor of `signupData`:

- It makes it clear that these attributes would only be considered during signup
- It is not specific to additional attributes (e.g we can pass other core properties such as `name` or `image`)

jonathansamines: Actually, are you asking whether or not we support additional data? We do support that and it is documented I think, but please let me know if that is what you meant.

mhamid3d:
This solution still returns back a hard-coded user schema. Why can't we use `parseUserOutput` as is done in `sign-up.ts`?

https://github.com/better-auth/better-auth/blob/b2c52eae63c88fe9bd36431e27bfe841effac8f6/packages/better-auth/src/api/routes/sign-up.ts#L313-L319

Also was expecting `signupData` to just be of type `User`, if it's running through `parseUserInput` anyways?

PR #5942, opened 2025-11-12
feat(oidc-provider): add RP-Initiated Logout endpoint

himself65:


packages/better-auth/src/plugins/oidc-provider/endpoints/end-session.ts
```diff
@@ -0,0 +1,177 @@
...
+					},
+				);
+			}
+
+			// Otherwise, return a success response
+			return ctx.json(
+				{ message: "Logout successful" },
+				{
+					status: 200,
+				},
```
Bekacru, line 174: status doesn't work

Issue #5549, opened 2025-10-24
PG/Drizzle: useNumberId add IDENTITY option


jose-biescas:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

When using the `useNumberId` option for PG/Drizzle, the `id` column is generated as:
```typescript
id: serial("id").primaryKey()
```
These two sources explain why `serial` shouldn't be used for newer applications (unless supporting Postgresql 10 or older)
And that `IDENTITY` is the new "SQL-standard identity column feature"

Sources:
https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_serial
https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-SERIAL

I understand having to support older versions as well, but versions before Postgresql 10 are no longer supported and version 10 had its' last updated almost 3 years ago: https://www.postgresql.org/support/versioning/

### Describe the solution you'd like

Newer applications should be using the IDENTITY column

Change:
```typescript
if (databaseType === "pg") {
   id = `serial("id").primaryKey()`;
} 
```

To:
```typescript
if (databaseType === "pg") {
   id = `integer("id").generatedAlwaysAsIdentity().primaryKey()`;
} 
```

Provide a `useLegacySerial` option if people would rather use serial columns.

### Describe alternatives you've considered

Remove serial entirely, and to accommodate either wanting Always or ByDefault
`useNumberId` can optionally take an object that describes the options

```typescript
useNumberId: {
   byDefault: true // default is false
}
```

PR #5943, opened 2025-11-12
fix: use `identity` instead of `serial` for pg schema

ping-maxwell:
closes:  https://github.com/better-auth/better-auth/issues/5549
linear: https://linear.app/better-auth/issue/ENG-661/fix-use-identity-over-serial-for-pg-number-ids

Converts `pg` with `generateId === 'serial'` to use integer with `IDENTITY` instead of `serial`

---
## Summary by cubic
Switches Postgres numeric IDs from serial to identity across migrations and CLI schema generation, aligning with modern PG behavior and fixing #5549 and ENG-661. Generated SQL now uses integer GENERATED BY DEFAULT AS IDENTITY.

- **Bug Fixes**
  - Migration generator: use sql`integer GENERATED BY DEFAULT AS IDENTITY` for Postgres when number IDs are enabled; adjust column builder to avoid double identity config.
  - CLI (Drizzle): replace serial("id") with integer("id").generatedByDefaultAsIdentity().primaryKey(); remove serial import.
  - Tests: add Postgres identity test and update snapshots to ensure no SERIAL appears.

<sup>Written for commit 45beb7a1373ec7e74915cdcdbad775f050948e4f. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/email-otp/index.ts
+++ b/packages/better-auth/src/plugins/email-otp/index.ts
@@ -30,7 +30,7 @@ export interface EmailOTPOptions {
 			otp: string;
 			type: "sign-in" | "email-verification" | "forget-password";
 		},
-		request?: Request | undefined,
+		ctx?: GenericEndpointContext | undefined,
 	) => Promise<void>;
 	/**
 	 * Length of the OTP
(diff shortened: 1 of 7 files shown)


packages/better-auth/src/db/get-migration-schema.test.ts
```diff
@@ -238,3 +238,62 @@ describe.runIf(isPostgresAvailable)(
...
+					database: {
+						generateId: "serial",
+					},
+				},
+			};
+
+			const { compileMigrations } = await getMigrations(config);
+			const migrations = await compileMigrations();
+
+			console.log(migrations);
```
himself65: remove console.log

ping-maxwell: ah yes ty

PR #5986, opened 2025-11-14
feat(oauth-proxy): OAuth proxy that works even in stateless mode

bytaesu:
Context -> https://discord.com/channels/1288403910284935179/1438696563936333946

### Tasks

- [x] Add support for stateless mode
- [x] Add test cases
- [ ] Test with a real app

---
## Summary by cubic
Enabled the OAuth proxy to work in stateless mode by reading and decrypting the OAuth state from a cookie. This removes database state lookups while keeping the existing database strategy intact.

- **New Features**
  - Added cookie-based state handling (storeStateStrategy: "cookie") with symmetric decryption using the app secret.
  - Skips internalAdapter.findVerificationValue in stateless mode; only sets skipStateCookieCheck for the database strategy.
  - Added tests for stateless and secondary storage flows, validating redirect to oauth-proxy-callback with encrypted cookies.

<sup>Written for commit d8400318a857fb7f5257ea5dbf5575efaebd309c. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/oauth-proxy/oauth-proxy.test.ts
+++ b/packages/better-auth/src/plugins/oauth-proxy/oauth-proxy.test.ts
@@ -53,9 +53,9 @@ afterEach(() => {
 
 afterAll(() => server.close());
 
-describe("oauth-proxy", async () => {
+describe("oauth-proxy", () => {
 	it("should redirect to proxy url", async () => {
-		const { client, cookieSetter } = await getTestInstance({
+		const { client } = await getTestInstance({
 			plugins: [
 				oAuthProxy({
 					currentURL: "http://preview-localhost:3000",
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/oauth-proxy/index.ts
```diff
@@ -268,23 +268,52 @@ export const oAuthProxy = (opts?: OAuthProxyOptions | undefined) => {
...
-							parsedState = parseJSON<{ callbackURL?: string }>(data.value);
-						} catch {
-							parsedState = undefined;
+
+						if (ctx.context.oauthConfig.storeStateStrategy === "cookie") {
+							//
+							// Cookie strategy (stateless mode)
+							//
+							const stateCookie = ctx.context.createAuthCookie("oauth_state");
+							const encryptedData = ctx.getCookie(stateCookie.name);
```
billyjanitsch, line 278: Can you clarify where this cookie would come from? This code is being run on the production origin, but the auth request might have been initiated from some other origin, so how would the cookie have been set here?

bytaesu: Right, this code isn’t working right now, so I’m exploring other options. Stateless strategy relies on cookies, and I’m not sure how to handle cross-origin properly without a database.

bytaesu: I fixed the previous `state_mismatch` issue by using the internalAdapter(database).. do you have any ideas on how to handle this without it??

billyjanitsch: Did you consider a solution like https://github.com/better-auth/better-auth/issues/1819#issuecomment-3003038278 where the data gets encrypted into the `state` query param sent to the OAuth provider from the non-production env, and then decrypted by the production env when the provider sends it back?

bytaesu: I was hesitant to put sensitive data in the URL, even if encrypted. After discussing with the team, we think it’s worth trying. I’ll close this PR for now and try this out soon. Thanks for the suggestion 😄

PR #5988, opened 2025-11-14
Use "user id" as _id in mongo adapter if defined when creating user

MathiasWP:
IMO when you define a user id when using the MongoAdapter, then the id of user user should be the same on the user object that is created. To achieve this the _id field has to be the same, see https://www.mongodb.com/docs/drivers/node/current/crud/insert/#a-note-about-_id

---
## Summary by cubic
Use the provided user id as the MongoDB _id when creating a user. Falls back to the insertedId if no id is provided, keeping user ids consistent across the app and database.

<sup>Written for commit bbc37f14442eb8387f4a1fb772f3cd4c24f4d84d. Summary will update automatically on new commits.</sup>


MathiasWP:
closed in favour of https://github.com/better-auth/better-auth/pull/6033, #6033 has the correct solution

PR #5993, opened 2025-11-14
fix(oidc-provider): implement proper OIDC prompt parameter handling

himself65:
Fixes: https://github.com/better-auth/better-auth/issues/4594

---
## Summary by cubic
Implements spec-compliant OIDC prompt and max_age handling, fixes silent login/consent behavior, and prevents unintended redirects after OIDC flows.

- **Bug Fixes**
  - prompt=none: return login_required when unauthenticated; consent_required when consent is needed; otherwise proceed without UI.
  - max_age: treat 0 as prompt=login and require reauth when session age exceeds max_age.
  - Clear oidc_login_prompt cookie after resuming flow so later normal logins don’t redirect to the OIDC client.

- **New Features**
  - Added parsePrompt utility with validation (none cannot be combined) and tests.
  - Introduced OIDCProviderError/InvalidRequest, broadened prompt type, and resumed flow after login by stripping prompt=login before re-authorizing.

<sup>Written for commit cd8effe79ce7486dad2196e234d55e0b89ff018d. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/types.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/types.ts
@@ -358,13 +358,6 @@ export interface CodeVerificationValue {
 	 */
 	requireConsent: boolean;
 	/**
-	 * Whether the user needs to be logged in to use the client.
-	 *
-	 * If this is true, then the user will be redirected to the login page.
-	 * If the user is already logged in, they will be redirected to the redirect URI.
-	 */
-	requireLogin: boolean;
-	/**
 	 * The state parameter from the request
 	 *
 	 * If the prompt is set to `consent`, then the state


packages/better-auth/src/plugins/oidc-provider/error.ts
```diff
@@ -0,0 +1,13 @@
+import { APIError } from "better-call";
+
+export class OIDCProviderError extends APIError {}
+
+export class InvalidRequest extends OIDCProviderError {
+	constructor(error_description: string, error_detail?: string) {
```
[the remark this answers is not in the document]
Bekacru: @himself65

packages/better-auth/src/plugins/oidc-provider/types.ts
```diff
@@ -354,6 +357,13 @@ export interface CodeVerificationValue {
 	 * with the actual code.
 	 */
 	requireConsent: boolean;
+	/**
+	 * Whether the user needs to be logged in to use the client.
+	 *
+	 * If this is true, then the user will be redirected to the login page.
+	 * If the user is already logged in, they will be redirected to the redirect URI.
+	 */
+	requireLogin: boolean;
```
Bekacru: not sure why this option exist? if prompt is set to login or the user doesn't have a session we require login. What does this option change?

packages/better-auth/src/plugins/oidc-provider/index.ts
Bekacru, line 269: I think we should try to get the session from `ctx.context.newSession`, and if it's null, then fall back to fetching the session using the sessionToken

packages/better-auth/src/plugins/oidc-provider/authorize.ts
```diff
@@ -58,6 +59,19 @@ export async function authorize(
...
 	if (!session) {
+		// Handle prompt=none per OIDC spec - must return error instead of redirecting
+		const query = ctx.query as AuthorizationQuery;
+		const promptSet = parsePrompt(query.prompt ?? "");
+		if (promptSet.has("none") && query.redirect_uri) {
+			return handleRedirect(
+				formatErrorURL(
+					query.redirect_uri,
+					"login_required",
+					"Authentication required but prompt is none",
```
Bekacru, line 70: I think we should still redirect to login if prompt is still none. This will be a breaking change otherwise

Bekacru (changes requested):

packages/better-auth/src/plugins/oidc-provider/types.ts
himself65: When the prompt is like `contest login`, FYI, the prompt could be a list of items, not just contest or login

packages/better-auth/src/plugins/oidc-provider/authorize.ts
himself65: If so, that would violate the OIDC spec. Should we?

packages/better-auth/src/plugins/oidc-provider/types.ts
himself65: but yeah, I think I don't use it now... I will remove it

packages/better-auth/src/plugins/oidc-provider/index.ts
himself65: yeah, I see... Maybe this is why there's bug, it could be stateless session

packages/better-auth/src/plugins/mcp/authorize.ts
himself65, line 1: sync from oidc-provider code

himself65:
I think mcp plugin is not following the odic spec, let me do that in another PR

packages/better-auth/src/plugins/oidc-provider/authorize.ts
```diff
@@ -205,9 +217,40 @@ export async function authorize(
...
+	if (promptSet.has("none")) {
+		// If consent is required, return consent_required error
+		if (!skipConsentForTrustedClient && !hasAlreadyConsented) {
+			return handleRedirect(
+				formatErrorURL(
+					query.redirect_uri,
+					"consent_required",
+					"Consent required but prompt is none",
+				),
+			);
```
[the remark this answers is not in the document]
himself65: > OAuth 2.0 state value. REQUIRED if the Authorization Request included the state parameter. Set to the value received from the Client.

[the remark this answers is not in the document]
himself65: I dont want to fix this in this PR. seems too large

himself65:
I'm going to merge this and test it in https://github.com/better-auth/better-auth/pull/6083

PR #5995, opened 2025-11-14
chore: improve zod imports tree shaking

sabaturgay:
Adjust Zod imports to avoid bundling locale files and reduce the esbuild bundle size by ~260kb.
With the current import { z } from 'zod', the Zod locales end up included in the bundle, which unnecessarily increases the output size.

---
## Summary by cubic
Switched all Zod imports to namespace form to improve tree-shaking and exclude locale files, reducing the esbuild bundle by ~260 KB.

- **Refactors**
  - Replaced `import { z } from 'zod'` with `import * as z from 'zod'` across core, auth routes, plugins, docs, and scripts; added a Biome rule to enforce this import style.
  - No API or runtime behavior changes.

<sup>Written for commit 309204a7a8966ad0e6d0827268cab5f91649243a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -14,6 +14,19 @@
   "linter": {
     "enabled": true,
     "rules": {
+      "style": {
+        "noRestrictedImports": {
+          "options": {
+            "paths": {
+              "zod": {
+                "importNames": ["z"],
+                "message": "Use `import * as z from \"zod\"` instead of `import { z }`."
+              }
+            }
+          },
+          "level": "error"
+        }
+      },
       "recommended": false,
       "suspicious": {
         "noImplicitAnyLet": "warn",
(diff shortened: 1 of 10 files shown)


sabaturgay:
> do you have benchmark for this?

I do use import * as z from 'zod' in all of my codebase
with better-auth (ignore zod/v3 it is related to another package)
Screenshot 2025-11-14 at 23 10 49

when i remove better-auth 
Screenshot 2025-11-14 at 23 13 59

Also here is another issue from the zod repo, others also refer that: https://github.com/colinhacks/zod/issues/4369#issuecomment-2976883065

himself65:
I updated the biome linter so we can use' import * as z' forever!

sabaturgay:
⚠️ Regression

Between **v1.4.4 → v1.4.7**, a regression was introduced again for this issue, also adds kysely even though I do use `better-auth/minimal` with drizzle
Screenshot 2025-12-24 at 18 16 55

PR #5998, opened 2025-11-15
fix(haveibeenpwned): check for limited set of paths

Bekacru:
## Summary by cubic
Limit Have I Been Pwned checks to specific auth routes to avoid unnecessary lookups and improve performance. By default, checks run only on /sign-up/email, /change-password, and /reset-password.

- **New Features**
  - Added paths option to configure which routes trigger the password breach check.
  - Keeps customPasswordCompromisedMessage support.

- **Bug Fixes**
  - Run checks only when the current request path matches the whitelist (via getCurrentAuthContext).
  - Prevents unintended checks during unrelated hashing, reducing external calls and false positives.

<sup>Written for commit 0ccc4946c3e88fc4431f4f7fee78738055c4bf2e. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/haveibeenpwned/index.ts
+++ b/packages/better-auth/src/plugins/haveibeenpwned/index.ts
@@ -6,95 +6,95 @@ import { betterFetch } from "@better-fetch/fetch";
 import { APIError } from "../../api";
 
 const ERROR_CODES = defineErrorCodes({
-  PASSWORD_COMPROMISED:
-    "The password you entered has been compromised. Please choose a different password.",
+	PASSWORD_COMPROMISED:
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/haveibeenpwned/index.ts
```diff
@@ -57,17 +58,33 @@ async function checkPasswordCompromise(
 
 export interface HaveIBeenPwnedOptions {
 	customPasswordCompromisedMessage?: string | undefined;
+	/**
+	 * Paths to check for password
+	 *
+	 * by default "/sign-up/email", "/change-password", "/reset-pasword"
```
himself65: @default

PR #5999, opened 2025-11-15
Add support for additionalFields to API key

gricha:
Hey folks! The problem I'm trying to solve with this PR is an ability to scope API keys to organizations. I happen to use both organization and API key plugin and I wanted to make sure that there's a way to associate the two easily. This allows me to add an `organizationId` field directly to API key in a similar manner that I would extend organization/user/session. I considered using metadata field, but I figured that this may end up being a field I want to index on, and there's prior art to how these extensions are being made.

---
## Summary by cubic
Adds support for custom additional fields on API keys, enabling first-class fields like organizationId with validation, typing, and persistence. This avoids overloading metadata and allows indexing on these fields.

- **New Features**
  - apiKey plugin now accepts schema.apikey.additionalFields to define extra DB fields.
  - create/update API routes validate, persist, and return these fields.
  - Added helpers to build/parse additional field schemas for request payloads.
  - mergeSchema now merges additionalFields into table fields.
  - Types updated to infer additional fields on ApiKey and plugin option schema.
  - Tests cover creating and updating an API key with organizationId.

<sup>Written for commit a32d4b66960e361fb412679a7aad5d884746e619. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/api-key/api-key.test.ts
+++ b/packages/better-auth/src/plugins/api-key/api-key.test.ts
@@ -753,8 +753,8 @@ describe("api-key", async () => {
 		});
 
 		expect(apiKey).not.toBeNull();
-		expect(apiKey.metadata.test).toBeDefined();
-		expect(apiKey.metadata.test).toEqual(metadata.test);
+		expect(apiKey.metadata?.test).toBeDefined();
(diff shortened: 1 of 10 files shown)


thomasmol:
can i ask why this was closed? i would like to see this feature

PR #6013, opened 2025-11-16
feat: support storing account data in a cookie

Bekacru:
- Support storing OAuth access and refresh tokens directly in a cookie so that `getAccessToken` and other IDP-token methods work without requiring a database
- move `oauthConfig` to `accounts` options instead
- intorduce new option `storeAccountCookie` for enabling account cookie data

- [x] Implementation
- [x] Test
- [x] Docs


Blitz2145:
Tested this PR locally and it works well, thanks for the help.

Issue #6018, opened 2025-11-16
Bug: hex string expected, got unpadded hex of length 253


firdausai:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. I have fork this repo and modified the demo nextjs directory to demonstrate the error [this repo](https://github.com/firdausai/better-auth/tree/demo/oauth-encryption-bug). I just added a component so its easier to test this.
2. Login using Google OAuth
3. After being redirected to the dashboard on the demo, scroll to the bottom and click the first button with the name `1. Add google drive scope`. You can modify this to any other scope that require the access token, but this scope is what I am using on my app
4. Click the second button with the name `2. Call google drive api (list file)`. This should result in a success (expected)
5. Go to `auth.ts` and change `encryptOAuthTokens` from `False` to `True`
6. Wait for the access token to be expired
7. Click the second button with the name `2. Call google drive api (list file)` again. This should result in an error

```
Error fetching OAuth tokens: [Error [APIError]: Failed to get a valid access token] {
  status: 'BAD_REQUEST',
  body: [Object],
  headers: {},
  statusCode: 400,
  [cause]: Error: hex string expected, got unpadded hex of length 253
      at symmetricDecrypt (../../packages/better-auth/dist/crypto-DZW3hrq1.mjs:64:32)
      at async (../../packages/better-auth/dist/api-fDFKThbG.mjs:664:43)
      at async (../../packages/better-auth/dist/api-fDFKThbG.mjs:2870:21)
      at async GET (app/api/token/route.ts:12:29)
    62 | const symmetricDecrypt = async ({ key, data }) => {
    63 | 	const keyAsBytes = await createHash("SHA-256").digest(key);
  > 64 | 	const dataAsBytes = hexToBytes(data);
       | 	                              ^
    65 | 	const chacha = managedNonce(xchacha20poly1305)(new Uint8Array(keyAsBytes));
    66 | 	return new TextDecoder().decode(chacha.decrypt(dataAsBytes));
    67 | };
}
```

### Current vs. Expected behavior

Current:
When changing the `encryptOAuthTokens` from `False` to `True`, `getAccessToken` function will call the `decryptOAuthToken` function and the if statement will evaluate to `True` even though the token is not encrypted. Hence, the error.

Expected Behavior:
Token should be decrypted only if they are encrypted in the first place.

### What version of Better Auth are you using?

1.3.34

### System info

```bash
{
  "system": {
    "platform": "darwin",
[...]

PR #6019, opened 2025-11-16
Fix: hex string expected, got unpadded hex of length 253

firdausai:
Propose solution to bug reported in [#6018](https://github.com/better-auth/better-auth/issues/6018)

This PR fixes a bug where toggling `encryptOAuthTokens` from `false` to `true` causes a `"hex string expected, got unpadded hex of length 253"` error when refreshing expired tokens. The issue occurred because `decryptOAuthToken` attempted to decrypt plain text tokens that were stored before encryption was enabled. The fix adds a prefix-based check (`enc:`) to distinguish encrypted tokens from plain text ones. Encrypted tokens are now prefixed with `enc:` when stored with setTokenUtil, and `decryptOAuthToken` only attempts decryption if the prefix is present.

---
## Summary by cubic
Prefix encrypted OAuth tokens with 'enc:' and decrypt only when the prefix is present. Prevents errors when handling plaintext or mixed tokens.

- **Bug Fixes**
  - Stops decryption attempts on plaintext tokens, avoiding failures when encryption is disabled or tokens vary.

- **Migration**
  - setTokenUtil is now async; update callers to await its result.

<sup>Written for commit e293957d651494ad554bd79fab7d735b1accbd14. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/oauth2/utils.ts
+++ b/packages/better-auth/src/oauth2/utils.ts
@@ -3,24 +3,25 @@ import { symmetricDecrypt, symmetricEncrypt } from "../crypto";
 
 export function decryptOAuthToken(token: string, ctx: AuthContext) {
 	if (!token) return token;
-	if (ctx.options.account?.encryptOAuthTokens) {
+	if (ctx.options.account?.encryptOAuthTokens && token.startsWith('enc:')) {
 		return symmetricDecrypt({
 			key: ctx.secret,
-			data: token,
+			data: token.slice(4),
 		});
 	}
 	return token;
 }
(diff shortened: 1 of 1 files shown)


Bekacru:
intersting but this would break all currently encrypted tokens without `enc:` prefix

firdausai:
@Bekacru You're right. Would adding a CLI command to either encrypt/decrypt the token make more sense? Or maybe this is something that the user has to do outside of better-auth?

himself65:

himself65:

PR #6031, opened 2025-11-17
fix: resolve SESSION_IS_NOT_FRESH error with cookieCache

### LINKED ISSUE #6028: SESSION_IS_NOT_FRESH thrown, when cookieCache is enabled

### To Reproduce

1. Enable cookieCache
2. Go to webpage /api/auth/passkey/generate-register-options?authenticatorAttachment=platform
3. You'll get SESSION_IS_NOT_FRESH
4. Wait beyond CookieCache.maxAge (In my case, 5 minutes is enough)
5. Refresh page and you'll get proper results - not SESSION_IS_NOT_FRESH

I expect to not have freshness error right after logging in especially when default freshness is set to 1 day.

when cookieCache is enabled, updatedAt is in the wrong format (2025-11-16T18:07:42.283Z instead of 1763327105228) which results in a bad comparison which results in error.

1.4.0 beta

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:40 PDT 2025;",
    "release": "24.6.0",
    "cpuCount": 16,
    "cpuModel": "redacted",
    "totalMemory": "128.00 GB",
    "freeMemory": "20.62 GB"
  },
  "node": {
    "version": "v24.7.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.5.1"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "0.1.0",
    "config": null
  }
}
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
  cookieCache: {
        enabled: true,
        maxAge: 5 * 60, // 5 min
        strategy: 'jwe',
  }
});
```

### Additional context

https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/api/routes/session.ts#L548

GautamBytes:
Closes #6028

### Description
Fixes a bug where `freshSessionMiddleware` incorrectly threw `SESSION_IS_NOT_FRESH` when `cookieCache` was enabled. 

The issue was caused by the `updatedAt` field being retrieved from the cookie as an ISO string, while the freshness logic used `.valueOf()` directly, which failed to perform the correct math against `Date.now()`.

### The Fix
Updated `freshSessionMiddleware` in `session.ts` to explicitly wrap `updatedAt` (or `createdAt`) in `new Date(...).getTime()`. This ensures a numeric timestamp is always used for comparison, regardless of whether the input is a Date object or an ISO string.

---
## Summary by cubic
Fixes the session freshness check to stop false SESSION_IS_NOT_FRESH errors when cookieCache is enabled. Normalizes cookie dates to numeric timestamps before comparison.

- **Bug Fixes**
  - In freshSessionMiddleware, compute lastUpdated with new Date(updatedAt || createdAt).getTime() so ISO strings from cookies compare correctly against Date.now().

<sup>Written for commit e3498d79bb4920e43788107c62e0507cda16c799. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/session.ts
+++ b/packages/better-auth/src/api/routes/session.ts
@@ -544,8 +544,9 @@ export const freshSessionMiddleware = createAuthMiddleware(async (ctx) => {
 		};
 	}
 	const freshAge = ctx.context.sessionConfig.freshAge;
-	const lastUpdated =
-		session.session.updatedAt?.valueOf() || session.session.createdAt.valueOf();
+	const lastUpdated = new Date(
+		session.session.updatedAt || session.session.createdAt,
+	).getTime();
 	const now = Date.now();
 	const isFresh = now - lastUpdated < freshAge * 1000;
 	if (!isFresh) {


ping-maxwell:
This is good. we should also consider finding the the code which grabs the cookie cache in the first place and making that convert date strings. Might be a little complicated for this PR though, so maybe for now this is fine.

ping-maxwell (approved):

GautamBytes:
> we should also consider finding the the code which grabs the cookie cache in the first place and making that convert date strings.

will consider that as my next task then 🙌

PR #6045, opened 2025-11-17
feat: move react-start to tanstack-start

Bekacru:
## Summary by cubic
Fix the TanStack setCookie import by switching to "@tanstack/react-start/server" in packages/better-auth/src/integrations/tanstack-start.ts. Prevents runtime errors and ensures Set-Cookie headers are applied.

- **Bug Fixes**
  - Cast the SQLite DB to SqliteDatabase in the Kysely adapter to avoid type issues.

- **Dependencies**
  - Expose "./tanstack-start" (replacing "./react-start"), rename the plugin to tanstackStartCookies, add "@tanstack/react-start-server", and update docs to use "@tanstack/start/react-router" with "better-auth/tanstack-start".

<sup>Written for commit e36ba536e4d3beb6ca5208ff37dbaa999732382a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5384,11 +5384,8 @@ packages:
     cpu: [x64]
     os: [win32]
 
-  '@rolldown/pluginutils@1.0.0-beta.45':
-    resolution: {integrity: sha512-Le9ulGCrD8ggInzWw/k2J8QcbPz7eGIOWqfJ2L+1R0Opm7n6J37s2hiDWlh6LJN0Lk9L5sUzMvRHKW7UxBZsQA==}
-
-  '@rolldown/pluginutils@1.0.0-beta.46':
-    resolution: {integrity: sha512-xMNwJo/pHkEP/mhNVnW+zUiJDle6/hxrwO0mfSJuEVRbBfgrJFuUSRoZx/nYUw5pCjrysl9OkNXCkAdih8GCnA==}
+  '@rolldown/pluginutils@1.0.0-beta.50':
+    resolution: {integrity: sha512-5e76wQiQVeL1ICOZVUg4LSOVYg9jyhGCin+icYozhsUzM+fHE7kddi1bdiE0jwVqTfkjba3jUFbEkoC9WkdvyA==}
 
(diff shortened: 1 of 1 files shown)


packages/better-auth/package.json
```diff
@@ -493,6 +493,7 @@
     "@prisma/client": "^5.22.0",
     "@sveltejs/kit": "^2.37.1",
     "@tanstack/react-start": "^1.131.3",
+    "@tanstack/start": "^1.120.20",
```
himself65: Should we add this to peer? I think we missed many peer deps in the package.json

Bekacru: yeah but optional right?

Bekacru: also we should probably remove react-start

himself65: see https://github.com/better-auth/better-auth/blob/9571c96e409939cf332e4a395c2325f586ca9156/packages/expo/package.json#L71-L81

Bekacru: yeah yeah

himself65 (approved):

packages/better-auth/src/integrations/tanstack-start.ts
```diff
@@ -20,7 +20,7 @@ export const reactStartCookies = () => {
 							const setCookies = returned?.get("set-cookie");
 							if (!setCookies) return;
 							const parsed = parseSetCookieHeader(setCookies);
-							const { setCookie } = await import("@tanstack/start-server-core");
+							const { setCookie } = await import("@tanstack/start/server");
```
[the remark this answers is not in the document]
austinm911: the import should be "@tanstack/react-start/server"

https://github.com/TanStack/router/blob/main/packages/react-start/src/server.tsx

andybarron:
hi! original issue reporter here. ([discord link](https://discord.com/channels/1288403910284935179/1439906110252253314))

unfortunately, i couldn't test this PR locally (pnpm git+monorepo issue). but i confirmed via `pnpm patch` that updating the import from `@tanstack/start-server-core` -> `@tanstack/react-start/server` fixes the vite crash i was getting before :+1:

PR #6048, opened 2025-11-17
fix: support `WantAuthnRequestsSigned="true"` for sso SAML SPs

jzylks:
The `sso` package cannot currently authenticate with a SAML IDP that requires signed authentication requests. There are a couple problems:

- If the SP metadata includes `AuthnRequestsSigned="true"`, samlify will throw an error when constructing the ServiceProvider object because the private key (and optional private key password) are not included in the samlConfig object. Only the metadata and `allowCreate` are currently set.
- After adding the private key for signing, the request to the IDP cannot be validated because the `RelayState` is added to the URL after the request has been signed.

This adds the private key and private key password to the constructor for the samlify ServiceProvider object so that it is able to sign the request, and moves the relayState to that same constructor so that it is included in the signature.

---
## Summary by cubic
Adds support for SAML IDPs that require signed Authn requests (WantAuthnRequestsSigned="true"). Authn requests are now signed by the SP and include RelayState in the signature so IDP validation succeeds.

- **Bug Fixes**
  - Pass privateKey and privateKeyPass from spMetadata to the samlify ServiceProvider.
  - Set relayState in the ServiceProvider constructor and stop appending RelayState after signing.
  - Added tests covering IDPs that require signed Authn requests.

<sup>Written for commit 735f5fd7009da949689fd19c929a2f9cf9bb7bf6. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/sso/src/routes/sso.ts
+++ b/packages/sso/src/routes/sso.ts
@@ -984,6 +984,9 @@ export const signInSSO = (options?: SSOOptions) => {
 				}
 				const sp = saml.ServiceProvider({
 					metadata: parsedSamlConfig.spMetadata.metadata,
+					privateKey: parsedSamlConfig.spMetadata.privateKey,
+					privateKeyPass: parsedSamlConfig.spMetadata.privateKeyPass,
+					relayState: body.callbackURL,
 					allowCreate: true,
 				});
 
@@ -1004,9 +1007,7 @@ export const signInSSO = (options?: SSOOptions) => {
 					});
 				}
 				return ctx.json({
-					url: `${loginRequest.context}&RelayState=${encodeURIComponent(
-						body.callbackURL,
-					)}`,
+					url: loginRequest.context,
 					redirect: true,
 				});
 			}
--- a/packages/sso/src/routes/sso.ts
+++ b/packages/sso/src/routes/sso.ts
@@ -984,16 +984,16 @@ export const signInSSO = (options?: SSOOptions) => {
 				}
 				const sp = parsedSamlConfig.spMetadata.metadata
 					? saml.ServiceProvider({
-						metadata: parsedSamlConfig.spMetadata.metadata,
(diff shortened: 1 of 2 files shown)


Bekacru (dismissed):

packages/sso/src/routes/sso.ts
```diff
@@ -968,10 +968,18 @@ export const signInSSO = (options?: SSOOptions) => {
...
-				});
+				const sp = parsedSamlConfig.spMetadata.metadata
+					? saml.ServiceProvider({
+						metadata: parsedSamlConfig.spMetadata.metadata,
+						privateKey: parsedSamlConfig.spMetadata.privateKey,
+						privateKeyPass: parsedSamlConfig.spMetadata.privateKeyPass,
+						relayState: body.callbackURL,
+						allowCreate: true,
+					})
+					: saml.ServiceProvider({
```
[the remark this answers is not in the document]
jzylks: Does it ever make sense to call `saml.ServiceProvider` without metadata? Before my update, the metadata was always included. I think maybe that suggestion from Copilot wasn't necessary and [61e61f4](https://github.com/better-auth/better-auth/pull/6048/commits/61e61f43639aa47be6165b9a25ceb728ef4010d5) should be reverted.

packages/sso/src/saml.test.ts
```diff
@@ -585,6 +691,89 @@ describe("SAML SSO with defaultSSO array", async () => {
 	});
 });
 
+describe("SAML SSO with defaultSSO array and IDP requires signed Authn requests", async () => {
+	const data = {
```
himself65, line 695: why don't we use test utils here?

jzylks: I just copied the tests from above and changed out the metadata. I don't have any objection to changing the tests if there's a better or more preferred way to write them.

himself65: I see. I'm gonna update test case later

PR #6056, opened 2025-11-18
feat: support custom id field names

### LINKED ISSUE #5807: Map custom id field names

It's our best practice to use ids such as tableNameId instead of just id.

Better auth allows for renaming all fields except the id fields.

We hack around this by having duplicated drizzle schema definitions just to satisfy better auth.

```ts
  user: {
    fields: {
      id: text("user_id")
    }
  }
});
```

If we could map id field also like this.

Right now we include a seperately defined db schema

like so:
```ts
export const baseSchema = {
  user: sqliteTable("user", {
    id: text("user_id")
      .primaryKey()
      .$defaultFn(() => `u-${ulid()}`),
    name: text("name").notNull(),
    email: text("email").notNull().unique(),
    emailVerified: integer("email_verified", { mode: "boolean" }).notNull(),
    image: text("image"),
    role: text("role").default("user"),
    banned: integer("banned", { mode: "boolean" }).default(false),
    banReason: text("ban_reason"),
    banExpires: integer("ban_expires", { mode: "timestamp" }),
    createdAt: integer("created_at", { mode: "timestamp" }).notNull(),
    updatedAt: integer("updated_at", { mode: "timestamp" }).notNull(),
  }),
```

that is only used for the better auth config. So essentially we have duplicate schema definitions for better auth related tables.

### Additional context

_No response_

Ridhim-RR:
Closes #5807
### Summary
Adds support for configuring custom database column names for primary ID fields (e.g., user_id) while preserving that mapping at runtime. This ensures that generated IDs are correctly inserted into the configured columns, while callers still use the canonical id field.

### Changes

- Updated types and options in init-options.ts to support custom ID field names.
- Added ID entries to the runtime schema generated by getAuthTables for user, session, and account schemas.
- Modified Adapter Factory to preserve fieldName when injecting generated ID attributes, ensuring transformInput inserts generated IDs into the configured DB column.
- Extended Zod schemas to include ID fields in userSchema, accountSchema, and sessionSchema.

**Note**:  Support for custom Id columns is currently implemented only  for the user, session and account tables.

---
## Summary by cubic
Adds support for custom database column names for primary ID fields on user, session, and account tables. Generated IDs are written to the mapped columns, while callers continue to use the canonical id field.

- **New Features**
  - Configure custom ID columns via options.user/session/account.fields.id.
  - Runtime schema includes id.fieldName and the adapter writes generated IDs to that column.
  - Added id to user, session, and account Zod schemas; tests verify mapping and adapter behavior.

- **Migration**
  - If you use non-standard ID columns, set fields.id in your user/session/account options. No changes needed otherwise.

<sup>Written for commit 12b720acb5a5faf052b8fde8428dbf5d112851d1. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/adapters/adapter-factory/adapter-factory.id-field.test.ts
+++ b/packages/better-auth/src/adapters/adapter-factory/adapter-factory.id-field.test.ts
@@ -1,86 +1,93 @@
 // packages/better-auth/src/adapters/adapter-factory/adapter-factory.id-field.test.ts
-import { describe, it, expect } from "vitest";
+import { describe, expect, it } from "vitest";
 import { getAuthTables } from "../../db/get-tables";
+import { createAdapterFactory } from ".";
+import { initGetDefaultFieldName } from "./get-default-field-name";
 import { initGetDefaultModelName } from "./get-default-model-name";
 import { initGetFieldName } from "./get-field-name";
-import { initGetDefaultFieldName } from "./get-default-field-name";
-import { createAdapterFactory } from ".";
 
 describe("custom id field mapping", () => {
-  it("getAuthTables should include fields.id.fieldName from options", () => {
-    const options = {
-      user: { fields: { id: "userId" } },
-      session: { fields: { id: "sessionId" } },
-      account: { fields: { id: "accountId" } },
-    } as any;
-    const schema = getAuthTables(options);
-    expect(schema?.user?.fields.id?.fieldName).toBe("userId");
-    expect(schema?.session?.fields.id?.fieldName).toBe("sessionId");
-    expect(schema?.account?.fields.id?.fieldName).toBe("accountId");
-  });
+	it("getAuthTables should include fields.id.fieldName from options", () => {
+		const options = {
+			user: { fields: { id: "userId" } },
+			session: { fields: { id: "sessionId" } },
+			account: { fields: { id: "accountId" } },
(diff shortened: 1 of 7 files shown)


himself65 (commented):
/cc @Bekacru @ping-maxwell

himself65 (changes requested):
This is a breaking change, and we may need to modify more under-the-hood code beyond this PR.

ping-maxwell (changes requested):
In Mysql we do not always have a way to return data of what was inserted into the DB as mysql does not return such data (unlike other DBs).

Because of this, in adapters like Drizzle (and possibily others) there is logic to take the `id` from what was used to insert and try to fetch the row and use that as the response, just make sure that this PR also updates such case if the `id` is changed to a different field name to use the right field while fetching.

Ridhim-RR:
> In Mysql we do not always have a way to return data of what was inserted into the DB as mysql does not return such data (unlike other DBs).
> 
> Because of this, in adapters like Drizzle (and possibily others) there is logic to take the `id` from what was used to insert and try to fetch the row and use that as the response, just make sure that this PR also updates such case if the `id` is changed to a different field name to use the right field while fetching.

ok, ,Let me look in this as its been a month since I have raised this pr , will analyse the whole flow again and then share the flow to implement this

PR #6060, opened 2025-11-18
docs: clarify database mapping

### LINKED ISSUE #6059: Clarify field renaming

The BetterAuth lib allows to rename user, session, or plugin fields (for instance in my case, I want fields in my database and in my User objects to be snake_case instead of the default BetterAuth camelCase)

so I've added renaming like eg:

```ts
  user: {
    modelName: "usersTable",
    fields: {
      emailVerified: "email_verified",
      createdAt: "created_at",
      updatedAt: "updated_at",
    },
```

but when I get a user in a BetterAuth middleware, the fields are not yet renamed:

```json
contextUser={
  "name": "Gouach Admin",
  "email": "admin@gouach.com",
  "emailVerified": true,
  "image": null,
  "createdAt": "2025-11-18T10:23:12.377Z",
  "updatedAt": "2025-11-18T10:23:12.377Z",
  "role": "admin",
  "banned": null,
  "banReason": null,
  "banExpires": null,
  "description": "System administrator",
  "status": "active",
  "last_login_at": null,
  "partner_id": null,
  "user_role": "internal:admin",
  "id": "internal-user-0"
}
```

is that expected?

1. Perhaps it should be clearly documented somewhere (how / when fields are renamed)
2. This also means (I guess?) that we can't "swap" fields, eg if I want that the BetterAuth admin plugin "role" be called "better_auth_role", and that I define my own (user) "role" field, it's going to conflict?``

I'd like that once I decide to rename fields, everytime I get a "User" object from BetterAuth, or everytime I'm sending a User object to some BetterAuth method or API, it is in the "renamed field" format, not the "default format"

Create some helper methods to do the conversion, and add "logs" to know what format a data is given in at each method, but this is not clear. At the very least, it should be documented in the BetterAuth docs

### Additional context

_No response_

okisdev:
This PR adds more details information to clarify database mapping.

This PR closes #6059.

---
## Summary by cubic
Clarified how field mapping works: the fields config only changes database column names, while your code always uses the original camelCase fields. Added examples, an info callout, and an FAQ entry to reduce confusion when using snake_case schemas.

<sup>Written for commit 976b2886affbce611561c34d15cf868281a7855a. Summary will update automatically on new commits.</sup>


himself65 (approved):
Yes, this is a nice catch. But please double check if we need this @Bekacru

bytaesu (commented):
This is standard field-mapping behavior, and the existing callout already explains it sufficiently IMO, so I'm going to close this PR.

PR #6085, opened 2025-11-19
feat(generic-oauth): provide pre configured provider helpers

Paola3stefania:
---
## Summary by cubic
Adds provider helpers to generic-oauth for Okta, Auth0, Microsoft Entra ID, Slack, and Keycloak for quick setup. Also sets emailVerified to false by default across built-in providers for safer, consistent handling.

- **New Features**
  - okta, auth0, microsoftEntraId, slack, keycloak helpers return pre-configured GenericOAuthConfig (authorization/token/userInfo or discovery URLs, default scopes).
  - Supports overrides: scopes, redirectURI, pkce, disableImplicitSignUp, disableSignUp, overrideUserInfo; getUserInfo maps profile fields and sets emailVerified when available.

- **Bug Fixes**
  - Updated facebook, gitlab, kick, linear, microsoft-entra-id, polar, and roblox providers to default emailVerified to false (or use provided claim) for consistency.

<sup>Written for commit 15eb9d26a8954e1410db525a2d14130b9839dc1a. Summary will update automatically on new commits.</sup>

https://github.com/user-attachments/assets/66c8213d-54af-432c-a0da-70d5a66dd103

### DIFF
--- a/packages/better-auth/src/plugins/generic-oauth/index.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/index.ts
@@ -662,13 +662,19 @@ export const genericOAuth = (options: GenericOAuthOptions) => {
 							}&error_description=${ctx.query.error_description}`,
 						);
 					}
+					const providerId = ctx.params?.providerId;
+					if (!providerId) {
+						throw new APIError("BAD_REQUEST", {
+							message: "Provider ID is required",
+						});
+					}
 					const provider = options.config.find(
-						(p) => p.providerId === ctx.params.providerId,
+						(p) => p.providerId === providerId,
 					);
 
 					if (!provider) {
 						throw new APIError("BAD_REQUEST", {
-							message: `No config found for provider ${ctx.params.providerId}`,
+							message: `No config found for provider ${providerId}`,
 						});
 					}
 					let tokens: OAuth2Tokens | undefined = undefined;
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/generic-oauth/providers/github.ts
```diff
@@ -0,0 +1,170 @@
+import type { OAuth2Tokens, OAuth2UserInfo } from "@better-auth/core/oauth2";
+import { betterFetch } from "@better-fetch/fetch";
+import type { GenericOAuthConfig } from "../index";
+
+/**
+ * Provider definition based on Auth.js/NextAuth.js
```
Bekacru: we don't need to mention auth.js

packages/better-auth/src/plugins/generic-oauth/providers/keycloak.ts
```diff
@@ -0,0 +1,134 @@
...
+ * Provider definition based on Auth.js/NextAuth.js
+ * Source: https://github.com/nextauthjs/next-auth
+ * Adapted for Better Auth's GenericOAuthConfig format
+ */
+
+export interface KeycloakOptions {
+	/** Keycloak OAuth client ID */
+	clientId: string;
+	/** Keycloak OAuth client secret */
+	clientSecret: string;
```
Bekacru: Instead of defining these common options for each provider, can we either extend the default `genericOAuth` config type or introduce another shared type?

Paola3stefania: shared type added 👌

packages/better-auth/src/plugins/generic-oauth/providers/github.ts
Paola3stefania: great then , removing those

PR #6087, opened 2025-11-19
chore: refactor internal endpoints

Bekacru:
## Summary by cubic
Refactored and hardened auth endpoints to reduce information leaks, prevent timing attacks, and improve session handling. Adds security checks and tests for XSS and unsafe callback URLs.

- **Bug Fixes**
  - Password reset: simulate work on missing users, send email in the background, and always return a generic success message.
  - Email verification: generate a token when user is not found and return early if the email is already verified.
  - Sessions: revoke only when owned by the current user; sign-out handles missing tokens and DB errors while always clearing the cookie.
  - Security checks: sanitize error page query params and reject untrusted callback/error/new user URLs.

- **Refactors**
  - Refresh token endpoint returns only explicit token fields (access/refresh and expirations).
  - Added tests for XSS sanitization, password reset non-disclosure, and URL validation.

<sup>Written for commit 99cd38b2016b9c52545e9ef545f45a9c9dfcae2e. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/sign-out.ts
+++ b/packages/better-auth/src/api/routes/sign-out.ts
@@ -1,6 +1,4 @@
 import { createAuthEndpoint } from "@better-auth/core/api";
-import { BASE_ERROR_CODES } from "@better-auth/core/error";
-import { APIError } from "better-call";
 import { deleteSessionCookie } from "../../cookies";
 
 export const signOut = createAuthEndpoint(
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/api/routes/error.test.ts
himself65, line 1: Maybe we should move such a security test into a single test file.

Bekacru: no it's fine. it's specefic to error page

PR #6106, opened 2025-11-19
fix: remove active session requirement for change email verification

### LINKED ISSUE #5978: Change email flow impossible without active session
Currently the change email flow has been build by always requiring a session within the active context. This currently breaks, this flow if I create a change email request from expo. For this app I will never have a session on my betterauth server, since it will be mobile first and the session is managed in local storage on the mobile app.

My proposal is to remove the requirement for a session when using change email. I think this is possible because you already need a session to first request the email change. Second you need to have access to the email account that is already verified. Third it is using a very short-lived jwt token of an hour. I think this already covering most of the security and the extra session check is redundant and we can make it optional.

Bekacru:
closes #5978 

---
## Summary by cubic
Removed the active session requirement for change-email verification so users can confirm email changes without being logged in. Keeps security by rejecting when a logged-in session doesn't match the token email, and creates a session if needed to complete the flow.

- **Bug Fixes**
  - Allow verification without an active session; only deny when a session exists for a different email.
  - Create a session for the user if none exists to complete change-email verification; return an error if creation fails.
  - Send verification email using the token’s user data instead of the session user.

<sup>Written for commit 72e2df4314c8cc8d0c7b269ee25981199bd26daa. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/email-verification.ts
+++ b/packages/better-auth/src/api/routes/email-verification.ts
@@ -309,20 +309,10 @@ export const verifyEmail = createAuthEndpoint(
 			return redirectOnError("user_not_found");
 		}
 		if (parsed.updateTo) {
-			const session = await getSessionFromCtx(ctx);
-			if (!session) {
-				if (ctx.query.callbackURL) {
-					throw ctx.redirect(`${ctx.query.callbackURL}?error=unauthorized`);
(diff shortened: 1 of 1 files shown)


tomdenboon:
Thanks!

PR #6108, opened 2025-11-19
feat(client): introduce `disableSignal` client option

ping-maxwell:
This PR adds the addition of `disableSignal` to prevent specific endpoints from triggering hooks.

Also, `refetch` from hooks return a promise, allowing users to optionally await a given `refetch()` call.

```ts title="auth-client.ts"
// As the second argument
await authClient.updateUser({
    name: "New Name"
}, {
    disableSignal: true
})

// Or within fetchOptions
await authClient.updateUser({
    name: "New Name",
    fetchOptions: {
        disableSignal: true
    }
})
```

---
## Summary by cubic
Adds a disableSignal option to client fetch requests to prevent hook rerenders (e.g., useSession) on specific calls. Updates docs and makes refetch in useAuthQuery awaitable for manual control.

- **New Features**
  - Introduced ClientFetchOption.disableSignal to skip atom signals and hook rerenders on successful requests.
  - Documented usage with examples and manual refetch guidance.
  - useAuthQuery.refetch now returns a Promise.

- **Refactors**
  - Replaced BetterFetchOption with ClientFetchOption across clients (core, better-auth, expo, one-tap, passkey).
  - Updated proxy logic to respect fetch options when triggering atom listeners.

<sup>Written for commit dcae949eb8ffb2db8ca0e156b098b571f3e52317. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/session-api.test.ts
+++ b/packages/better-auth/src/api/routes/session-api.test.ts
@@ -547,11 +547,11 @@ describe("cookie cache", async () => {
 		});
 		expect(session.data?.user.emailVerified).toBe(true);
 		expect(session.data).not.toBeNull();
-		expect(fn).toHaveBeenCalledTimes(3);
+		expect(fn).toHaveBeenCalledTimes(2);
 	});
 
 	it("should reset cache when expires", async () => {
-		expect(fn).toHaveBeenCalledTimes(3);
+		expect(fn).toHaveBeenCalledTimes(2);
 		await client.getSession({
 			fetchOptions: {
 				headers,
@@ -567,7 +567,7 @@ describe("cookie cache", async () => {
 				},
 			},
 		});
-		expect(fn).toHaveBeenCalledTimes(5);
+		expect(fn).toHaveBeenCalledTimes(3);
 		await client.getSession({
 			fetchOptions: {
 				headers,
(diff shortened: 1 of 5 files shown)


packages/better-auth/src/client/query.ts
```diff
@@ -113,18 +118,18 @@ export const useAuthQuery = <T>(
...
 	for (const initAtom of initializedAtom) {
-		initAtom.subscribe(() => {
+		initAtom.subscribe(async () => {
 			if (isServer()) {
 				// On server, don't trigger fetch
 				return;
 			}
 			if (isMounted) {
-				fn();
+				await fn();
```
himself65, line 127: why adding await here?

ping-maxwell: lint CI fails without that.

daveycodez: This way we can now use await refetch from useSession(), really useful for SPA based sign in flow. We can ensure that the refetch completes before navigating to the authenticated page.

ping-maxwell: I can't lie, the dx is pretty nice with this imo 🥲

CleanShot 2025-11-20 at 08 24 52
CleanShot 2025-11-20 at 08 25 30

Bekacru (approved):

PR #6110, opened 2025-11-19
chore: session refetch cooldown for window focus

ping-maxwell:
Every time you focus the browser, we make a get-session request. If you alt+tab between the browser frequently, you would hit your server several times (and thus the DB too) within the span of a couple seconds.

It's unnecessary to get the session each browser refocus within the span of a couple seconds.

---
## Summary by cubic
Add a 5-second cooldown for session refetch on window focus to prevent rapid, repeated get-session calls. This reduces server and DB load when users frequently alt‑tab, while still refetching if the session is missing or the cooldown has expired.

- **New Features**
  - Rate-limit focus-triggered refetch to 5s; bypass if session is null or undefined.
  - Track lastSessionRequest; update on poll and visibilitychange events.
  - Respect offline state: skip refetch when offline unless refetchWhenOffline is true.
  - Add tests covering cooldown, expiration, null/undefined session, and offline behavior.

<sup>Written for commit bb51390a364c322d6400a52ffb44e8977cc70803. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/client/session-refresh.test.ts
+++ b/packages/better-auth/src/client/session-refresh.test.ts
@@ -1,10 +1,17 @@
 // @vitest-environment happy-dom
 
 import { atom } from "nanostores";
-import { describe, expect, it, vi } from "vitest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { getGlobalOnlineManager } from "./online-manager";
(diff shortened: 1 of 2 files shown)


himself65 (approved):

himself65:
Nice catch

PR #6111, opened 2025-11-19
chore: bump better-fetch

dvanmali:
Properly encodes JSON body as a `application/x-www-form-urlencoded` when sent from a client. For example, when `allowedMediaTypes: ["application/x-www-form-urlencoded"]` (meaning json is not accepted), this requires the client to send the `application/x-www-form-urlencoded` header.

Example use case:
```ts
const tokens = await client.oauth2.token({
	grant_type: "client_credentials",
	client_id
	client_secret,
	scope,
	redirect_uri,
}, {
	headers: {
		accept: "application/json",
		'content-type': "application/x-www-form-urlencoded"
	},
});
```

---
## Summary by cubic
Send client request bodies as application/x-www-form-urlencoded when the content-type header is set. This fixes requests to endpoints that require form-encoded payloads (e.g., OAuth2 token).

- **Bug Fixes**
  - Serialize the body with URLSearchParams when content-type is application/x-www-form-urlencoded; otherwise send the original body.

<sup>Written for commit cba5010d09ce675574fb9306b66ac594215dcbc4. Summary will update automatically on new commits.</sup>

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7,8 +7,8 @@ settings:
 catalogs:
   default:
     '@better-fetch/fetch':
-      specifier: 1.1.18
-      version: 1.1.18
+      specifier: 1.1.21
+      version: 1.1.21
     better-call:
       specifier: 1.1.5
       version: 1.1.5
@@ -586,7 +586,7 @@ importers:
         version: 0.3.0
       '@better-fetch/fetch':
         specifier: 'catalog:'
-        version: 1.1.18
+        version: 1.1.21
       '@hookform/resolvers':
         specifier: ^5.2.1
         version: 5.2.1(react-hook-form@7.62.0(react@19.2.1))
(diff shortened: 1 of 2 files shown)


Bekacru:
I’m assuming the only reason the `token` endpoint would ever be called directly from a client is for testing, right? In a normal case, the OAuth client should be the one calling it externally

I don’t think there’s any endpoint callable from the client that expects anything other than JSON. That said, the fix should belong in better-fetch. Better Fetch should handle sending the request body according to the specified `Content-Type`.

dvanmali:
> I’m assuming the only reason the `token` endpoint would ever be called directly from a client is for testing, right? In a normal case, the OAuth client should be the one calling it externally

@Bekacru Not just for testing, if you want an oauth application to utilize the oauth provider plugin then it needs `application/x-www-form-urlencoded`. For example a public web client, can simply use the oauth provider plugin as a server-based client.

Also `/introspect` and `/revoke` endpoints are additionally restricted to `application/x-www-form-urlencoded` type.

You do bring up a good idea though for #4163 to create a different client specifically the oauth application clients.

> I don’t think there’s any endpoint callable from the client that expects anything other than JSON. That said, the fix should belong in better-fetch. Better Fetch should handle sending the request body according to the specified `Content-Type`.

I checked better fetch and it implements it correctly in that package library. I noticed this because the `client.$fetch` passed with the header but `client.oauth2.token` would fail.

PR #6114, opened 2025-11-20
chore: update docu about auth secret

Paola3stefania:
This PR improves the Installation documentation by adding clear requirements for BETTER_AUTH_SECRET.
Several developers ran into cryptic runtime errors caused by using secrets that were too short or low-entropy.

This improves onboarding reliability, aligns with industry standards, and prevents common configuration mistakes.

**Changes**

	•	Added a call-out under Installation → Set Environment Variables clarifying that:
	•	BETTER_AUTH_SECRET must be 32+ characters
	•	Should be generated using a high-entropy source (e.g., openssl rand -base64 32)
	•	Short/weak secrets can lead to signing or encryption errors

image

---
## Summary by cubic
Add runtime validation for BETTER_AUTH_SECRET and update installation docs with a clear callout on length and entropy. This enforces a 32+ char secret, rejects the default in production, and warns on low entropy.

- **New Features**
  - Validate secret on startup: error if default used in production or <32 chars.
  - Warn when the secret appears low-entropy.
  - Updated tests and e2e fixtures to use a valid-length secret.

- **Migration**
  - Set BETTER_AUTH_SECRET to a 32+ character random value; do not use the default in production.
  - Use the generator in the docs or run: npx @better-auth/cli secret.

<sup>Written for commit 66b6c9bf3516ac03b07323a824af69bbb932ba8a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/installation.mdx
+++ b/docs/content/docs/installation.mdx
@@ -26,17 +26,13 @@ Create a `.env` file in the root of your project and add the following environme
 
 1. **Secret Key**
 
-Random value used by the library for encryption and generating hashes. **You can generate one using the button below**.
+A secret value used for encryption and hashing. It must be at least 32 characters and generated with high entropy. Click the button below to generate one. You can also use `openssl rand -base64 32` to generate one.
 
 ```txt title=".env"
 BETTER_AUTH_SECRET=
 ```
 
 <GenerateSecret />
-
-<Callout type="warn">
-Your `BETTER_AUTH_SECRET` **must be at least 32 characters long** and generated with **high entropy**. Using a short or weak secret can cause **cryptic runtime errors**, invalid signatures, or weakened security.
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/context/create-context.ts
```diff
@@ -24,6 +24,57 @@ import {
...
+
+/**
+ * Validates that the secret meets minimum security requirements.
+ * Throws BetterAuthError if the secret is invalid.
+ * Skips validation for DEFAULT_SECRET in test environments only.
+ */
+function validateSecret(
+	secret: string,
+	logger: ReturnType<typeof createLogger>,
+	isDefaultSecret: boolean,
```
Bekacru: nit:

instead of taking this as an argument, better to check if it's default secret inside the function itself

Issue #5025, opened 2025-10-01
OIDC "consent" (hasAlreadyConsented) does not take into consideration if the scope have changed


jippi:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Request scope `openid profile` (user consents) 
2. Request scope `openid profile email` and the consent screen is bypassed entirely

The `scopes` field is not updated to reflect `email` added, but the `oauth access tokens` table for the new token now has `openid profile email` instead - basically allowing escalating permissions without further consent

### Current vs. Expected behavior

Expanding consent should require a new consent prompt

Today this does not happen

### What version of Better Auth are you using?

1.3.24

### System info

```bash
N/A
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});
```

jippi:
Specifically, https://github.com/better-auth/better-auth/blob/19f39cb400d1b7afe925528bb58bccb8c3f077c3/packages/better-auth/src/plugins/oidc-provider/authorize.ts#L191C8-L191C27 should include `scope` in the check to see if consent has happened - or compare the consent value in DB and new scopes and return false if not the same

PR #6116, opened 2025-11-20
fix(oidc-provider): redirect to consent when scope changed

himself65:
Fixes: https://github.com/better-auth/better-auth/issues/5025

---
## Summary by cubic
Redirects to the OIDC consent page when requested scopes exceed prior consent, and skips consent when scopes are unchanged. Adds unit tests for initial consent, unchanged scopes, and scope expansion to verify redirect behavior.

<sup>Written for commit 7684f5f47a0164f43d275632221984e1490c3a39. Summary will update automatically on new commits.</sup>

### DIFF
--- a/test/unit/oidc/scope-consent.spec.ts
+++ b/test/unit/oidc/scope-consent.spec.ts
@@ -75,7 +75,7 @@ describe("oidc scope consent", async () => {
 
 	it("should require new consent when scopes are expanded", async () => {
 		// 1. Create a user and login
-		const user = await auth.api.signUpEmail({
+		await auth.api.signUpEmail({
 			body: {
 				email: "test@example.com",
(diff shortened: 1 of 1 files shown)


test/unit/oidc/scope-consent.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+			return handler(req, res);
+		});
+		await new Promise<void>((resolve) => {
+			server.listen(0, () => {
+				resolve();
+			});
+		});
+		url = `https://localhost:${(server.address() as any).port}`;
+		// @ts-expect-error
+		auth.options.baseURL = url;
```
himself65, line 61: @Bekacru, do you think we have a better way to set the baseURL after the server is mounted?

PR #6122, opened 2025-11-20
feat(jwt): add JWT verification endpoint and refactor verification logic

himself65:
## Summary by cubic
Adds a server-only /verify-jwt endpoint and a shared verifyJWT helper to validate JWTs against JWKS and return the payload. OIDC Provider now uses this helper, removing duplicate logic and making verification consistent.

- **New Features**
  - POST /verify-jwt accepts { token, issuer? } and returns { payload } where payload can be null.
  - verifyJWT exported from the JWT plugin for programmatic use.
  - Optional issuer override; falls back to plugin options or baseURL.

- **Refactors**
  - OIDC Provider replaces its local JWKS verification with jwt.verifyJWT.
  - Removed redundant jose import and verification code.
  - Centralizes JWT verification in the JWT plugin.

<sup>Written for commit f3eb18406648c45891ff07461e064b44af852770. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/index.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/index.ts
@@ -1620,7 +1620,11 @@ export const oidcProvider = (options: OIDCOptions) => {
 								);
 								if (verified) {
 									validatedUserId = verified.sub;
-									validatedClientId = verified.aud;
+									validatedClientId = verified.aud
+										? typeof verified.aud === "string"
+											? verified.aud
+											: verified.aud[0]!
+										: null;
 								}
 							} else {
 								// For HS256 tokens, we need the client_id to verify


packages/better-auth/src/plugins/jwt/verify.ts
```diff
@@ -0,0 +1,64 @@
...
+import type { JwtOptions } from "./types";
+
+/**
+ * Verify a JWT token using the JWKS public keys
+ * Returns the payload if valid, null otherwise
+ */
+export async function verifyJWT(
+	token: string,
+	options?: JwtOptions,
+): Promise<{ sub: string; aud: string; [key: string]: any } | null> {
```
Bekacru: I think we should imporve the return type here. You can use Jose types maybe

LightTab2:
I know this is probably of no use, because in #4748 I was more radical (and code got complicated; is outdated cuz no time) and I think it's actually more suited for separate community plugin than core one, but you're more than welcome to take any inspiration/code/tests/jsdocs from there.

himself65:
> I know this is probably of no use, because in #4748 I was more radical (and code got complicated; is outdated cuz no time) and I think it's actually more suited for separate community plugin than core one, but you're more than welcome to take any inspiration/code/tests/jsdocs from there.

Thanks for the comment. I will check it later. Right now this PR is mainly focus on the oauth-provider

PR #6125, opened 2025-11-20
feat: Add support for "jsonb" type in additional fields

GautamBytes:
### Description
This PR adds explicit support for the `jsonb` field type in `additionalFields`. While `json` currently maps to `jsonb` in Postgres, adding this explicit type improves clarity and allows for specific Postgres optimizations while maintaining backward compatibility.
image

### Changes
- **Core:** Added `jsonb` to `DBFieldType`.
- **Runtime:** Updated field inference to treat `jsonb` as an object (`Record<string, any>`).
- **Migrations:** Mapped `jsonb` to the `JSONB` SQL type for Postgres (with fallbacks to `JSON` or `TEXT` for MySQL/SQLite/MSSQL).
- **Validation:** Updated `to-zod.ts` to generate `z.any()` for `jsonb` fields to prevent runtime validation errors.
- **CLI:** Updated Drizzle and Prisma generators to correctly map and import the `jsonb` type.

### Impact
This is a purely additive, opt-in change. Existing `json` fields are unaffected.

---
## Summary by cubic
Adds explicit support for the jsonb type in additionalFields. Improves clarity and enables Postgres-specific optimizations while keeping existing json fields unchanged.

- **New Features**
  - Added jsonb to DBFieldType with runtime inference as Record<string, any>.
  - Mapped jsonb to Postgres JSONB with fallbacks: MySQL JSON, SQLite TEXT, MSSQL VARCHAR.
  - Updated Zod schema generation to handle jsonb via z.json() or z.any().
  - Updated Drizzle and Prisma generators to output jsonb correctly and adjust imports.

<sup>Written for commit 2a73d6e494fcc66873b9951df41c83a46fefa365. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/db/field.ts
+++ b/packages/better-auth/src/db/field.ts
@@ -28,13 +28,15 @@ export type InferValueType<T extends DBFieldType> = T extends "string"
 				? Date
 				: T extends "json"
 					? Record<string, any>
-					: T extends `${infer U}[]`
-						? U extends "string"
-							? string[]
-							: number[]
-						: T extends Array<any>
-							? T[number]
-							: never;
+					: T extends "jsonb"
+						? Record<string, any>
+						: T extends `${infer U}[]`
+							? U extends "string"
+								? string[]
+								: number[]
+							: T extends Array<any>
+								? T[number]
+								: never;
 
 export type InferFieldsOutput<Field> = Field extends Record<
 	infer Key,
--- a/packages/better-auth/src/db/get-migration.ts
+++ b/packages/better-auth/src/db/get-migration.ts
@@ -29,6 +29,7 @@ const postgresMap = {
 	boolean: ["bool", "boolean"],
 	date: ["timestamptz", "timestamp", "date"],
 	json: ["json", "jsonb"],
+	jsonb: ["jsonb", "json"],
 };
 const mysqlMap = {
 	string: ["varchar", "text", "uuid"],
@@ -44,6 +45,7 @@ const mysqlMap = {
 	boolean: ["boolean", "tinyint"],
 	date: ["timestamp", "datetime", "date"],
 	json: ["json"],
+	jsonb: ["json"],
 };
 
 const sqliteMap = {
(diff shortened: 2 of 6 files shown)


antonvishal:
Thanks for the PR @GautamBytes

packages/better-auth/src/db/field.ts
```diff
@@ -28,13 +28,15 @@ export type InferValueType<T extends DBFieldType> = T extends "string"
 				? Date
 				: T extends "json"
```
Bekacru, line 29: json and jsonb are the same concept at the js level. DBs decide weather to store it as jsonb or text

GautamBytes: The intention here was to allow users to be explicit in their schema definition. While Better Auth currently defaults `json` to `jsonb` in Postgres (which is great), some users (like the one who requested this) prefer seeing `type: "jsonb"` in their config to be 100% sure they are utilizing the binary format without relying on implicit adapter logic. Since I added `jsonb` to the core `DBFieldType` to support that explicit preference, I had to add it here in `field.ts` as well to ensure TypeScript infers it correctly as an object, just like `json`.

Bekacru: I think it’ll just introduce confusing options. Like what’s the difference between setting it to json vs jsonb in Postgres? If there’s no real difference other than being explicit, it doesn’t make sense to add it

antonvishal: Hey guys, if it is already using jsonb under the hood. It's fine maybe we can add a tiny section in the docs under /database mentioning this @Bekacru @GautamBytes

GautamBytes: > I think it’ll just introduce confusing options. Like what’s the difference between setting it to json vs jsonb in Postgres? If there’s no real difference other than being explicit, it doesn’t make sense to add it

Fair point, do you suggest adding it under docs and reverting these changes or i should close this pr? @Bekacru

antonvishal:
@himself65 this PR should be closed I believe as per the above comment by Bekacru

PR #6129, opened 2025-11-20
fix(anonymous): improve user name generation logic and type assertion hooks.after  AuthMiddleware

### LINKED ISSUE #6127: Using `exactOptionalPropertyTypes: true` in TSConfig Causes Type Errors in Anonymous Plugin

### To Reproduce

Here's the full minimal reproduction project, [using the documented usage of the plugin](https://www.better-auth.com/docs/plugins/anonymous):

`package.json`
```json
{
	"name": "repro",
	"version": "1.0.0",
	"type": "module",
	"scripts": {
		"types": "tsc --noEmit"
	},
	"devDependencies": {
		"typescript": "5.9.3"
	},
	"dependencies": {
		"better-auth": "1.3.34"
	}
}
```

`tsconfig.json`
```json
{
	"compilerOptions": {
		"target": "ESNext",
		"module": "esnext",
		"moduleResolution": "bundler",

		"skipLibCheck": true,
		"noEmit": true,
		"strictNullChecks": true,
		"exactOptionalPropertyTypes": true
	}
}
```

`src/index.ts`
```typescript
import { anonymous } from "better-auth/plugins";

const auth = betterAuth({
	plugins: [anonymous()], // Produces the type error
});

export { auth };
```

When running the `npm run types` command, I would expect the TypeScript compiler to produce no errors. However, the highlighted line produces the following error:

```
Type '{ id: "anonymous"; endpoints: { signInAnonymous: { <AsResponse extends boolean = false, ReturnHeaders extends boolean = false>(inputCtx_0?: ({ body?: undefined; } & { method?: "POST" | undefined; } & { query?: Record<string, any> | undefined; } & ... 4 more ... & { ...; }) | undefined): Promise<...>; options: { ...;...' is not assignable to type 'BetterAuthPlugin'.
The types of 'hooks.after' are incompatible between these types.
```

1.3.34

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "release": "25.1.0",
    "cpuCount": 14,
    "cpuModel": "Apple M4 Pro",
    "totalMemory": "48.00 GB",
    "freeMemory": "7.90 GB"
  },
  "node": {
    "version": "v24.10.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.1"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "1.3.34",
    "config": null
  }
}
```

Backend

```typescript
import { anonymous } from "better-auth/plugins";

const auth = betterAuth({
	plugins: [anonymous()], // Produces the type error
});

export { auth };
```

### Additional context

The type error only appears when `exactOptionalPropertyTypes` in `tsconfig.json` is set to `true`. With the option set to `false`, the type check passes successfully. **It affects both backend and frontend parts**.

 I have tracked the issue back a few versions, and this error started appearing from `1.3.28` onward. `1.3.27` is working as expected without any type errors. Looking at the [1.3.28 changelog](https://github.com/better-auth/better-auth/releases/tag/v1.3.28), I don't see an immediate cause of the issue. But I have only taken a brief look.

rovertrack:
solves #6127 

---
## Summary by cubic
Improves anonymous user creation by safely generating names and tightening types in the after hook middleware. Fixes #6127.

- **Bug Fixes**
  - Only call generateName when it’s a function; otherwise default to "Anonymous".
  - Assert the after hook as AuthMiddleware to resolve type issues.

<sup>Written for commit ea1131d2cd6bcab85894b8c77101f78b7786b66a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/concepts/database.mdx
+++ b/docs/content/docs/concepts/database.mdx
@@ -5,9 +5,13 @@ description: Learn how to use a database with Better Auth.
 
 ## Adapters
 
-Better Auth requires a database connection to store data. The database will be used to store data such as users, sessions, and more. Plugins can also define their own database tables to store data.
+Better Auth connects to a database to store data. The database will be used to store data such as users, sessions, and more. Plugins can also define their own database tables to store data.
 
 You can pass a database connection to Better Auth by passing a supported database instance in the database options. You can learn more about supported database adapters in the [Other relational databases](/docs/adapters/other-relational-databases) documentation.
+
+<Callout type="info">
+Better Auth also works without any database. For more details, see [Stateless Session Management](/docs/concepts/session-management#stateless-session-management).
+</Callout>
 
 ## CLI
 
--- a/packages/better-auth/src/plugins/oidc-provider/authorize.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/authorize.ts
@@ -202,6 +202,7 @@ export async function authorize(
 	const hasAlreadyConsented = await ctx.context.adapter
 		.findOne<{
 			consentGiven: boolean;
+			scopes: string;
 		}>({
 			model: "oauthConsent",
 			where: [
@@ -215,7 +216,16 @@ export async function authorize(
 				},
 			],
 		})
-		.then((res) => !!res?.consentGiven);
+		.then((res) => {
+			if (!res?.consentGiven) {
+				return false;
+			}
+			const consentedScopes = res.scopes ? res.scopes.split(" ") : [];
+			const hasConsented = requestScope.every((scope) =>
+				consentedScopes.includes(scope),
+			);
+			return hasConsented;
+		});
 
 	const promptSet = parsePrompt(query.prompt ?? "");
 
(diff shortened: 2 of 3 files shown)


packages/better-auth/src/plugins/anonymous/index.ts
```diff
@@ -199,7 +203,7 @@ export const anonymous = (options?: AnonymousOptions | undefined) => {
 						if (!options?.disableDeleteAnonymousUser) {
 							await ctx.context.internalAdapter.deleteUser(session.user.id);
 						}
-					}),
+					}) as AuthMiddleware,
```
himself65, line 206: this seems unncessary

packages/better-auth/src/plugins/anonymous/index.ts
```diff
@@ -84,7 +85,10 @@ export const anonymous = (options?: AnonymousOptions | undefined) => {
 					}
 
 					const email = await getAnonUserEmail(options);
-					const name = (await options?.generateName?.(ctx)) || "Anonymous";
+					const name =
```
himself65, line 88: This also seems unnecessary. If you really want to check if generateName is a function, please put it into the init step

himself65 (changes requested):

PR #6132, opened 2025-11-20
chore(fix): usePlural should work wtith joins

ping-maxwell:
This PR fixes:

- Prisma CLI generator doesn’t respect `usePlural`
- No schema unit tests for usePlural.
- Drizzle usePlural will cause relations to generate with double `s`’s (eg “sessionss”)

---
## Summary by cubic
Fixes usePlural handling across adapters and CLI generators so joins and relation keys are correct and schemas generate consistently. Prevents double “s” and mismatched singular/plural mapping in Drizzle and Prisma.

- **Bug Fixes**
  - Drizzle adapter: correct join key suffix based on isUnique/usePlural; remap join results to expected keys under usePlural; only apply limit/offset when provided.
  - Prisma adapter: join key naming respects usePlural; relation field names toggle between singular/plural accordingly.
  - CLI generators: Drizzle relations avoid double “s”; Prisma and Drizzle use initGetModelName/initGetFieldName for consistent naming; added snapshots to lock schema output with usePlural.
  - Utilities/tests: removed noisy schema error logs; test adapters and cleanup resolve model names with usePlural.

<sup>Written for commit 6fbecc8f8f3a76c17a66ab3d2346bb7093e0065a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/adapters/adapter-factory/get-default-field-name.ts
+++ b/packages/better-auth/src/adapters/adapter-factory/get-default-field-name.ts
@@ -50,7 +50,9 @@ export const initGetDefaultFieldName = ({
 			}
 		}
 		if (!f) {
-			throw new BetterAuthError(`Field ${field} not found in model ${model}`);
+			throw new BetterAuthError(
+				`Field ${field} not found in model ${model}: ${new Error().stack?.replace("Error:", "")}`,
+			);
 		}
 		return field;
 	};
--- a/packages/better-auth/src/adapters/adapter-factory/get-default-model-name.ts
+++ b/packages/better-auth/src/adapters/adapter-factory/get-default-model-name.ts
@@ -42,7 +42,9 @@ export const initGetDefaultModelName = ({
 		}
 
 		if (!m) {
-			throw new BetterAuthError(`Model "${model}" not found in schema`);
+			throw new BetterAuthError(
+				`Model "${model}" not found in schema: ${new Error().stack?.replace("Error:", "")}`,
+			);
 		}
 		return m;
 	};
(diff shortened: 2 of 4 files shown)


packages/better-auth/src/adapters/adapter-factory/get-default-field-name.ts
```diff
@@ -6,13 +6,16 @@ import { initGetDefaultModelName } from "./get-default-model-name";
 export const initGetDefaultFieldName = ({
 	schema,
 	usePlural,
+	suppressErrorLogs = false,
```
Bekacru: we shulnd't add option like this. Either read from the logger config or If it’s specific to the CLI, that behavior shouldn’t leak into the internal logic

packages/better-auth/src/adapters/adapter-factory/get-default-field-name.ts
```diff
@@ -51,9 +54,11 @@ export const initGetDefaultFieldName = ({
...
 		}
 		if (!f) {
-			logger.error(`Field "${field}" not found in schema`);
-			logger.error(`Schema:`, schema);
-			logger.error(`Error stack:`, new Error().stack?.replace("Error:", ""));
+			if (!suppressErrorLogs) {
+				logger.error(`Field "${field}" not found in schema`);
+				logger.error(`Schema:`, schema);
+				logger.error(`Error stack:`, new Error().stack?.replace("Error:", ""));
+			}
```
Bekacru: yeah this should be determined by the logger config

packages/better-auth/src/adapters/adapter-factory/get-default-model-name.ts
```diff
@@ -5,9 +5,11 @@ import { BetterAuthError } from "@better-auth/core/error";
 export const initGetDefaultModelName = ({
 	usePlural,
 	schema,
+	suppressErrorLogs = false,
```
Bekacru: same here...remove this option everywhere

packages/better-auth/src/adapters/adapter-factory/get-default-field-name.ts
Bekacru: maybe we should remove this logs altogthere and jsut throw the error

PR #6136, opened 2025-11-20
fix(device-authorization): make verificationUri a required option

bytaesu:
This PR introduces a breaking change in the device-authorization plugin by making verificationUri a required option.

According to [RFC 8628 Section 3.2](https://datatracker.ietf.org/doc/html/rfc8628#section-3.2)￼, `verification_uri` is required, but we also need to allow it to be customized. We are changing it from an internally defaulted value to a required option.

---
## Summary by cubic
Makes verificationUri a required option in the device-authorization plugin to comply with RFC 8628. Removes the internal default; you must provide the verification URL.

- **Migration**
  - Pass verificationUri in deviceAuthorization({ verificationUri: "/device" }) or use an absolute URL.
  - verification_uri and verification_uri_complete are derived from the provided value (user_code is appended to the complete URI).

<sup>Written for commit 6f0f110fd163b8d306bd3d2e83756148c9c80132. Summary will update automatically on new commits.</sup>

### DIFF
--- a/e2e/smoke/test/fixtures/tsconfig-declaration/src/demo.ts
+++ b/e2e/smoke/test/fixtures/tsconfig-declaration/src/demo.ts
@@ -192,6 +192,7 @@ export const auth = betterAuth({
 			],
 		}),
 		deviceAuthorization({
+			verificationUri: "/device",
 			expiresIn: "3min",
 			interval: "5s",
 		}),


docs/content/docs/plugins/device-authorization.mdx
```diff
@@ -39,14 +39,11 @@ This will demonstrate the complete device authorization flow by:
...
         export const auth = betterAuth({
           // ... other config
-          plugins: [ // [!code highlight]
+          plugins: [
             deviceAuthorization({ // [!code highlight]
-              // Optional configuration
-              verificationUri: "/device/verify", // Custom URL where users verify their device code (default: /device) // [!code highlight]
-              expiresIn: "30m", // Device code expiration time // [!code highlight]
-              interval: "5s",    // Minimum polling interval // [!code highlight]
+              verificationUri: "/device", // [!code highlight]
```
himself65, line 44: Should we set `verificationUri` to /device by default? I see it's from spec, but we could help developer to set it automatally

bytaesu: I think this should be customizable, not a default. For example, if Better Auth is used from a backend separate from the frontend, a fixed relative path wouldn’t work

himself65: Yeah, I agree it should be customizable, but I think what we need to do is not set it to required. We just mention it in the doc that you should add /device page or add you custom url

Bekacru:
I think for now since this would be breaking change let's keep it as it is

docs/content/docs/plugins/device-authorization.mdx
himself65, line 1: Could you please cherry-pick the doc update to a new PR?

bytaesu: I’ll quickly reopen it with a brief explanation of verification uri

Issue #5965, opened 2025-11-13
[Feature request] Implement a new hubspot as social-provider


realtebo:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

Hubspot is quite a standard Oauth provider.

We can succesfully configure it using oauth plugin, but it needed customization for a function

### Describe the solution you'd like

use the code from the next section to implement a new social provider for hubspot

it's quite standard, except the implementation of `getUserInfo` that requires to differently map the hubspot-provided data to user profile object required by your package

### Describe alternatives you've considered

currently we are doing this
- use the `genericOAuth` plugin
- customize `getUserInfo`

```
genericOAuth({
      config: [
        {
          providerId: "hubspot",
          clientId: "<app client id redacted>", 
          clientSecret: "<app client secret>",
          authorizationUrl: "https://app.hubspot.com/oauth/<portal_id>/authorize",
          tokenUrl: "https://api.hubapi.com/oauth/v1/token",

          //do not use this
          // userInfoUrl: "https://api.hubapi.com/oauth/v1/access-tokens/${token}",

          // use this implementation
          async getUserInfo(tokens) {
            const response = await fetch(
              `https://api.hubapi.com/oauth/v1/access-tokens/${tokens.accessToken}`
            );
            if (!response.ok) {
              throw new Error("Failed to retrieve HubSpot user info");
            }

            const data = await response.json();

            const rawId = data.signed_access_token?.userId as string;

            const email = data.user;

            return {
              id: rawId,
              email,
              name: email,
              emailVerified: true,
            };
          },
        },
      ],
    }),
```

Issue #5493, opened 2025-10-22
Line provider: Support for multiple countries (server-side)


JeongJuhyeon:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

Line is the most popular messenger and super-app in Japan, Thailand and Taiwan, but the current `better-auth` LINE provider only supports a single country.

<img width="608" height="107" alt="Image" src="https://github.com/user-attachments/assets/703de92e-e384-4643-82cb-1ae30429d2ac" />

^ You need to create one channel per country, each will have its own client id and client secret. But betterAuth's Line config only supports a single clientId and clientSecret.

### Describe the solution you'd like

Accepting an array of objects, or other schema through which we can accept multiple countries for applications that have many users in countries like Japan, Thailand and Taiwan

### Describe alternatives you've considered

Forking the provider I guess..

### Additional context

From a Western developer perspective this may sound like it's extremely niche but these are wealthy countries representing 200+ million people where it's effectively the most used app.

PR #6138, opened 2025-11-20
feat(extend auth providers): hubspot and line providers

Paola3stefania:
Closes Issues [#5965](https://github.com/better-auth/better-auth/issues/5965) and [#5493](https://github.com/better-auth/better-auth/issues/5493)

---------------

## Summary by cubic
Added HubSpot and LINE OAuth providers to the generic OAuth plugin to enable sign-in and basic profile retrieval from these platforms.

- **New Features**
  - HubSpot provider: default scope "oauth"; retrieves user via access token info endpoint; returns id and email (no profile picture).
  - LINE provider: supports multiple channels via providerId; default scopes "openid", "profile", "email"; uses ID token when available, otherwise falls back to userinfo.
  - Providers exported in the generic-oauth index for easy import (line, hubspot).

<sup>Written for commit 13462d82922e3b10392807bf3ab6b04ac88a42d3. Summary will update automatically on new commits.</sup>


packages/better-auth/src/plugins/generic-oauth/providers/line.ts
```diff
@@ -0,0 +1,136 @@
+import type { OAuth2Tokens, OAuth2UserInfo } from "@better-auth/core/oauth2";
```
Bekacru, line 1: it's worth adding documentation for this inside the current line docs

packages/better-auth/src/plugins/generic-oauth/providers/index.ts
```diff
@@ -5,7 +5,7 @@
  *
  * @example
  * ```ts
- * import { genericOAuth, okta, auth0, microsoftEntraId, slack, keycloak } from "better-auth/plugins/generic-oauth";
+ * import { genericOAuth, okta, auth0, microsoftEntraId, slack, keycloak, line, hubspot } from "better-auth/plugins/generic-oauth";
```
Bekacru, line 8: we should document this under supported providers on generic oauth docs

Bekacru (changes requested):

packages/better-auth/src/plugins/generic-oauth/providers/line.ts
Paola3stefania: 100 %

Paola3stefania:
@Bekacru  Updated documentation. Wondering if in the future we want a dropdown as we have for social providers?

Bekacru:
> @Bekacru Updated documentation. Wondering if in the future we want a dropdown as we have for social providers?

I think we should have drop down under plugins at somepoint. Plugins like organization for example are wayy to long to be just one page. Let's figure out the design and we'll do it then

docs/content/docs/plugins/generic-oauth.mdx
```diff
@@ -193,6 +213,43 @@ All providers support the same optional fields:
 - `disableSignUp?: boolean` - Disable sign-up entirely
 - `overrideUserInfo?: boolean` - Override user info on sign in
 
+### Multi-Channel Support (LINE)
```
Bekacru: I think this should be mentioned in line provider docs instead...show that they can use generic oauth plugin for multi channel support

Paola3stefania: This makes sense. Change made.

realtebo:
Sorry, but I cannot find any evidence of this new social provider in the doc

realtebo:
ensure that hubspot user id AND hubspot portal id are BOTH exposed

PR #6143, opened 2025-11-20
fix(device-authorization): sanitize user code input on device approve

Bekacru:
## Summary by cubic
Allows device approval with hyphenated user codes and deletes device codes only after successful authorization. Prevents lookup failures and premature cleanup during the device flow.

- **Bug Fixes**
  - Normalize user code by removing dashes before lookup on approve.
  - Move device code deletion to run after authorization succeeds.

<sup>Written for commit 779272243bc202d5b438aec713dccb8a63d53127. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/device-authorization/index.ts
+++ b/packages/better-auth/src/plugins/device-authorization/index.ts
@@ -554,17 +554,6 @@ Follow [rfc8628#section-3.4](https://datatracker.ietf.org/doc/html/rfc8628#secti
 						deviceCodeRecord.status === "approved" &&
 						deviceCodeRecord.userId
 					) {
-						// Delete the device code after successful authorization
-						await ctx.context.adapter.delete({
-							model: "deviceCode",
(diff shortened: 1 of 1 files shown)


himself65 (approved):

packages/better-auth/src/plugins/device-authorization/index.ts
```diff
@@ -780,14 +780,15 @@ Follow [rfc8628#section-3.4](https://datatracker.ietf.org/doc/html/rfc8628#secti
 					}
 
 					const { userCode } = ctx.body;
+					const cleanUserCode = userCode.replace(/-/g, "");
```
himself65, line 783: Originally, I didn't want the user to send code including the `-` character. But, it's okay

Bekacru: I think it's in the rfc to expect that if im correct

PR #6144, opened 2025-11-20
refactor(haveibeenpwned): add try catch and timeout

Bekacru:
## Summary by cubic
Add a 5s timeout to the Have I Been Pwned API call and wrap the password compromise check in try/catch so hashing always proceeds if the request fails. This prevents auth flows from hanging or breaking when the HIBP service is slow or unreachable, while still checking on allowed paths.

<sup>Written for commit 2c2b990aa20154b121f413009f4359199c4e825d. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/haveibeenpwned/index.ts
+++ b/packages/better-auth/src/plugins/haveibeenpwned/index.ts
@@ -29,6 +29,7 @@ async function checkPasswordCompromise(
 					"Add-Padding": "true",
 					"User-Agent": "BetterAuth Password Checker",
 				},
+				timeout: 5000, // 5s timeout
 			},
 		);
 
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/haveibeenpwned/index.ts
```diff
@@ -81,15 +82,19 @@ export const haveIBeenPwned = (options?: HaveIBeenPwnedOptions | undefined) => {
...
+								const c = await getCurrentAuthContext();
+								if (!c.path || !paths.includes(c.path)) {
+									return ctx.password.hash(password);
+								}
+								await checkPasswordCompromise(
+									password,
+									options?.customPasswordCompromisedMessage,
+								);
+								return ctx.password.hash(password);
+							} catch (e) {
```
himself65, line 95: The try-catch block here will omit any errors inside `checkPasswordCompromise`, which renders error caching in that function useless. So we definitely should remove one  of the two

himself65 (changes requested):

PR #6147, opened 2025-11-20
feat(jwt): add key rotation

Bekacru:
Introduces key rotation


LightTab2:
This is cool, but shouldn't it be generalized so we can hold multiple keysets with their own unique ids that expire independently (maybe with custom time too)?

Bekacru:
> This is cool, but shouldn't it be generalized so we can hold multiple keysets with their own unique ids that expire independently (maybe with custom time too)?

probably. That'd require introducing keysetId or something to group them. Didn't want to introduce required field for next release

PR #6155, opened 2025-11-21
feat(multi-session): use database for multi session

Bekacru:
Multi-session previously relied on cookies. But cookie-based multi-session comes with a lot of limitations. It’s better to introduce a multi-session approach that relies on server-side state instead

---
## Summary by cubic
Switch multi-session from cookie-based to server-side, device-bound sessions. Improves security, simplifies logic, and enforces per-device session limits.

- **New Features**
  - Add a signed device_id cookie (configurable name/options) to bind sessions to a device.
  - Database hook on session creation sets deviceId and prunes oldest sessions when exceeding the maximum (default 5).
  - List device sessions now queries by deviceId and returns session + user data.
  - Set active validates device ownership (deviceId) before switching.
  - Revoke deletes the session and clears the active cookie if it was the current session.

- **Migration**
  - Add deviceId (string, optional, input: false) to the session schema.
  - Remove reliance on “_multi-” session cookies; multi-session is now database-driven.
  - Existing client API calls (list/setActive/revoke) continue to work; cookie handling is internal.

<sup>Written for commit 0bbe0266aaff82505642c54f1d0b65f0b8442513. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/multi-session.mdx
+++ b/docs/content/docs/plugins/multi-session.mdx
@@ -62,7 +62,7 @@ type listDeviceSessions = {
 
 ### Set active session
 
-To set the active session, you can call the `setActive` method.
+To set the active session, you can call the `setActiveSession` method.
 
 <APIMethod
   path="/multi-session/set-active"
@@ -81,7 +81,7 @@ type setActiveSession = {
 
 ### Revoke a session
 
-To revoke a session, you can call the `revoke` method.
+To revoke a session, you can call the `revokeDeviceSession` method.
 
 <APIMethod
   path="/multi-session/revoke"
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/multi-session/index.ts
```diff
@@ -20,21 +18,122 @@ interface MultiSessionConfig {
...
+
+										// Delete oldest sessions until we fit in the limit
+										const toDelete = sorted.slice(
+											0,
+											validSessions.length - (opts.maximumSessions - 1),
+										);
+
+										await Promise.all(
+											toDelete.map((s) =>
+												ctx.context.internalAdapter.deleteSession(s.token),
```
himself65, line 120: Use a transaction?

himself65 (dismissed):

Bekacru:
I'm gonna close this. It actually requires a little bit more thinknig than I thought :)

PR #6167, opened 2025-11-21
chore(docs): remove db requirement info in installation docs

ping-maxwell:
## Summary by cubic
Update installation docs to clarify that a database is optional. Renamed the step to "Configure Database (optional)" and added a callout noting stateless mode support and that plugins require a database.

<sup>Written for commit 382b24b4d9291de99b54f57615b33ec0ea6fe195. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/jwt.mdx
+++ b/docs/content/docs/plugins/jwt.mdx
@@ -338,6 +338,12 @@ Table Name: `jwks`
       type: "Date", 
       description: "Timestamp of when the web key was created" 
     },
+    { 
+      name: "expiresAt", 
+      type: "Date", 
+      description: "Timestamp of when the web key expires",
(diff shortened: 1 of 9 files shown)


docs/content/docs/installation.mdx
```diff
@@ -66,9 +66,10 @@ export const auth = betterAuth({
 </Step>
 
 <Step>
-### Configure Database
+### Configure Database (optional)
+
+You can optionally configure a database to store user data. (Recommended)
```
himself65: Perhaps we should declare that if no database is provided, certain plugins may not function properly. Or we list db is required in each plugin page

ping-maxwell: updated.

PR #6181, opened 2025-11-22
fix: lockfile

bytaesu:
## Summary by cubic
Cleaned up pnpm-lock.yaml to match the current dependency graph and fix formatting for stable installs. Removed stale rolldown references, bumped better-call to 1.1.0-beta.2, and added missing bufferutil/utf-8-validate as transitive peer deps for React Native dev middleware to prevent warnings.

<sup>Written for commit 73dc4cc5e6069a1edd422f00895154cf1f77bdfa. Summary will update automatically on new commits.</sup>


bytaesu:
=> resolved

PR #6186, opened 2025-11-22
fix(client): get-session gets triggered twice on foucs

Bekacru:
## Summary by cubic
Stop double getSession calls when returning to the tab. Removed the window focus listener and now rely on document visibilitychange to set focus only when the page becomes visible.

<sup>Written for commit 978cf9858fa73ac8ec935ef3fed4171c526cedaf. Summary will update automatically on new commits.</sup>


himself65 (commented):
I think we should use debounce here? which might more make sense

Bekacru:
> I think we should use debounce here? which might more make sense

yeah we should probably do debounce

PR #6227, opened 2025-11-23
chore: fix plugin import

hyoban:
```
The inferred type of 'baseAuthPlugins' cannot be named without a reference to '../../../../node_modules/better-auth/dist/index-SBjk_96V.mjs'. This is likely not portable. A type annotation is necessary.ts(2742)
```

---
## Summary by cubic
Fixed the two-factor client to import the plugin type from "." instead of "../../plugins/two-factor", resolving the ts(2742) portability error. Also switched additional-fields client type imports to "@better-auth/core" to avoid non-portable inferred types.

<sup>Written for commit 4b0021b0ceb18c6eba15283fbd0ac3c45a58c639. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/additional-fields/client.ts
+++ b/packages/better-auth/src/plugins/additional-fields/client.ts
@@ -1,6 +1,9 @@
-import type { BetterAuthClientPlugin } from "@better-auth/core";
+import type {
+	BetterAuthClientPlugin,
+	BetterAuthOptions,
+	BetterAuthPlugin,
+} from "@better-auth/core";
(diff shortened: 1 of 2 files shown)


jslno:
@Bekacru This error is still ocurring just with a different named file. We need to export the necessary types to make them accessible to the consumer

`The inferred type of 'clientOptions' cannot be named without a reference to '../../node_modules/better-auth/dist/index-DXMBWrNT.mjs'. This is likely not portable. A type annotation is necessary. (ts 2742)`

This is also occurs for some other client plugins
See #6230

PR #6235, opened 2025-11-23
fix: support @tanstack/solid-start in tanstackStartCookies plugin

jakst:
The react specific import doesn't work with solid-start. But it's just a re-export from '@tanstack/start-server-core', so if we import directly from there instead, solid-start is automatically supported.

---
## Summary by cubic
Fixed tanstackStartCookies to import setCookie from @tanstack/start-server-core instead of @tanstack/react-start/server. This adds support for @tanstack/solid-start and prevents import errors.

<sup>Written for commit 67c6b5f7dd6dae15c85c28af37fc3ad5c1689033. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/package.json
+++ b/packages/better-auth/package.json
@@ -492,7 +492,7 @@
     "@lynx-js/react": "^0.114.0",
     "@prisma/client": "^5.22.0",
     "@sveltejs/kit": "^2.37.1",
-    "@tanstack/react-start": "^1.131.3",
+    "@tanstack/start-server-core": "^1.131.3",
     "@types/better-sqlite3": "^7.6.13",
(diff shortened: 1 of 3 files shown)


jakst:
@Bekacru did you notice the merge to main failed?

CarlosZiegler:
Nice, but tried on my React App and looks like still have a Bug when using better-auth > 1.4.3

image

austinm911:
> Nice, but tried on my React App and looks like still have a Bug when using better-auth > 1.4.3
> 
> 
> 
> image
> 
> 

Yes here are more details - this reintroduces the problem

https://github.com/TanStack/router/issues/5795#issuecomment-3579298576

PR #6240, opened 2025-11-23
feat: unique where requirement, and other adapter improvements

ping-maxwell:
## Features / Improvements
1. Unique Where Requirement - Ensures that where clauses for single-query operations meet the requirement to at least include 1 unique field in the where clause.
2. Unit Test Improvements:  
    - Kysely `node-sqlite dialect` was using the old better-auth adapter test system, it's now moved to use the new test suite version.
    - Currently **some** adapter test names start with `adapter.<adapter_name>.<dialect>.test.ts` this makes for a long test name. Using something like `drizzle.sqlite.test.ts` is enough, we do not need `adapter.drizzle.sqlite.test.ts`
3. Type definition improvements, under the `JoinConfig`. We currently have the following: `JoinConfig.on.from` & `JoinConfig.on.to`, this is now changed to `JoinConfig.on.field` & `JoinConfig.on.referencing` for more accurate naming.

### Unique Where Requirement fixes the following plugins:
(Technically speaking, everything here will fail for prisma users)

- **TwoFactor**:
  - `viewBackupCodes` tries to find twoFactors by userId, but uid fk is not unique.
  - `disableTwoFactor` tries to delete a twoFactor by userId, but uid fk is not unique, must use deleteMany
- **Organization**:
  - `findMemberByEmail` attempts to find a user by email, but `email` is not `unique`.
  - `findMemberByOrgId` tries to find an org by `organizationId` & `userId`, neither are `unique`.
  - `findOrCreateTeamMember` tries to find an org by `teamId` & `userId`, neither are `unique`.
  - `checkMembership` tries to find a member by `userId` & `organizationId`
  - `findTeamMember` tries to find by `teamId` & `userId`, neither of which are `unique`.
- **Device Authorization**:
  - `deviceCode` should be `unique`
  - `deviceVerify`, `deviceDeny` & `deviceApproves` tries to find `deviceCode` but `userCode` is not unique
- **API Key**:
  - `key` should be unique
- **username**:
  - `signInWithUsername` tries to find 1 account but `providerId` and `userId` are not unique
 - **oidc-provider**:
   - `authorize` tries to find 1 `oauthConsent` via `clientId` & `userId`, but none are `unique`.
- **sso**:
  - `verifyDomain`: tries to find verification via `identifier` but it's not unique
  - `signInSSO`: tries to get the `ssoProvider` but `organizationId` is not unique
  - `registerSSOProvider`: tries to check if a user is a member of an org but `userId` & `orId` is not uniuqe
  - `callbackSSO` finds member but not using unique keys
- **stripe**:
  - `onSubscriptionUpdated` tries to find a subscription by `stripeSubscriptionId` but it's not unique
  - `upgradeSubscription` tries to find a subscription using `referenceId` but it's not unique.
- **swie**:
  - `verifySwieMessage` tries to find a `walletAddress` using `chainId` or `address` which isn't unique
- **scim**:
  - `generateSCIMToken` tries to find `scimProvider` from `organizationId` which is not unique
  - `createSCIMUser` tries to find existing accounts  but `accountId` & `providerId` are not unique
  - `findUserById` tries to find  existing accounts & members but `userId` & `providerId` are not unique

And Internal adapter which touches several core endpoints:
  - `findAccount` tries to find 1 account, but `accountId` is not unique 
  - `deleteVerificationByIdentifier` tries to delete 1 `identifier` from verification model, but `identifier` is not unique.

---
## Summary by cubic
Enforces a unique-field requirement for single-record queries in the adapter factory to prevent ambiguous matches. Applies to findOne, update, and delete, with schema and call-site updates; also refines join config naming.

- **New Features**
  - Requires an equality (eq) on a unique field for single-record queries; operator defaults to eq and throws BetterAuthError when no unique-eq filter is provided.
  - Adds disableTransformWhere for advanced/testing.
  - Marks apiKey.key and deviceCode.deviceCode as unique.
  - Prisma schema generator disallows defaults on unique fields; createdAt gets a default only when not unique.
  - Renames JoinConfig.on.from/to to on.field/references.

- **Migration**
  - For findOne, update, and delete, include id or another unique field with eq in where.
  - Use findMany with limit: 1 for non-unique lookups, and deleteMany/updateMany when the where clause isn’t unique.
  - In Prisma schemas, do not set default values on unique fields.

<sup>Written for commit dbe94bf59df5d976c538f2021eaf577cf750f298. Summary will update automatically on new commits.</sup>


packages/stripe/src/index.ts
```diff
@@ -285,10 +285,13 @@ export const stripe = <O extends StripeOptions>(options: O) => {
...
-								model: "subscription",
-								where: [{ field: "referenceId", value: referenceId }],
-							})
+						? await ctx.context.adapter
+								.findMany<Subscription>({
+									model: "subscription",
+									where: [{ field: "referenceId", value: referenceId }],
+									limit: 1,
+								})
+								.then((subs) => subs[0])
```
Bekacru, line 294: why are we doing find many? find one should get us the first element

Issue #6262, opened 2025-11-24
Add option to rename relations


antonykamp:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

When I enable the experimental join feature and generate the schema, I am required to use the generated names of the relations in my Prisma schema. The generated ones doesn't match the case style I use.

### Describe the solution you'd like

I'd love to have an option in the better auth configuration object to rename the relation, like a field:

```ts
 user: {
    fields: {
      name: "firstName"
      usersessions: "UserSessions"
    }
}
```

### Describe alternatives you've considered

Right now I have to rename the relations in my prisma schema.

antonykamp:
Thank you for the (automated) response 😉 We use pascal case.

GautamBytes:
Will work on this @antonykamp

PR #6271, opened 2025-11-24
feat(prisma): add option to rename relations in schema and adapter

GautamBytes:
This PR addresses #6262 by allowing users to customize relation names in their Prisma schema configuration. Previously, relation names were hardcoded (e.g., sessions or user), which caused conflicts with existing databases or strict naming conventions (e.g., requiring PascalCase UserSessions).

Changes
- packages/core: Added optional relations?: Record<string, string> configuration to user, session, and account options.
- packages/cli: Updated generators/prisma.ts to respect custom relation names for both:
- Forward Relations: (e.g., user field on the Session model).
- Back Relations: (e.g., sessions field on the User model).
- packages/better-auth: Updated prisma-adapter.ts runtime logic to dynamically look up the correct relation key during joins.
- Added safety checks to handle singular/plural mismatch (e.g., if config maps "session" but internal logic looks for "sessions").

---
## Summary by cubic
Adds support for custom Prisma relation field names in the CLI generator and adapter so projects can match existing DB conventions. Aligns with Linear #6262 by removing hardcoded names and handling singular/plural safely.

- **New Features**
  - core: Added relations?: Record<string, string> to user, session, and account options (key = default name, value = custom name).
  - cli: Prisma schema generator applies custom relation names for forward and back relations, with singular/plural based on FK uniqueness and usePlural.
  - adapter: Runtime join logic resolves the correct relation key from config and falls back cleanly, including singular/plural mismatch handling.

- **Migration**
  - To rename a relation, set options.<model>.relations = { "<default>": "<custom>" } (e.g., { "sessions": "UserSessions" }). No change needed if defaults are fine.

<sup>Written for commit 62c6da0d54d302e26043ff8618dfd4f8cb0d853c. Summary will update on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/one-tap.mdx
+++ b/docs/content/docs/plugins/one-tap.mdx
@@ -107,20 +107,21 @@ await authClient.oneTap({
 
 ### Client Options
 
-- **clientId**: The client ID for your Google One Tap API.
-- **autoSelect**: Automatically select the account if the user is already signed in. Default is false.
-- **context**: The context in which the One Tap API should be used (e.g., "signin"). Default is "signin".
-- **cancelOnTapOutside**: Cancel the One Tap popup when the user taps outside it. Default is true.
-- additionalOptions: Extra options to pass to Google's initialize method as per the [Google Identity Services docs](https://developers.google.com/identity/gsi/web/reference/js-reference#google.accounts.id.prompt).
-- **promptOptions**: Configuration for the prompt behavior and exponential backoff:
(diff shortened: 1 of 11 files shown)


GautamBytes:
> Could you please add test cases?
> 
> /cc @ping-maxwell take a look

yeah sure, will add tomorrow!

ping-maxwell (changes requested):
Please add documentation at `prisma.mdx`

ping-maxwell:
On a side note, In the future we will also need to support drizzle and potentially other adapters that could generate relations

Issue #5034, opened 2025-10-02
expo - state_mismatch error for google


fdemir:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Use the following setup and sign in using `authClign.signIn.social`. Nothing different.

### Current vs. Expected behavior

Current:
BetterAuth gives `Error Code: state_mismatch`.

Expected behavior:
User should be redirect to the app.

### What version of Better Auth are you using?

1.2.12

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.4.0: Fri Apr 11 18:33:47 PDT 2025; root:xnu-11417.101.15~117/RELEASE_ARM64_T6030",
    "release": "24.4.0",
    "cpuCount": 11,
    "cpuModel": "Apple M3 Pro",
    "totalMemory": "18.00 GB",
    "freeMemory": "0.24 GB"
  },
  "node": {
    "version": "v20.12.2",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.5.0"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null
  }
```

### Which area(s) are affected? (Select all that apply)

Backend, Client

### Auth config (if applicable)

```typescript
import { expo } from "@better-auth/expo";
import { prisma } from "@wac/db";
import { betterAuth } from "better-auth";
import { prismaAdapter } from "better-auth/adapters/prisma";
import { magicLink } from "better-auth/plugins";
import { Resend } from "resend";

const isProduction = process.env.NODE_ENV === "production";

const resend = new Resend(process.env.RESEND_API_KEY);

export const auth = betterAuth({
  trustedOrigins: [
    "yourapp://",
    "https://yourapp.in",
    "https://www.yourapp.com",
  ],
  advanced: {
    ...(isProduction && {
      crossSubDomainCookies: {
        enabled: true,
        domain: ".yourapp.com",
      },
    }),
  },
  plugins: [
    expo({}),
    magicLink({
      sendMagicLink: async ({ email, token, url }, request) => {
        await resend.emails.send({
          from: "noreply@yourapp.com",
          to: email,
          subject: "Login to YourApp",
          html: `<p>Login to your account with this link: <a href="${url}">Click here to sign in</a></p>`,
        });
      },
    }),
  ],
  database: prismaAdapter(prisma, {
[...]

PR #6278, opened 2025-11-24
set default to cookie to match JSDoc

van-sprundel:
Would fix https://github.com/better-auth/better-auth/issues/5034 since after "1.3.13" this was broken for people (see my last comment)

Pre "1.3.13" it was essentially cookie-based, so this would be similar. Users who want the added security of database storage can still explicitly set storeStateStrategy: "database" and use skipStateCookieCheck: true if cookies are unreliable in their environment.

I'm not sure what the team thinks so let me know if this is desired or not

---
## Summary by cubic
Restore oauthConfig.storeStateStrategy default to "cookie" to match the JSDoc and previous behavior. This fixes a regression after 1.3.13 that broke OAuth state handling for apps not setting this option.

- **Bug Fixes**
  - Default storeStateStrategy is now "cookie" when not provided.
  - Aligns runtime with docs and prevents state mismatches during sign-in.

<sup>Written for commit 5390502fe1b0a2e19a15ef79d55c26fdff71b028. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/index.ts
+++ b/packages/better-auth/src/index.ts
@@ -9,6 +9,7 @@ export * from "@better-auth/core/oauth2";
 export * from "@better-auth/core/utils";
 //#endregion
 export * from "./auth";
+// @ts-expect-error
 export * from "./types";
 export * from "./utils";
 // export this as we are referencing OAuth2Tokens in the `refresh-token` api as return type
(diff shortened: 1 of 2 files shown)


himself65 (approved):

van-sprundel:
> There is no CLA to sign for better-auth/better-auth
> (could not find linked item for owner better-auth and repo better-auth)

van-sprundel:
seems like the strategy got fixed in the meantime so I'm closing this PR

PR #6281, opened 2025-11-25
feat: option to redirect after `/sign-up/email` finished

jslno:
## Summary by cubic
Adds an optional redirectTo parameter to the sign-up API to enable redirect after a successful sign-up. The response now returns redirect and url to guide client-side navigation; no redirect when email verification is required.

- **New Features**
  - Request: accepts redirectTo in the body.
  - Response: adds redirect (boolean) and url (nullable).
  - Behavior: if email verification is required, returns redirect: false and token: null; otherwise sets redirect/url based on redirectTo.

<sup>Written for commit 14e097e53df3b3cddcc78736798e9ed5f1875e56. Summary will update automatically on new commits.</sup>


himself65 (changes requested):
could you please add a test case

Issue #6250, opened 2025-11-24
OpenAPI plugin generates invalid schema for sign-in endpoint causing syntax errors in code generation tools


viniciuspatzer:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Set up better-auth with the openAPI() plugin enabled
2. Generate the OpenAPI schema using auth.api.generateOpenAPISchema()
3. Export the schema to a JSON file and use it with a code generation tool like Orval
4. The syntax error will occur when Orval attempts to parse the schema

### Current vs. Expected behavior

**Current behavior:**
When using the OpenAPI schema generated by better-auth's openAPI() plugin with code generation tools like Orval, the following validation errors occur:

```text
⚠️  SyntaxError: Swagger schema validation failed.

#/paths/~1api~1auth~1sign-in~1email/post/responses/200/content/application~1json/schema/properties/url/type must be equal to one of the allowed values
#/paths/~1api~1auth~1sign-in~1email/post/responses/200/content/application~1json/schema/properties/url must have required property '$ref'
#/paths/~1api~1auth~1sign-in~1email/post/responses/200/content/application~1json/schema/properties/url must match exactly one schema in oneOf
#/paths/~1api~1auth~1sign-in~1email/post/responses/200/content/application~1json/schema must have required property '$ref'
#/paths/~1api~1auth~1sign-in~1email/post/responses/200/content/application~1json/schema must match exactly one schema in oneOf
#/paths/~1api~1auth~1sign-in~1email/post/responses/200 must have required property '$ref'
#/paths/~1api~1auth~1sign-in~1email/post/responses/200 must match exactly one schema in oneOf
```

The generated schema for the /sign-in/email endpoint contains invalid schema definitions where the url property in the response has an invalid or missing type field, and the overall response schema structure doesn't properly comply with OpenAPI specification requirements.

**Expected behavior:**
The OpenAPI schema generated by auth.api.generateOpenAPISchema() should produce valid OpenAPI 3.0/3.1 compliant schemas that pass validation in standard tools. Specifically:

- All properties should have valid type values
- Response schemas should properly reference components or follow oneOf requirements correctly
- The schema should be consumable by code generation tools like Orval

### What version of Better Auth are you using?

1.4.1

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
[...]

PR #6292, opened 2025-11-25
fix: clean up incorrect null type in OpenAPI

bytaesu:
Closes https://github.com/better-auth/better-auth/issues/6250

---
## Summary by cubic
Fixed incorrect OpenAPI schema types by replacing “null” with proper types and marking them as nullable. This resolves spec validation issues and prevents broken SDK/codegen.

- **Bug Fixes**
  - sign-in email: url is now type "string" with nullable: true.
  - organization rejectInvitation: member is now type "object" with nullable: true.

<sup>Written for commit 4db1da6d643ee68b53a9fe6173b8b9b586c422f6. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/sign-in.ts
+++ b/packages/better-auth/src/api/routes/sign-in.ts
@@ -380,7 +380,7 @@ export const signInEmail = createAuthEndpoint(
 											description: "Session token",
 										},
 										url: {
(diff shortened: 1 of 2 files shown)


bytaesu:
-> rewrite

PR #6314, opened 2025-11-25
feat: support form data for email sign-in/sign-up and fallback to checking fetch Metadata for first login

Paola3stefania:
## Summary by cubic
Add Fetch Metadata CSRF protection for first login. Blocks cross-site navigation during email sign-in/sign-up when the request has no cookies, while preserving origin checks and backward compatibility.

- **New Features**
  - Use Sec-Fetch-Site/Mode/Dest to detect cross-site navigation on first login to email sign-in/sign-up and return 403 with CROSS_SITE_NAVIGATION_LOGIN_BLOCKED.
  - Allow same-origin/same-site navigation and fetch/XHR requests (mode=cors).
  - Cookies/metadata handling: with cookies, use Origin/Referer validation; without cookies and with Fetch Metadata, also validate Origin/Referer against trustedOrigins; without cookies and no Fetch Metadata, fall back to previous behavior.
  - Accept application/x-www-form-urlencoded on email sign-in/sign-up to support browser form submissions.

- **Refactors**
  - Added formCsrf middleware and a stronger validateOrigin with wildcard and dynamic trustedOrigins; applied to email sign-in/sign-up routes.
  - Updated security docs and added tests covering email sign-in/sign-up flows, form submissions, and metadata scenarios.

<sup>Written for commit 108d1988cc20dc3ff2a44bdeb80694641acf6ffc. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/oauth-provider.mdx
+++ b/docs/content/docs/plugins/oauth-provider.mdx
@@ -49,7 +49,8 @@ This plugin is in active development and may not be suitable for production use.
 
     ```ts title="auth.ts"
     import { betterAuth } from "better-auth";
-    import { oauthProvider, jwt } from "@better-auth/oauth-provider";
+    import { jwt } from "better-auth/plugins";
+    import { oauthProvider } from "@better-auth/oauth-provider";
 
     const auth = betterAuth({
       disabledPaths: [
@@ -1776,6 +1777,21 @@ To improve lookup performance, database adapters may map the field `client_id` o
 Previously `oauthApplication`
 
 - If `storeClientSecret` was unset or `plain`, you must hash all the stored `clientSecret` values into its "SHA-256" representation then convert it into base64Url format or use another storage method specified by `storeClientSecret`.
+The following function will convert a `plain` representation into the default hash:
+```ts
+import { createHash } from "@better-auth/utils/hash";
+import { base64Url } from "@better-auth/utils/base64";
+
+const defaultHasher = async (value: string) => {
+	const hash = await createHash("SHA-256").digest(
+		new TextEncoder().encode(value),
+	);
+	const hashed = base64Url.encode(new Uint8Array(hash), {
+		padding: false,
+	});
+	return hashed;
+};
+```
 - `type` field is no longer a required field. Instead, the schema requires `public` of type `boolean`. Migrate with the following rules:
     - Clients with `type: "public"`: set `type: undefined`, `public: true`, and `clientSecret: undefined`
     - Clients with `type: "native"`: set `public: true` and `clientSecret: undefined`
@@ -1795,7 +1811,7 @@ Option 2 (more complex):
 - Convert `oauthAccessToken` with `refreshToken` field into a new `oauthRefreshToken` entry.
 ```ts
 {
-  token: hashed(refreshToken),
+  token: defaultHasher(refreshToken),
   expiresAt: refreshTokenExpiresAt,
   clientId: clientId,
   scopes: scopes,
@@ -1808,7 +1824,7 @@ Option 2 (more complex):
 - Keep `oauthAccessToken` but reference new `oauthRefreshToken`.
 ```ts
 {
-  token: hashed(accessToken),
+  token: defaultHasher(accessToken),
   expiresAt: accessTokenExpiresAt,
   clientId: clientId,
   scopes: scopes,
@@ -1828,3 +1844,4 @@ The MCP endpoints moved from `/mcp` to the `/oauth2` equivalent.
 - `/oauth2/register` (previously `/mcp/register`)
 - `/mcp/get-session` removed as not OAuth 2 compliant, use `/oauth2/introspect` instead
 - `/.well-known/oauth-protected-resource` removed, use the helper `mcpHandler` (or manually with the server `api.oAuth2introspectVerify` or the resource client `verifyAccessToken`)
+-  Database changes are equivalent to the [From OIDC Provider Plugin](#from-oidc-provider-plugin) section.


Paola3stefania:
## Breaking change — limited scope

**What changed:**
- Before: Cross-origin first-login (no session cookie) → no origin validation → worked without `trustedOrigins`
- After: Cross-origin first-login (browsers send Fetch Metadata) → origin validation now happens → requires `trustedOrigins` to include the frontend origin

**Who is affected:**

1. Cross-origin setups (frontend and backend on different origins):
   - Different ports: `http://localhost:5173` → `http://localhost:3000`
   - Different domains: `https://app.example.com` → `https://api.example.com`
   - Different subdomains: `https://www.example.com` → `https://api.example.com`
   - Only for first-login to email sign-in/sign-up endpoints

2. Not affected:
   - Same-origin setups: `baseURL` is automatically added to `trustedOrigins` (line 49 in `base.ts`)
   - Subsequent logins (session cookie exists → uses existing origin validation)
   - OAuth flows (different endpoints with their own security)
   - Non-browser clients (no Fetch Metadata → fallback to old behavior)

**Why tests need it:**
The tests use cross-origin setups:
- Integration test: Frontend on `localhost:5173`, backend on random port
- Smoke tests: Server on random port, requests from `localhost:PORT`

This matches real cross-origin scenarios that now require explicit `trustedOrigins` configuration.

**Summary:**
- This is a breaking change only for cross-origin first-login scenarios.
- Same-origin setups continue to work automatically.
- The test changes reflect what users with cross-origin setups need to configure.

packages/better-auth/src/api/middlewares/origin-check.ts
```diff
@@ -18,10 +20,19 @@ export const originCheckMiddleware = createAuthMiddleware(async (ctx) => {
...
 	}
-	const headers = ctx.request?.headers;
 	const request = ctx.request;
 	const { body, query, context } = ctx;
-	const originHeader = headers?.get("origin") || headers?.get("referer") || "";
+
+	// Only apply Fetch Metadata CSRF protection to email sign-in and sign-up endpoints
+	const isEmailAuthEndpoint =
+		ctx.path === "/sign-in/email" || ctx.path === "/sign-up/email";
+
```
Bekacru: We don’t know which plugins or endpoints will need this now or in the future, so hard-coding paths isn’t the right approach. Each endpoint should be able to declare whether it requires these additional security checks.

Instead of making this a global middleware, we should make it an endpoint-level middleware that each endpoint can import and apply as needed.

Bekacru (changes requested):

packages/better-auth/src/api/middlewares/origin-check.ts
Paola3stefania: Makes sense 100%, refactored !

packages/better-auth/src/api/middlewares/origin-check.ts
```diff
@@ -159,3 +153,171 @@ export const originCheck = (
 			validateURL(url, "callbackURL");
 		}
 	});
+
+/**
+ * Checks if a Better Auth session cookie exists in the request (cookie presence only, not validation).
+ */
+function hasBetterAuthSessionCookie(ctx: GenericEndpointContext): boolean {
```
Bekacru: we're not using this helper anywhere right? if so, we should remove it

Paola3stefania: Ohh yes I totally missed it. TY

Bekacru:
can we add test for both sign in and sign up?

Paola3stefania:
> can we add test for both sign in and sign up?

Yes ! I was missing this. 
**Added**
4 new tests in sign-in CSRF protection
4 new tests in sign-up CSRF protection 

The tests verify the **formCsrfMiddleware** is working correctly on both endpoints by testing:
Cross-site navigation blocking (CSRF attack pattern)
Same-origin navigation allowance (legitimate forms)
CORS mode requests (fetch/XHR API calls)
Cookie-based origin validation fallback

e2e/integration/solid-vinxi/src/lib/auth.ts
```diff
@@ -11,6 +11,11 @@ export const auth = betterAuth({
 	emailAndPassword: {
 		enabled: true,
 	},
+	trustedOrigins: [
+		baseURL,
```
jonathansamines: I seem to remember that `baseURL` is automatically added to the trusted origins list.

packages/better-auth/src/api/middlewares/origin-check.ts
```diff
@@ -88,3 +81,134 @@ export const originCheck = (
...
+
+	const shouldValidate =
+		forceValidate ||
+		(useCookies && !ctx.context.skipCSRFCheck && !ctx.context.skipOriginCheck);
+
+	if (!shouldValidate) {
+		return;
+	}
+
+	if (!originHeader || originHeader === "null") {
```
jonathansamines, line 167: Curious about this, is the check against the "null" string intentional?

packages/better-auth/src/api/middlewares/origin-check.ts
```diff
@@ -88,3 +81,134 @@ export const originCheck = (
...
+			}
+			return wildcardMatch(pattern)(host);
+		}
+		const protocol = getProtocol(url);
+		return protocol === "http:" || protocol === "https:" || !protocol
+			? pattern === getOrigin(url)
+			: url.startsWith(pattern);
+	};
+
+	const isTrustedOrigin = trustedOrigins.some((origin) =>
```
jonathansamines, line 180: I think you can now use the `context.isTrustedOrigin()` helper instead:

```suggestion
	const isTrustedOrigin = ctx.context.isTrustedOrigin(originHeader)
```

packages/better-auth/src/api/middlewares/origin-check.ts
Bekacru: yeah, origin header will be set to "null" in some scenarios

e2e/integration/solid-vinxi/src/lib/auth.ts
Bekacru: good catch, no need to add it manually.

Paola3stefania: Oh great ! changing that now. ty @jonathansamines

packages/better-auth/src/api/middlewares/origin-check.ts
Paola3stefania: some examples are
1. **Sandboxed iframes** - When an iframe has the `sandbox` attribute without `allow-same-origin`:
   ```html
   <iframe sandbox="allow-scripts" src="..."></iframe>
   ```
2. **Data URLs** - Requests originating from pages loaded via `data:` URLs

3. **File URLs** - Requests from local files (`file://`) in most browsers

4. **Cross-origin redirects** - After certain cross-origin redirect chains, the origin can become `"null"`

5. **Privacy-sensitive contexts** - Some browsers use `"null"` in privacy modes or when the origin is considered opaque

is part of the [HTML spec for opaque origins](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque) when serialized, opaque origins become the string `"null"`.

PR #6321, opened 2025-11-26
feat: add keyGenerator option for rate limit

k8adev:
Currently, the rate limiter generates keys based on the client's IP address configurable via `ipAddressHeaders`.

This works well for most cases, but there are scenarios where IP-based rate limiting is not ideal:

- NAT/CGNAT: Multiple users behind the same NAT gateway share the same public IP, causing false positives 
- Corporate networks: Thousands of users may share the same corporate IP
- VPN users: Multiple users share the same VPN exit node IP

I encountered this issue on Vercel, where IPv6 clients connect via NAT64 gateways with shared IPs since the platform doesn't support IPv6 directly. This causes multiple unrelated users to be rate-limited together.

## Solution

Add an optional `keyGenerator` function to the rate limit configuration:

```ts
    rateLimit: {
      keyGenerator: (request) => {
        return request.headers.get("x-fingerprint"); // or any other custom header
      },
    },
  });
```

The request path is automatically appended to the returned value.

## Changes
- Added keyGenerator option to BetterAuthRateLimitOptions type
- Implemented keyGenerator logic in rate limiter
- Added test for custom key generator
- Added documentation

## Backwards Compatibility
This is a non-breaking change. If `keyGenerator` is not provided, the current behavior is preserved.

---
## Summary by cubic
Adds an optional rateLimit.keyGenerator to build custom rate limit keys from the request (e.g., a header) instead of the client IP. This avoids shared-IP throttling and falls back to IP if the generator is missing, empty, or errors; the request path is always appended.

- **New Features**
  - Added rateLimit.keyGenerator to BetterAuthRateLimitOptions.
  - Implemented key generation logic in the rate limiter with path suffixing.
  - Added tests and documentation for custom key generation.

<sup>Written for commit f151c19345f83d0101e9dcc0fba61aba095d114d. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/rate-limiter/index.ts
+++ b/packages/better-auth/src/api/rate-limiter/index.ts
@@ -141,22 +141,21 @@ export async function onRequestRateLimit(req: Request, ctx: AuthContext) {
 	);
 	let window = ctx.rateLimit.window;
 	let max = ctx.rateLimit.max;
-	let key: string = "";
+	let key: string | null = null;
 
 	if (ctx.options.rateLimit?.keyGenerator) {
 		try {
 			const generatedKey = await ctx.options.rateLimit.keyGenerator(req);
 
-			if (!generatedKey?.length) {
-				return;
+			if (generatedKey?.length) {
+				key = generatedKey;
 			}
-
-			key = generatedKey;
 		} catch (e) {
 			ctx.logger.error("Error generating rate limit key", e);
-			return;
 		}
-	} else {
+	}
+
+	if (!key) {
 		const ip = getIp(req, ctx.options);
 
 		if (!ip) {
--- a/docs/content/docs/concepts/rate-limit.mdx
+++ b/docs/content/docs/concepts/rate-limit.mdx
@@ -85,7 +85,7 @@ export const auth = betterAuth({
 ```
 
 <Callout>
-The request path is automatically appended to the returned value.
+The request path is automatically appended to the returned value. If the function returns an empty value or throws an error, the IP address will be used as fallback.
 </Callout>
 
 ### Rate Limit Window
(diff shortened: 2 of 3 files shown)


packages/better-auth/src/api/rate-limiter/index.ts
```diff
@@ -141,11 +141,33 @@ export async function onRequestRateLimit(req: Request, ctx: AuthContext) {
...
+			const generatedKey = await ctx.options.rateLimit.keyGenerator(req);
+
+			if (!generatedKey?.length) {
+				return;
+			}
+
+			key = generatedKey;
+		} catch (e) {
+			ctx.logger.error("Error generating rate limit key", e);
+			return;
```
himself65: fallback to default logic

packages/better-auth/src/api/rate-limiter/index.ts
```diff
@@ -141,11 +141,33 @@ export async function onRequestRateLimit(req: Request, ctx: AuthContext) {
...
-	if (!ip) {
-		return;
+	let key: string = "";
+
+	if (ctx.options.rateLimit?.keyGenerator) {
+		try {
+			const generatedKey = await ctx.options.rateLimit.keyGenerator(req);
+
+			if (!generatedKey?.length) {
+				return;
```
himself65: shouldn't early return

packages/better-auth/src/api/rate-limiter/index.ts
k8adev: Done! I also added a test case to cover the fallback

k8adev:
Hi @himself65, could you please review it again? Is there anything else I need to do?

Issue #6297, opened 2025-11-25
Redirect Issue with `CallbackURL` in Username Plugin


OwenChia:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

This library is incredibly well-designed and has significantly enhanced the authentication process in my applications. 
Thanks for your  excellent work.

While Better-Auth offers many useful features, I have encountered some unexpected behavior.
When using the username plugin from Better-Auth in a Next.js application, the callbackURL parameter fails to function correctly. 
After examining the codebase, it appears that the logic for processing the callbackURL is missing, leading to unexpected behavior where the user is not redirected to the desired URL upon successful login.

Steps to Reproduce
  1. Add username() and usernameClient() to the plugin list in your configuration.
  2. Attempt to log in using valid username and password credentials.
  3. After a successful login, observe that no redirection occurs to the specified callbackURL.

In this code, it only return two field: `token` and `user`, no `redirect` and `url`,
https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/plugins/username/index.ts#L395-L408
```javascript
return ctx.json({
		token: session.token,
		user: {
			id: user.id,
			email: user.email,
			emailVerified: user.emailVerified,
			username: user.username,
			displayUsername: user.displayUsername,
			name: user.name,
			image: user.image,
			createdAt: user.createdAt,
			updatedAt: user.updatedAt,
		},
	});
```

I think it should be like this:
```javascript
return ctx.json({
		redirect: !!ctx.body.callbackURL,
		url: ctx.body.callbackURL,
		token: session.token,
		user: {
			id: user.id,
			email: user.email,
			emailVerified: user.emailVerified,
			username: user.username,
			displayUsername: user.displayUsername,
			name: user.name,
			image: user.image,
			createdAt: user.createdAt,
			updatedAt: user.updatedAt,
		},
	});
```

Also, the test case for `callbackURL` is not included.

### Current vs. Expected behavior

+ Current Behavior
  The application does not redirect users to the specified callbackURL after a successful login with the username plugin. As a result, users remain on the same page.

+ Expected Behavior
  After a successful login, users should be redirected to the callbackURL specified in the request body. Additionally, there should be test cases included to verify the functionality of the callbackURL.

### What version of Better Auth are you using?

1.4.1

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
[...]

PR #6330, opened 2025-11-26
fix(userName-Plugin) implement-callback-url-handling

Ridhim-RR:
**Closes** #6297

This PR implements the handling of the callback URL in the Username plugin.

**Changes**:
- In the index.ts file within the sign-in with Username flow, implement handling of the callback URL if it is sent from the frontend.
- If no callback URL is provided, fall back to the root path "/".
- The response now includes url field which will include the callback url along with token and duser Information.

Updated Response
```
return ctx.json({
  url: callbackURL,
  token: session.token,
  user: {
    id: user.id,
    email: user.email,
    emailVerified: user.emailVerified,
    username: user.username,
    displayUsername: user.displayUsername,
    name: user.name,
    image: user.image,
    createdAt: user.createdAt,
    updatedAt: user.updatedAt,
  },
});
```
Previous Response
```
return ctx.json({
  token: session.token,
  user: {
    id: user.id,
    email: user.email,
    emailVerified: user.emailVerified,
    username: user.username,
    displayUsername: user.displayUsername,
    name: user.name,
    image: user.image,
    createdAt: user.createdAt,
    updatedAt: user.updatedAt,
  },
});

```

- Added two test cases in username.test.ts:
      - One to verify that, if a URL is provided in the request body, it is returned in the response.
      - Another to check that, if no URL is provided, the fallback URL ("/") is sent back.

---
## Summary by cubic
Adds callback URL handling to the Username sign-in flow, validates its origin, and returns it as url in the response for post-auth redirects. If no callbackURL is provided, the response defaults to "/" (addresses #6297).

<sup>Written for commit d0f8865e3f50f27179eec6af4436a604c75658e7. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/components/sidebar-content.tsx
+++ b/docs/components/sidebar-content.tsx
@@ -495,7 +495,6 @@ export const contents: Content[] = [
 			{
 				title: "Atlassian",
 				href: "/docs/authentication/atlassian",
-				isNew: true,
 				icon: () => (
 					<svg
 						xmlns="http://www.w3.org/2000/svg"
@@ -518,7 +517,6 @@ export const contents: Content[] = [
 			{
 				title: "Cognito",
 				href: "/docs/authentication/cognito",
-				isNew: true,
 				icon: () => (
 					<svg
 						xmlns="http://www.w3.org/2000/svg"
@@ -571,7 +569,6 @@ export const contents: Content[] = [
 			{
 				title: "Figma",
 				href: "/docs/authentication/figma",
-				isNew: true,
 				icon: () => (
 					<svg
 						width="1.2em"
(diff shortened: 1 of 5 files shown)


ping-maxwell:
Maybe we should be also sending a 302 redirect if the callbackURL is provided and if it was a successful sign-in

Ridhim-RR:
> Maybe we should be also sending a 302 redirect if the callbackURL is provided and if it was a successful sign-in

@ping-maxwell  So  when the redirtect url is provided then we will sending 302 redirect to the user and user will get the token , user info from the session. And in case redirect url is not provided then we will send our default json resposne . Should I implement this or is there something else

himself65:
Closing this PR as duplicate. Newer PRs #6331 and #7200 implement the same fix (username callbackURL handling). Please review and comment on the newer PRs instead. Thank you for your contribution!

PR #6345, opened 2025-11-27
fix(stripe): properly merge subscription_data to preserve free trial

mohamede1945:
## Summary
- Fixes an issue where using `getCheckoutSessionParams` to add custom `subscription_data` (e.g., description or metadata) would overwrite the free trial settings
- Ensures `subscription_data` from user params is properly merged with `freeTrial` instead of replacing it entirely

## Problem
When adding custom subscription data like this:
```ts
return {
  params: {
    subscription_data: {
      metadata: {
        organization_name: organization.slug,
      },
      description: `Product XYZ ${planName} subscription for ${organization.slug}`,
    },
  },
};
```

The free trial would be disabled because `...params?.params` was spread after `subscription_data`, completely replacing it.

## Solution
Move the `subscription_data` object to be defined after `...params?.params` and properly merge both the `freeTrial` settings and any user-provided `subscription_data` options.

## Test plan
- [x] Verify that free trial works when no custom `subscription_data` is passed
- [x] Verify that free trial works when custom `subscription_data` (description, metadata) is passed via `getCheckoutSessionParams`

---
## Summary by cubic
Fixes a bug where custom subscription_data passed to getCheckoutSessionParams disabled the free trial. We now merge user subscription_data with the freeTrial settings, preserving trials while keeping custom description/metadata.

<sup>Written for commit 1367ab728681b9d46afbf1647e72a07d14585014. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/index.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/index.ts
@@ -408,10 +408,13 @@ export const oidcProvider = (options: OIDCOptions) => {
 
 					if (!consentCode) {
 						// Check for cookie-based consent flow
-						consentCode = await ctx.getSignedCookie(
+						const cookieValue = await ctx.getSignedCookie(
 							"oidc_consent_prompt",
 							ctx.context.secret,
 						);
+						if (cookieValue) {
+							consentCode = cookieValue;
+						}
 					}
 
 					if (!consentCode) {


mohamede1945:
I've used cc to generate the PR description and commit, but I did the change myself and verified it fixed my issue. 

Thanks!

PR #6346, opened 2025-11-27
fix(oidc-provider): resolve getSignedCookie return type

bytaesu:
Return type updated -> https://github.com/Bekacru/better-call/pull/81

## Summary by cubic
Handled nullable return from getSignedCookie in the OIDC consent flow. We now set consentCode only when the "oidc_consent_prompt" cookie exists, resolving type errors and preventing undefined assignments.

<sup>Written for commit 4d5853c890332a8cd96136794622b41ec08cd002. Summary will update automatically on new commits.</sup>


bytaesu:
This PR resolves all remaining type mismatches from the `better-call` update. cc @Bekacru

PR #6352, opened 2025-11-27
Fix/admin plugin get user additional fields

### LINKED ISSUE #6318: Admin plugin getUser endpoint does not return typed additional fields

### To Reproduce

Hi!
I'm using the admin plugin and noticed that the getUser endpoint is not returning the full typing for custom/additional user fields.

```typescript
import { headers } from "next/headers"
import { auth } from "@/lib/auth/auth"

export async function getUser(userId: string) {
  try {
    const user = await auth.api.getUser({
      query: { id: userId },
      headers: await headers(),
    })
    return user
  } catch (error) {
    console.error(`User not found: ${userId}`, error)
    return null
  }
}
```

### Actual typing returned

The response is typed only as:

```tyepscript
const user: {
  id: string
  createdAt: Date
  updatedAt: Date
  email: string
  emailVerified: boolean
  name: string
  image?: string | null | undefined
}
```

auth.api.getUser should return the full inferred type of the User model, including any added fields defined in the Better Auth configuration.

1.4.2

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "x64",
    "version": "Darwin Kernel Version 25.1.0: Mon Oct 20 19:26:51 PDT 2025; root:xnu-12377.41.6~2/RELEASE_X86_64",
    "release": "25.1.0",
    "cpuCount": 12,
    "cpuModel": "Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz",
    "totalMemory": "16.00 GB",
    "freeMemory": "0.04 GB"
  },
  "node": {
    "version": "v24.11.1",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.2"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.0.3"
    },
    {
      "name": "react",
      "version": "19.2.0"
    }
  ],
  "databases": [
    {
      "name": "@prisma/client",
      "version": "6.18.0"
    }
  ],
  "betterAuth": {
    "version": "1.4.2",
    "config": null
  }
}
```

Backend

```typescript
  appName: "Template",
  database: prismaAdapter(db, {
    provider: "postgresql",
  }),
  user: {
    deleteUser: {
      enabled: true,
    },
    additionalFields: {
      bio: {
        type: "string",
        required: false,
        fieldName: "bio",
      },
    },
  },
...
```

### Additional context

_No response_

Ridhim-RR:
Closes #6318
## Problem

The `getUser` endpoint in the admin plugin was not returning custom/additional user fields defined in `user.additionalFields` configuration. While the fields existed in the database, they were not included in the API response.

**Example:**
```typescript
// Config
user: {
additionalFields: {
bio: { type: "string", required: false }
}
}


// API call
const user = await auth.api.admin.getUser({ query: { id: userId } });
// ❌ user.bio is undefined, even though it exists in the database
```

## Root Cause

The `findUserById` method in the internal adapter was constrained to return only the base `User` type:

```typescript
// Before
findUserById: async (userId: string) => {
const user = await adapter.findOne<User>({ ... }); // ❌ Only base User type
return user;
}
```

This type constraint caused the adapter to potentially filter out fields that weren't part of the base `User` schema, even though `parseUserOutput` was designed to handle additional fields.

## Solution

Updated `findUserById` to return `User & Record<string, any>` instead of just `User`, allowing all fields from the database to be included in the response.

### 1. Updated Type Definition
**File:** `packages/core/src/types/context.ts`

```typescript
// Before
findUserById(userId: string): Promise<User | null>;


// After
findUserById(userId: string): Promise<User & Record<string, any> | null>;
```

### 2. Updated Implementation
**File:** `packages/better-auth/src/db/internal-adapter.ts`

```typescript
// Before
findUserById: async (userId: string) => {
const user = await adapter.findOne<User>({ ... });
return user;
}


// After
findUserById: async (userId: string) => {
const user = await adapter.findOne<User & Record<string, any>>({ ... });
return user;
}
```

## How It Works

1. **Database Query**: `findOne<User & Record<string, any>>` tells the adapter to return all fields from the database, including additional fields like `bio`.

2. **Type Safety**: `User & Record<string, any>` preserves the base `User` type while allowing additional fields, maintaining type safety for known fields.

3. **Field Processing**: `parseUserOutput` receives the complete user object (including `bio`) and processes it according to the schema configuration, respecting `returned: false` flags and applying output transformations.

4. **Response**: The endpoint returns the full user object with all additional fields included.

## Testing

Added test case to verify additional fields are returned:

```typescript
it("should return additional fields when getting user", async () => {
// Update user with bio field
await client.admin.updateUser({
userId: testNonAdminUser.id,
data: { bio: "This is my bio" },
}, { headers: adminHeaders });


// Get user
const res = await client.admin.getUser({
query: { id: testNonAdminUser.id },
}, { headers: adminHeaders });


// Assert bio is present
expect(res.data?.bio).toBe("This is my bio");
});
```

---
## Summary by cubic
Fixes admin.getUser so it returns custom fields defined in user.additionalFields (e.g., bio), ensuring the full user object is included in the response.

- **Bug Fixes**
  - Widened InternalAdapter.findUserById to return User & Record<string, any> so extra fields aren’t dropped.
  - Updated admin.getUser to parse and return User & Record<string, any>.
  - Fixed admin.updateUser to handle additional fields by widening the update type.
  - Added a test to confirm additional fields like bio are returned.

<sup>Written for commit a5e5343d99b81c2fc83f2f3f30db643ea52ab2b1. Summary will update automatically on new commits.</sup>


packages/better-auth/src/plugins/admin/admin.ts
```diff
@@ -350,7 +350,9 @@ export const admin = <O extends AdminOptions>(options?: O | undefined) => {
 						});
 					}
 
-					return parseUserOutput(ctx.context.options, user);
+					return parseUserOutput(ctx.context.options, user) as InferUser<
+						typeof ctx.context.options
```
Bekacru: `ctx.context.options` doesnt actually infer the options the user passed

douglasndm:
Is there any chance this be available in the next release? I'm having this issue for some time

Ridhim-RR:
> Is there any chance this be available in the next release? I'm having this issue for some time

tag himself65 or bekacru

douglasndm:
Hi @himself65, could you check that?

Issue #5539, opened 2025-10-24
Failed to fetch dynamically imported module: node:async_hooks


qweered:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

I've updated from 1.3.27 to 1.3.29 and saw that error. toNextjsHandler configured correctly. I have proxy.ts file instead of middleware.ts. 1.3.27 with proxy.ts also works correctly, not tested 1.3.29 

```ts
export async function proxy(request: NextRequest) {
    const session = await auth.api.getSession({
        headers: await headers()
    })

    if (!session) {
        return NextResponse.redirect(new URL('/login', request.url))
    }

    return NextResponse.next()
}

export const config = {
    matcher: ['/dashboard', '/profile', '/settings', '/create-organization', '/invitations']
}
```

### Current vs. Expected behavior

Do not have error

```bash
 ✓ Ready in 3.4s
[browser] [better-auth] Warning: AsyncLocalStorage is not available in this environment. Some features may not work as expected. (file:///home/qweered/Projects/automaspec/.next/dev/static/chunks/02ec3_next_dist_91e776ba._.js:2294:13)
[browser] [better-auth] Please read more about this warning at https://better-auth.com/docs/installation#mount-handler (file:///home/qweered/Projects/automaspec/.next/dev/static/chunks/02ec3_next_dist_91e776ba._.js:2294:13)
[browser] [better-auth] If you are using Cloudflare Workers, please see: https://developers.cloudflare.com/workers/configuration/compatibility-flags/#nodejs-compatibility-flag (file:///home/qweered/Projects/automaspec/.next/dev/static/chunks/02ec3_next_dist_91e776ba._.js:2294:13)
[browser] ⨯ unhandledRejection: TypeError: Failed to fetch dynamically imported module: node:async_hooks
TypeError: Failed to fetch dynamically imported module: node:async_hooks
```

### What version of Better Auth are you using?

1.3.29

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#1-NixOS SMP PREEMPT_DYNAMIC Wed Oct 15 10:04:23 UTC 2025",
    "release": "6.17.3-cachyos",
    "cpuCount": 8,
    "cpuModel": "AMD Ryzen 3 5300U with Radeon Graphics",
    "totalMemory": "14.95 GB",
    "freeMemory": "5.20 GB"
  },
  "node": {
    "version": "v22.19.0",
    "env": "development"
[...]

PR #6361, opened 2025-11-27
fix: avoid throwing on client side

landoncolburn:
When importing `async_hooks` on client side, the `AsyncLocalStoragePromise` throws since `AsyncLocalStorage` is not available in the browser. This is semi-intended behaviour since `AsyncLocalStoragePromise` should not be used on the browser, but the throw should take place on usage, not on import (some files may run on both client and server).

I changed the type signature of `AsyncLocalStoragePromise` to `Promise<typeof AsyncLocalStorage | null>`, and then throw in the accessor method `getAsyncLocalStorage` if `AsyncLocalStoragePromise` is `null`.

This should resolve https://github.com/better-auth/better-auth/issues/5539

---
## Summary by cubic
Prevent client-side imports of node:async_hooks from throwing by deferring the error to runtime. The error now only occurs when getAsyncLocalStorage is called in the browser. Resolves better-auth/better-auth#5539.

- **Bug Fixes**
  - Updated AsyncLocalStoragePromise type to Promise<typeof AsyncLocalStorage | null>.
  - Return null on client (window detected); keep server warnings and throw only when unavailable server-side.
  - Throw a clear error from getAsyncLocalStorage when used in browser; server behavior unchanged.

<sup>Written for commit abe97ae310af898e515491b37a432f546ceec6a4. Summary will update automatically on new commits.</sup>


packages/core/src/async_hooks/index.ts
```diff
@@ -6,30 +6,39 @@ import type { AsyncLocalStorage } from "node:async_hooks";
...
+		.catch((err) => {
+			if ("AsyncLocalStorage" in globalThis) {
+				return (globalThis as any).AsyncLocalStorage;
+			}
+			if (typeof window !== "undefined") {
+				return null;
+			}
+			console.warn(
+				"[better-auth] Warning: AsyncLocalStorage is not available in this environment. Some features may not work as expected.",
+			);
```
fugisakimatheus: **issue:** This console prompt is not needed right now, as you have already added it below:

```
if (mod === null) {
		throw new Error("getAsyncLocalStorage is only available in server code");
}
```

landoncolburn, line 25: We return early in the catch block, so the console.warn that you are referencing is never called. We return the error because there is no AsyncLocalStorage to return, and returning a stub feels bad practice as it shouldn't be used on the client.

TL;DR throwing an error there is the only correct way to handle this.

landoncolburn:
@himself65 @Bekacru This bug is affecting users trying to use the Polar integration along with various other client side plugins. Can we get this looked at when you have a moment?

PR #6366, opened 2025-11-27
Fix: mysql date generating

MarkForLoop:
https://dev.mysql.com/doc/refman/9.4/en/datetime.html
MySQL date fields generated as `timestamp` instead of `datetime`

DATETIME
- Format: YYYY-MM-DD hh:mm:ss
- Range: '1000-01-01 00:00:00' → '9999-12-31 23:59:59'

TIMESTAMP
- Range: '1970-01-01 00:00:01' UTC → '2038-01-19 03:14:07' UTC

Because `timestamp` has a far more limited valid range, this causes issues for users who need broader date support. The schema generator should default to `datetime()` instead of `timestamp()` for MySQL date fields.

---
## Summary by cubic
Fix MySQL date/time generation to use datetime with fsp: 3, proper CURRENT_TIMESTAMP(3) defaults, and correct .$onUpdate behavior. This produces valid schemas with millisecond precision and correct imports.

- **Bug Fixes**
  - Generate datetime(...) for MySQL date fields instead of timestamp.
  - Use default(sql`CURRENT_TIMESTAMP(3)`) and ensure .$onUpdate works for MySQL dates (sets CURRENT_TIMESTAMP(3) when using onUpdate(new Date())).
  - Update imports to include datetime and sql for MySQL.
  - Harden cookie parsing in multi-session and OIDC consent to avoid false values.

<sup>Written for commit e833fa27b4e63c8c82e520234f57b796fa02c7a1. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/cli/test/__snapshots__/auth-schema-mysql-enum.txt
+++ b/packages/cli/test/__snapshots__/auth-schema-mysql-enum.txt
@@ -35,6 +35,7 @@ export const session = mysqlTable(
       .default(sql`CURRENT_TIMESTAMP(3)`)
       .notNull(),
     updatedAt: datetime("updated_at", { fsp: 3 })
+      .default(sql`CURRENT_TIMESTAMP(3)`)
       .$onUpdate(() => /* @__PURE__ */ new Date())
       .notNull(),
     ipAddress: text("ip_address"),
@@ -66,6 +67,7 @@ export const account = mysqlTable(
       .default(sql`CURRENT_TIMESTAMP(3)`)
       .notNull(),
     updatedAt: datetime("updated_at", { fsp: 3 })
+      .default(sql`CURRENT_TIMESTAMP(3)`)
       .$onUpdate(() => /* @__PURE__ */ new Date())
       .notNull(),
   },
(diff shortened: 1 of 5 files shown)


ping-maxwell:
we used to use datetime for mysql I believe.
The reason why we can't use datetime is because it doesn't support milliseconds.

MarkForLoop:
> we used to use datetime for mysql I believe.
> The reason why we can't use datetime is because it doesn't support milliseconds.
@ping-maxwell 

MySQL's datetime actually does support fractional seconds, including milliseconds. You can specify the precision like this:
```SQL
DATETIME(3) -- milliseconds (3 decimal places)
DATETIME(6) -- microseconds (6 decimal places)
```
https://dev.mysql.com/doc/refman/9.5/en/fractional-seconds.html

that being said, fractional second support in `datatime` was added in MySQL `5.6.4` (which was on 2011-12-20), so if you were working with an older version, that would have been a real limitation.

I included the `fps` for the fractional second here: https://github.com/better-auth/better-auth/blob/e833fa27b4e63c8c82e520234f57b796fa02c7a1/packages/cli/src/generators/drizzle.ts#L139

ping-maxwell:
Oh you're right, I did discover this a few months ago when I changed it. By default however it doesn't, so that must have been what confused me. If I recall, after finding out that even if I can get ms to store, it doesn't support timezone conversions either - which is a requirement we need. 

We need to find a good place in code to convert dates to UTC before storing it.

himself65:
Closing this PR due to merge conflicts that have not been resolved for over a month.

If you'd like to continue with this fix:
- Please rebase your branch on the latest `canary`
- Resolve all merge conflicts
- Open a new PR

Thank you for your contribution!

PR #6377, opened 2025-11-28
feat(anonymous): delete anonymous user endpoint

ping-maxwell:
Currently, there is no way to delete an anonymous user. Calling authClient.deleteUser requires a password, which anonymous users have not set.

This PR introduces a `/delete-anonymous-user` endpoint to allow such.

https://discord.com/channels/1288403910284935179/1443668195037679616

---
## Summary by cubic
Adds a POST /delete-anonymous-user endpoint so anonymous users can delete their account and session without a password. Updates docs and error codes, and respects the plugin option to disable deletion.

- **New Features**
  - Endpoint: POST /delete-anonymous-user; requires a valid session; deletes the user and clears the session cookie; returns { success: true }.
  - Guards: only works for anonymous users; errors for disabled deletion, non‑anonymous users, or failures.
  - Error codes: FAILED_TO_DELETE_ANONYMOUS_USER, USER_IS_NOT_ANONYMOUS, DELETE_ANONYMOUS_USER_DISABLED.
  - Docs: added “Delete Anonymous User” section and notes.

- **Migration**
  - disableDeleteAnonymousUser now also blocks access to /delete-anonymous-user, in addition to stopping automatic deletion after account linking. Review this option if you rely on the new endpoint.

<sup>Written for commit 9c1b12faa917b225e4805963c4ea8c7021f22314. Summary will update on new commits.</sup>

### DIFF
--- a/docs/content/docs/concepts/session-management.mdx
+++ b/docs/content/docs/concepts/session-management.mdx
@@ -247,6 +247,53 @@ await auth.api.getSession({
 });
 ```
 
+## Sessions in Secondary Storage
+
+By default, if you provide a [secondary storage](/docs/concepts/database#secondary-storage) in your auth configuration, the session will be stored in the secondary storage.
+
+```ts
+betterAuth({
+  // ... other options
+  secondaryStorage: {
+    // Your implementation here
+  },
+});
+```
+
+### Storing Sessions in the Database
+
+By default, Better Auth already stores sessions in the database, however if you provide a secondary storage,
+Better Auth will store sessions in the secondary storage instead of the database.
+
+You can choose to store sessions in the database instead of secondary storage by passing
+`storeSessionInDatabase: true` in the session configuration.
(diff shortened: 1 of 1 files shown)


Bartek532 (approved):

packages/better-auth/src/plugins/anonymous/index.ts
```diff
@@ -126,6 +134,95 @@ export const anonymous = (options?: AnonymousOptions | undefined) => {
...
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+				async (ctx) => {
+					const session = ctx.context.session as typeof ctx.context.session & {
```
Bekacru: define `AnonymousSession` or some kind of helper type

ping-maxwell: done

PR #6384, opened 2025-11-28
feat(email): add requireVerificationOnOldEmail option

### LINKED ISSUE #3742: Configurable Email-Address Change Verification Flow

Currently, when users change their email address in Better Auth, the system automatically sends two verification emails:
1. A verification email to the OLD email address (via `sendVerificationEmail`)
2. A verification email to the NEW email address (via `changeEmail.sendChangeEmailVerification`)

This dual-verification approach creates significant real-world problems:

### User Impact
- **Blocked email changes**: Users who no longer have access to their old email address cannot complete the verification process and must contact support for manual email changes
- **Support burden**: Organizations must handle support tickets for what should be a self-service user action
- **Confusing experience**: Users receive two verification emails and may not understand why both are necessary
- **Security concerns**: Users may ignore legitimate verification emails thinking they're duplicates

### Developer Impact
- **Fragile workarounds required**: Developers must implement URL-based detection (`request?.url.includes('/profile')`) to prevent unwanted verification emails
- **Undocumented coupling**: The relationship between signup and email change verification flows is not clearly documented
- **Limited customization**: No built-in way to control when old email verification is sent
- **Testing difficulties**: Email verification flows are hard to test, making these issues harder to discover

### Primary Feature: `sendOldEmailVerification` Option

Add a new configuration option `sendOldEmailVerification` (default: `true` for backward compatibility) that allows developers to control whether verification emails are sent to the old email address during email changes:

```javascript
  emailAndPassword: {
    enabled: true,
    requireEmailVerification: true
  },
  emailVerification: {
    sendOldEmailVerification: false, // NEW OPTION - prevents old email verification
      // Would only be called for actual signups when sendOldEmailVerification: false
      console.log("Sending signup verification email");
      await sendCustomEmail(user.email, url);
    },
    sendChangeEmailVerification: async ({ user, newEmail, url }, request) => {
      // Always called for email changes - this is the primary customization point
      console.log("Sending email change verification");
      await sendCustomEmail(newEmail, url);
    }
  }
});
```

### Current Workaround (Fragile)

Developers currently must implement URL-based detection to achieve similar behavior:

```javascript
  // Workaround: Check URL to detect email change flow
  const isEmailChange = request?.url.includes('/profile')
  if (!isEmailChange) {
    // Only send verification email for actual signups, not email changes
    await sendCustomEmail(user.email, url);
  }
}
```

### Secondary Improvements

1. **Context Detection**: Provide a built-in way to detect email change vs signup contexts without URL inspection

2. **Clear Documentation**: Document the email change flow, including when each email is sent and how to customize the behavior

3. **Better Debugging**: Add context information to email handlers to make flow identification easier

4. **Migration Path**: Provide clear guidance for existing applications that may rely on the current dual-email behavior

### Alternative 1: Always Single Email
- **Pros**: Simpler user experience
- **Cons**: Breaking change for applications that rely on old email verification

### Alternative 2: Configuration at Call Time
- **Pros**: Maximum flexibility per email change
- **Cons**: More complex API, harder to configure consistently

### Alternative 3: Role-Based Configuration
- **Pros**: Different behavior for different user types
- **Cons**: Adds complexity without addressing the core use case

### Real-World Use Cases for Configurable Verification
- **Consumer applications**: Users frequently lose access to old email addresses (work changes, expired accounts, etc.)
- **Internal tools**: Old email verification adds unnecessary friction for authenticated users
- **Migration scenarios**: Importing users from other systems where old emails may be invalid
- **Self-service portals**: Authenticated users should be trusted to change their email without dual verification
- **Support reduction**: Eliminate support tickets for users who can't access old email addresses

### Current Behavior Analysis
- Both `sendVerificationEmail` and `sendChangeEmailVerification` are called during email changes
- The `sendOnSignUp: false` setting does NOT prevent email change verification (contrary to expectations)
- `sendChangeEmailVerification` works correctly and is the primary customization point for new email verification
- The coupling between signup and email change flows is undocumented and counterintuitive

### Backward Compatibility Strategy
- Default `sendOldEmailVerification: true` maintains current dual-email behavior
- Existing applications continue working without changes
- Migration is opt-in and controlled by developers
- No breaking changes to existing APIs

### Implementation Considerations
- Should work seamlessly with existing `changeEmail.sendChangeEmailVerification`
- Must eliminate the need for URL-based workarounds: `request?.url.includes('/profile')`
- Should provide clear context to email handlers about the verification type
- Consider adding a `context` parameter to email handlers: `{ context: 'signup' | 'email-change' }`

### Expected Benefits
- **Reduced support burden**: Users can change emails without manual intervention
- **Better UX**: Clear, single-purpose verification emails
- **Cleaner code**: No need for fragile URL-based workarounds
- **Flexible security**: Organizations can choose appropriate verification levels
- **Easier testing**: More predictable email verification behavior

This feature would significantly improve both developer experience and user experience for email address changes while maintaining full backward compatibility for existing applications.

yoshifumi-kondo:
## Summary

- Add `requireVerificationOnOldEmail` option to `user.changeEmail` config
- When set to `false`, skips sending confirmation email to the old email address and sends verification directly to the new email
- Useful when users no longer have access to their old email address
- Default is `true` to maintain backward compatibility and security

Closes #3742

### Naming Decision
The original issue proposed `sendOldEmailVerification` under `emailVerification`, but I chose `requireVerificationOnOldEmail` under `user.changeEmail` because:
1. It sits alongside related options (`sendChangeEmailConfirmation`, `sendChangeEmailVerification`)
2. The name describes *what is required* rather than *what is sent*
3. Maintains consistency with the existing config structure

Happy to change the naming if maintainers prefer the original proposal.

### Security Consideration
As @himself65 noted in the issue, old email verification adds security against account hijacking. The default `true` preserves this security. Developers opting out should understand the trade-off.

## Test Plan

- [x] Added test for `requireVerificationOnOldEmail: false` - verifies old email confirmation is skipped and new email gets direct verification
- [x] Added test for default behavior (`true`) - verifies old email confirmation is still called
- [x] Existing tests pass (1187 passed, 7 failed in unrelated Vercel Provider tests)

---
## Summary by cubic
Added the requireVerificationOnOldEmail option to user.changeEmail to control whether we confirm via the old email before verifying the new one. When false, we skip the old email confirmation and send verification directly to the new email; default stays true for security and backward compatibility.

- **New Features**
  - New config: user.changeEmail.requireVerificationOnOldEmail (default true).
  - changeEmail endpoint skips old-email confirmation when false and sends new-email verification.
  - Types updated and tests added for both behaviors.

- **Migration**
  - No changes needed. To allow direct new-email verification, set requireVerificationOnOldEmail: false in user.changeEmail.

<sup>Written for commit 156ecfe6fddea6927e6d028008f1ad2c4a4e447a. Summary will update automatically on new commits.</sup>


ping-maxwell:
Hey it's been a while since this PR has made progress, I'll close it for now as stale.

PR #6386, opened 2025-11-28
fix(org): update client schema types to allow all server schema options

jonathansamines:
**What is changing?**
Updating organization client to support al server schema options. Otherwise, inference breaks as soon as you use an option that is only available to the server.

See [related issue](https://discord.com/channels/1288403910284935179/1443985134452998239).

---
## Summary by cubic
Aligns the organization client schema types with the server schema so type inference doesn’t break when using server-only options. The client now accepts all server schema options.

- **Bug Fixes**
  - Use OrganizationOptions["schema"] for OrganizationClientOptions.schema to match server types.
  - Update test to include organizationRole.modelName to reflect the expanded schema.

<sup>Written for commit da9784fd353af230f0887e5779f6befbbcfc6f96. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/organization/client.ts
+++ b/packages/better-auth/src/plugins/organization/client.ts
@@ -1,5 +1,4 @@
 import type { BetterAuthClientPlugin } from "@better-auth/core";
-import type { DBFieldAttribute } from "@better-auth/core/db";
 import { atom } from "nanostores";
 import { useAuthQuery } from "../../client";
 import type {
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/plugins/organization/client.ts
```diff
@@ -42,35 +41,7 @@ interface OrganizationClientOptions {
...
-					};
-				};
-				organizationRole?: {
-					additionalFields?: {
-						[key: string]: DBFieldAttribute;
-					};
-				};
-		  }
-		| undefined;
+	schema?: OrganizationOptions["schema"];
```
jonathansamines, line 44: As an alternative, if we don't want to explicitly allow the use of all other server supported options, we can extend the existing client options to allow any other field, e.g:

```diff
organization?: {
					additionalFields?: {
						[key: string]: DBFieldAttribute;
					};
+                    [key: string]: any,
				};
```

PR #6395, opened 2025-11-28
feat(sso): add OIDC discovery

Paola3stefania:
Add new discovery module for fetching and validating OIDC configuration from .well-known/openid-configuration endpoints.

---
## Summary by cubic
Add OIDC discovery to SSO to fetch, validate, and hydrate provider config from .well-known/openid-configuration. This simplifies provider setup with registration-time validation and enables runtime discovery for legacy configs.

- **New Features**
  - discoverOIDCConfig: computes the discovery URL, fetches the document, validates issuer and required fields, selects token auth, and returns a hydrated config (existing values take precedence).
  - Structured errors via DiscoveryError codes mapped to API errors during registration: discovery_timeout, discovery_not_found, discovery_invalid_url, discovery_invalid_json, issuer_mismatch, discovery_incomplete, unsupported_token_auth_method, discovery_unexpected_error.
  - Token endpoint auth selection with sensible defaults (prefers client_secret_basic; supports client_secret_post if advertised).
  - needsRuntimeDiscovery helper to trigger discovery when token/jwks endpoints are missing.
  - Integrated into registerSSOProvider: performs discovery at registration, hydrates endpoints, and fails fast on misconfigured IdPs (user overrides win).
  - 10s default timeout using betterFetch’s built-in timeout, plus comprehensive tests for success and failure cases.
  - New skipDiscovery option on registerSSOProvider to bypass discovery and require manual endpoints when needed.

<sup>Written for commit e6bcb4408b64296c6d22e4442ddcf607128b92b1. Summary will update automatically on new commits.</sup>


packages/sso/src/oidc/discovery.ts
```diff
@@ -0,0 +1,349 @@
...
+ * Future phases will:
+ * - Resolve relative URLs against issuer base
+ * - Normalize URL formats
+ * - Apply security transformations
+ *
+ * @param doc - The discovery document
+ * @param _issuerBase - The base issuer URL (unused in Phase 1)
+ * @returns The normalized discovery document
+ */
+export function normalizeDiscoveryUrls(
```
Paola3stefania, line 283: @jonathansamines here stub for you to work on

packages/sso/src/oidc/discovery.ts
```diff
@@ -0,0 +1,349 @@
...
+/**
+ * Normalize a single URL endpoint.
+ *
+ * Phase 1: This is a stub that returns the endpoint unchanged.
+ *
+ * @param endpoint - The endpoint URL to normalize
+ * @param _issuerBase - The base issuer URL (unused in Phase 1)
+ * @returns The normalized endpoint URL
+ */
+export function normalizeUrl(endpoint: string, _issuerBase: string): string {
```
Paola3stefania, line 297: @jonathansamines here stub for you to work on as well :)

packages/sso/src/oidc/discovery.ts
```diff
@@ -0,0 +1,390 @@
...
+ * @returns The parsed discovery document
+ * @throws DiscoveryError on network errors, timeouts, or invalid responses
+ */
+export async function fetchDiscoveryDocument(
+	url: string,
+	timeout: number = DEFAULT_DISCOVERY_TIMEOUT,
+): Promise<OIDCDiscoveryDocument> {
+	try {
+		const controller = new AbortController();
+		const timeoutId = setTimeout(() => controller.abort(), timeout);
```
jonathansamines: What do you think of using [`AbortSignal.timeout()`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static) here?

jonathansamines: Actually, I wonder if we can use the built-in [timeout](https://better-fetch.vercel.app/docs/timeout-and-retry#timeout) option.

packages/sso/src/oidc/discovery.ts
```diff
@@ -0,0 +1,390 @@
...
+
+		if (error instanceof Error && error.name === "AbortError") {
+			throw new DiscoveryError(
+				"discovery_timeout",
+				`Discovery request timed out after ${timeout}ms`,
+				{ url, timeout },
+			);
+		}
+
+		if (error instanceof SyntaxError) {
```
jonathansamines: Would better-fetch actually throw in case of parsing errors? I think there is some sort of fallback behavior that may fallback to plain text in case of parsing error.

See:
https://github.com/better-auth/better-fetch/blob/main/packages/better-fetch/src/utils.ts#L63
https://github.com/better-auth/better-fetch/blob/main/packages/better-fetch/src/fetch.ts#L120

But to be honest, I am not a 100% sure, so I'd double check.

packages/sso/src/oidc/discovery.ts
```diff
@@ -0,0 +1,390 @@
...
+			signal: controller.signal,
+		});
+
+		clearTimeout(timeoutId);
+
+		if (response.error) {
+			const status = response.error.status;
+
+			if (status === 404) {
+				throw new DiscoveryError(
```
jonathansamines, line 149: Do you intend for `DiscoveryError` to be a public-facing error (e.g bubble up as-is to the handler). If we intend to let this sort of error bubble up, we might want to extend `APIError` as it'll allow us to communicate response specific info (e.g status code, etc).

packages/sso/src/oidc/discovery.ts
```diff
@@ -0,0 +1,390 @@
...
+				{ url, originalError: error.message },
+			);
+		}
+
+		throw new DiscoveryError(
+			"discovery_unexpected_error",
+			`Unexpected error during discovery: ${error instanceof Error ? error.message : String(error)}`,
+			{
+				url,
+				originalError: error instanceof Error ? error.message : String(error),
```
jonathansamines: I wonder if we want to use error-cause instead to expose the whole causal chain instead of only the top-most error?

packages/sso/src/oidc/discovery.test.ts
```diff
@@ -0,0 +1,850 @@
...
+				DiscoveryError,
+			);
+			expect(() => validateDiscoveryUrl("ftp://example.com/config")).toThrow(
+				"must use HTTP or HTTPS",
+			);
+		});
+
+		it("should throw DiscoveryError with discovery_invalid_url code for invalid URL", () => {
+			try {
+				validateDiscoveryUrl("not-a-url");
```
jonathansamines: I wonder if [`toThrow()`](https://vitest.dev/api/expect.html#tothrowerror) would be a better assertion here. I think it does accept an asymmetric matcher, which would allow to perform partial deep comparisons:

```ts
expect(() => validateDiscoveryUrl("not-a-url"), expect.objectContaining({
  code: "discovery_invalid_urls",
  details: expect.objectContaining({
     url: "not-a-url"
  })
}));
```

packages/sso/src/oidc/discovery.ts
Paola3stefania: Ah this is a great catch ! Yes, betterFetch already has a built-in timeout option. 
Nobody's used it in the repo yet, but this is the right place to start. 

I'll simplify to use that instead of the manual AbortController.

packages/sso/src/oidc/discovery.ts
Paola3stefania: I did double check, thanks for this. I added the SyntaxError catch assuming standard fetch + response.json() behavior, but betterFetch handles parsing internally and intentionally swallows parse errors. That means the catch block was dead code.

Updating the code

Paola3stefania: So ...
response.error is still used for HTTP/network failures (404, 500, DNS, timeout), and we map those to discovery_not_found / discovery_unexpected_error / discovery_timeout.

The typeof response.data === "string" is specifically for the case where the IdP returns a 2xx response but the body isn’t valid JSON – betterFetch falls back to returning the raw text with error = null, so the only way to detect “invalid JSON” is by checking that the data is a string, not an object.

```// node_modules/@better-fetch/fetch/dist/index.js

function jsonParse(text) {
  try {
    return JSON.parse(text);
  } catch (error) {
    return text;  // <-- Falls back to raw text, doesn't throw!
  }
}```

File: packages/sso/src/oidc/discovery.ts (hunk 5 above)
Comment (Paola3stefania) [reply packages/sso/src/oidc/discovery.ts:149]: Good question !The idea behind keeping DiscoveryError separate from APIError is that the discovery module shouldn’t know anything about HTTP or response semantics. 
It only needs to report what went wrong at the OIDC-protocol level (discovery_not_found, issuer_mismatch, discovery_timeout, etc.).

Later, at the SSO route layer (registration or callback), we catch DiscoveryError and decide how to show it — either as an APIError wit, or as a redirect with error / error_description.

This follows the same pattern we use elsewhere (elike BetterAuthError)

File: packages/sso/src/oidc/discovery.ts (hunk 6 above)
Comment (Paola3stefania) [reply packages/sso/src/oidc/discovery.ts]: Reviewing the repo I changed it to error-cause, keeping the discovery errors consistent with the rest of the codebase. Great !

File: packages/sso/src/oidc/discovery.test.ts (hunk 7 above)
Comment (Paola3stefania) [reply packages/sso/src/oidc/discovery.test.ts]: Absolutely, thanks for this.

File: packages/sso/src/oidc/discovery.ts [hunk 8]
```diff
@@ -0,0 +1,401 @@
...
+	// TODO(Phase 2): Implement URL normalization
+	return doc;
+}
+
+/**
+ * Normalize a single URL endpoint.
+ *
+ * Phase 1: This is a stub that returns the endpoint unchanged.
+ *
+ * TODO(Phase 2): Implement normalization of relative URLs based on issuer/discovery endpoint.
```
Bekacru: let's remove todo

packages/sso/src/oidc/discovery.ts
```diff
@@ -0,0 +1,401 @@
...
+/**
+ * Normalize URLs in the discovery document.
+ *
+ * Phase 1: This is a stub that returns the document unchanged.
+ * Future phases will:
+ * - Resolve relative URLs against issuer base
+ * - Normalize URL formats
+ * - Apply security transformations
+ *
+ * TODO(Phase 2): Implement normalization of relative URLs based on issuer/discovery endpoint.
```
Bekacru: lets remove todo

packages/sso/src/oidc/discovery.ts
Paola3stefania: https://github.com/better-auth/better-auth/pull/6503/files  This todo is solved in this PR of @jonathansamines its not from phase 2 from OIDC. 

Should we remove it anyway? its removed there as well

packages/sso/src/oidc/discovery.ts
Paola3stefania: same as above :)

PR #6401, opened 2025-11-29
docs: clarify credential account requirement for phone number plugin 

### LINKED ISSUE #6396: Clarify in the docs for phone number plugin that the User must have a respective record in Account with `providerId` equal to `"credential"`

### To Reproduce

I had hard time migrating from Auth.js (v5) to Better Auth and encountered some hassles which may be worth document.

In my platform, admins pre-create accounts for users to login. Users log in with phone number and password they got sent privately.

The logical thing to do was to use the built-in `phoneNumber` plugin. Unfortuantely it couldn't work out of the box so while examining the source code I learned that:

1. The user must have a record in the `Account` table, otherwise they couldn't authenticate. In my original setup with Auth.js the account table was literally empty - I put the `password` field on `User`.
2. Not only they need to have a record in `Account`, the `providerId` needs to be precisely equal to `"credential"`, as examined from the code.
3. Perhaps a trivial thing, but the password must be set in the account model (and not directly in User model...)

```typescript
// packages\better-auth\src\plugins\phone-number\index.ts line 130-178
const user = await ctx.context.adapter.findOne<UserWithPhoneNumber>({
	model: "user",
	where: [
		{
			field: "phoneNumber",
			value: phoneNumber,
		},
	],
});
if (!user) {
		message:
			PHONE_NUMBER_ERROR_CODES.INVALID_PHONE_NUMBER_OR_PASSWORD,
	});
}
if (opts.requireVerification) {
   .....
}
// 1. It searches for the accounts the user might have
const accounts = await ctx.context.internalAdapter.findAccountByUserId(user.id);
const credentialAccount = accounts.find(
	(a) => a.providerId === "credential", // 2. It filters to get only the ones the explicitly have provider "credential"
);
if (!credentialAccount) {
	ctx.context.logger.error("Credential account not found", {
		phoneNumber,
	});
		message:
			PHONE_NUMBER_ERROR_CODES.INVALID_PHONE_NUMBER_OR_PASSWORD,
	});
}
```

I would expect this to be explained directly in the docs or even in the migration guide. I never populated the accounts table - and even if I did I don't think I would expect to set providerId to be as specific as `"credential"`. Maybe I present a rare case, but I think it's worth clarifying things in the documentation to make the migration smoother.

I would have added clarification to both the migration articles and phone number plugin.

1.3.34

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 10 Home",
    "release": "10.0.19045",
    "cpuCount": 8,
    "cpuModel": "Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz",
    "totalMemory": "7.82 GB",
    "freeMemory": "0.52 GB"
  },
  "node": {
    "version": "v22.11.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.8.2"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "^16.0.5"
    },
    {
      "name": "react",
      "version": "^19.2.0"
    }
  ],
  "databases": [
    {
      "name": "@prisma/client",
      "version": "^6.14.0"
    }
  ],
  "betterAuth": {
    "version": "^1.3.34",
    "config": null
  }
}
```

Client, Backend

```typescript

```

### Additional context

_No response_

GautamBytes:
# Description

Closes #6396

This PR addresses a documentation gap regarding the **Phone Number** plugin's password sign-in flow.

**Context:**
As identified in #6396, the `phoneNumber` plugin strictly requires a user to have a linked record in the `account` table with `providerId: "credential"` to authenticate via password. This logic is hardcoded in the plugin source but was not previously documented.

**Changes:**
1.  **Plugin Docs (`phone-number.mdx`):** Added a `warn` callout in the "Sign In with Phone Number" section to explicitly state this requirement.
2.  **Migration Guide (`next-auth-migration-guide.mdx`):** Added a `warn` callout in the Database Models comparison section. This is critical for users migrating from Auth.js (NextAuth), where passwords were often stored on the `User` model, causing silent failures if the `Account` record is missing.

---
## Summary by cubic
Clarified docs that phone number + password sign-in requires an `Account` record with `providerId: "credential"`.
Added warn callouts in the Phone Number plugin docs and migration guide to prevent silent failures and highlight that passwords must be stored in `Account`, not `User` (for Auth.js migrations).

<sup>Written for commit 1e8cb52f0500492aaf4e5a3ff6012a88271fa0b6. Summary will update on new commits.</sup>


GautamBytes:
@himself65 just pinging to have a review!

ping-maxwell (approved):

Issue #6257, opened 2025-11-24
Origin Header Issue


milhim:
### Is this suited for github?

- [ ] Yes, this is suited for github

### To Reproduce

1. Setup the Project With NodeJS and Fastify
2. Enable Email & Password Auth
3. Add trusted origin
4. Send Signup request from postman 

### Current vs. Expected behavior

Current :
I send first request to signup with postman without the Origin header it pass without any issue and the account was created but the second request response was NULL_OR_MISSING_ORIGIN

Expected:
All request must response with NULL_OR_MISSING_ORIGIN If I did not attach the origin header

### What version of Better Auth are you using?

^1.3.34

### System info

```bash
I did not run this command I added the config below
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

/* eslint-disable sonarjs/todo-tag */
/* eslint-disable @typescript-eslint/no-unused-vars */
/* eslint-disable no-unused-vars */
import { betterAuth } from "better-auth";
import { prismaAdapter } from "better-auth/adapters/prisma";

import prisma from "./prisma.js";
import { env } from "../config/env.js";
import { passkey } from "better-auth/plugins/passkey";
import { captcha, magicLink, twoFactor } from "better-auth/plugins";
import { referralPlugin } from "./plugins/index.js";
import {
  BETTER_AUTH_APP_NAMES,
  CAPTCHA_PROVIDER,
  MAGIC_LINK_EXPIRES_IN,
  PASS_KEY_ORIGIN,
  PASS_KEY_RP_ID,
  PASS_KEY_RP_NAME,
  TRUSTED_ORIGINS,
} from "../constants/index.js";

export const auth = betterAuth({
  secret: env.BETTER_AUTH_SECRET,
  database: prismaAdapter(prisma, {
    provider: "postgresql",
  }),
  appName: BETTER_AUTH_APP_NAMES,
  emailAndPassword: {
    enabled: true,
    minPasswordLength: 8,
    },
  ],
  trustedOrigins: TRUSTED_ORIGINS,
});
export const TRUSTED_ORIGINS = [
  "http://localhost:3000",
  "http://localhost:5000",
];

milhim:
@better-auth-agent

Paola3stefania:
Hi @milhim 

[...]

Issue #5536, opened 2025-10-24
@better-auth/expo with phoneNumber - MISSING_OR_NULL_ORIGIN


ArvidAnderson:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Upgrade @better-auth/expo and better-auth to 1.3.29
2. Try to sign in using phoneNumber.verify
3. Get response of `{"code": "MISSING_OR_NULL_ORIGIN", "message": "Missing or null Origin", "status": 403, "statusText": ""}`

### Current vs. Expected behavior

**Current Behavior:**
When attempting to sign in using phoneNumber.verify with @better-auth/expo and better-auth version 1.3.29, the request fails with a 403 Forbidden error. The response returns `{"code": "MISSING_OR_NULL_ORIGIN", "message": "Missing or null Origin", "status": 403, "statusText": ""}, indicating that the authentication service is rejecting the request due to a missing or null Origin header.`

**Expected Behavior:**
The phoneNumber.verify request should return a successful response, allowing the user to complete the sign-in process.

### What version of Better Auth are you using?

1.3.29

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.0.0: Wed Sep 17 21:41:50 PDT 2025; root:xnu-12377.1.9~141/RELEASE_ARM64_T6030",
    "release": "25.0.0",
    "cpuCount": 11,
    "cpuModel": "Apple M3 Pro",
    "totalMemory": "36.00 GB",
    "freeMemory": "2.45 GB"
  },
  "node": {
    "version": "v22.13.1",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.1"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "1.3.29",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Client, Backend

### Auth config (if applicable)

```typescript

```

R5dan:
@ArvidAnderson Can you show an attempt to make the request? Is it failing for any other requests? Do you know if you are sending [cookies](https://www.better-auth.com/docs/integrations/expo#making-authenticated-requests-to-your-server)

PR #6420, opened 2025-11-30
fix(expo): Due to an incorrect type of the copied request object, the…

GuangZhouzhu:
Due to an incorrect type of the copied request object, the subsequent logic did not use this duplicated object which had the origin field added. This resulted in the MISSING_OR_NULL_ORIGIN error.

Based on my investigation so far, the affected logic is within the handler method returned by the createRouter function in the better-call package [source code at line 248](https://github.com/Bekacru/better-call/blob/main/src/router.ts#L248). Here is the relevant source code:
```ts
return {
    handler: async (request: Request) => {
        const onReq = await config?.onRequest?.(request);
        if (onReq instanceof Response) {
            return onReq;
        }
        // onReq instanceof Request = false, so req = request
        const req = onReq instanceof Request ? onReq : request;
        const res = await processRequest(req);
        const onRes = await config?.onResponse?.(res);
        if (onRes instanceof Response) {
            return onRes;
        }
        return res;
    },
    endpoints,
};
```

Please note this line of code: `const req = onReq instanceof Request ? onReq : request;`

The root cause is that the req object, which was previously created using `const req = request.clone();`, was not of the Request type. Consequently, the type check onReq instanceof Request in better-call evaluated to false. This forced the handler to fall back to using the original request object instead of the onReq object that contained the added origin field.

As a result, the originCheckMiddleware could not detect the origin field in the request, ultimately throwing the "Missing or null Origin" error.

link issue: #5536 #6257

---
## Summary by cubic
Fix origin header handling in Expo by creating a real Request (new Request(request)) instead of using request.clone(). This ensures the better-call router uses the modified request and resolves MISSING_OR_NULL_ORIGIN errors.

<sup>Written for commit fe6553e2c7a40b6e69995ffce5ff9a80291bc054. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/expo/src/index.ts
+++ b/packages/expo/src/index.ts
@@ -39,7 +39,7 @@ export const expo = (options?: ExpoOptions | undefined) => {
 			if (!expoOrigin) {
 				return;
 			}
-			const req = request.clone();
+			const req = new Request(request);
 			req.headers.set("origin", expoOrigin);
 			return {
 				request: req,


himself65:
This looks weird. Is this an expo-only issue?

GuangZhouzhu:
> This looks weird. Is this an expo-only issue?

Yes, this is an Expo-specific issue. It occurs because HTTP requests made from Expo applications, by default, lack the standard origin header and instead use a custom expo-origin header. Therefore, the @better-auth/expo plugin needs to add the origin header to the request on the server side when it receives it.

GuangZhouzhu:
I don't know why `onReq instanceof Request = false`, but i read the mdn [If you intend to modify the request, you may prefer the Request constructor.](https://developer.mozilla.org/en-US/docs/Web/API/Request/clone). so I try to change that, and it works

himself65 (approved):

gustavovalverde:
Closing in favor of #7070 (merged) and #8405 (merged). #7070 fixed plugin `onRequest` chaining so the origin header set by the Expo plugin is actually seen by the router. #8405 then handles both mutable headers (Bun/Node/Deno) and immutable headers (Cloudflare Workers) via try/catch fallback, covering the scenario this PR targets.

PR #6425, opened 2025-11-30
fix(prisma-adapter): extract `id` to root level for delete operations

ping-maxwell:
Prisma requires `id` at the root level for delete operations. When multiple conditions were provided (e.g., `id` + `userId`), they were all placed in an AND array, causing delete operations to fail.

Now, for delete actions, the `id` condition is extracted to the root level while other AND conditions remain in the AND array.

---
## Summary by cubic
Fix delete operations in the Prisma adapter by moving the id filter to the root where clause for delete. Prevents delete failures when id is combined with other filters.

- **Bug Fixes**
  - Extracted id to the root for delete; other filters stay under AND/OR.
  - Updated convertWhereClause to accept action and handle delete-specific logic.
  - Refactored adapter methods to pass the action to convertWhereClause.

<sup>Written for commit ed97e0300373b975f939155acca20a4633acdc49. Summary will update automatically on new commits.</sup>


himself65 (commented):
LGTM if we have test cases

guilhermejansen:
+1 for this fix!

I ran into this exact issue when calling `twoFactor.disable()` - the adapter tries to delete by `userId` which isn't unique in the schema, causing Prisma to throw a validation error.

Currently using `@unique` on `userId` as a workaround, but this fix would be much cleaner. Thanks for working on this!

Related: #6807

PR #6452, opened 2025-12-01
fix: preserve user ID in cookie cache during stateless sessions

### LINKED ISSUE #6447: `generateId: false` causes 401 on API endpoints in stateless mode

### To Reproduce

1. Configure better-auth with stateless session (no database, JWT cookie cache)
2. Set `generateId: false` to use OAuth provider's ID as user ID
3. Configure genericOAuth with Microsoft Entra ID
4. Return `id: idToken.oid` from `mapProfileToUser`
5. Log in successfully
6. Call `auth.api.getAccessToken({ headers, body: { providerId: "..." } })`
7. Receive 401 UNAUTHORIZED

**Use case:** Single OAuth provider (Microsoft Entra ID), no database, stateless sessions. I want to use the Entra ID `oid` as the user ID instead of a generated UUID.

**Expected:** Setting `generateId: false` and returning `id: idToken.oid` from `mapProfileToUser` should preserve that ID in the session. API endpoints like `getAccessToken` should work.

**Actual:**
- `getSession()` returns user data but **without the `id` field**
- `getAccessToken()` and `accountInfo()` fail with 401 UNAUTHORIZED

The `id` is filtered out by `parseUserOutput` in `setCookieCache`, so it's missing from the JWT session cache. When `getAccessToken` checks:

```ts
let resolvedUserId = session?.user?.id || userId;
if (!resolvedUserId) throw ctx.error("UNAUTHORIZED");
```

It throws 401 because `session.user.id` is undefined.

1.4.4

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.1.0: Mon Oct 20 19:33:00 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6020",
    "release": "25.1.0",
    "cpuCount": 10,
    "cpuModel": "Apple M2 Pro",
    "totalMemory": "16.00 GB",
    "freeMemory": "2.39 GB"
  },
  "node": {
    "version": "v22.15.0",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.3"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null
  }
```

Backend

```typescript

```

### Additional context

**Root cause:** In `setCookieCache` (cookies module), the user is filtered through `parseUserOutput`:

```ts
const filteredUser = parseUserOutput(ctx.context.options, session.user);
```

When `generateId: false` is set, the `id` field is not recognized as a configured field and gets filtered out before storing in the JWT cookie.

**Suggested fix:** Always preserve `id` in `setCookieCache`:

```ts
const filteredUser = { id: session.user.id, ...parseUserOutput(ctx.context.options, session.user) };
```

**Workaround:** Store the external ID in a custom field instead:

```ts
// Don't use generateId: false
// Instead, add custom field:
user: {
  additionalFields: {
    entraId: { type: "string", required: false, input: false },
  },
},

// In mapProfileToUser:
return {
  // Let better-auth generate id
  entraId: idToken.oid, // Store Entra ID separately
  ...
};
```

GautamBytes:
Fixes #6447

This PR fixes a bug where the user.id was being stripped from the session cookie (JWT) when running in stateless mode with generateId: false.

**The Issue:** In setCookieCache, the parseUserOutput helper filters the user object based on the configuration. When generateId is set to false (e.g., using an external provider's ID like Entra ID), the internal schema definition does not recognize id as a generated field, causing it to be filtered out. This resulted in 401 Unauthorized errors on API endpoints because the session was restored without a User ID.

**The Fix:** Modified setCookieCache in packages/better-auth/src/cookies/index.ts to explicitly merge session.user.id back into the filtered user object before tokenization.

**Changes:**
- packages/better-auth/src/cookies/index.ts: Explicitly preserve id during cookie creation.
- packages/better-auth/src/cookies/cookies.test.ts: Added a regression test to ensure user.id is always present in the cookie cache.

---
## Summary by cubic
Fixes a bug where user.id was stripped from the session cookie in stateless mode (generateId: false), causing 401 errors when restoring sessions. The cookie cache now always preserves user.id.

- **Bug Fixes**
  - Always preserve user.id in parseUserOutput so it isn't filtered out before cookie creation.
  - Add a regression test to ensure user.id is present in the cookie cache.

<sup>Written for commit 7062dcfcead746e96565f31017d8ead03e80c1bf. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/cookies/index.ts
+++ b/packages/better-auth/src/cookies/index.ts
@@ -135,10 +135,7 @@ export async function setCookieCache(
 		);
 
 		// Apply field filtering to user data
-		const filteredUser = {
-			...parseUserOutput(ctx.context.options, session.user),
-			id: session.user.id,
-		};
+		const filteredUser = parseUserOutput(ctx.context.options, session.user);
 
 		// Compute version
 		const versionConfig = ctx.context.options.session?.cookieCache?.version;
--- a/packages/better-auth/src/db/schema.ts
+++ b/packages/better-auth/src/db/schema.ts
@@ -60,7 +60,10 @@ function getAllFields(options: BetterAuthOptions, table: string) {
 
 export function parseUserOutput(options: BetterAuthOptions, user: User) {
 	const schema = getAllFields(options, "user");
-	return parseOutputData(user, { fields: schema });
+	return {
+		...parseOutputData(user, { fields: schema }),
+		id: user.id,
+	};
 }
 
 export function parseAccountOutput(


Bekacru:
We should implement this at parseOutput level to always add the id...this also applies to sign in endpoint returned fields

PR #6454, opened 2025-12-01
feat(admin): prevent impersonating admins by default [breaking]

### LINKED ISSUE #6448: Prevent admins from impersonating other admins

Currently, if an admin impersonates another admin, then it is difficult to audit what actions were taken during impersonation.

For example:
- Admin A impersonates Admin B
- As Admin B, they then impersonate customer User C
- As User C, they perform an action

If you are logging sessionIds in stdout for audit purposes, then you read `impersonatedBy` in the sessions table to see if which user impersonated User C... but it will say Admin B and not Admin A.

Theoretically, you _could_ look in the session table to see if Admin B had been impersonated in that timeframe, but it will be inconclusive on whether or not that particular action was done by Admin B or someone impersonating Admin B.

I think the easiest solution is to simply prevent admins from impersonating other admins because it would prevent permission escalation out of the box.

There are very few use cases where we want admins to impersonate other admins. However, if some existing users want this capability, then we can make an `allowAdminImpersonation` option in the admin plugin that defaults to false. We should probably make the settings "as secure as possible" by default.

An existing stop-gap solution would be to use hooks:

```typescript
hooks: {
        console.log("ctx.path is", ctx.path);
        if (ctx.path.includes("/admin/impersonate-user")) {
          const body = ctx.body as { userId?: string } | undefined;
          const userId = body?.userId;

          if (userId) {
            const [targetUser] = await db.select().from(users).where(eq(users.id, userId)).limit(1);

            if (targetUser?.role === "admin") {
              throw new Error("Cannot impersonate another admin");
            }
          }
        }
    }),
    ...
}
```

However, I don't think this is a good solution. We can't expect users to write this hook every time and we should try to make everything "secure by default".

Another alternative is to make an option in the `admin` plugin called "canImpersonate`, which could be typed as `(opts: { requestorUserId: string, targerUserId: string }) => Promise<Boolean>`. If we go this route, we could add a default implementation that prevents admin impersonation. The nice thing about this approach is that it solves #3056.

### Additional context

_No response_

jslno:
closes #6448

---
## Summary by cubic
Prevent impersonation of admin accounts in the admin plugin to close a security gap. Impersonation requests targeting admins (by role or adminUserIds, with defaultRole fallback) now return FORBIDDEN unless allowImpersonatingAdmins is true, with support for array and comma-separated adminRoles.

<sup>Written for commit f2abdb4174a735392880d7dc8c7d25946a989703. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/integrations/convex.mdx
+++ b/docs/content/docs/integrations/convex.mdx
@@ -4,7 +4,7 @@ description: Integrate Better Auth with Convex.
 ---
 
 <Callout>
-  This documentation comes from the [Convex documentation](https://convex-better-auth.netlify.app/),
+  This section is adapted from the [Convex + Better Auth guide](https://convex-better-auth.netlify.app/),
   for more information, please refer to their documentation.
 </Callout>
 
@@ -340,8 +340,3 @@ export async function updatePassword({
   );
 }
 ```
-
-<Callout>
-  This documentation comes from the [Convex documentation](https://convex-better-auth.netlify.app/),
-  for more information, please refer to their documentation.
-</Callout>
--- a/docs/content/docs/plugins/oauth-proxy.mdx
+++ b/docs/content/docs/plugins/oauth-proxy.mdx
@@ -68,6 +68,8 @@ This plugin requires skipping the state cookie check. This has security implicat
 
 ## Options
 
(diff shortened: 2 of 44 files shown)


speciankarel2-code:
``
```

himself65 (approved):

PR #6462, opened 2025-12-01
feat: add `ctx.isTrustedDomain` helper

jonathansamines:
**What is changing?**
Refactoring the internal logic to identify URLs that match a trusted domain into a context util (`context.isTrustedDomain(url)`), so that it can be reused in other contexts (e.g SSRF checks).

I plan to use is in a follow-up PR that will further restrict OIDC discovery and will make it less prone to SSRF attacks.

---
## Summary by cubic
Refactored trusted origin matching into a reusable context helper and added support for dynamic trusted origins. This simplifies origin checks and strengthens validation for callback URLs, preparing for stricter OIDC discovery.

- **Refactors**
  - Moved origin matching into a shared utility and exposed context.isTrustedOrigin(url, { allowRelativePaths }).
  - Simplified origin-check middleware to use the context helper and removed duplicate pattern logic.

- **New Features**
  - Supports per-request trustedOrigins via an async function. Added tests and updated docs with examples, plugin usage, and a per-request warning.

<sup>Written for commit 8000cbf9494a8ef22485a44a27530e4a8e30f915. Summary will update automatically on new commits.</sup>


packages/better-auth/src/auth/trusted-origins.ts
Paola3stefania, line 1: How do you feel about unit testing this one? trusted-origins.test.ts

docs/content/docs/reference/security.mdx
Paola3stefania, line 1: how about adding an example on the docs?
Lerts say in the plugin area maybe?


`
**Example: Using `isTrustedOrigin` in a plugin**

You can use `isTrustedOrigin` to validate callback URLs or redirects in your custom endpoints:

```ts title="plugin.ts"
import { createAuthEndpoint } from "better-auth/api";
import type { BetterAuthPlugin } from "better-auth";
import { z } from "zod";

const myPlugin = () => {
    return {
        id: "my-plugin",
        endpoints: {
            customRedirect: createAuthEndpoint("/my-plugin/redirect", {
                method: "POST",
                body: z.object({
                    redirectUrl: z.string(),
                }),
            }, async(ctx) => {
                const { redirectUrl } = ctx.body;
                
                // Validate the redirect URL against trusted origins
                if (!ctx.context.isTrustedOrigin(redirectUrl)) {
                    return ctx.json(
                        { error: "Invalid redirect URL" },
                        { status: 403 }
                    );
                }
                
                // Safe to redirect - URL is trusted
                return ctx.redirect(redirectUrl);
            })
        }
    } satisfies BetterAuthPlugin
}
```

This ensures that your plugin respects the same trusted origins configuration as the rest of Better Auth, preventing open redirect vulnerabilities.
`

and then her ein security part crossreferencing

`
### Using `isTrustedOrigin` in Plugins

If you're building custom plugins or endpoints, you can use the `isTrustedOrigin()` method available on the auth context to validate URLs against your trusted origins configuration. This ensures your custom endpoints respect the same security settings as Better Auth's built-in endpoints.

```typescript
// In a plugin endpoint
async (ctx) => {
  const redirectUrl = ctx.body.redirectUrl;
  
  if (!ctx.context.isTrustedOrigin(redirectUrl)) {
    return ctx.json({ error: "Invalid redirect URL" }, { status: 403 });
  }
  
  return ctx.redirect(redirectUrl);
}
```

See the [Plugins documentation](/docs/concepts/plugins) for more details on accessing context methods.
`

packages/better-auth/src/api/middlewares/origin-check.test.ts
Paola3stefania, line 1: How about adding a regression test for dynamic trustedOrigins (per-request behavior)?

Atest that explicitly verifies that trustedOrigins functions are evaluated per request and receive the actual Request object: **“trustedOrigins: function must be called with the request object, for every request, and its return value must affect origin validation.”**

Something like:
```
	it("should call dynamic trustedOrigins function per-request", async () => {

		// Scenario:
		// - baseURL is http://localhost:3000
		// - We send origin: https://dynamic.example.com (deliberately different from baseURL)
		// - trustedOrigins is a function that returns [request.headers.get("origin")]
		//
		// If trustedOrigins is evaluated per-request:
		//   → Function runs with this request → returns ["https://dynamic.example.com"]
		//   → Origin is trusted → sign-in succeeds → test PASSES
		//
		// If trustedOrigins is only evaluated once (static):
		//   → Function never runs with this request → only baseURL is trusted
		//   → Origin is rejected → 403 error → test FAILS
```

This test will fail if **trustedOrigins** stops being evaluated per-request (or without the req object) and will catch regressions where only static origins are used.

Like a safety net for apps that rely on dynamic origin policies.

packages/better-auth/src/auth/trusted-origins.ts
jonathansamines: yeah, good idea, let me add some tests over the public API we expose.

docs/content/docs/reference/security.mdx
jonathansamines: I think this is a good idea, though I don't think we generally document every property available at the auth context anywhere, so maybe we can start by expanding on the plugins docs.

Let me add some changes here.

packages/better-auth/src/api/middlewares/origin-check.test.ts
jonathansamines: I think we already cover this with the test I've added.

docs/content/docs/reference/security.mdx
jonathansamines: done.

packages/better-auth/src/auth/trusted-origins.ts
jonathansamines: done

docs/content/docs/concepts/plugins.mdx
```diff
@@ -355,6 +357,46 @@ const myPlugin = () => {
...
 ```
 
+### Trusted origins
+
+If you're building custom plugins or endpoints, you can use the `isTrustedOrigin()` method available on the auth context to validate URLs against your trusted origins configuration.
+This ensures your custom endpoints respect the same security settings as Better Auth's built-in endpoints.
+
+```ts title="plugin.ts"
+import { createAuthEndpoint, APIError } from "better-auth/api";
+import z from "zod";
```
himself65: import * as z from "zod"

packages/better-auth/src/auth/trusted-origins.test.ts
```diff
@@ -0,0 +1,289 @@
+import { createAuthEndpoint } from "@better-auth/core/api";
+import { describe, expect, it } from "vitest";
+import z from "zod";
```
himself65: same

docs/content/docs/concepts/plugins.mdx
jonathansamines: There must be a way to lint our code examples 🤣 . Let me fix this.

packages/better-auth/src/auth/trusted-origins.test.ts
jonathansamines: done.

docs/content/docs/concepts/plugins.mdx
jonathansamines: done

himself65 (approved):

PR #6493, opened 2025-12-03
test: fix regression test for generateId: false scenario

GautamBytes:
This PR updates the test suite to correctly verify the fix for issue #6447 (where id was filtered out when generateId: false).
- A targeted unit test for parseUserOutput. This creates a cleaner, faster verification that explicitly proves the id field is preserved when generateId: false is configured, without relying on the database layer.

---
## Summary by cubic
Replaces the DB-dependent cookie cache test with a focused unit test for parseUserOutput to verify user.id is preserved when generateId: false. This makes the regression check faster and more reliable.

<sup>Written for commit 4512e392bdacb7ab58e89d5b229ed0a95bcb6da3. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/cookies/cookies.test.ts
+++ b/packages/better-auth/src/cookies/cookies.test.ts
@@ -544,11 +544,11 @@ describe("Cookie Cache Field Filtering", () => {
 		expect(cache?.user?.adminFlags).toBeUndefined();
 	});
 
-	it("should include id in parseUserOutput when generateId is false", () => {
+	it("should always include id in parseUserOutput", () => {
 		const options = {
-			advanced: {
-				database: {
-					generateId: false,
+			user: {
+				additionalFields: {
+					id: { type: "string", returned: false },
 				},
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/cookies/cookies.test.ts
```diff
@@ -481,36 +482,6 @@ describe("Cookie Cache Field Filtering", () => {
 		expect(cache?.user?.internalNote).toBeUndefined();
 	});
 
-	it("should always preserve the user.id in cookie cache", async () => {
```
himself65, line 484: what is the reason to remove this test case?

GautamBytes: The removed test was a false positive because it didn't actually configure generateId: false. so it failed to reproduce the issue as defined [here ](https://github.com/better-auth/better-auth/issues/6447#issuecomment-3606661971).

PR #6503, opened 2025-12-03
feat(sso): add URL normalization and validation to all discovery URLs

jonathansamines:
**What is changing?**

This PR is build on top of #6456 and further improves our security posture by:

- Normalizing all discovery related URLs to support relative paths
- Validating all discovery related URLs
- Making sure we apply the `trustedOrigins` policy to all downstream requests, improving our security posture by automatically protecting against [`SSRF` ](https://portswigger.net/web-security/ssrf) vulnerabilities.

> **Note**: This PR is blocked by #6462, which will introduces the `ctx.isTrustedOrigin()` helper that we will use to apply the trusted origins policy to discovery urls.

---
## Summary by cubic
Adds URL normalization and trusted-origin validation to OIDC discovery to resolve relative endpoints and block SSRF risks. Discovery now requires an isTrustedOrigin predicate and returns clearer errors.

- **New Features**
  - Normalize authorization, token, JWKS, userinfo, revocation, end_session, and introspection URLs relative to the issuer (base path preserved).
  - Validate URLs and protocols; throw discovery_invalid_url for malformed or non-http/https.
  - Apply trustedOrigins to the main discovery URL and every discovered URL; throw discovery_untrusted_origin when untrusted.
  - Added isTrustedOrigin to discoverOIDCConfig params and mapped new errors to API responses.
  - Updated docs with examples and trustedOrigins guidance.

- **Migration**
  - If you call discoverOIDCConfig directly, pass isTrustedOrigin (e.g., a predicate wired to your trustedOrigins).
  - Add your IdP domain(s) to trustedOrigins (e.g., https://your-org.okta.com).
  - registerSSOProvider now uses ctx.isTrustedOrigin; configure trustedOrigins or discovery will fail for untrusted URLs.

<sup>Written for commit 5e99ceb9bab5709af4e0c6c9620da1ed5a476fd4. Summary will update automatically on new commits.</sup>


packages/sso/src/oidc/discovery.ts
```diff
@@ -102,27 +107,20 @@ export function computeDiscoveryUrl(issuer: string): string {
  * Future phases will add SSRF protection, allowlist checks, etc.
  *
  * @param url - The discovery URL to validate
```
Paola3stefania, line 101: I think we can remove this comment of phase 1 as this pr is adding SSRF protection :)

packages/sso/src/oidc/discovery.ts
```diff
@@ -279,42 +277,147 @@ export function validateDiscoveryDocument(
...
- * - Resolve relative URLs against issuer base
- * - Normalize URL formats
- * - Apply security transformations
- *
- * TODO(Phase 2): Implement normalization of relative URLs based on issuer/discovery endpoint.
- * For now, this is an identity function so we can wire discovery into flows incrementally.
- *
  * @param doc - The discovery document
- * @param _issuerBase - The base issuer URL (unused in Phase 1)
+ * @param issuer - The base issuer URL (unused in Phase 1)
```
Paola3stefania: Here also I think we can remove the "unused in Phase 1" part :)

packages/sso/src/oidc/discovery.ts
```diff
@@ -279,42 +277,147 @@ export function validateDiscoveryDocument(
...
+			issuer,
+			isTrustedOrigin,
+		);
+	}
+
 	return doc;
 }
 
+/**
+ * Normalizes and validates a single URL endpoint
```
Paola3stefania, line 349: Can we fix the JSON Doc to match the params in the function? Missing isTrustedOrigin and what we return

packages/sso/src/oidc/discovery.ts
```diff
@@ -279,42 +277,147 @@ export function validateDiscoveryDocument(
...
  * @returns The normalized discovery document
  */
 export function normalizeDiscoveryUrls(
 	doc: OIDCDiscoveryDocument,
-	_issuerBase: string,
+	issuer: string,
+	isTrustedOrigin: DiscoverOIDCConfigParams["isTrustedOrigin"],
 ): OIDCDiscoveryDocument {
-	// TODO(Phase 2): Implement URL normalization
+	doc.token_endpoint = normalizeAndValidateUrl(
```
Paola3stefania, line 289: This works, but Im concerned about this is as we are in a shared pipeline...
What if we normalized into a shallow copy? Like


```suggestion	
	const normalized: OIDCDiscoveryDocument = { ...doc };
    
    normalized.token_endpoint = normalizeAndValidateUrl(...);
     // ...

     return normalized;
```
In this way callers can safely reuse the original doc if they need it later without worrying about hidden mutations.

docs/content/docs/plugins/sso.mdx
```diff
@@ -209,11 +209,37 @@ Better Auth fills in the following fields by reading the IdP's discovery documen
...
+- `issuer`: `"https://your-org.okta.com"`
+- `token_endpoint`: `"/v1/tokens"`
+- `token_url`: `"https://your-org.okta.com/v1/tokens"`
+
+
+Example of relative endpoint and issuer with base path:
+
+- `issuer`: `"https://your-org.okta.com/v1"`
+- `token_endpoint`: `"/tokens"`
+- `token_url`: `"https://your-org.okta.com/v1/tokens"`
```
Paola3stefania: This **token_url** field seems confusing.  Someone reading this could think: “Wait, is token_url part of OIDC discovery?” “Do I need to provide both?” “Is this a Better Auth custom field?”.

Maybe changing it to something like **normalized token_endpoint: "..."** ?

packages/sso/src/oidc/discovery.test.ts
```diff
@@ -326,18 +350,232 @@ describe("OIDC Discovery", () => {
...
+				normalizeUrl("url", endpoint, "https://idp.example.com/base/"),
+			).toBe("https://idp.example.com/base/oauth2/token");
+		});
+
+		it("should return endpoint as an absolute url preserving the issuer base path (multiple slashes)", () => {
+			const endpoint = "//oauth2/token";
+			expect(
+				normalizeUrl("url", endpoint, "https://idp.example.com/base//"),
+			).toBe("https://idp.example.com/base/oauth2/token");
+		});
```
Paola3stefania: This are great test scenarios , what do you think reducing it (as almost same test) with this: 
```suggestion
		it.each([
			["/oauth2/token", "https://idp.example.com/base", "endpoint with leading slash"],
			["oauth2/token", "https://idp.example.com/base", "endpoint without leading slash"],
			["/oauth2/token", "https://idp.example.com/base/", "issuer with trailing slash"],
			["//oauth2/token", "https://idp.example.com/base//", "multiple slashes"],
		])(
			"should resolve relative endpoint preserving issuer base path (%s, %s) - %s",
			(endpoint, issuer) => {
				expect(normalizeUrl("url", endpoint, issuer)).toBe(
					"https://idp.example.com/base/oauth2/token",
				);
			},
		);
		
```

packages/sso/src/routes/sso.ts
```diff
@@ -619,6 +619,8 @@ export const registerSSOProvider = <O extends SSOOptions>(options: O) => {
 							tokenEndpointAuthentication:
 								body.oidcConfig.tokenEndpointAuthentication,
 						},
+						// TODO: This is a stub until the trusted origins refactor is merged
```
Paola3stefania: maybe add the link to the PR. https://github.com/better-auth/better-auth/pull/6462
so we can track it

packages/sso/src/oidc/discovery.ts
Paola3stefania, line 349:
```suggestion
 * Normalizes and validates a single URL endpoint
 *
 * Only normalizes endpoints that Better Auth actually uses:
 * - token_endpoint, authorization_endpoint, jwks_uri (required)
 * - userinfo_endpoint, revocation_endpoint (optional)
 *
 * Other OIDC endpoints like end_session_endpoint or introspection_endpoint
 * are intentionally skipped as they are not consumed by Better Auth today.
 * Add normalization here when support for those endpoints is added.
 *
```

packages/sso/src/oidc/discovery.ts
jonathansamines: 😱 I was pretty sure I had removed these comments, let me fix this.

packages/sso/src/oidc/discovery.ts
jonathansamines: agreed.

packages/sso/src/oidc/discovery.test.ts
jonathansamines: To be honest I am not a big fan of introducing any kind of logic or iteration to tests 😅 , it just makes them harder to reason about and debug.

packages/sso/src/oidc/discovery.ts
jonathansamines: Also not super fan of overly documenting functions, specially internal functions, in my experience this sort of comment that heavily describes the implementation tend to quickly get outdated as the implementation changes.

packages/sso/src/oidc/discovery.ts
jonathansamines: Done.

packages/sso/src/oidc/discovery.ts
jonathansamines: Done.

packages/sso/src/oidc/discovery.ts
jonathansamines: Done.

packages/sso/src/oidc/discovery.ts
jonathansamines: Done.

docs/content/docs/plugins/sso.mdx
jonathansamines: Fixed.

packages/sso/src/routes/sso.ts
jonathansamines: Done.

packages/sso/src/oidc/discovery.test.ts
Paola3stefania: I would agree on that if its a super complex test, but in a string check I defer :)

packages/sso/src/oidc/discovery.ts
Paola3stefania: Im fine with both options, just we need to track this behavior somewhere, and sometimes the codebase its the best one

packages/sso/src/oidc/discovery.test.ts
jonathansamines: Fair, let me update this.

jonathansamines:
> We only normalize the endpoints that Better Auth actually uses today (**userinfo_endpoint**, **revocation_endpoint**). Others like end_session_endpoint or introspection_endpoint are optional in OIDC and not consumed yet, so skipping them is intentional right?.

Yeah, that's was the intent, but can see how we could forget to normalize other URLs as we start to consume them, so I think it's probably best to normalize everything right now. Let me do that.

PR #6509, opened 2025-12-04
fix(core): unify account linking trust checks across SSO and core OAuth

Paola3stefania:
---
## Summary by cubic
Unifies account linking trust checks across core OAuth and SSO. Adds an explicit auto-linking policy; core defaults to verified_email_trusted for backward compatibility, while SSO only auto-links when the provider is trusted or the domain is verified.

- **New Features**
  - Introduced account.accountLinking.linkingPolicy: never, trusted_providers_only, verified_email_trusted (core default).
  - Added canAutoLinkExistingUser for SSO (OIDC/SAML) using trustedProviders or verified domain match; SSO treats verified_email_trusted as requiring a trust signal.
  - Added skipTrustCheck to handleOAuthUserInfo so SSO can bypass core trust checks after its own validation.

- **Bug Fixes**
  - Consistent redirects with error=account_not_linked when auto-linking is denied.
  - Unified trust logic for OIDC and SAML to prevent unintended linking from untrusted providers.
  - Enforce bans during OAuth sign-in and clear expired bans automatically.

<sup>Written for commit 1e2f73f60a4dc922cdf33b7133ba5a1f1891c9ba. Summary will update automatically on new commits.</sup>

**Backward compatible**

Existing applications that never configured accountLinking will continue to behave exactly as before. Only applications that opt in to stricter modes will get the hardened behavior.

### DIFF
--- a/packages/sso/src/utils.ts
+++ b/packages/sso/src/utils.ts
@@ -1,7 +1,7 @@
 import type { GenericEndpointContext } from "better-auth";
 import type { SSOOptions, SSOProvider } from "./types";
 
-export const validateEmailDomain = (email: string, domain: string) => {
+const validateEmailDomain = (email: string, domain: string) => {
 	const emailDomain = email.split("@")[1]?.toLowerCase();
 	const providerDomain = domain.toLowerCase();
 	if (!emailDomain || !providerDomain) {


Paola3stefania:
The admin plugin and generic OAuth tests exposed a bug: banned users were being treated as **account_not_linked**. Fixed the order of checks so that banned users are always rejected with **error=banned** before any linking logic runs.

Paola3stefania:
Existing applications that never configured accountLinking will continue to behave exactly as before.

Only applications that opt in to stricter modes will get the hardened behavior.

packages/better-auth/src/oauth2/link-account.ts
```diff
@@ -40,31 +42,76 @@ export async function handleOAuthUserInfo(
...
-			) {
+			if (c.context.options.account?.accountLinking?.enabled === false) {
 				if (isDevelopment()) {
 					logger.warn(
-						`User already exist but account isn't linked to ${account.providerId}. To read more about how account linking works in Better Auth see https://www.better-auth.com/docs/concepts/users-accounts#account-linking.`,
+						`Account linking is disabled. User exists but account isn't linked to ${account.providerId}. To enable account linking, see https://www.better-auth.com/docs/concepts/users-accounts#account-linking.`,
 					);
 				}
 				return {
 					error: "account not linked",
```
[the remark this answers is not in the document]
Paola3stefania: This is folowing the repo pattern like "unable to create session", and many others.

joeyorlando:
hi @Paola3stefania @himself65, looking forward to this change!

any chance you can merge this so that I can test? 🙂

docs/content/docs/concepts/users-accounts.mdx
```diff
@@ -380,6 +380,37 @@ export const auth = betterAuth({
 });
 ```
 
+#### Existing User Auto-Linking Policy
```
jonathansamines, line 368: I wonder if we could move this section above the "trusted providers" section? Feels like something I should read before I learn about trusted providers or manual account linking.

docs/content/docs/concepts/users-accounts.mdx
```diff
@@ -380,6 +380,37 @@ export const auth = betterAuth({
...
+
+The `existingUserMode` option controls when OAuth/SSO providers can auto-link to an existing user with the same email address. This provides fine-grained control over account linking security.
+
+```ts title="auth.ts"
+export const auth = betterAuth({
+    account: {
+        accountLinking: {
+            enabled: true,
+            existingUserMode: "trusted_providers_only", // recommended
```
jonathansamines: What do you think of renaming this option to `existingUserPolicy` or `linkingPolicy`?

docs/content/docs/concepts/users-accounts.mdx
```diff
@@ -380,6 +380,37 @@ export const auth = betterAuth({
...
+        }
+    },
+});
+```
+
+Available modes:
+
+- **`"never"`**: Never auto-link. Users must explicitly link accounts (e.g., via a signed-in "link account" flow).
+
+- **`"trusted_providers_only"`** (recommended): Only auto-link if the provider passes a trust check. This means the provider must be listed in `trustedProviders`. For SSO providers, domain verification also counts as a trust signal.
```
jonathansamines: I wonder if we could rephrase this in a way that is not specific to SSO and then we can further elaborate on the SSO plugin docs (which I think we already do).


```suggestion
- **`"trusted_providers_only"`** (recommended): Only auto-link if the provider passes a trust check. This means the provider must be listed in `trustedProviders` or the associated domain has been verified.
```

docs/content/docs/concepts/users-accounts.mdx
```diff
@@ -380,6 +380,37 @@ export const auth = betterAuth({
...
+
+- **`"email_match_any"`** (default): Legacy behavior for backward compatibility. Auto-link when either the provider is trusted OR the provider returns `email_verified: true`.
```
jonathansamines: I found it hard to understand what this policy does only by the name. I wonder if something like `trusted_implicit` would make sense?

docs/content/docs/plugins/sso.mdx
```diff
@@ -704,6 +704,54 @@ mapping: {
...
 
+## Account Linking for SSO
+
+When a user signs in via SSO and an account with the same email already exists, Better Auth applies a **strict trust check** before auto-linking. This prevents account takeover attacks where a malicious SSO provider could claim any email address.
+
+<Callout type="warn">
+**Important**: SSO account linking is stricter than regular OAuth linking. Even if `existingUserMode` is set to `"email_match_any"` in your core config, SSO providers **still require a trust signal** before auto-linking to existing accounts.
+</Callout>
+
+For SSO providers, auto-linking is allowed when **any** of the following conditions are met:
```
jonathansamines, line 715: I think this can be interpreted as a completely different policy, even though all we are doing is applying the `trusted_providers_only` policy. I wonder if we could reinforce that in the docs by saying that SSO only supports `trusted_providers_only` and `never` and that we'll defaul to `trusted_providers_only` even if the user setting is `email_match_any`?

packages/better-auth/src/oauth2/link-account.ts
```diff
@@ -13,12 +13,14 @@ export async function handleOAuthUserInfo(
 		callbackURL,
 		disableSignUp,
 		overrideUserInfo,
+		forceTrusted,
```
jonathansamines:
```suggestion
		implicitTrusted,
```

packages/sso/src/account-linking.test.ts
```diff
@@ -0,0 +1,670 @@
...
+		// Verify SSO account was linked
+		const accounts = await ctx.adapter.findMany({
+			model: "account",
+			where: [{ field: "providerId", value: "sso-domain-verified" }],
+		});
+		expect(accounts.length).toBe(1);
+	});
+});
+
+describe("SSO Account Linking - configuration tests", () => {
```
jonathansamines: These tests feel like they don't really belong to SSO, but rather to the account linking feature as a whole?

jonathansamines (commented):
Changes LGTM, left a few comments with questions/suggestion, mostly on docs and about naming :)

packages/sso/src/account-linking.test.ts
Paola3stefania: Thank you for his.  Those were leftover tests I wrote initially to verify the config was being passed through correctly. Forgot to remove them once I had the actual SSO flow tests in place.

Removed them now

docs/content/docs/concepts/users-accounts.mdx
Paola3stefania: Thinking about this... `linkingPolicy` feels better because It’s shorter and clear.
Also speaks of security/trust decision and its alraedy nested under accountLinking.

packages/better-auth/src/oauth2/link-account.ts
Paola3stefania: Humm I think this changes the meaning.
**forceTrusted = "force this provider to be treated as trusted (bypass checks)"
implicitTrusted = "provider is implicitly trusted (via email_verified)"**

They're different concepts. **forceTrusted** is used when you know the caller is trusted (like from an internal flow). implicitTrusted suggests trust was inferred.

Maybe **skipTrustCheck** its clearer.

docs/content/docs/concepts/users-accounts.mdx
Paola3stefania: I updated this :)

docs/content/docs/concepts/users-accounts.mdx
Paola3stefania: The policy does two things:
Links if provider is in trustedProviders (same as trusted_providers_only)
Also links if provider returns email_verified: true

So the name should convey "we also trust providers that verify email".

Maybe:
verified_email_trusted - "a verified email is treated as trusted"
implicit_email_trust - "implicit trust via email"

Leaning to verified_email_trusted

docs/content/docs/plugins/sso.mdx
Paola3stefania: done !

docs/content/docs/concepts/users-accounts.mdx
Paola3stefania: done

docs/content/docs/concepts/users-accounts.mdx
jonathansamines: `verified_email_trusted ` is a great name!

jonathansamines (commented):
Changes LGTM!

packages/better-auth/src/oauth2/link-account.ts
```diff
@@ -40,31 +42,76 @@ export async function handleOAuthUserInfo(
...
 					error: "account not linked",
 					data: null,
 				};
 			}
+
+			if (!skipTrustCheck) {
+				const trustedProviders =
+					c.context.options.account?.accountLinking?.trustedProviders;
+				const isTrustedProvider = trustedProviders?.includes(
+					account.providerId as "apple",
```
himself65, line 88: `as "google"`?

packages/core/src/types/init-options.ts
```diff
@@ -878,6 +878,39 @@ export type BetterAuthOptions = {
...
+					 *    Legacy behavior for backward compatibility. Auto-link when either:
+					 *    - the provider is trusted, or
+					 *    - the provider returns `email_verified: true`.
+					 *
+					 *    Note: For SSO plugins, this policy still requires a trust signal
+					 *    (trusted provider or domain-verified) for security.
+					 *
+					 * @default "verified_email_trusted"
+					 */
+					linkingPolicy?:
```
himself65, line 910: should be renamed to `autoLinking?: false | "trusted_providers_only" | "verified_email_trusted"`

Paola3stefania:
Closing this PR, as future enhancement. Not priotiry right now

joeyorlando:
@Paola3stefania why? _Honest feedback_ - the SSO plugin is rather buggy right now, I think this PR would've been a step in the right direction.

(it also looked like the PR was practically approved?)

r1me75:
Any updates on this?

PR #6542, opened 2025-12-05
fix: respect returnHeaders and returnStatus options in auth.api endpoints

### LINKED ISSUE #6520: [Bug] `auth.api.getSession()` returns wrapped `{ response: Response }` instead of session data
## Environment
- **Better Auth version**: 1.4.5 (also verified behavior in 1.4.6-beta.2 source)
- **Framework**: NestJS with ORPC
- **Node version**: 20.x

## Description

When calling `auth.api.getSession()` from server-side code, the function returns a wrapped object `{ response: Response }` instead of the expected `{ session, user }` data, even when **not** using `asResponse: true`.

## Expected Behavior

```typescript
const result = await auth.api.getSession({ headers });
// Expected: { session: {...}, user: {...} } or null
```

## Actual Behavior

```typescript
const result = await auth.api.getSession({ headers });
// Actual: { response: Response } where Response is a web Response object
// Need to do: JSON.parse(await result.response.text()) to get actual data
```

## Root Cause Found

Looking at the source code in `to-auth-endpoints.ts` (lines ~94-97), I found that `returnHeaders` and `returnStatus` are **hardcoded to `true`** internally:

```typescript
internalContext.asResponse = false;
internalContext.returnHeaders = true;  // Always true
internalContext.returnStatus = true;   // Always true
```

This causes the response to always be wrapped as `{ headers, response, status }` regardless of what the caller expects.

## Reproduction

Using Better Auth with ORPC in NestJS, extracting session in a middleware/plugin:

```typescript
// ORPC Plugin to inject session into context
const authPlugin = new StandardHandlerPlugin({
  rootInterceptors: [
    async (input, context, meta, next) => {
      const headers = context.request?.headers;
      
      // This returns { response: Response } instead of { session, user }
      const result = await auth.api.getSession({
        headers: new Headers(Object.entries(headers || {})),
      });
      
      // WORKAROUND: Have to manually extract the session
      if (result && 'response' in result && result.response instanceof Response) {
        const body = await result.response.text();
        const parsed = JSON.parse(body);
        context.session = parsed.session;
        context.user = parsed.user;
      }
      
      return next({ input, context, meta });
    },
  ],
});
```

## Current Workaround

Helper to handle the wrapped response:

```typescript
async function extractSession(result: unknown): Promise<{ session: Session; user: User } | null> {
  if (!result) return null;
  
  // Handle the { response: Response } wrapper
  if (typeof result === 'object' && 'response' in result) {
    const wrapped = result as { response: Response };
    if (wrapped.response instanceof Response) {
      const text = await wrapped.response.text();
      if (!text) return null;
      return JSON.parse(text);
    }
  }
  
  // Direct session data (expected behavior)
  if ('session' in (result as object)) {
    return result as { session: Session; user: User };
  }
  
  return null;
}
```

## Questions

1. Is this intentional behavior? If so, what's the recommended way to get session data server-side?
2. Is there a configuration option to disable the response wrapping?
3. Will there be a fix to respect the caller's intent (return raw data when `asResponse` is not set)?

Thank you for the great library!

yoshifumi-kondo:
## Summary

Fixed a bug where `auth.api.getSession()` and other server-side API endpoints always returned wrapped responses `{ headers, response, status }` even when `returnHeaders` and `returnStatus` options were not specified.

## Problem

In `to-auth-endpoints.ts` (lines 109-110), `returnHeaders` and `returnStatus` were hardcoded to `true`, overriding the caller's intent:

```typescript
internalContext.returnHeaders = true;  // Always true
internalContext.returnStatus = true;   // Always true
```

This caused all `auth.api.*` calls to return wrapped responses instead of the actual data.

## Solution

Changed lines 109-110 to respect the caller's options:

```typescript
internalContext.returnHeaders = context?.returnHeaders ?? false;
internalContext.returnStatus = context?.returnStatus ?? false;
```

Now:
- Without options: `auth.api.getSession()` returns `{ session, user } | null`
- With `returnHeaders: true`: Returns `{ headers, response: { session, user } }`
- With `asResponse: true`: Returns `Response` object

## Changes

- Modified `packages/better-auth/src/api/to-auth-endpoints.ts` (lines 109-110)
- Added test case in `packages/better-auth/src/api/routes/session-api.test.ts` to verify unwrapped response behavior

## Testing

- [x] All existing tests pass
- [x] Added new test case for default behavior (no wrapping)
- [x] Existing test case verifies `returnHeaders: true` still works
- [x] Ran `pnpm format` and `pnpm lint` - no issues

## Issue

Fixes #6520

## Impact

This restores the expected behavior for server-side `auth.api` calls, making it consistent with the documented API and type definitions.

---
## Summary by cubic
Fixes auth.api endpoints to respect returnHeaders and returnStatus, so getSession returns unwrapped data by default. Restores the documented behavior and prevents unintended response wrapping.

- **Bug Fixes**
  - Use caller-provided returnHeaders/returnStatus (default false) instead of hardcoded true.
  - getSession now returns { session, user } unless wrapping is requested.
  - Updated response handling: headers may be undefined; status defaults to 200 and uses APIError.statusCode when applicable. Tests cover default, status-only, both, and error cases.

<sup>Written for commit ae4e013ce5a25056017403ce1204b39cd9aa0692. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/session-api.test.ts
+++ b/packages/better-auth/src/api/routes/session-api.test.ts
@@ -369,6 +369,214 @@ describe("session", async () => {
 			},
 		);
 	});
+
+	it("should return session data directly without wrapping when returnHeaders is not set", async () => {
+		const context = await auth.$context;
+		await runWithEndpointContext(
+			{
+				context,
+			} as unknown as GenericEndpointContext,
+			async () => {
+				const signInRes = await auth.api.signInEmail({
+					body: {
+						email: testUser.email,
+						password: testUser.password,
+					},
+					returnHeaders: true,
+				});
+
+				const signInHeaders = new Headers();
+				signInHeaders.set("cookie", signInRes.headers.getSetCookie()[0]!);
+
+				const sessionRes = await auth.api.getSession({
+					headers: signInHeaders,
+				});
+
+				// Should return session data directly, not wrapped in { response: ... }
+				expect(sessionRes).toHaveProperty("user");
+				expect(sessionRes).toHaveProperty("session");
+				// @ts-expect-error: response property should not exist
+				expect(sessionRes.response).toBeUndefined();
+				// @ts-expect-error: headers property should not exist
+				expect(sessionRes.headers).toBeUndefined();
+				// @ts-expect-error: status property should not exist
+				expect(sessionRes.status).toBeUndefined();
+
+				expect(sessionRes?.user).toBeDefined();
+				expect(sessionRes?.session).toBeDefined();
+				expect(sessionRes?.user.email).toBe(testUser.email);
+			},
+		);
+	});
+
+	it("should return status without headers when only returnStatus is true", async () => {
+		const context = await auth.$context;
+		await runWithEndpointContext(
+			{
+				context,
+			} as unknown as GenericEndpointContext,
+			async () => {
+				const signInRes = await auth.api.signInEmail({
+					body: {
+						email: testUser.email,
+						password: testUser.password,
+					},
+					returnHeaders: true,
+				});
+
+				const signInHeaders = new Headers();
+				signInHeaders.set("cookie", signInRes.headers.getSetCookie()[0]!);
+
+				const sessionRes = await auth.api.getSession({
+					headers: signInHeaders,
+					returnStatus: true,
+				});
+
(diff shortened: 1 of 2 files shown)


yoshifumi-kondo:
I gave up

Issue #6481, opened 2025-12-03
SAML account linking requires trustedProviders but OIDC does not (inconsistent behavior)


joeyorlando:
## Description

When a user logs in via SSO and an account with their email already exists in the database, Better Auth handles OIDC and SAML account linking differently. OIDC automatically links accounts based on email match, while SAML requires the provider to be in `trustedProviders` or have `domainVerified: true`.

## Expected Behavior

Both OIDC and SAML should have the **same** account linking behavior. Either:
1. Both should require trusted providers for automatic linking, OR
2. Both should allow automatic linking based on email match (current OIDC behavior)

## Current Behavior

**OIDC Callback** (~line 1719-1750 in `@better-auth/sso/src/routes/sso.ts`):
```typescript
if (existingUser) {
  user = existingUser;
}
// ... later ...
if (!account) {
  // Just creates the account link - no trusted provider check!
  await ctx.context.internalAdapter.createAccount({...});
}
```

**SAML Callback** (~line 2076-2105 in `@better-auth/sso/src/routes/sso.ts`):
```typescript
if (existingUser) {
  const account = await findAccount(...);
  if (!account) {
    // Requires trusted provider OR domain verification!
    const isTrustedProvider = trustedProviders?.includes(provider.providerId) ||
      ("domainVerified" in provider && provider.domainVerified && validateEmailDomain(...));
    
    if (!isTrustedProvider) {
      throw ctx.redirect(`${callbackUrl}?error=account_not_found`);
    }
    await ctx.context.internalAdapter.createAccount({...});
  }
}
```

## Steps to Reproduce

1. Create a user account with email `user@example.com`
2. Configure a custom SAML provider (not in `trustedProviders`)
3. Login via SAML with a user whose email is `user@example.com`
4. Observe: Redirected with `?error=account_not_found`

Compare with OIDC:
1. Create a user account with email `user@example.com`
2. Configure a custom OIDC provider (not in `trustedProviders`)
3. Login via OIDC with a user whose email is `user@example.com`
4. Observe: Account is linked successfully ✅

## Impact

- Custom OIDC providers work fine for account linking
- Custom SAML providers fail with `account_not_found` unless explicitly added to `trustedProviders` or domain verification is enabled and completed
- This creates a confusing UX where OIDC "just works" but SAML requires extra configuration

## Workarounds

1. Add all custom provider IDs to `trustedProviders` (requires knowing them ahead of time)
[...]

PR #6551, opened 2025-12-05
fix(saml): enforce trusted provider check

Paola3stefania:
---
## Summary by cubic
Fixes SAML callback trust checks to prevent auto-linking accounts unless the provider is trusted or the domain is verified. This blocks unintended links and redirects with error=account_not_linked when trust is missing.

- **Bug Fixes**
  - For existing users, only create an account link if the provider is in trustedProviders or domainVerified and the email domain matches; otherwise redirect with error=account_not_linked.
  - Removed unconditional account linking on callback; now respects accountLinking trust rules.
  - Added tests covering deny (untrusted provider, unverified domain) and allow (trusted provider) scenarios.

<sup>Written for commit ce4e81154a2b71497d879cfc1833f12fc9eddcec. Summary will update automatically on new commits.</sup>

This fixes https://github.com/better-auth/better-auth/issues/6481

### DIFF
--- a/packages/sso/src/saml.test.ts
+++ b/packages/sso/src/saml.test.ts
@@ -1184,7 +1184,6 @@ describe("SAML SSO", async () => {
 	});
 
 	it("should deny account linking when provider is not trusted and domain is not verified", async () => {
-		// Create a separate auth instance for this test
 		const {
 			auth: authUntrusted,
 			signInWithTestUser,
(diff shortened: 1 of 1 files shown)


jonathansamines (commented):
Changes LGTM, is there any docs that we might need to update?

Paola3stefania:
> Changes LGTM, is there any docs that we might need to update?

Yesss, account linking in docs. Just waiting to see if we want to make OIDC and SAML trust system consistent and go with that update. ty

Bekacru (approved):

PR #6557, opened 2025-12-05
feat(sso): add SAML replay attack protection via InResponseTo validation

Paola3stefania:
---
## Summary by cubic
Adds SAML replay attack protection by validating InResponseTo against stored AuthnRequest IDs. Prevents replay and cross-provider injection, and can block IdP-initiated (unsolicited) responses while keeping existing setups backward compatible.

- **New Features**
  - InResponseTo validation for SP-initiated SAML; enable via saml.enableInResponseToValidation or by providing saml.authnRequestStore.
  - AuthnRequestStore interface with default persistence via the verification table when no store is provided; in-memory test store and DEFAULT_AUTHN_REQUEST_TTL_MS exported.
  - Validation in SAML callback/ACS with allowIdpInitiated and requestTTL options, plus clear redirect errors on failure.
  - Docs updated with single vs multi-instance setup examples.

- **Migration**
  - Single instance: set enableInResponseToValidation: true (no custom store needed). Optionally set allowIdpInitiated: false.
  - Multi instance: provide a shared authnRequestStore (e.g., Redis) so AuthnRequest IDs are available across nodes.

<sup>Written for commit c17b01b4d1de9c04077995a3809380c54a10efd4. Summary will update automatically on new commits.</sup>

## Summary

Adds InResponseTo validation for SAML SP-initiated flows to prevent replay attacks, unsolicited responses, and cross-provider injection.

## Breaking Changes

**None** — This feature is opt-in to ensure backward compatibility.

## How to Enable

Users can opt-in to this security enhancement in two ways:

```
**Single-instance deployments:**
sso({
    saml: {
        enableInResponseToValidation: true,
    },
})**Multi-instance deployments (automatically enabled):**
sso({
    saml: {
        authnRequestStore: redisStore, // providing a store enables validation
    },
})
```

## What's Included

- AuthnRequest ID tracking for SP-initiated SAML logins
- InResponseTo validation against stored requests
- One-time use enforcement (deleted after validation)
- Configurable IdP-initiated flow restriction
- Unit and integration tests
- Documentation updates


packages/sso/src/authn-request-store.ts
himself65, line 1: What is the reason we don't use secondary storage? I feel like it's duplicate

Bekacru: yeah we should rely on secondary storage

Bekacru: and the plugin needs to work without any secondary sotrage or in memory db (most people deploy apps on serverless memory isn't available) so we need to fallback to using db (verification table)

Paola3stefania: The in-memory default was a quick solution for the initial implementation, but it should use secondary storage with verification table fallback like everything else does. I'll fix it. Thank you

PR #6589, opened 2025-12-07
feat(scim): add Microsoft Entra ID SCIM Compatibility

cemcevik:
The current SCIM implementation does not fully cover the SCIM protocol requirements for Microsoft Entra ID.

This PR addresses those gaps by implementing the necessary changes to achieve full compatibility.  The implementation has been validated using Microsoft's official SCIM Validator tool (https://scimvalidator.microsoft.com/) and passes all required tests.

#### 1. Added `active` Field to User Schema
- Added an `active` boolean field to the user table to track user status
- This enables soft-delete functionality, which is essential for Microsoft Entra ID user deprovisioning workflows

#### 2. Case-Insensitive Operation Handling
- Microsoft sends SCIM operations (`Replace`, `Add`, `Remove`) with uppercase first letters
- Implemented normalization to convert operations to lowercase before processing

#### 3. Empty Media Type Support for DELETE Requests
- Microsoft does not send a media type header for DELETE user requests
- Added empty string (`""`) as an allowed media type for the delete endpoint

#### 4. Path Normalization for Microsoft's Format
- Microsoft uses dot notation without leading slashes (e.g., `name.formatted` instead of `/name/formatted`)
- Implemented a path normalizer that converts Microsoft's format to the expected internal format

#### 5. Added Support for `add` Operation
- The previous implementation only handled the `replace` operation
- Added full support for the `add` operation, including nested object value handling

---
## Summary by cubic
Adds full Microsoft Entra ID SCIM compatibility across user CRUD and patch operations. Normalizes Microsoft-specific payloads; this implementation passes Microsoft’s SCIM Validator.

- **New Features**
  - Supports add operations, nested values, operations without a path, and dot-notation paths.
  - Normalizes operation casing and allows empty media type headers on DELETE.
  - Improves name sub-attribute updates and skips no-op add operations.

<sup>Written for commit ac3e689f1874058c5319c5c27f4316228685c863. Summary will update on new commits.</sup>


packages/scim/src/patch-operations.ts
```diff
@@ -56,24 +61,82 @@ const userPatchMappings: Record<string, Mapping> = {
...
+		if (operation.op !== "add" && operation.op !== "replace") {
 			continue;
 		}
 
-		const mapping = userPatchMappings[operation.path];
-		if (mapping) {
-			const resource = resources[mapping.resource];
-			resource[mapping.target] = mapping.map(user, operation);
-		}
+		applyPatchValue(
```
himself65, line 138: This seems like a big change. I think we should add more tests for this

packages/scim/src/patch-operations.ts
```diff
@@ -56,24 +61,82 @@ const userPatchMappings: Record<string, Mapping> = {
...
+
+	resources[mapping.resource][mapping.target] = newValue;
+};
+
+const applyPatchValue = (
+	user: User,
+	resources: Resources,
+	value: unknown,
+	op: "add" | "replace",
+	path = "",
```
himself65: remove `= ""`?

plyhdrn: I've replaced it with `path?: string | undefined`

packages/scim/src/patch-operations.ts
plyhdrn: thanks for the feedback, I’ve added more tests

himself65 (approved):

plyhdrn:
I missed adding the `active` property to the docs in the `Create new user`, `Update an existing user` and `SCIM attribute mapping` sections. I'll update it shortly.

packages/scim/src/index.ts
```diff
@@ -59,6 +59,15 @@ export const scim = (options?: SCIMOptions) => {
 					},
 				},
 			},
+			user: {
```
jonathansamines: I really wonder whether the SCIM plugin should be extending the users schema to fit any particular provider's needs. I'd argue that the fact that better-auth does not support it is really the problem here and this is not necessarily a problem that should be solved by the SCIM plugin.

So, I wonder if there is a design that would allow us to provision arbitrary new fields, and let custom plugins handle the missing schema parts (by extending or adding new schemas).

I think the reasons why I don't necessarily feel comfortable adding this here, is:

- This fields does not really have any meaning for anything but the SCIM plugin. I'd argue whether this is useful if neither better-auth itself or any of the ecosystem plugins will respect it.
- For the same reasons, other third party plugins could be provisioning their own active field already, and that could lead to conflicts here.

jonathansamines: Other than this aspect, all other changes seem like a good fit for the SCIM plugin though :) 

cc @Bekacru

plyhdrn: Agreed, the active attribute is optional in the SCIM spec (https://datatracker.ietf.org/doc/html/rfc7643#section-4.1.1) and isn't strictly required by Microsoft Entra ID, even though it is commonly used by other identity providers.

Since users often want to sync other fields like manager or department with SCIM anyway, in my opinion, allowing users or other plugins to dynamically define additional fields would be a much more flexible approach.

plyhdrn: @jonathansamines  What if we allow users to define their own schema for the users table along with field mappings? Something like this:
```[js]
plugins: [
    scim({
        schema: {
            user: {
                fields: {
                    active: {
                        type: "boolean",
                        defaultValue: false,
                        mapping: "active",
                    },
                    address: {
                        type: "string",
                        mapping: 'addresses[type eq "work"].primary',
                    },
                    department: {
                        type: "string",
                        mapping:
                            "urn:ietf:params:scim:schemas:extension:enterprise:2.0:User:department",
                    },
                },
            },
        },
    }),
],
```

This way users define their own columns, so there's no risk of conflicts with other plugins. And since they also control the mapping, it doesn't matter which SCIM provider they're using, they can change it however they want.

From what I can tell, better-auth plugin structure should support this since plugins can already extend schemas dynamically.

plyhdrn:
Since it’s unclear how we’ll be extending SCIM, I’m thinking I should just remove the active field from this PR. That way, we’d have a version that more compatible with Microsoft Entra ID while we continue discussing possible solutions.

Thoughts?

jonathansamines:
> Since it’s unclear how we’ll be extending SCIM, I’m thinking I should just remove the active field from this PR. That way, we’d have a version that more compatible with Microsoft Entra ID while we continue discussing possible solutions.
> 
> Thoughts?

That sounds like a great idea, sorry I haven't been able to come back to this PR, but that should do it as a way to unblock the rest of the compat issues with Microsoft Entra!

plyhdrn:
> > Since it’s unclear how we’ll be extending SCIM, I’m thinking I should just remove the active field from this PR. That way, we’d have a version that more compatible with Microsoft Entra ID while we continue discussing possible solutions.
> > Thoughts?
> 
> That sounds like a great idea, sorry I haven't been able to come back to this PR, but that should do it as a way to unblock the rest of the compat issues with Microsoft Entra!

Perfect, I’ve removed the active field from the PR. I’ve also opened a feature request on Discord for supporting dynamically defined fields and mapping.

jonathansamines (approved):
Changes LGTM.

plyhdrn:
If there are no other issues, can we merge this PR? @Bekacru

himself65:
We are going to merge this

PR #6611, opened 2025-12-08
fix(username): await username validator

jslno:
## Summary by cubic
Await the username validator in the username plugin so async validators run correctly in availability checks and during sign-up/sign-in. Invalid usernames now fail with 422 USERNAME_IS_INVALID.

<sup>Written for commit cb36ade4a3e4f853bc45017146685114a24eeb08. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/username/username.test.ts
+++ b/packages/better-auth/src/plugins/username/username.test.ts
@@ -501,24 +501,24 @@ describe("isUsernameAvailable with custom validator", async (it) => {
 	});
 
 	it("should reject username that doesn't match custom validator during sign-up/sign-in", async () => {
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/username/index.ts
```diff
@@ -275,7 +275,7 @@ export const username = (options?: UsernameOptions | undefined) => {
 					const validator =
 						options?.usernameValidator || defaultUsernameValidator;
 
-					if (!validator(username)) {
+					if (!(await validator(username))) {
```
Bekacru: lets add a test and also move it to a variable instead of awaiting inside if block

PR #6618, opened 2025-12-08
feat(sso): use domain verified flag to trust providers automatically

Paola3stefania:
## Summary by cubic
Unifies SSO trust checks for OIDC and SAML so domain-verified providers are trusted for automatic account linking when the email domain matches. Updates docs to state that verified domains enable automatic linking.

- **New Features**
  - Treat domain-verified providers as trusted for auto account linking (OIDC + SAML) when email domain matches.
  - Compute isTrustedProvider in the SSO callback using provider.domainVerified and validateEmailDomain.
  - Pass isTrustedProvider into handleOAuthUserInfo to allow linking without email_verified.
  - Add tests for OAuth2 linking (trusted vs untrusted) and OIDC domain-verified linking.

<sup>Written for commit b0f9c52d7c9ce38ebb13a0923bba18c8fbd52724. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,7 +163,7 @@ jobs:
         env:
           TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
           TURBO_TEAM: ${{ vars.TURBO_TEAM || github.repository_owner }}
-        run: pnpm test
+        run: pnpm coverage
 
       - name: Stop Docker Containers
         run: docker compose down
--- a/biome.json
+++ b/biome.json
@@ -81,6 +81,7 @@
     "includes": [
       "**",
       "!**/dist",
+      "!**/coverage",
       "!**/build",
       "!**/.next",
       "!**/.svelte-kit",
--- a/package.json
+++ b/package.json
@@ -18,14 +18,22 @@
     "release:canary": "turbo build --filter=./packages/* && bumpp && pnpm -r publish --access public --tag canary --no-git-checks",
     "bump": "bumpp",
     "test": "turbo test --continue --filter=./packages/* --filter=./test",
+    "coverage": "turbo coverage --filter=./packages/* --filter=./test -- --coverage.reporter=json --coverage.provider=istanbul",
+    "postcoverage": "pnpm coverage:collect && pnpm coverage:merge && pnpm coverage:report",
+    "coverage:merge": "nyc merge coverage/raw coverage/merged/merged-coverage.json",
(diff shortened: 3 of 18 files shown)


docs/content/docs/plugins/sso.mdx
```diff
@@ -707,7 +707,13 @@ mapping: {
...
 
 Domain verification allows your application to automatically trust a new SSO provider
-by automatically validating ownership via the associated domain:
+by automatically validating ownership via the associated domain.
+
+When a provider's domain is verified, it is also trusted for **automatic account linking**. This means that if a user signs in with an SSO provider (OIDC or SAML) and an existing account with the same email exists, the accounts will be linked automatically — as long as the user's email domain matches the provider's verified domain.
+
+<Callout type="info">
+This is a stronger trust mechanism than `trustEmailVerified` (deprecated) because it requires proof of domain ownership rather than relying on the provider's `email_verified` claim.
+</Callout>
```
Bekacru: let's remove this callout here

Bekacru: since it's deprecated no need to mention it in the docs

packages/better-auth/src/oauth2/link-account.ts
```diff
@@ -13,12 +13,14 @@ export async function handleOAuthUserInfo(
...
 		disableSignUp,
 		overrideUserInfo,
+		isTrustedByDomain,
 	}: {
 		userInfo: Omit<User, "createdAt" | "updatedAt">;
 		account: Omit<Account, "id" | "userId" | "createdAt" | "updatedAt">;
 		callbackURL?: string | undefined;
 		disableSignUp?: boolean | undefined;
 		overrideUserInfo?: boolean | undefined;
+		isTrustedByDomain?: boolean | undefined;
```
Bekacru: instead of isTrustedByDomain we should rename this to `isTrustedProvider` and on line 53 

```ts
const isTrustedProvider = opts.isTrustedProvider || trustedProviders?.includes(
				account.providerId as "apple",
)
```

docs/content/docs/plugins/sso.mdx
Paola3stefania: Ah great. makes sense

packages/better-auth/src/oauth2/link-account.ts
Paola3stefania: this way would look cleaner ! ty

PR #6622, opened 2025-12-09
fix: use async storage pollyfill as defualt fallback

Bekacru:
## Summary by cubic
Default to the AsyncLocalStorage polyfill when Node’s async_hooks is unavailable. This prevents crashes and keeps context features working in Cloudflare Workers and other non-Node runtimes.

- **Bug Fixes**
  - Use polyfill fallback in getAsyncLocalStorage; no longer throw on import failure.
  - Remove Convex-specific env checks; always fallback when async_hooks can’t load.
  - Keep browser returning null and show helpful warnings.

- **Dependencies**
  - Bump all packages to v1.4.6-beta.4.

<sup>Written for commit e55a3ad70a971082e2f59c52fb3f9761ab525aa3. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/core/src/async_hooks/index.ts
+++ b/packages/core/src/async_hooks/index.ts
@@ -1,5 +1,4 @@
 import type { AsyncLocalStorage } from "node:async_hooks";
-import { env } from "../env";
 
 export type { AsyncLocalStorage };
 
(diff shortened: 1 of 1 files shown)


packages/core/src/async_hooks/index.ts
```diff
@@ -55,7 +51,7 @@ const AsyncLocalStoragePromise: Promise<typeof AsyncLocalStorage | null> =
 			console.warn(
 				"[better-auth] If you are using Cloudflare Workers, please see: https://developers.cloudflare.com/workers/configuration/compatibility-flags/#nodejs-compatibility-flag",
 			);
-			throw err;
```
erquhart, line 58: @Bekacru note: Convex runtime won't make it here, it will throw an uncatchable system error when the import is attempted.

Bekacru: is it because `node:` prefix or if a module doesn't exist and it's imported it just fails?

himself65: Do you have the runtime key for package.json? like `workerd` or `browser`...

himself65:
Closing, see https://github.com/better-auth/better-auth/pull/6630

PR #6626, opened 2025-12-09
feat(admin): add `/admin/has-role` endpoint

### LINKED ISSUE #6625: Implement `hasRole` endpoint in admin plugin

There is no clean or native way to check whether a user (other than the one currently logged in) has one or more specific roles. I am aware that it makes sense and is safer to check for individual permissions, but in some scenarios it can be helpful to perform a simple role comparison.

An endpoint in the admin plugin that checks whether a user has a role. In my opinion, this should be relatively easy to implement. I am happy to implement the feature myself.

My workaround so far has always been to check whether the target user has a permission that is only held by a specific role. However, this means that certain permissions can no longer be distributed flexibly and breaks the workflow if other groups also need this permission.

### Additional context

_No response_

chris23lngr:
Closes #6625 

---
## Summary by cubic
Adds an admin hasRole endpoint to check if a user has one or more roles. Exposes server and client APIs so admin flows can gate actions by role.

- **New Features**
  - POST /admin/has-role with admin middleware.
  - Accepts userId and role(s) (string or array); validates input.
  - Requires an authenticated session; handles not-found users and missing roles.
  - Checks comma-separated user.role; returns success only if all requested roles are present.
  - Exposes auth.api.userHasRole and authClient.admin.hasRole with OpenAPI metadata.

<sup>Written for commit ef71e571af16b7c7ac32f9b24c88e55437d6b534. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/admin/routes.ts
+++ b/packages/better-auth/src/plugins/admin/routes.ts
@@ -1715,7 +1715,10 @@ export const userHasRole = <O extends AdminOptions>(opts: O) => {
 											description: "The id of the user",
 										},
 										role: {
-											type: "string",
+											oneOf: [
+												{ type: "string" },
+												{ type: "array", items: { type: "string" } },
+											],
 											description: "The role to check for",
 										},
 									},
--- a/packages/better-auth/src/plugins/admin/routes.ts
+++ b/packages/better-auth/src/plugins/admin/routes.ts
@@ -1657,3 +1657,140 @@ export const userHasPermission = <O extends AdminOptions>(opts: O) => {
 		},
 	);
 };
+
+const userHasRoleBodySchema = z.object({
+	userId: z.coerce.string().meta({
+		description: "The user id",
+	}),
+	role: z
+		.union([
+			z.string().meta({
+				description: "The role to check for",
+			}),
+			z.array(
+				z.string().meta({
+					description: "The roles to check for",
+				}),
+			),
+		])
+		.meta({
+			description:
+				"The role to check for, this can be a string or an array of strings. Eg: `admin` or `[admin, user]`",
+		}),
+});
+
+/**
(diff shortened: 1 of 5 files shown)


chris23lngr:
As mentioned in #6625 better-auth is missing a way to check if a user has a specific role. From the original issue:

> My workaround so far has always been to check whether the target user has a permission that is only held by a specific role. However, this means that certain permissions can no longer be distributed flexibly and breaks the workflow if other groups also need this permission.

This PR should be ready for review. Happy to hear your feedback.

packages/better-auth/src/plugins/admin/routes.ts
```diff
@@ -1657,3 +1657,143 @@ export const userHasPermission = <O extends AdminOptions>(opts: O) => {
...
+				});
+			}
+
+			const session = await getSessionFromCtx(ctx);
+
+			if (!session && (ctx.request || ctx.headers)) {
+				throw new APIError("UNAUTHORIZED");
+			}
+			const user = (await ctx.context.internalAdapter.findUserById(
+				ctx.body.userId as string,
```
himself65, line 1773: There is a security issue. The API leaks the information in the public endpoint; anyone who logs in can access the endpoint

himself65 (changes requested):

packages/better-auth/src/plugins/admin/routes.ts
chris23lngr: Would you consider adding a new default permission like `view` oder should i just use one of the existing ones like `list`?

I think adding a own permission could make sense for cases, where regular users should be able to see other users roles without giving them the ability to list all users.

chris23lngr: @himself65 Any updates on this?

PR #6633, opened 2025-12-09
feat: add code property for api error instance

himself65:
Fixes: https://github.com/better-auth/better-auth/issues/6938

---
## Summary by cubic
Standardized API error responses to always include a machine-readable code, and surfaced plugin error codes to clients for consistent handling.

- **New Features**
  - Introduced APIError in @better-auth/core/error with from and fromStatus helpers.
  - Added isAPIError utility and updated docs to use it for safe error detection.
  - defineErrorCodes now returns { code, message }; BASE_ERROR_CODES expanded (origin/callback/redirect, email verification, sessions, etc.).
  - Updated all routes and plugins to return consistent { code, message } errors.
  - Origin/callback/redirect/errorCallbackURL/newUserCallbackURL validation now throws specific error codes.
  - Exposed error codes to clients via $ERROR_CODES across core plugins (admin, anonymous, api-key, email-otp, generic-oauth, multi-session, organization, phone-number, two-factor, username), plus Stripe and Passkey.

- **Migration**
  - Client code should read error.code alongside error.message in responses.
  - Use isAPIError instead of instanceof checks when handling errors.
  - Custom plugins: update $ERROR_CODES to the new shape { code, message } and use defineErrorCodes accordingly.
  - When throwing errors in custom code, use APIError.from(status, errorCode) or APIError.fromStatus(status).

<sup>Written for commit 7c39df9b61665169fadc373b5694892a2bbcdfd2. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/stripe/src/stripe.test.ts
+++ b/packages/stripe/src/stripe.test.ts
@@ -1,11 +1,7 @@
-import type { GenericEndpointContext } from "@better-auth/core";
 import { runWithEndpointContext } from "@better-auth/core/context";
 import type { Auth, User } from "better-auth";
-import { betterAuth } from "better-auth";
 import { memoryAdapter } from "better-auth/adapters/memory";
-import { createAuthClient } from "better-auth/client";
-import { setCookieToHeader } from "better-auth/cookies";
-import { bearer } from "better-auth/plugins";
+import { getTestInstance } from "better-auth/test";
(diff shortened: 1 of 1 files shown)


himself65 (commented):
I think we might need to unify the code prefix.

For example: `INVALID_`, `USER_`, `NOT_`...

Issue #4972, opened 2025-09-29
[SSO] Users signing in with different methods don't get added to the organization


jakst:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Set up Better Auth with the SSO plugin
2. Create an organization and set up an SSO configuration with a domain for it
3. Sign up a user with the configured SSO domain, but using a different sign-in method

The user will not be invited to the organization tied to the domain of the users email-address.

Note 1: I had to set `account.accountLinking.enabled: true` and `plugins[sso].trustEmailVerified: true` in order to get signup to work with SSO. This might play a role in the beahviour here.

Note 2: I can't upgrade to latest better-auth, because then SAML config is required when creating an SSO configuration, and we're not using SAML, just regular OIDC.

### Current vs. Expected behavior

I expect either of these behaviours when using the SSO plugin:

1. Signing in/up with other auth methods is disabled when SSO is configured for a domain
2. If the domain matches, users who sign up with other methods than SSO still get added to the organization.

### What version of Better Auth are you using?

1.3.9

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.6.0: Mon Jul 14 11:28:30 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6030",
    "release": "24.6.0",
    "cpuCount": 12,
    "cpuModel": "Apple M3 Pro",
    "totalMemory": "36.00 GB",
    "freeMemory": "0.16 GB"
  },
  "node": {
    "version": "v24.8.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "10.17.1"
  },
  "frameworks": [
    {
      "name": "solid",
      "version": "1.9.9"
    }
  ],
  "databases": [
    {
      "name": "drizzle",
      "version": "0.44.5"
    }
  ],
  "betterAuth": {
    "version": "1.3.9",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"

export const auth = betterAuth({
  baseURL,
  secret,

  session: {
    cookieCache: {
      enabled: true,
      maxAge: 15 * 60,
    },
  },

  user: {
    deleteUser: {
      enabled: true,
    },
  },

  account: {
    accountLinking: {
      enabled: true,
    },
  },

  advanced: {
[...]

PR #6652, opened 2025-12-10
Unify SSO/OAuth account linking and add domain-based org assignment to all sign-in flows

Paola3stefania:
This PR refactors SSO flows (SAML and OIDC) to use the shared `handleOAuthUserInfo()` pipeline, ensuring consistent identity linking, signup behavior, and organization assignment across all protocols.

## Bug Fix

**SAML ACS (IdP-initiated) ignored `disableImplicitSignUp`**  
Users were created even when implicit sign-up was disabled.  
This is now fixed by routing SAML callbacks through the unified linking pipeline.

## Breaking Change

SSO errors now follow standard redirect-based protocol behavior:

**Before:**  
`HTTP 401 UNAUTHORIZED` thrown when `disableImplicitSignUp` was enabled and the user did not exist.

**After:**  
A `302` redirect occurs with `?error=signup_disabled` in the callback URL.

This aligns with how SAML and OIDC flows expect error handling and prevents breaking the browser redirect chain.

---------
fixes https://github.com/better-auth/better-auth/issues/4972

### DIFF
--- a/packages/better-auth/src/db/field.ts
+++ b/packages/better-auth/src/db/field.ts
@@ -102,10 +102,10 @@ export type InferFieldsInputClient<Field> =
 					: Field[key]["required"] extends false
 						? key
 						: Field[key]["defaultValue"] extends
-								| string
-								| number
-								| boolean
-								| Date
+									| string
+									| number
+									| boolean
+									| Date
 							? key
 							: never]?: InferFieldInput<Field[key]> | undefined | null;
 			}


packages/sso/src/routes/sso.ts
```diff
@@ -1687,73 +1667,43 @@ export const callbackSSOSAML = (options?: SSOOptions) => {
...
-				}
-				user = existingUser;
-			} else {
-				// if implicit sign up is disabled, we should not create a new user nor a new account.
-				if (options?.disableImplicitSignUp) {
-					throw new APIError("UNAUTHORIZED", {
-						message:
-							"User not found and implicit sign up is disabled for this provider",
-					});
-				}
```
Paola3stefania, line 1906: <p class="p2"><span class="s2"><h3><b>What changed</b></h3></span></p>
<p class="p1">When <span class="s3">disableImplicitSignUp: true</span> and a user does not exist, SSO endpoints now <span class="s1"><b>redirect</b></span> with an error instead of throwing an HTTP error.</p>

Before | After
-- | --
HTTP 401 UNAUTHORIZED with JSON body | 302 Redirect to ?error=signup_disabled

<p class="p2"><span class="s1"><h3><b>Why</b></h3></span></p>
<p class="p1"><span class="s1"><ul><li>

<p class="p1">User is in a browser-based redirect flow</p>
</li><li>
<p class="p1">Throwing HTTP errors breaks the flow and shows ugly error pages</p>
</li><li>
<p class="p1">Redirecting to callback URL lets your app handle errors gracefully</p>
</li><li>
<p class="p1">Consistent with OAuth callback behavior in better-auth
</p>

</li></ul></span></p>
<p class="p1"><span class="s1"><h3><b>Notes</b></h3></span></p>
<p class="p3">users now should update the code to check the <span class="s2">error</span> query parameter in the callback URL:</p>
<pre><code>// Before
try {
  await auth.api.callbackSSOSAML(...);
} catch (e) {
  if (e.status === &quot;UNAUTHORIZED&quot;) {
    // handle
  }
}

// After — inspect callback URL
// Example: /callback?error=signup_disabled
const error = new URL(request.url).searchParams.get(&quot;error&quot;);
if (error === &quot;signup_disabled&quot;) {
  // handle
}</code></pre>


</body>
</html>

PR #6661, opened 2025-12-10
feat: add support for a custom betterAuth initializer that support intercepting and modifying the context

yacobmole:
Summary
- add a `betterAuth` initializer variant that allows intercepting and mutating the resolved auth context
- expose the custom initializer through the package export map and cover the behavior with vitest

```ts
export const auth = betterAuth(
  {
    appName: "My App",
  },
  (ctx) => {
    ctx.appName = "My Better App";
    return ctx;
  },
);
```

---
## Summary by cubic
Added a custom `betterAuth` initializer to intercept and mutate the `AuthContext` after init. Covered by tests.

- **New Features**
  - New `./custom` entry: `betterAuth(options, interceptor?)` applies a context interceptor after `init()`.
  - Validates interceptor and throws `BetterAuthError` if not a function.
  - Adds `typesVersions` for `./custom`; preserves `Auth` types and reflects mutations in `auth.$context`.

<sup>Written for commit 862bba4c652e11e3ca8cfebcb40da8dd4fb4b756. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/package.json
+++ b/packages/better-auth/package.json
@@ -51,6 +51,11 @@
       "types": "./dist/auth/minimal.d.mts",
       "default": "./dist/auth/minimal.mjs"
     },
+    "./custom": {
+      "dev-source": "./src/auth/custom.ts",
+      "types": "./dist/auth/custom.d.mts",
+      "default": "./dist/auth/custom.mjs"
+    },
     "./social-providers": {
       "dev-source": "./src/social-providers/index.ts",
       "types": "./dist/social-providers/index.d.mts",
--- a/packages/better-auth/package.json
+++ b/packages/better-auth/package.json
@@ -335,6 +335,9 @@
       "minimal": [
         "./dist/auth/minimal.d.mts"
       ],
+      "custom": [
+        "./dist/auth/custom.d.mts"
+      ],
       "node": [
         "./dist/integrations/node.d.mts"
       ],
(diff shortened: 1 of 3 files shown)


himself65 (changes requested):
ctx is for internal use. I believe there re many ways like before and after hooks to do the same thing. Adding more endpoints are not worth

yacobmole:
> ctx is for internal use. I believe there re many ways like before and after hooks to do the same thing. Adding more endpoints are not worth

You are right, the hooks do cover the context for the requests however; The $context property (`betterAuth({}).$context`) will not have the context updated.

```ts
  appName: "Better Auth",
  hooks: {
      ctx.context.appName = "My New App";
    }),
      console.log(ctx.context.appName); // "My New App"
    }),
  },
});

const authContext = await auth.$context;
console.log(authContext.appName); // "Better Auth"
```

This initializer lets us make one-time, deterministic mutations to the auth context at boot so that consuming context at Request Level and also Application Level is the exact same context.

This is important for better-auth compatible library authors that may consume the auth.$context

yacobmole:
@himself65 What if instead of adding a new endpoint, would it be ideal to instead export both the `createBetterAuth` and `init` (including `initMinimal`) methods allowing more advanced better-auth setups instead of predefined fixed setups that lack customizing fully (i.e. how `/minimal` is without kysely)

Could export them under `/custom`?

PR #6664, opened 2025-12-10
docs: useSession hook is a field not a method

zekariyasamdu:
When using useSession as suggested in the docs like this.

```
 const { data: session } = authClient.useSession();
```
It causes a typescript error saying " This expression is not callable ".

fix: call the get() method on useSession to get session data,
```
 const { data: session } = authClient.useSession.get();
```

---
## Summary by cubic
Corrected the docs for useSession: it’s a field with a get() method, not a callable hook. Updated examples to use authClient.useSession.get() to avoid the TypeScript “not callable” error, and added a Framework-Specific Clients section with correct import sources (client, react, svelte-kit, vue, tanstack-start) plus a note that framework packages expose reactive features like React Hooks.

<sup>Written for commit aff3b05b5056ae6758617f225a488592f5be3e37. Summary will update on new commits.</sup>


docs/content/docs/concepts/session-management.mdx
```diff
@@ -101,7 +101,7 @@ The `useSession` action provides a reactive way to access the current session.
 ```ts client="client.ts"
 import { authClient } from "@/lib/client"
 
-const { data: session } = authClient.useSession()
+const { data: session } = authClient.useSession.get()
```
himself65: This means you are using the vanilla client or something similar. This API exists in a React client. So I would recommend to add section for selection different client

zekariyasamdu: Yes! actually it looks like I imported:
```
import { createAuthClient } from "better-auth/client";
```

instead of:
```
```

I will add a section that talks about the different clients on: _docs/concepts/client_

himself65: nice

PR #6669, opened 2025-12-10
docs: add IdP-initiated SAML SSO workaround for Next.js

### LINKED ISSUE #6615: IdP-initiated SAML SSO returns 404

### To Reproduce

1. Set up Better Auth with standard catch-all route:
 ```typescript
 // app/api/auth/[...all]/route.ts
 export const { GET, POST } = toNextJsHandler(auth.handler)
```
2. Register a SAML SSO provider with:
- callbackUrl: https://example.com/api/auth/sso/saml2/callback/my-provider
3. Test SP-initiated flow:
- Call authClient.signIn.sso({ email: "user@example.com" })
- Result: ✅ Works
4. Test IdP-initiated flow:
- Click app icon from Okta dashboard
- Okta POSTs SAMLResponse to callback URL
- Better Auth processes and returns 302 redirect to same callback URL
- Browser follows 302 with GET request
- Result: ❌ 404 Not Found (catch-all doesn't match the specific path)
5. Workaround: Create explicit route with BOTH GET and POST:
```ts
// app/api/auth/sso/saml2/callback/[providerId]/route.ts
export async function POST(req: NextRequest) {
  return auth.handler(req)
}

export async function GET() {
  // Required because POST returns 302 to same URL
  return NextResponse.redirect(new URL('/', process.env.BETTER_AUTH_URL))
}
```
- POST only → 405 Method Not Allowed (GET fails)
- GET only → POST fails
- Both → ✅ Works
6. Request flow (from server logs):
- POST /api/auth/sso/saml2/callback/my-provider 302
- GET  /api/auth/sso/saml2/callback/my-provider 307 (redirect to `/`)
- GET  / 200

  **Current behavior:**
  1. Register SAML provider with catch-all route only (`app/api/auth/[...all]/route.ts`)
  2. IdP-initiated POST to callback URL
  3. Better Auth processes and returns **302 redirect to same callback URL**
  4. Browser follows 302 with GET request
  5. **GET returns 404** (catch-all not matched) or **405** (if explicit POST-only route exists)

  **Expected behavior:**
  Better Auth should either:
  1. Return 302 redirect to a different URL (like `/` or configured callback)
  2. Or document that both GET and POST handlers are required for SAML callback

better-auth: ^1.4.4, @better-auth/sso: ^1.4.4 (included in better-auth), Next.js: 15.5.7, React: 19.1.0, Node.js: v24.10.0

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64"
  },
  "node": {
    "version": "v24.10.0"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.0"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "15.5.7"
    },
    {
      "name": "react",
      "version": "19.1.0"
    }
  ],
  "databases": [
    {
      "name": "@prisma/client",
      "version": "6.16.2"
    }
  ],
  "betterAuth": {
    "version": "^1.4.4",
    "config": {
      "appName": "[REDACTED]",
      "baseURL": "[REDACTED]",
      "advanced": {
        "disableCSRFCheck": true
      },
      "session": {
        "cookieCache": {
          "enabled": true,
          "maxAge": 300
        },
        "expiresIn": 604800,
        "updateAge": 86400
      },
      "plugins": [
        { "name": "next-cookies" },
        { "name": "organization" },
        {
          "name": "sso",
          "config": {
            "id": "sso",
            "schema": {
              "ssoProvider": {
                "fields": {
                  "issuer": { "type": "string", "required": true },
                  "samlConfig": { "type": "string", "required": false },
                  "providerId": { "type": "string", "required": true, "unique": true },
                  "organizationId": { "type": "string", "required": false },
                  "domain": { "type": "string", "required": true }
                }
              }
            }
          }
        }
      ]
    }
  }
}
```

Client, Documentation, Backend

```typescript
// lib/auth.ts (sanitized for GitHub issue)
import { prismaAdapter } from 'better-auth/adapters/prisma'
import { nextCookies } from 'better-auth/next-js'
import { organization } from 'better-auth/plugins'
import { sso } from '@better-auth/sso'

  appName: 'My App',
  baseURL: process.env.BETTER_AUTH_URL, // e.g., https://example.com

    provider: 'postgresql',
  }),

  session: {
    cookieCache: {
      enabled: true,
      maxAge: 60 * 5,
    },
    expiresIn: 60 * 60 * 24 * 7,
    updateAge: 60 * 60 * 24,
  },

  plugins: [
    nextCookies(),
    organization({
      // ... organization config
    }),
    sso({
      providerRegistration: {
        disabled: false,
      },
      organizationProvisioning: {
        disabled: false,
        defaultRole: 'member',
      },
      errorCallbackURL: '/auth/sign-in',
    }),
  ],

})

// SAML Provider registration example
await auth.api.registerSSOProvider({
  body: {
    providerId: 'my-okta-provider',
    issuer: 'https://example.com',
    domain: 'example.com',
    samlConfig: {
      issuer: 'https://example.com',
      entryPoint: 'https://my-company.okta.com/app/xxx/sso/saml',
      cert: '[REDACTED - IdP Certificate]',
      callbackUrl: 'https://example.com/api/auth/sso/saml2/callback/my-okta-provider',
      audience: 'https://example.com',
      wantAssertionsSigned: true,
      signatureAlgorithm: 'sha256',
      identifierFormat: 'urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress',
      idpMetadata: {
        entityID: 'http://www.okta.com/xxx',
        cert: '[REDACTED]',
        singleSignOnService: [{
          Binding: 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect',
          Location: 'https://my-company.okta.com/app/xxx/sso/saml',
        }],
      },
      spMetadata: {
        entityID: 'https://example.com',
        binding: 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST',
      },
      mapping: {
        id: 'nameID',
        email: 'email',
        name: 'displayName',
      },
    },
  },
})
```

### Additional context

Why IdP-initiated SSO matters:
- Enterprise customers often access apps from their IdP dashboard (Okta, Azure AD, Google Workspace)
- IT admins prefer centralized app access through IdP portals
- Many enterprise SSO requirements mandate IdP-initiated flow support

**Real-world debugging experience:**

I spent several hours debugging why SSO worked perfectly in my local environment but customers reported 404 errors. The challenge was:

- SP-initiated flow (which I tested locally) worked fine
- IdP-initiated flow (which customers used from their Okta dashboard) failed with 404
- I couldn't ask enterprise customers to open DevTools and debug for me
- The only feedback I had was "it shows 404"

The root cause was that Better Auth's POST handler returns 302 redirect to the same callback URL, which requires a GET handler that doesn't exist by default.

Adding documentation or auto-registering this callback route would save other developers from the same confusion.

GautamBytes:
### Description
This PR addresses the issue where IdP-initiated SAML SSO flows (e.g., clicking the app icon from an Okta dashboard) fail with a 404  error in Next.js App Router.

### The Issue
As discussed in #6615, when the IdP POSTs to the callback URL, Better Auth redirects (302). The browser follows this with a `GET` request to the same URL. Since the default setup relies on a catch-all route that may not handle this specific `GET` context, the flow breaks.

### The Fix
I have added a new documentation section **"IdP-Initiated SSO (Next.js)"** to `docs/content/docs/plugins/sso.mdx`. It provides the code snippet for creating an explicit route handler (`route.ts`) that manages both the `POST` (handshake) and the `GET` (redirect) methods.

### Related Issue
Fixes #6615

---
## Summary by cubic
Adds documentation for a Next.js workaround to support IdP-initiated SAML SSO and prevent 404s after the IdP redirect. Includes a route handler snippet that handles both POST (handshake) and GET (redirect) on the callback URL.

<sup>Written for commit 30cc9d558c2b3e96ff8c848962f747979b2d7623. Summary will update on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/sso.mdx
+++ b/docs/content/docs/plugins/sso.mdx
@@ -430,23 +430,29 @@ await auth.api.registerSSOProvider({
     </Tab>
 </Tabs>
 
-### IdP-Initiated SSO (Next.js)
+### IdP-Initiated SSO
 
-For IdP-initiated flows (e.g., via Okta dashboard), Next.js requires an explicit route handler to manage the redirect. Create this file to prevent 404 errors:
+ For IdP-initiated flows (e.g., via Okta dashboard), your framework may require an explicit route handler to manage the redirect if the default handler doesn't support the `GET` request following the SAML POST.
 
-```ts title="app/api/auth/sso/saml2/callback/[providerId]/route.ts"
-import { auth } from "@/lib/auth";
-import { NextResponse } from "next/server";
+ <Tabs items={["next-js-app-router"]}>
+   <Tab value="next-js-app-router">
+     Create this file to prevent 404 errors:
 
-export async function POST(req: Request) {
-    return auth.handler(req);
-}
+     ```ts title="app/api/auth/sso/saml2/callback/[providerId]/route.ts"
+     import { auth } from "@/lib/auth";
+     import { NextResponse } from "next/server";
 
-export async function GET(req: Request) {
-    // Required: IdP-initiated flows redirect to this URL after POST
-    return NextResponse.redirect(new URL("/", req.url));
-}
-```
+     export async function POST(req: Request) {
+         return auth.handler(req);
+     }
+
+     export async function GET(req: Request) {
+         // Required: IdP-initiated flows redirect to this URL after POST
+         return NextResponse.redirect(new URL("/", req.url));
+     }
+     ```
+   </Tab>
+</Tabs>
 
 ### Get Service Provider Metadata
 


EthanKim88:
I'd like to clarify some misunderstandings in the review comments.

I only had a single catch-all route file in (/api/auth/):

```typescript
// app/api/auth/[...all]/route.ts
import { auth } from '@/lib/auth'
import { toNextJsHandler } from 'better-auth/next-js'

export const { GET, POST } = toNextJsHandler(auth.handler)
```

This is the standard Better Auth setup as documented. No explicit SAML callback route was created.

The Core Issue: IdP-Initiated vs SP-Initiated Flow

SP-Initiated Flow (works correctly):
1. User calls authClient.signIn.sso({ callbackURL: '/dashboard' })
2. Better Auth redirects to IdP with RelayState=/dashboard
3. IdP authenticates and POSTs SAMLResponse + RelayState back to callback
4. Better Auth redirects to /dashboard ✅

IdP-Initiated Flow (404 error):
1. User clicks app icon directly from Okta/Azure AD dashboard
2. IdP POSTs SAMLResponse to callback URL without RelayState (SP was never involved)
3. Better Auth executes this code:
const callbackUrl = RelayState || parsedSamlConfig.callbackUrl || ctx.context.baseURL;
throw ctx.redirect(callbackUrl);
4. Since RelayState is empty, it falls back to parsedSamlConfig.callbackUrl
5. callbackUrl is the ACS URL itself (e.g., /api/auth/sso/saml2/callback/my-provider)
6. Browser follows 302 redirect with GET request to the same URL → 404

Addressing the Review Comments

"GET requests to the SAML callback endpoint are not part of the normal flow"

This is incorrect for IdP-initiated flows. When RelayState is absent and callbackUrl equals the ACS URL, Better Auth redirects to itself, causing a GET request.

"The catch-all route should handle this"

The catch-all route does export both GET and POST via toNextJsHandler(). However, Better Auth's internal SAML callback endpoint (/sso/saml2/callback/:providerId) only registers a POST method handler. When the 302 redirect happens and the browser sends a GET request, Better Auth has no GET handler registered for that specific endpoint, resulting in 404.

"callbackUrl should be set to something else"

This is not possible. Per SAML specification, callbackUrl must be the ACS (Assertion Consumer Service) URL where the IdP sends the SAMLResponse. You cannot register a different URL - the IdP won't accept it.

The workaround I documented (adding an explicit GET handler in a separate route file) is a valid solution until Better Auth addresses this design issue. #6615

EthanKim88:
The root cause is that Better Auth's SAML callback endpoint only registers a POST handler
- https://github.com/better-auth/better-auth/blob/main/packages/sso/src/routes/sso.ts#L1458-L1487

When IdP-initiated flow redirects back to the same URL with a GET request (due to empty `RelayState`), there's no GET handler to catch it, resulting in 404.

docs/content/docs/plugins/sso.mdx
```diff
@@ -258,6 +258,24 @@ await auth.api.registerSSOProvider({
     </Tab>
 </Tabs>
 
+### IdP-Initiated SSO (Next.js)
```
himself65: Can we use tabs here? And it might not only happen in Next.js.

for example: https://www.better-auth.com/docs/installation#mount-handler

kanarian:
I had the same problem happens for `api/auth/sso/saml2/sp/acs/:providerId]` and used the same solution as here to fix it (just manually add the api route to my next app).

himself65 (approved):

PR #6675, opened 2025-12-10
fix(saml): IdP-Initiated Callback Routing

Paola3stefania:
---
## Summary by cubic
Fix SAML IdP-initiated callback routing with GET support, safe redirects, and persisted RelayState. IdP- and SP-initiated flows now redirect reliably without loops and respect requestSignUp.

- **Bug Fixes**
  - Allow GET and POST on /sso/saml2/callback/:providerId; handle RelayState via body or query.
  - Validate session on GET; redirect to error URL if missing.
  - Generate and validate RelayState; fall back to app origin or configured callbackUrl; prevent open redirects and callback-to-callback loops.
  - Require SAMLResponse on POST; perform safe redirect after session is set.
  - Respect requestSignUp when disableImplicitSignUp is true.
  - Skip origin checks for SAML callback and ACS endpoints to allow external IdP POSTs; other routes remain protected.

- **Refactors**
  - Introduce generic state helpers shared by OAuth2 and SAML RelayState.

<sup>Written for commit c5e51003649d7c46d4b8b4bdea9b5b93e73b3f17. Summary will update on new commits.</sup>

This fixes https://github.com/better-auth/better-auth/issues/6615

### DIFF
--- a/packages/better-auth/src/state.ts
+++ b/packages/better-auth/src/state.ts
@@ -36,12 +36,14 @@ export class StateError extends BetterAuthError {
 
 	constructor(
 		message: string,
-		opts: { code: StateErrorCode; details?: Record<string, any> },
-		errorOptions?: ErrorOptions,
+		options: ErrorOptions & {
+			code: StateErrorCode;
+			details?: Record<string, any>;
+		},
 	) {
-		super(message, errorOptions?.cause as string | undefined);
-		this.code = opts.code;
-		this.details = opts.details;
+		super(message, options);
+		this.code = options.code;
+		this.details = options.details;
 	}
 }
 
@@ -151,8 +153,8 @@ export async function parseGenericState(
 				{
 					code: "state_invalid",
 					details: { state },
+					cause: error,
 				},
-				{ cause: error },
 			);
 		}
 


rajat1saxena:
Eagerly waiting for this fix.

For now, going with the suggested fix mentioned in #6615.

packages/better-auth/src/state.ts
```diff
@@ -0,0 +1,218 @@
+import type { GenericEndpointContext } from "@better-auth/core";
```
Bekacru, line 1: why did we make 2 state.ts files? can't we move this to the one inside `oauth2`?

Paola3stefania: Yeah I get why you are pointing this, but the separation is intentional though — the generic **state.ts** is shared by both OAuth2 and SAML.

The SSO package imports these functions from better-auth to handle relay state, so it needs to live at the top level rather than inside oauth2/.

packages/better-auth/src/oauth2/state.ts
```diff
@@ -24,200 +21,55 @@ export async function generateState(
...
 		codeVerifier,
 		errorURL: c.body?.errorCallbackURL,
 		newUserURL: c.body?.newUserCallbackURL,
 		link,
-		/**
-		 * This is the actual expiry time of the state
-		 */
 		expiresAt: Date.now() + 10 * 60 * 1000,
 		requestSignUp: c.body?.requestSignUp,
-		state,
```
himself65, line 42: Why do we remove `state` here?

Paola3stefania: Its was the result of the refactor, the state field was removed from stateData because it was redundant.

The **state** is an identifier, not part of the payload data.
For database strategy: state is stored as the identifier in the verification table - no need to duplicate it inside the JSON value.
For cookie strategy: state is returned in the URL query param and validated separately - not needed inside the encrypted cookie.

https://github.com/better-auth/better-auth/blob/b057064e922d5d6f1c69ea2f20e1b73675d7eb7a/packages/better-auth/src/social.test.ts#L725 -> this test confirms the state behavior is working correctly.

- generateState() → stores data, returns state id
- OAuth provider redirects back with state in URL
- parseState() → uses state to look up and return the original data

(saml and social tests also address this)

--------

**before**:
```
const state = generateRandomString(32);
const stateData = {
  ...
  state,  // state was stored INSIDE stateData
};
```

The state was generated first, then included inside stateData, and then was stored (encrypted in a cookie or in the database with state as id)

**after**
````
const stateData: StateData = {
  ...
  // no state field
};
return await generateGenericState(c, stateData);
````

In **generateGenericState** [See Here](https://github.com/better-auth/better-auth/pull/6675/changes/BASE..665bd1b0a0545a038ec4554677fe830a1b0420dc#diff-311f70c350a8c0d2245c083645a327451d04cb39e54ab51acc33749c15bb74a1R47:)

````
export async function generateGenericState(c, stateData, settings?) {

  const state = generateRandomString(32);  // state is generated HERE now

  // ...stores stateData, returns { state, codeVerifier }
}
````

File: packages/better-auth/src/oauth2/state.ts [hunk 3]
```diff
@@ -24,200 +21,55 @@ export async function generateState(
...
 
 	await setOAuthState(stateData);
 
-	if (storeStateStrategy === "cookie") {
-		// Store state data in an encrypted cookie
-		const encryptedData = await symmetricEncrypt({
-			key: c.context.secret,
-			data: JSON.stringify(stateData),
+	try {
+		return await generateGenericState(c, stateData);
```
himself65: @claude do not use return await

packages/better-auth/src/utils/url.ts
```diff
@@ -183,3 +183,47 @@ export function getHost(url: string) {
...
+ *          or "/" if URL parsing fails
+ *
+ * @example
+ * normalizePathname("http://localhost:3000/api/auth/sso/saml2/callback/provider1", "/api/auth")
+ * // Returns: "/sso/saml2/callback/provider1"
+ *
+ * normalizePathname("http://localhost:3000/sso/saml2/callback/provider1/", "/")
+ * // Returns: "/sso/saml2/callback/provider1"
+ */
+export function normalizePathname(
```
himself65: @claude let's move this util to `@better-auth/core/utils/url`

packages/better-auth/src/state.ts
```diff
@@ -0,0 +1,218 @@
...
+
+export type StateData = z.infer<typeof stateDataSchema>;
+
+export type StateErrorCode =
+	| "state_generation_error"
+	| "state_invalid"
+	| "state_mismatch"
+	| "state_security_mismatch";
+
+export class StateError extends Error {
```
himself65: @claude Please make it extends BetterAuthError

PR #6685, opened 2025-12-11
feat(siwe): allow linking SIWE wallets to existing accounts

META-DREAMER:
## Summary

Adds support for linking SIWE wallets to existing authenticated user accounts, with cross-chain wallet identity protection.

## Changes

- **Account Linking Support:** Respects accountLinking.enabled and accountLinking.trustedProviders configuration. When authenticated and linking is enabled (with siwe in trusted providers), wallets are linked to the existing account instead of creating a new user.
    - Creates new users when not authenticated (original behaviour preserved)
    - Returns success if wallet is already linked to current user (idempotent)
    - Throws `WALLET_ALREADY_LINKED` error if wallet belongs to another user
    - Cascades deletion of walletAddress when account is unlinked
- **Cross-Chain Identity Protection:** Prevents different users from linking the same wallet address on different chains. A wallet address represents the same key pair across all chains, so linking 0x123... on Ethereum should not be allowed if it's already linked to another user on Polygon.

## Test plan

- [x] Added test: links wallet when `accountLinking.trustedProviders` includes `siwe`
- [x] Added test: links wallet when `trustedProviders` is undefined (default behavior)
- [x] Added test: does NOT link when `trustedProviders` is set but doesn't include `siwe`
- [x] Added test: does NOT link when `accountLinking.enabled` is `false`
- [x] Added test: returns success if wallet already linked to current user (idempotent)
- [x] Added test: verify ENS lookup functionality correctly sets user name / photo on account creation
- [x] Added test: rejects if wallet already linked to another user
- [x] Added test: Rejects linking if wallet exists on another chain for a different user (Same user can link same wallet on multiple chains)
- [x] Added test: creates new user when not authenticated (no session)
- [x] Added test: ensure walletAddress records are deleted when account is unlinked from user
- [x] All existing SIWE tests continue to pass (40 tests total)

---
## Summary by cubic
Enables linking SIWE wallets to an authenticated user's existing account when account linking is enabled and SIWE is trusted. Keeps the original behavior for unauthenticated users and adds safeguards for idempotency and cross-chain conflicts.

- **New Features**
  - Link wallet to the current session user when accountLinking.enabled is true and trustedProviders includes siwe (or is unset/empty).
  - Return success if the wallet is already linked to the current user (idempotent).
  - Throw WALLET_ALREADY_LINKED when the wallet belongs to another user, including if the same address exists on a different chain for another user.
  - Preserve existing flow: create a new user when no session.
  - Include the current session token in successful wallet linking responses.
  - Create an account record for each verified wallet (including linked wallets).
  - Cascade delete the corresponding walletAddress when a SIWE account is unlinked or the user is deleted.

- **Refactors**
  - Extract shared walletAddress and chainId schemas.
  - Add session detection via getSessionFromCtx.
  - Streamline nonce handling; standardize SIWE error codes and wallet account ID construction.

<sup>Written for commit 3d3e701ad63e9aa24846dc115f4b2edb41e17b1b. Summary will update on new commits.</sup>


META-DREAMER:
cc @rokitgg

packages/better-auth/src/plugins/siwe/index.ts
```diff
@@ -95,206 +95,236 @@ export const siwe = (options: SIWEPluginOptions) =>
...
-							status: 400,
+					// Verify nonce exists and is not expired
+					const verification =
+						await ctx.context.internalAdapter.findVerificationValue(
+							`siwe:${walletAddress}:${chainId}`,
+						);
+
+					if (!verification || new Date() > verification.expiresAt) {
+						throw new APIError("UNAUTHORIZED", {
+							message: "Unauthorized: Invalid or expired nonce",
```
himself65: please use ERROR_CODEs

packages/better-auth/src/plugins/siwe/index.ts
```diff
@@ -95,206 +95,236 @@ export const siwe = (options: SIWEPluginOptions) =>
...
+						});
+					}
+
+					// Clean up used nonce
+					await ctx.context.internalAdapter.deleteVerificationValue(
+						verification.id,
+					);
+
+					const existingWallet: WalletAddress | null =
+						await ctx.context.adapter.findOne({
```
himself65:
```suggestion
						await ctx.context.adapter.findOne<WalletAddress>({
```

packages/better-auth/src/plugins/siwe/index.ts
```diff
@@ -95,206 +95,236 @@ export const siwe = (options: SIWEPluginOptions) =>
...
+							trustedProviders.includes("siwe"));
+
+					if (shouldLink) {
+						const sessionUser = currentSession.user;
+
+						if (existingWallet) {
+							if (existingWallet.userId !== sessionUser.id) {
+								throw new APIError("BAD_REQUEST", {
+									message: "Wallet already linked to another account",
+									code: "WALLET_ALREADY_LINKED",
```
himself65: same

packages/better-auth/src/plugins/siwe/index.ts
```diff
@@ -95,206 +95,236 @@ export const siwe = (options: SIWEPluginOptions) =>
...
-								},
-								s: { t: "eip191", s: signature },
+						await ctx.context.adapter.create({
+							model: "walletAddress",
+							data: {
+								userId: sessionUser.id,
+								address: walletAddress,
+								chainId,
+								isPrimary: false,
+								createdAt: new Date(),
```
himself65, line 279: please remove this, `createdAt` is automatically created

packages/better-auth/src/plugins/siwe/index.ts
```diff
@@ -95,206 +95,236 @@ export const siwe = (options: SIWEPluginOptions) =>
...
-						if (!verified) {
-							throw new APIError("UNAUTHORIZED", {
-								message: "Unauthorized: Invalid SIWE signature",
-								status: 401,
-							});
-						}
+						await ctx.context.internalAdapter.linkAccount({
+							userId: sessionUser.id,
+							providerId: "siwe",
+							accountId: `${walletAddress}:${chainId}`,
```
himself65: let's abstract `${walletAddress}:${chainId}` to a top level variable

packages/better-auth/src/plugins/siwe/index.ts
META-DREAMER: Doesn't seem to be the case, at least for the walletAddress. When I remove it the tests fail with SQLITE error NON NULL failed assertion. Maybe the test sqlite DB doesn't handle automatically setting createdAt? 

 Removed it for the createAccount call though and tests still passed, but that seems like it only worked because there's this code here: https://github.com/better-auth/better-auth/blob/b56d7b8eaad9c8cc222f274bbb0d3d6fe1e8a750/packages/better-auth/src/db/internal-adapter.ts#L89-L91

META-DREAMER:
@himself65 Fixed all the merge conflicts and addressed all the feedback. Let me know if we can get this merged soon, ty!

PR #6696, opened 2025-12-11
feat: remove timestamp assignments un internalAdapter

Ridhim-RR:
### 🎯 **Description**
Removes redundant manual timestamp assignments from `internal-adapter.ts` as per the TODO comment:

> `// todo: we should remove auto setting createdAt and updatedAt in the next major release, since the db generators already handle that`

---
## Summary by cubic
Removed manual createdAt/updatedAt assignments in internal-adapter and set updatedAt on create via table defaults for session and account. This centralizes timestamp logic and prevents mismatches.

- **Refactors**
  - Dropped timestamp fields from createWithHooks (user, account, session, verification).
  - Added defaultValue for updatedAt in getAuthTables (session, account); onUpdate continues to handle updates.

<sup>Written for commit ac443998bf0f60aa5cf30a249d35673f1b91f734. Summary will update automatically on new commits.</sup>


packages/core/src/db/adapter/utils.ts
```diff
@@ -24,6 +24,14 @@ export function withApplyDefault(
 				}
 				return field.defaultValue;
 			}
+			// If no defaultValue but onUpdate exists, use onUpdate for creation
+			// This handles fields like updatedAt that should be set on both create and update
+			if (field.onUpdate !== undefined) {
```
himself65: What does this mean? Could you please add a test case to cover this?

Ridhim-RR: After removing the manual assignments the test cases of the adapters failed ::
Serialized Error: { code: 'SQLITE_CONSTRAINT_NOTNULL' } 
  Error: SqliteError: NOT NULL constraint failed: account.updatedAt
  
  As in account and session table the updatedAt does not have default value only onUpdate.
  
  As the code in the factory.ts 
  if (
       value === undefined &&
       ((fieldAttributes!.defaultValue === undefined &&
         !fieldAttributes!.transform?.input &&
         !(action === "update" && fieldAttributes!.onUpdate)) ||  
         (action === "update" && !fieldAttributes!.onUpdate))
     ) {
       continue; 
     }
Fields with only onUpdate (no defaultValue) were skipped
due to which it was null thats why I have make changes in the fatctory.ts as well in the utils

Ridhim-RR: But now i think if i add defaultValue in the account as well as in the session then no need for these changes in the factory.ts as well in the utils

himself65: LGTM

PR #6719, opened 2025-12-12
feat(saml): enforce one-time use of SAML assertions

Paola3stefania:
## Summary by cubic
Adds SAML assertion replay protection to enforce one-time use of assertions during SSO callback and ACS. Assertion IDs are stored with a TTL and replays are rejected with an error redirect.

- **New Features**
  - Tracks each SAML Assertion ID in the verification store until NotOnOrAfter + clock skew (or a default TTL) and rejects re-use with error=replay_detected.
  - Replay protection is always enabled (no configuration toggle).
  - Centralized SAML constants in constants.ts (key prefixes, TTLs, clock skew) and added USED_ASSERTION_KEY_PREFIX.
  - Extracts assertionId from SAML XML using fast-xml-parser.
  - Expanded tests covering callback, ACS, and cross-endpoint replay scenarios with a shared mock IdP.

- **Migration**
  - DEFAULT_CLOCK_SKEW_MS and DEFAULT_AUTHN_REQUEST_TTL_MS are now exported from constants.ts and are no longer re-exported from the package root. Update imports accordingly.
  - Removed AuthnRequestStore, createInMemoryAuthnRequestStore, and the saml.authnRequestStore option. InResponseTo validation now always uses the verification table—remove any custom store usage.

<sup>Written for commit 24ffbc6bca2051e36526dcef411a3f4573ee133c. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/integrations/convex.mdx
+++ b/docs/content/docs/integrations/convex.mdx
@@ -63,7 +63,7 @@ If you're not using Next.js, support for other frameworks is documented in the [
         </Callout>
 
         ```package-install
-        npm install better-auth@1.3.27 --save-exact
+        npm install better-auth@1.4.7 --save-exact
         npm install convex@latest @convex-dev/better-auth
         ```
     </Step>
@@ -87,17 +87,13 @@ If you're not using Next.js, support for other frameworks is documented in the [
 
         Add a `convex/auth.config.ts` file to configure Better Auth as an authentication provider.
 
-
-
         ```ts title="convex/auth.config.ts"
+        import { getAuthConfigProvider } from "@convex-dev/better-auth/auth-config";
+        import type { AuthConfig } from "convex/server";
+
         export default {
-            providers: [
-                {
-                    domain: process.env.CONVEX_SITE_URL,
-                    applicationID: "convex",
-                },
-            ],
-        };
+            providers: [getAuthConfigProvider()],
+        } satisfies AuthConfig;
         ```
     </Step>
     <Step>
@@ -128,7 +124,7 @@ If you're not using Next.js, support for other frameworks is documented in the [
         NEXT_PUBLIC_CONVEX_SITE_URL=https://adjective-animal-123.convex.site # [!code ++]
 
         # Your local site URL // [!code ++]
-        SITE_URL=http://localhost:3000 # [!code ++]
+        NEXT_PUBLIC_SITE_URL=http://localhost:3000 # [!code ++]
         ```
 
         ```shell title=".env.local" tab="Self hosted"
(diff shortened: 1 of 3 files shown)


packages/sso/src/types.ts
```diff
@@ -339,5 +339,18 @@ export interface SSOOptions {
...
+		 *
+		 * This prevents replay attacks where an attacker captures a valid
+		 * SAMLResponse and resubmits it to gain unauthorized access.
+		 *
+		 * Storage: Uses the verification table in the database.
+		 * Assertions are tracked until their NotOnOrAfter expiration time.
+		 *
+		 * @default true
+		 */
+		enableReplayProtection?: boolean;
```
Bekacru: there shouldn't be an option to disable this

packages/sso/src/authn-request-store.ts
Bekacru: let's rename this to set to align is secondary storage

packages/sso/src/authn-request-store.ts
```diff
@@ -22,15 +22,9 @@ export interface AuthnRequestStore {
...
-/**
- * Default TTL for AuthnRequest records (5 minutes).
- * This should be sufficient for most IdPs while protecting against stale requests.
- */
-export const DEFAULT_AUTHN_REQUEST_TTL_MS = 5 * 60 * 1000;
-
 /**
  * In-memory implementation of AuthnRequestStore.
- * ⚠️ Only suitable for testing or single-instance non-serverless deployments.
+ * @note Only suitable for testing or single-instance non-serverless deployments.
```
Bekacru: if we only need authnRequestStore to store saml asseration, we should actually remove it and just rely on the `verification` table implementation alone

Paola3stefania: Humm this is a nice discussion to have

I added **authnRequestStore** thinking about enterprise usage : large organizations doing high-volume SAML (10,000+ logins/hour) might want to offload ephemeral auth state to Redis instead of hitting the DB on every login, getting faster lookups, native TTL cleanup, and reduced DB connection pressure in serverless environment

I agree with you that for most users the verification table is perfectly fine so this is really just an optional escape hatch for power users who hit scale issues

packages/sso/src/types.ts
Paola3stefania: Yeah ... Agreed: removed the option. Replay protection is now always on. Even though SAML spec says SHOULD, we're treating it as MUST. The DB write is minimal and there's no legitimate reason to disable it

packages/sso/src/authn-request-store.ts
Paola3stefania: ahh great , ty fot his.

packages/sso/src/authn-request-store.ts
Paola3stefania: I think is nice to support that edge case, but im also fine removing it.

Paola3stefania: Removing it as we'll soon support using secondary storage as a fallback to the verificaiton table instead

Bekacru (approved):

PR #6744, opened 2025-12-14
fix: properly validate body on email signup

### LINKED ISSUE #6743: `signUp.email` throws unhandled TypeError when password is undefined instead of returning validation error

### To Reproduce

1. Enable email/password authentication
2. Call the signup endpoint with a request body missing the password field:

```json
{
  "name": "John Doe",
  "email": "john@example.com"
}
```

**Current behavior**

When calling the email signup endpoint without a `password` field, the handler throws an unhandled `TypeError` instead of returning a proper 400 validation error, resulting in a 500 response with an empty body, and server console prints:
```
TypeError: undefined is not an object (evaluating 'password.length')
    at <anonymous> (/node_modules/better-auth/dist/api-CkmycQ2x.mjs:2342:7)
```

**Expected behavior**
Return a 400 Bad Request with a validation error message like "Password is required".

1.4.5

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:34 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T8112",
    "release": "25.2.0",
    "cpuCount": 8,
    "cpuModel": "Apple M2",
    "totalMemory": "16.00 GB",
    "freeMemory": "0.13 GB"
  },
  "node": {
    "version": "v24.7.0",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.0"
  },
  "frameworks": [
    {
      "name": "hono",
      "version": "^4.7.11"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "^1.3.18",
    "config": null
  }
}
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

**Root Cause**
In `api-CkmycQ2x.mjs` line 2342, password.length is accessed before validating that password exists:
```javascript
const minPasswordLength = ctx.context.password.config.minPasswordLength;
if (password.length < minPasswordLength) {  // crashes here if password is undefined
```

**Suggested Fix**
Add a null check before accessing password.length:
```javascript
if (!password) {
  throw new APIError("BAD_REQUEST", { message: "Password is required" });
}
if (password.length < minPasswordLength) {
  // ...
}
```

jslno:
closes #6743

---
## Summary by cubic
Enforces strict body validation for email sign-up so invalid payloads return 400. Requires name, email, and password while still allowing optional and custom fields.

- **Bug Fixes**
  - Replace permissive body schema with explicit z.object: name, email, password required; image, callbackURL, rememberMe optional; extra fields allowed.
  - Add test that asserts a 400 error when required fields are missing.

<sup>Written for commit 4ae32e721541a1b50ccc89ff9db0b8dacfd317cb. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/mcp.mdx
+++ b/docs/content/docs/plugins/mcp.mdx
@@ -7,6 +7,10 @@ description: MCP provider plugin for Better Auth
 
 The **MCP** plugin lets your app act as an OAuth provider for MCP clients. It handles authentication and makes it easy to issue and manage access tokens for MCP applications.
 
+<Callout type="warn">
+This plugin is based on OIDC Provider plugin. It's currently not ready for production use. We are working on it and will update this documentation when it's ready.
+</Callout>
+
 ## Installation
 
 <Steps>
--- a/docs/content/docs/plugins/phone-number.mdx
+++ b/docs/content/docs/plugins/phone-number.mdx
@@ -134,6 +134,10 @@ export const auth = betterAuth({
 })
 ```
 
+<Callout>
+We highly recommend not awaiting the `sendOTP` function. If you await it, it'll slow down the request and could cause timing attacks. For serverless platforms, you can use `waitUntil` to ensure the OTP is sent.
+</Callout>
+
 ### Sign In with Phone Number
 
 In addition to signing in a user using send-verify flow, you can also use phone number as an identifier and sign in a user using phone number and password.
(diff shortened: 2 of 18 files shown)


rozenmd:
is it just me or was this a breaking change?

forms that previously allowed signups with an empty string name will no longer work

PR #6749, opened 2025-12-14
fix: revert token masking in listSessions route

bytaesu:
## Summary by cubic
Reverted token masking in the listSessions route, returning active sessions with their original token field to restore expected behavior for clients. Expired sessions are still filtered out.

<sup>Written for commit d09ec35fecb335d6193569116415ddc419d57d65. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/session.ts
+++ b/packages/better-auth/src/api/routes/session.ts
@@ -600,16 +600,9 @@ export const listSessions = <Option extends BetterAuthOptions>() =>
 				const sessions = await ctx.context.internalAdapter.listSessions(
 					ctx.context.session.user.id,
 				);
-				const activeSessions = sessions
-					.filter((session) => {
-						return session.expiresAt > new Date();
-					})
-					.map((session) => {
-						return {
-							...session,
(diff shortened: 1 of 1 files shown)


bytaesu:
> What’s the reason we revert it?

If the token is explicitly removed in `/list-sessions`, we should also update the `/revoke-session` request body. This would be a breaking change, so we need to figure out a better approach for now. I think Bereket may have changed this incorrectly.

cc @himself65

mghizzo:
@bytaesu When can we expect this revert change to be published on NPM? Latest version breaks the /list-sessions endpoint and took me quite some time to debug the reason of that empty string.

bytaesu:
> @bytaesu When can we expect this revert change to be published on NPM? Latest version breaks the /list-sessions endpoint and took me quite some time to debug the reason of that empty string.

Hello @mghizzo,
Now you can try this version! [v1.4.8-beta.1](https://github.com/better-auth/better-auth/releases/tag/v1.4.8-beta.1)
Thanks for your patience 🙏

PR #6786, opened 2025-12-15
feat(organization): Add function support to membershipLimit option in organization plugin

### LINKED ISSUE #5517: Remove or make configurable the hard-coded default limit of 100 members per organization
### Description

The organization plugin currently has a hard-coded default limit of 100 members per organization in the `acceptInvitation` function:

https://github.com/better-auth/better-auth/blob/2a02166eefd714e8d599201b8d979e2f82bc462b/packages/better-auth/src/plugins/organization/routes/crud-invites.ts#L525

While this limit can be overridden via `orgOptions.membershipLimit`, the default of 100 is arbitrary and may be too restrictive for many production applications, especially enterprise SaaS platforms that need to support larger organizations.

### Problem

1. Restrictive Default: 100 members is a relatively low limit for many real-world use cases (enterprise teams, large organizations, etc.)
2. Silent Failure Risk: Developers may not be aware of this default until they encounter the limit in production
3. Arbitrary Choice: The number 100 appears to be arbitrary without clear justification for why this specific limit was chosen

### Proposed Solutions

Option 1: Remove the default limit entirely

```typescript
const membershipLimit = ctx.context.orgOptions?.membershipLimit;

if (membershipLimit !== undefined && membersCount >= membershipLimit) {
  throw new APIError("FORBIDDEN", {
    message: ORGANIZATION_ERROR_CODES.ORGANIZATION_MEMBERSHIP_LIMIT_REACHED,
  });
}
```

Option 2: Significantly increase the default

```typescript
const membershipLimit = ctx.context.orgOptions?.membershipLimit || 10000;
```

Option 3: Make it explicit in the configuration

Require developers to explicitly set `membershipLimit` in their organization configuration, with no default value.

### Use Case

We're building an enterprise SaaS platform where organizations regularly exceed 100 members. While we can set membershipLimit explicitly, we believe the default should be either much higher or removed entirely to avoid surprising developers with this limitation.

### Additional Context

A similar pattern exists for invitationLimit which also defaults to 100. It might be worth reviewing all hard-coded limits in the organization plugin.

### Breaking Change Consideration

Changing or removing the default limit could be considered a breaking change if existing users rely on the 100-member limit for their business logic. If that's a concern, this could be introduced as part of a major version bump with clear migration documentation.

amrnt:
## Summary

This PR enhances the `membershipLimit` option in the organization plugin to accept a function that dynamically determines the maximum number of members allowed in an organization. This provides flexibility similar to other limit options like `invitationLimit`.

Updated `OrganizationOptions.membershipLimit` to accept either a `number` or a function `(user: User, organization: Organization) => Promise<number> | number`.

Fixes #5517

---
## Summary by cubic
Adds function support to the organization plugin’s membershipLimit so orgs can set dynamic member caps. Limits are checked when adding members and when accepting invitations; docs and tests updated.

- **New Features**
  - membershipLimit now accepts a number or a function: (user, organization) => Promise<number> | number.
  - Runtime limit enforcement added to addMember and acceptInvitation routes.

- **Bug Fixes**
  - Adapter safeguards ensure a numeric limit is used in queries when membershipLimit is a function (defaults to 100).

<sup>Written for commit 24f5fb8511516e82c3f69850b981f3549c1f4561. Summary will update on new commits.</sup>


packages/better-auth/src/plugins/organization/adapter.ts
amrnt, line 1: I thought the default limit shall be 100 when a function is provided for `membershipLimit`.

Or we could do this: 🤔 

```ts
const org = await adapter.findOrganizationById(data.organizationId);
const defaultLimit = typeof options.membershipLimit === "number" 
  ? options.membershipLimit 
  : await options.membershipLimit(session.user, org);
const limit = data.limit || defaultLimit || 100;
```

amrnt: @himself65 let me know what you think of this part? Otherwise I can go ahead and merge it :)

PR #6787, opened 2025-12-15
chore: dispaly warning when baseURL isn't available 

### LINKED ISSUE #6780: A warning should be printed when the user does not provide the baseURL.

I meet an unreasonable baseURL behavior. If I don't define the baseURL, the google provider always defaults to providing a redirectURI starting with http( the hostname for URL is correct),  which results in a URI MISMATCH error in the production environment. 

> **One of my services on Kubernetes is bound to two domain names simultaneously. However, when I log in using https://a.com (Google Auth), a redirect link to http://b.com is generated** 

```yaml

---

apiVersion: traefik.io/v1alpha1
kind: IngressRoute
spec:
  entryPoints:
    - web
  routes:
    - match: Host(`a.com`)
      kind: Rule
      services:
        - name: api
          port: 3000  
          
    - match: Host(`b.com`)
      kind: Rule
      services:
        - name: api
          port: 3000            
```

It took a long time to figure out why the http protocol appeared, and I checked my codebase, Google Cloud, and environment variables everywhere. I hope this can be mentioned in the documentation or a warning log can be written in the console, which will save a lot of time. 
```typescript
 socialProviders: {
        google: {
            clientId: process.env.GOOGLE_CLIENT_ID || '',
            clientSecret: process.env.GOOGLE_CLIENT_SECRET,
            accessType: "offline",
            prompt: "select_account consent",

        },
    },
baseURL: process.env.PROD_URL,
```

console.warn  it and mention the possible consequences of undefined behavior in the docs

.

### Additional context

_No response_

Bekacru:
- Display warning when 
- Add more tests for base url, disabled paths

closes #6780 

---
## Summary by cubic
Adds a warning when baseURL can’t be determined and hardens path/URL handling to prevent disabled path bypasses and unsafe proxy header usage. Improves reliability of redirects/callbacks and blocks common encoded path and traversal tricks.

- **New Features**
  - Logs an error when baseURL is missing, with guidance to configure baseURL or BETTER_AUTH_BASE_URL.

- **Bug Fixes**
  - Normalizes and double-decodes request paths to enforce disabledPaths, returning 404 for encoded slash and traversal variants.
  - Validates X-Forwarded-Host/Proto only when trustedProxyHeaders is true; accepts only http/https and safe host formats, otherwise falls back to the request URL.
  - Adds comprehensive tests for baseURL resolution, disabled paths, and proxy header edge cases.

<sup>Written for commit 1bde5775da1c59d2f8175dd64ad378fc9e370441. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/context/create-context.ts
+++ b/packages/better-auth/src/context/create-context.ts
@@ -104,12 +104,10 @@ export async function createAuthContext(
 	const logger = createLogger(options.logger);
 	const baseURL = getBaseURL(options.baseURL, options.basePath);
 
-	let displayError = false;
-	if (!baseURL && !displayError) {
+	if (!baseURL && !options.advanced?.trustedProxyHeaders) {
 		logger.error(
 			`[better-auth] Base URL could not be determined. Please set a valid base URL using the baseURL config option or the BETTER_AUTH_BASE_URL environment variable. Without this, callbacks and redirects may not work correctly.`,
 		);
-		displayError = true;
 	}
 
 	const secret =


packages/better-auth/src/context/create-context.ts
```diff
@@ -104,6 +104,12 @@ export async function createAuthContext(
 	const logger = createLogger(options.logger);
 	const baseURL = getBaseURL(options.baseURL, options.basePath);
 
+	let displayError = false;
```
himself65: I think this is unnecessary since it will only be called once per `betterAuth` call. And it won't work if there are multiple initial unless you put it in `globalThis`

Bekacru: done

nmcc1212:
Isn't the env var called `BETTER_AUTH_URL` not `BETTER_AUTH_BASE_URL` ?

A quick search of the codebase and this is the only mention of the `BETTER_AUTH_BASE_URL` variable https://github.com/search?q=repo%3Abetter-auth%2Fbetter-auth+BETTER_AUTH_BASE_URL&type=code

whereas the getBaseURL function uses `BETTER_AUTH_URL` https://github.com/better-auth/better-auth/blob/50990564441f870f1c91c910952f36f60d54ba1e/packages/better-auth/src/utils/url.ts#L104-L122 

am I missing something?

nmcc1212:
also, this warning is shown in contexts where it doesn't seem right to have it, such as when running
`npx @better-auth/cli@latest generate` the warning is shown

```
npx @better-auth/cli@latest generate

2025-12-27T15:24:46.018Z WARN [Better Auth]: [better-auth] Base URL could not be determined. Please set a valid base URL using the baseURL config option or the BETTER_AUTH_BASE_URL environment variable. Without this, callbacks and redirects may not work correctly.
Your schema is already up to date.
```

himself65:
> also, this warning is shown in contexts where it doesn't seem right to have it, such as when running `npx @better-auth/cli@latest generate` the warning is shown
> 
> ```
> npx @better-auth/cli@latest generate
> 
> 2025-12-27T15:24:46.018Z WARN [Better Auth]: [better-auth] Base URL could not be determined. Please set a valid base URL using the baseURL config option or the BETTER_AUTH_BASE_URL environment variable. Without this, callbacks and redirects may not work correctly.
> Your schema is already up to date.
> ```

this is expected. Please add it to the environment

nmcc1212:
> > also, this warning is shown in contexts where it doesn't seem right to have it, such as when running `npx @better-auth/cli@latest generate` the warning is shown
> 
> > 
> 
> > ```
> 
> > npx @better-auth/cli@latest generate
> 
> > 
> 
> > 2025-12-27T15:24:46.018Z WARN [Better Auth]: [better-auth] Base URL could not be determined. Please set a valid base URL using the baseURL config option or the BETTER_AUTH_BASE_URL environment variable. Without this, callbacks and redirects may not work correctly.
> 
> > Your schema is already up to date.
> 
> > ```
> 
> 
> 
> this is expected. Please add it to the environment 

When running the cli command to generate prisma schema, the environment variable to set the base url isn't needed at all.

Also, in my project (like pretty much any other) I store environment variables in a .env file which is loaded at runtime (both Next.JS and Vite behave like this) so the environment variable wouldn't be seen when running a command through the terminal/npx

PR #6794, opened 2025-12-16
fix(anonymous): missing path breaks anonymous hooks

ping-maxwell:
Recently, we supported `createAuthEndpoint` to not require a `path`, however we forgot to update a specific type which caused the anonmous plugin to have breaking code.

Any users who use the anonymous plugin and accessed an endpoint with not path (such as `auth.api.setPassword`) would run into an error.

---
## Summary by cubic
Fixes anonymous plugin errors when an endpoint has no path. Anonymous hooks now work with pathless endpoints like auth.api.setPassword, and hook matcher failures are handled safely.

- **Bug Fixes**
  - Made HookEndpointContext.path optional to match createAuthEndpoint.
  - Guarded matcher checks with ctx.path?.startsWith(...) and a final || false to avoid runtime errors.
  - Wrapped hook matcher in try/catch, log with hook source, and return a generic API error to avoid leaking internal details.

<sup>Written for commit d2f5017521c6a78acdfcd9488b82a5afef15eed6. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/to-auth-endpoints.ts
+++ b/packages/better-auth/src/api/to-auth-endpoints.ts
@@ -222,7 +222,7 @@ async function runBeforeHooks(
 				error,
 			);
 			throw new APIError("INTERNAL_SERVER_ERROR", {
-				message: `An error occured during hook matcher execution. Check the logs for more details.`,
+				message: `An error occurred during hook matcher execution. Check the logs for more details.`,
 			});
 		}
 		if (matched) {


packages/better-auth/src/plugins/anonymous/index.ts
```diff
@@ -132,15 +132,16 @@ export const anonymous = (options?: AnonymousOptions | undefined) => {
...
-							ctx.path.startsWith("/sign-in") ||
-							ctx.path.startsWith("/sign-up") ||
-							ctx.path.startsWith("/callback") ||
-							ctx.path.startsWith("/oauth2/callback") ||
-							ctx.path.startsWith("/magic-link/verify") ||
-							ctx.path.startsWith("/email-otp/verify-email") ||
-							ctx.path.startsWith("/one-tap/callback") ||
-							ctx.path.startsWith("/passkey/verify-authentication") ||
-							ctx.path.startsWith("/phone-number/verify")
+							ctx.path?.startsWith("/sign-in") ||
```
himself65, line 137: Do we have something to declare a endpoint to be a group of something like this?

packages/core/src/types/plugin.ts
```diff
@@ -20,7 +20,7 @@ type DeepPartial<T> = T extends Function
 export type HookEndpointContext = Partial<
 	EndpointContext<string, any> & Omit<InputContext<string, any>, "method">
 > & {
-	path: string;
+	path?: string;
```
himself65, line 23: Is this optional indeed? I thought `path` is required

ping-maxwell: Not anymore, recently we made `createAuthEndpoint` not require a path

Eg:
https://github.com/better-auth/better-auth/blob/63b961a5a919e03fdc88c35a24f235996ae52873/packages/better-auth/src/api/routes/update-user.ts#L315C1-L328C1

Ran into this error while on my test app, so I know indeed that `path` can be undefined

packages/better-auth/src/plugins/anonymous/index.ts
ping-maxwell: not sure what you mean

PR #6805, opened 2025-12-16
feat(saml): add XML parser hardening with configurable size limits

Paola3stefania:
## Summary by cubic
Hardened SAML handling by adding size limits and disabling XML entity processing. This blocks oversized payloads and XXE-style attacks, with configurable limits.

- **Security Hardening**
  - Reject SAML responses larger than 256KB (ACS and callback).
  - Reject IdP metadata larger than 100KB during provider setup.
  - Disable XML entity processing (processEntities: false) in the XML parser.

- **New Features**
  - Config options: saml.maxResponseSize (default 256KB) and saml.maxMetadataSize (default 100KB).
  - Exported default size limit constants for reuse.

<sup>Written for commit 9cb51a87a6a4d0149227681eb93423f994bcf489. Summary will update automatically on new commits.</sup>


packages/sso/src/routes/sso.ts
```diff
@@ -40,6 +40,9 @@ const AUTHN_REQUEST_KEY_PREFIX = "saml-authn-request:";
 /** Default clock skew tolerance: 5 minutes */
 export const DEFAULT_CLOCK_SKEW_MS = 5 * 60 * 1000;
 
+export const DEFAULT_MAX_SAML_RESPONSE_SIZE = 256 * 1024;
+export const DEFAULT_MAX_SAML_METADATA_SIZE = 100 * 1024;
+
```
Paola3stefania: after https://github.com/better-auth/better-auth/pull/6719 is merged, those constants belong to the new constant file

packages/sso/src/routes/sso.ts
```diff
@@ -628,6 +631,17 @@ export const registerSSOProvider = <O extends SSOOptions>(options: O) => {
 					message: "Invalid issuer. Must be a valid URL",
 				});
 			}
+
+			if (body.samlConfig?.idpMetadata?.metadata) {
+				const maxMetadataSize =
+					options?.saml?.maxMetadataSize ?? DEFAULT_MAX_SAML_METADATA_SIZE;
+				if (body.samlConfig.idpMetadata.metadata.length > maxMetadataSize) {
+					throw new APIError("BAD_REQUEST", {
+						message: `IdP metadata exceeds maximum allowed size (${maxMetadataSize} bytes)`,
```
himself65, line 680: This is not comparing bytes size, use byteLength with buffer or something

Paola3stefania: Thank you for this... I followed the pattern used in the cookie chunking logic (cookies/index.ts:208, session-store.ts:294) which uses .length against byte limits - but after investigating now I realized that those work because cookie data is base64-encoded (pure ASCII), where .length equals byte length. 

SAML metadata is UTF-8, so that assumption doesn't hold. Will switch to Buffer.byteLength(). TY

packages/sso/src/routes/sso.ts
```diff
@@ -1679,6 +1695,15 @@ export const callbackSSOSAML = (options?: SSOOptions) => {
 		async (ctx) => {
 			const { SAMLResponse, RelayState } = ctx.body;
 			const { providerId } = ctx.params;
+
+			const maxResponseSize =
+				options?.saml?.maxResponseSize ?? DEFAULT_MAX_SAML_RESPONSE_SIZE;
+			if (Buffer.byteLength(SAMLResponse, "utf8") > maxResponseSize) {
```
himself65: Still, Buffer is only available in node.js

Paola3stefania: addresed, added text encoder for consistency

PR #6825, opened 2025-12-17
chore: remove unnecessary custom word

bytaesu:
## Summary by cubic
Removed the hardcoded WorkOS user id from cspell custom words and added a local cspell ignore in the migration guide. This keeps the dictionary lean and prevents spellcheck noise around the example id.

<sup>Written for commit 79ce197ca553f3692aa03acab48c1f61077421c4. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/content/docs/guides/workos-migration-guide.mdx
+++ b/docs/content/docs/guides/workos-migration-guide.mdx
@@ -123,6 +123,7 @@ export const auth = betterAuth({
 
 ### Additional Fields
 
+{/* cspell:disable-next-line */}
 You probably used metadata in WorkOS. To preserve that metadata and the user id from WorkOS (e.g., user_01KBT4BMFF7ASGRDD0WZ6W63FF), extend the `user` schema as shown below. Better Auth provides a more flexible way to store user data. For more information, see [here](/docs/concepts/database#extending-core-schema).
 
 ```ts title="auth.ts"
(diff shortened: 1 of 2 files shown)


docs/content/docs/guides/workos-migration-guide.mdx
```diff
@@ -123,6 +123,7 @@ export const auth = betterAuth({
 
 ### Additional Fields
 
+{/* cspell:disable-next-line */}
 You probably used metadata in WorkOS. To preserve that metadata and the user id from WorkOS (e.g., user_01KBT4BMFF7ASGRDD0WZ6W63FF), extend the `user` schema as shown below. Better Auth provides a more flexible way to store user data. For more information, see [here](/docs/concepts/database#extending-core-schema).
```
jonathansamines, line 127: If user_randomletters is a common identifiable pattern, adding a new regex pattern to identify it may also work and could work better long term instead of adding manual exclusions.

See https://github.com/better-auth/better-auth/blob/canary/.cspell.jsonc#L58

bytaesu: This is an example showing WorkOS id format that we don’t use, so it’s unlikely to be reused!
Thanks for explaining 😁

jonathansamines: Got it 🫡

PR #6836, opened 2025-12-17
feat(saml): reject SAML responses containing multiple assertions

Paola3stefania:
## Summary by cubic
Enforces single-assertion SAML responses and fails fast when a response has zero or multiple assertions to block XSW-style injection. Validation runs before parsing in both SAML callback paths and supports plain and encrypted assertions.

- **New Features**
  - Added validateSingleAssertion and countAssertions with clear API errors (SAML_NO_ASSERTION, SAML_MULTIPLE_ASSERTIONS, SAML_INVALID_ENCODING, SAML_INVALID_XML).
  - Applied validation in callbackSSOSAML and acsEndpoint before parsing; ACS redirects with error=multiple_assertions or error=no_assertion and include error_description; exported helpers from saml/index.ts.

- **Refactors**
  - Centralized XML parsing and helpers (xmlParser, findNode, countAllNodes) in saml/parser.ts; hardened parser with processEntities: false and reused in algorithms.ts.
  - Added tests covering single/multiple/none assertions, namespaces, encrypted assertions, XSW patterns, and ACS/callback integration paths.

<sup>Written for commit a5e20a39a1305107eb51707889d9bfd260c15d76. Summary will update on new commits.</sup>

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1456,6 +1456,9 @@ importers:
 
   packages/sso:
     dependencies:
+      '@better-auth/utils':
+        specifier: 'catalog:'
+        version: 0.3.0
       '@better-fetch/fetch':
         specifier: 'catalog:'
         version: 1.1.21
@@ -1472,9 +1475,6 @@ importers:
         specifier: ^4.1.12
         version: 4.1.13
     devDependencies:
-      '@better-auth/utils':
-        specifier: 'catalog:'
-        version: 0.3.0
       '@types/body-parser':
         specifier: ^1.19.6
         version: 1.19.6


packages/sso/src/saml/assertions.ts
```diff
@@ -0,0 +1,61 @@
...
+		assertions,
+		encryptedAssertions,
+		total: assertions + encryptedAssertions,
+	};
+}
+
+export function validateSingleAssertion(samlResponse: string): void {
+	let xml: string;
+	try {
+		xml = Buffer.from(samlResponse, "base64").toString("utf-8");
```
himself65: no buffer because it's Node.js only feature, Please use the class from our utils

Paola3stefania: Ah its great that we alredy have a class with this. Missed that, ty

packages/sso/package.json
```diff
@@ -79,6 +79,7 @@
     "tsdown": "catalog:"
   },
   "peerDependencies": {
+    "@better-auth/utils": "catalog:",
```
himself65, line 82: We should move this into a different PR; otherwise, it's hard to cherry-pick. I let me do this

himself65: Maybe no

PR #6837, opened 2025-12-17
fix(oidc): add support for public clients with no client secret

marcoskichel:
## Summary

Fixes a few related bugs in the OIDC provider that prevented public OAuth clients (like ChatGPT) from authenticating using PKCE without client secrets.

## Problem

The OIDC provider had several issues:

1. **Public client ID token signing failure**: Attempted to sign ID tokens with undefined `clientSecret` for public clients, causing crashes
2.  **Foreign key constraint for trusted clients**: Attempted to create database records for trusted clients that don't exist in the database

## Solution

1. **Fixed ID token signing**: Use server secret for public clients when `clientSecret` is unavailable
2. **Fixed trusted clients**: Skip database record creation for trusted clients that are configured in memory

---
## Summary by cubic
Fixes OIDC to fully support public OAuth clients using PKCE. Authorization codes are invalidated on first use, public client ID tokens use the server secret when no client secret is present, and access tokens aren’t stored for trusted clients.

- **Bug Fixes**
  - Delete verification value immediately to invalidate the code after any use attempt.
  - Sign public client ID tokens with the server secret when clientSecret is missing.
  - Skip access token DB records for trusted clients to avoid foreign key issues.

<sup>Written for commit 2a133526bdbf36c1339b76107cc1db83cbf82a40. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/index.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/index.ts
@@ -809,9 +809,14 @@ export const oidcProvider = (options: OIDCOptions) => {
 						});
 					}
 
+					const value = JSON.parse(
+						verificationValue.value,
+					) as CodeVerificationValue;
+
 					await ctx.context.internalAdapter.deleteVerificationValue(
 						verificationValue.id,
 					);
+
 					if (!client_id) {
 						throw new APIError("UNAUTHORIZED", {
 							error_description: "client_id is required",
@@ -852,9 +857,6 @@ export const oidcProvider = (options: OIDCOptions) => {
 						});
 					}
 
-					const value = JSON.parse(
-						verificationValue.value,
-					) as CodeVerificationValue;
 					if (value.clientId !== client_id.toString()) {
 						throw new APIError("UNAUTHORIZED", {
 							error_description: "invalid client_id",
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/plugins/oidc-provider/index.ts
```diff
@@ -918,25 +920,31 @@ export const oidcProvider = (options: OIDCOptions) => {
...
-					});
+
+					// Check if this is a trusted client (not in database)
+					const isTrustedClient = trustedClients.some(
+						(tc) => tc.clientId === client_id.toString(),
+					);
+
+					// Only create database record for non-trusted clients
+					if (!isTrustedClient) {
+						await ctx.context.adapter.create({
```
marcoskichel, line 933: Note to self: can't do that, the proper fix is one of:

- Keep trusted clients in sync with the database
- Remove the `clientId` fk constraint and validate it in runtime against the clients in the database and the runtime ones defined in the config file (probably the best approach)

dvanmali:
Hi all, we released the new [OAuth Provider Plugin](https://www.better-auth.com/docs/plugins/oauth-provider) which supports public clients. Feel free to let us know how it works :)

PR #6839, opened 2025-12-17
chore(sso): clean up SAML test suite

Paola3stefania:
## Summary by cubic
Cleaned up the SAML SSO test suite and tightened SSO provider types. This removes type errors and stabilizes the tests.

- **Refactors**
  - Added SSOProviderResponse and narrowed SSOProviderReturn to return parsed oidcConfig/samlConfig with explicit fields.
  - Refactored registerSSOProvider to build a typed result before returning via ctx.json.
  - Simplified customFetchImpl usage in tests for consistency.

<sup>Written for commit 224f2c6f0b1e3b390a3c65fb33560fe6d9b93dad. Summary will update automatically on new commits.</sup>


packages/sso/src/saml.test.ts
```diff
@@ -610,7 +620,10 @@ describe("SAML SSO", async () => {
 		baseURL: "http://localhost:3000",
 		plugins: [bearer(), ssoClient()],
 		fetchOptions: {
-			customFetchImpl: async (url, init) => {
+			customFetchImpl: async (
```
himself65: I think we don't need to declare types here? since `createAuthClient` should already do that

packages/sso/src/saml.test.ts
```diff
@@ -28,6 +28,16 @@ import {
...
 import { ssoClient } from "./client";
 import { DEFAULT_CLOCK_SKEW_MS } from "./constants";
+import type { OIDCConfig, SAMLConfig, SSOProvider } from "./types";
+
+type SSOProviderResponse = SSOProvider<{}> & {
+	samlConfig?: SAMLConfig;
+	oidcConfig?: OIDCConfig;
+};
+
+function asSSOProvider<T>(response: T): SSOProviderResponse {
```
himself65: why do we need this type in the test case? should we update the API response type?

Paola3stefania: Ahh yes !  I'll fix the API response type properly so this cast isn't needed in tests.

himself65 (approved):

PR #6841, opened 2025-12-17
chore(better-auth): add context to organization hooks

jonathansamines:
## Summary by cubic
Add endpoint context (ctx) to all organization hooks so custom logic can access request/session details. Updated types, routes, docs, and tests to pass ctx consistently.

- **New Features**
  - All organization hooks now receive ctx: GenericEndpointContext.
  - Routes for orgs, members, teams, and invites pass ctx to before/after hooks; docs and tests updated.

- **Migration**
  - Update your organizationHooks to use (data, ctx) for every hook.
  - If using TypeScript, add ctx: GenericEndpointContext to hook signatures.

<sup>Written for commit 5ae178be5a474938564b708639be1b62eeb28d7c. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/organization/types.ts
+++ b/packages/better-auth/src/plugins/organization/types.ts
@@ -409,7 +409,7 @@ export interface OrganizationOptions {
 				 *
 				 * @example
 				 * ```ts
-				 * beforeCreateOrganization: async (data) => {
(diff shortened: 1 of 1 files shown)


himself65 (changes requested):
We use `getCurrentAuthContext` for this case.

Issue #6848, opened 2025-12-18
accountLinking is NOT disabled by default


mhamid3d:
### To Reproduce

1. Do NOT add any `accountLinking` option in your auth instance
2. Perform a social sign in on the client (not OneTap, as that plugin handles it properly, explained more below)
3. See that account linking happens automatically

### Current vs. Expected behavior

I expect the oauth callback to block the sign in request because `accountLinking.enabled` is mentioned to be `false` by default. 

### What version of Better Auth are you using?

1.4.6

### Additional context

`accountLinking.enabled` will return `undefined` if the user doesn't explicitly set it. In several areas throughout the codebase, `accountLinking.enabled` is checked like this `accountLinking?.enabled === false` which will return false. But in OneTap, they just check if enabled is truthy, which is why the OneTap plugin properly honors the default false setting.

Also I will note, the typehint `@default` for `accountLinking.enabled` is `true` but in the docs it says false, this is misleading and confusing.

bytaesu:
Hello @mhamid3d,

I'm looking into this

PR #6851, opened 2025-12-18
fix: disable account linking by default

bytaesu:
The documentation states `default: false`, but the internal logic was implemented as `default: true`. These should be aligned. I think account linking should only work when explicitly enabled, so I’m opening this PR.

Closes https://github.com/better-auth/better-auth/issues/6848

docs

---
## Summary by cubic
Disable account linking by default to match the docs and prevent unintended linking. Linking now only happens when explicitly enabled.

- **Bug Fixes**
  - Require account.accountLinking.enabled === true in account, callback, and OAuth linking flows.
  - Update option docs/types: default is now false.
  - Adjust tests to set enabled: true where needed; add coverage ensuring default-disabled behavior (returns account_not_linked).

- **Migration**
  - If you use account linking, set account.accountLinking.enabled: true in your config.

<sup>Written for commit 3c79387c98811c3c34902943a2c53bf1ce75d5ca. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/account.test.ts
+++ b/packages/better-auth/src/api/routes/account.test.ts
@@ -74,6 +74,7 @@ describe("account", async () => {
 		},
 		account: {
 			accountLinking: {
+				enabled: true,
 				allowDifferentEmails: true,
 			},
 			encryptOAuthTokens: true,
--- a/packages/better-auth/src/oauth2/link-account.test.ts
+++ b/packages/better-auth/src/oauth2/link-account.test.ts
@@ -486,3 +486,89 @@ describe("oauth2 - override user info on sign-in", async () => {
 		expect(session.data?.user.name).toBe("Updated Name");
 	});
 });
+
+describe("oauth2 - account linking disabled by default", async () => {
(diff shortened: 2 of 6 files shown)


bytaesu:
> I'm afraid this will be a regression, since people already use `accountLinking: true` in their production environment.
> 
> How about we change the document? /cc @Bekacru

Right. We should clarify the approach for this option:
whether this should be explicitly enabled or disabled.

Bekacru:
Yeah I think changing the docs make more sense

Bekacru:
account linking beind enabled by default is intended behavior if i recall properly

bytaesu:
I'll close this PR and update the documentation!

cc @Bekacru @himself65

mhamid3d:
@bytaesu If default is true then patterns like this aren't valid anymore as `undefined` would return `false`

https://github.com/better-auth/better-auth/blob/a345badd731b809f95989747f8c662c8871c7eb3/packages/better-auth/src/plugins/one-tap/index.ts#L150-L153

bytaesu:
Hello @mhamid3d,

Thanks for checking. I'll update the documentation and review the related logic.

PR #6857, opened 2025-12-18
fix(organization): allow internal organization creation when disabled for client

### LINKED ISSUE #6791: Cannot use `auth.api.createOrganization` in database hooks when `allowUserToCreateOrganization` is false

### To Reproduce

I am trying to implement a flow where a generic user cannot create organizations manually, but the system creates a "Personal Organization" for them automatically upon signup via a database hook.

When I set `allowUserToCreateOrganization: false` in the Organization plugin configuration, it prevents `auth.api.createOrganization` from being used within the `databaseHooks.user.create.after` hook.

Current Behavior:
If `allowUserToCreateOrganization` is false, and we call `auth.api.createOrganization` in `databaseHooks.user.create.after`, it fails with `You are not allowed to create a new organization`.

Expected Behavior:
`auth.api.createOrganization` called from the server-side (without user headers) should bypass the `allowUserToCreateOrganization: false` restriction.

1.4.7

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.1.0: Mon Oct 20 19:32:56 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T8132",
    "release": "25.1.0",
    "cpuCount": 10,
    "cpuModel": "Apple M4",
    "totalMemory": "16.00 GB",
    "freeMemory": "0.17 GB"
  },
  "node": {
    "version": "v24.12.0",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.4"
  },
  "frameworks": [
    {
      "name": "react",
      "version": "^19.2.3"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.16.3"
    },
 {
      "name": "drizzle",
      "version": "^0.39.3"
    }
  ],
  "betterAuth": {
    "version": "^1.4.7",
    "config": null
  }
}
```

Backend

```typescript
const auth = betterAuth({
	emailAndPassword: {
		enabled: true,
	},
	plugins: [
		organization({
			allowUserToCreateOrganization: false,
		}),
	],
	databaseHooks: {
		user: {
			create: {
				after: async (user, ctx) => {
					await auth.api.createOrganization({
						body: {
							name: `${user.name}'s Workspace`,
							slug: `${slugify(user.name)}-${randomStr(6)}`,
							logo: undefined,
							metadata: {
								personal: true
							},
							userId: user.id,
							keepCurrentActiveOrganization: false,
						},
					})
				}
			},
		}
	},
});
```

### Additional context

_No response_

GautamBytes:
This fixes an issue where using auth.api.createOrganization inside database hooks (like user.create.after) would fail if allowUserToCreateOrganization was set to false.

The Change I updated createOrganization to detect "system actions" (calls made without a session but with an explicit userId, which happens during internal API calls). These trusted internal calls now bypass the allowUserToCreateOrganization config check.

Fixes #6791

---
## Summary by cubic
Allow internal system calls to create organizations even when user org creation is disabled. Detects system actions (no session + explicit userId) in createOrganization and bypasses allowUserToCreateOrganization, fixing failures in hooks like user.create.after.

<sup>Written for commit 15a12fa5e8f0a5bda77343bc21bb4fbb7dba2fc8. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-org.test.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-org.test.ts
@@ -444,4 +444,32 @@ describe("organization hooks", async () => {
 		});
 		expect(afterCreateTeam).toHaveBeenCalled();
 	});
+
+	it("should allow internal organization creation when disabled for users", async () => {
+		const { auth } = await getTestInstance({
+			plugins: [
+				organization({
+					allowUserToCreateOrganization: false,
+				}),
+			],
+		});
+
+		const newUser = await auth.api.signUpEmail({
+			body: {
+				email: "internal@test.com",
+				password: "password",
+				name: "Internal User",
+			},
+		});
+
+		const internalOrg = await auth.api.createOrganization({
+			body: {
+				name: "Internal Org",
+				slug: "internal-org",
+				userId: newUser.user.id,
+			},
+		});
+		expect(internalOrg).toBeDefined();
+		expect(internalOrg?.name).toBe("Internal Org");
+	});
 });


himself65 (commented):
This is fine. But I think up an idea that we should improve the code that handle client side call / internal call

himself65 (changes requested):
Could you please update the test case? Just like what you mentioned in the description

PR #6867, opened 2025-12-18
fix: passkey - correct verification deletion when using useNumberId

tbveloso:
## Summary by cubic
Fixes incorrect cleanup of passkey verification challenges by deleting records by identifier, ensuring proper removal when useNumberId is enabled in both registration and authentication flows.

- **Bug Fixes**
  - Replace internalAdapter.deleteVerificationValue with deleteVerificationByIdentifier(challengeId) in verifyPasskeyRegistration and verifyPasskeyAuthentication to prevent stale verification entries.

<sup>Written for commit 95cc3065d314037e4220f9572cf5f3651087f919. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/passkey/src/routes.ts
+++ b/packages/passkey/src/routes.ts
@@ -506,7 +506,9 @@ export const verifyPasskeyRegistration = (options: RequiredPassKeyOptions) =>
 					model: "passkey",
 					data: newPasskey,
 				});
-				await ctx.context.internalAdapter.deleteVerificationValue(challengeId);
+				await ctx.context.internalAdapter.deleteVerificationByIdentifier(
+					challengeId,
+				);
 				return ctx.json(newPasskeyRes, {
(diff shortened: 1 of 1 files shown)


tbveloso:
I am configuring Passkey in a Next.js project and, possibly due to using the `useNumberId: true` option, the deletion of verification records (from the `verifications` table) was consistently failing, reporting that it was not possible to delete a record with an `id` of `NaN`.

After reviewing the code, I identified the root cause and changed the deleteVerificationValue function to the deleteVerificationByIdentifier function to query by the `credentialId` column instead. I am not sure whether this change has any impact when `useNumberId` is set to `false`.

In my case, everything is now working as expected, and as an optimization, I also created an index on the `credentialId` column.

himself65 (changes requested):
Could you please update the test cases?

PR #6868, opened 2025-12-18
fix(sso): enforce domain verification in assignOrganizationByDomain

Paola3stefania:
## Summary by cubic
Enables optional domain verification in assignOrganizationByDomain so users are auto-assigned only when the SSO provider’s domain is verified and verification is enabled. Passes domainVerification from sso() options and adds tests for verified/unverified domains, disabled verification, missing org links, existing memberships, and conflicting providers.

- **Bug Fixes**
  - SSO provider lookup now requires a domain match and domainVerified: true when domain verification is enabled.

<sup>Written for commit debc2e0e9fd65b077ce86e1694a660f820915eb6. Summary will update automatically on new commits.</sup>


packages/sso/src/linking/org-assignment.ts
```diff
@@ -119,7 +119,10 @@ export async function assignOrganizationByDomain(
 		SSOProvider<SSOOptions>
 	>({
 		model: "ssoProvider",
-		where: [{ field: "domain", value: domain }],
+		where: [
+			{ field: "domain", value: domain },
+			{ field: "domainVerified", value: true },
+		],
```
Bekacru: but unless domain verification is enabled specifically this would fail as `domainVerified` field wouldn't exist in the db... we should check if the sso provider has domain verified value and it's set to true instead

Paola3stefania: Ah yes ! Adding the check. TY

Issue #6799, opened 2025-12-16
Session refresh writes on GET request


mnlfischer:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

I have the problem that I am using the LiteFS proxy from Fly.io, and this proxy cannot perform write operations during a GET request. When my session has expired, better-auth performs an update. However, if there are multiple databases, the GET request may be routed to a replica database instance, and writes on a replica are forbidden. The proxy only forwards POST requests to the primary database, since only the primary is allowed to perform writes.

Is it possible to send the session GET request as a POST request as an optional configuration?

This issue is not specific to Fly. It also impacts database replica configurations that rely on correct HTTP methods to ensure requests are routed to the correct primary or replica database.

Fly Reference:
For most web applications, you can take advantage of load balancer configuration to route writes to the primary node, assuming your application follows the convention of avoiding write operations (INSERT, UPDATE, etc.) on GET requests.

I disabled session refresh for now.

### Current vs. Expected behavior

getSession refreshes the session in the database via GET Request , I expect writing operations with the corret http method.

### What version of Better Auth are you using?

1.4.6

### System info

```bash
no outpud for whatever reason
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
export const auth = betterAuth({
  database: prismaAdapter(prisma, {
    provider: "sqlite",
  }),
  user: {
    additionalFields: {
      first_name: {
        type: "string",
        input: false,
      },
      last_name: {
        type: "string",
        input: false,
      },
    },
  },
  advanced: {
    database: {
      generateId: () => cuid(),
    },
  },
  session: {
    disableSessionRefresh: true,
    expiresIn: 60 * 60 * 24 * 7,
  },
  telemetry: {
    enabled: false,
  },
  plugins: [
    magicLink({...
      },
    }),
    admin(),
  ],
});
```

PR #6871, opened 2025-12-19
feat(session): add `deferSessionRefresh` option to support read-replica setups

Paola3stefania:
Needs https://github.com/better-auth/better-auth/pull/6891 to be merged before it can be reviewed

---
## Summary by cubic
Adds a deferSessionRefresh option to defer session refresh writes to POST so GET is read-only, enabling read‑replica setups. Off by default.

- **New Features**
  - Added session.deferSessionRefresh?: boolean to BetterAuthOptions (default false). When enabled, GET is read-only and refresh occurs on POST.
  - /get-session now supports POST (only when deferSessionRefresh is enabled). GET performs no writes, includes needsRefresh (respects disableSessionRefresh), and does not delete expired sessions; deletion/refresh happen on POST. OpenAPI exposes both GET and POST.
  - Client refresh manager detects needsRefresh and posts to /get-session when needed; otherwise no POST.

<sup>Written for commit 15827656e5a41ffcc6a65f57a67d75f58144a25b. Summary will update on new commits.</sup>

Fixes https://github.com/better-auth/better-auth/issues/6799


bun.lock
himself65, line 1: Remove this

Paola3stefania: Oh didnt saw this coming through. TY

Issue #6798, opened 2025-12-16
trustedOrigins disregarded in v1.4.7


giuliog-GR:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Upgrade from v1.4.4 to v1.4.7 
2. Better Auth breaks due to invalid origin
3. Downgrade to v1.4.4 everything works again

### Current vs. Expected behavior

I would expect the system to continue working instead it is currently throwing the following error:

`2025-12-16T12:02:08.788Z ERROR [Better Auth]: Invalid origin: http://localhost:5173`

I did not have any trustedOrigin specified, I have tried to add a list of trusted origins but that does not seem to have any effect. Authentication is effectively broken. 

### What version of Better Auth are you using?

1.4.7

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#101-Ubuntu SMP Tue Nov 14 13:30:08 UTC 2023",
    "release": "5.15.0-91-generic",
    "cpuCount": 32,
    "cpuModel": "AMD EPYC 73F3 16-Core Processor",
    "totalMemory": "1003.36 GB",
    "freeMemory": "896.62 GB"
  },
  "node": {
    "version": "v22.21.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.2"
  },
  "frameworks": [
    {
      "name": "svelte",
      "version": "^5.45.10"
    },
    {
      "name": "@sveltejs/kit",
      "version": "^2.49.2"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.16.3"
    },
    {
      "name": "postgres",
      "version": "^3.4.7"
    },
    {
      "name": "drizzle",
      "version": "^0.45.1"
    }
  ],
  "betterAuth": {
    "version": "1.4.7",
    "config": {
      "trustedOrigins": [
        "https://REDACTED",
        "https://REDACTED",
        "http://localhost:5173"
      ],
      "session": {
        "expiresIn": 86400
      },
      "user": {
        "additionalFields": {
          "groups": {
            "type": "string[]",
            "required": true,
            "defaultValue": [],
            "input": false
          }
        }
      },
      "plugins": [
        {
          "name": "sveltekit-cookies",
          "config": {
            "id": "sveltekit-cookies",
            "hooks": {
              "after": [
                {}
              ]
            }
          }
        },
        {
          "name": "generic-oauth",
          "config": {
[...]

PR #6882, opened 2025-12-19
fix(auth): respect trustedOrigins when baseURL is inferred

Paola3stefania:
## Summary by cubic
Fixes origin checks to honor trustedOrigins when baseURL is inferred from the first request, preventing false “Invalid origin” errors. Config arrays and BETTER_AUTH_TRUSTED_ORIGINS are now respected even without a configured baseURL.

- **Bug Fixes**
  - Rebuilds trustedOrigins after inferring baseURL, adding the inferred origin, config array entries, and BETTER_AUTH_TRUSTED_ORIGINS.
  - Keeps function-based trustedOrigins working at request time.
  - Adds regression tests covering trusted config/env origins, rejection of untrusted origins, and acceptance of the inferred baseURL origin.

- **Docs**
  - Corrects env var name to BETTER_AUTH_URL.
  - Adds a warning recommending explicit baseURL configuration for security and stability.

<sup>Written for commit 77401926927b2fc902cafda5de889813b9cbbbf5. Summary will update automatically on new commits.</sup>

fixes https://github.com/better-auth/better-auth/issues/6798

### DIFF
--- a/packages/better-auth/src/auth/base.ts
+++ b/packages/better-auth/src/auth/base.ts
@@ -1,8 +1,8 @@
 import type { AuthContext, BetterAuthOptions } from "@better-auth/core";
 import { runWithAdapter } from "@better-auth/core/context";
-import { env } from "@better-auth/core/env";
 import { BASE_ERROR_CODES, BetterAuthError } from "@better-auth/core/error";
 import { getEndpoints, router } from "../api";
+import { getTrustedOrigins } from "../context/helpers";
 import type { Auth } from "../types";
 import { getBaseURL, getOrigin } from "../utils/url";
 
@@ -27,8 +27,7 @@ export const createBetterAuth = <Options extends BetterAuthOptions>(
 		handler: async (request: Request) => {
 			const ctx = await authContext;
 			const basePath = ctx.options.basePath || "/api/auth";
-			const baseURLWasInferred = !ctx.options.baseURL;
-			if (baseURLWasInferred) {
+			if (!ctx.options.baseURL) {
 				const baseURL = getBaseURL(
 					undefined,
 					basePath,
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/auth/base.ts
```diff
@@ -26,7 +27,8 @@ export const createBetterAuth = <Options extends BetterAuthOptions>(
 		handler: async (request: Request) => {
 			const ctx = await authContext;
 			const basePath = ctx.options.basePath || "/api/auth";
-			if (!ctx.options.baseURL) {
+			const baseURLWasInferred = !ctx.options.baseURL;
```
Bekacru: no need to define a new constant, lets just use ctx.options.baseURL makes it clearer

packages/better-auth/src/auth/base.ts
```diff
@@ -44,6 +46,27 @@ export const createBetterAuth = <Options extends BetterAuthOptions>(
...
+				// Add array-based trustedOrigins from config
+				if (options.trustedOrigins && Array.isArray(options.trustedOrigins)) {
+					ctx.trustedOrigins.push(...options.trustedOrigins);
+				}
+				// Add env-based trustedOrigins
+				const envTrustedOrigins = env.BETTER_AUTH_TRUSTED_ORIGINS;
+				if (envTrustedOrigins) {
+					ctx.trustedOrigins.push(...envTrustedOrigins.split(","));
+				}
+			}
```
Bekacru: why dont we use the helper function https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/context/helpers.ts#L64

Paola3stefania: Ahh this way looks cleaner. missed that one, txss!

PR #6887, opened 2025-12-19
fix: trusted origins resolving

Bekacru:
## Summary by cubic
Fixes trusted origin resolution so redirect and origin checks use the correct list on every request. Origins from baseURL, env, and per-request functions are now merged in one place.

- **Bug Fixes**
  - Centralized origin resolution in async getTrustedOrigins(options, request) and use it in context and request handler.
  - Support options.trustedOrigins as an array or per-request function; merge with BETTER_AUTH_TRUSTED_ORIGINS.
  - Handle missing baseURL gracefully; still resolve other origins and log a warning instead of an error.

<sup>Written for commit cdebe4207440b2e36429b1c019892fc8acd08842. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/context/create-context.ts
+++ b/packages/better-auth/src/context/create-context.ts
@@ -104,16 +104,10 @@ export async function createAuthContext(
 	const logger = createLogger(options.logger);
 	const baseURL = getBaseURL(options.baseURL, options.basePath);
 
-	let showBaseURLWarning = false;
-	if (
-		!baseURL &&
-		!options.advanced?.trustedProxyHeaders &&
-		!showBaseURLWarning
-	) {
-		logger.error(
+	if (!baseURL) {
+		logger.warn(
 			`[better-auth] Base URL could not be determined. Please set a valid base URL using the baseURL config option or the BETTER_AUTH_BASE_URL environment variable. Without this, callbacks and redirects may not work correctly.`,
 		);
-		showBaseURLWarning = true;
 	}
 
 	const secret =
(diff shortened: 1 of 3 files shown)


packages/better-auth/src/context/create-context.ts
```diff
@@ -188,8 +182,13 @@ export async function createAuthContext(
...
-		trustedOrigins: getTrustedOrigins(options),
-		isTrustedOrigin(url: string, settings?: { allowRelativePaths: boolean }) {
+		trustedOrigins: await getTrustedOrigins(options),
+		isTrustedOrigin(
+			url: string,
+			settings?: {
+				allowRelativePaths: boolean;
+			},
+		) {
 			return ctx.trustedOrigins.some((origin) =>
```
Paola3stefania, line 192: I think Copilot its quire right about soemthing

mutating the shared `ctx` object can cause race conditions when multiple requests arrive simultaneously with different dynamic trusted origins.

creating a new object won't help because the method still reads from the original ctx closure.

**Scenario:**

Request A arrives from https://app-a.com
- trustedOrigins is a function that returns ["https://app-a.com"]
- Starts await getTrustedOrigins() → returns ["https://app-a.com"]

Request B arrives from https://app-b.com (while A is still processing)
- trustedOrigins function returns ["https://app-b.com"]
 - **Sets ctx.trustedOrigins = ["https://app-b.com"]**  **overwrites**

Request A continues
- Now ctx.trustedOrigins is ["https://app-b.com"]
- Origin check fails: https://app-a.com is not in ["https://app-b.com"]
- Request A gets rejected incorrectly

This happens because both requests share the same ctx object and overwrite each other's trustedOrigins.

WDYT? Or Im mising something here

Issue #5473, opened 2025-10-22
[plugins][organization]: Wrong migrations order when dynamicAccessControl is enabled


brenodanyel:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. enable `dynamicAccessControl` from the organization plugin
2. try to run the migrations `npx @better-auth/cli migrate`

### Current vs. Expected behavior

the migrate command attempts to run the migrations that create the **organizationRole** table before the **organization** table, which naturally causes the migration to fail, since **organizationRole** references **organization**.

### What version of Better Auth are you using?

1.3.28

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#202412060638~1748542656~22.04~663e4dc SMP PREEMPT_DYNAMIC Thu M",
    "release": "6.12.10-76061203-generic",
    "cpuCount": 16,
    "cpuModel": "AMD Ryzen 7 5700G with Radeon Graphics",
    "totalMemory": "15.41 GB",
    "freeMemory": "5.83 GB"
  },
  "node": {
    "version": "v22.19.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.3"
  },
  "frameworks": [
    {
      "name": "hono",
      "version": "^4.10.1"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.15.6"
    },
    {
      "name": "kysely",
      "version": "^0.28.2"
    }
  ],
  "betterAuth": {
    "version": "^1.3.28",
    "config": {
      "database": {
        "dialect": {},
        "type": "postgres"
      },
      "emailAndPassword": {
        "enabled": true
      },
      "socialProviders": {
        "google": {
          "clientId": "[REDACTED]",
          "clientSecret": "[REDACTED]",
          "prompt": "select_account"
        }
      },
      "baseURL": "http://localhost:3000",
      "trustedOrigins": [
        "http://localhost:5173"
      ],
      "plugins": [
        {
          "name": "admin",
          "config": {
            "id": "admin",
            "hooks": {
              "after": [
                {}
              ]
            },
            "endpoints": {},
            "$ERROR_CODES": {
              "FAILED_TO_CREATE_USER": "Failed to create user",
              "USER_ALREADY_EXISTS": "User already exists.",
              "USER_ALREADY_EXISTS_USE_ANOTHER_EMAIL": "User already exists. Use another email.",
[...]

PR #6911, opened 2025-12-20
Fix migration ordering when organization plugin is used with teams en…

### LINKED ISSUE #6832: [organization] Migration fails: team table created before organization table
Description:

When using the organization plugin with teams enabled, the migrate command fails because it attempts to create the team table (which has a foreign key to organization) before creating the organization table.

  Versions:
  - better-auth: 1.4.7
  - @better-auth/cli: 1.4.7
  - Database: PostgreSQL 16

  Configuration:
  import { organization } from "better-auth/plugins";

    database: new Pool({ connectionString: process.env.DATABASE_URL }),
    plugins: [
      organization({
        teams: { enabled: true }
      })
    ],
  });

  Steps to reproduce:

  1. Configure better-auth with the organization plugin (teams enabled)
  2. Run npx @better-auth/cli migrate -y against an empty PostgreSQL database

  Expected behavior:

  Tables created in dependency order:
  1. organization table (no dependencies)
  2. team table (depends on organization)

  Actual behavior:

  The CLI outputs migration plan showing team before organization:
  🔑 The migration will affect the following:
  -> name, organizationId, createdAt, updatedAt fields on team table.
  -> name, slug, logo, createdAt, metadata fields on organization table.

  Then fails with:
  error: relation "organization" does not exist

  Workaround:

  Use better-auth generate to get schema, create organization first, then run migrate:

  npx @better-auth/cli generate -y --output /tmp/schema.sql
  grep 'create table "organization"' /tmp/schema.sql | psql -d mydb
  npx @better-auth/cli migrate -y  # now succeeds

  Related:
  - #5473 fixed similar ordering for dynamicAccessControl, but base organization plugin still has this issue

Adityakk9031:
This PR fixes the issue reported in:

👉 [organization] Migration fails: team table created before organization table #6832

When using the organization plugin with teams: { enabled: true }, the migration command fails on a fresh database.

What was happening

team table has a foreign key reference to organization

Migration engine was creating team table first

PostgreSQL throws error:

relation "organization" does not exist

Root cause

Plugin tables do not have explicit migration order

Both organization and team tables default to order: Infinity

Migration relied on array order, which is not dependency-aware

Result: foreign key table created before parent table

What this PR does
✅ 1. Adds dependency-aware table sorting

Introduces a new utility:

sortTablesByDependencies()

This function:

Builds dependency graph from fields[].references.model

Ensures parent tables are created before dependent tables

Safely handles missing tables and circular references

✅ 2. Adds unit test (no DB required)

New test verifies correct ordering:

organization → team

This test:

Reproduces the exact failure scenario from #6832

Runs without PostgreSQL

Prevents regression in future plugin changes

Files added / changed

packages/better-auth/src/utils/sort-tables-by-dependencies.ts

packages/better-auth/src/utils/index.ts

packages/better-auth/src/migrations/getmigration.test.ts

Result

organization table is always created before team

npx @better-auth/cli migrate works on empty databases

Fix applies to all plugins with foreign key dependencies

Related Issues

Fixes: #6832

Related: #5473 (dynamicAccessControl ordering issue)

---
## Summary by cubic
Fixes #6832. Ensures migrations create parent tables before dependents, so organization is created before team when the organization plugin runs with teams enabled.

- **Bug Fixes**
  - Added dependency-aware table sorting in get-migration (topological sort that respects numeric order; warns and falls back if cycles).
  - Introduced sortTablesByDependencies utility and a unit test verifying organization → team ordering.

<sup>Written for commit 733429b8b8c79dbc6714dfa7ecce173a92d35cc9. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/db/get-migration.ts
+++ b/packages/better-auth/src/db/get-migration.ts
@@ -93,14 +93,109 @@ export function matchType(
 }
 
 /**
+ * Topological sort helper:
+ * - Inspects table.fields for attr.references.model (or attr.references.table)
+ * - Respects explicit numeric `order` values (missing => Infinity)
+ * - Throws an error describing remaining tables when a cycle is detected
+ */
+
+type TableMeta = {
+	table: string;
+	fields: Record<string, any>;
+	order?: number;
+};
+
+function getFKsFromFields(fields: Record<string, any>): string[] {
+	const fks = new Set<string>();
+	for (const f in fields) {
+		const attr = fields[f];
+		if (!attr) continue;
+		// support .references.model or .references.table
+		const ref = attr.references?.model || attr.references?.table;
+		if (typeof ref === "string") fks.add(ref);
+	}
+	return Array.from(fks);
+}
+
+function sortTablesByDependencies(tables: TableMeta[]) {
+	const nameMap = new Map<string, TableMeta>();
+	for (const t of tables) nameMap.set(t.table, t);
+
+	const adj = new Map<string, Set<string>>();
+	const indegree = new Map<string, number>();
+	for (const t of tables) {
+		indegree.set(t.table, 0);
+		adj.set(t.table, new Set());
+	}
+
+	for (const t of tables) {
+		const fks = getFKsFromFields(t.fields);
+		for (const ref of fks) {
+			if (!nameMap.has(ref)) {
+				// reference to an external table (e.g., core table); ignore for ordering
+				continue;
+			}
+			// edge: ref -> t.table (ref must be created before t.table)
+			adj.get(ref)!.add(t.table);
+			indegree.set(t.table, (indegree.get(t.table) ?? 0) + 1);
+		}
+	}
+
+	const orderVal = (name: string) => {
+		const o = nameMap.get(name)?.order;
+		return typeof o === "number" ? o : Number.POSITIVE_INFINITY;
+	};
+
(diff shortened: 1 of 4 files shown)


packages/better-auth/src/db/get-migration.ts
```diff
@@ -92,15 +92,110 @@ export function matchType(
...
+/**
+ * Topological sort helper:
+ * - Inspects table.fields for attr.references.model (or attr.references.table)
+ * - Respects explicit numeric `order` values (missing => Infinity)
+ * - Throws an error describing remaining tables when a cycle is detected
+ */
+
+type TableMeta = {
+	table: string;
+	fields: Record<string, any>;
```
himself65, line 104: Let’s dont use any type

packages/better-auth/src/db/get-migration.ts
```diff
@@ -360,7 +475,7 @@ export async function getMigrations(config: BetterAuthOptions) {
 				mssql: useNumberId
 					? "integer"
 					: useUUIDs
-						? "varchar(36)" /* Should be using `UNIQUEIDENTIFIER` but Kysely doesn't support it */
+						? "varchar(36)"
```
himself65, line 478: Leave the comment here

himself65:
I think topography sort is great, but this is vibe coded PR, obviously.

PR #6914, opened 2025-12-20
fix(adapter): apply `customTransformInput` to where clause values

### LINKED ISSUE #6912: Where clause transformed without input mapping

### To Reproduce

1. Use a custom database adapter
2. Set `supportsDates: false`
3. Use custom transform methods to map dates to/from something custom, like a number
4. Use that field in a where clause

Those repro steps aren't super helpful, but the concept is you now have transformed data in the database, and you're using non-transformed values to query. So in this case my date fields in the database are numbers, and I'm querying using strings.

1.4.7

### System info

```bash
This isn't setup specific.
```

Backend

```typescript

```

### Additional context

This is a regression from when adapter configuration like `supportsDates` started being applied to where clauses, but other relevant adapter config like transformations were not. There may be other effects to this, but currently database operations are haywire if you're transforming values that are affected by supports* config properties.

erquhart:
Fixes #6912 

---
## Summary by cubic
Apply customTransformInput to where clause values so query filters are correctly transformed across all actions. Expand action types and adjust Mongo adapter logic to avoid failures on non-create operations.

- **Bug Fixes**
  - Where values now pass through customTransformInput for findOne, findMany, update, updateMany, delete, deleteMany, and count.
  - Expanded action union in DBAdapterFactoryConfig and updated MongoDB adapter to only run create-specific logic on create.

<sup>Written for commit 76ff42b917f2a7d2458634706c5460bd5b4b52ad. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/package.json
+++ b/packages/better-auth/package.json
@@ -1,6 +1,6 @@
 {
   "name": "better-auth",
-  "version": "1.4.8-beta.4",
+  "version": "1.4.8-beta.5",
   "description": "The most comprehensive authentication framework for TypeScript.",
   "type": "module",
   "license": "MIT",


ping-maxwell:
hey I'm looking into the mongo CI failing, hope you don't mind if I make changes on this PR

erquhart:
@ping-maxwell please do!

PR #6924, opened 2025-12-21
feat(stripe): handle `customer.subscription.created` webhook event

bytaesu:
This PR changes:

- Handles `customer.subscription.created` webhook to sync subscriptions created outside the checkout flow (e.g. Stripe Dashboard or API)
- Updated Stripe plugin to use `logger` from context instead of global

### To-do
- [x] Verify behavior in the real app setup

---
## Summary by cubic
Add support for the Stripe customer.subscription.created webhook to sync subscriptions created outside checkout with our database and prevent broken subscription states.

- **New Features**
  - Handle customer.subscription.created in the webhook route.
  - Create subscription records for users matched by stripeCustomerId; map plan via priceId/lookup_key; set seats, period, and trial dates.
  - Add subscription.onSubscriptionCreated callback and document it; docs updated to include the event and getCheckoutSessionParams.

- **Bug Fixes**
  - Skip creation if the subscription already exists; guard when user or plan is missing; use request-scoped logging.
  - Add tests for success, duplicate, missing user, and missing plan scenarios.

<sup>Written for commit cf7243d24925a373c90d6aeab6227a5a269d1d37. Summary will update on new commits.</sup>


packages/stripe/src/types.ts
```diff
@@ -257,6 +257,18 @@ export type SubscriptionOptions = {
...
+	/**
+	 * A callback to run when a subscription is created
+	 * @returns
+	 */
+	onSubscriptionCreated?:
+		| ((data: {
+				event: Stripe.Event;
+				stripeSubscription: Stripe.Subscription;
+				subscription: Subscription;
+				plan?: StripePlan;
```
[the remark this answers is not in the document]
bytaesu: good

bytaesu:
=> I'm improving docs

packages/stripe/src/hooks.ts
```diff
@@ -93,7 +92,124 @@ export async function onCheckoutSessionCompleted(
...
+			});
+		if (existingSubscription) {
+			ctx.context.logger.info(
+				`Stripe webhook: Subscription ${subscriptionCreated.id} already exists in database, skipping creation`,
+			);
+			return;
+		}
+
+		// Find user by stripeCustomerId
+		const user = await ctx.context.adapter.findOne<{ id: string }>({
```
himself65: Use `User` type here

packages/stripe/src/hooks.ts
```diff
@@ -93,7 +92,124 @@ export async function onCheckoutSessionCompleted(
...
+		const priceLookupKey = subscriptionItem.price.lookup_key || null;
+		const plan = await getPlanByPriceInfo(options, priceId, priceLookupKey);
+		if (!plan) {
+			ctx.context.logger.warn(
+				`Stripe webhook warning: No matching plan found for priceId: ${priceId}`,
+			);
+			return;
+		}
+
+		const seats = subscriptionItem.quantity || 1;
```
himself65: allow it to be `undefined`

packages/stripe/src/hooks.ts
bytaesu: Thanks

packages/stripe/src/hooks.ts
bytaesu: I updated it so the hook accepts `undefined`.

In the routes it's being converted to 1 using `|| 1`. What's the difference between these?

packages/stripe/src/hooks.ts
```diff
@@ -1,5 +1,4 @@
-import type { GenericEndpointContext } from "better-auth";
-import { logger } from "better-auth";
+import type { GenericEndpointContext, User } from "better-auth";
```
himself65: can we import type from `@better-auth/core` ?

bytaesu: https://github.com/bytaesu/better-auth/blob/561f88d88b8d0c9c034df8b8dc18211b74ddc8a8/packages/stripe/src/types.ts#L6

The User type exists in `better-auth`, and other files already import it from there, so it seems cleaner to keep this consistent 🤔

screen

bytaesu:
This PR needs overall updates to align with recent merged changes. I'll prioritize https://github.com/better-auth/better-auth/pull/6961 first.

PR #6950, opened 2025-12-23
fix(stripe): prevent trial abuse by checking all user subscriptions

### LINKED ISSUE #6863: [Stripe] Trial abuse: `hasEverTrialed` check uses wrong subscription set

### To Reproduce

I've had issues where a user is able to get multiple trials, and this seems to be what's happening.

The PR #4562 fixed cross-plan trial abuse but I think it missed this: `hasEverTrialed` check can still return false for users who have trialed before.

https://github.com/better-auth/better-auth/blob/canary/packages/stripe/src/routes.ts#L202-L206
  ```ts
  // findOne returns arbitrary subscription for referenceId
  let subscriptionToUpdate = ...
      : referenceId
          ? await ctx.context.adapter.findOne<Subscription>({
              where: [{ field: "referenceId", value: referenceId }],
          })

  //  subscriptions becomes just that ONE record
  const subscriptions = subscriptionToUpdate
      ? [subscriptionToUpdate]
      : await ctx.context.adapter.findMany...
````

When user has multiple subscriptions (e.g. one canceled w/ trial history, one new incomplete), `findOne` returns whichever the DB returns first. If it's the new incomplete one, `hasEverTrialed` is false.

How to reproduce

1. User subscribes with 7-day trial, completes checkout
2. Subscription record has trialStart/trialEnd set
3. User cancels -> status becomes "canceled"
4. User starts new subscription -> new "incomplete" record created
5. `findOne` returns the new incomplete subscription
6. `subscriptions = [subscriptionToUpdate]` contains only that one
7. `hasEverTrialed` returns false → user gets another trial

Fix The trial check should always query ALL subscriptions:

```ts
  // Always check ALL subscriptions for trial history, not just `subscriptionToUpdate`
  const allSubscriptionsForTrialCheck = await ctx.context.adapter.findMany<Subscription>({
      model: "subscription",
      where: [{ field: "referenceId", value: referenceId }],
  });

  const hasEverTrialed = allSubscriptionsForTrialCheck.some((s) => {
      const hadTrial = !!(s.trialStart || s.trialEnd) || s.status === "trialing";
      return hadTrial;
  });
```

This separates the "which subscription to update" logic from the "has user ever trialed" logic.

The workaround is to have something like this:
```
async function hasUserEverTrialed(userId: string): Promise<boolean> {
  const subscriptions = await db.query.subscription.findMany({
    where: eq(schema.subscription.referenceId, userId),
  });

  return subscriptions.some(
    (s) =>
      s.status === "trialing" ||
      s.status === "active" ||
      s.status === "canceled"
  );
}
```
And call it inside `getCheckoutSessionParams`:
```
        getCheckoutSessionParams: async ({ user }, request) => {
          // Workaround: check trial history across ALL subscriptions
          const hadTrial = await hasUserEverTrialed(user.id);
          const trialOverride = hadTrial
            ? { subscription_data: { trial_period_days: undefined } }
            : {};
...
```

Related: #4539, #4046, #4562

Users should not be able to trial abuse.

1.4.7

### System info

```bash
n/a
```

Backend

```typescript

```

### Additional context

_No response_

GautamBytes:
### Description
Fixes #6863

**The Issue:**
Previously, the `hasEverTrialed` check relied on the `subscriptions` variable, which in some flows (like upgrading/updating) was optimized to contain only a single record (`subscriptionToUpdate`). This meant that if a user canceled a previous trial and started a new subscription, the check would only see the new "incomplete" record, fail to find past trial history, and grant a new trial.

**The Fix:**
I decoupled the trial history check from the update logic. The code now explicitly executes a `findMany` query to fetch **all** subscriptions associated with the `referenceId` (User ID). This ensures that `hasEverTrialed` validates the user's entire history, regardless of which specific subscription is currently being processed.

---
## Summary by cubic
Prevent trial abuse by checking a user's full subscription history when upgrading or creating Stripe subscriptions. We now query all subscriptions for the user before granting any trial.

- **Bug Fixes**
  - Fetch all subscriptions by referenceId via adapter.findMany and compute hasEverTrialed across any plan.
  - Decoupled the trial check from the update flow so it no longer relies on a possibly partial subscriptions array.
  - Added a test to ensure a canceled past trial blocks new trials even when an incomplete subscription exists.

<sup>Written for commit db89b695ccb2d35108150594756bf41571b1e2bd. Summary will update automatically on new commits.</sup>


bytaesu:
Thanks @GautamBytes,
It looks good, I'll make sure to test it today!

himself65 (approved):

bytaesu (changes requested):
It seems this test doesn't cover the issue.

Image

bytaesu:
Hi @GautamBytes,

I wanted the test case to cover the actual issue, so I updated it! Thanks for the contribution 🙂

---

### Before
before

### After
after

bytaesu (approved):

PR #6954, opened 2025-12-23
fix: BETTER_AUTH_SECRET length check should not throw

Diabl0570:
In this MR: https://github.com/better-auth/better-auth/pull/6114 there is a check added that throws if the secret is <32 chars.

However for implementation currently running with a secret of <32 this now breaks the application.

I could also not find straight forward migration steps on how to rotate your secret.

---
## Summary by cubic
Stop throwing when BETTER_AUTH_SECRET is shorter than 32 characters. We now log a warning to avoid breaking existing deployments while still encouraging stronger secrets.

- **Bug Fixes**
  - validateSecret logs a warning instead of throwing when secret < 32.
  - Prevents startup failures in apps using shorter secrets.

- **Migration**
  - Generate a 32+ char secret: npx @better-auth/cli secret or openssl rand -base64 32.
  - Update BETTER_AUTH_SECRET and restart; sessions signed with the old secret may require re-auth.

<sup>Written for commit 38fcd41562b7b9c724704f924efd672607507912. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/context/create-context.test.ts
+++ b/packages/better-auth/src/context/create-context.test.ts
@@ -1254,13 +1254,23 @@ describe("base context creation", () => {
 			vi.unstubAllEnvs();
 		});
 
-		it("should throw error when secret is too short", async () => {
+		it("should log a warning when secret is too short", async () => {
 			vi.stubEnv("BETTER_AUTH_SECRET", "");
 			vi.stubEnv("AUTH_SECRET", "");
 			const originalNodeEnv = process.env.NODE_ENV;
-
-			const expectedErrorMessage =
-				"Invalid BETTER_AUTH_SECRET: must be at least 32 characters long for adequate security. Generate one with `npx @better-auth/cli secret` or `openssl rand -base64 32`.";
+			const log = vi.fn();
+			const res = await initBase({
+				logger: {
+					level: "warn",
+					log,
+				} as any,
+				database: new Database(":memory:"),
+				session: {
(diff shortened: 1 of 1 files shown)


bytaesu:
Hi @Diabl0570,

This is necessary for proper security. Since it’s a breaking change, we’ll just warn for now and make it mandatory with proper guidance in v1.5 🙂

AyushSehrawat:
> Hi @Diabl0570,
> 
> This is necessary for proper security. Since it’s a breaking change, we’ll just warn for now and make it mandatory with proper guidance in v1.5 🙂

Hi, like i mentioned in discord thread [here](https://discord.com/channels/1288403910284935179/1288403910284935182/1453033817333043210) , this gives issue when you are building public docker images, workaround i found is to have a runtime secret generation. 

Or maybe some env to disable the check

bytaesu:
Hi @AyushSehrawat,

Could you share the link with me? I want to take a look!

AyushSehrawat:
> Hi @AyushSehrawat,
> 
> Could you share the link with me? I want to take a look!

Yea. I build public docker images for my project using GitHub actions, so i try to avoid setting any default ENV/ARG in docker. But with the throw error and entropy check it was failing,

[rivenmedia/riven-frontend/actions/runs/20438120803/job/58724382835](https://github.com/rivenmedia/riven-frontend/actions/runs/20438120803/job/58724382835)

Workaround was runtime generation of env.

PR #6966, opened 2025-12-23
docs: add notes about cookie cache session revocation

### LINKED ISSUE #6893: Even if the cookieCache is configured and the authClient.revokeOtherSessions() is called, the session will only become invalid when the cache expires.

### To Reproduce

1. Multi-device login to the same account
2. When one of the devices successfully logged in after resetting the password, the following code was executed: authClient.revokeOtherSessions()
3. The database only has one piece of data for the current user's session. This is normal.
4.However, the other device was still in a logged-in state. After my test, I found that it would only expire after 5 minutes, with the configuration being cookieCache: {enabled:true, maxAge: 5 * 60,}. This is not the result I was expecting.

After calling `authClient.revokeOtherSessions()`, the sessions of other devices remain valid within the cache period.

1.3.14

### System info

```bash
"better-auth": "^1.3.14",
```

Client

```typescript
await authClient.revokeOtherSessions();

  session: {
    cookieCache: {
      enabled: false,
      maxAge: 5 * 60, // Cache duration in seconds
    },
  },
```

### Additional context

_No response_

bytaesu:
Make users aware in advance of behaviors caused by cookie caching and allow them to balance between performance and security.

Closes #6893

---
## Summary by cubic
Added notes to Session Management docs about how cookieCache can delay session revocation across devices. Includes mitigation guidance (disable cookieCache, shorten maxAge, or use disableCookieCache for sensitive actions).

<sup>Written for commit d109bf7effb10aebdab4a964348189305056ed51. Summary will update automatically on new commits.</sup>

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -401,7 +401,7 @@ importers:
         version: 12.23.12(react-dom@19.2.3(react@19.2.3))(react@19.2.3)
       geist:
         specifier: ^1.4.2
-        version: 1.4.2(next@16.0.10(@babel/core@7.28.4)(@opentelemetry/api@1.9.0)(@playwright/test@1.56.1)(babel-plugin-react-compiler@1.0.0)(react-dom@19.2.3(react@19.2.3))(react@19.2.3)(sass@1.90.0))
+        version: 1.4.2(next@16.0.10(@opentelemetry/api@1.9.0)(@playwright/test@1.56.1)(babel-plugin-react-compiler@1.0.0)(react-dom@19.2.3(react@19.2.3))(react@19.2.3)(sass@1.90.0))
       input-otp:
         specifier: ^1.4.2
         version: 1.4.2(react-dom@19.2.3(react@19.2.3))(react@19.2.3)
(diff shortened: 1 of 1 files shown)


docs/content/docs/concepts/session-management.mdx
```diff
@@ -177,6 +177,21 @@ export const auth = betterAuth({
 });
 ```
 
+<Callout type="info">
+**Notes**
+
+When `cookieCache` is enabled, revoked sessions may remain active on other devices until the cookie cache expires (`maxAge`). This is because:
+
+- Cookie cache stores session data in the client's browser for performance
```
himself65: Actually, it's not (only) for performance. I think it's for like stateless or distributed design

bytaesu: Okay, I removed it since it felt sufficient to describe the behavior rather than go deep into the reason here 🙂

pnpm-lock.yaml
himself65, line 1: remove this change

himself65 (changes requested):

Issue #6291, opened 2025-11-25
SSO: List Providers


XavierGeerinck:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

I am currently looking for a way to list existing SSO Providers. Is this something that is supported already?

### Describe the solution you'd like

I'd love to see the possibility to list SSO Providers

### Describe alternatives you've considered

Adding my custom extension

PR #6967, opened 2025-12-23
feat(sso): add provider list and detail endpoints

Paola3stefania:
Closes https://github.com/better-auth/better-auth/issues/6291

---
## Summary by cubic
Adds SSO provider list, detail, update, and delete endpoints with sanitized OIDC/SAML data and org-aware access.

- **New Features**
  - Endpoints: GET /sso/providers and GET/PATCH/DELETE /sso/providers/:providerId return sanitized OIDC/SAML data (masked clientId, no clientSecret, parsed cert metadata, spMetadataUrl) and enforce access for org admins/owners or provider owners.
  - Context: Added hasPlugin to check enabled plugins (used in SSO and SCIM); partial updates validate schemas/algorithms and reset domainVerified on domain changes.

<sup>Written for commit b6be50d955cacaf95be1e5c272743e7cc9e867f6. Summary will update on new commits.</sup>

### DIFF
--- a/packages/expo/src/focus-manager.ts
+++ b/packages/expo/src/focus-manager.ts
@@ -6,6 +6,7 @@ import { AppState } from "react-native";
 class ExpoFocusManager implements FocusManager {
 	listeners = new Set<FocusListener>();
 	subscription?: ReturnType<typeof AppState.addEventListener>;
+	isFocused: boolean | undefined;
 
 	subscribe(listener: FocusListener) {
 		this.listeners.add(listener);
@@ -15,6 +16,8 @@ class ExpoFocusManager implements FocusManager {
 	}
 
 	setFocused(focused: boolean) {
+		if (this.isFocused === focused) return;
+		this.isFocused = focused;
 		this.listeners.forEach((listener) => listener(focused));
 	}
 
--- a/packages/expo/src/online-manager.ts
+++ b/packages/expo/src/online-manager.ts
@@ -14,6 +14,7 @@ class ExpoOnlineManager implements OnlineManager {
 	}
 
 	setOnline(online: boolean) {
+		if (this.isOnline === online) return;
 		this.isOnline = online;
 		this.listeners.forEach((listener) => listener(online));
 	}
--- a/packages/expo/test/expo.test.ts
+++ b/packages/expo/test/expo.test.ts
@@ -876,3 +876,47 @@ describe("expo deep link cookie injection for verify-email", async () => {
 		expect(error).toBeDefined();
 	});
 });
+
+describe("ExpoFocusManager duplicate notification prevention", () => {
+	it("should not notify listeners when setFocused is called with the same value", async () => {
+		const { setupExpoFocusManager } = await import("../src/focus-manager");
+		const focusManager = setupExpoFocusManager();
+
+		const listener = vi.fn();
+		focusManager.subscribe(listener);
+
+		focusManager.setFocused(true);
+		expect(listener).toHaveBeenCalledTimes(1);
+
+		focusManager.setFocused(true);
+		expect(listener).toHaveBeenCalledTimes(1);
+
+		focusManager.setFocused(false);
+		expect(listener).toHaveBeenCalledTimes(2);
+
+		focusManager.setFocused(false);
+		expect(listener).toHaveBeenCalledTimes(2);
+	});
+});
+
+describe("ExpoOnlineManager duplicate notification prevention", () => {
+	it("should not notify listeners when setOnline is called with the same value", async () => {
+		const { setupExpoOnlineManager } = await import("../src/online-manager");
+		const onlineManager = setupExpoOnlineManager();
+
+		const listener = vi.fn();
+		onlineManager.subscribe(listener);
+
+		onlineManager.setOnline(false);
+		expect(listener).toHaveBeenCalledTimes(1);
+
+		onlineManager.setOnline(false);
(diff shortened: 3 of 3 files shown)


packages/sso/src/types.ts
```diff
@@ -343,3 +343,10 @@ export interface SSOOptions {
 		algorithms?: AlgorithmValidationOptions;
 	};
 }
+
+export interface Member {
+	id: string;
+	userId: string;
+	organizationId: string;
+	role: string;
+}
```
Paola3stefania, line 364: I dont want to couple the SSO package to the organization package's types

packages/sso/src/routes/providers.ts
```diff
@@ -0,0 +1,238 @@
...
+import type { Member, OIDCConfig, SAMLConfig } from "../types";
+import { maskClientId, parseCertificate, safeJsonParse } from "../utils";
+
+const ADMIN_ROLES = ["owner", "admin"];
+
+function isOrgPluginEnabled(ctx: {
+	context: { options: { plugins?: { id: string }[] } };
+}): boolean {
+	return (
+		ctx.context.options.plugins?.some((p) => p.id === "organization") ?? false
```
himself65: nit: we need a better way to check plugin exist

Paola3stefania: Going deeper I think we have this 3 options

1. **Add `hasPlugin(id)` method to `AuthContext`** - cleanest 
   ```ts
   ctx.context.hasPlugin("organization")
   ```

2. **Standalone utility function in core**
   ```ts
   hasPlugin(ctx, "organization")
   ```

3. **Add `enabledPlugins: Set<string>` property to context**
   ```ts
   ctx.context.enabledPlugins.has("organization")
   ```

I'd lean toward option 1 - most discoverable and self-documenting. Would implement it

packages/core/src/types/context.ts
```diff
@@ -294,4 +294,18 @@ export type AuthContext<Options extends BetterAuthOptions = BetterAuthOptions> =
...
+		 * @returns `true` if the plugin is enabled, `false` otherwise
+		 *
+		 * @example
+		 * ```ts
+		 * if (ctx.context.hasPlugin("organization")) {
+		 *   // organization plugin is enabled
+		 * }
+		 * ```
+		 */
+		hasPlugin: (pluginId: string) => boolean;
```
himself65: I think we should put `hasPlugin` API into a different PR

packages/sso/src/utils.ts
```diff
@@ -39,3 +41,30 @@ export const validateEmailDomain = (email: string, domain: string) => {
 		emailDomain === providerDomain || emailDomain.endsWith(`.${providerDomain}`)
 	);
 };
+
+export function parseCertificate(certPem: string) {
+	try {
+		const normalized = certPem.includes("-----BEGIN")
```
himself65: we do we check `-----BEGIN` here? Does the developer often pass the cretPerm without a prefix?

packages/sso/src/utils.ts
```diff
@@ -39,3 +41,30 @@ export const validateEmailDomain = (email: string, domain: string) => {
...
+
+		return {
+			fingerprintSha256: cert.fingerprint256,
+			notBefore: cert.validFrom,
+			notAfter: cert.validTo,
+			publicKeyAlgorithm:
+				cert.publicKey.asymmetricKeyType?.toUpperCase() || "UNKNOWN",
+		};
+	} catch {
+		return { certParseError: "Failed to parse certificate" };
```
himself65: Why not throw error

packages/sso/src/routes/providers.ts
```diff
@@ -0,0 +1,230 @@
...
+		samlConfig: samlConfig
+			? {
+					entryPoint: samlConfig.entryPoint,
+					callbackUrl: samlConfig.callbackUrl,
+					audience: samlConfig.audience,
+					wantAssertionsSigned: samlConfig.wantAssertionsSigned,
+					identifierFormat: samlConfig.identifierFormat,
+					signatureAlgorithm: samlConfig.signatureAlgorithm,
+					digestAlgorithm: samlConfig.digestAlgorithm,
+					certificate: parseCertificate(samlConfig.cert),
```
himself65: certificate could be `{ certPastError: '...' }`

himself65 (dismissed):

packages/core/src/types/context.ts
Paola3stefania: I will create a new pr and rebase this one

Paola3stefania: https://github.com/better-auth/better-auth/pull/7157

packages/sso/src/utils.ts
Paola3stefania: Yes, they often pass certificates without PEM headers - specifically when the certificate comes from SAML metadata XML, where <X509Certificate> elements contain raw base64 only (no -----BEGIN/-----END)

packages/sso/src/utils.ts
Paola3stefania: As its a read-only endpoint that returns provider info, throwing would fail the entire request if one provider has a malformed certificate... 
 
It Returns { certParseError: "Failed to parse certificate" } so the API response still succeeds
Admin can see which provider has a bad cert and fix it
Other providers in the list are still returned

But for type safety and error handling it would look better to throw instead. Changing that.

Issue #6958, opened 2025-12-23
MCP redirect bug with magic link auth


amirilovic:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Reproduction steps can be found here: https://github.com/amirilovic/better-auth-magic-link-mcp-issue

## Expected behavior

When auth flow is started by mcp client, redirect back to the client should happen after successful login.

## Current behavior

User is redirected to /api/auth/error?error=invalid_client

### What version of Better Auth are you using?

1.4.8

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.5.0: Tue Apr 22 19:53:27 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6041",
    "release": "24.5.0",
    "cpuCount": 14,
    "cpuModel": "Apple M4 Pro",
    "totalMemory": "48.00 GB",
    "freeMemory": "0.19 GB"
  },
  "node": {
    "version": "v22.21.1",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.4"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "^15.1.0"
    },
    {
      "name": "react",
      "version": "^19.0.0"
    }
  ],
  "databases": [
    {
      "name": "better-sqlite3",
      "version": "^12.0.0"
    }
  ],
  "betterAuth": {
    "version": "1.4.8",
    "config": {
      "database": {
        "name": "./auth.db",
        "open": true,
        "inTransaction": false,
        "readonly": false,
        "memory": false
      },
      "emailAndPassword": {
        "enabled": false
      },
      "plugins": [
        {
          "name": "magic-link",
          "config": {
            "id": "magic-link",
            "endpoints": {},
            "rateLimit": [
              {
                "window": 60,
                "max": 5
              }
            ],
            "options": {}
          }
        },
        {
          "name": "mcp",
          "config": {
            "id": "mcp",
            "hooks": {
              "after": [
                {}
              ]
            },
            "endpoints": {},
            "schema": {
              "oauthApplication": {
                "modelName": "oauthApplication",
                "fields": {
                  "name": {
                    "type": "string"
                  },
                  "icon": {
                    "type": "string",
                    "required": false
                  },
                  "metadata": {
                    "type": "string",
                    "required": false
                  },
                  "clientId": {
                    "type": "string",
                    "unique": true
[...]

PR #6974, opened 2025-12-24
fix(mcp): restore ctx.query from cookie in OAuth flow

bytaesu:
Closes https://github.com/better-auth/better-auth/issues/6958

---
## Summary by cubic
Fixes MCP OAuth login redirects so unauthenticated users who log in are sent back to the client/consent flow, not /api/auth/error.

- **Bug Fixes**
  - Restore ctx.query from the persisted oidc_login_prompt cookie during OAuth to continue with the original authorization params.
  - Add a test that asserts cookie persistence and a valid redirect (consent or callback with code) without invalid_client or error.

<sup>Written for commit 3eb948786a32ea3d63bf05127903fc0c72bc9606. Summary will update on new commits.</sup>


Bekacru:
mcp issues can now be closed and redirect the user to use the `oauth-provider` instead

Issue #6810, opened 2025-12-16
[BUG] [Expo] Fail in useSession due to invalid SecureStore cookie


MarkLee425:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

In expo development build and expo go, when i use ios for testing google sso, better auth successfully saved a cookie inside SecureStore, but the format is like this:
`{"null":{"value":"","expires":null},"etigo.state":{"value":"i_BhlUCZY-M3ol74KfE_zxZiqfCv-11s.l%2FM%2BJlmkYs4D1THnd41a%2Fu5B5wOjhvDhUulGEBrCDVI%3D","expires":"2025-12-20T10:09:37.426Z"}}`

I am not sure if it is because of invalid cookie format or something else, eventually it will always get NULL from authClient.useSession.

For expo web, it works with no issues. 

### Current vs. Expected behavior

Just found the issue:
Somehow in backend, console logging the cookie before sending to expo app is 
etigo.session_token=4YXFfpe99PBGu9xrgP6pRS8PdFIwnqFF.C%2BrrkiE7DUTHvZwOnx134VmRsp4Z92MCYZ0VOTkenG4%3D; etigo.state=i_BhlUCZY-M3ol74KfE_zxZiqfCv-11s.l%2FM%2BJlmkYs4D1THnd41a%2Fu5B5wOjhvDhUulGEBrCDVI%3D

But eventually in expo securestore, it will become 
`{"null":{"value":"","expires":null},"etigo.state":{"value":"TEjfBa5W4mii9rdKTe09lJ8upQgLZ47v.QUBHaetxm%2F0%2BaJff7Ucb9Nm8tk2eOQ5xyDniIKHUy1s%3D","expires":"2025-12-20T10:36:10.857Z"}}
`

The session token is missing

Expected behavior:
`{"etigo.session_token":{"value":"the-token","expires":"2025-12-20T10:36:10.857Z"},"etigo.state":{"value":"TEjfBa5W4mii9rdKTe09lJ8upQgLZ47v.QUBHaetxm%2F0%2BaJff7Ucb9Nm8tk2eOQ5xyDniIKHUy1s%3D","expires":"2025-12-20T10:36:10.857Z"}}`

### What version of Better Auth are you using?

1.4.7

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
[...]

PR #6990, opened 2025-12-24
fix(expo): improve `parseSetCookieHeader`

bytaesu:
Related to https://github.com/better-auth/better-auth/issues/6810

---

**_Note:_**  
Currently, `parseSetCookieHeader` already exists in `better-auth/cookies`, but the Expo plugin defines and uses it separately. I think It should be unified, but since `better-auth/cookies` also contains server-side utilities, I haven’t touched it for now.

---
## Summary by cubic
Update Expo’s parseSetCookieHeader to skip cookies with empty names, safely handle missing name/value segments, and ignore empty attributes. Prevents invalid cookies from being stored, avoids parsing errors from malformed Set-Cookie headers, and preserves existing session cookies when malformed headers are received.

<sup>Written for commit ec94d64f7543bf9995108956dc206e74a0d7424a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/expo/test/expo.test.ts
+++ b/packages/expo/test/expo.test.ts
@@ -151,7 +151,7 @@ describe("expo", async () => {
 		);
 
 		// Simulate malformed cookie header starting with semicolon
-		const malformedHeader = "; etigo.state=xyz; Path=/";
+		const malformedHeader = "; abc.state=xyz; Path=/";
 		const parsed = parseSetCookieHeader(malformedHeader);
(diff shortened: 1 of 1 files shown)


himself65 (commented):
Add test

bytaesu:
> Add test

Just added!

I hadn't committed it earlier because having cookie header parsing tests in multiple places felt wrong.. but, need it for now 🤔... Anyway, it’s better to unify parseSetCookieHeader in one place

himself65 (approved):

PR #7009, opened 2025-12-26
fix(build): externalize @better-auth/core to fix Expo/Metro resolution

anantasharma510:
When using Expo/Metro, the bundler was incorrect rewriting @better-auth/core imports to relative paths causing module resolution errors. This PR adds @better-auth/core to external in tsdown config to preserve correct import paths.

---
## Summary by cubic
Externalizes @better-auth/core in tsdown config so bundlers keep the package import and don’t rewrite it to relative paths. Fixes Expo/Metro module resolution errors.

<sup>Written for commit 30a862bc5b7b14da27f96db66b22acd5b2bed179. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/tsdown.config.ts
+++ b/packages/better-auth/tsdown.config.ts
@@ -64,4 +64,5 @@ export default defineConfig({
 	treeshake: true,
 	clean: true,
 	unbundle: true,
+	external: ["@better-auth/core"],
 });


packages/better-auth/tsdown.config.ts
```diff
@@ -64,4 +64,5 @@ export default defineConfig({
 	treeshake: true,
 	clean: true,
 	unbundle: true,
```
himself65, line 66: This is not possible, `unbundle` mode is already externalize all imports

himself65:
I think it's from your build config issue. Closing

packages/better-auth/tsdown.config.ts
anantasharma510: Thanks for the review!

You are right that unbundle: true should handle this, but for some reason, the workspace:* dependency seems to be causing the bundler to rewrite @better-auth/core imports into relative paths in the final output.

I encountered this error in an Expo project using v1.4.9:

Unable to resolve "../error.mjs" from "node_modules\better-auth\dist\utils\url.mjs"
This suggests that dist/utils/url.mjs contained a relative import ../error.mjs (which doesn't exist in the package), instead of preserving the external import import { ... } from "@better-auth/core/error".

Explicitly adding it to external fixed the output to correctly keep the package import. It might be due to how tsdown handles workspace resolution vs package dependencies.

anantasharma510:
Thanks for the review!

You are right that unbundle: true should handle this, but for some reason, the workspace:* dependency seems to be causing the bundler to rewrite @better-auth/core imports into relative paths in the final output.

I encountered this error in an Expo project using v1.4.9:

Unable to resolve "../error.mjs" from "node_modules\better-auth\dist\utils\url.mjs"
This suggests that dist/utils/url.mjs contained a relative import ../error.mjs (which doesn't exist in the package), instead of preserving the external import import { ... } from "@better-auth/core/error".

Explicitly adding it to external fixed the output to correctly keep the package import. It might be due to how tsdown handles workspace resolution vs package dependencies.

himself65:
> Thanks for the review!
> 
> You are right that unbundle: true should handle this, but for some reason, the workspace:* dependency seems to be causing the bundler to rewrite @better-auth/core imports into relative paths in the final output.
> 
> I encountered this error in an Expo project using v1.4.9:
> 
> Unable to resolve "../error.mjs" from "node_modules\better-auth\dist\utils\url.mjs" This suggests that dist/utils/url.mjs contained a relative import ../error.mjs (which doesn't exist in the package), instead of preserving the external import import { ... } from "@better-auth/core/error".
> 
> Explicitly adding it to external fixed the output to correctly keep the package import. It might be due to how tsdown handles workspace resolution vs package dependencies.

Thanks for the following up. I will take a look

himself65:
image

Can you check if you have install `@better-auth/core` in your locally?

PR #7013, opened 2025-12-27
fix(client): signal race condition fix in the createDynamicPathProxy function

mayur9210:
Tests for the signal race condition fix in the `createDynamicPathProxy` function and refactors the signal toggling logic to use the microtask queue for improved async coordination. The changes ensure that signal updates are reliable, prevent race conditions, and improve the robustness of the proxy client.

**Signal update logic:**

* Refactored signal toggling in `createDynamicPathProxy` to use `queueMicrotask` instead of `setTimeout`, ensuring atomic updates and eliminating race conditions during async operations.

**Testing improvements:**

* Added a new test suite `proxy.test.ts` to validate signal toggling behavior, concurrent API calls, disabling signal updates, handling missing signals, listener matching, and custom callbacks in `createDynamicPathProxy`.

---
## Summary by cubic
Fixed signal race conditions in createDynamicPathProxy by toggling atom signals immediately in onSuccess with deduplication. Added tests for concurrent calls, disabled/missing signals, listener matching, custom onSuccess, and deduplicated signals.

- **Bug Fixes**
  - Removed microtask/setTimeout scheduling and toggle signals synchronously to avoid races.
  - Deduplicated toggles when multiple listeners target the same signal.
  - Added safety for missing atoms and only toggle when signal.set exists.

<sup>Written for commit 1b740d64142b589d64fd015e8218b9c8e6079d43. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/client/proxy.test.ts
+++ b/packages/better-auth/src/client/proxy.test.ts
@@ -1,5 +1,3 @@
-// packages/better-auth/src/client/proxy.test.ts
-
 import type { BetterAuthClientPlugin } from "@better-auth/core";
 import { atom } from "nanostores";
 import { beforeEach, describe, expect, it, vi } from "vitest";
@@ -38,7 +36,7 @@ describe("createDynamicPathProxy - Signal Race Condition Fix", () => {
 		];
 	});
 
-	it("should toggle signal without race condition using queueMicrotask", async () => {
+	it("should toggle signal correctly on API call", async () => {
 		const proxy = createDynamicPathProxy(
 			{},
 			mockFetch,
--- a/packages/better-auth/src/client/proxy.ts
+++ b/packages/better-auth/src/client/proxy.ts
@@ -108,12 +108,9 @@ export function createDynamicPathProxy<T extends Record<string, any>>(
 							const signal = atoms[signalName as any];
 							if (!signal) continue;
 
-							// microtask queue for proper async coordination
-							queueMicrotask(() => {
-								if ("set" in signal) {
-									signal.set(!signal.get());
-								}
-							});
+							if ("set" in signal) {
+								signal.set(!signal.get());
+							}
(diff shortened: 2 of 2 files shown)


packages/better-auth/src/client/proxy.test.ts
```diff
@@ -0,0 +1,246 @@
+// packages/better-auth/src/client/proxy.test.ts
```
himself65: remove this comment

packages/better-auth/src/client/proxy.ts
```diff
@@ -100,17 +100,20 @@ export function createDynamicPathProxy<T extends Record<string, any>>(
...
+
+						// Deduplicate signals to avoid toggling the same signal multiple times
+						const uniqueSignals = new Set(matches.map((m) => m.signal));
+
+						for (const signalName of uniqueSignals) {
+							const signal = atoms[signalName as any];
+							if (!signal) continue;
+
+							// microtask queue for proper async coordination
+							queueMicrotask(() => {
```
himself65: i don't think this gonna helps the race condition

packages/better-auth/src/client/proxy.test.ts
mayur9210: Done, removed the comment

packages/better-auth/src/client/proxy.ts
mayur9210: You're right, The actual race condition protection comes from deduplicating signals upfront using `new Set(matches.map(m => m.signal))`, which ensures each signal is only toggled once per onSuccess call regardless of how many listeners match. `queueMicrotask` doesn't prevent race conditions - it just defers execution. I've removed it.

himself65:
Landed in https://github.com/better-auth/better-auth/pull/7018

PR #7076, opened 2025-12-31
fix(api-key): remove double stringify/parse of metadata field

xiaoyu2er:
## Problem

The metadata field in the API key plugin was being transformed twice:

1. **Manual stringify**: `schema.apikey.fields.metadata.transform.input(metadata)` in the route handlers
2. **Automatic stringify**: The adapter factory applies `transform.input` to all fields with transforms defined in the schema

This caused metadata to be double-stringified when stored:
```
Input: { organizationId: "abc" }
After first stringify: '{"organizationId":"abc"}'
After second stringify: '"{\\"organizationId\\":\\"abc\\"}"'
```

## Solution

- Remove manual `transform.input` calls in `create-api-key.ts` and `update-api-key.ts`
- Remove manual `transform.output` calls in `list-api-keys.ts`, `get-api-key.ts`, `update-api-key.ts`, and `verify-api-key.ts`
- Let the adapter factory handle all metadata transformations automatically

## ⚠️ Breaking Change

This is a **breaking change** for users who have existing API keys with metadata created using the buggy double-stringify behavior.

**Before this fix:**
- Metadata was double-stringified on write
- Double-parsed on read (adapter + manual transform) → returned as object ✅

**After this fix:**
- Metadata is properly stringified once on write
- Parsed once on read (adapter only) → returned as object ✅
- **However, OLD double-stringified data** will only be parsed once → returned as a **string** instead of object ❌

## Testing

All 120 API key tests pass.

### DIFF
--- a/packages/better-auth/src/plugins/api-key/api-key.test.ts
+++ b/packages/better-auth/src/plugins/api-key/api-key.test.ts
@@ -3203,7 +3203,7 @@ describe("api-key", async () => {
 	// =========================================================================
 
 	describe("legacy double-stringified metadata migration", async () => {
-		const { client, auth, signInWithTestUser, db } = await getTestInstance(
+		const { auth, signInWithTestUser, db } = await getTestInstance(
 			{
 				plugins: [
 					apiKey({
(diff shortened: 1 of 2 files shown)


xiaoyu2er:
@ping-maxwell @himself65 could you take a second look 🙏

PR #7077, opened 2025-12-31
docs(last-login-method): add tips about hydration mismatch & layout shift

moshetanzer:
Added note about SSR and hydration mismatches in Nuxt applications.

---
## Summary by cubic
Add a Nuxt SSR tip to the last-login-method docs to prevent hydration mismatches and layout shifts. In the Nuxt tab, recommend reading the cookie via `useCookie('better-auth.last_used_login_method')` (default cookie name) during SSR instead of calling `getLastUsedLoginMethod()`.

<sup>Written for commit 248346425a81621b0826df0aa152bdb5a744ce77. Summary will update on new commits.</sup>


docs/content/docs/plugins/last-login-method.mdx
```diff
@@ -60,6 +60,11 @@ const wasGoogle = authClient.isLastUsedLoginMethod("google")
 // Clear the stored method
 authClient.clearLastUsedLoginMethod()
 ```
+<Callout type="info">
```
himself65: We should also provide tips for other runtimes like Next.js, plase change it to tabs

himself65 (changes requested):

docs/content/docs/plugins/last-login-method.mdx
Bekacru: why do we need to provide tips for other frameworks? this shouldn't create hydration mismatch in other frameworks if im not mistake?

moshetanzer: just thought to make it easier for users :)? If ssr unless you use a ssr compatible cookie composable it will cause layout shifts and possiblehydration mismatches

himself65 (approved):

himself65:
I will update the content later

jd-solanki:
I agree with @Bekacru 

Fetching last login method using `getLastUsedLoginMethod` should work on SSR and not end up with a hydration error.

ping-maxwell (approved):

PR #7133, opened 2026-01-05
fix(last-login-method): correctly handle multiple Set-Cookie headers

dngpng:
## Problem

The last-login-method plugin may fail to set the login method cookie when multiple `Set-Cookie` headers were present in the response.

The issue occurred in the after hook handler:

https://github.com/better-auth/better-auth/blob/0827fc3f49c4968ea37f4a42a26a309edf75bea1/packages/better-auth/src/plugins/last-login-method/index.ts#L139-L143

### Why this may fail:

- HTTP responses can have multiple Set-Cookie headers (one per cookie)
- The `.get("set-cookie")` method only returns the first `Set-Cookie` header value
- If the session token cookie was not in the first header, the check would fail
- This caused the last-login-method cookie to not be set

### Real-world scenario:

When Better Auth sets multiple cookies (session token, CSRF token, additional plugin cookies, etc.), they each get their own Set-Cookie header. If the session token wasn't first, the last-login-method cookie would be silently skipped. (See screenshot below)

SCR-20260105-hvoi

## Solution

Replace `.get("set-cookie")` with `.getSetCookie()` which returns an array of all `Set-Cookie` headers:

```
  const setCookieHeaders = ctx.context.responseHeaders?.getSetCookie?.() || [];
  const hasSessionToken = setCookieHeaders.some((cookie) =>
    cookie.includes(sessionTokenName),
  );
```

This approach:

- ✅ Checks all Set-Cookie headers, not just the first one
- ✅ Uses the standard Headers API .getSetCookie() method designed for this purpose
- ✅ Includes optional chaining and fallback for compatibility
- ✅ Matches the pattern used elsewhere in the codebase (e.g., mcp.test.ts:577)

## Testing

Added a comprehensive test case that:

1. Creates a custom plugin that sets an additional cookie (simulating multiple Set-Cookie headers)
2. Verifies that multiple cookies are present in the response using getSetCookie()
3. Confirms the last-login-method cookie is still correctly set
4. Validates both cookies are present in the final response

### Test results:

  ✓ src/plugins/last-login-method/last-login-method.test.ts (10 tests)
    ✓ should handle multiple set-cookie headers correctly

  All existing tests continue to pass.

## Files Changed

- `packages/better-auth/src/plugins/last-login-method/index.ts` - Fixed Set-Cookie header handling
- `packages/better-auth/src/plugins/last-login-method/last-login-method.test.ts` - Added test for multiple Set-Cookie headers

## Related

This fix aligns with the Headers API standard and follows the same pattern used in other parts of the codebase for handling multiple Set-Cookie headers.

---
## Summary by cubic
Fixes the last-login-method plugin so it correctly sets the last_used_login_method cookie when responses include multiple Set-Cookie headers. Prevents skips when the session cookie isn’t the first header.

- **Bug Fixes**
  - Replaced get("set-cookie") with getSetCookie() to read all Set-Cookie headers.
  - Checked for the session token across headers before setting the cookie.
  - Added a test that simulates multiple cookies and verifies both cookies are set.

<sup>Written for commit fd1eb14a2cce498d6853a042c10a112192d02641. Summary will update on new commits.</sup>


dngpng:
@himself65 do i need to file an separate issue for this? the last login method will not work reliably until someone gives this PR some love 🥹

Bekacru (approved):

Issue #7135, opened 2026-01-05
"[object Object]" is not valid JSON when creating an OAuth client with metadata


bauerj:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Enable OAuth 2.1 provider plugin.
2. Create a new client using admin endpoint:

```
const client = await auth.api.adminCreateOAuthClient({
      headers: {
        cookie: cookie ?? '',
      },
      body: {
        redirect_uris: [`https://example.com/auth/callback`],
        skip_consent: true,
        metadata: {
          a: 'b',
        },
      },
    });
```

### Current vs. Expected behavior

I expected this to work since metadata is typed as Record<string, unkown>. However, I get this exception:

```
SyntaxError: "[object Object]" is not valid JSON
    at JSON.parse (<anonymous>)
    at schemaToOAuth (file:///home/bauerj/project/api/node_modules/.pnpm/@better-auth+oauth-provider@1.4.10_@better-auth+core@1.4.10_@better-auth+utils@0.3.0_@b_56bb7b0be66b3a227cf5ae860359839b/node_modules/@better-auth/oauth-provider/dist/index.mjs:1468:22)
    at createOAuthClientEndpoint (file:///home/bauerj/project/api/node_modules/.pnpm/@better-auth+oauth-provider@1.4.10_@better-auth+core@1.4.10_@better-auth+utils@0.3.0_@b_56bb7b0be66b3a227cf5ae860359839b/node_modules/@better-auth/oauth-provider/dist/index.mjs:1405:18)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async internalHandler (/home/bauerj/project/api/node_modules/.pnpm/better-call@1.1.7_zod@4.2.1/node_modules/better-call/src/endpoint.ts:507:20)
    at async <anonymous> (/home/bauerj/project/api/node_modules/.pnpm/better-auth@1.4.10_drizzle-kit@0.31.7_drizzle-orm@0.45.1_@types+pg@8.16.0_kysely@0.28.9_e69517855bdf5a28be3e1242deda0275/node_modules/better-auth/src/api/to-auth-endpoints.ts:119:22)
    at async AuthService.createOAuthClient (/home/bauerj/project/api/src/auth/auth.service.ts:21:20)
```

Apparently, it's trying to parse an object as JSON.

### What version of Better Auth are you using?

1.4.10

### System info

```bash
{
  "system": {
    "platform": "linux",
[...]

PR #7147, opened 2026-01-06
fix(oauth-provider): metadata field handling in client registration

bauerj:
The `metadata` field in `adminCreateOAuthClient` was causing a `SyntaxError: "[object Object]" is not valid JSON` error when creating OAuth clients. This occurred because:

1. The `metadata` field was not explicitly destructured in `oauthToSchema`, causing it to be included in the `...rest` spread
2. This resulted in double-wrapping: `{"metadata": {"foo": "bar"}}` instead of `{"foo": "bar"}`. This would result in storing JSON objects like { "metadata": ... } in the metadata column and returning double-nested metadata when querying.
3. Database adapters auto-parse JSON columns, returning objects instead of strings
4. `schemaToOAuth` attempted to `JSON.parse()` an already-parsed object, causing the error

This change enables setting the metadata field and returns data in a more useful format.

Closes: #7135 

---
## Summary by cubic
Fix metadata handling in OAuth client registration to stop double-wrapped data. Metadata is now stored and returned as a single nested object under the metadata field, and redirect_uris now defaults to an empty array.

- **Bug Fixes**
  - Explicitly destructure metadata in oauthToSchema to avoid double-wrapping.
  - Return metadata as a dedicated field in schemaToOAuth.
  - Default redirect_uris to [] to prevent missing field errors.
  - Type jwks as JSONWebKeySet and allow client_secret_expires_at as string or number (default 0 in create).

<sup>Written for commit 93b4681af4b89f1f766c12a8bcaa1d927b3132e7. Summary will update on new commits.</sup>

### DIFF
--- a/packages/oauth-provider/src/register.ts
+++ b/packages/oauth-provider/src/register.ts
@@ -252,8 +252,7 @@ export function oauthToSchema(input: OAuthClient): SchemaClient<Scope[]> {
 		skip_consent: skipConsent,
 		enable_end_session: enableEndSession,
 		reference_id: referenceId,
-		// Metadata field
-		metadata: _metadata,
+		// All other metadata
 		...rest
 	} = input;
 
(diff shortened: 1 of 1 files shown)


bauerj:
Hi @himself65,

I've seen you made some substantial changes. Is there any way I can help or do you want to finish this?

gustavovalverde:
Already fixed in https://github.com/better-auth/better-auth/pull/7232

PR #7169, opened 2026-01-07
feat(drizzle): generate drizzle schema with schema namespace

### LINKED ISSUE #6606: CLI generate command: Support for PostgreSQL custom schema (pgSchema) in Drizzle output
## Problem

When using a custom PostgreSQL schema (e.g., `auth` instead of `public`), the CLI `generate` command outputs Drizzle schema using `pgTable()`:

```typescript
export const user = pgTable("user", {...});
```

Instead of using Drizzle's `pgSchema()` for explicit schema qualification:

```typescript
const authSchema = pgSchema("auth");
export const user = authSchema.table("user", {...});
```

## Current Workaround

The docs recommend using `search_path` at connection time:

```typescript
connectionString: "postgres://...?options=-c search_path=auth"
```

This works for **runtime** queries, but the generated Drizzle schema doesn't reflect the actual database structure. This causes confusion and potential issues with `drizzle-kit` migrations.

## Expected Behavior

If Better Auth detects a custom `search_path` in the database connection (or via a CLI flag), the `generate` command should output:

```typescript
import { pgSchema } from "drizzle-orm/pg-core";

const authSchema = pgSchema("auth");

export const user = authSchema.table("user", {...});
export const session = authSchema.table("session", {...});
// etc.
```

## Proposed Solutions

1. **Auto-detect**: If the CLI config has a real database connection, detect `search_path` and generate accordingly
2. **CLI flag**: Add `--pg-schema auth` flag to explicitly specify the schema
3. **Config option**: Add `schema: "auth"` option to the Drizzle adapter config

## Environment

- Better Auth version: 1.4.x
- Database: PostgreSQL
- ORM: Drizzle

## Related Docs

- [PostgreSQL Adapter - Non-default schema](https://www.better-auth.com/docs/adapters/postgresql#use-a-non-default-schema)
- [CLI - Generate command](https://www.better-auth.com/docs/concepts/cli)

The docs mention custom schema support, but it only applies to `migrate` (Kysely) and runtime queries, not the `generate` command output for Drizzle users.

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/6606

## Custom Schema namespace

If you're using PostgreSQL and you want to generate the schema with a custom schema namespace,
you can pass the `schemaName` option to the Drizzle adapter.

```ts
    provider: "pg",
    schemaName: "auth", // [!code highlight]
  }),
});
```

Then when using the Better Auth CLI, it will generate the schema that looks something like this:

```bash
npx @better-auth/cli@latest generate
```

```ts
export const authSchema = pgSchema("auth");

export const user = authSchema.table("user", {...});
export const session = authSchema.table("session", {...});
```

---
## Summary by cubic
Add schemaName support to the Drizzle adapter so the CLI can generate PostgreSQL schemas under a custom namespace. MySQL and SQLite remain unchanged.

- **New Features**
  - Added schemaName to DrizzleAdapterConfig.
  - CLI generator imports pgSchema only when schemaName is set, declares const <identifier>Schema = pgSchema("<name>"), and uses <identifier>Schema.table(...) for PostgreSQL.
  - Converts schemaName to a safe identifier (handles hyphens, non-alphanumerics, leading numbers).
  - Updated docs and added tests/snapshots for pg with/without schemaName; sqlite/mysql ignore schemaName.

- **Migration**
  - PostgreSQL: set schemaName in drizzleAdapter config and re-run the Better Auth CLI.
  - No action needed for MySQL or SQLite.

<sup>Written for commit 47c7b2eb8ce463cdbc01bd65ab3dff4968bed904. Summary will update on new commits.</sup>


himself65 (approved):

SaizFerri:
Would it be possible to pass an existing pgSchema to the config instead of just a string?
This current change seems too rigid as it implements 2 things on top of the `schemaName` specification:
* concatenate auth+Schema
* create the schema on the db schema file and export it

I think most use cases are from apps that already use a pgSchema across the app and want to reuse it for the auth. For this cases this implementation would not be usable.

alexanderniebuhr:
Is this PR also blocked by the drizzle v1 release?

Issue #7115, opened 2026-01-04
Optional peer dependency `expo-network` crashes app when not installed


marcospgp:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

  `@better-auth/expo` declares `expo-network` as an optional peer dependency, but the app crashes at runtime if it's not installed.

  ### Expected behavior

  Optional dependencies should gracefully degrade when not present, not crash the app.

  ### Actual behavior

  ```
  ERROR  [Error: Cannot find native module 'ExpoNetwork']
  ```

  ### Root cause

  In `dist/client.mjs`, there's an unconditional dynamic import:

  ```js
  import("expo-network").then(({ addNetworkStateListener }) => {
  ```

  Since it's a dynamic import, Metro still bundles the package if it exists in `node_modules` (which it does as a transitive dependency). When the import resolves, `expo-network`'s JS tries to access the native module, which isn't linked because the user never ran `expo install expo-network`.

  ### Suggested fix

  Wrap the dynamic import in a try-catch:

  ```js
  import("expo-network")
    .then(({ addNetworkStateListener }) => {
      // use it
    })
    .catch(() => {
      // expo-network not available, gracefully degrade
    });
  ```

  Or check if the native module exists before using it.

  ### Environment

  - `@better-auth/expo`: 1.4.6
  - `expo-network`: 8.0.8 (optional peer)
  - Expo SDK: 54

  ### Workaround

  Explicitly install `expo-network`:

  ```bash
  npx expo install expo-network
  ```

### Current vs. Expected behavior

see above

### What version of Better Auth are you using?

see above

### System info

```bash
see above
```

### Which area(s) are affected? (Select all that apply)

Package

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
});
```

bytaesu:
Thanks @marcospgp for letting me know

PR #7179, opened 2026-01-07
chore: make `expo-network` required and update expo docs content

SergioJuniorCE:
This single config makes the expo package not even run, by making the expo network package not optional we can start using again the expo package

- Closes https://github.com/better-auth/better-auth/issues/7115

---
## Summary by cubic
Require expo-network in packages/expo/package.json so the expo package installs and runs without missing-module errors. Update the Expo integration docs to include expo-network in the install steps and add a note for API Routes.

<sup>Written for commit 9a0afb2514b72e1008a51d012c3e1026127ab39c. Summary will update on new commits.</sup>


bytaesu:
Hi @SergioJuniorCE,

we also need to document this. I’ll make the changes in this PR!

PR #7186, opened 2026-01-07
fix(github): surface OAuth token exchange errors

Paola3stefania:
## Summary by cubic
Surface GitHub OAuth token exchange errors during authorization code validation. This prevents silent failures and maps token exchange failures to a clear invalid_code redirect.

- **Bug Fixes**
  - Build the token request with createAuthorizationCodeRequest and call the token endpoint via betterFetch.
  - On fetch errors or GitHub data.error, log and return null; the callback redirects with invalid_code.
  - Update OAuthProvider.validateAuthorizationCode to return OAuth2Tokens | null; successful flows still return tokens via getOAuth2Tokens.

<sup>Written for commit 5d9ec3e01e5da1ee09ce48bd8eac05757b719c63. Summary will update on new commits.</sup>


packages/core/src/social-providers/github.ts
```diff
@@ -86,13 +87,33 @@ export const github = (options: GithubOptions) => {
 			});
 		},
 		validateAuthorizationCode: async ({ code, codeVerifier, redirectURI }) => {
-			return validateAuthorizationCode({
+			const { body, headers: requestHeaders } = createAuthorizationCodeRequest({
```
bytaesu, line 91: Can we centralize this inside `validateAuthorizationCode`?

packages/core/src/social-providers/github.ts
```diff
@@ -86,13 +87,33 @@ export const github = (options: GithubOptions) => {
...
+				body: body,
+				headers: requestHeaders,
+			});
+
+			if (error) {
+				throw error;
+			}
+
+			if ("error" in data) {
+				throw new Error(
```
bytaesu: And can we just keep throwing the `error` while using `logger.error`?

bytaesu: I mean not restructuring the `Error` 🤔

packages/core/src/social-providers/github.ts
Paola3stefania: We can as its just for github. Keep it GitHub-specific since not all OAuth providers return errors the same way. I see this a quick fix for github for now, we need to have a whole refactor and check how providers are returning errors.

packages/core/src/social-providers/github.ts
Paola3stefania: I did this on purpouse, throwing error_description is more helpful for debugging since it contains the human-readable message.

packages/core/src/social-providers/github.ts
bytaesu: As we discussed before, we'll need to find a better approach later, but for now I get it 🙏

packages/core/src/social-providers/github.ts
```diff
@@ -86,13 +87,33 @@ export const github = (options: GithubOptions) => {
...
-				tokenEndpoint,
 			});
+
+			const { data, error } = await betterFetch<
+				| { access_token: string; token_type: string; scope: string }
+				| { error: string; error_description?: string; error_uri?: string }
+			>(tokenEndpoint, {
+				method: "POST",
+				body: body,
+				headers: requestHeaders,
```
himself65, line 104: we have `throwError: true`

packages/core/src/social-providers/github.ts
```diff
@@ -86,13 +87,33 @@ export const github = (options: GithubOptions) => {
...
+
+			if (error) {
+				throw error;
+			}
+
+			if ("error" in data) {
+				throw new Error(
+					data.error_description || data.error || "Failed to exchange code",
+				);
+			}
```
Bekacru, line 115: maybe we should log this to the server and just return null instead or something than throwing new Error

packages/core/src/social-providers/github.ts
Paola3stefania: Yeah.. the thing is that `throw: true` would auto-throw on HTTP errors and GitHub returns OAuth errors as 200 OK responses with an error field in the body (e.g., {"error": "bad_verification_code", "error_description": "..."}), so throw: true wouldn't catch those - we'd still need to check if ("error" in data)

packages/core/src/social-providers/github.ts
Paola3stefania: makes sense, adding it

packages/core/src/social-providers/github.ts
```diff
@@ -1,10 +1,12 @@
 import { betterFetch } from "@better-fetch/fetch";
+import { logger } from "../env";
```
himself65, line 2: I remembered this is not working for some reason in our cli package

PR #7204, opened 2026-01-08
fix(core): separate CSRF and origin checks

Paola3stefania:
## Summary by cubic
Separated CSRF protection from origin URL validation to give precise control over security and prevent misconfiguration. Flags now target one area each, with backward compatibility for existing disableOriginCheck setups.

- **Bug Fixes**
  - CSRF check now controls origin header validation and Fetch Metadata protections.
  - Origin check only validates redirect-related URLs (callbackURL, redirectTo, etc.).
  - Middleware honors skipCSRFCheck and skipOriginCheck independently.
  - disableOriginCheck also disables CSRF for backward compatibility; added deprecation warning and docs explaining the scope.
  - Added tests covering flag combinations and expected behaviors.

<sup>Written for commit 2b637daf90ae5a553c77481128dee84c6899f6fc. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/middlewares/origin-check.test.ts
+++ b/packages/better-auth/src/api/middlewares/origin-check.test.ts
@@ -858,6 +858,7 @@ describe("disableCSRFCheck and disableOriginCheck separation", async (it) => {
 			},
 		});
 
+		expect(warnFn).toHaveBeenCalledTimes(0);
 		// disableOriginCheck: true also disables CSRF for backward compatibility
 		// so this should succeed even with an untrusted origin
 		const res = await client.signIn.email({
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/api/middlewares/origin-check.ts
```diff
@@ -3,6 +3,32 @@ import { createAuthMiddleware } from "@better-auth/core/api";
...
+	);
+}
+
+/**
+ * Logs deprecation warning for users relying on coupled behavior.
+ * Only logs if user explicitly set disableOriginCheck (not test environment default).
+ */
+function logBackwardCompatWarning(ctx: GenericEndpointContext): void {
+	if (ctx.context.options.advanced?.disableOriginCheck === true) {
+		ctx.context.logger?.warn(
```
himself65: We should only emit this warning once. I already added `deprecated` utils in the core package. We can use that I think

PR #7231, opened 2026-01-09
fix(two-factor): updated backup codes respect `storeBackupCodes` option

Byte-Biscuit:
## Summary

After using a backup code, the remaining codes were always re-saved with the built-in `symmetricEncrypt`, ignoring the `storeBackupCodes` configuration. If a developer configured `"plain"` or a custom `{ encrypt, decrypt }` pair, the first successful verification would silently corrupt the storage format, making all remaining codes unusable.

## Fix

Extracted a new `encodeBackupCodes` helper that handles the 3-way storage branch (plain, encrypted, custom) in one place. Both `generateBackupCodes` and the verify endpoint now call it, so the encode logic can't drift out of sync.

Parameterized integration tests cover all three storage modes through a full generate/verify/re-verify cycle.

### DIFF
--- a/.changeset/fix-backup-codes-storage.md
+++ b/.changeset/fix-backup-codes-storage.md
@@ -1,0 +1,7 @@
+---
+"better-auth": patch
+---
+
+fix(two-factor): preserve backup codes storage format after verification
+
+After using a backup code, remaining codes are now re-saved using the same `storeBackupCodes` strategy (plain, encrypted, or custom) configured by the user. Previously, codes were always re-encrypted with the built-in symmetric encryption, breaking subsequent verifications for plain or custom storage modes.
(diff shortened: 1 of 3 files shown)


himself65 (approved):

Byte-Biscuit:
> [CLA assistant check](https://cla-assistant.io/better-auth/better-auth?pullRequest=7231) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/better-auth/better-auth?pullRequest=7231) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/better-auth/better-auth?pullRequest=7231) it.

image

The CLA link seems broken. It returns an error saying 'could not find linked item'. Please help.

gustavovalverde (approved):

PR #7249, opened 2026-01-10
Refactor: Move runWithEndpointContext to top-level wrapper

Ridhim-RR:
**Description**
This PR refactors the endpoint handler logic to improve readability and maintainability. Previously, runWithEndpointContext was being called inline for every endpoint definition, leading to code duplication.

I have introduced a withEndpointContext helper function that wraps handlers consistently. This ensures all endpoints benefit from the same context injection logic without cluttering the route definitions.

**Changes**
Added withEndpointContext: A reusable higher-order function that handles the runWithEndpointContext boilerplate.

```
/**
 * Higher-Order Function to wrap handlers with the necessary endpoint context.
 * This removes the need to manually call runWithEndpointContext inside every handler.
 */
const withEndpointContext = (handler: Handler) => {
  return async (ctx: any) => {
    return runWithEndpointContext(ctx as any, () => handler(ctx));
  };
};

// Usage is now clean and uniform
return createEndpoint(path, options, withEndpointContext(handler));
```

---
## Summary by cubic
Refactored endpoint context handling by introducing a withEndpointContext wrapper and applying it to createAuthEndpoint to remove duplication and keep route definitions clean. Also added async custom message support to the HaveIBeenPwned plugin and updated docs.

- **Refactors**
  - Added withEndpointContext(handler) to wrap handlers with runWithEndpointContext.
  - Replaced inline calls in createAuthEndpoint for consistent context injection.

- **New Features**
  - HaveIBeenPwned: customPasswordCompromisedMessage now supports async functions with a safe extractor and logging fallback.
  - Docs updated with a usage example.

<sup>Written for commit abd89327b82b0f656a0c60c95b8abc8be5c84b6d. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/src/api/index.ts
+++ b/packages/core/src/api/index.ts
@@ -39,6 +39,13 @@ type EndpointHandler<
 	R,
 > = (context: EndpointContext<Path, Options, AuthContext>) => Promise<R>;
 
+type Handler = (ctx: any) => Promise<any>;
+export const withEndpointContext = (handler: Handler) => {
+	return async (ctx: any) => {
+		return runWithEndpointContext(ctx as any, () => handler(ctx));
+	};
+};
+
 export function createAuthEndpoint<
 	Path extends string,
 	Options extends EndpointOptions,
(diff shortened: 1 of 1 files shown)


himself65:
No, this is a very low-level API, I don't want to expose it for now

Ridhim-RR:
> No, this is a very low-level API, I don't want to expose it for now

Ok, I saw the comment so implement it.

Issue #7196, opened 2026-01-08
changeEmail Vulnerability: Account Takeover via Typo in New Email Address


mohhh-ok:
The current `auth.api.changeEmail` implementation creates a critical security vulnerability that allows unintended account takeover when users make a typo in their new email address.

## Current Flow

1. User requests email change to a new address
2. Confirmation link sent to **old email**
3. User clicks old email confirmation link → **Database immediately updated** to new email
4. Verification email sent to new email address

## Vulnerability Scenario

1. User A wants to change email to `alice@example.com` but accidentally types `aliceee@example.com` (typo)
2. User A receives confirmation link in their old email
3. User A clicks the confirmation link (assuming the new email is correct)
4. **Account email is immediately changed to `aliceee@example.com` in the database**
5. The owner of `aliceee@example.com` (unknown third party) receives the verification email
6. Third party clicks verification link OR uses "Forgot Password" flow
7. **Account takeover complete** - User A is now locked out of their own account

## Additional Security Issue: No Session Validation

The verification links do not validate the user's session or require authentication:

- **Verification links can be opened in incognito/private browser windows**
- **No cookie or session check is performed**
- Anyone with access to the link can complete the verification without proving they own the account

This means:
- If someone gains access to the email (forwarded email, compromised email account, shared computer), they can complete the verification without knowing the account password
- The link itself is the only authentication factor, with no additional validation

## Why This is Critical

- Users typically don't re-verify the new email address when clicking the old email confirmation
- Typos in email addresses are extremely common
- Once the database is updated (step 4), the original owner loses control
- The third party who receives the verification email can complete the takeover
- **No session validation means links can be used by anyone, anywhere**

## Expected Behavior

The email change should follow a two-step verification process:

1. User requests email change
2. Confirmation link sent to old email
3. User clicks old email link → **Change marked as "pending" (no database update yet)**
4. Verification link sent to new email
5. User clicks new email link → **Only now update the database**
6. Invalidate all tokens to prevent reuse
7. **Verify that the user is authenticated when clicking verification links**

## Recommended Solution: OTP-based Verification

Instead of verification links, use OTP (One-Time Password) codes:

1. **User must be logged in to change email**
2. Send OTP to new email → user enters code in authenticated session
3. Update database only after OTPs verified

mohhh-ok:
@better-auth-agent Are there any other major services that complete email address changes with just a confirmation link to the new email address?

mohhh-ok:
[...]

PR #7252, opened 2026-01-10
fix: require session for `change-email-verification`

bytaesu:
> [!NOTE]
> To avoid confusion between flows, I replaced the `if` with a `switch`, allowing each case to be clearly separated, even if it introduces some code duplication. 

Closes https://github.com/better-auth/better-auth/issues/7196

---
## Summary by cubic
Require an active session to complete change-email verification to prevent third‑party verification attempts. Also adds an UNAUTHORIZED error and refactors the handler for clarity.

- **Bug Fixes**
  - Enforce session for requestType: change-email-verification; return UNAUTHORIZED when missing.
  - Added BASE_ERROR_CODES.UNAUTHORIZED ("Authentication required") and hooked into redirectOnError.
  - Test covers verification failure without a session.

- **Refactors**
  - Replaced nested conditionals with a switch on requestType, separating change-email-confirmation, change-email-verification, and legacy flow.

<sup>Written for commit 2c21d73a7232bac8c496ae983ad4c4fe4c955a22. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/email-verification.test.ts
+++ b/packages/better-auth/src/api/routes/email-verification.test.ts
@@ -248,6 +248,72 @@ describe("Email Verification", async () => {
 		);
 	});
 
+	it("should NOT allow email change verification without an active session", async () => {
+		let confirmationToken = "";
+		let verificationToken = "";
+
+		const { client, auth, signInWithTestUser } = await getTestInstance({
+			emailAndPassword: {
+				enabled: true,
+			},
+			emailVerification: {
+				async sendVerificationEmail({ token: _token }) {
+					verificationToken = _token;
+				},
+			},
+			user: {
+				changeEmail: {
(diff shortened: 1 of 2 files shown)


bytaesu:
> What about the server side? Have you considered the developer to call wrap this API on the server side?

- Devs using `updateEmailWithoutVerification: false` expect a secure two-step flow, and session validation meets that expectation.
- This endpoint is for the client flow where users interact by clicking an email link, so requiring a session is natural.

bytaesu:
Let me know if there are any side effects I might have overlooked - cc @himself65

bytaesu:
> [!NOTE]
> I’m reconsidering this in terms of the end-user experience.

bytaesu:
It might be safe, but realistically no one opens such links in the same browser when they received the email.

Issue #4490, opened 2025-09-06
[BUG] multiSession: listDeviceSessions() only returns 2 sessions when maximumSessions: 3


Maxime-RiseVerse:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Configure multiSession with maximumSessions: 3
2. Login with Account A
3. Login with Account B
4. Login with Account C (currently active)
5. Call listDeviceSessions() → Returns only 2 sessions (A and B, not the active C)
6. Switch to Account A using multiSession.setActive()
7. Call listDeviceSessions() → Still only 2 sessions (Account C disappeared)

### Current vs. Expected behavior

Current behavior:
- I have 3 accounts logged in
- listDeviceSessions() only returns 2 accounts
- The currently active account is missing from the list

Expected behavior:
- I have 3 accounts logged in
- listDeviceSessions() should return all 3 accounts
- All logged-in sessions should be visible in the list

Note: All 3 sessions exist in the database, but the API only returns 2.

Workaround: Setting maximumSessions: 4 allows 3 accounts to work correctly.

### What version of Better Auth are you using?

1.3.8

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Pro",
    "release": "10.0.26100",
    "cpuCount": 16,
    "cpuModel": "Intel(R) Core(TM) i7-10700K CPU @ 3.80GHz",
    "totalMemory": "31.88 GB",
    "freeMemory": "6.50 GB"
  },
  "node": {
    "version": "v22.14.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.2"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "15.5.2"
    },
    {
      "name": "react",
      "version": "19.1.1"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.16.3"
    },
    {
      "name": "drizzle",
      "version": "^0.44.5"
    }
  ],
  "betterAuth": {
    "version": "0.1.0",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from 'better-auth';
  import { drizzleAdapter } from
  'better-auth/adapters/drizzle';
  import { emailOTP, multiSession, organization
   } from 'better-auth/plugins';
  import { db } from '@/lib/db';
  import {
    account,
[...]

Issue #6184, opened 2025-11-22
Multi Session plugin creates multiple cookies for the same account when signing in multipe times.


at-channel:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Enabled the Multi Session plugin. Sign in with the same account multiple times without signing out.

### Current vs. Expected behavior

When signing in with the same account multiple times, I would expect the old cookie to be replaced with the new cookie for that one account.

Instead, what happens is a new cookie is created in addition to that same user's old cookie, so the user will end up having multiple cookies just for that same account.

### What version of Better Auth are you using?

1.4

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.1.0: Mon Oct 20 19:34:05 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6041",
    "release": "25.1.0",
    "cpuCount": 12,
    "cpuModel": "Apple M4 Pro",
    "totalMemory": "24.00 GB",
    "freeMemory": "0.75 GB"
  },
  "node": {
    "version": "v24.10.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.0"
  },
  "frameworks": [
    {
      "name": "vue",
      "version": "^3.5.16"
    },
    {
      "name": "nuxt",
      "version": "^4.2.1"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.16.3"
    }
  ],
  "betterAuth": {
    "version": "^1.4.0",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Client

### Auth config (if applicable)

```typescript

```

Adityakk9031:
@dosubot can you tell me the possible code part that causing this error

Adityakk9031:
@dosubot please ask anyone to review   pr and close  this issue

rahultumma-jnext:
Any update on this issue.

bytaesu:
- Closing this as a result of #7256

PR #7256, opened 2026-01-11
fix(multi-session): prevent duplicate cookies when same user signs in multiple times

Paola3stefania:
Closes https://github.com/better-auth/better-auth/issues/4490 and https://github.com/better-auth/better-auth/issues/6184

---
## Summary by cubic
Fixes duplicate multi-session cookies when the same user signs in multiple times on the same device. Old session cookies and records are removed, and max session limits are enforced; also adds refetch(queryParams) to useSession across clients.

- **Bug Fixes**
  - Delete existing multi-session cookies and sessions for the same user on sign-in, then set a single new cookie.
  - Recalculate and enforce maximumSessions after cleanup; added tests to verify cookie deletion and single-session state.

- **New Features**
  - Added refetch(queryParams) to useSession across clients with updated typings and tests.

<sup>Written for commit c93dd33bf1d9678dcfbeab3421438e1d1de95f15. Summary will update on new commits.</sup>

### DIFF
--- a/packages/scim/src/scim.test.ts
+++ b/packages/scim/src/scim.test.ts
@@ -2071,7 +2071,7 @@ describe("SCIM", () => {
 				totalResults: 2,
 				Resources: [userA, userC],
 			});
-		}, 10000);
+		});
 
 		it("should filter the list of users", async () => {
 			const { auth, getSCIMToken } = createTestInstance();


packages/scim/src/scim.test.ts
```diff
@@ -2071,7 +2071,7 @@ describe("SCIM", () => {
 				totalResults: 2,
 				Resources: [userA, userC],
 			});
-		});
+		}, 10000);
```
himself65: let's remove this here

PR #7258, opened 2026-01-11
docs(fastify): add session retrieval section

### LINKED ISSUE #7107: Docs: Add minimal server-side session retrieval example for non-framework Node servers
Hi 
First of all, thanks for the great work on Better Auth — the API and framework integrations are very clear.

While reading the docs, I noticed that server-side session retrieval is demonstrated only in framework-managed environments (Next.js, Nuxt, Astro, etc.), where request context and cookies are handled automatically.

For non-framework Node servers (e.g. Express, Fastify, NestJS), the same logic still applies, but it requires explicitly passing request headers to retrieve the session. This is straightforward for experienced backend developers, but can be confusing for users who are new to backend auth or coming from framework-heavy environments.

What’s currently clear in the docs

How to initialize Better Auth

How to use sessions on the client (useSession in React/Vue/etc.)

How server-side session access works within framework integrations

What’s missing (small gap)

A minimal reference example showing how to retrieve a session on the server when there is no automatic request context, e.g.:

passing req.headers explicitly

checking for an existing session

optionally attaching it to the request

Proposal (very minimal)

Add a short, non-opinionated example such as:

  headers: req.headers
});

if (!session) {
  return res.status(401).end();
}

Optionally followed by brief examples for:

Express middleware

Fastify preHandler

NestJS Guard

This would be clearly labeled as a reference example (not a required pattern or full middleware guide), and would complement the existing framework docs rather than replace them.

Questions

Would this kind of minimal server-side example be welcome in the docs?

If yes, would you prefer:

a single “Non-framework Node servers” section, or

brief examples under existing integrations?

Is it okay to include Express / Fastify / NestJS together, or should it start with one?

I’m happy to follow any preferred structure or scope before opening a PR.

Thanks!

nathannewyen:
## Summary

- Add "Getting the User Session" section to Fastify integration docs
- Show how to use `auth.api.getSession()` with proper header conversion
- Include reusable helper function example for header conversion

Closes #7107

---
## Summary by cubic
Adds a "Getting the User Session" section to the Fastify integration docs with examples for using auth.api.getSession and converting Fastify request headers to a standard Headers object. Addresses Linear #7107 by clarifying session retrieval in Fastify routes.

<sup>Written for commit c947ed94f0db3ede94af67d4aed0f5c6a20c16fa. Summary will update on new commits.</sup>

### DIFF
--- a/docs/content/docs/integrations/fastify.mdx
+++ b/docs/content/docs/integrations/fastify.mdx
@@ -122,52 +122,15 @@ fastify.register(fastifyCors, {
 
 ### Getting the User Session
 
-To retrieve the user's session in your Fastify routes, use the `auth.api.getSession` method. You need to convert Fastify's request headers to the standard `Headers` object format:
+To retrieve the user's session in your Fastify routes, use the `auth.api.getSession` method. Better Auth provides a `fromNodeHeaders` helper function that converts Node.js request headers to the format expected by Better Auth.
 
 ```ts title="server.ts"
+import { fromNodeHeaders } from "better-auth/node";
 import { auth } from "./auth"; // Your Better Auth instance
 
 fastify.get("/api/me", async (request, reply) => {
-  // Convert Fastify headers to standard Headers object
-  const headers = new Headers();
-  Object.entries(request.headers).forEach(([key, value]) => {
-    if (value) headers.append(key, value.toString());
-  });
-
   const session = await auth.api.getSession({
-    headers,
-  });
-
-  if (!session) {
-    return reply.status(401).send({ error: "Unauthorized" });
-  }
-
-  return reply.send(session);
-});
-```
-
-You can also create a reusable helper function to simplify header conversion:
-
-```ts title="utils.ts"
-import type { IncomingHttpHeaders } from "http";
-
-export function toHeaders(headers: IncomingHttpHeaders): Headers {
-  const result = new Headers();
-  Object.entries(headers).forEach(([key, value]) => {
-    if (value) result.append(key, value.toString());
-  });
-  return result;
-}
-```
-
-Then use it in your routes:
-
-```ts title="server.ts"
-import { toHeaders } from "./utils";
-
-fastify.get("/api/me", async (request, reply) => {
-  const session = await auth.api.getSession({
-    headers: toHeaders(request.headers),
+    headers: fromNodeHeaders(request.headers),
   });
 
   if (!session) {


docs/content/docs/integrations/fastify.mdx
```diff
@@ -119,3 +119,61 @@ fastify.register(fastifyCors, {
...
+
+```ts title="server.ts"
+import { auth } from "./auth"; // Your Better Auth instance
+
+fastify.get("/api/me", async (request, reply) => {
+  // Convert Fastify headers to standard Headers object
+  const headers = new Headers();
+  Object.entries(request.headers).forEach(([key, value]) => {
+    if (value) headers.append(key, value.toString());
+  });
```
himself65: Do we have a better way to get the header instead of Object.entries.forEach?

nathannewyen:
Thanks for the feedback! I've updated the PR to use the official `fromNodeHeaders` helper from `better-auth/node` instead of manual header conversion. This:

1. Addresses the Node 16 compatibility concern (the helper handles this internally)
2. Simplifies the code significantly
3. Aligns with the Express integration docs pattern

Let me know if there's anything else needed!

docs/content/docs/integrations/fastify.mdx
nathannewyen: Updated to use `fromNodeHeaders` from `better-auth/node` - much cleaner!

docs/content/docs/integrations/fastify.mdx
```diff
@@ -119,3 +119,61 @@ fastify.register(fastifyCors, {
...
+### Getting the User Session
+
+To retrieve the user's session in your Fastify routes, use the `auth.api.getSession` method. You need to convert Fastify's request headers to the standard `Headers` object format:
+
+```ts title="server.ts"
+import { auth } from "./auth"; // Your Better Auth instance
+
+fastify.get("/api/me", async (request, reply) => {
+  // Convert Fastify headers to standard Headers object
+  const headers = new Headers();
```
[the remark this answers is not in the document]
nathannewyen: Fixed! Now using `fromNodeHeaders` from `better-auth/node` which handles the conversion internally without relying on global `Headers`.

nathannewyen:
@himself65 I've updated the code to use `fromNodeHeaders` from `better-auth/node` as suggested. Could you please take another look when you have a chance?

Issue #6392, opened 2025-11-28
Allow explicit 'linkSocial' while disabling implicit auto-linking on Sign In


Dayql:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

I am trying to implement a strict security flow regarding OAuth account linking (OWASP recommendation).

My requirements are:
1. **Strict Sign In:** If a user tries to sign in with **any social provider** (Google, GitHub, etc.) but already has a Password account with the same email, I want to **BLOCK** the automatic login/merge. The user should be prompted to login with their password first.
2. **Manual Linking:** Once the user is authenticated (logged in via password), they should be able to explicitly link **that provider** from their settings page using `authClient.linkSocial`.

**The Conflict:**
Currently, the `accountLinking` configuration seems to control both behaviors globally, creating a deadlock:

* **Scenario A:** If I set `accountLinking: { enabled: false }`:
  * ✅ **Sign In:** It correctly blocks the implicit login (Good).
  * ❌ **Settings:** `authClient.linkSocial` fails with an error (`Unable to link account`), even though the user is explicitly requesting the link.

* **Scenario B:** If I set `accountLinking: { enabled: true }`:
  * ✅ **Settings:** `authClient.linkSocial` works perfectly.
  * ❌ **Sign In:** It automatically merges the account on sign-in if emails match, which I want to avoid for security reasons.

### Describe the solution you'd like

I believe `authClient.linkSocial` should be allowed to proceed if the user is **already authenticated**, regardless of the `accountLinking.enabled` global setting.

Alternatively, a more granular configuration would be great, for example:

```typescript
accountLinking: {
    enabled: true, 
    implicitLinking: false, // New option? To prevent auto-merge on sign-in
    trustedProviders: ["google", "github", "discord"]
}
```
### Describe alternatives you've considered

I have considered trying to intercept the sign-in flow manually using hooks to check for existing users in the database before allowing the OAuth flow to complete, but this feels like a workaround. A native configuration option to separate "implicit merging" from "explicit linking" would be much cleaner and safer.

### Additional context

Here is my current config attempting to achieve this. It fails for any provider listed:

```typescript
export const auth = betterAuth({
    accountLinking: {
        enabled: false, // I need this to prevent auto-merge
        trustedProviders: ["google", "github"] // Fails for both when attempting manual link
    }
});
```
When I call authClient.linkSocial with this config, I get the following error: "Unable to link account"

[...]

PR #7270, opened 2026-01-11
feat: add `disableImplicitLinking` to accountLinking

Paola3stefania:
Closes https://github.com/better-auth/better-auth/issues/6392

---
## Summary by cubic
Adds disableImplicitLinking to accountLinking to stop auto-linking during OAuth sign-in. When enabled, users must link providers explicitly; new user sign-up still works.

- **New Features**
  - account.accountLinking.disableImplicitLinking (default false).
  - Blocks implicit linking on sign-in, even for trusted providers or verified emails.
  - Explicit linking supported via linkSocial() while authenticated.

- **Migration**
  - Set account.accountLinking.disableImplicitLinking: true to opt in.
  - Update flows to use linkSocial() (e.g., from Settings) to link providers.
  - Expect existing users signing in with an unlinked provider to be redirected with error=account_not_linked.

<sup>Written for commit 3c421711a9eba36454b1fdf843f3eab923853ee5. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/oauth2/link-account.ts
+++ b/packages/better-auth/src/oauth2/link-account.ts
@@ -46,7 +46,7 @@ export async function handleOAuthUserInfo(
 					acc.accountId === account.accountId,
 			);
 		if (!linkedAccount) {
-      const accountLinking = c.context.options.account?.accountLinking;
+			const accountLinking = c.context.options.account?.accountLinking;
 			const trustedProviders =
 				c.context.options.account?.accountLinking?.trustedProviders;
 			const isTrustedProvider =


packages/better-auth/src/oauth2/link-account.ts
```diff
@@ -44,14 +44,15 @@ export async function handleOAuthUserInfo(
 				a.accountId === account.accountId,
 		);
 		if (!hasBeenLinked) {
-			const trustedProviders =
-				c.context.options.account?.accountLinking?.trustedProviders;
+			const accountLinking = c.context.options.account?.accountLinking;
+			const trustedProviders = accountLinking?.trustedProviders;
```
Bekacru: wouldn't this break manual account linking using `authClient.linkSocial` or `oauth2.linkAccount`

Paola3stefania: **linkSocial** uses a separate code path in **callback.ts** that doesn't go through **handleOAuthUserInfo**, so **disableImplicitLinking** only affects implicit linking during sign-in, not explicit linking

Everything looks correct, there's even a test in **link-account.test.ts** that proves explicit linking still works:

```
it("should allow explicit linkSocial when disableImplicitLinking is true", async () => {
    // Creates user with email/password
    // Calls linkSocial while authenticated
    // Verifies it works without errors
    // Confirms account was linked
});
```

PR #7295, opened 2026-01-12
feat(utils): export additional utility functions from index.ts

0-Sandy:
Exporting all utilities allows plugin developers to import the same functions used internally by Better Auth, making it easier to build plugins that integrate smoothly with the core.

---
## Summary by cubic
Exported additional utility modules from utils/index.ts (e.g., boolean, hashing, password, time, url, get-request-ip) so plugins can use the same helpers as core. This removes deep imports and improves DX.

<sup>Written for commit e54ba0fda3bc90a4a6f29a38b343639820e54bf7. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/utils/index.ts
+++ b/packages/better-auth/src/utils/index.ts
@@ -1,3 +1,4 @@
+export { generateId } from "@better-auth/core/utils";
 export * from "../oauth2/state";
 export * from "./boolean";
 export * from "./constants";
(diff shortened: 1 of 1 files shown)


himself65 (changes requested):
No, we don't want to expose these to the user. We are not lodash.

You can easily copy/paste these into your codebase.

himself65:
> No, we don't want to expose these to the user. We are not lodash.
> 
> You can easily copy/paste these into your codebase.

But it's arguable that maybe some of them are useful because it's betterAuth only helper

PR #7308, opened 2026-01-12
fix(core): remove dual module waring

himself65:
## Summary by cubic
Removed the dual-module warning and made AsyncLocalStorage shared across package copies to prevent isolated context and transaction state.

- **Bug Fixes**
  - Reuse AsyncLocalStorage via global Symbol.for for endpoint context, request state, and DB transaction.
  - Removed the “Better Auth was already imported” guard in context/index.ts.

<sup>Written for commit 927676a0fbcf0cda14bd46a7dcc96598373526b4. Summary will update on new commits.</sup>


himself65 (commented):
`globalThis[symbol]` looks not great but it's the only solution to hide the root issue

PR #7313, opened 2026-01-13
chore: improve rate limiting type definitions

himself65:
## Summary by cubic
Standardized rate limiting types and storage to make configuration safer and more flexible. Adds explicit storage and rule types, updates customRules to receive the current rule, and normalizes storage get to return null on misses.

- **Refactors**
  - Introduced and exported BetterAuthRateLimitStorage and BetterAuthRateLimitRule.
  - AuthContext.rateLimit now flattens window, max, and storage; duplicates are omitted from BetterAuthRateLimitOptions.
  - customRules can be a function (request, currentRule) returning false or a rule; async supported.
  - Unified memory/secondary storage behavior: get returns null on miss/expiry; TTL derives from window.
  - Renamed createDBStorage to createDatabaseStorageWrapper; getRateLimitStorage now returns BetterAuthRateLimitStorage.

<sup>Written for commit 69039b94e38eff2c8831cd4cfe4f98c458c77154. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -13,15 +13,17 @@ on:
 jobs:
   claude:
     if: |
-      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) ||
-      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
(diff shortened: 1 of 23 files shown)


packages/core/src/types/init-options.ts
himself65, line 1: @Claude, can you double-check if the prev/after type result is the same?

PR #7323, opened 2026-01-13
feat(adapters): add advanced.database.usePlural as fallback for adapter-specific option

Bekacru:
- Add usePlural to advanced.database config in init-options.ts
- Update adapter factory to use config.usePlural ?? options.advanced?.database?.usePlural
- Update Drizzle adapter to use the fallback for internal usePlural usages
- Update Prisma adapter to use the fallback for internal usePlural usages
- Mark adapter-specific usePlural options as deprecated with guidance to use advanced.database.usePlural instead

This allows users to set usePlural once in advanced.database and have it apply to all adapters, while still allowing adapter-specific overrides if needed

---
## Summary by cubic
Add a global advanced.database.usePlural option that acts as a fallback for adapter-specific usePlural. This unifies plural table handling across adapters while still allowing per-adapter overrides.

- **New Features**
  - Introduced advanced.database.usePlural (default false) to control plural table names across all adapters.
  - Adapter factory, Drizzle, and Prisma now resolve usePlural via adapterConfig.usePlural ?? advanced.database.usePlural.
  - Marked usePlural in Drizzle/Prisma/Kysely adapter configs as deprecated.

- **Migration**
  - Set advanced.database.usePlural in your main config.
  - Keep adapter-specific usePlural only if you need an override.
  - Plan to remove adapter-level usePlural in a future release.

<sup>Written for commit 0e80707ad855ac5435075824a093cd955977a3cd. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/src/types/init-options.ts
+++ b/packages/core/src/types/init-options.ts
@@ -270,6 +270,16 @@ export type BetterAuthAdvancedOptions = {
 				 * function.
 				 */
 				generateId?: GenerateIdFn | false | "serial" | "uuid";
+				/**
+				 * Use plural table names (e.g., "users" instead of "user").
+				 *
+				 * This option is used as a fallback for database adapters
+				 * that support plural table names. If set here, it will apply
+				 * to all adapters unless overridden in the adapter-specific config.
+				 *
+				 * @default false
+				 */
+				usePlural?: boolean;
 		  }
 		| undefined;
 	/**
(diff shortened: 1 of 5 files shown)


packages/core/src/types/init-options.ts
```diff
@@ -332,6 +270,16 @@ export type BetterAuthAdvancedOptions = {
...
+				/**
+				 * Use plural table names (e.g., "users" instead of "user").
+				 *
+				 * This option is used as a fallback for database adapters
+				 * that support plural table names. If set here, it will apply
+				 * to all adapters unless overridden in the adapter-specific config.
+				 *
+				 * @default false
+				 */
+				usePlural?: boolean;
```
bytaesu, line 282: I agree that `usePlural` should be a global option.

PR #7328, opened 2026-01-13
fix(two-factor): use parseUserOutput for consistent user response

krismach:
The two-factor verification endpoints (/two-factor/verify-totp, /two-factor/verify-otp) return a hardcoded user object with only basic fields, which excludes additional user fields configured through plugins like the admin plugin's role field.

This PR updates the two-factor verification response to use parseUserOutput(), ensuring consistency with other authentication endpoints

---
## Summary by cubic
Use parseUserOutput for two-factor verification responses so user data matches other auth endpoints. This returns all configured fields (including plugin fields like role) instead of a hardcoded subset.

<sup>Written for commit 9d04d28f4910c8a0b3872a3e7d17f2ba5d8bbc99. Summary will update on new commits.</sup>


himself65 (approved):

bytaesu:
It seems like the base branch wasn’t synced - cc @himself65 

img

PR #7329, opened 2026-01-13
fix(mongodb-adapter): use duck-typing for ObjectId validation

krismach:
The MongoDB adapter uses instanceof ObjectId to validate ObjectId values. This fails when the ObjectId originates from a different instance of the mongodb package (e.g., when mongoose bundles its own mongodb).

This PR replaces instanceof ObjectId checks with duck-typing by checking for the presence of the toHexString method.

---
## Summary by cubic
Switch ObjectId validation in the MongoDB adapter to duck-typing to avoid instanceof failures across different mongodb package instances. This improves compatibility with Mongoose and prevents invalid ID errors.

- **Bug Fixes**
  - Added isObjectIdLike helper that checks for toHexString.
  - Replaced instanceof ObjectId checks in input/output transforms and array handling.
  - Accepts ObjectId-like values directly to avoid "INVALID_ID" errors.

<sup>Written for commit b07db1cb2f468dd17b86479711eec51c931d6acb. Summary will update on new commits.</sup>

### DIFF
--- a/e2e/adapter/test/mongo-adapter/adapter.mongo-db.test.ts
+++ b/e2e/adapter/test/mongo-adapter/adapter.mongo-db.test.ts
@@ -1,6 +1,7 @@
-import { mongodbAdapter } from "@better-auth/mongo-adapter";
+import { isObjectIdLike, mongodbAdapter } from "@better-auth/mongo-adapter";
 import { testAdapter } from "@better-auth/test-utils/adapter";
 import { MongoClient, ObjectId } from "mongodb";
+import { describe, expect, it } from "vitest";
 import {
 	authFlowTestSuite,
 	joinsTestSuite,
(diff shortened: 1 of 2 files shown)


bytaesu:
Hi @krismach 🙂

This case seems similar to the fix I wrote before  https://github.com/better-auth/better-call/pull/92

Could you provide a reference showing that this improved logic has been battle-tested and can be safely used? 🙏

+) adding test case would be great!

krismach:
Hi @bytaesu :)

> Could you provide a reference showing that this improved logic has been battle-tested and can be safely used? 🙏

Sure.

The duck-typing approach for ObjectId validation is a well established pattern used by:

1. **MongoDB's Official js-bson Library**

   The official [js-bson source](https://github.com/mongodb/js-bson/blob/main/src/objectid.ts) uses duck-typing in multiple places:
```javascript
   if ('toHexString' in inputId && typeof inputId.toHexString === 'function') {
     workingId = ByteUtils.fromHex(inputId.toHexString());
   }
```

2. **connect-mongo**

   [Issue #172](https://github.com/jdesboeufs/connect-mongo/issues/172) documents the exact same problem. The npm maintainer @othiym23 explicitly states:

   > "That's not an error, and that's why instanceof checks are generally considered a bad idea in larger Node apps where a module may not have control over all its dependencies."

   The recommendation:

   > "Use duck-typing or branding instead, if you want to check the type of an object."

3. **Mongoose**

   [PR #11841](https://github.com/Automattic/mongoose/pull/11841) fixed the same issue by replacing `instanceof ObjectId` checks with `isBsonType()` which uses duck-typing internally.


> +) adding test case would be great!

---

### Test Coverage Added

| Test Case | Expected |
|-----------|----------|
| Real ObjectId instances | `true` |
| ObjectId from hex string | `true` |
| Objects with `toHexString()` method | `true` |
| Custom class with `toHexString()` | `true` |
| `null` | `false` |
| `undefined` | `false` |
| Plain strings | `false` |
| Numbers | `false` |
| Objects without `toHexString` | `false` |
| Objects with non-function `toHexString` | `false` |
| Arrays | `false` |
| Date objects | `false` |

---

PR #7331, opened 2026-01-13
refactor: return linked account in findOAuthUser

OscarCornish:
## Summary
- Query for user-account combination with a `findOne` instead of the slightly less performant & less semantic `findMany` and `find`
- `findOauthUser` now returns a `linkedAccount` when it found a match to the given parameters - this allows for the `hasBeenLinked` logic to be cleaned up (and given a more semantic name)

Key motivation here is that the values for accountId are stored case-insensitively in my database, but in the existing setup the search for the matching accountId is done outside of the db context. If we move it into the db context, then my adapter will handle this for me.

## Notes
- `linkedAccount` is optional to preserve compatibility for older adapters/mocks, I could be more aggressive with the refactoring if you're happy to break this

---
## Summary by cubic
Refactored OAuth account lookup to use a single, DB-level query and return the matched linked account, improving correctness for case-insensitive accountId matching and simplifying linking logic.

- **Refactors**
  - Use findOne with accountId + providerId (with user join) instead of findMany + filter.
  - findOAuthUser now returns linkedAccount; handleOAuthUserInfo uses it for linking, token updates, and cookies.
  - Context type updated to include linkedAccount: Account | null in findOAuthUser return.

<sup>Written for commit cde37ab21b094a988aed454a3555a5865cf49776. Summary will update on new commits.</sup>


packages/core/src/types/context.ts
```diff
@@ -113,7 +113,11 @@ export interface InternalAdapter<
 		email: string,
 		accountId: string,
 		providerId: string,
-	): Promise<{ user: User; accounts: Account[] } | null>;
+	): Promise<{
+		user: User;
+		linkedAccount?: Account;
```
himself65: @claude Please update the type to `Account | null` and fix the related code

himself65: @claude Please update the type to `Account | null` and fix the related code

himself65: I think Claude is not working well with thrid third-party developer. We need to install the claude app

Issue #7324, opened 2026-01-13
SSO: Support multiple domains per provider (enterprise multi-domain SSO)


ManuelCortes23:
## Problem

We're using Better Auth's SSO plugin for enterprise SAML authentication where a single IdP (e.g., Okta) serves multiple email domains belonging to the same organization. For example, a company might have:

- `company.com`
- `subsidiary.com` 
- `acquired-company.com`

All three domains authenticate through the same Okta instance with a single ACS URL.

## Current Behavior

The `ssoProvider` schema expects one domain per provider, with `providerId` marked as unique. 

We tried creating multiple `ssoProvider` rows with the same `providerId` but different `domain` values. This breaks because the SAML callback uses `findOne` by `providerId` only:

```javascript
// @better-auth/sso/dist/index.mjs ~line 1802
provider = await ctx.context.adapter.findOne({
  model: "ssoProvider",
  where: [{ field: "providerId", value: providerId }]
})
```

With multiple rows sharing the same `providerId`, `findOne` returns an arbitrary row (e.g., `company.com`). Then domain validation fails for users from other domains:

```javascript
// ~line 1974
const isTrustedProvider = ... || validateEmailDomain(userInfo.email, provider.domain);
// validateEmailDomain("user@subsidiary.com", "company.com") → false
```

Result: `account_not_linked` error for users from non-primary domains.

## Expected Behavior

Enterprise SSO should support multiple email domains per provider, since this is a common real-world configuration where:
- One IdP instance serves multiple corporate domains
- All domains share the same SAML configuration and ACS URL
- Users from any configured domain should authenticate successfully

## Current Workaround

Adding the provider to `trustedProviders` bypasses domain validation entirely, but this removes an important security check.

## Related Issues/PRs

- #6481 - SAML account linking requires trustedProviders (related behavior)
- #6551 - Fix for SAML trusted provider check
- #2296 - Multiple domains configuration (different issue - about trustedOrigins)

## Environment

- `better-auth`: 1.4.10
- `@better-auth/sso`: 1.4.10

ulic75:
I admit that I might be out of touch as I've not worked with better-auth in a while.  However it looks like the `authClient.signIn.sso()` method allows for specifying the domain.  Could build some logic or a db call to do some kind of domain mapping so they would all sign-in using a single domain provider?

https://www.better-auth.com/docs/plugins/sso#sign-in-with-sso

Paola3stefania:
Into this @ManuelCortes23

[...]

PR #7335, opened 2026-01-13
fix(sso): support multi-domain providers in account linking

thewh1teagle:
## Summary

Fixes 

- #7324

When multiple domains are registered for the same SSO `providerId` (e.g., `domain1.com` and `domain2.com` sharing one Okta provider), account linking would fail with `account_not_linked` errors for users from secondary domains.

## Problem

The issue occurs because domain validation only checks against a single provider row returned by `findOne`, even when multiple domains exist for the same `providerId`. This is common in enterprise scenarios where one IdP (like Okta) serves multiple email domains.

## Solution

- Added `validateEmailDomainForProvider()` to check email domain against **all** provider domains for a given `providerId`
- Extracted trust validation logic into `isProviderTrusted()` helper for better readability
- Updated both SAML callback endpoints to use the new multi-domain validation

## Changes

- Query all provider rows with `findMany` when validating domain
- Check if user email matches **any** of the registered domains
- Maintains backward compatibility for single-domain providers

## Testing

Tested manually with multi-domain Okta setup.

## Backward Compatibility

✅ Single-domain providers continue to work identically (array with 1 item)
✅ No breaking changes to API or behavior

### DIFF
--- a/packages/sso/src/routes/sso.ts
+++ b/packages/sso/src/routes/sso.ts
@@ -1,5 +1,10 @@
 import { BetterFetchError, betterFetch } from "@better-fetch/fetch";
-import type { User, Verification } from "better-auth";
+import type {
+	BetterAuthOptions,
+	DBAdapter,
+	User,
+	Verification,
+} from "better-auth";
 import {
 	createAuthorizationURL,
 	generateState,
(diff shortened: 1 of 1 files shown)


ManuelCortes23:
Bumping this, perhaps @Paola3stefania I believe you modified code related to this recently at https://github.com/better-auth/better-auth/pull/6551

PR #7346, opened 2026-01-14
feat(oauth-provider): add OAuth 2.1 compatible support for optional PKCE

OscarCornish:
Implements configurable PKCE requirements in the `oauth-provider` plugin to support OAuth 2.1 compliance while maintaining backward compatibility with legacy confidential clients.

## Problem

The new `oauth-provider` mandates PKCE for all clients, but the older `oidc-provider` (which supports optional PKCE) will be deprecated. This leaves no migration path for
applications with confidential clients that don't support PKCE.

## Solution

Implements a hybrid strategy:
- **Public clients** (native, user-agent-based, or `public: true`) ALWAYS require PKCE (OAuth 2.1 compliant)
- **Confidential clients** can opt out of PKCE via `requirePKCE: false` configuration
- **`offline_access` scope** ALWAYS requires PKCE (even for confidential clients)

---
## Summary by cubic
Make PKCE configurable for confidential clients to keep legacy OIDC apps working, while still enforcing PKCE for public clients and for offline_access. Adds optional support for the plain code challenge and reflects it in OIDC metadata.

- New Features
  - New options: requirePKCE (default true) and allowPlainCodeChallengeMethod (default false).
  - Authorize: PKCE always for public; confidential respects requirePKCE; offline_access always needs PKCE; validate S256 or plain (when enabled).
  - Token: Public must use PKCE; confidential must use PKCE or client_secret; verify S256/plain when PKCE was used.
  - Metadata: code_challenge_methods_supported includes "plain" when enabled.

- Migration
  - To support legacy confidential clients without PKCE, set requirePKCE=false (not for public clients).
  - Enable allowPlainCodeChallengeMethod=true only if you must accept plain.

<sup>Written for commit 3c5293612db92c02608dfccfb9c2d6038989a95f. Summary will update on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/oauth-provider.mdx
+++ b/docs/content/docs/plugins/oauth-provider.mdx
@@ -424,7 +424,7 @@ The Authorization Endpoint is the entry point for initiating an OAuth 2.1 author
 
 Important notes:
 - In OAuth 2.1, only `response_type: "code"` is supported.
-- `code_challenge_method: "plain"` will not be supported since this is a security vulnerability.
+- `code_challenge_method: "plain"` is not supported by default, but can be enabled via `allowPlainCodeChallengeMethod: true` configuration option.
 
 **State**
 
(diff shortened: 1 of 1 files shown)


dvanmali:
Hiya @OscarCornish 👋 I'm not familiar of a client without PKCE support nowadays. From the best-security practices perspective, I'd suggest the OAuth Client to support PKCE. For example, the [MCP Auth spec](https://modelcontextprotocol.io/specification/draft/basic/authorization#authorization-code-protection) requires it and support is widely adopted.

OscarCornish:
Hey @dvanmali, I understand this isn't a common scenario - but I am currently stuck between a rock and a hard place (and I'm sure that I won't be the only one):

I have a client which doesn't support PKCE (as it is a confidential client and implements OAuth2.0, it is not required by the spec) - I've asked this 3rd-party to introduce PKCE but they seem unwilling... I do understand it is less secure, and I'd really like to not have to use it, but I don't think I'll get a choice unfortunately.

Currently my only option is to use the oidc-provider, but that is already deprecated with the guidance to move to oauth-provider.

PR #7351, opened 2026-01-14
fix: should return dates for expiration fields

ping-maxwell:
`getAccesToken`'s return of `accessTokenExpiresAt` is a string, we should convert this to a date for users who use auth.api for conviniance.

## Warning:
This could be a breaking change.

### DIFF
--- a/packages/better-auth/src/api/routes/account.ts
+++ b/packages/better-auth/src/api/routes/account.ts
@@ -582,7 +582,7 @@ export const getAccessToken = createAuthEndpoint(
 				}
 			}
 
-			let accessTokenExpiresAt = (() => {
+			const accessTokenExpiresAt = (() => {
 				if (newTokens?.accessTokenExpiresAt) {
 					if (typeof newTokens.accessTokenExpiresAt === "string") {
 						return new Date(newTokens.accessTokenExpiresAt);


packages/better-auth/src/api/routes/account.ts
```diff
@@ -592,10 +585,11 @@ export const getAccessToken = createAuthEndpoint(
...
 					(await decryptOAuthToken(account.accessToken ?? "", ctx.context)),
-				accessTokenExpiresAt:
-					newTokens?.accessTokenExpiresAt ??
-					account.accessTokenExpiresAt ??
-					undefined,
+				accessTokenExpiresAt: newTokens?.accessTokenExpiresAt
+					? new Date(newTokens.accessTokenExpiresAt)
+					: account.accessTokenExpiresAt
+						? new Date(account.accessTokenExpiresAt)
+						: undefined,
```
Bekacru: can we have check if it's already a date object before converting it?

ping-maxwell: sure, although passing a date into `new Date()` returns the same date

ping-maxwell: done

himself65 (approved):

PR #7354, opened 2026-01-14
fix(sign-up): ensure verification email sent only when both sendOnSignUp and sendVerificationEmail are set

mrgrauel:
## Description

This PR fixes the email verification logic during sign-up to ensure verification emails are only sent when both `sendOnSignUp` and `sendVerificationEmail` are configured.

## Problem

Previously, the condition used an OR operator (`||`), which meant verification emails could be sent when either `sendOnSignUp` was true OR `requireEmailVerification` was true, even if `sendVerificationEmail` was not provided. This could lead to errors or unexpected behavior.

## Solution

Changed the condition to use AND (`&&`) to require both `sendOnSignUp` and `sendVerificationEmail` to be truthy:

```typescript
// Before:
if (
  ctx.context.options.emailVerification?.sendOnSignUp ||
  ctx.context.options.emailAndPassword.requireEmailVerification
) {
  // ...
}

// After:
if (
  ctx.context.options.emailVerification?.sendOnSignUp &&
  ctx.context.options.emailVerification?.sendVerificationEmail
) {
  // ...
}
```

## Changes

- Updated sign-up route to require both `sendOnSignUp` AND `sendVerificationEmail` before sending verification emails
- Added comprehensive tests for email verification logic
- Fixed unused spy in test (removed unused mock that was never wired into auth instance)

## Behavior

- Verification email is sent only when both `sendOnSignUp: true` AND `sendVerificationEmail` function is provided
- If either is missing/undefined, no verification email is sent during sign-up
- This ensures the email sending logic only runs when properly configured

### DIFF
--- a/packages/better-auth/src/api/routes/sign-up.test.ts
+++ b/packages/better-auth/src/api/routes/sign-up.test.ts
@@ -419,34 +419,6 @@ describe("sign-up email verification logic", async (it) => {
 		);
 	});
 
-	it("should NOT send verification email when only sendVerificationEmail is set (without sendOnSignUp)", async () => {
-		const mockSendVerificationEmail = vi.fn();
-		const { auth } = await getTestInstance(
-			{
-				emailVerification: {
-					sendOnSignUp: false,
-					sendVerificationEmail: mockSendVerificationEmail,
-				},
-				emailAndPassword: {
-					enabled: true,
-				},
-			},
-			{
-				disableTestUser: true,
-			},
-		);
-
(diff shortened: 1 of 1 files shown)


himself65 (changes requested):
I think this is intended.

mrosberghaus:
@himself65 it's nice that it's intended, but how can I as a user define that an email is not send automatically on sign-up even if `requiredEmailVerification` is enabled?

PR #7392, opened 2026-01-15
Add support for Stripe Elements (Embedded)

Jumaron:
Here is the updated PR description, adjusted to reflect the dedicated endpoints and the new client-side helper (`embedded-checkout.ts`) that handles the Stripe.js loading dynamically.

## Summary
This Pull Request adds full support for **Stripe Embedded Checkout** (`ui_mode: 'embedded'`) to the `@better-auth/stripe` plugin. 

It introduces server-side endpoints to generate embedded sessions and, significantly, a **client-side helper** that dynamically handles Stripe.js loading and mounting. This allows developers to integrate embedded payments without manually managing script tags or installing the heavy `@stripe/stripe-js` dependency.

## Changes
- **New Endpoint:** `POST /subscription/create-embedded-checkout`
  - Generates a Stripe Checkout Session with `ui_mode: 'embedded'`.
  - Returns the `clientSecret` required to mount the form on the client.
  - Handles subscription upgrades, seat adjustments, and metadata inheritance.
- **New Endpoint:** `GET /subscription/checkout-status`
  - Retrieves the status of a session (open, complete, expired) to verify payments on the return page without relying solely on webhooks.
- **New Client Helper:** `createStripeEmbeddedCheckout` (`embedded-checkout.ts`)
  - A lightweight wrapper that dynamically loads `stripe.js` only when needed.
  - Provides a clean API (`mount`, `unmount`, `destroy`) to manage the checkout lifecycle.
  - Fully typed configuration for appearance and locale.
- **Client SDK Update:**
  - Updated `stripeClient` definition to include type inference for the new routes.
  - Exported the new client helpers.
- **Tests:**
  - Added comprehensive unit tests for `embedded-checkout.ts` to ensure lifecycle management (mount/unmount) works correctly.

## Motivation
Previously, the plugin only supported the `hosted` UI mode, which forces a redirect. Embedded Checkout is the modern standard for Stripe integrations.

**Benefits:**
1.  **Better UX:** The user never leaves the application.
2.  **Zero-Dependency Client:** The new helper functions allow usage without adding `@stripe/stripe-js` to the project's `package.json`.
3.  **Higher Conversion:** Fewer redirects reduce drop-off.
4.  **Modern Security:** Handles 3D Secure and SCA authentication natively within the iframe.

### 1. Client Side (Mounting the Checkout)
Using the new built-in helper:

```typescript
import { authClient } from "@/lib/auth-client";
import { createStripeEmbeddedCheckout } from "@better-auth/stripe/client";

// 1. Initialize the helper (loads Stripe.js automatically)
const stripeCheckout = createStripeEmbeddedCheckout({
    publishableKey: "pk_test_...",
    onComplete: () => console.log("Checkout complete!"),
});

// 2. Get the Client Secret from Better Auth
const { data } = await authClient.subscription.createEmbeddedCheckout({
    plan: "pro",
    annual: true,
    returnUrl: `${window.location.origin}/return?session_id={CHECKOUT_SESSION_ID}`,
});

// 3. Mount the form
if (data) {
    await stripeCheckout.mount({
        clientSecret: data.clientSecret,
        container: "#checkout-container", // ID of your div
    });
}
```

### 2. Client Side (Return Page Verification)
```typescript
const sessionId = searchParams.get("session_id");

const { data } = await authClient.subscription.getCheckoutStatus({
    query: { sessionId }
});

if (data?.status === 'complete') {
    console.log("Payment successful for:", data.customerEmail);
}
```

## Checklist
- [x] Added `embedded-checkout.ts` helper and types.
- [x] Added new routes to `stripeClient` type definition.
- [x] Implemented Zod schemas for input validation.
- [x] Added unit tests for the embedded checkout client wrapper.
- [x] Tested with `user` and `organization` customer types.
- [x] Verified trial logic and duplicate subscription prevention.

---

Closes Nothing, this is personal.

Disclosure: Opus was cooking.

### DIFF
--- a/packages/stripe/src/client.ts
+++ b/packages/stripe/src/client.ts
@@ -31,6 +31,8 @@ export const stripeClient = <
 		pathMethods: {
 			"/subscription/billing-portal": "POST",
 			"/subscription/restore": "POST",
+			"/subscription/create-embedded-checkout": "POST",
+			"/subscription/checkout-status": "GET",
 		},
 		$ERROR_CODES: STRIPE_ERROR_CODES,
 	} satisfies BetterAuthClientPlugin;
--- a/packages/stripe/src/client.ts
+++ b/packages/stripe/src/client.ts
@@ -37,4 +37,6 @@ export const stripeClient = <
 		$ERROR_CODES: STRIPE_ERROR_CODES,
 	} satisfies BetterAuthClientPlugin;
 };
+
+export * from "./embedded-checkout";
 export * from "./error-codes";
--- a/packages/stripe/src/client.ts
+++ b/packages/stripe/src/client.ts
@@ -31,8 +31,6 @@ export const stripeClient = <
 		pathMethods: {
 			"/subscription/billing-portal": "POST",
 			"/subscription/restore": "POST",
-			"/subscription/create-embedded-checkout": "POST",
-			"/subscription/checkout-status": "GET",
 		},
 		$ERROR_CODES: STRIPE_ERROR_CODES,
 	} satisfies BetterAuthClientPlugin;
--- a/packages/stripe/src/embedded-checkout.ts
+++ b/packages/stripe/src/embedded-checkout.ts
@@ -123,7 +123,7 @@ export interface StripeEmbeddedCheckoutOptions {
 
 export interface MountOptions {
 	/**
-	 * The client secret returned from createEmbeddedCheckout endpoint
+	 * The client secret returned from upgrade endpoint with embeddedCheckout: true
 	 */
 	clientSecret: string;
 	/**
@@ -273,9 +273,10 @@ async function loadStripeJs(
  *   },
  * });
  *
- * // After getting clientSecret from createEmbeddedCheckout endpoint
- * const { data } = await authClient.subscription.createEmbeddedCheckout({
+ * // After getting clientSecret from upgrade endpoint with embeddedCheckout option
+ * const { data } = await authClient.subscription.upgrade({
  *   plan: "pro",
+ *   embeddedCheckout: true,
  *   returnUrl: `${window.location.origin}/success?session_id={CHECKOUT_SESSION_ID}`,
  * });
  *
@@ -314,7 +315,7 @@ export function createStripeEmbeddedCheckout(
 
 		if (!clientSecret) {
 			throw new Error(
-				"clientSecret is required. Get it from the createEmbeddedCheckout endpoint.",
+				"clientSecret is required. Get it from the upgrade endpoint with embeddedCheckout: true.",
 			);
 		}
 
@@ -398,8 +399,9 @@ export function createStripeEmbeddedCheckout(
  * ```ts
  * import { mountEmbeddedCheckout } from "@better-auth/stripe/client";
  *
- * const { data } = await authClient.subscription.createEmbeddedCheckout({
+ * const { data } = await authClient.subscription.upgrade({
  *   plan: "pro",
+ *   embeddedCheckout: true,
  *   returnUrl: `${window.location.origin}/success?session_id={CHECKOUT_SESSION_ID}`,
  * });
  *
--- a/packages/stripe/src/embedded-checkout.ts
+++ b/packages/stripe/src/embedded-checkout.ts
@@ -1,0 +1,436 @@
+/**
+ * Stripe Embedded Checkout Client
+ *
+ * A client-side helper that handles Stripe.js loading and embedded checkout mounting,
+ * eliminating the need to manually import @stripe/stripe-js.
+ *
+ * @example
+ * ```ts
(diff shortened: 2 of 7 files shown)


Jumaron:
@bytaesu Just a rough Idea how this could potentially look like, would like to know if something like this would make sense.

packages/stripe/src/client.ts
```diff
@@ -31,8 +31,12 @@ export const stripeClient = <
 		pathMethods: {
 			"/subscription/billing-portal": "POST",
 			"/subscription/restore": "POST",
+			"/subscription/create-embedded-checkout": "POST",
+			"/subscription/checkout-status": "GET",
```
bytaesu: `pathMethods` is not required for all endpoints

packages/stripe/src/embedded-checkout.ts
bytaesu, line 1: These are only exported in `client.ts` and aren't used anywhere in the actual auth flow 🧐
Do we need these?

packages/stripe/src/routes.ts
```diff
@@ -754,6 +754,514 @@ export const upgradeSubscription = (options: StripeOptions) => {
...
+ *   returnUrl: 'https://example.com/success?session_id={CHECKOUT_SESSION_ID}'
+ * });
+ *
+ * const checkout = await stripe.initEmbeddedCheckout({ clientSecret: data.clientSecret });
+ * checkout.mount('#checkout-container');
+ * ```
+ *
+ * @see [Stripe Embedded Checkout](https://docs.stripe.com/checkout/embedded/quickstart)
+ */
+export const createEmbeddedCheckout = (options: StripeOptions) => {
```
bytaesu: Wouldn't it be enough to just add options to the existing endpoint? 🧐

Jumaron: Good Morning,
before Investing more time into making this into a cleaner Version, would this "feature" be something you would support to see in the BetterAuth Stripe Plugin?

bytaesu: It would be nice to have for more flexibility, but we should reduce complexity and leverage existing implementations. I think there's too much overlap with `/subscription/upgrade`

Jumaron: Adressed in newer commit.

packages/stripe/src/client.ts
Jumaron: Adressed in newer commit.

packages/stripe/src/embedded-checkout.ts
Jumaron: The idea was to make a convenience wrapper so your not required to use any Stripe API Directly.
I am tbh very unsure where we wanna go with this, since based from the docs, i thought you would rather prefer this instead of users using the naked @stripe/stripe-js package.

packages/stripe/src/routes.ts
Jumaron:
```
/**
 * ### Endpoint
 *
 * POST `/subscription/create-embedded-checkout`
 *
 * Creates an embedded Stripe Checkout session that can be used with Stripe Elements.
 * Returns a `clientSecret` that you use to initialize the embedded checkout on the client.
 *
 * ### API Methods
 *
 * **server:**
 * `auth.api.createEmbeddedCheckout`
 *
 * **client:**
 * `authClient.subscription.createEmbeddedCheckout`
 *
 * ### Usage (Recommended)
 *
 * Use the built-in helper that handles Stripe.js loading automatically:
 *
 * ```typescript
 * import { authClient } from './auth-client';
 * import { createStripeEmbeddedCheckout } from '@better-auth/stripe/client';
 *
 * // Create the checkout instance once
 * const stripeCheckout = createStripeEmbeddedCheckout({
 *   publishableKey: 'pk_...',
 *   onComplete: () => console.log('Checkout submitted!'),
 * });
 *
 * // Get the client secret from the server
 * const { data } = await authClient.subscription.createEmbeddedCheckout({
 *   plan: 'pro',
 *   returnUrl: `${window.location.origin}/success?session_id={CHECKOUT_SESSION_ID}`,
 * });
 *
 * // Mount the checkout (Stripe.js loads automatically)
 * await stripeCheckout.mount({
 *   clientSecret: data.clientSecret,
 *   container: '#checkout-container',
 * });
 * ```
 *
 * ### Alternative: Manual Stripe.js
 *
 * If you need more control, you can use Stripe.js directly:
 *
 * ```typescript
 * import { loadStripe } from '@stripe/stripe-js';
 *
 * const stripe = await loadStripe('pk_...');
 * const { data } = await authClient.subscription.createEmbeddedCheckout({
 *   plan: 'pro',
 *   returnUrl: 'https://example.com/success?session_id={CHECKOUT_SESSION_ID}'
 * });
 *
 * const checkout = await stripe.initEmbeddedCheckout({ clientSecret: data.clientSecret });
 * checkout.mount('#checkout-container');
 * ```
```

PR #7407, opened 2026-01-16
fix(client): deep merge plugin actions to preserve all methods

gustavovalverde:
## Summary

Fixes a bug where client plugin actions with the same top-level key would overwrite each other instead of merging.

When multiple plugins return actions under the same namespace (e.g., `signIn`), only the last plugin's methods were accessible due to `Object.assign` performing a shallow merge.

**Before:**
```ts
// Plugin A: { signIn: { methodA: fn } }
// Plugin B: { signIn: { methodB: fn } }
// Result: { signIn: { methodB: fn } }  methodA is lost
```

**After:**
```ts
// Plugin A: { signIn: { methodA: fn } }
// Plugin B: { signIn: { methodB: fn } }
// Result: { signIn: { methodA: fn, methodB: fn } }  both preserved
```

This aligns the runtime behavior with the TypeScript types, which already use `UnionToIntersection` to merge plugin action types; implying deep merge semantics.

## Solution

This PR switches to `defu` for deep merging, to preserve all methods from all plugins.

---
## Summary by cubic
Deep-merge client plugin actions so methods aren’t lost when multiple plugins share the same namespace (e.g., signIn). Uses defu for deep merge and adds tests to confirm methods from all plugins remain available.

- **Bug Fixes**
  - Replace Object.assign with defu to deep-merge plugin actions.
  - Preserve all methods under the same top-level key, regardless of plugin order.
  - Add tests covering merge behavior and order independence.

<sup>Written for commit 98450509c160d9f2e7574053ea3872ed4be34f40. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/client/config.ts
+++ b/packages/better-auth/src/client/config.ts
@@ -3,6 +3,7 @@ import type {
 	ClientAtomListener,
 } from "@better-auth/core";
 import { createFetch } from "@better-fetch/fetch";
+import { defu } from "defu";
 import type { WritableAtom } from "nanostores";
 import { getBaseURL } from "../utils/url";
 import { redirectPlugin } from "./fetch-plugins";
@@ -62,7 +63,7 @@ export const getClientConfig = (
 	});
 	const { $sessionSignal, session } = getSessionAtom($fetch, options);
 	const plugins = options?.plugins || [];
-	const pluginsActions = {} as Record<string, any>;
+	let pluginsActions = {} as Record<string, any>;
 	const pluginsAtoms = {
 		$sessionSignal,
 		session,
@@ -124,9 +125,9 @@ export const getClientConfig = (
 
 	for (const plugin of plugins) {
 		if (plugin.getActions) {
-			Object.assign(
+			pluginsActions = defu(
+				plugin.getActions?.($fetch, $store, options) ?? {},
 				pluginsActions,
-				plugin.getActions?.($fetch, $store, options),
 			);
 		}
 	}
--- a/packages/better-auth/src/client/client.test.ts
+++ b/packages/better-auth/src/client/client.test.ts
@@ -20,7 +20,12 @@ import {
 import { createAuthClient as createReactClient } from "./react";
 import { createAuthClient as createSolidClient } from "./solid";
 import { createAuthClient as createSvelteClient } from "./svelte";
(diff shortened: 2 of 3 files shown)


packages/better-auth/src/client/config.ts
```diff
@@ -124,9 +125,9 @@ export const getClientConfig = (
...
 	for (const plugin of plugins) {
 		if (plugin.getActions) {
-			Object.assign(
+			pluginsActions = defu(
+				plugin.getActions?.($fetch, $store, options) ?? {},
 				pluginsActions,
-				plugin.getActions?.($fetch, $store, options),
 			);
 		}
 	}
```
jslno, line 133: I don't think it's ideal to deep merge each entry. Instead we should do a single merge:

```ts
const pluginsActions = defu(
	{},
	...plugins.map((p) => p.getActions?.($fetch, $store, options)),
) as Record<string, any>;
```

Since defu reads from right to left, this will also make the last plugin win for conflicting fields (expected behavior), instead of the first one (current behavior).

gustavovalverde: I tested both approaches and found that the behavior is actually the inverse of what you described.

```javascript
import { defu } from 'defu';

const pluginA = { signIn: { shared: 'A wins' } };
const pluginB = { signIn: { shared: 'B wins' } };

// Single spread approach
const single = defu({}, pluginA, pluginB);
console.log(single.signIn.shared); // "A wins" ← First plugin wins

// Iterative approach (this PR)
let iterative = {};
iterative = defu(pluginA, iterative);
iterative = defu(pluginB, iterative);
console.log(iterative.signIn.shared); // "B wins" ← Last plugin wins
```

If you look at [defu's source](https://github.com/unjs/defu/blob/main/src/defu.ts):

```javascript
function _defu(baseObject, defaults) {
  const object = Object.assign({}, defaults);  // Start with defaults
  for (const key in baseObject) {
    object[key] = value;  // baseObject OVERWRITES defaults
  }
  return object;
}
```

The **first argument** (`baseObject`) takes precedence over the **second argument** (`defaults`). Your approach  would make the **first plugin win** (as shown in the test above), which inverts the expected behavior.

jslno: oh, you're right. Thanks for pointing this out 🙏🏻, been a while since I used defu and I should've taken a look at the source code beforehand! Feel free to resolve this thread

PR #7424, opened 2026-01-16
feat(api-keys): pagination support for list-api-keys

ping-maxwell:
## Breaking Change

Note that this changes the return type/value of the endpoint.

---
## Summary by cubic
Adds pagination and sorting to the list API keys endpoint. This is a breaking change: the endpoint now returns a paginated object instead of an array.

- **New Features**
  - Query params: limit, offset, sortBy, sortDirection ("asc" | "desc").
  - Returns { apiKeys, total, limit?, offset? } for easier paging.
  - Works in both database and secondary-storage modes (in-memory sorting/pagination when needed).
  - OpenAPI docs and usage examples updated; tests cover pagination and sorting.

- **Migration**
  - Update client code to use result.apiKeys instead of the previous array response.
  - Read result.total for the full count, and result.limit/result.offset for the current page.

<sup>Written for commit 7926223b0dd17b0cd066c3188dfe4e87fd8d0a47. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/api-key/routes/list-api-keys.ts
+++ b/packages/better-auth/src/plugins/api-key/routes/list-api-keys.ts
@@ -11,6 +11,39 @@ import {
 import type { apiKeySchema } from "../schema";
 import type { PredefinedApiKeyOptions } from ".";
 
+const listApiKeysQuerySchema = z
+	.object({
+		limit: z.coerce
+			.number()
+			.int()
+			.nonnegative()
+			.meta({
+				description: "The number of API keys to return",
+			})
+			.optional(),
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/plugins/api-key/routes/list-api-keys.ts
```diff
@@ -26,6 +28,39 @@ export function listApiKeys({
 		{
 			method: "GET",
 			use: [sessionMiddleware],
+			query: z
```
himself65: move the zod schema top level

himself65 (approved):

PR #7427, opened 2026-01-16
feat(auth): add optional session middleware

0-Sandy:
## What does this PR do?

This PR introduces a new `optionalSessionMiddleware` that attaches the session
to the request context **only if it exists**, without requiring authentication.

## Why is this needed?

Some endpoints need to:
- work for unauthenticated users
- still have access to `ctx.context.session` when a session is present

The existing `sessionMiddleware` always enforces authentication, which makes
these use cases impossible without duplicating logic.

## What changes?

- Added `optionalSessionMiddleware`
- Allows endpoints to safely read `ctx.context.session` or `undefined`
- Does not throw `UNAUTHORIZED` when no session is present

---
## Summary by cubic
Add optionalSessionMiddleware to attach a session to the request context only if it exists, without requiring authentication. This lets public endpoints read ctx.context.session when available and avoids UNAUTHORIZED errors for unauthenticated users.

<sup>Written for commit 8290a9191cad183033f369e420f74521242d14d6. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/session.ts
+++ b/packages/better-auth/src/api/routes/session.ts
@@ -556,7 +556,7 @@ export const sessionMiddleware = createAuthMiddleware(async (ctx) => {
 });
 
 /**
- * The middleware attaches the session to the request context if present, without requiring authentication.
+ * The middleware attaches the user session to the request context if present, without requiring authentication.
  */
 export const optionalSessionMiddleware = createAuthMiddleware(async (ctx) => {
 	const session = await getSessionFromCtx(ctx);


0-Sandy:
@Bekacru @himself65

himself65 (changes requested):
Can you show me the use case?

0-Sandy:
This is an example of how I would use it in my plugin.

### Accepting an invite with or without an active session

Invite links are sent by email, so the user can open them **logged in or not**.  
The same endpoint must handle both cases.

```ts
export const acceptInvite = createAuthEndpoint(
  "/invite/accept",
  {
    method: "POST",
    middleware: [optionalSessionMiddleware],
  },
  async (ctx) => {
    const session = ctx.context.session;
    const invite = await getInviteFromToken(ctx.body.token);

    if (!invite) {
      throw APIError.from("BAD_REQUEST", {
        message: "Invalid or expired invite",
      });
    }

    // A) User is already logged in → apply role immediately
    if (session) {
      await addUserRole({
        userId: session.user.id,
        role: invite.role,
        organizationId: invite.organizationId,
      });

      await markInviteAsAccepted(invite.id);

      return { status: "accepted" };
    }

    // B) User is not logged in → redirect to login
    // Invite is finalized after login via a hook
    return {
      status: "login_required",
      redirectTo: `/login?callback=/invite/finalize?token=${ctx.body.token}`,
    };
  }
);
```

0-Sandy:
I'm seeing a CI failure in  
`src/plugins/organization/organization.test.ts > should respect membershipLimit when adding members to organization`.

I didn’t change this test or the organization logic directly, and the logic still looks correct to me.

Not sure yet why this test is failing.

0-Sandy:
It looks like this was just a bug, and it should be fixed now.

@himself65 @Bekacru when you have a moment, could you please take a look and let me know if everything looks good?

0-Sandy:
We could call getSessionFromCtx directly, but the middleware keeps things cleaner and consistent with the existing auth model.

It avoids repeating the same logic in each endpoint and makes the intent explicit: this endpoint supports an *optional* session, similar to how `sessionMiddleware` expresses a *required* one.

0-Sandy:
Also, I think getSessionFromCtx should remain as an internal utility, rather than being called directly from endpoints.

PR #7459, opened 2026-01-19
fix(stripe): allow re-subscribing to the same plan when subscription has expired

DIYgod:
Add periodEnd check to ALREADY_SUBSCRIBED_PLAN validation to allow users to re-subscribe to the same plan when their subscription period has ended.

---
## Summary by cubic
Allow users to re-subscribe to the same plan after their previous subscription period has ended. Updated ALREADY_SUBSCRIBED_PLAN validation to check periodEnd, so it only blocks when the active subscription is still unexpired.

<sup>Written for commit 650bf64df5d7fbb971018d5dd230a0df6a37376c. Summary will update on new commits.</sup>


packages/stripe/src/routes.ts
```diff
@@ -511,7 +511,9 @@ export const upgradeSubscription = (options: StripeOptions) => {
 				activeOrTrialingSubscription &&
 				activeOrTrialingSubscription.status === "active" &&
```
bytaesu, line 512: Do we have cases where the `status` cannot be trusted?

DIYgod: While using in https://github.com/RSSNext/Folo, we found that sometimes the status isn't updated correctly after expiration. Although the webhook is configured correctly, individual events are occasionally not received or not processed properly. We run a daily scheduled task to recheck the status, but we can't avoid this for now.

bytaesu: I'll add this as a defensive measure for now.

Could you share the specific cases where webhook sync fails? If it's something I can fix on our end, I'd like to address it 🧐

DIYgod: I still have no clue right now. The only possibility I can think of is that the webhook event was not received when the server was unstable. In other cases, I may need to add more logs for troubleshooting.

PR #7466, opened 2026-01-19
Resolve OAuth/OIDC wrapped response bug for HTTP requests

### LINKED ISSUE #7355: OAuth/OIDC token endpoints return wrapped JSON (response field) instead of spec-pure payload

### To Reproduce
1. Run Better Auth on canary (or latest) with default config.
2. Call any OAuth/OIDC token-style endpoint over HTTP (e.g. `/oauth2/token`).
3. Inspect the JSON response body.

**Current:** JSON is wrapped as `{ "response": { ...actual_oauth_fields... } }`. OAuth/OIDC clients expect `access_token`, `token_type`, `c_nonce`, etc. at the top level, so they fail (e.g., sending `Bearer undefined` or raising `invalid_token`).

**Expected:** HTTP responses for OAuth/OIDC endpoints should be spec-pure JSON with top-level fields, e.g. `{ "access_token": "...", "token_type": "bearer", ... }`.

Canary branch tip (also reproducible on 1.5.0-beta.6)

### System info
```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
  },
  "node": {
    "version": "v25.2.1",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.2"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null
  }
}
```

- [x] Backend
- [x] Package
- [x] Other

```ts

});
```

### Additional context
Root cause appears to be the endpoint pipeline returning `{ response, headers, status }` for HTTP requests, which then gets serialized directly by the router. This is fine for internal API calls but breaks spec requirements for OAuth/OIDC endpoints that must return top-level JSON fields.

**Proposed fix:** when endpoints are invoked with a real HTTP `Request` context, return a `Response` at the outermost layer (after hooks/cookies) so the JSON body is the raw `response` payload rather than wrapped. This keeps internal API behavior intact while ensuring spec-pure HTTP responses for OAuth/OIDC (and similar) endpoints.

sidd190:
# Fix OAuth/OIDC wrapped response bug for HTTP requests

**Fixes #7355**

## What changes were made and why

Modified the response handling logic in `packages/better-auth/src/api/to-auth-endpoints.ts` to distinguish between HTTP requests from the router and direct JavaScript API calls.

**Changes:**
- Added router context detection using `context._flag === "router"`
- Return proper `Response` objects for HTTP requests via `toResponse()`
- Preserve existing wrapped format for JavaScript API calls with `returnHeaders: true`

**The problem was :**
OAuth/OIDC endpoints were returning wrapped JSON responses like `{"response": {"access_token": "..."}}` instead of spec-compliant top-level fields like `{"access_token": "..."}`. This broke OAuth2/OIDC clients that expect standard response formats according to RFC specifications.

## A little context

The root cause was in the endpoint pipeline returning `{ response, headers, status }` structures for all calls, which then got serialized directly by the router for HTTP requests. While this worked fine for internal JavaScript API calls, it violated OAuth2/OIDC specifications that require top-level JSON fields.

The better-call router sets `_flag: "router"` in the context for HTTP requests, allowing us to detect and handle them differently.

**No breaking changes.** This fix maintains full backward compatibility.

## Before/After Behavior

**Before (Broken):**
```json
{"response": {"issuer": "http://localhost:3000", "token_endpoint": "..."}}
```

**After (Fixed):**
```json
{"issuer": "http://localhost:3000", "token_endpoint": "..."}
```

---
## Summary by cubic
Fixed OAuth/OIDC HTTP responses to return spec-compliant top‑level JSON by returning Response objects for OAuth/OIDC endpoints on HTTP requests. JS API behavior stays the same; no breaking changes.

- **Bug Fixes**
  - Detect OAuth/OIDC endpoints via context.path; return Response only when context.request is present.
  - Respect asResponse/returnHeaders for non‑OAuth/OIDC endpoints and non‑HTTP calls.
  - Removes wrapped {"response": {...}} payloads for token and discovery endpoints.

<sup>Written for commit 69bb9b9461ee8b079b26fcc52902181a1147c0e5. Summary will update on new commits.</sup>


packages/better-auth/src/api/to-auth-endpoints.ts
```diff
@@ -42,7 +42,9 @@ const hooksSourceWeakMap = new WeakMap<
 
 type UserInputContext = Partial<
 	InputContext<string, any> & EndpointContext<string, any>
->;
+> & {
+	_flag?: string;
+};
```
himself65: Where does this field come from?

himself65 (changes requested):

packages/better-auth/src/api/to-auth-endpoints.ts
sidd190: Updated! I mistakenly thought it was being set by the server while it was actually being left to the user as an option to be set it or not, so I updated to check by path and target specific endpoints only. Hope that fixed it.

gustavovalverde: I opened this PR which I forgot to do earlier: https://github.com/better-auth/better-auth/pull/7521
> Context https://github.com/better-auth/better-auth/issues/7355#issuecomment-3777179981

I'll let you guys decide what to do, but I wanted to share some thoughts on this approach after working on my similar fix:

- The list in this PR might need future updates.

- It also creates some constraints: I have some OIDC-compliant plugins, that would require to make changes here as it's missing endpoints (e.g.: `/oidc4vci/*`) that I'm already using. This will also be true for any future spec-compliant endpoints or external plugins.

- The fix I made is broader, and should be considered because it uses the correct semantics (and expectations): `HTTP Request in → HTTP Response out`. That's the contract, regardless of which endpoint.

sidd190: Ohh yeah, you seem like a more seasoned contributor as well, this is my first contribution here, so I really just followed through as the comments under the issue suggested.

About the broader fix, I was going to implement a similar fix which would keep all http requests from getting wrapped but I was asked to just keep it to the specific endpoints for now, to resolve the specific issue.

No worries though, I'll pick another issue to work on.

PR #7474, opened 2026-01-19
fix: handle serial and false cases in generateId

bytaesu:
## Summary by cubic
Fix generateId to respect database config: 'uuid' returns a UUID; 'serial' or false return false so the DB assigns IDs. Update API key creation to fall back when generateId returns false, and add tests for these cases.

<sup>Written for commit 5cdabbe5cf0356196fc07df7caa4ac83a322efdc. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/api-key/routes/create-api-key.ts
+++ b/packages/better-auth/src/plugins/api-key/routes/create-api-key.ts
@@ -447,7 +447,7 @@ export function createApiKey({
 				const generatedId = ctx.context.generateId({
 					model: API_KEY_TABLE_NAME,
 				});
-				const id = generatedId !== false ? generatedId : generateId();
+				const id = generatedId || generateId();
 				apiKey = {
 					...data,
 					id,


packages/better-auth/src/context/create-context.ts
```diff
@@ -159,8 +159,15 @@ export async function createAuthContext(
...
 			return (options.advanced as any).generateId({ model, size });
 		}
-		if (typeof options?.advanced?.database?.generateId === "function") {
-			return options.advanced.database.generateId({ model, size });
+		const dbGenerateId = options?.advanced?.database?.generateId;
+		if (typeof dbGenerateId === "function") {
+			return dbGenerateId({ model, size });
+		}
+		if (dbGenerateId === "uuid") {
+			return crypto.randomUUID();
```
himself65, line 167: import from better auth utils

bytaesu: Could you let me know where the UUID utils are? It looks like the codebase is just using `crypto.randomUUID()` everywhere 🧐

himself65: you are right

himself65 (approved):

PR #7500, opened 2026-01-20
chore(test): use `onTestFinished`

himself65:
## Summary by cubic
Switches test cleanup to Vitest’s onTestFinished for per-test teardown and better isolation. Removes the custom cleanup set and afterAll to avoid shared state and simplify the test utils.

<sup>Written for commit 081e6f20fc414f2adaa28f1090978db87e79cefe. Summary will update on new commits.</sup>


himself65:
I think this is not gonna work. We need to cleanup many things before doing this

Issue #7489, opened 2026-01-20
Property "returned" does not work as expected on additionaFields


mmorper0:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Add the organization plugin to your backend instance
2. Add an additional field with the property returned to false and input to true
3. Make a request to the backend, the field will never be saved

### Current vs. Expected behavior

The documentation says this property is for the field to not be returned, but in fact it also makes that, at least through API, you cannot input it. Supposedly, the input property is the one who manages that.

### What version of Better Auth are you using?

1.4.11

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#37~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 20 10:25:38 UTC 2",
    "release": "6.14.0-37-generic",
    "cpuCount": 14,
    "cpuModel": "Intel(R) Core(TM) Ultra 7 155U",
    "totalMemory": "30.86 GB",
    "freeMemory": "14.09 GB"
  },
  "node": {
    "version": "v22.0.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.5.1"
  },
  "frameworks": [
    {
      "name": "express",
      "version": "^5.1.0"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.16.3"
    },
    {
      "name": "drizzle",
      "version": "^0.44.7"
    }
  ],
  "betterAuth": {
    "version": "^1.4.11",
    "config": {
      "emailAndPassword": {
        "enabled": true,
        "disableSignUp": true
      },
      "plugins": [
        {
          "name": "open-api",
          "config": {
            "id": "open-api",
            "endpoints": {}
          }
        },
        {
          "name": "admin",
          "config": {
            "id": "admin",
            "hooks": {
              "after": [
                {}
              ]
            },
            "endpoints": {},
            "$ERROR_CODES": {
              "FAILED_TO_CREATE_USER": "Failed to create user",
              "USER_ALREADY_EXISTS": "User already exists.",
              "USER_ALREADY_EXISTS_USE_ANOTHER_EMAIL": "User already exists. Use another email.",
              "YOU_CANNOT_BAN_YOURSELF": "You cannot ban yourself",
              "YOU_ARE_NOT_ALLOWED_TO_CHANGE_USERS_ROLE": "You are not allowed to change users role",
[...]

PR #7504, opened 2026-01-20
fix(db): only exclude `returned: false` fields from output schemas

Paola3stefania:
https://github.com/better-auth/better-auth/issues/7489 fix

---
## Summary by cubic
Exclude fields marked returned:false only from output schemas (isClientSide: false). Input schemas now keep these fields, so they can be validated and saved when creating/updating records.

<sup>Written for commit b286da60db49695a251d9afa98148ce764f353cf. Summary will update on new commits.</sup>


Paola3stefania:
> can you add one test case?

Im on that :)

PR #7534, opened 2026-01-21
feat(db): store verification data in secondary storage

Paola3stefania:
## Summary by cubic
Store verification data in secondary storage by default to reduce database writes and improve performance, with an option to also persist to the database. Adds TTL handling from expiresAt and smarter fallback reads.

- **New Features**
  - verification.storeInDatabase option (default false).
  - Write verification to secondaryStorage with TTL derived from expiresAt.
  - Read verification from secondaryStorage first; fallback to DB only when storeInDatabase is true.
  - Fallback to plain identifier in secondaryStorage for migration.
  - Delete by identifier removes from secondaryStorage and, when enabled, the database.
  - Shared getTTLSeconds helper applied to session caching.

- **Migration**
  - findVerificationValue returns null (not undefined) when not found.
  - If secondary storage is enabled and storeInDatabase is false, verification data is not stored in the DB and the verification table is omitted from the schema. Enable storeInDatabase if your tooling queries the DB or you need the table.
  - deleteVerificationValue is a no-op in secondary-only mode; use deleteVerificationByIdentifier instead.

<sup>Written for commit 9b6501f4e6a1225ca899b1ec12b4c9b92199e440. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/src/db/get-tables.ts
+++ b/packages/core/src/db/get-tables.ts
@@ -50,6 +50,46 @@ export const getAuthTables = (
 	const { user, session, account, verification, ...pluginTables } =
 		pluginSchema;
 
+	const verificationTable = {
+		verification: {
+			modelName: options.verification?.modelName || "verification",
+			fields: {
+				identifier: {
+					type: "string",
+					required: true,
+					fieldName: options.verification?.fields?.identifier || "identifier",
+					index: true,
+				},
+				value: {
+					type: "string",
+					required: true,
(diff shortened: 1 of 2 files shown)


himself65 (commented):
nit: We should also update the CLI db schema generation

Paola3stefania:
> nit: We should also update the CLI db schema generation

ah yes ! Opening a new PR for schema generation

himself65:
can you rebase the code?

PR #7552, opened 2026-01-22
fix(expo): prevent duplicate listener notifications in FocusManager and OnlineManager

kimchi-developer:
## Problem

`ExpoOnlineManager.setOnline()` and `ExpoFocusManager.setFocused()` notify listeners on every call, even when the state value hasn't actually changed.

This causes issues when `expo-network` fires multiple events with the same `isInternetReachable: true` value during app initialization. Each notification triggers a session refetch via the session refresh manager, leading to:
- Unnecessary `/get-session` API calls
- Infinite re-render loops in React Native apps
- Poor performance and potential rate limiting

## Root Cause

```typescript
// Current code - always notifies
setOnline(online: boolean) {
  this.isOnline = online;
  this.listeners.forEach((listener) => listener(online)); // Called even if value unchanged
}
```

## Solution

Add state comparison before notifying listeners:

```typescript
setOnline(online: boolean) {
  if (this.isOnline === online) return; // Skip if no change
  this.isOnline = online;
  this.listeners.forEach((listener) => listener(online));
}
```

Same fix applied to `ExpoFocusManager.setFocused()`.

## Testing

Tested in a React Native Expo app where the issue was causing continuous session refetches. After the fix, listeners are only notified on actual state changes.

---
## Summary by cubic
Prevent duplicate listener notifications by only notifying on actual state changes in ExpoOnlineManager and ExpoFocusManager. This stops redundant session refetches and render loops in React Native Expo apps.

- **Bug Fixes**
  - Added isFocused state and early return in setFocused.
  - Added early return in setOnline.
  - Reduces duplicate expo-network init events and unnecessary /get-session calls.

<sup>Written for commit 90973c6502e5b3390a627c8f1587b6f97ddc8120. Summary will update on new commits.</sup>

### DIFF
--- a/packages/expo/test/expo.test.ts
+++ b/packages/expo/test/expo.test.ts
@@ -876,3 +876,47 @@ describe("expo deep link cookie injection for verify-email", async () => {
 		expect(error).toBeDefined();
 	});
 });
+
+describe("ExpoFocusManager duplicate notification prevention", () => {
+	it("should not notify listeners when setFocused is called with the same value", async () => {
+		const { setupExpoFocusManager } = await import("../src/focus-manager");
+		const focusManager = setupExpoFocusManager();
+
+		const listener = vi.fn();
+		focusManager.subscribe(listener);
+
+		focusManager.setFocused(true);
+		expect(listener).toHaveBeenCalledTimes(1);
+
+		focusManager.setFocused(true);
+		expect(listener).toHaveBeenCalledTimes(1);
+
+		focusManager.setFocused(false);
+		expect(listener).toHaveBeenCalledTimes(2);
+
+		focusManager.setFocused(false);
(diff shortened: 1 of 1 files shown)


packages/expo/src/focus-manager.ts
```diff
@@ -6,6 +6,7 @@ import { AppState } from "react-native";
 class ExpoFocusManager implements FocusManager {
 	listeners = new Set<FocusListener>();
 	subscription?: ReturnType<typeof AppState.addEventListener>;
+	isFocused = true;
```
bytaesu: Wouldn't this make it hard to distinguish the initial state from a duplicate state? How about using undefined as the initial value?

```ts
isFocused: boolean | undefined;
```

PR #7556, opened 2026-01-22
fix: centralize cookie parsing and handle Expires dates correctly

bytaesu:
## Summary by cubic
Centralized cookie utilities in the core package and removed duplicates from the Expo client. Improves Set-Cookie parsing (handles Expires commas correctly) while keeping Expo APIs compatible.

- **Refactors**
  - Moved splitSetCookieHeader to better-auth/cookies and used it in parseSetCookieHeader.
  - Updated setCookieToHeader to use parseSetCookieHeader instead of splitting on commas.
  - Removed local cookie helpers from Expo; Expo now imports and re-exports parseSetCookieHeader from better-auth/cookies for backwards compatibility.

<sup>Written for commit d3ede1f24fa668eec5a818c44bb65a0f6ef6fb8a. Summary will update on new commits.</sup>

### DIFF
--- a/packages/expo/src/client.ts
+++ b/packages/expo/src/client.ts
@@ -466,7 +466,6 @@ export const expoClient = (opts: ExpoClientOptions) => {
 	} satisfies BetterAuthClientPlugin;
 };
 
-// Re-export parseSetCookieHeader for backwards compatibility
 export { parseSetCookieHeader } from "better-auth/cookies";
 export * from "./focus-manager";
 export * from "./online-manager";


bytaesu:
-> will check manually

bytaesu:
@cubic-dev-ai, what do you think about `splitSetCookieHeader` logic

packages/expo/src/client.ts
```diff
@@ -539,5 +466,7 @@ export const expoClient = (opts: ExpoClientOptions) => {
 	} satisfies BetterAuthClientPlugin;
 };
 
+// Re-export parseSetCookieHeader for backwards compatibility
```
himself65:
```suggestion
```

PR #7576, opened 2026-01-23
fix: refresh `account_data` cookie when session is refreshed

bytaesu:
> [!NOTE]
> In stateless mode, `storeAccountCookie` is enabled by default

Related to https://discord.com/channels/1288403910284935179/1443514112205262960/1463820896316948563

---
## Summary by cubic
Refreshes the account_data cookie whenever the session cookie is refreshed to keep both in sync. Prevents stale account info, including in stateless mode where storeAccountCookie is enabled by default.

- **Bug Fixes**
  - Refresh account_data after session cookie cache updates (compact/jwt/jwe strategies).
  - Add tests for standard and stateless flows to verify cookie refresh.

- **Refactors**
  - Use getAccountCookie and setAccountCookie helpers to manage account cookie updates.
  - Centralize output field filtering in core utils (filterOutputFields) and reuse in cookies and organization plugin.

<sup>Written for commit ec6ab2bead49c6cb87f918380207ea39a7ef6f3a. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/cookies/index.ts
+++ b/packages/better-auth/src/cookies/index.ts
@@ -3,7 +3,6 @@ import type {
 	BetterAuthCookies,
 	BetterAuthOptions,
 	GenericEndpointContext,
-	SessionCookieData,
 } from "@better-auth/core";
 import { env, isProduction } from "@better-auth/core/env";
 import { BetterAuthError } from "@better-auth/core/error";
--- a/packages/better-auth/src/api/routes/session.ts
+++ b/packages/better-auth/src/api/routes/session.ts
@@ -1,7 +1,6 @@
 import type {
 	BetterAuthOptions,
 	GenericEndpointContext,
-	SessionCookieData,
 } from "@better-auth/core";
 import {
 	createAuthEndpoint,
--- a/packages/core/src/types/cookie.ts
+++ b/packages/core/src/types/cookie.ts
@@ -1,4 +1,3 @@
-import type { Session, User } from "@better-auth/core/db";
 import type { CookieOptions } from "better-call";
 
 export type BetterAuthCookie = { name: string; attributes: CookieOptions };
@@ -8,11 +7,4 @@ export type BetterAuthCookies = {
 	sessionData: BetterAuthCookie;
 	accountData: BetterAuthCookie;
 	dontRememberToken: BetterAuthCookie;
-};
-
-export type SessionCookieData = {
-	session: Session & Record<string, any>;
-	user: User & Record<string, any>;
-	updatedAt: number;
-	version?: string;
 };
--- a/packages/core/src/types/index.ts
+++ b/packages/core/src/types/index.ts
@@ -11,7 +11,6 @@ export type {
 export type {
 	BetterAuthCookie,
 	BetterAuthCookies,
-	SessionCookieData,
 } from "./cookie";
 export type * from "./helper";
 export type {


packages/better-auth/src/cookies/index.ts
```diff
@@ -118,116 +123,117 @@ export async function setCookieCache(
...
+		} else if (typeof versionConfig === "function") {
+			const result = versionConfig(session.session, session.user);
+			version = result instanceof Promise ? await result : result;
+		}
+	}
 
-		const expiresAtDate = getDate(options.maxAge || 60, "sec").getTime();
-		const strategy =
-			ctx.context.options.session?.cookieCache?.strategy || "compact";
+	const sessionData = {
```
himself65, line 150: Use a type here

`const sessionData: What`

bytaesu:
@himself65

I refactored this without changing the existing logic, but there was no type defined before. I’ll add a commit to introduce a type for cookie cache payload.

himself65:
> @himself65
> 
> I refactored this without changing the existing logic, but there was no type defined before. I’ll add a commit to introduce a type for cookie cache payload.

I will update this PR. thanks

bytaesu:
@himself65 nice 🙏

If you look at the session routes file, you’ll see that generics are all provided inline.

packages/better-auth/src/cookies/index.ts
```diff
@@ -118,116 +124,111 @@ export async function setCookieCache(
...
-				: ctx.context.authCookies.sessionData.attributes.maxAge,
-		};
+	const versionConfig = ctx.context.options.session?.cookieCache?.version;
+	let version = "1";
+	if (versionConfig) {
+		if (typeof versionConfig === "string") {
+			version = versionConfig;
+		} else if (typeof versionConfig === "function") {
+			const result = versionConfig(session.session, session.user);
+			version = result instanceof Promise ? await result : result;
```
himself65:
```suggestion
			version = isPromise(result) ? await result : result;
```

packages/better-auth/src/types/models.ts
```diff
@@ -54,6 +54,13 @@ export type InferPluginTypes<O extends BetterAuthOptions> =
 			>
 		: {};
 
+export type SessionData = {
+	session: Session & Record<string, any>;
+	user: User & Record<string, any>;
+	updatedAt: number;
+	version?: string;
+};
+
```
bytaesu: @himself65 

Could you name this something related to a cookie cache payload? It’s not the actual SessionData, but an internal object.

himself65: Yeah you are right. I'm still reviewing it

bytaesu: 🫡

packages/core/src/utils/db.ts
```diff
@@ -0,0 +1,20 @@
+import type { DBFieldAttribute } from "../db";
+
+/**
+ * Filters output data by removing fields with the `returned: false` attribute.
+ * This ensures sensitive fields are not exposed in API responses.
+ */
+export function filterOutputFields<T extends Record<string, unknown> | null>(
```
himself65, line 7: I found that we have some duplicate logics with `parseOutputData` function. I will handle that in another PR

bytaesu: @himself65 
I noticed that many repetitive logic pieces aren't centralized, making it hard to fix the root issues. Let's handle these gradually when we spot them 😄

Issue #6993, opened 2025-12-25
Session ID missing in KV storage when using secondaryStorage + storeSessionInDatabase


ethan-huo:
## Description

When using `secondaryStorage` with `storeSessionInDatabase: true`, the session stored in KV storage lacks the `id` field. This causes OAuth token exchange to fail when `@better-auth/oauth-provider` tries to query the session by ID.

## Steps to Reproduce

1. Configure better-auth with:
```typescript
secondaryStorage: {
  get: (key) => kv.get(key),
  set: (key, value, ttl) => kv.put(key, value, { expirationTtl: ttl }),
  delete: (key) => kv.delete(key),
},
session: {
  storeSessionInDatabase: true,
},
```

2. Add `@better-auth/oauth-provider` plugin
3. Complete OAuth authorization flow
4. Token exchange fails with:
```
params: [ undefined ]
Query: select ... from "session" where "session"."id" = $1
```

## Root Cause

In `packages/better-auth/src/db/with-hooks.ts:47-57`, the custom storage function (`fn`) is executed **before** the database `create` operation:

```typescript
// Step 1: Custom function executes FIRST
const customCreated = customCreateFn
  ? await customCreateFn.fn(actualData)  // actualData has NO id yet
  : null;

// Step 2: Database create executes SECOND
const created =
  !customCreateFn || customCreateFn.executeMainFn
    ? await adapter.create({ model, data: actualData })  // id generated HERE
    : customCreated;
```

This means:
1. `fn(sessionData)` stores session in KV **without** `id`
2. Database `create` generates `id`, but KV is **never updated**
3. `getSessionFromCtx()` returns KV data (cache hit) → **no `id`**
4. `oauth-provider/token.ts:694` queries `WHERE id = undefined` → **fails**

## Expected Behavior

Session stored in KV should include the `id` field after database creation.

## Suggested Fix

**Option A**: Update KV after DB create (recommended)

Modify `internal-adapter.ts:createSession` to update KV storage after database creation:

```typescript
const dbSession = await createWithHooks(data, "session", {
  executeMainFn: options.session?.storeSessionInDatabase,
});

if (secondaryStorage && dbSession.id) {
  await secondaryStorage.set(
    data.token,
    JSON.stringify({ session: dbSession, user }),
    sessionTTL,
  );
}
```

**Option B**: Pre-generate ID before storage

Generate session `id` before calling `createWithHooks`:

```typescript
const sessionId = ctx.generateId("session");
[...]

Issue #7454, opened 2026-01-18
listSessions returns sessions without `id` field when using secondaryStorage (Redis)


phuctm97:
## Description

When using `secondaryStorage` (Redis), the `authClient.listSessions()` API returns sessions without the `id` field, despite the TypeScript type definitions indicating that `id` should be present.

This appears to be related to #6993 - the root cause is likely that sessions are stored in secondary storage (KV/Redis) **before** the database generates the `id`, so the cached sessions never receive the `id` field.

## Expected Behavior

`listSessions()` should return sessions with all fields including `id`:

```typescript
const { data: sessions } = await authClient.listSessions();
// sessions[0].id should be defined
```

## Actual Behavior

Sessions are returned without the `id` field:

```typescript
const { data: sessions } = await authClient.listSessions();
console.log(sessions[0].id); // undefined
console.log(sessions[0]); // { token: "...", userId: "...", expiresAt: "...", ... } - no id field
```

TypeScript does not catch this because the type definition includes `id` as a required field.

## Steps to Reproduce

1. Configure better-auth with secondaryStorage (Redis):
```typescript
export const auth = betterAuth({
  database: drizzleAdapter(database, { provider: "pg" }),
  secondaryStorage: {
    get: (key) => redis.get(key),
    set: async (key, value, ttl) => {
      await (typeof ttl === "number"
        ? redis.set(key, value, { EX: ttl })
        : redis.set(key, value));
    },
    delete: async (key) => {
      await redis.del(key);
    },
  },
  // ... other config
});
```

2. Authenticate a user (creates a session)
3. Call `authClient.listSessions()`
4. Observe that returned sessions have no `id` field

## Root Cause Analysis

Based on #6993, the issue is in the session creation flow:

1. Session data is stored in secondaryStorage (Redis) **before** database insertion
2. Database `create` operation generates the `id`
3. Secondary storage is **never updated** with the generated `id`
4. When `listSessions` reads from cache, it gets session data without `id`

## Impact

- Cannot use session `id` for operations like `revokeSession({ id: session.id })`
- Type safety is compromised (runtime values don't match TypeScript types)
- Related to OAuth provider issues in #6993

## Environment

- better-auth: ^1.4.15
- @better-auth/cli: ^1.4.15
- Runtime: Node.js
- Secondary storage: Redis (using `redis` npm package v5.x)

## Related Issues

- #6993 - Session ID missing in KV storage when using secondaryStorage + storeSessionInDatabase
[...]

PR #7583, opened 2026-01-23
refactor: remove duplicate session id generation

himself65:
Fixes: https://github.com/better-auth/better-auth/issues/6993

## Summary by cubic
Removed duplicate session ID generation in the internal adapter. Sessions now rely on the database for IDs, header handling is simplified, and create hooks run with the created record.

- **Refactors**
  - Removed local sessionId creation; no id is set on new sessions.
  - Simplified header retrieval from the auth context; use headers only.
  - Parse session fields with options; compute ipAddress and userAgent from headers.
  - Updated with-hooks create flow: run main create first when enabled, then pass the created record to customCreateFn; its result becomes the final output.

<sup>Written for commit 900ee895ef4d7685b608ab740039c3b0d23146d2. Summary will update on new commits.</sup>

### DIFF
--- a/e2e/smoke/test/redis.spec.ts
+++ b/e2e/smoke/test/redis.spec.ts
@@ -106,6 +106,48 @@ test("Redis secondary storage integration", async (t) => {
 	);
 
 	await t.test(
+		"should have session id in Redis when `storeSessionInDatabase` is true",
+		async (t: TestContext) => {
+			const auth = betterAuth({
+				database: new DatabaseSync(":memory:"),
+				emailAndPassword: {
+					enabled: true,
+				},
+				secondaryStorage: redisStorage({
+					client: redisClient,
+					keyPrefix: `${id}|`,
+				}),
+				session: {
+					storeSessionInDatabase: true,
+				},
+			});
+
+			const { runMigrations } = await getMigrations(auth.options);
+			await runMigrations();
+
+			const { token } = await auth.api.signUpEmail({
+				body: {
+					email: "himself65@outlook.com",
+					password: "123456789",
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/db/internal-adapter.ts
```diff
@@ -269,39 +269,21 @@ export const createInternalAdapter = (
...
 				id: _,
 				...rest
 			} = override || {};
 
-			// determine session id
-			let sessionId: string | undefined;
-			const generatedId = ctx.generateId({ model: "session" });
-			if (generatedId !== false) {
-				sessionId = generatedId;
-			} else if (secondaryStorage && storeInDb === false) {
```
himself65, line 286: there's no meaning to generate the ID for session if user doesn't store in db and secondaryStorage enabled

bytaesu: This could be a regression for this issue, as the return value may differ between the DB and secondary storage. Documentation might be needed.
- https://github.com/better-auth/better-auth/issues/7454

himself65: No, this is still fixed by switching the order of the function call.

https://github.com/better-auth/better-auth/pull/7583/changes#diff-b006d07cf9e80d206bd13a2e42de381b8788c04ad5281f9b6dec0e6fbf42b542R48-R58

himself65: I will add test coverage for this case

bytaesu: 🫡

PR #7588, opened 2026-01-24
docs: prevent content overflow

bytaesu:
### Before
before

### After
after

---
## Summary by cubic
Prevent content overflow in docs by wrapping long inline code, links, and titles to avoid layout breakage on smaller screens.

- **Bug Fixes**
  - Apply overflow-wrap: anywhere to .prose code:not(pre code) and .prose a.
  - Add wrap-break-word to DocsTitle (H1) to break long words.

<sup>Written for commit 35797be1d7639902b52dbd56ed15baa272fb63ca. Summary will update on new commits.</sup>

### DIFF
--- a/docs/app/global.css
+++ b/docs/app/global.css
@@ -357,6 +357,7 @@ html:not([data-anchor-scrolling]) {
 	animation: stream-pulse 1s ease-in-out infinite;
 }
 
-code:not(pre code) {
-	@apply break-words;
+.prose code:not(pre code),
+.prose a {
+	overflow-wrap: anywhere;
 }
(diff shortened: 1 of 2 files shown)


bytaesu:
> Let's use this one?
> 
> https://github.com/shuding/react-wrap-balancer

Oh let me see 👀

bytaesu:
@himself65
I used `overflow-wrap: anywhere` because things like underscores or long inline code seems treated as a single word. But, making error codes into normal sentences would be weird 🤔

bytaesu:
e.g. https://www.better-auth.com/docs/plugins/sso#error-handling

PR #7600, opened 2026-01-25
feat: add magic-link prefix to token storage

xiaoyu2er:
Add 'magic-link:' prefix to token identifiers when storing verification values, matching the pattern used in one-time-token plugin. This helps distinguish token types in the database.

## Changes
- Add prefix when creating verification value
- Add prefix when finding verification value  
- Update tests to use prefix when looking up stored tokens

---
## Summary by cubic
Add a "magic-link:" prefix to verification token identifiers in the magic link plugin to prevent collisions and match one-time-token storage. Verification also supports old tokens without the prefix for backward compatibility.

- **Refactors**
  - Apply prefix when creating verification values.
  - Check prefixed identifier first when verifying, then fallback to non-prefixed tokens.
  - Update tests to use the prefixed identifier and cover backward compatibility.

<sup>Written for commit ffe47e9d12738db820179de6bd71ca8129d26e90. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/magic-link/index.ts
+++ b/packages/better-auth/src/plugins/magic-link/index.ts
@@ -209,7 +209,7 @@ export const magicLink = (options: MagicLinkOptions) => {
 						: generateRandomString(32, "a-z", "A-Z");
 					const storedToken = await storeToken(ctx, verificationToken);
 					await ctx.context.internalAdapter.createVerificationValue({
-						identifier: storedToken,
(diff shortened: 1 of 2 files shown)


himself65 (approved):
LGTM, please sign CLA

gustavovalverde:
Closing in favor of #8326. #8326 generalizes namespacing into the verification adapter's method shape, adding a `namespace` parameter to all four methods across 33 files; this PR's magic-link-only inline prefix is a strict subset.

PR #7604, opened 2026-01-25
fix(org): dynamic access control fails with Prisma operations

ping-maxwell:
Prisma requires a unique field in the where condition during update or delete operations, we don't have any unqiue conditions if the user is querying with a role instead of an id. this PR fixes this bug.

---
## Summary by cubic
Fixes Prisma errors in organization role access control when queries use non-unique fields. Uses *Many operations and limited lookups so get/update/delete work with role-based filters.

- **Bug Fixes**
  - Replace findOne with findMany + limit: 1; read first result.
  - Replace update/delete with updateMany/deleteMany to avoid unique-field requirement.

<sup>Written for commit 9ec2766f27439637cac31eac7e8914090b397333. Summary will update on new commits.</sup>


gustavovalverde:
Closing in favor of #6429 (merged), #8314 (merged), and #8524 (merged). The trio fixed the non-unique-field constraint at the Prisma adapter layer, making this PR's call-site workarounds in `crud-access-control.ts` redundant: #6429 switched `findOne` to `findFirst`, #8314 added the `deleteMany` fallback, and #8524 added the `updateMany` fallback via `hasRootUniqueWhereCondition`.

PR #7609, opened 2026-01-25
feat(oauth): add support for legacy OAuth clients that don't use PKCE

OscarCornish:
- **feat(oauth): Add support for non-PKCE capable clients**
- **chore(oauth): Add test for optional PKCE**
- **docs(oauth): update docs**

---
## Summary by cubic
Adds per-client PKCE configuration with an admin-only opt-out for legacy confidential clients, while keeping PKCE mandatory for public clients and any request with offline_access. Tightens authorize/token flows with explicit S256-only checks, robust PKCE validation, and clearer errors; updates registration rules, docs, and tests.

- New Features
  - Added require_pkce to client schema and adminCreateOAuthClient (default true; admin-only; ignored on non-admin/dynamic registration).
  - Authorization endpoint uses isPKCERequired, validates optional PKCE params (must be provided together), enforces S256-only, and returns specific reasons when PKCE is required (public client, offline_access, per-client).
  - Token endpoint requires either code_verifier or client_secret, applies PKCE rules per client and scope, and adds strict auth/token PKCE consistency checks with fail-fast 400/401 errors.

- Migration
  - Removed global requirePKCE; use per-client require_pkce via the admin API for legacy confidential clients.
  - PKCE remains required for public clients and any request with offline_access; only set require_pkce: false for confidential legacy clients that cannot be updated.

<sup>Written for commit 0706f9766ff5dd683563cdf1445c00f3b68093e3. Summary will update on new commits.</sup>


packages/oauth-provider/src/register.ts
```diff
@@ -251,6 +251,7 @@ export function oauthToSchema(input: OAuthClient): SchemaClient<Scope[]> {
 		disabled,
 		skip_consent: skipConsent,
 		enable_end_session: enableEndSession,
```
Paola3stefania, line 260: **Security: Dynamic registration should not allow `require_pkce: false`**

Because OAuthClient has [key: string]: unknown, a client can pass require_pkce: false through the /oauth2/register endpoint. If unauthenticated registration is enabled, anyone could register a non-PKCE client..

I feel like stripping require_pkce in registerEndpoint would be nice, so only the admin endpoint can opt out of PKCE.

packages/oauth-provider/src/utils/index.ts
```diff
@@ -425,3 +425,41 @@ export function deleteFromPrompt(query: URLSearchParams, prompt: Prompt) {
...
+ * PKCE is always required for:
+ * 1. Public clients (cannot securely store client_secret)
+ * 2. Requests with offline_access scope (refresh token security)
+ *
+ * For confidential clients without offline_access:
+ * - Uses client.requirePKCE if set (defaults to true)
+ *
+ * @internal
+ */
+export function isPKCERequired(
```
Paola3stefania, line 448: **Nit: unused `opts` parameter**

`opts` is accepted but never referenced in the function body. 

Maybe we should remove it or perhaps add a comment explaining planned future use (e.g. a future global override). If we dont do something with this, will result in dead code that may confuse devs

packages/oauth-provider/src/token.ts
```diff
@@ -609,14 +610,6 @@ async function handleAuthorizationCodeGrant(
 	const isAuthCodeWithSecret = client_id && client_secret;
 	const isAuthCodeWithPkce = client_id && code && code_verifier;
```
Paola3stefania, line 611: **Auth code gets consumed before credential check**

The removed validation used to reject requests missing both **client_secret** and **code_verifier** early — before the auth code was deleted... 
Now a malformed request will consume and delete the one-time auth code before failing.

Since either client_secret or code_verifier is always needed regardless of PKCE settings, what do you think about keeping a simple `if (!client_secret && !code_verifier)` check before `checkVerificationValue()`

```ts
if (!client_secret && !code_verifier) {
        error_description: "Either code_verifier or client_secret is required",
        error: "invalid_request",
    });
}
```

packages/oauth-provider/src/authorize.ts
```diff
@@ -218,30 +220,52 @@ export async function authorizeEndpoint(
...
+	const pkceRequired = isPKCERequired(opts, client, requestedScopes);
+
+	// Validate PKCE parameters if required
+	if (pkceRequired) {
+		if (!query.code_challenge || !query.code_challenge_method) {
+			throw ctx.redirect(
+				formatErrorURL(
+					query.redirect_uri,
+					"invalid_request",
+					"pkce is required for this client",
```
Paola3stefania: **Suggestion: distinct error for `offline_access`**

Before, requesting `offline_access` without PKCE returned `invalid_scope`. Now it returns this generic message, which doesn't explain *why* PKCE is needed when the client explicitly has `require_pkce: false`.

What aboout writing a specific message when the reason is `offline_access` like
```ts
"pkce is required when requesting offline_access scope"
```
I think this will be more useful to devs

packages/oauth-provider/src/pkce-optional.test.ts
```diff
@@ -0,0 +1,635 @@
...
+				skip_consent: true,
+				token_endpoint_auth_method: "none",
+			},
+		});
+		publicClient = pubResponse;
+	});
+
+	it("public client without PKCE should fail", async () => {
+		if (!publicClient?.client_id) {
+			throw Error("beforeAll not run properly");
```
Paola3stefania: **Suggestion: improve test assertions**

The `throw Error("beforeAll not run properly")` could be `expect(publicClient?.client_id).toBeDefined()` giving better Vitest output on failure

Consistency test  only checks `expect(tokenResponse.error).toBeDefined()` ... I think If we add something like this
```ts
expect(tokenResponse.error.error_description).toContain("code_verifier required")
```
would catch regressions in error messages too , it could be a great win.

Paola3stefania (commented):
Thank you for the hard work !! Overall its a great PR, just want to mention some things :)

Paola3stefania (changes requested):
.

packages/oauth-provider/src/pkce-optional.test.ts
OscarCornish: I had the checks in there to try and prod typescript in the right direction, but I should've (and have now) just changed the types to not be nullable

As for the error_description, my client wasn't expecting the endpoint to be able to throw an error with `error_description`? Not really related to this PR, but just flagging it

packages/oauth-provider/src/authorize.ts
OscarCornish: Agree the error description is bad, but there are a few scenarios that lead to this error (offline_access, public client, and client that requires pkce)

I've tried to map these properly so we can throw an accurate error depending on why PKCE is required

Issue #424, opened 2024-11-05
Please make the name optional on the signUp on both the client and server 


firstaxel:

firstaxel:
not all signUp as i have seen doesnt ask for the user name

itsderinger:
@firstaxel, how did you resolve this? facing the same issue.

firstaxel:
you can bypass this error by 
`
name: ""
`
the error is going to go

itsderinger:
> you can bypass this error by `name: ""` the error is going to go

I figured, but that's really counter-intuitive, hence making the field optional in the config is still the better approach, imo.

itsderinger:
@Bekacru What's your take? Would be cool to have it before v1?

chris-verclytte:
Same here, you might easily have a user flow that just ask for the email and password and provide further details afterwards to limit the UX friction.

Maybe something like : 
```js 
export const auth = betterAuth({
   user: {
    fields: {
      name: { required: false } // provide options instead of a simple string for mapping
    },
  },
})

// or 

export const auth = betterAuth({
   user: {
    additionalFields: {
      name: { required: false } // override default behavior when set in additionalFields
    },
  },
})
```

But it might even be preferable to get rid of the name in `better-auth` as you might want to split name with first_name / last_name in your DB and having name create an unnecessary column in the db.

firstaxel:
Thanks this is true please fix it

ping-maxwell:
Totally agree with this, in my case, I have a `first_name` and `last_name`. I don't need the `name` field at all.

zispidd:
@Bekacru any updates on this and "image" field?

ping-maxwell:
There are future plans to make the fields dynamic.
Not sure when this will be a thing, but it's definitely on @Bekacru radar and apart of the plan.

zispidd:
still relevant

Makakusnik:
+1 
Plugins should do bare minimum but they should be opened to any adjustments.. for example emailAndPassword plugin should be only responsible for emailAndPassword i really really dont understand need of name being in there too, and image aswell but that's optional at least..

firstaxel:
Still relevant it has been changed

firstaxel:
I mean it hasn't been changed

tomleidy:
I’m really interested in this change, too. Requiring a name field when it’s "emailAndPassword" seems very strange. Looking forward to this getting implemented.

JPustkuchen:
[...]

PR #7617, opened 2026-01-26
feat: make `name` field optional

bytaesu:
> [!NOTE]
>  This PR make the `name` field optional in the core `user` schema.
>
> - Email–password sign-ups are now allowed without a name.
> - Remove logic that masked missing names with empty strings or emails.
> - Since name is not used as an identifier, the impact on core logic is minimal.
> - This is a breaking change because name is no longer always returned.

Closes https://github.com/better-auth/better-auth/issues/424
Closes https://github.com/better-auth/better-auth/issues/5139

---
## Summary by cubic
Make the user name optional across the platform. Sign-up no longer requires a name, and all APIs, types, docs, and integrations now handle null or missing names safely.

- **Refactors**
  - User schema: name is now nullish; CLI generators and snapshots updated to make the column nullable.
  - API/OpenAPI: sign-up body accepts optional name; responses allow null; removed fallbacks to email/empty string.
  - Integrations: OAuth/social providers, Magic Link, Email OTP, OIDC/MCP, SCIM, and Stripe updated to work without a name.
  - Client types, tests, and docs updated (including OIDC claims parsing and SCIM name formatting).

<sup>Written for commit 66f62c9bda0ec8bbc3280c2dc24f435d9a2de457. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/organization/organization.test.ts
+++ b/packages/better-auth/src/plugins/organization/organization.test.ts
@@ -2379,7 +2379,7 @@ describe("Additional Fields", async () => {
 				user: {
 					id: string;
 					email: string;
-					name: string;
+					name?: string | undefined;
 					image?: string;
 				};
 				memberRequiredField: string;
@@ -2443,7 +2443,7 @@ describe("Additional Fields", async () => {
 			user: {
 				id: string;
 				email: string;
-				name: string;
+				name?: string | undefined;
 				image?: string;
 			};
 			memberRequiredField: string;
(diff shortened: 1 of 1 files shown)


Paola3stefania:
Thank you @bytaesu for this, Im going through files changed. 

One thing, we should add to the release notes something like
```
Breaking: name field is now optional on the user table

The name field on the user model is no longer required. If your database has a NOT NULL constraint on the name column, you'll need to manually alter it:


-- PostgreSQL
ALTER TABLE "user" ALTER COLUMN "name" DROP NOT NULL;

-- MySQL
ALTER TABLE `user` MODIFY COLUMN `name` VARCHAR(255) NULL;
SQLite does not enforce NOT NULL at the column level in the same way, so no action is needed for SQLite databases.

Note: The CLI migrate command does not currently detect changes to column constraints (e.g., NOT NULL → nullable). This migration must be applied manually.
```

bytaesu:
This should be introduced in v1.5 and explicitly mentioned in the release blog post!

This can be handled by `@better-auth/cli`

IMO, this isn't a breaking change at the DB level since it actually makes the database more flexible, but the API response type does change, so users need to handle that.

cc @Paola3stefania

himself65 (approved):

bytaesu:
> [!NOTE]
> We'll either find a way to do this without a breaking change or try again in a major release like v2 🙂

PR #7619, opened 2026-01-26
fix(next-js): skip session refresh in RSC context to prevent cookie desync

### LINKED ISSUE #2115: Session cookie maxAge is not refreshed

### To Reproduce

1. Have a short session maxAge and updateAge:
`session:{
        expiriesIn:60,
	updateAge:10
}`

2. When the updateAge has passed, call getSession

Currently the session in the database is refreshed and extended but the session cookie is not.
In the above example, the cookie expires after 60 seconds and you will be logged out.

Expected behaviour is that whenever the session is updated the cookie is also updated with the new expiry.

In my app I have 30 minute maxAge and 5 minute updateAge. It's expected that as you use the app your session is constantly extended until you stop using at and the app stops calling getSession. However what actually happens is the user is always logged out after 30 minutes.

1.2.5

```bash
- OS: Mac OS 15.3.1 (24D70)
- Browser: AVersion 1.89.0 (60953)
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

adamsbytes:
Closes #2115

`getSession()` in RSC would refresh the session in DB but couldn't update the cookie, causing desync. This detects RSC context via the `RSC` header and skips the refresh. Server Actions still work since they can set cookies.

---
## Summary by cubic
Skip session refresh in Next.js RSC to prevent DB/cookie desync. Server Actions still refresh since they can set cookies. Addresses #2115.

- **Bug Fixes**
  - Detect RSC via the RSC header in nextCookies and set a _skipSessionRefresh flag.
  - Respect the flag in getSession to avoid DB refresh when cookies can’t be updated.
  - Allow refresh during Server Actions when Next-Action is present.
  - Added tests for RSC skip, Server Action refresh, and normal refresh.

<sup>Written for commit 8d915c8afce05d3768d92d13cdf44e8624f1464f. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/session-api.test.ts
+++ b/packages/better-auth/src/api/routes/session-api.test.ts
@@ -1461,3 +1461,111 @@ describe("date field type consistency", async () => {
 		);
 	});
 });
+
+describe("RSC context session refresh", async () => {
+	const { nextCookies } = await import("../../integrations/next-js");
+
+	it("should skip session refresh in RSC context to prevent DB/cookie mismatch", async () => {
+		const { auth, testUser } = await getTestInstance({
+			session: {
+				updateAge: 0, // Update on every request
+			},
+			plugins: [nextCookies()],
+		});
+
+		const signInRes = await auth.api.signInEmail({
+			body: { email: testUser.email, password: testUser.password },
+			returnHeaders: true,
+		});
+
+		const headers = new Headers();
+		headers.set("cookie", signInRes.headers.getSetCookie().join("; "));
+
+		const initialSession = await auth.api.getSession({ headers });
+		expect(initialSession).not.toBeNull();
+		const initialExpiresAt = initialSession!.session.expiresAt;
+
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/api/routes/session.ts
```diff
@@ -384,8 +384,22 @@ export const getSession = <Option extends BetterAuthOptions>() =>
 					updateAge * 1000;
 				const shouldBeUpdated = sessionIsDueToBeUpdatedDate <= Date.now();
 
+				const skipSessionRefresh = (
+					ctx.context as typeof ctx.context & {
+						_skipSessionRefresh?: boolean;
+					}
+				)._skipSessionRefresh;
```
himself65, line 391: we shouldn't use this

himself65: Maybe we should use `defineRequestState`?

PR #7639, opened 2026-01-27
feat(i18n): type inference for error codes

himself65:
## Summary by cubic
Adds type-safe i18n error code translations inferred from installed plugins, improving DX and reducing missing-key errors. Also unifies error code typing in core and adjusts error string behavior.

- **New Features**
  - i18n TranslationDictionary now suggests error code keys from all registered plugins (via BetterAuthPluginRegistry).
  - Plugins can expose $ERROR_CODES typed as RawError; base error codes are registered via an internal "$internal:base" entry.
  - defineErrorCodes returns RawError objects; BetterAuthPluginErrorCodePart is exported for plugin authors.

- **Migration**
  - If you relied on error.toString() for the message, use error.message. toString() now returns the error code.

<sup>Written for commit 80778ba27df6d59a546b70b535144a5fea3c4660. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/src/error/codes.ts
+++ b/packages/core/src/error/codes.ts
@@ -1,7 +1,6 @@
 import { defineErrorCodes } from "../utils/error-codes";
 
 declare module "@better-auth/core" {
-	// biome-ignore lint/correctness/noUnusedVariables: AuthOptions and Options need to be same as declared in the module
 	interface BetterAuthPluginRegistry<AuthOptions, Options> {
 		/**
 		 * This plugin does not exist, do not use it in runtime.


packages/core/src/utils/error-codes.ts
```diff
@@ -44,21 +44,23 @@ type ValidateErrorCodes<T> = {
...
+	[K in keyof T]: RawError;
 } {
 	return Object.fromEntries(
 		Object.entries(codes).map(([key, value]) => [
 			key,
 			{
 				code: key,
 				message: value,
-				toString: () => value,
+				toString: () => key,
```
himself65, line 64: small change, since we have an error code. So the key is necessary than message

himself65:
> LGTM
> 
> Quick question, is there's a reason for widening the `defineErrorCodes` **return type** to RawError ? Im curious :)

We want to add a `.code` property to every APIError in the future (during 1.5). So the user can do i18n or catch some errors  handily

himself65:
Also wrap the result type from `{...}` to a defined type `RawError` will helps type inference and performance

PR #7650, opened 2026-01-27
fix: check jsconfig.json in getPathAliases

### LINKED ISSUE #7649: CLI: `getPathAliases` ignores `jsconfig.json`
## Description

`@better-auth/cli generate` fails on JSDoc-based projects because `getPathAliases()` only checks for `tsconfig.json`, not `jsconfig.json`. When no `tsconfig.json` is found, it returns `null` — so `addSvelteKitEnvModules` never runs and aliases like `$env/dynamic/private`, `$app/server`, `$lib/*` are never registered for jiti.

TypeScript projects work fine.

## Error

```
Couldn't read your auth config. Error: Cannot find module '$env/dynamic/private'
```

## Reproduction

1. Create a SvelteKit project with JSDoc: `npx sv create`
2. Add better-auth with drizzle
3. Run `npx @better-auth/cli generate --config src/lib/server/auth.js`
4. Fails with `Cannot find module '$env/dynamic/private'`

I'll provide a PR for this asap :+1:

jycouet:
## Summary

- Fall back to `jsconfig.json` when `tsconfig.json` is not found in `getPathAliases`
- Fixes JSDoc-based SvelteKit projects where `@better-auth/cli generate` fails because aliases (`$env/*`, `$app/*`, `$lib/*`) are never registered

Fixes #7649

---
## Summary by cubic
getPathAliases now falls back to jsconfig.json when tsconfig.json is missing. This fixes @better-auth/cli generate for JSDoc-based SvelteKit projects by correctly registering $env/*, $app/*, and $lib/* aliases.

<sup>Written for commit ff5d5e1ea96fbd7e3e928dc59df976ccfb60619a. Summary will update on new commits.</sup>

### DIFF
--- a/packages/cli/src/utils/get-config.ts
+++ b/packages/cli/src/utils/get-config.ts
@@ -120,7 +120,10 @@ function getPathAliasesRecursive(
 }
 
 function getPathAliases(cwd: string): Record<string, string> | null {
-	const tsConfigPath = path.join(cwd, "tsconfig.json");
+	let tsConfigPath = path.join(cwd, "tsconfig.json");
+	if (!fs.existsSync(tsConfigPath)) {
+		tsConfigPath = path.join(cwd, "jsconfig.json");
+	}
 	if (!fs.existsSync(tsConfigPath)) {
 		return null;
 	}


jycouet:
Do you know in what version this will be deployed?
Thank you

Issue #6853, opened 2025-12-18
[Issue] NextJS 16, Server Error


niassuh1:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1.Run `pnpm run dev`
2. Login using **Keycloak** after proper setup
3. Login successful
4. Shut down the app
5. Run the app again
6. Gets an error

## Screenshot
<img width="596" height="746" alt="Image" src="https://github.com/user-attachments/assets/aee1d7a7-4a0a-42e7-ac15-92f7cd5df1ab" />

## Logs
<img width="880" height="242" alt="Image" src="https://github.com/user-attachments/assets/506fd5ec-3ca0-4401-953e-b7c6e0e74c44" />

### Current vs. Expected behavior

It should open the "/" page correctly

### What version of Better Auth are you using?

1.4.7

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Home",
    "release": "10.0.26100",
    "cpuCount": 16,
    "cpuModel": "AMD Ryzen 7 6800H with Radeon Graphics         ",
    "totalMemory": "31.25 GB",
    "freeMemory": "6.62 GB"
  },
  "node": {
    "version": "v22.20.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.2"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.0.10"
    },
    {
      "name": "react",
      "version": "19.2.1"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "^1.4.7",
    "config": {
      "plugins": [
        {
          "name": "generic-oauth",
          "config": {
            "id": "generic-oauth",
            "endpoints": {},
            "$ERROR_CODES": {
              "INVALID_OAUTH_CONFIGURATION": "Invalid OAuth configuration",
              "TOKEN_URL_NOT_FOUND": "Invalid OAuth configuration. Token URL not found.",
              "PROVIDER_CONFIG_NOT_FOUND": "No config found for provider",
              "PROVIDER_ID_REQUIRED": "Provider ID is required",
              "INVALID_OAUTH_CONFIG": "Invalid OAuth configuration.",
              "SESSION_REQUIRED": "Session is required"
            }
          }
        },
        {
          "name": "next-cookies",
          "config": {
            "id": "next-cookies",
            "hooks": {
              "after": [
                {}
              ]
            }
          }
        }
[...]

PR #7651, opened 2026-01-27
fix: improve Headers detection with instanceof check and cross-realm fallback

bytaesu:
Related to https://github.com/better-auth/better-auth/issues/6853#issuecomment-3690299673

test

---
## Summary by cubic
Update getSessionCookie and getCookieCache to work with native Headers, Next.js ReadonlyHeaders, and cross-realm Headers-like objects. Uses an instanceof check with a duck-typing fallback to reliably read cookies and avoid false positives from prototype-level headers.

- **Bug Fixes**
  - Detect Headers with request instanceof Headers or when no headers property; otherwise use request.headers.
  - Apply the same logic to getCookieCache.
  - Added tests for direct Headers, inherited headers on prototype, and cross-realm Headers-like objects.

<sup>Written for commit ec571a48f7dad6c57a06af6bef42e806f8cb583c. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/cookies/cookies.test.ts
+++ b/packages/better-auth/src/cookies/cookies.test.ts
@@ -336,6 +336,22 @@ describe("getSessionCookie", async () => {
 		expect(cookies).toBeDefined();
 	});
 
+	it("should work with cross-realm Headers-like object (instanceof fails)", async () => {
+		const { signInWithTestUser } = await getTestInstance();
+		const { headers } = await signInWithTestUser();
+
+		// Simulate cross-realm Headers where instanceof check fails
+		// See: https://github.com/better-auth/better-auth/pull/1838
+		const crossRealmHeaders = {
+			get: (name: string) => headers.get(name),
+			has: (name: string) => headers.has(name),
+		};
+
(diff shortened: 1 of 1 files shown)


himself65 (changes requested):
Regression of https://github.com/better-auth/better-auth/pull/1838

packages/better-auth/src/cookies/index.ts
```diff
@@ -359,7 +359,7 @@ export const getSessionCookie = (
 			config.cookiePrefix = `${config.cookiePrefix}.`;
 		}
 	}
-	const headers = "headers" in request ? request.headers : request;
+	const headers = request instanceof Headers ? request : request.headers;
```
himself65: According to the next.js impl. Please do `request instanceof Headers` before this syntax will fix the issue

https://github.com/vercel/next.js/blob/02d8ff2cf3b3cdb0101981321ee7a31730d6a843/packages/next/src/server/web/spec-extension/adapters/headers.ts#L36-L37

himself65: @claude please do this

bytaesu: what are you doing claude 🤖🪓

Issue #7219, opened 2026-01-09
SSO plugin type inference issue


anzev:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Frontend
```
const authClient = createAuthClient({
  plugins: [ssoClient()],
});
```

Backend
```
const auth = betterAuth({
  plugins: [sso()],
});
```

### Current vs. Expected behavior

There's an issue with type inference when including the sso plugin in the configuration for backend and frontend.

For example, `authClient` is resolved as type `any`.
<img width="356" height="126" alt="Image" src="https://github.com/user-attachments/assets/c05163de-0ec2-460d-a560-2662bb0da8b6" />

However, if I remove the plugin from the list, the type gets resolved correctly:
<img width="903" height="301" alt="Image" src="https://github.com/user-attachments/assets/a61ec97e-d98b-462b-b1f9-cad0e3f195f0" />

Similar thing happens on the backend, where the `api` object gets infered as `any`:

```
const auth = betterAuth({
  plugins: [sso()],
});

type AuthType = typeof auth;
```
In this case `AuthType` is inferred as (note the incorrect `api: any` type):
<img width="583" height="304" alt="Image" src="https://github.com/user-attachments/assets/07fdfc0d-d7f4-4e03-b4b5-2a8fa51f0485" />

and if you remove `sso()` from the plugins, you get the expected inference of the `api` object:
<img width="786" height="303" alt="Image" src="https://github.com/user-attachments/assets/df80c2de-89d9-4394-902f-3bef5dc97174" />

If it helps, it seems that this broke in the version **1.4.4**, because in **1.4.3** the types are correctly inferred.

### What version of Better Auth are you using?

1.4.10

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:40 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6000",
    "release": "25.2.0",
    "cpuCount": 10,
    "cpuModel": "Apple M1 Pro",
    "totalMemory": "32.00 GB",
    "freeMemory": "0.20 GB"
  },
  "node": {
    "version": "v20.19.1",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.8.2"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.0.10"
    },
    {
      "name": "react",
[...]

PR #7663, opened 2026-01-28
fix(sso): remove direct `better-call` runtime dependency

bytaesu:
- Closes https://github.com/better-auth/better-auth/issues/7219

---
## Summary by cubic
Removes the direct better-call dependency from the SSO package and uses APIError from better-auth instead. This simplifies runtime dependencies and avoids shipping better-call.

- **Dependencies**
  - Removed better-call from packages/sso/package.json and pnpm-lock.
  - Switched APIError import in saml-state.ts to better-auth/api.
  - Updated knip.jsonc to stop ignoring better-call.

<sup>Written for commit d0b619e6c478341ba2433601fd28b83d5d4f234c. Summary will update on new commits.</sup>

### DIFF
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -90,8 +90,7 @@
 		"packages/sso": {
 			"ignoreDependencies": [
 				// type-only dependencies
-				"@better-auth/core",
-				"better-call"
+				"@better-auth/core"
 			]
 		},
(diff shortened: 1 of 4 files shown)


bytaesu:
-> something wrong, will open new pr

PR #7665, opened 2026-01-28
fix(oauth-provider): honor prompt=none for OIDC

NefixEstrada:
The OIDC standard allows prompt=none to be a valid value. This commit fixes hte plugin to honor the standard

---
## Summary by cubic
Accepts prompt=none in OIDC requests to match the spec and support silent authentication flows. Prevents validation errors from clients using prompt=none.

<sup>Written for commit 6c4da7ef1b996aacd98fdf7f1212ff69618c3353. Summary will update on new commits.</sup>

### DIFF
--- a/packages/oauth-provider/src/oauth.ts
+++ b/packages/oauth-provider/src/oauth.ts
@@ -341,6 +341,7 @@ export const oauthProvider = <O extends OAuthOptions<Scope[]>>(options: O) => {
 						nonce: z.string().optional(),
 						prompt: z
 							.enum([
+								"none",
 								"consent",
 								"login",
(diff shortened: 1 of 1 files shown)


himself65 (approved):

packages/oauth-provider/src/oauth.ts
himself65, line 333: nit: we should move the zod schema to top level to reduce memory usage with multiple auth instance

dvanmali:
@NefixEstrada (cc @himself65) prompt="none" is not currently supported internally, PR is incomplete will make an issue for this.

PR #7702, opened 2026-01-30
chore: add api-extractor for type signature tracking in CI

bytaesu:
> [!NOTE]
> Adding this to catch important type changes in core packages.
> reference: https://github.com/cloudflare/workers-sdk/blob/main/packages/miniflare/scripts/types.mjs#L85 

---
## Summary by cubic
Add API Extractor to track and gate public type signature changes for @better-auth/core. CI now fails if the exported API changes without updating the report.

- **Dependencies**
  - Added @microsoft/api-extractor and config for packages/core.
  - New API report file: packages/core/api-extractor.api.md.
  - CI runs pnpm api-extractor:check; turbo tasks added for check/update.

- **Migration**
  - When intentionally changing exported types, run: pnpm build && pnpm api-extractor:update.
  - Commit the updated packages/core/api-extractor.api.md.

<sup>Written for commit a6e1d4a04d650b130e0d0a9b80d2dff5a7acd9af. Summary will update on new commits.</sup>

### DIFF
--- a/.remarkignore
+++ b/.remarkignore
@@ -1,4 +1,6 @@
 packages/**/CHANGELOG.md
+**/*.api.md
+**/.temp/**
 **/dist/**
 **/.expo/**
 .expo/


bytaesu:
-> Let's discuss, and if this is useful, refine the setup and merge it 🙂

bytaesu:
> I don't think we should do this. If we want to ensure that some types are clearly defined. We can use vitest `expectTypeOf`. Adding a markdown file won't helps regression

Alright, I'll add tests for important types 🧐

PR #7705, opened 2026-01-30
fix: getAccessToken does not refresh access token in stateless setup (#7703)

### LINKED ISSUE #7703: on stateless setup, calling `getAccessToken` does not refresh the access token when needed

## Critical bug: access token never refreshes after expiry (breaks auth flow)

This has also been mentioned on Discord but didn’t get traction:
- https://discord.com/channels/1288403910284935179/1464286520776069326
- https://discord.com/channels/1288403910284935179/1461083247160328393

### Summary
This appears to be a critical bug because it breaks the authentication flow entirely.

Once the `accessToken` expires, it is **never refreshed**. Instead, calls that should transparently refresh the token fail with a `400 BAD_REQUEST`.

### Observed behavior
When the token is expired, calling `getAccessToken` throws:

```js
{
  status: 'BAD_REQUEST',
  body: [Object],
  headers: {},
  statusCode: 400,
  [cause]: [Object]
}
```

## Current issue
The `accessToken` expires and is not automatically refreshed. In practice, the token expires every time and the auth flow breaks once it does.

## Expected behavior
Calling `getAccessToken` should automatically refresh the expired `accessToken`, as described in the documentation, and return a valid token without requiring a full re-authentication flow.

1.4.10

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "release": "25.1.0",
    "cpuCount": 14,
    "cpuModel": "Apple M4 Pro",
    "totalMemory": "48.00 GB",
    "freeMemory": "0.38 GB"
  },
  "node": {
    "version": "v22.17.1",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.2"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "~16.1.4"
    },
    {
      "name": "react",
      "version": "~19.2.1"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "^1.4.10",
    "config": null
  }
}
```

Client, Backend

```typescript
  plugins: [
    genericOAuth({
      config: [
        keycloak({
            clientId: process.env.KEYCLOAK_ID ?? '',
            clientSecret: process.env.KEYCLOAK_SECRET ?? '',
            issuer: process.env.KEYCLOAK_ISSUER ?? '',
          }),
      ]
    })
  ],

  session: {
    expiresIn: 10 * 60 * 60,
    updateAge: 30 * 60,
    cookieCache: {
      enabled: true,
      strategy: 'jwe',
      maxAge: 10 * 60 * 60,
      refreshCache: true
    }
  },
});
```

### Additional context

_No response_

biniam0:
## Summary
Fixes #7703. In stateless (or cookie-cache) setups, calling `getAccessToken` when the access token was expired did not refresh the token and returned 400 BAD_REQUEST instead of a valid token.

## Changes
- **Refresh when expiry is unknown:** Attempt token refresh when we have a refresh token but no `accessTokenExpiresAt` (e.g. when the OAuth provider omits `expires_in`), so stateless and providers like Keycloak still get refreshed tokens.
- **Stateless-safe refresh:** Wrapped `updateAccount` in try/catch. When it fails (e.g. no DB or account not in this instance’s store), we still return the new tokens and update the account cookie when `storeAccountCookie` is enabled, so the client gets a valid token.
- **Cookie matching:** When resolving account from the cookie, match the request’s `accountId` to `accountData.accountId` (OAuth provider user id) instead of `accountData.id` (DB primary key), so `getAccessToken({ providerId, accountId })` works when using the account cookie.

## Testing
- All 18 tests in `packages/better-auth/src/api/routes/account.test.ts` pass, including:
  - "should get access token using accountId from listAccounts"
  - "should refresh account_data cookie in stateless mode"
- No API or config changes; existing behavior is preserved where a DB is used.

---
## Summary by cubic
Fixes getAccessToken not refreshing in stateless/cookie-cache setups and returning 400. Tokens now refresh when expired or when expiry is missing, and cookie-based account matching uses accountId.

- **Bug Fixes**
  - Refresh when a refresh token exists and the access token is expired or expiry is unknown; only when `provider.refreshAccessToken` is available.
  - Wrap `updateAccount` in try/catch; still return refreshed tokens and update the account cookie when `storeAccountCookie` is enabled, even if the DB update fails.
  - Match cookie account by `accountId` (provider user id) instead of DB id so `getAccessToken({ providerId, accountId })` works with the account cookie.
  - No API or config changes.

<sup>Written for commit f2c9b64fcefef2495b2d93313ca714049471788e. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/account.ts
+++ b/packages/better-auth/src/api/routes/account.ts
@@ -550,8 +550,8 @@ export const getAccessToken = createAuthEndpoint(
 				new Date(account.accessTokenExpiresAt).getTime() - Date.now() < 5_000;
 			if (
 				account.refreshToken &&
-				accessTokenExpired &&
-				provider.refreshAccessToken
+				provider.refreshAccessToken &&
+				(accessTokenExpired || !account.accessTokenExpiresAt)
 			) {
 				const refreshToken = await decryptOAuthToken(
 					account.refreshToken,
@@ -570,10 +570,15 @@ export const getAccessToken = createAuthEndpoint(
 				};
 				let updatedAccount: Record<string, any> | null = null;
 				if (account.id) {
-					updatedAccount = await ctx.context.internalAdapter.updateAccount(
-						account.id,
-						updatedData,
-					);
+					try {
+						updatedAccount = await ctx.context.internalAdapter.updateAccount(
+							account.id,
+							updatedData,
+						);
+					} catch {
+						// In stateless or multi-instance setups the account may not exist in this instance's store.
+						// Still update the account cookie with refreshed tokens so the client gets a valid token.
+					}
 				}
 				if (ctx.context.options.account?.storeAccountCookie) {
 					await setAccountCookie(ctx, {


packages/better-auth/src/api/routes/account.ts
```diff
@@ -568,10 +568,15 @@ export const getAccessToken = createAuthEndpoint(
...
-					);
+					try {
+						updatedAccount = await ctx.context.internalAdapter.updateAccount(
+							account.id,
+							updatedData,
+						);
+					} catch {
+						// In stateless or multi-instance setups the account may not exist in this instance's store.
+						// Still update the account cookie with refreshed tokens so the client gets a valid token.
+					}
```
bytaesu, line 581: Can we check the options instead of using try-catch?

`if (account.id && ctx.context.options.database)`

packages/better-auth/src/api/routes/account.ts
```diff
@@ -552,8 +552,8 @@ export const getAccessToken = createAuthEndpoint(
 				new Date(account.accessTokenExpiresAt).getTime() - Date.now() < 5_000;
 			if (
 				account.refreshToken &&
-				accessTokenExpired &&
-				provider.refreshAccessToken
+				provider.refreshAccessToken &&
+				(accessTokenExpired || !account.accessTokenExpiresAt)
```
bytaesu, line 554: Changing this could affect when it's triggered, which might cause other issues. If needed, it's better to fix this in a separate PR!

https://github.com/better-auth/better-auth/blob/canary/packages/core/src/oauth2/utils.ts#L14-L19

bytaesu (commented):
+) can we have a test? 🙂

biniam0:
alright let me check things out

Lookwe69:
Is there a way to merge? This is a major obstacle.

daniele-20tab:
Heads up: I opened #9351 with the change @bytaesu suggested in their review here (`if (account.id && ctx.context.options.database)` instead of try/catch), plus a regression test and the changeset to unblock the Verify Changesets check. The `accessTokenExpiresAt`-missing change from this PR is intentionally left out, per @bytaesu's request that it belongs in a separate PR.

@biniam0 you're co-authored on the commit to credit the original work here. If you'd rather pick this up yourself I'm happy to close #9351 — just let me know.

daniele-20tab:
Heads up: I opened #9351 with the change @bytaesu suggested in their review here (`if (account.id && ctx.context.options.database)` instead of try/catch), plus a regression test and the changeset to unblock the Verify Changesets check. The `accessTokenExpiresAt`-missing change from this PR is intentionally left out, per @bytaesu's request that it belongs in a separate PR.

@biniam0 you're co-authored on the commit to credit the original work here. If you'd rather pick this up yourself I'm happy to close #9351 — just let me know.

bytaesu:
@daniele-20tab Good!

It looks like biniam0 is not currently active, and since biniam0 is listed as a co-author on #9351, I'll continue the work in that PR. Thanks for working on this @biniam0 🫡

Issue #6609, opened 2025-12-08
SSO SAML AuthnRequest Empty AssertionConsumerServiceURL


ColaIan:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Set ACS URL as callbackUrl
```ts
		sso({
			defaultSSO: [
				{
					domain: 'localhost:5173', // Your domain
					providerId: 'pydrox-dev',
					samlConfig: {
						// SP Configuration
						issuer: 'http://localhost:5173/api/auth/sso/saml2/sp/metadata',
						entryPoint:
							'https://auth.pydrox.com/application/saml/development-better-auth-5173/sso/binding/redirect/',
						callbackUrl: 'http://localhost:5173/api/auth/sso/saml2/sp/acs/pydrox-dev',
						...
```
Attempt to sign in with that SAML SSO provider

### Current vs. Expected behavior

After base64decoding and xml deflating the SAMLRequest param, AssertionConsumerServiceURL shows empty string

<img width="825" height="172" alt="Image" src="https://github.com/user-attachments/assets/2cdcf58f-30e4-4844-b313-89b296321e23" />

Expected value of AssertionConsumerServiceURL should be the value of callbackURL

### What version of Better Auth are you using?

1.4.5

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "arm64",
    "version": "#17~24.04.1-Ubuntu SMP Tue Nov  4 16:40:23 UTC 2025",
    "release": "6.14.0-1017-oracle",
    "cpuCount": 4,
    "cpuModel": "Neoverse-N1",
    "totalMemory": "23.42 GB",
    "freeMemory": "10.68 GB"
  },
  "node": {
    "version": "v24.7.0",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.4"
  },
  "frameworks": [
    {
      "name": "svelte",
      "version": "^5.45.6"
    },
    {
      "name": "@sveltejs/kit",
      "version": "^2.49.1"
    }
  ],
  "databases": [
    {
      "name": "@prisma/client",
      "version": "^7.1.0"
    }
  ],
  "betterAuth": {
    "version": "^1.4.5",
    "config": {
      "experimental": {
        "joins": true
      },
      "plugins": [
        {
          "name": "organization",
          "config": {
            "id": "organization",
            "endpoints": {},
            "schema": {
              "organization": {
                "fields": {
                  "name": {
                    "type": "string",
[...]

PR #7708, opened 2026-01-30
fix(sso): correct IdentityProvider configuration in signInSSO

theNailz:
## Problem

The `signInSSO` endpoint incorrectly constructs the samlify `IdentityProvider`, causing:

- Azure AD (and other IdPs) to reject the SAML AuthnRequest with:
  `AADSTS7500511: XML attribute 'AssertionConsumerServiceURL' in the SAML message must be a URI`

## Root Cause

In `signInSSO`, the `IdentityProvider` was constructed with:

1. Wrong field name: `encryptCert` instead of `signingCert` - samlify ignores `encryptCert` for signing
2. Missing fallback: `entityID` doesn't fall back to `parsedSamlConfig.issuer`
3. Missing fallback: `singleSignOnService` doesn't fall back to `idpData?.singleSignOnService` or `parsedSamlConfig.entryPoint`
4. Wrong config: `wantAuthnRequestsSigned` used `wantAssertionsSigned` instead of `authnRequestsSigned`

The `callbackSSOSAML` endpoint had similar inconsistencies (items 3 & 4).

## Solution

Align the `signInSSO` and `callbackSSOSAML` implementations by:

1. Using `signingCert` instead of `encryptCert`
2. Adding fallback to `parsedSamlConfig.issuer` for `entityID`
3. Adding fallback to `idpData?.singleSignOnService` then `parsedSamlConfig.entryPoint` for `singleSignOnService`
4. **Using `authnRequestsSigned` for `wantAuthnRequestsSigned`** - this must match the SP's `authnRequestsSigned` flag or samlify throws `ERR_METADATA_CONFLICT_REQUEST_SIGNED_FLAG`
5. Handling both metadata and non-metadata IdP configurations
6. Including `privateKey`/`privateKeyPass` in the metadata branch

## Testing

- Added test suite for SAML SSO with idpMetadata but without metadata XML
- All 84 SAML tests pass
- `pnpm lint`, `pnpm typecheck` pass

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>


theNailz:
## Regarding the link to #6609

@Paola3stefania raised a good question on the original PR about why this fix references issue #6609, since that issue was closed as a "configuration issue".

After investigating, I believe #6609 actually describes the **same bug** this PR fixes:

**Issue #6609 symptom:**
> "After base64decoding and xml deflating the SAMLRequest param, AssertionConsumerServiceURL shows empty string"

**Root cause (this PR):**
The `signInSSO` function was using `encryptCert` instead of `signingCert`. Since samlify ignores `encryptCert` for signing purposes, the `IdentityProvider` was essentially constructed without proper certificate configuration, causing the SAML AuthnRequest to be malformed (empty `AssertionConsumerServiceURL` and `Issuer`).

The issue was likely misdiagnosed as a configuration problem because providing `spMetadata` with explicit values could work around the bug in some cases. However, the actual fix needed was in the `signInSSO` code itself.

This PR aligns `signInSSO` with the correct implementation already present in `callbackSSOSAML` and `acsEndpoint`, which properly handle both metadata-based and field-based IdP configurations.

theNailz:
## Additional Fix: `wantAuthnRequestsSigned` Configuration

This update also fixes an additional bug discovered during code review:

**Issue:** The IdP's `wantAuthnRequestsSigned` was incorrectly set to `parsedSamlConfig.wantAssertionsSigned` instead of `parsedSamlConfig.authnRequestsSigned`.

**Why this matters:** Samlify validates that `SP.authnRequestsSigned === IdP.wantAuthnRequestsSigned` in `createLoginRequest()`. When these don't match, it throws `ERR_METADATA_CONFLICT_REQUEST_SIGNED_FLAG`.

**Example failure scenario:**
- `authnRequestsSigned: true` (SP says "I will sign requests")  
- `wantAssertionsSigned: false` (incorrectly used for IdP)
- IdP says "I don't require signatures" → **MISMATCH → ERROR**

---

## Known Issues (Not Fixed in This PR)

The `callbackSSOSAML` endpoint (lines 1962-1976) has similar issues but they **don't cause runtime bugs** because it only calls `parseLoginResponse()`, which doesn't validate these properties:

1. **Line 1964**: Missing `idpData?.singleSignOnService ||` fallback
2. **Line 1971-1972**: Uses `wantAssertionsSigned` instead of `authnRequestsSigned`

These should be fixed in a follow-up PR for code consistency.

theNailz:
**Update:** The "Known Issues" mentioned in my previous comment have now been fixed in the latest commit (`cc408fc`). 

Both `signInSSO` and `callbackSSOSAML` now have consistent IdP construction:
- ✅ `singleSignOnService` fallback to `idpData?.singleSignOnService`
- ✅ `wantAuthnRequestsSigned` uses `authnRequestsSigned` (not `wantAssertionsSigned`)

All 84 SAML tests pass.

PR #7709, opened 2026-01-30
Add auth handler for latest version

tnspacetime:
The latest version treats the request as immutable so we need to create a copy of the request.

---
## Summary by cubic
Updated the Hono integration docs for better-auth 1.4.18 to handle immutable Request objects. Adds an example that copies the Request before calling the auth handler to prevent runtime errors.

<sup>Written for commit 4b39cba797386ce07bb85fe792598db7a1aab73b. Summary will update on new commits.</sup>

### DIFF
--- a/docs/content/docs/integrations/hono.mdx
+++ b/docs/content/docs/integrations/hono.mdx
@@ -51,6 +51,12 @@ app.on(["POST", "GET"], "/api/auth/*", (c) => {
 	return auth.handler(c.req.raw);
 });
 
+// On better-auth 1.4.18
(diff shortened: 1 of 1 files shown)


himself65 (changes requested):
I don't think this is on our side. we don't make it immutable

tnspacetime:
Hey @himself65 I don't know which side it comes from. But I've upgraded to the latest hono and better-auth and the doc concerns the integration of the two and it failed with the old code, so the doc (which is an integration doc) should point that to the users right?

Issue #7707, opened 2026-01-30
Even if `user.emailVerified = true`, calling the `/send-verification-email` endpoint will still send a verification email.


shinexoh:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

```typescript
import { drizzleAdapter } from 'better-auth/adapters/drizzle';
import { betterAuth, type BetterAuthOptions } from 'better-auth/minimal';

const betterAuthOptions = {
  basePath: '/auth',
  appName: envConfig.APP_NAME,
  baseURL: envConfig.BETTER_AUTH_URL,
  secret: envConfig.BETTER_AUTH_SECRET,
  trustedOrigins: envConfig.TRUSTED_ORIGINS,
  plugins: [
    admin({
      defaultRole: 'user',
      adminRoles: ['admin'],
    }),
    openAPI(),
  ],
  database: drizzleAdapter(db, {
    provider: 'pg',
    usePlural: true,
    schema,
  }),
  secondaryStorage: {
    // ...
  },
  emailAndPassword: {
    enabled: true,
    autoSignIn: true,
    requireEmailVerification: true,
    revokeSessionsOnPasswordReset: true,
    sendResetPassword: async ({ user, url }) => {
      void sendEmail({
        to: user.email,
        subject: 'Reset your password',
        html: resetPasswordTemplate({ url }),
      });
    },
    password: {
      hash: hashPassword,
      verify: async (data) => verifyPassword(data.password, data.hash),
    },
  },
  emailVerification: {
    sendOnSignIn: false,
    sendOnSignUp: false,
    expiresIn: 60 * 60 * 12,
    autoSignInAfterVerification: true,
    sendVerificationEmail: async ({ user, url }) => {
      void sendEmail({
        to: user.email,
        subject: 'Verify your email',
        html: verifyEmailTemplate({ url }),
      });
    },
    afterEmailVerification: async (user) => {
      // ...
    },
  },
  advanced: {
    cookiePrefix: envConfig.APP_NAME,
    database: {
      generateId: false,
    },
  },
  user: {
    deleteUser: {
      enabled: true,
      sendDeleteAccountVerification: async ({ user, url }) => {
        // ...
      },
    },
  },
  session: {
    expiresIn: 60 * 60 * 24 * 15,
    cookieCache: {
      enabled: isProduction,
      maxAge: 5 * 60,
    },
  },
} satisfies BetterAuthOptions;

const auth = betterAuth(betterAuthOptions);
```

[...]

PR #7712, opened 2026-01-30
fix: skip sending email verification to already verified users without a session

bytaesu:
- Closes https://github.com/better-auth/better-auth/issues/7707

---
## Summary by cubic
Prevents sending verification emails when an unauthenticated request targets an already verified email. The endpoint now short-circuits and returns status: true to avoid user enumeration.

- **Bug Fixes**
  - Updated sendVerificationEmail to early-return when there’s no session and the user is missing or already verified (no email sent).
  - Added a test to confirm third-party requests don’t trigger emails for verified users.

<sup>Written for commit 9847e2d1ee5dcebcef032960e4f0a5c9c2778d78. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/email-verification.ts
+++ b/packages/better-auth/src/api/routes/email-verification.ts
@@ -171,19 +171,13 @@ export const sendVerificationEmail = createAuthEndpoint(
 		const session = await getSessionFromCtx(ctx);
 		if (!session) {
 			const user = await ctx.context.internalAdapter.findUserByEmail(email);
-			if (!user) {
+			if (!user || user.user.emailVerified) {
 				await createEmailVerificationToken(
 					ctx.context.secret,
 					email,
 					undefined,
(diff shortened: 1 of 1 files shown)


packages/better-auth/src/api/routes/email-verification.ts
```diff
@@ -171,14 +171,14 @@ export const sendVerificationEmail = createAuthEndpoint(
 		const session = await getSessionFromCtx(ctx);
 		if (!session) {
 			const user = await ctx.context.internalAdapter.findUserByEmail(email);
-			if (!user) {
+			if (!user || user.user.emailVerified) {
```
himself65, line 174: why do we still create token if email verified

bytaesu: I think it's because otherwise it would have to be inferred from the request time 🤔

Issue #7710, opened 2026-01-30
oauthProvider: callback triggered twice causing state_mismatch (worked with oidcProvider)


StipleAdmin:
### Is this suited for github?
Yes, this is suited for github

### To Reproduce
1. Set up auth server with `oauthProvider` plugin (migrated from `oidcProvider`)
2. Set up client app with `genericOAuth` plugin
3. Both apps on subdomains (e.g., auth.example.com, app.example.com)
4. Attempt OAuth login

### Current vs. Expected behavior

**Expected:** User authenticates and redirects to callbackURL

**Actual:** OAuth callback endpoint is hit twice in quick succession:
- First request: consumes state verification successfully
- Second request: fails with `state_mismatch: verification not found`

This worked perfectly with `oidcProvider` - the issue started after migrating to `oauthProvider`.

### Logs
```
=== CALLBACK DEBUG ===
State in URL: RW9cg1d_8VmS5HT6zrAGvPesRVdFa1ti
All cookies: __Secure-app.state=RW9cg1d_8VmS5HT6zrAGvPesRVdFa1ti.xxx
=== CALLBACK DEBUG ===
State in URL: RW9cg1d_8VmS5HT6zrAGvPesRVdFa1ti
All cookies: __Secure-app.state=RW9cg1d_8VmS5HT6zrAGvPesRVdFa1ti.xxx
ERROR [Better Auth]: Failed to parse state [Error [BetterAuthError]: State mismatch: verification not found]
```

### Workaround
Added deduplication in route handler to block duplicate callbacks.

### System info
- better-auth: latest
- Next.js: 16.1.1
- Node.js: 20.x

### Auth server config
```typescript
import { oauthProvider } from "@better-auth/oauth-provider";
import { betterAuth } from "better-auth";
import { jwt } from "better-auth/plugins";

export const auth = betterAuth({
  plugins: [
    jwt(),
    oauthProvider({
      loginPage: "/",
      consentPage: "/consent",
    }),
    nextCookies(),
  ],
});
```

### Client config
```typescript
import { genericOAuth } from "better-auth/plugins";

export const auth = betterAuth({
  plugins: [
    genericOAuth({
      config: [{
        providerId: "my-provider",
        discoveryUrl: "https://auth.example.com/api/auth/.well-known/openid-configuration",
        // ...
      }],
    }),
    nextCookies(),
  ],
});
```

StipleAdmin:
@better-auth-agent This is not a cookie naming issue. 

[...]

PR #7716, opened 2026-01-31
fix(oauth-provider): detect fetch requests using sec-fetch-mode header

bytaesu:
> [!NOTE]
>  Reference: https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/plugins/oidc-provider/authorize.ts#L31

- Closes https://github.com/better-auth/better-auth/issues/7710

---
## Summary by cubic
Detects fetch-based OAuth authorize requests using the sec-fetch-mode header and returns a JSON redirect payload, while keeping 302 redirects for normal navigation. This makes SPA/fetch clients handle redirects cleanly.

- **Bug Fixes**
  - handleRedirect returns JSON when sec-fetch-mode=cors or Accept: application/json is present.
  - Falls back to 302 redirect when no fetch headers are set.
  - Added tests covering both JSON and 302 behaviors.

<sup>Written for commit 08e3573b6cd771d63c4360bba3f437ea3fb92539. Summary will update on new commits.</sup>

### DIFF
--- a/packages/oauth-provider/src/authorize.test.ts
+++ b/packages/oauth-provider/src/authorize.test.ts
@@ -298,3 +298,130 @@ describe("oauth authorize - authenticated", async () => {
 		expect(issParam).toBe(metadataIssuer);
 	});
 });
+
+describe("handleRedirect - fetch detection", async () => {
+	const authServerBaseUrl = "http://localhost:3000";
+	const rpBaseUrl = "http://localhost:5000";
+	const { auth, signInWithTestUser, customFetchImpl } = await getTestInstance({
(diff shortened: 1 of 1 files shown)


dvanmali:
@bytaesu `sec-fetch-mode` is a CORS concept and shouldn't be utilized by the OAuth Provider plugin for redirection. The response type should be determined by the `accept` header.

bytaesu:
Hey @dvanmali,

Thanks for letting me know 🤩
Could you check what the correct way to handle this issue would be?
The current change was based on code taken from the OIDC Provider.

gustavovalverde:
This was fixed in #8519

PR #7718, opened 2026-01-31
feat(oauth-proxy): add replicateData option

bytaesu:
testing

---
## Summary by cubic
Adds a replicateData option to the oauth-proxy plugin to replicate user, account, and session data across isolated databases during OAuth sign-in. This lets preview or multi-env setups (e.g., Convex previews) complete sign-in without a shared database.

- **New Features**
  - New option: replicateData (default false).
  - When enabled, production includes user/session/account in the encrypted payload; target env recreates missing records and persists the session with the same IDs/token.
  - Supports cookie- or database-backed OAuth state; proxies state as needed. Skips creation if records exist and logs errors without blocking sign-in.

<sup>Written for commit 83d0e9f28cfbc1240703d260e3d410dfe4e053c4. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/oauth-proxy/index.ts
+++ b/packages/better-auth/src/plugins/oauth-proxy/index.ts
@@ -257,29 +257,49 @@ export const oAuthProxy = <O extends OAuthProxyOptions>(opts?: O) => {
 						const { user, session, account } = payload.replicationData;
 
 						try {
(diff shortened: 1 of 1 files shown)


bytaesu:
-> This flow seems better https://github.com/better-auth/better-auth/pull/7720

PR #7732, opened 2026-01-31
fix(organization): update path matching for active member signals

LovelessCodes:
* Enhanced the path matching logic to include `/organization/set-active` for both `$activeMemberSignal` and `$activeMemberRoleSignal`.
* This change ensures that the signals are triggered correctly for the active member functionality, when switching the active organization.

---
## Summary by cubic
Fixes active member and role not updating when switching organizations by matching paths that set the active org. Signals now fire correctly on org switch.

- **Bug Fixes**
  - Updated matchers to include /organization/set-active (startsWith) alongside /organization/update-member-role.
  - Triggers $activeMemberSignal and $activeMemberRoleSignal when the active org changes.

<sup>Written for commit 2531b3de43e519f153224bb629420a807bdead88. Summary will update on new commits.</sup>


LovelessCodes:
Is this not relevant anymore? @himself65

PR #7735, opened 2026-02-01
fix(organization): sync members role if the role name was changed

### LINKED ISSUE #7727: when we update a role name in the dynamic role the member table is not updated
I’m using dynamic-access-control.

It sets up an org role table and uses these roles in the member table’s role column. However, here’s the problem: we’re adding the role name here, not the ID. As a result, if we change the role name, it doesn’t get updated here, which creates a gap between the roles and members. The update role method doesn’t update the role in the member table.

bytaesu:
- Closes #7727

---
## Summary by cubic
Renaming an organization role now automatically updates all member records and pending invitations using that role, including multi‑role cases, keeping access control consistent. The client also reacts to role updates so active member and role state stay in sync. Addresses #7727.

- **Bug Fixes**
  - UpdateOrgRole cascades the new role name to members in the org, including multi‑role members.
  - Pending invitations are updated to the new role name, including multi‑role invites.
  - Client listeners watch /organization/update-role to refresh active member and role signals.
  - Added tests for role rename updates for members and invitations.

<sup>Written for commit 2717b40288a46bbc67405d138332e02e373c62ba. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/organization/client.ts
+++ b/packages/better-auth/src/plugins/organization/client.ts
@@ -258,13 +258,19 @@ export const organizationClient = <CO extends OrganizationClientOptions>(
 			},
 			{
 				matcher(path) {
-					return path.includes("/organization/update-member-role");
+					return (
+						path.includes("/organization/update-member-role") ||
+						path.includes("/organization/update-role")
+					);
 				},
 				signal: "$activeMemberSignal",
 			},
 			{
 				matcher(path) {
-					return path.includes("/organization/update-member-role");
+					return (
+						path.includes("/organization/update-member-role") ||
+						path.includes("/organization/update-role")
+					);
 				},
 				signal: "$activeMemberRoleSignal",
 			},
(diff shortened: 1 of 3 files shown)


bytaesu:
-> updating for array case

bytaesu:
@ping-maxwell,

Since roles are comma-separated, it seems like we need to call the update multiple times when updating a specific role.

Will the new org plugin handle this more cleanly, or is this an unavoidable trade-off for flexibility?

pranavgoel29:
> @ping-maxwell,
> 
> Since roles are comma-separated, it seems like we need to call the update multiple times when updating a specific role.
> 
> Will the new org plugin handle this more cleanly, or is this an unavoidable trade-off for flexibility?

Just adding more context here from what I know from the discord chat with @ping-maxwell we will be switching to roleid instead of the role name completely avoiding these pitfalls. 😊

Thanks for the quick update!

pranavgoel29:
Just curious when can we expect this to be released?

himself65:

bytaesu:
@ping-maxwell, Could this issue also exist in the new org plugin?

PR #7736, opened 2026-02-01
fix(organization): prevent deletion of roles assigned to members

bytaesu:
Related to https://github.com/better-auth/better-auth/issues/7727#issuecomment-3829065189

---
## Summary by cubic
Prevent deleting organization roles that are assigned to members. Shows a clear error and requires reassigning members before deletion.

- **Bug Fixes**
  - Added server-side check in deleteOrgRole to block deletion if any member uses the role (supports single or multiple roles per member).
  - Introduced ROLE_IS_ASSIGNED_TO_MEMBERS error code with a helpful message.
  - Added tests for assigned role deletion and multi-role member scenarios.

<sup>Written for commit 31741c47fb9927e3e37ff5ce3ba4ac6fbea03f07. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-access-control.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-access-control.ts
@@ -473,18 +473,24 @@ export const deleteOrgRole = <O extends OrganizationOptions>(options: O) => {
 			);
 
 			// Check if any members are assigned to this role
-			const membersInOrg = await ctx.context.adapter.findMany<Member>({
+			const roleToDelete = existingRoleInDB.role;
+			const members = await ctx.context.adapter.findMany<Member>({
 				model: "member",
 				where: [
 					{
 						field: "organizationId",
 						value: organizationId,
 						operator: "eq",
+						connector: "AND",
+					},
+					{
+						field: "role",
+						value: roleToDelete,
+						operator: "contains",
 					},
 				],
 			});
-			const roleToDelete = existingRoleInDB.role;
-			const memberWithRole = membersInOrg.find((member) => {
+			const memberWithRole = members.find((member) => {
 				const memberRoles = member.role.split(",").map((r) => r.trim());
 				return memberRoles.includes(roleToDelete);
 			});


packages/better-auth/src/plugins/organization/routes/crud-access-control.ts
```diff
@@ -472,6 +472,36 @@ export const deleteOrgRole = <O extends OrganizationOptions>(options: O) => {
...
+						value: organizationId,
+						operator: "eq",
+					},
+				],
+			});
+			const roleToDelete = existingRoleInDB.role;
+			const memberWithRole = membersInOrg.find((member) => {
+				const memberRoles = member.role.split(",").map((r) => r.trim());
+				return memberRoles.includes(roleToDelete);
+			});
```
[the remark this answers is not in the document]
bytaesu: “findMany → parse roles in TS and filter” is already the pattern used in crud-members.

packages/better-auth/src/plugins/organization/routes/crud-access-control.ts
```diff
@@ -472,6 +472,36 @@ export const deleteOrgRole = <O extends OrganizationOptions>(options: O) => {
 				existingRoleInDB.permission as never as string,
 			);
 
+			// Check if any members are assigned to this role
+			const membersInOrg = await ctx.context.adapter.findMany<Member>({
```
himself65: I think this would have performance when there are many members. Can we use like foreign key to avoid the deletion?

bytaesu: > https://github.com/better-auth/better-auth/pull/7735#issuecomment-3829908611

Since role can be stored as a comma-separated array, applying FK is difficult. For now, I've at least added a contains operator.

pranavgoel29:
Just curious whats the usecase of allowing multiple roles to be assigned to a user?

bytaesu:
> Just curious whats the usecase of allowing multiple roles to be assigned to a user?

For example, someone can be both an `engineer` and handle `sales`. It depends on the business logic!

PR #7749, opened 2026-02-01
chore: correct environment variable name in baseURL warning

### LINKED ISSUE #7748: ENV for Base URL has name mismatch in logs and an actual value

### To Reproduce

Create a backend without setting BETTER_AUTH_URL.

In console you would receive message:
```
2026-02-01T11:58:08.244Z WARN [Better Auth]: [better-auth] Base URL could not be determined. Please set a valid base URL using the baseURL config option or the BETTER_AUTH_BASE_URL environment variable. Without this, callbacks and redirects may not work correctly.
```

Set BETTER_AUTH_BASE_URL and message is still here

Actual name for variable according to documentation is BETTER_AUTH_URL so console output should use it because provided name has no effect. 

1.4.18

### System info

```bash
Not needed
```

Backend

```typescript

```

### Additional context

_No response_

cucumber-sp:
Fix: ENV variable name in warning for Base URL
Misc: add auto-claude files to gitignore

Closes: #7748

---
## Summary by cubic
Updated the base URL warning to use BETTER_AUTH_URL, fixing the incorrect env var reference (closes #7748).

<sup>Written for commit e6f2999937d7fc81f461a3cd26b950b1378f6b23. Summary will update on new commits.</sup>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -201,14 +201,3 @@ test-results/
 state.txt
 
 .cursor/
-
-# Auto Claude data directory
-.auto-claude/
-
-# Auto Claude generated files
-.auto-claude-security.json
-.auto-claude-status
-.claude_settings.json
-.worktrees/
-.security-key
-logs/security/


bytaesu (changes requested):
-> moved to the comment

.gitignore
```diff
@@ -201,3 +201,14 @@ test-results/
...
+# Auto Claude data directory
+.auto-claude/
+
+# Auto Claude generated files
+.auto-claude-security.json
+.auto-claude-status
+.claude_settings.json
+.worktrees/
+.security-key
+logs/security/
```
bytaesu: We don't need these

cucumber-sp: Removed

Issue #6564, opened 2025-12-06
Dev dependency `expo-network` is set to optional, but imported on expo package


dccarmo:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Add `@better-auth/expo` package
2. Import it
3. Try to run the Expo project

### Current vs. Expected behavior

**Current**

Metro returns an error: `Unable to resolve module expo-network`.

**Expected**

No errors.

### What version of Better Auth are you using?

1.4.0

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.1.0: Mon Oct 20 19:32:41 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6000",
    "release": "25.1.0",
    "cpuCount": 8,
    "cpuModel": "Apple M1 Pro",
    "totalMemory": "32.00 GB",
    "freeMemory": "0.22 GB"
  },
  "node": {
    "version": "v22.18.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "10.24.0"
  },
  "frameworks": [
    {
      "name": "react",
      "version": "19.1.0"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "1.4.0",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Package

### Auth config (if applicable)

```typescript

```

### Additional context

The `expo-network` peer dependency is set as optional here: https://github.com/better-auth/better-auth/blob/c79b6e3aae409bee98653e3b85a5f24ea3f91663/packages/expo/package.json#L81

But is imported here: https://github.com/better-auth/better-auth/blob/c79b6e3aae409bee98653e3b85a5f24ea3f91663/packages/expo/src/online-manager.ts#L22

I'm guessing the fix is to have it not be optional, so the package manager can fetch it when installing dependencies.

SergioJuniorCE:
bump?

hubingkang:
same problem

bytaesu:
> [!NOTE]
> https://www.better-auth.com/docs/integrations/expo#install-client-dependencies 

Hi everyone,

We use `expo-network` on the Expo client side, and we updated the documentation accordingly. Based on my testing, the issue is resolved when following the latest guide.

I'm closing this. Feel free to reopen this or mention me if the issue persists.

PR #7752, opened 2026-02-01
fix(expo): `expo-web-browser` is not installed as a dependency

### LINKED ISSUE #7739: [Expo][Android] Error: Uncaught (in promise, id: 0) Error: "expo-web-browser" is not installed as a dependency!

### To Reproduce

Just follow the steps from the official docs: https://www.better-auth.com/docs/integrations/expo

It should not throw the `Error: Uncaught (in promise, id: 0) Error: "expo-web-browser" is not installed as a dependency!` error

1.4.18

### System info

```bash
- NX Monorepo
- @better-auth/expo v1.4.18
- expo v54.0.32
```

Client

```typescript
import { createAuthClient } from 'better-auth/client';

  baseURL: `${SharedConstants.Common.AppHttpUriProtocol}${SharedConstants.Common.ServerUrl}`,
  basePath: `${SharedConstants.Common.ServerApiPrefix}/auth`,
  plugins: [
    expoClient({
      scheme: 'myapp',
      storagePrefix: 'myapp',
      storage: SecureStore,
    }),
  ],
});
```

### Additional context

_No response_

AdiMarianMutu:
Fix: Error: "expo-web-browser" is not installed as a dependency!
- Closes #7739

## Unit Tests

Don't see how this could be unit tested as the `expo-web-browser` import is mocked.

---
## Summary by cubic
Fixes the false “expo-web-browser is not installed” error by trying require() when the dynamic import fails, and adds non-null assertions around Browser calls. Restores compatibility in Nx monorepos and mixed CJS/ESM setups.

- **Bug Fixes**
  - If import("expo-web-browser") fails, fall back to require(); only throw if both fail.
  - Use Browser! for dismissAuthSession and openAuthSessionAsync to prevent undefined access.

<sup>Written for commit d8a6620d823fbd3b2556addff48ad2da4e6927e3. Summary will update on new commits.</sup>

### DIFF
--- a/packages/expo/src/client.ts
+++ b/packages/expo/src/client.ts
@@ -375,7 +375,7 @@ export const expoClient = (opts: ExpoClientOptions) => {
 
 							if (Platform.OS === "android") {
 								try {
-									Browser.dismissAuthSession();
+									Browser!.dismissAuthSession();
 								} catch {}
 							}
 
@@ -391,7 +391,7 @@ export const expoClient = (opts: ExpoClientOptions) => {
 								params.append("oauthState", oauthStateValue);
 							}
 							const proxyURL = `${context.request.baseURL}/expo-authorization-proxy?${params.toString()}`;
-							const result = await Browser.openAuthSessionAsync(
+							const result = await Browser!.openAuthSessionAsync(
 								proxyURL,
 								to,
 								opts?.webBrowserOptions,


himself65 (approved):

AdiMarianMutu:
> why do you try catch it twice

If the dynamic ES `import` fails (which would throw the error I've mentioned in the PR itself) it means that we can try via the common `require` import which may fix the issue within some codebase. (like mine with an NX monorepo).

himself65:
> > why do you try catch it twice
> 
> If the dynamic ES `import` fails (which would throw the error I've mentioned in the PR itself) it means that we can try via the common `require` import which may fix the issue within some codebase. (like mine with an NX monorepo).

How is it possible that there are ESM/CJS in a single file? What's the bundler that expo are using

bytaesu:
@himself65,

Isn’t https://github.com/better-auth/better-auth/issues/6564 related to `expo-network` and therefore a separate issue? -> I updated description

AdiMarianMutu:
> > > why do you try catch it twice
> > 
> > 
> > If the dynamic ES `import` fails (which would throw the error I've mentioned in the PR itself) it means that we can try via the common `require` import which may fix the issue within some codebase. (like mine with an NX monorepo).
> 
> How is it possible that there are ESM/CJS in a single file? What's the bundler that expo are using

Expo primarily uses the Metro bundler, which converts ESM imports to CommonJS for native platforms since native doesn't support ESM fully.

In NX monorepos (like mine), shared libraries or mixed module setups can cause dynamic import("expo-web-browser") to fail due to ESM resolution issues in Metro's stricter package.json exports enforcement or interop problems between CJS/ESM. (Being a dynamic import, Metro bundler can't statically find it) 

The fallback to require("expo-web-browser") succeeds because require handles CJS resolution better in these environments, avoiding the "not installed" error despite the package being present.

himself65:
> > > > why do you try catch it twice
> > > 
> > > 
> > > If the dynamic ES `import` fails (which would throw the error I've mentioned in the PR itself) it means that we can try via the common `require` import which may fix the issue within some codebase. (like mine with an NX monorepo).
> > 
> > 
> > How is it possible that there are ESM/CJS in a single file? What's the bundler that expo are using
> 
> Expo primarily uses the Metro bundler, which converts ESM imports to CommonJS for native platforms since native doesn't support ESM fully.
> 
> In NX monorepos (like mine), shared libraries or mixed module setups can cause dynamic import("expo-web-browser") to fail due to ESM resolution issues in Metro's stricter package.json exports enforcement or interop problems between CJS/ESM. (Being a dynamic import, Metro bundler can't statically find it)
> 
> The fallback to require("expo-web-browser") succeeds because require handles CJS resolution better in these environments, avoiding the "not installed" error despite the package being present.

Should we emit the CJS output? Is this a better way?

AdiMarianMutu:
> > > > > why do you try catch it twice
> > > > 
> > > > 
> > > > If the dynamic ES `import` fails (which would throw the error I've mentioned in the PR itself) it means that we can try via the common `require` import which may fix the issue within some codebase. (like mine with an NX monorepo).
> > > 
> > > 
> > > How is it possible that there are ESM/CJS in a single file? What's the bundler that expo are using
> > 
> > 
> > Expo primarily uses the Metro bundler, which converts ESM imports to CommonJS for native platforms since native doesn't support ESM fully.
> > In NX monorepos (like mine), shared libraries or mixed module setups can cause dynamic import("expo-web-browser") to fail due to ESM resolution issues in Metro's stricter package.json exports enforcement or interop problems between CJS/ESM. (Being a dynamic import, Metro bundler can't statically find it)
> > The fallback to require("expo-web-browser") succeeds because require handles CJS resolution better in these environments, avoiding the "not installed" error despite the package being present.
> 
> Should we emit the CJS output? Is this a better way?

That would definitely be a much better solution which could catch (and automatically fix) other latent issues like this one.

himself65:
TBH im not fimilar with epxo. Let's merge it and see

PR #7756, opened 2026-02-01
refactor: remove unused trailing AND connectors

TheUntraceable:
## Summary by cubic
Removed unnecessary "connector: 'AND'" fields from organization role access-control queries. This aligns with the filter API and avoids malformed filters.

- **Refactors**
  - Dropped trailing AND connectors from single-condition filters across role create, delete, list, get, and update routes.
  - No behavior change; queries use the builder’s default AND handling.

<sup>Written for commit c8e52e28320586b4e6448abe51a2f20647587a03. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-access-control.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-access-control.ts
@@ -156,7 +156,6 @@ export const createOrgRole = <O extends OrganizationOptions>(options: O) => {
 						field: "userId",
 						value: user.id,
 						operator: "eq",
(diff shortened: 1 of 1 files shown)


bytaesu:
Hi @TheUntraceable,

The connector determines an AND/OR group rather than linking list items, so this code causes no harm. I think it actually makes the intent clearer, so I’d like to keep it.

If you have further questions, feel free to mention me!

TheUntraceable:
Ah OK, I wrongly diagnosed this as an issue i was having with a convex integration. Thanks!

PR #7769, opened 2026-02-03
feat: add account object to session

cvienna:
Note: This is my first ever PR, so I'd really appreciate some feedback, and maybe also make sure to double check my code.

### Intro
I've been struggling with multiple accounts lately, I found it a nightmare to deal with multiple accounts per user with `allowDifferentEmails` set to true. Because there was no way to know which account was currently _in use_.

### What did I do?
I added the account object to the session. So when you call the: `getSession()` aka `/get-session` you get following:
```typescript
{
  account: Account
  session: Session
  user: User
}
```
Where before you would only get:
```typescript
{
  session: Session
  user: User
}
```

---
## Summary by cubic
Adds the account object to session data so clients always know which account is active. getSession now returns { account, session, user } and OAuth sign-ins create sessions tied to the account.

- **New Features**
  - getSession returns account alongside session and user.
  - Session cookies/JWT/JWE and cookie cache include account.
  - Internal adapter fetches and returns account with session (single and batch).
  - OAuth flow links/creates the account, returns it, and creates a session with that accountId.
  - Types updated to expose account in inferred Session; tests adjusted.

- **Migration**
  - Database: add a non-null accountId field on sessions with a FK to accounts; backfill existing sessions.
  - Ensure all session creation paths provide accountId when creating a session.
  - Update any consumers expecting only { session, user } to handle { account, session, user }.

<sup>Written for commit 12fefd946a785c934545e3185768751cc3c0814e. Summary will update on new commits.</sup>

### DIFF
--- a/docs/content/docs/authentication/email-password.mdx
+++ b/docs/content/docs/authentication/email-password.mdx
@@ -40,6 +40,10 @@ To sign a user up, you can use the `signUp.email` function provided by the clien
 ```ts
 type signUpEmail = {
     /**
+     * The name of the user.
+     */
+    name: string = "John Doe"
+    /**
      * The email address of the user.
      */
     email: string = "john.doe@example.com"
@@ -47,10 +51,6 @@ type signUpEmail = {
      * The password of the user. It should be at least 8 characters long and max 128 by default.
      */
     password: string = "password1234"
-    /**
-     * An optional name of the user.
-     */
-    name?: string = "John Doe"
     /**
      * An optional profile image of the user.
      */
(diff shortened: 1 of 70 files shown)


bytaesu:
Hi @cvienna, thanks for your first PR!

We don't modify the core schema in minor versions.. this would be huge breaking change.

And are there use cases where we need to know what the current account is? Feel free to let me know.

cvienna:
Thank you for responding @bytaesu.

Imagine a scenario where you have 2 google accounts linked to your main user. In `accountLinking` you have set `allowDifferentEmails` to true.
When the user is authenticated, you don't know which account is used in your current session. There is no way to tell which of these accounts is currently in use. Because the session only contains the user.

This is a very frequent issue I see popping up, seems like a lot of people have the same issue.

I mean, what's the point of: `allowDifferentEmails` if you can't even use it properly.

PR #7772, opened 2026-02-03
feat(email-otp): add `persistOTP` option to resend existing OTP

### LINKED ISSUE #1568: Email OTP Plugin: Option to Accept Multiple OTPs

One common customer flow is: The app sends an email OTP to the user. If there's a delay receiving the email, the user might request another code. When both OTP emails eventually arrive, the user may try using the code from the first email, which better-auth will reject.

No matter how reliable the email provider is, I’ve seen this happen repeatedly in real-world apps. In fact, I’ve been asked twice by clients to build custom auth systems to have better control on things like this, which has been their top support request from non-tech-savvy users.

In that context, it would be nice for Better Auth to support a new option to solve this issue.

Add an option like `allowMultipleOTPs` to let all OTPs sent within the expiry period be valid.

```ts
emailOTP({
  allowMultipleOTPs: true,
  // other options...
})
```

Also, after a successful login, all outstanding OTPs should be deleted to prevent any reuse.

Building a custom auth implementation.

### Additional context

This would dramatically improve the end user login experience. I don’t think there are official guidelines against having multiple valid OTPs at the same time. The key is ensuring all OTPs have an expiry in place and are properly deleted on a successful login.

bytaesu:
I chose this approach because it felt more reliable than sending multiple OTPs and validating them all together. It addresses the same case where a user requests multiple OTPs while emails are delayed.

- Closes #1568

---
## Summary by cubic
Add a persistOTP option to email-otp to resend the same OTP and extend its expiry instead of generating new codes. Addresses #1568 by preventing multiple valid codes when emails are delayed.

- **New Features**
  - persistOTP (default false): reuses the existing unexpired OTP and extends its expiry; otherwise generates a new one.
  - Works when the OTP is recoverable (plain/encrypted/custom decrypt); falls back when hashed/custom-hashed.
  - Added retrieveOTP helper, route logic, docs, and tests.
  - Deletes existing verification when the user doesn’t exist in the persistOTP path.

- **Migration**
  - Opt-in only; no breaking changes.
  - Ignored when storeOTP is "hashed" or uses a custom hasher.

<sup>Written for commit 726bf13b85463ec96906c0746fa7856928f8a75c. Summary will update on new commits.</sup>


packages/better-auth/src/plugins/email-otp/routes.ts
```diff
@@ -7,7 +7,7 @@ import { setCookieCache, setSessionCookie } from "../../cookies";
 import { generateRandomString, symmetricDecrypt } from "../../crypto";
```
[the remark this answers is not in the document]
bytaesu: For now, I’ll keep it as is. If the duplication grows further, I’ll refactor and separate it.

packages/better-auth/src/plugins/email-otp/types.ts
```diff
@@ -79,4 +79,12 @@ export interface EmailOTPOptions {
...
 	 */
 	overrideDefaultEmailVerification?: boolean | undefined;
+	/**
+	 * Resend existing OTP with extended expiry instead of generating a new one.
+	 *
+	 * Ignored when storeOTP is "hashed" since OTP cannot be recovered.
+	 *
+	 * @default false
+	 */
+	persistOTP?: boolean | undefined;
```
himself65, line 89: Should we use `stoteXXXInDatabase` to keep the naming consistency

bytaesu: This isn't an option that controls storing data in the database. I aligned it for consistency with `storeOTP` naming 🤔

himself65: Yeah, that's another question. I think it would be confusing. Let's rename it to `storeInDatabase` to align with verification and others

bytaesu: `storeInDatabase` doesn’t fit this option. How about `reuseOTP`?

himself65: uhmm, how about `sendStrategy: enum xxx`

maoosi: can I suggest `resendStrategy: 'rotate' | 'reuse'`?

bytaesu: > can I suggest `resendStrategy: 'rotate' | 'reuse'`?

I think this makes sense ❤️

I'll finalize this PR today

himself65:

maoosi:
@bytaesu correct me if I’m wrong, but there might be a small issue when used together with [`allowedAttempts`](https://better-auth.com/docs/plugins/email-otp#options)?

After the final allowed failed attempt, the verification record stays in the database with a maxed-out counter and is only deleted on the next verify attempt. So if the user requests a new OTP during that window, the system re-sends the same OTP and extends the expiry, but does not check the attempt counter.

Not sure what the behaviour should be, but might be worth checking the attempt count before extending the expiry? If already maxed out, maybe generate a fresh OTP (and/or delete the old record) instead of reusing the exhausted one.

maoosi:
@bytaesu quick check: when do you think you might have time to apply the latest changes?

Happy to help finalise if you're short on time. I'm waiting on this PR before deploying related changes on my side.

bytaesu:
Hi @maoosi, sorry for delay 🙏

I just opened a new PR https://github.com/better-auth/better-auth/pull/8560

Issue #7558, opened 2026-01-22
Incomplete flow in the new OAuth 2.1 Provider plugin


pxtrickb:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Create two next.js applications, both with better-auth
2. On one application implement the new oauth-provider plugin (we'll call this the auth provider)
3. On the second application implement the generic oauth plugin (we'll call this the auth consumer)
4. Configure the two to work with eachother
5. Call `authClient.signIn.oauth2` from the auth consumer app.
6. Try completing the sign in process on the auth provider app. <--- This is where the issue appears.

### Current behavior:

When users get redirected to the authentication provider app for completing the sign in process (e.g.: filling an email/password form), they are not redirected back to the authentication consumer app after a valid completion. Instead, the form's usual logic takes over and redirects user's to the "dashboard", in my case. But even if I disable the redirect, the page just hangs on the sign in form. The session is created, but the OAuth flow does not go further.

### Expected behavior:

After the users successfully sign in, they should be redirected back to the consumer app.

### What version of Better Auth are you using?

1.4.17

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Pro",
    "release": "10.0.26100",
    "cpuCount": 20,
    "cpuModel": "12th Gen Intel(R) Core(TM) i7-12700H",
    "totalMemory": "15.62 GB",
    "freeMemory": "1.99 GB"
  },
  "node": {
    "version": "v24.0.2",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.4.1"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.1.4"
    },
    {
      "name": "react",
      "version": "19.2.3"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.17.2"
    },
    {
      "name": "drizzle",
      "version": "^0.45.1"
    }
  ],
  "betterAuth": {
    "version": "^1.4.17",
    "config": {
      "trustedOrigins": [
        "[REDACTED]"
      ],
      "emailAndPassword": {
        "enabled": true
      },
      "advanced": {
        "cookiePrefix": "[REDACTED]"
      },
      "plugins": [
        {
          "name": "oauthProvider",
          "config": {
            "id": "oauthProvider",
            "options": {
              "codeExpiresIn": 600,
              "accessTokenExpiresIn": 3600,
[...]

PR #7159, opened 2026-01-07
fix(cookies): fallback to isProduction when `baseURL` is not set

bytaesu:
docs

---

- Closes https://github.com/better-auth/better-auth/issues/7156

---
## Summary by cubic
Default to secure cookies in production when baseURL is not set. Prevents insecure cookies in prod. Closes #7156.

- **Bug Fixes**
  - Use NODE_ENV=production as fallback to enable Secure cookies when baseURL is missing.
  - Added test to assert Secure is present in Set-Cookie for prod with no baseURL.

<sup>Written for commit bd26c93dd68081f2ffe38e8f10ef2cb30b163620. Summary will update on new commits.</sup>

PR #7778, opened 2026-02-03
fix(oauth-provider): login via a sign-in provider redirects

dvanmali:
Additionally uses `isProduction` for an empty string which is set in the `createAuthContext` function for baseURL. Fixes redirection from an oauth flow back to the original application for the oauth-provider plugin.

Related: #7558

### DIFF
--- a/packages/better-auth/src/cookies/index.ts
+++ b/packages/better-auth/src/cookies/index.ts
@@ -35,10 +35,8 @@ export function createCookieGetter(options: BetterAuthOptions) {
 	const secure =
 		options.advanced?.useSecureCookies !== undefined
 			? options.advanced?.useSecureCookies
-			: options.baseURL !== undefined
+			: options.baseURL?.length
 				? options.baseURL.startsWith("https://")
-					? true
-					: false
 				: isProduction;
 	const secureCookiePrefix = secure ? SECURE_COOKIE_PREFIX : "";
 	const crossSubdomainEnabled =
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/cookies/index.ts
```diff
@@ -35,10 +35,8 @@ export function createCookieGetter(options: BetterAuthOptions) {
 	const secure =
 		options.advanced?.useSecureCookies !== undefined
 			? options.advanced?.useSecureCookies
-			: options.baseURL
+			: options.baseURL?.length
```
himself65, line 38: why use length here?

dvanmali: Empty string gets set in `createAuthContext` [here](https://github.com/better-auth/better-auth/blob/9f40c35ee428b0313c90089d857b219b911a51ca/packages/better-auth/src/context/create-context.ts#L137) and [here](https://github.com/better-auth/better-auth/blob/9f40c35ee428b0313c90089d857b219b911a51ca/packages/better-auth/src/context/create-context.ts#L205). Traced back to changes on #7159 where `options.baseURL !== undefined` was removed.

dvanmali:
@himself65 marking ready

dvanmali:
@himself65 auto merge doesn't work without Vercel in CI. To fix, could Vercel be skippable as a CI requirement and simply just check that demo and docs can be built instead?

dvanmali:
@himself65 bump

dvanmali:
@himself65 bump

PR #7800, opened 2026-02-04
fix(sso): preserve OAuth provider state through SSO authentication flows

rbayliss:
When using the SSO plugin (SAML in my case) with the OAuth provider plugin, the OAuth authorization flow would break after SSO authentication. Users would authenticate successfully via SAML or OIDC, but instead of being directed to the consent screen or registration flow, they would be redirected to the callback URL without completing the OAuth flow.

The root cause is that OAuth authorization parameters (client_id, redirect_uri, scope, etc.) were not preserved through the SSO authentication round-trip. The SSO sign-in endpoint was explicitly passing false for additionalData when generating state, and there was no mechanism to restore OAuth state after the IdP callback.

Note: This PR builds on work in https://github.com/better-auth/better-auth/pull/7781

This PR makes the following changes:
1. OAuth provider "before" hook: Extended to inject oauth_query into additionalData for /sign-in/sso (in addition to existing /sign-in/social and /sign-in/oauth2)
2. SSO sign-in endpoint: Now accepts additionalData in the request body and passes it through to generateState() (OIDC) and generateRelayState() (SAML)
3. SSO callback endpoints: After parsing state/RelayState, if OAuth query params are present, calls setOAuthState() to make them available to the OAuth provider's "after" hook
4. Export setOAuthState: Exposed from better-auth/api to enable cross-plugin OAuth state management

---
## Summary by cubic
Preserves OAuth provider state across SSO (SAML/OIDC) so the OAuth flow resumes after IdP login. Users now reach consent/registration instead of being redirected to the callback early.

- **Bug Fixes**
  - Added query to additionalData for /sign-in/sso in the OAuth provider "before" hook.
  - SSO sign-in now accepts additionalData and forwards it to generateState (OIDC) and generateRelayState (SAML).
  - SSO callbacks restore OAuth context by calling setOAuthState when OAuth params are present.
  - Exported setOAuthState from better-auth/api for cross-plugin state handoff.

<sup>Written for commit 7fb22332b5a33323ec41714272ff05a41e64c6d2. Summary will update on new commits.</sup>

### DIFF
--- a/packages/sso/src/routes/sso.ts
+++ b/packages/sso/src/routes/sso.ts
@@ -13,6 +13,7 @@ import {
 	createAuthEndpoint,
 	getSessionFromCtx,
 	sessionMiddleware,
+	setOAuthState,
 } from "better-auth/api";
 import { setSessionCookie } from "better-auth/cookies";
 import { generateRandomString } from "better-auth/crypto";
@@ -970,6 +971,12 @@ const signInSSOBodySchema = z.object({
 		})
 		.optional(),
 	providerType: z.enum(["oidc", "saml"]).optional(),
+	additionalData: z
+		.record(z.string(), z.any())
+		.meta({
+			description: "Additional data to be passed through the SSO flow",
+		})
+		.optional(),
 });
 
 export const signInSSO = (options?: SSOOptions) => {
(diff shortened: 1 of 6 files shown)


dvanmali:
Great solution, solves without breaking oauth-provider plugin!

@himself65 (cc @Paola3stefania for SSO confirmation)

packages/sso/src/routes/sso.ts
```diff
@@ -2935,40 +2774,20 @@ export const acsEndpoint = (options?: SSOOptions) => {
...
+			// set the OAuth state so the oauth-provider's after hook can continue the flow
+			if (relayState?.query) {
+				await setOAuthState({
+					callbackURL: relayState.callbackURL,
+					codeVerifier: relayState.codeVerifier,
+					errorURL: relayState.errorURL,
+					newUserURL: relayState.newUserURL,
+					expiresAt: relayState.expiresAt,
+					query: relayState.query,
+				});
```
Paola3stefania, line 2787: you can simplify this with


```suggestion
					if (relayState?.query) {
    await setOAuthState(relayState);
}
```

Paola3stefania (changes requested):
Thank you for this work :) @rbayliss 

Can you please address the feedback? ty

himself65:

PR #7811, opened 2026-02-05
fix(oauth-provider): return `url` instead of `uri` in continue and consent endpoints

### LINKED ISSUE #7807: `/oauth2/continue` endpoint returns `uri` instead of `url`, breaking   client-side redirects

### To Reproduce

1. create a setup with the new oauth-provider package
2. add signup.page and signup.shouldRedirect option, and always return true for shouldRedirect (for simulation)
3. on the client-side, handle the signup.page url, and call the continue endpoint through better-auth client. you will notice that you will not automatically redirect

current behaviour: the continue endpoint returns a response which we have to handle inside an if check and assign to window.location

expectation: the redirect should be handled seamlessly inside better-auth. 

example: /sign-in/email inside an oauth flow automatically handles redirect even if it returns json

1.5.0-beta.9

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "cpuCount": 16,
    "cpuModel": "Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz",
    "totalMemory": "31.34 GB",
    "freeMemory": "10.01 GB"
  },
  "node": {
    "version": "v24.5.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "10.28.1"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "~16.0.6"
    },
    {
      "name": "react",
      "version": "19.2.0"
    },
    {
      "name": "fastify",
      "version": "5.6.2"
    }
  ],
  "databases": [
    {
      "name": "mongodb",
      "version": "7.0.0"
    }
  ],
  "betterAuth": {
    "version": "1.5.0-beta.9",
    "config": null
  }
}
```

Package, Client, Backend

```typescript

```

### Additional context

the client-side redirectPlugin in `packages/better-auth/src/client/fetch-plugins.ts` expects url:
```ts
  if (context.data?.url && context.data?.redirect) {
      window.location.href = context.data.url;
  }
```

bytaesu:
> [!NOTE]
> This is a breaking change to the API response field names, but it needs to be fixed because the existing code already breaks compatibility with the DefaultFetchPlugins
>
> https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/client/fetch-plugins.ts#L8

 - Closes #7807

---
## Summary by cubic
Switch consent and continue endpoint responses from uri to url to match client fetch plugins and fix redirect handling. This is a breaking change; tests updated to expect url.

- **Migration**
  - Read data.url instead of data.uri from consent and continue endpoints.
  - Update any redirects that use res.uri to use res.url (e.g., new URL(res.data.url)).

<sup>Written for commit 3f2330961b689aeaf11b889dd80010985084be15. Summary will update on new commits.</sup>

### DIFF
--- a/packages/oauth-provider/src/consent.ts
+++ b/packages/oauth-provider/src/consent.ts
@@ -43,7 +43,7 @@ export async function consentEndpoint(
 	if (!accepted) {
 		return {
 			redirect: true,
-			uri: formatErrorURL(
+			url: formatErrorURL(
 				query.get("redirect_uri") ?? "",
 				"access_denied",
 				"User denied access",
@@ -121,6 +121,6 @@ export async function consentEndpoint(
 	const { url } = await authorizeEndpoint(ctx, opts);
 	return {
 		redirect: true,
-		uri: url,
+		url,
 	};
 }
--- a/packages/oauth-provider/src/continue.ts
+++ b/packages/oauth-provider/src/continue.ts
@@ -41,7 +41,7 @@ async function selected(
 	const { url } = await authorizeEndpoint(ctx, opts);
 	return {
 		redirect: true,
-		uri: url,
+		url,
 	};
 }
 
@@ -61,7 +61,7 @@ async function created(
 	const { url } = await authorizeEndpoint(ctx, opts);
 	return {
 		redirect: true,
-		uri: url,
+		url,
 	};
 }
 
@@ -84,6 +84,6 @@ async function postLogin(
 	});
 	return {
 		redirect: true,
-		uri: url,
+		url,
 	};
 }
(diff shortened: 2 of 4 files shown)


dvanmali:
To prevent breaking,

Option 1: use the following instead of changing the output format:

```
{
	redirect: true,
	url,
	uri: url,
}
```

Option 2: allow redirects from client using either the `uri` or `url` if redirect is true. [Redirect plugin](packages/better-auth/src/client/fetch-plugins.ts)

```
if ((context.data?.url || context.data?.uri) && context.data?.redirect) {
```

bytaesu:
Hi @dvanmali 😄

I think option 2 makes more sense!!

bytaesu:
> [!NOTE]
> We decided to go with this, even if it involves a breaking change. Expanding the global setup to allow a single exception could make the API more confusing later.

dvanmali:
Linking: #7558

PR #7812, opened 2026-02-05
fix(oauth): support case-insensitive email matching for social account linking

### LINKED ISSUE #7806: OAuth account linking fails when email casing differs between provider and existing account

### To Reproduce

1. Create an account using a standard email/password provider with a lowercase email: johndoe@email.com.
2. Attempt to sign in or link an account via an OAuth provider (e.g., Google or GitHub) where the primary email is returned with capital letters: JohnDoe@email.com.
3. Observe that the linking process fails or creates a duplicate user instead of linking to the existing record.

Account linking fails during the OAuth flow if the email address returned by the OAuth provider has different casing than the email address already stored in the database.

While the email addresses are logically the same (e.g., johndoe@email.com and JohnDoe@email.com), the system treats them as distinct, preventing the automatic or manual linking of the OAuth identity to the existing user profile.

1.4.13

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Pro",
    "release": "10.0.26200",
    "cpuCount": 12,
    "cpuModel": "AMD Ryzen 5 4600H with Radeon Graphics         ",
    "totalMemory": "31.37 GB",
    "freeMemory": "12.41 GB"
  },
  "node": {
    "version": "v24.13.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.2"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.1.5"
    },
    {
      "name": "react",
      "version": "19.1.0"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.17.2"
    },
    {
      "name": "@prisma/client",
      "version": "7.3.0"
    }
  ],
  "betterAuth": {
    "version": "1.4.13",
    "config": null
  }
}
```

Backend

```typescript
  database: prismaAdapter(db, {
    provider: "postgresql",
  }),
  emailAndPassword: {  
    enabled: true
  },
  socialProviders: {
    google: {
      clientId: process.env["AUTH_GOOGLE_ID"] as string,
      clientSecret: process.env["AUTH_GOOGLE_SECRET"] as string,
      disableSignUp: true,
    },
  },
});
```

### Additional context

_No response_

karuppusamy-d:
This ensures that social account linking succeeds even if the email
returned by the OAuth provider has different casing than the email
currently associated with the user.

- Update `linkSocialAccount` and `callbackOAuth` to use case-insensitive email comparison.
- Added comprehensive tests for both callback and idToken flows.
- Prevents "email_doesn't_match" errors due to case sensitivity.

Closes #7806

---
## Summary by cubic
Make email comparison case-insensitive during social account linking. This prevents mismatches when OAuth providers return different casing, so linking succeeds instead of showing "email_doesn't_match".

- **Bug Fixes**
  - Compare emails case-insensitively in linkSocialAccount, callbackOAuth, and the generic OAuth callback route.
  - Add tests for callback and idToken flows that cover casing differences.

<sup>Written for commit 199d42032f7b55f142a1e40090d5af99fa4f0562. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/account.ts
+++ b/packages/better-auth/src/api/routes/account.ts
@@ -309,7 +309,8 @@ export const linkSocialAccount = createAuthEndpoint(
 			}
 
 			if (
-				linkingUserInfo.user.email?.toLowerCase() !== session.user.email &&
+				linkingUserInfo.user.email?.toLowerCase() !==
+					session.user.email.toLowerCase() &&
 				c.context.options.account?.accountLinking?.allowDifferentEmails !== true
 			) {
 				throw APIError.from("UNAUTHORIZED", {
--- a/packages/better-auth/src/api/routes/callback.ts
+++ b/packages/better-auth/src/api/routes/callback.ts
@@ -187,7 +187,7 @@ export const callbackOAuth = createAuthEndpoint(
 			}
 
 			if (
-				userInfo.email?.toLowerCase() !== link.email &&
+				userInfo.email?.toLowerCase() !== link.email.toLowerCase() &&
 				c.context.options.account?.accountLinking?.allowDifferentEmails !== true
 			) {
 				return redirectOnError("email_doesn't_match");
--- a/packages/better-auth/src/oauth2/link-account.test.ts
+++ b/packages/better-auth/src/oauth2/link-account.test.ts
@@ -231,6 +231,9 @@ describe("oauth2 - email verification on link", async () => {
 	});
 });
 
+/**
+ * @see https://github.com/better-auth/better-auth/issues/7806
+ */
 describe("oauth2 - account linking with case insensitive email", async () => {
 	const { auth, client, cookieSetter } = await getTestInstance({
 		socialProviders: {
--- a/packages/better-auth/src/plugins/generic-oauth/routes.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/routes.ts
@@ -454,7 +454,7 @@ export const oAuth2Callback = (options: GenericOAuthOptions) =>
 				if (
 					ctx.context.options.account?.accountLinking?.allowDifferentEmails !==
 						true &&
-					link.email !== userInfo.email
+					link.email.toLowerCase() !== userInfo.email.toLowerCase()
 				) {
 					return redirectOnError("email_doesn't_match");
 				}


packages/better-auth/src/api/routes/callback.ts
```diff
@@ -187,7 +187,7 @@ export const callbackOAuth = createAuthEndpoint(
 			}
 
 			if (
-				userInfo.email !== link.email &&
+				userInfo.email?.toLowerCase() !== link.email &&
```
bytaesu:
```suggestion
				userInfo.email?.toLowerCase() !== link.email.toLowerCase() &&
```

packages/better-auth/src/api/routes/account.ts
```diff
@@ -309,7 +309,7 @@ export const linkSocialAccount = createAuthEndpoint(
 			}
 
 			if (
-				linkingUserInfo.user.email !== session.user.email &&
+				linkingUserInfo.user.email?.toLowerCase() !== session.user.email &&
```
bytaesu:
```suggestion
				linkingUserInfo.user.email?.toLowerCase() !== session.user.email.toLowerCase() &&
```

packages/better-auth/src/oauth2/link-account.test.ts
```diff
@@ -231,6 +231,178 @@ describe("oauth2 - email verification on link", async () => {
 	});
 });
 
+describe("oauth2 - account linking with case insensitive email", async () => {
```
bytaesu, line 237: Could you add this comment on the describe block? 🙏
```ts
/**
 * @see https://github.com/better-auth/better-auth/issues/7806
 */
```

karuppusamy-d:
> https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/plugins/generic-oauth/routes.ts#L457
> 
> There's one part missing 🙏

Thanks for the ping! I believe userInfo.email is already handled on #L432. Should we still add a manual conversion here as a redundant check, or are we good to leave it as is?

packages/better-auth/src/api/routes/callback.ts
karuppusamy-d: @bytaesu link.email is already in lowercase as it was from database. should we still add a manual conversion here?

packages/better-auth/src/api/routes/account.ts
karuppusamy-d: @bytaesu session.user.email will be in lowercase as it was from database.

bytaesu:
Hi @karuppusamy-d, thanks for working on this!

I think it's better to compare the two values in lowercase.

---

> https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/plugins/generic-oauth/routes.ts#L457

Also, while `link.email` is always lowercase via the internalAdapter, this isn't 100% guaranteed when using the community DB adapter or DB migration, so the issue could reoccur 🤔

what do you think?

karuppusamy-d:
Hi @bytaesu , 

Agreed, handling that edge case now will definitely save some headaches later.

I’ve pushed the update. Thanks for the catch.

PR #7827, opened 2026-02-06
fix(admin): apply listUsers filter when filterValue is defined

coderrshyam:
Previously, listUsers only applied filters when `filterValue` was truthy. This caused valid filters such as `false`, `0`, or empty strings to be ignored (e.g. filtering `banned = false` or `emailVerified = false`).

Updated the condition to explicitly check for `undefined`, ensuring all valid filter values are handled correctly.

Closes: #7837 

> I can see one pr is opened to release `1.4.19`, please make sure tha this fix should also be landed on the upcoming minor version

---
## Summary by cubic
Fix listUsers so filters apply when filterValue is defined, allowing falsy values like false, 0, and ''. Examples: banned=false and emailVerified=false.

- **Bug Fixes**
  - Use filterValue !== undefined; add a focused regression test (with temp user ban/unban) and clearer test name.

<sup>Written for commit 90b4e0e046fc924e120d66e84f185c1a069bf9fe. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/admin/admin.test.ts
+++ b/packages/better-auth/src/plugins/admin/admin.test.ts
@@ -374,7 +374,7 @@ describe("Admin plugin", async () => {
 	/**
 	 * @see https://github.com/better-auth/better-auth/issues/7837
 	 */
-	it("should allow filtering with falsy filterValue", async () => {
+	it("should apply filter when filterValue is defined", async () => {
 		// Create a dedicated user for this test to avoid invalidating shared sessions
 		const { data: tempUser } = await client.signUp.email({
 			email: "falsy-filter-test@test.com",


coderrshyam:
## Note for better-auth maintainers (separate from the fix)

While working with the admin APIs, I noticed a few edge cases and small issues that might be worth revisiting. None of these are blockers, but they can lead to confusing behavior in real applications.

- In `listUsers`, filtering previously depended on truthy checks. This can accidentally skip valid filters like `banned = false`, `emailVerified = false`, or numeric values such as `0`. <--> `fixed in this pr`

- `limit` and `offset` are coerced using `Number(value) || undefined`. If a client sends an invalid value, it gets silently ignored instead of failing fast, which makes pagination bugs harder to spot.

- `adminUpdateUser` allows updating any field via a generic `data` object. This makes it easy to accidentally modify fields like `id`, `createdAt`, or auth-related columns. A small allow-list could make this safer. This should be extend from `auth.ts`.

- In `createUser`, whether unauthenticated user creation is allowed depends on request context. Making this an explicit option would reduce the chance of accidental exposure.

- The impersonation flow assumes a specific cookie format when restoring the admin session. If the cookie is missing or malformed, this could fail in unexpected ways.

- `banUser` doesn’t check whether a user is already banned or whether an existing ban has expired, which can result in redundant updates.

- `userHasPermission` accepts both `permission` and `permissions`, but only one is actually used. Supporting a single shape would simplify the API and reduce confusion.

Overall, the implementation is solid — these are just small sharp edges that tend to show up when integrating this into a real admin dashboard.

coderrshyam:
Hey @himself65 @Bekacru @bytaesu 👋  
Could someone take a look at this when you have a moment?

I’m building an admin dashboard and was getting unexpected results from the users list. After debugging a bit, I found an issue in the filtering logic that skips valid values (like `false`). I’ve added a small fix for it.

Please fix test ad formatting issue.

bytaesu:
Hi @coderrshyam, I'm checking this 🙂

bytaesu:
Hi @coderrshyam,

Nice fix, thanks.
The test case wasn't reproducing the issue, so I've updated it 😄

coderrshyam:
Hi @bytaesu, thanks for the response.

Did you see this comment on the PR? - https://github.com/better-auth/better-auth/pull/7827#issuecomment-3861835700  
I’ve noted a few edge-case bugs there that showed up while building an admin dashboard.

One feature request I wanted to ask about:

Right now, filtering supports only a single condition, for example:
```
{ filterField: "role", filterValue: "user" }
```

In practice, it’s very common to filter by multiple conditions at once, like:
```
[
  { filterField: "role", filterValue: "user" },
  { filterField: "emailVerified", filterValue: true }
]
```

Currently this isn’t supported, which makes some dashboard use cases harder than necessary.  
Is multi-filter support something that’s planned or feasible to add?

Happy to help with a proposal or implementation if needed. Thanks!

PR #7839, opened 2026-02-07
fix: handle falsy filter values in admin listUsers API

### LINKED ISSUE #7837: Admin `listUsers` ignores valid falsy filter values

### To Reproduce

1. Enable the admin plugin in Better Auth.
2. Call the admin `listUsers` API with a filter that uses a falsy value.
   Example: `filterField = "banned", filterValue = false`.
3. Check the response from the API.

Current behavior:
When `filterValue` is a falsy value (such as false, 0, or an empty string),
the filter is ignored. The API behaves as if no filter was provided and
returns unfiltered results.

Expected behavior:
Filters should be applied whenever `filterValue` is defined, even if the
value itself is falsy. For example, filtering users with `banned = false`
or `emailVerified = false` should work correctly.

1.4.18

### System info

```bash
{
  "node": "v24.x.x",
  "os": "windows",
  "better-auth": "1.4.18"
}
```

Backend, Package

```typescript

  emailAndPassword: {  
    enabled: true
  },
  plugins: [
    admin() 
  ]
});
```

### Additional context

The issue comes from a truthy check on `ctx.query.filterValue` in the admin
`listUsers` route, which skips valid falsy values such as `false` or `0`.

Relevant code:
https://github.com/better-auth/better-auth/blob/dd6dc4601ff1c5a68f869fb285d3a9bd48074b66/packages/better-auth/src/plugins/admin/routes.ts#L658

Changing the condition to explicitly check for `undefined` fixes the issue.
A fix and related test update are included in PR #7827.

It would be good to have this fix included in the upcoming minor release as well.

liuxiaopai-ai:
## Summary

Fixes #7837

The `listUsers` API in the admin plugin was ignoring falsy filter values (`false`, `0`, `""`) because the code used a truthy check (`if (ctx.query?.filterValue)`) to determine whether to apply the filter. This caused valid filter values like `false`, `0`, and empty string to be silently skipped.

## Changes

- **`packages/better-auth/src/plugins/admin/routes.ts`**: Changed both `filterValue` and `searchValue` checks from truthy checks to strict `!== undefined && !== null` checks, so falsy values are properly handled.
- **`packages/better-auth/src/plugins/admin/admin.test.ts`**: Added test case covering `filterValue = false`, `filterValue = 0`, and `filterValue = ""` to ensure they are not ignored.

## Before

```typescript
if (ctx.query?.filterValue) {  // skips false, 0, ""
```

## After

```typescript
if (ctx.query?.filterValue !== undefined && ctx.query?.filterValue !== null) {  // only skips undefined/null
```

---
## Summary by cubic
Fixes listUsers in the admin plugin to correctly apply filters when values are falsy (false, 0, ""), so valid filters are no longer ignored and search behaves as expected.

- **Bug Fixes**
  - Use explicit undefined/null checks for filterValue and searchValue instead of truthy checks.
  - Add tests covering false, 0, and empty string to ensure filters are applied and prevent regressions.

<sup>Written for commit 47145e12d9a12689fe590987373564b68c5cad63. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/admin/routes.ts
+++ b/packages/better-auth/src/plugins/admin/routes.ts
@@ -647,7 +647,7 @@ export const listUsers = (opts: AdminOptions) =>
 
 			const where: Where[] = [];
 
-			if (ctx.query?.searchValue) {
+			if (ctx.query?.searchValue !== undefined && ctx.query?.searchValue !== null) {
 				where.push({
 					field: ctx.query.searchField || "email",
 					operator: ctx.query.searchOperator || "contains",
@@ -655,7 +655,7 @@ export const listUsers = (opts: AdminOptions) =>
 				});
 			}
 
-			if (ctx.query?.filterValue) {
+			if (ctx.query?.filterValue !== undefined && ctx.query?.filterValue !== null) {
 				where.push({
 					field: ctx.query.filterField || "email",
 					operator: ctx.query.filterOperator || "eq",
--- a/packages/better-auth/src/plugins/admin/admin.test.ts
+++ b/packages/better-auth/src/plugins/admin/admin.test.ts
@@ -437,6 +437,57 @@ describe("Admin plugin", async () => {
 		expect(res.data?.users[0]!.email).toBe("test@test.com");
 	});
 
+	it("should handle falsy filter values (false, 0, empty string)", async () => {
+		// filterValue = false should apply the filter, not be ignored
+		const resFalse = await client.admin.listUsers({
+			query: {
+				filterValue: false,
+				filterField: "emailVerified",
+				filterOperator: "eq",
+			},
+			fetchOptions: {
+				headers: adminHeaders,
+			},
+		});
+		// The filter should be applied (not ignored), so we get a filtered result
+		expect(resFalse.data).toBeDefined();
+		expect(resFalse.data?.users).toBeDefined();
+		// All returned users should have emailVerified === false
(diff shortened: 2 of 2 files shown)


packages/better-auth/src/plugins/admin/routes.ts
```diff
@@ -647,15 +647,15 @@ export const listUsers = (opts: AdminOptions) =>
 
 			const where: Where[] = [];
 
-			if (ctx.query?.searchValue) {
+			if (ctx.query?.searchValue !== undefined && ctx.query?.searchValue !== null) {
```
himself65, line 650:
```suggestion
			if (ctx.query?.searchValue != null) {
```

packages/better-auth/src/plugins/admin/routes.ts
```diff
@@ -647,15 +647,15 @@ export const listUsers = (opts: AdminOptions) =>
...
 				where.push({
 					field: ctx.query.searchField || "email",
 					operator: ctx.query.searchOperator || "contains",
 					value: ctx.query.searchValue,
 				});
 			}
 
-			if (ctx.query?.filterValue) {
+			if (ctx.query?.filterValue !== undefined && ctx.query?.filterValue !== null) {
```
himself65, line 658:
```suggestion
			if (ctx.query?.filterValue != null) {
```

bytaesu:
Duplicate of #7827

PR #7849, opened 2026-02-07
fix(oauth-provider): redirect with uri

dvanmali:
Redirects using the redirect plugin for return type.

```ts
{
    redirect: true,
    uri: string,
}
```

Alternative: #7811
Closes: #7807

---
## Summary by cubic
Update the redirect plugin to support responses that use uri, not just url. This fixes OAuth flows that return { redirect: true, uri } by redirecting to the correct location.

<sup>Written for commit 95dce467663987f29a46fa4bba96a4ff34475160. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/client/fetch-plugins.ts
+++ b/packages/better-auth/src/client/fetch-plugins.ts
@@ -5,11 +5,12 @@ export const redirectPlugin = {
 	name: "Redirect",
 	hooks: {
 		onSuccess(context) {
-			if (context.data?.url && context.data?.redirect) {
+			const redirectUri = context.data?.url ?? context.data?.uri;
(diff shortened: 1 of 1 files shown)


bytaesu:
I'm not sure if something misconfigured in the OAuth provider plugin should affect the global redirectPlugin in the first place. Let me think about it.

bytaesu:
If we do this, we'll end up mixing `url` and `uri`, which could implicitly tolerate inconsistency and cause more confusion.

bytaesu:
We decided to go with https://github.com/better-auth/better-auth/pull/7811.
Expanding the global setup to allow one exception could make the API more confusing later 🙏

PR #7851, opened 2026-02-08
fix(oauth-provider): add missing oauthClient createdAt/updatedAt values

dvanmali:
Adds missing `createdAt` and `updatedAt` values and fixes the `metadata` endpoint in some configurations.

---
## Summary by cubic
Adds missing createdAt/updatedAt for OAuth clients and completes provider metadata based on config. Fixes the metadata endpoint so it consistently advertises RFC 9207 and rounds timestamps to whole seconds.

- **New Features**
  - Metadata now includes public_client_supported, grant_types_supported, jwt_disabled; scopes_supported reflects configured scopes.
  - Advertises authorization_response_iss_parameter_supported; assertions moved to the metadata test suite.

- **Bug Fixes**
  - Set createdAt/updatedAt on client creation; update updatedAt on client updates and secret rotation.
  - Use floor-second precision for client_id_issued_at and timestamp conversions to match spec and tests.

<sup>Written for commit 687c4a749092894e90e0448fbd807f55ee030ba1. Summary will update on new commits.</sup>

### DIFF
--- a/packages/oauth-provider/src/oauthClient/endpoints.ts
+++ b/packages/oauth-provider/src/oauthClient/endpoints.ts
@@ -269,7 +269,10 @@ export async function updateClientEndpoint(
 					value: clientId,
 				},
 			],
-			update: oauthToSchema(updates),
+			update: {
+				...oauthToSchema(updates),
+				updatedAt: new Date(Math.floor(Date.now() / 1000) * 1000),
+			},
 		},
 	);
 	if (!updatedClient) {
@@ -351,8 +354,8 @@ export async function rotateClientSecretEndpoint(
 				},
 			],
 			update: {
-				...schemaToOAuth(client),
 				clientSecret: storedClientSecret,
+				updatedAt: new Date(Math.floor(Date.now() / 1000) * 1000),
 			},
 		},
 	);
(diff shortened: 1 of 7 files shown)


packages/oauth-provider/src/oauthClient/endpoints.ts
```diff
@@ -351,8 +354,8 @@ export async function rotateClientSecretEndpoint(
 				},
 			],
 			update: {
-				...schemaToOAuth(client),
 				clientSecret: storedClientSecret,
+				updatedAt: new Date(Math.floor(Date.now() / 1000) * 1000),
```
LovelessCodes, line 358: Is it not better to add to the `schema.ts`, so that it updates on database entry update instead?

packages/oauth-provider/src/register.ts
```diff
@@ -188,7 +188,11 @@ export async function createOAuthClientEndpoint(
 	});
 	const client = await ctx.context.adapter.create<SchemaClient<Scope[]>>({
 		model: "oauthClient",
-		data: schema,
+		data: {
+			...schema,
+			createdAt: new Date(iat * 1000),
```
LovelessCodes, line 193: Same goes for createdAt, I suppose - is it not better to add to the database schema as a defaultValue?

packages/oauth-provider/src/oauthClient/endpoints.ts
dvanmali: @LovelessCodes default values at `schema.ts` are best for fixed values, not those that change. Using a function needs to be handled by each adapter.

dvanmali:
@Bekacru Bump

PR #7855, opened 2026-02-08
fix(oauth-provider): resource indicators rfc 8707

### LINKED ISSUE #8298: Oauth Provider - Follow RFC 8707 Resource Implementation

### To Reproduce

Configure OAuth provider with resource indicators enabled (RFC 8707 implementation).

Initiate `/authorize` request with:

```http
GET /authorize?response_type=code
  &client_id=client
  &resource=https://api.example.com
```

Exchange authorization code at /token but:

Omit resource, or

Provide a different resource value (e.g., different order in array, or additional resource).

Observe token response.

OR (refresh flow):

Obtain refresh token issued with:

```json
{
  "resources": ["https://api.example.com"]
}
```

Call `/token` with `grant_type=refresh_token` and:

Request a new resource not present in the refresh token, or

Change the order of the resources array.

**Current behavior:**

JWT Access Tokens can change resource parameters later after authorize.

**Expected behavior:**

The strict enforcement is correct per RFC 8707.

>=1.4

Parent: #9178

dvanmali:
Follows the resource indicator spec [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html).

**Improvements**:
- Prevents resource value changes between `/authorize` and `/token`
- Restricts refresh and access tokens to resources specified at authorization
- `resource` supported across all grant types: authorization_code, client_credentials, refresh_token
- Access tokens with resource provides `aud` field at introspection
- Ensures consent re-prompting changes in resource

**Deprecations**:

- `customAccessTokenClaims` properly uses the `resources` field to indicate the resource at both `/token` and `/introspect` (`resource` field in this function is deprecated).

Follow the resource indicator spec RFC 8707.

Improvements:

- Prevents resource value changes between `/authorize` and `/token`
- Restricts refresh and access tokens to resources specified at authorization
- `resource` supported across all grant types: authorization_code, client_credentials, refresh_token
- Access tokens with resource provides `aud` field at introspection
- Ensures consent re-prompting changes in resource

Deprecations:

- `customAccessTokenClaims` properly uses the `resources` field to indicate the resource at both `/token` and `/introspect` (`resource` field in this function is deprecated).

Closes: #8298

After upgrading, users may also need to run one of the following to update the schema:

```bash
npx auth migrate
```
or
```bash
npx auth generate
```

---
## Summary by cubic
Implements RFC 8707 resource indicators in `@better-auth/oauth-provider` with strict URL validation (no fragments), consent-aware authorization, and resource-bound tokens across all grants. Adds `aud` to JWTs and to opaque token introspection, supports multiple resources, and enforces resource narrowing with safe fallbacks per spec.

- **New Features**
  - Accepts `resource` on `/authorize` and `/token` as string, array, or repeated param; invalid values return `invalid_target`. Validates against `validAudiences` and appends `userinfo` when `openid`.
  - Records requested resources in the authorization code and prevents widening at `/token`. Omitting `resource` falls back to the authorized set; provided values must be a subset. Refresh tokens retain the full original resources (RFC 8707 §2.2).
  - Consent saves selected resources and re-prompts if new ones are requested; `prompt=none` returns `consent_required`. Opaque tokens now expose `aud` via `/oauth2/introspect`. `customAccessTokenClaims` receives `resources` for both JWT and opaque tokens.

- **Migration**
  - Replace `customAccessTokenClaims.resource` with `customAccessTokenClaims.resources` (array).
  - Do not send `resource` to `/oauth2/introspect`; use `resources` via `customAccessTokenClaims` for audience-based claims.

<sup>Written for commit 0f4b2e6e9bc3a02347381144b275cd33ad9c5de2. Summary will update on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/oauth-provider.mdx
+++ b/docs/content/docs/plugins/oauth-provider.mdx
@@ -462,8 +462,7 @@ Important notes:
 * In OAuth 2.1, only `response_type: "code"` is supported.
 * `code_challenge_method: "plain"` will not be supported since this is a security vulnerability.
 * All authorization responses (success and error) include the `iss` parameter for issuer validation ([RFC 9207](https://datatracker.ietf.org/doc/html/rfc9207)).
-
-- Use the `resource` indicators to restrict tokens to a resource ([RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html)).
+* Use the `resource` indicators to restrict tokens to a resource ([RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html)).
 
 **State**
 
--- a/docs/content/docs/plugins/oauth-provider.mdx
+++ b/docs/content/docs/plugins/oauth-provider.mdx
@@ -462,6 +462,7 @@ Important notes:
 * In OAuth 2.1, only `response_type: "code"` is supported.
 * `code_challenge_method: "plain"` will not be supported since this is a security vulnerability.
 * All authorization responses (success and error) include the `iss` parameter for issuer validation ([RFC 9207](https://datatracker.ietf.org/doc/html/rfc9207)).
+- Use the `resource` indicators to restrict tokens to a resource ([RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html)).
 
 **State**
 
@@ -493,6 +494,9 @@ The token endpoint supports the following client authentication methods:
 * **`client_secret_post`** — Client credentials sent in the request body
 * **`private_key_jwt`** — Client authenticates with a signed JWT assertion ([RFC 7523](https://datatracker.ietf.org/doc/html/rfc7523))
 * **`none`** — Public client (PKCE required)
+
+Important Notes:
+- Use of the `resource` parameter results in a JWT access token (with JWT plugin enabled).
 
 #### Private Key JWT Authentication
 
@@ -615,7 +619,7 @@ Post login must be [configured](#post-login-screen) to perform post login select
 
 This endpoint provides details of the provided token. If the token is additionally tied to a session, the endpoint will ensure the session is `active`.
 
-To provide resource specific claims via `customAccessTokenClaims`, store the allowed resources that a confidential client can use in its `resources` field.
+Use the `resources` field in `customAccessTokenClaims` to add claims based on the audience.
 
 ### Revoke Endpoint
 
@@ -1079,9 +1083,10 @@ oauthProvider({
     };
   },
   // Attach claims to access tokens
-  customAccessTokenClaims: ({ user, scopes, referenceId, resource, metadata }) => {
+  customAccessTokenClaims: ({ user, scopes, referenceId, resources, metadata }) => {
     return {
       "https://example.com/org": referenceId,
+      "https://example.com/resources": resources,
       "https://example.com/roles": ["editor"],
     };
   },
--- a/docs/content/docs/plugins/oauth-provider.mdx
+++ b/docs/content/docs/plugins/oauth-provider.mdx
@@ -462,6 +462,7 @@ Important notes:
 * In OAuth 2.1, only `response_type: "code"` is supported.
 * `code_challenge_method: "plain"` will not be supported since this is a security vulnerability.
 * All authorization responses (success and error) include the `iss` parameter for issuer validation ([RFC 9207](https://datatracker.ietf.org/doc/html/rfc9207)).
+
 - Use the `resource` indicators to restrict tokens to a resource ([RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html)).
 
 **State**
@@ -496,7 +497,9 @@ The token endpoint supports the following client authentication methods:
 * **`none`** — Public client (PKCE required)
 
 Important Notes:
-- Use of the `resource` parameter results in a JWT access token (with JWT plugin enabled).
+
+* With the JWT plugin enabled (default), sending `resource` results in a JWT access token with the selected resource in the `aud` claim.
+* With `disableJwtPlugin: true`, access tokens remain opaque. Requested resources are still bound to the token and surfaced through `/oauth2/introspect` and `customAccessTokenClaims` (via `resources`).
 
 #### Private Key JWT Authentication
 
--- a/packages/oauth-provider/src/authorize.ts
+++ b/packages/oauth-provider/src/authorize.ts
@@ -7,7 +7,6 @@ import type { Verification } from "better-auth/db";
 import { APIError } from "better-call";
 import { oAuthState } from "./oauth";
 import type { OAuthErrorCode, OAuthRedirectOnError } from "./oauth-endpoint";
-import { checkResource } from "./token";
 import type {
 	OAuthAuthorizationQuery,
 	OAuthConsent,
@@ -17,6 +16,7 @@ import type {
 } from "./types";
 
 import {
+	checkResource,
 	getClient,
 	getJwtPlugin,
 	isPKCERequired,
--- a/packages/oauth-provider/src/authorize.ts
+++ b/packages/oauth-provider/src/authorize.ts
@@ -7,6 +7,7 @@ import type { Verification } from "better-auth/db";
 import { APIError } from "better-call";
 import { oAuthState } from "./oauth";
 import type { OAuthErrorCode, OAuthRedirectOnError } from "./oauth-endpoint";
+import { checkResource } from "./token";
 import type {
 	OAuthAuthorizationQuery,
 	OAuthConsent,
@@ -471,6 +472,26 @@ export async function authorizeEndpoint(
 		}
 	}
 
+	// Validate the resource sent to the authorize endpoint
+	const resource = query.resource;
+	try {
+		await checkResource(ctx, opts, resource, requestedScopes);
+	} catch (err) {
+		if (err instanceof APIError) {
+			return handleRedirect(
+				ctx,
+				formatErrorURL(
+					query.redirect_uri,
+					"invalid_target",
+					err?.message ?? err.body?.message ?? "invalid_resource",
+					query.state,
+					getIssuer(ctx, opts),
+				),
+			);
+		}
+		throw err;
+	}
+
 	// Check for session
 	const session = await getSessionFromCtx(ctx);
 	if (!session || promptSet?.has("login") || promptSet?.has("create")) {
@@ -615,6 +636,28 @@ export async function authorizeEndpoint(
 	if (
 		!consent ||
 		!requestedScopes.every((val) => consent.scopes.includes(val))
+	) {
+		if (promptNone) {
+			return redirectWithPromptNoneError(
+				ctx,
+				opts,
+				query,
+				"consent_required",
+				"End-User consent is required",
+			);
+		}
+		return redirectWithPromptCode(ctx, opts, "consent");
+	}
+
+	// Consent should be given for that resource if different than original consent
+	const consentedResources = consent?.resources;
+	const requestedResources =
+		typeof resource === "string" ? [resource] : resource;
+	if (
+		consentedResources &&
+		(!requestedResources ||
+			requestedResources.length === 0 ||
+			!requestedResources.every((v) => consentedResources.includes(v)))
 	) {
 		if (promptNone) {
 			return redirectWithPromptNoneError(
--- a/packages/oauth-provider/src/authorize.ts
+++ b/packages/oauth-provider/src/authorize.ts
@@ -474,22 +474,23 @@ export async function authorizeEndpoint(
 
 	// Validate the resource sent to the authorize endpoint
 	const resource = query.resource;
-	try {
-		await checkResource(ctx, opts, resource, requestedScopes);
-	} catch (err) {
-		if (err instanceof APIError) {
-			return handleRedirect(
-				ctx,
-				formatErrorURL(
-					query.redirect_uri,
-					"invalid_target",
-					err?.message ?? err.body?.message ?? "invalid_resource",
-					query.state,
-					getIssuer(ctx, opts),
-				),
-			);
-		}
-		throw err;
+	const resourceResult = await checkResource(
+		ctx,
+		opts,
+		resource,
+		requestedScopes,
+	);
+	if (!resourceResult.success) {
+		return handleRedirect(
+			ctx,
+			formatErrorURL(
+				query.redirect_uri,
+				"invalid_target",
+				"requested resource invalid",
+				query.state,
+				getIssuer(ctx, opts),
+			),
+		);
 	}
 
 	// Check for session
@@ -649,15 +650,17 @@ export async function authorizeEndpoint(
 		return redirectWithPromptCode(ctx, opts, "consent");
 	}
 
-	// Consent should be given for that resource if different than original consent
-	const consentedResources = consent?.resources;
-	const requestedResources =
-		typeof resource === "string" ? [resource] : resource;
+	// Prompt again whenever a requested resource is not covered by stored consent.
+	const requestedResources = Array.isArray(resource)
+		? resource
+		: resource
+			? [resource]
+			: [];
+	const consentedResources = consent?.resources ?? [];
 	if (
-		consentedResources &&
-		(!requestedResources ||
-			requestedResources.length === 0 ||
-			!requestedResources.every((v) => consentedResources.includes(v)))
+		requestedResources.some(
+			(requestedResource) => !consentedResources.includes(requestedResource),
+		)
 	) {
 		if (promptNone) {
 			return redirectWithPromptNoneError(
--- a/packages/oauth-provider/src/authorize.ts
+++ b/packages/oauth-provider/src/authorize.ts
@@ -22,6 +22,7 @@ import {
 	isPKCERequired,
 	parsePrompt,
 	storeToken,
+	toResourceList,
 } from "./utils";
 
 /**
@@ -492,6 +493,7 @@ export async function authorizeEndpoint(
 			),
 		);
 	}
+	const requestedResources = toResourceList(resource) ?? [];
 
 	// Check for session
 	const session = await getSessionFromCtx(ctx);
@@ -610,6 +612,7 @@ export async function authorizeEndpoint(
 			sessionId: session.session.id,
 			authTime: new Date(session.session.createdAt).getTime(),
 			referenceId,
+			resource: requestedResources,
 		});
 	}
 	const consent = await ctx.context.adapter.findOne<OAuthConsent<Scope[]>>({
@@ -651,11 +654,6 @@ export async function authorizeEndpoint(
 	}
 
 	// Prompt again whenever a requested resource is not covered by stored consent.
-	const requestedResources = Array.isArray(resource)
-		? resource
-		: resource
-			? [resource]
-			: [];
 	const consentedResources = consent?.resources ?? [];
 	if (
 		requestedResources.some(
@@ -681,6 +679,7 @@ export async function authorizeEndpoint(
 		sessionId: session.session.id,
 		authTime: new Date(session.session.createdAt).getTime(),
 		referenceId,
+		resource: requestedResources,
 	});
 }
 
@@ -709,6 +708,7 @@ async function redirectWithAuthorizationCode(
 		sessionId: string;
 		authTime: number;
 		referenceId?: string;
+		resource?: string[];
 	},
 ) {
 	const code = generateRandomString(32, "a-z", "A-Z", "0-9");
@@ -726,6 +726,7 @@ async function redirectWithAuthorizationCode(
 			sessionId: verificationValue?.sessionId,
 			referenceId: verificationValue.referenceId,
 			authTime: verificationValue.authTime,
+			resource: verificationValue.resource,
 		} satisfies VerificationValue),
 	};
 	await ctx.context.internalAdapter.createVerificationValue({
--- a/packages/oauth-provider/src/introspect.ts
+++ b/packages/oauth-provider/src/introspect.ts
@@ -218,6 +218,13 @@ async function validateOpaqueAccessToken(
 	const resources = Array.isArray(accessToken.resources)
 		? accessToken.resources
 		: undefined;
+	const audience = resources ? [...resources] : undefined;
+	if (audience?.length && accessToken.scopes?.includes("openid")) {
+		const userInfoEndpoint = `${ctx.context.baseURL}/oauth2/userinfo`;
+		if (!audience.includes(userInfoEndpoint)) {
+			audience.push(userInfoEndpoint);
+		}
+	}
 
 	// Add Custom Claims
 	const customClaims = opts.customAccessTokenClaims
@@ -240,7 +247,7 @@ async function validateOpaqueAccessToken(
 		...customClaims,
 		active: true,
 		iss: jwtPluginOptions?.jwt?.issuer ?? ctx.context.baseURL,
-		aud: toAudienceClaim(resources),
+		aud: toAudienceClaim(audience),
 		client_id: accessToken.clientId,
 		sub: user?.id,
 		sid: sessionId,
--- a/packages/oauth-provider/src/introspect.ts
+++ b/packages/oauth-provider/src/introspect.ts
@@ -20,6 +20,7 @@ import {
 	getStoredToken,
 	parseClientMetadata,
 	resolveSubjectIdentifier,
+	toAudienceClaim,
 	validateClientCredentials,
 } from "./utils";
 
@@ -214,6 +215,9 @@ async function validateOpaqueAccessToken(
 	if (accessToken.userId) {
 		user = await ctx.context.internalAdapter.findUserById(accessToken?.userId);
 	}
+	const resources = Array.isArray(accessToken.resources)
+		? accessToken.resources
+		: undefined;
 
 	// Add Custom Claims
 	const customClaims = opts.customAccessTokenClaims
@@ -221,7 +225,7 @@ async function validateOpaqueAccessToken(
 				user,
 				scopes: accessToken.scopes,
 				referenceId: accessToken?.referenceId,
-				resources: accessToken?.resources,
+				resources,
 				metadata: parseClientMetadata(client?.metadata),
 			})
 		: {};
@@ -236,6 +240,7 @@ async function validateOpaqueAccessToken(
 		...customClaims,
 		active: true,
 		iss: jwtPluginOptions?.jwt?.issuer ?? ctx.context.baseURL,
+		aud: toAudienceClaim(resources),
 		client_id: accessToken.clientId,
 		sub: user?.id,
 		sid: sessionId,
--- a/packages/oauth-provider/src/introspect.ts
+++ b/packages/oauth-provider/src/introspect.ts
@@ -221,6 +221,7 @@ async function validateOpaqueAccessToken(
 				user,
 				scopes: accessToken.scopes,
 				referenceId: accessToken?.referenceId,
+				resources: accessToken?.resources,
 				metadata: parseClientMetadata(client?.metadata),
 			})
 		: {};
@@ -296,7 +297,7 @@ async function validateRefreshToken(
 			where: [
 				{
 					field: "id",
-					value: refreshToken.sessionId,
+					value: sessionId,
 				},
 			],
 		});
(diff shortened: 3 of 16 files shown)


himself65:

packages/oauth-provider/src/token.ts
```diff
@@ -568,6 +585,36 @@ async function checkVerificationValue(
...
+		const authorizedSet = new Set(verificationResources);
+		for (const r of requestedSet) {
+			if (!authorizedSet.has(r)) {
+				throw new APIError("BAD_REQUEST", {
+					error_description: "requested resource not authorized",
+					error: "invalid_target",
+				});
+			}
+		}
+		if (requestedSet.size !== authorizedSet.size) {
```
gustavovalverde: RFC 8707 Section 2.2 says resources at the token endpoint can be a subset of what was authorized ("or a subset thereof"). The subset membership loop above is correct, but this size check forces an exact match, preventing legitimate downscoping.

Example: client authorizes with `resource=A&resource=B`, then exchanges the code with just `resource=A`. That should work, but this rejects it.

This block should be removed. The subset check alone is sufficient.

packages/oauth-provider/src/token.ts
```diff
@@ -1075,6 +1150,7 @@ async function handleRefreshTokenGrant(
 		undefined,
 		{
 			refreshToken,
+			resources: resources ?? refreshToken.resources,
```
gustavovalverde: RFC 8707 Section 2.2: "any refresh token that is returned is bound to the **full original grant**."

`createUserTokens()` uses this single `resources` value for both the access token and the new refresh token. If a client refreshes with `resource=A` when the original had `[A, B]`, the new refresh token only carries `[A]`. Each refresh narrows further until the client is locked to one resource.

This needs to be split: access token gets the requested subset, refresh token always keeps the full original set (`refreshToken.resources`).

packages/oauth-provider/src/authorize.ts
```diff
@@ -273,6 +274,25 @@ export async function authorizeEndpoint(
...
+	const resource = query.resource;
+	try {
+		await checkResource(ctx, opts, resource, requestedScopes);
+	} catch (err) {
+		if (err instanceof APIError) {
+			throw ctx.redirect(
+				formatErrorURL(
+					query.redirect_uri,
+					"invalid_target",
+					err.message,
```
gustavovalverde: `err.message` will be `undefined` here. `APIError` sets `this.message = body?.message`, but `checkResource()` throws with `{ error_description: "...", error: "..." }` — no `message` key. The redirect URL ends up as `error_description=undefined`.

Should be:
```typescript
err.body?.error_description ?? "invalid resource"
```

packages/oauth-provider/src/authorize.ts
```diff
@@ -385,6 +405,17 @@ export async function authorizeEndpoint(
 		return redirectWithPromptCode(ctx, opts, "consent");
 	}
 
+	// Consent should be given for that resource if different than original consent
+	const consentedResources = consent?.resources;
+	const requestedResources =
+		typeof resource === "string" ? [resource] : resource;
+	if (
+		consentedResources &&
+		!requestedResources?.every((v) => consentedResources.includes(v))
```
gustavovalverde: When `resource` is `undefined`, `requestedResources` is also `undefined`, so `undefined?.every(...)` returns `undefined`, and `!undefined` is `true`. This triggers a consent re-prompt even though the new request isn't asking for resources.

Add a guard:
```typescript
if (
    consentedResources &&
    requestedResources &&
    !requestedResources.every((v) => consentedResources.includes(v))
)
```

packages/oauth-provider/src/token.ts
```diff
@@ -568,6 +585,36 @@ async function checkVerificationValue(
...
 	}
+	const verificationResources =
+		typeof verificationValue.query.resource === "string"
+			? [verificationValue.query.resource]
+			: verificationValue.query.resource;
+	if (verificationResources) {
+		if (!resource) {
+			throw new APIError("BAD_REQUEST", {
+				error_description:
+					"resource not found in original authorization request",
```
gustavovalverde: This message is backwards. This branch fires when resources *were* in the authorization request but the token request *omits* them. Should be something like `"resource parameter required at token endpoint; was specified during authorization"`.

packages/oauth-provider/src/types/index.ts
```diff
@@ -788,6 +794,7 @@ export interface VerificationValue {
 	query: OAuthAuthorizationQuery;
 	sessionId: string;
 	userId: string;
+	resource?: string | string[];
```
gustavovalverde: Never populated or read. The code reads `verificationValue.query.resource` instead (since `OAuthAuthorizationQuery` now has `resource`). Either use this field or remove it.

packages/oauth-provider/src/types/index.ts
```diff
@@ -972,7 +983,7 @@ export interface OAuthRefreshToken<
 	Scopes extends readonly Scope[] = InternallySupportedScopes[],
 > {
 	token: string;
-	sessionId: string;
+	sessionId?: string;
 	userId: string;
 	referenceId?: string;
```
gustavovalverde, line 1126: Unrelated to resource indicators. Client credentials doesn't create refresh tokens (it blocks `offline_access`), so there's no case for a sessionless refresh token. Making this optional weakens downstream type safety. Should revert unless there's a specific reason for it.

packages/oauth-provider/src/token.ts
```diff
@@ -305,13 +310,14 @@ async function createRefreshToken(
...
 			authTime,
+			resources,
 			scopes,
 			createdAt: new Date(iat * 1000),
 			expiresAt: new Date(exp * 1000),
 		},
 	});
 	return {
-		id: refreshToken.id,
+		id: refreshToken.id as string,
```
gustavovalverde: If `refreshToken.id` isn't always a string, the cast masks the real issue. If it is, the cast is unnecessary.

packages/oauth-provider/src/oauth.ts
```diff
@@ -344,6 +344,9 @@ export const oauthProvider = <O extends OAuthOptions<Scope[]>>(options: O) => {
 						code_challenge: z.string().optional(),
 						code_challenge_method: z.enum(["S256"]).optional(),
 						nonce: z.string().optional(),
+						resource: z
```
gustavovalverde, line 394: RFC 8707 Section 2 requires `resource` values to be absolute URIs (RFC 3986 §4.3) without fragment components, and recommends against query components. This schema accepts any string.

Consider adding `.url()` or a `.refine()` that checks:
- Value is an absolute URI
- No fragment (`#`) component

Without this, malformed values only fail if they're not in `validAudiences`, which gives a less helpful error message.

packages/oauth-provider/src/token.ts
```diff
@@ -320,12 +326,12 @@ async function createRefreshToken(
  * Checks the resource parameter, if provided,
  * and returns a valid audience based on the request
  */
-async function checkResource(
+export async function checkResource(
 	ctx: GenericEndpointContext,
 	opts: OAuthOptions<Scope[]>,
+	resource: string | string[] | undefined,
```
gustavovalverde: Per RFC 8707 Section 2, resource values MUST be absolute URIs without fragment components. `checkResource()` validates against `validAudiences` but never checks URI format. A resource like `https://example.com/api#fragment` or a bare string `not-a-uri` would pass through to the audience check silently.

Consider validating before the audience loop:
```typescript
for (const aud of audience) {
    const parsed = new URL(aud); // throws on non-absolute URI
    if (parsed.hash) {
            error_description: "resource must not contain a fragment",
            error: "invalid_target",
        });
    }
}
```

gustavovalverde (changes requested):
Reviewed the full diff against [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html). Inline comments on specific locations below.

**Blockers:**

1. **Exact-match at token exchange is too strict.** RFC 8707 Section 2.2 allows the token request to use a *subset* of authorized resources. The size-equality check in `checkVerificationValue()` prevents that.

2. **Refresh tokens lose their original resource binding.** RFC 8707 Section 2.2: "any refresh token that is returned is bound to the full original grant." Currently, if a client refreshes with a narrower resource set, the new refresh token only carries the narrowed set, permanently losing the rest.

3. **`err.message` is `undefined` in the authorize redirect.** `APIError` sets `this.message` from `body?.message`, but the body has `error_description`, not `message`. The redirect URL ends up with `error_description=undefined`.

4. **No URI format validation.** RFC 8707 Section 2 requires resource values to be absolute URIs (RFC 3986) without fragment components. Neither the zod schema nor `checkResource()` enforces this.

5. **No tests.** This touches authorization, token exchange, refresh, consent, introspection, and schema. Needs test coverage, especially for the subset/superset boundary conditions.

**Non-blocking but should fix:**

6. Consent re-prompt fires incorrectly when the new request has no `resource` param but a previous consent stored resources.
7. `VerificationValue.resource` field added to the type but never populated or read.
8. `OAuthRefreshToken.sessionId` made optional, unrelated to this PR.
9. Error message in the missing-resource check says the opposite of what happened.

packages/oauth-provider/src/types/index.ts
dvanmali: Value is used in `authorize.ts` L321

packages/oauth-provider/src/types/index.ts
dvanmali: Unrelated but sessionless tokens are stored on the database according to `schema.ts` L182. I can make it a separate `chore` pr but it's small and unnecessary.

packages/oauth-provider/src/token.ts
dvanmali: Current type is `any`, but will move to type into adapter line though

packages/oauth-provider/src/oauth.ts
dvanmali: Will perform for all `SafeUrlSchema` values

packages/oauth-provider/src/token.ts
dvanmali: Adding this check via zod parsing `SafeUrlSchema`

packages/oauth-provider/src/token.ts
dvanmali: Great catch!!!

dvanmali:
@cubic-dev-ai @claude review

packages/oauth-provider/src/token.ts
```diff
@@ -572,6 +595,29 @@ async function checkVerificationValue(
 			error: "invalid_request",
 		});
 	}
+	const verificationResources =
+		typeof verificationValue.query.resource === "string"
+			? [verificationValue.query.resource]
+			: verificationValue.query.resource;
+	if (verificationResources) {
+		if (!resource) {
```
gustavovalverde: RFC 8707 §2.2 says when `resource` is omitted at the token endpoint but was specified during authorization, the AS **SHOULD** fall back to the authorized resources; not reject.

> "If the client does not specify a resource parameter when making an access token request, the authorization server SHOULD audience-restrict issued access tokens to the resource(s) indicated by the authorization request."

The `verificationResources` are already available here. Instead of throwing, fall back:

```ts
const effectiveResources = resource ?? verificationResources;
```

Then only validate the subset check when both are present.

packages/oauth-provider/src/authorize.ts
```diff
@@ -519,6 +540,27 @@ export async function authorizeEndpoint(
 		return redirectWithPromptCode(ctx, opts, "consent");
 	}
 
+	// Consent should be given for that resource if different than original consent
+	const consentedResources = consent?.resources;
+	const requestedResources =
+		typeof resource === "string" ? [resource] : resource;
+	if (
```
gustavovalverde, line 658: This only re-prompts when `requestedResources` is truthy. If a user previously consented to `resource=https://api-a` and the client later requests the same scopes **without** `resource`, this branch is skipped entirely. The old consent row is reused, and `createUserTokens` gets `resources === undefined`; silently dropping the audience restriction.

This is a security gap: a client can widen its token scope by omitting a parameter that was previously consented to.

packages/oauth-provider/src/introspect.ts
```diff
@@ -220,6 +220,7 @@ async function validateOpaqueAccessToken(
 				user,
 				scopes: accessToken.scopes,
 				referenceId: accessToken?.referenceId,
+				resources: accessToken?.resources,
```
gustavovalverde: The `resources` are passed to `customAccessTokenClaims` here, but the introspection response itself (lines 233–243) doesn't include `aud`. RFC 7662 §2.2 explicitly lists `aud` as an optional response field.

Without it, resource servers validating opaque tokens via introspection can't verify the token was issued for them; any RS with introspect credentials can validate tokens meant for a different RS.

Add `aud` from the stored resources to the response object.

packages/oauth-provider/src/types/index.ts
```diff
@@ -818,6 +824,7 @@ export interface VerificationValue {
 	query: OAuthAuthorizationQuery;
 	sessionId: string;
 	userId: string;
+	resource?: string | string[];
```
gustavovalverde: This field is declared but never written. `redirectWithAuthorizationCode` stores the resource inside `query.resource` (via `verificationValue.query`), not at this top level. The token endpoint reads it from `verificationValue.query.resource` (token.ts:598), never from `verificationValue.resource`.

Remove this dead field; it creates confusion about which is the source of truth.

packages/oauth-provider/src/types/index.ts
```diff
@@ -1004,7 +1015,7 @@ export interface OAuthRefreshToken<
 	Scopes extends readonly Scope[] = InternallySupportedScopes[],
 > {
 	token: string;
-	sessionId: string;
+	sessionId?: string;
```
gustavovalverde: This is unrelated to resource indicators; making `sessionId` optional changes the type contract for all `OAuthRefreshToken` consumers. If this is needed for client_credentials (which don't have sessions), it should be in a separate commit with its own justification.

packages/oauth-provider/src/oauth.ts
```diff
@@ -675,7 +688,7 @@ export const oauthProvider = <O extends OAuthOptions<Scope[]>>(options: O) => {
 												resource: {
 													type: "string",
 													description:
-														"Requested token resource (ie audience) to obtain a JWT formatted access token",
+														"Requested token resource(s) (ie audience) to obtain a JWT formatted access token",
```
gustavovalverde, line 732: The OpenAPI spec still says `type: "string"` for `resource` on the token endpoint, but the Zod schema (line 634) accepts `z.union([SafeUrlSchema, z.array(SafeUrlSchema).min(1)])`. Those should match: use `oneOf` with string and array, or at minimum document that it accepts both.

gustavovalverde (changes requested):
A few things need addressing before merge:

 **Blockers:**
- Token endpoint must fall back to authorized resources when `resource` is omitted (RFC 8707 §2.2), not reject
- Consent-widening gap: omitting `resource` on a subsequent request silently drops audience restriction
- No tests — project policy requires them for new features

**Should fix:**
- Opaque token introspection needs `aud` in the response for resource servers
- Dead `VerificationValue.resource` field (never written)
- OpenAPI type mismatch on token endpoint `resource` field
- `sessionId` optionality change is unrelated — separate it

**Note:** This PR will need a rebase after #9060; that fixes the query serialization layer so multi-valued `resource` params survive prompt redirects (login, consent, account selection).

packages/oauth-provider/src/types/index.ts
dvanmali: Needed for token.ts L316

dvanmali:
@gustavovalverde apologies for the wait, was out for the weekend. Updated

dvanmali:
@gustavovalverde any update for this?

dvanmali:
@gustavovalverde this should land in next minor

gustavovalverde:
@dvanmali, yes. There are a few PRs related to OAuth overall that will be merged alongside this one (hopefully for the next beta release). I have some notes in my local, I'll push those in a while.

packages/oauth-provider/src/oauth.ts
```diff
@@ -391,6 +391,9 @@ export const oauthProvider = <O extends OAuthOptions<Scope[]>>(options: O) => {
 							.pipe(z.enum(["S256"]))
 							.optional(),
 						nonce: z.string().optional(),
+						resource: z
```
gustavovalverde, line 394: Add `resource: "invalid_target"` to the `errorCodesByField` table at line 413 below. Post-#9277, zod failures map through `createOAuthEndpoint`'s field table; without a mapping for `resource` a malformed URL emits `invalid_request` via the default, which contradicts RFC 8707 §2.

packages/oauth-provider/src/oauth.ts
```diff
@@ -643,7 +654,9 @@ export const oauthProvider = <O extends OAuthOptions<Scope[]>>(options: O) => {
 						code_verifier: z.string().optional(),
 						redirect_uri: SafeUrlSchema.optional(),
 						refresh_token: z.string().optional(),
-						resource: z.string().optional(),
+						resource: z
+							.union([SafeUrlSchema, z.array(SafeUrlSchema).min(1)])
+							.optional(),
 						scope: z.string().optional(),
 					}),
 					errorCodesByField: {
```
gustavovalverde, line 663: Same as /authorize: add `resource: "invalid_target"` here so a bad resource on the token endpoint emits the RFC-mandated code.

packages/oauth-provider/src/authorize.ts
```diff
@@ -471,6 +472,26 @@ export async function authorizeEndpoint(
...
+	try {
+		await checkResource(ctx, opts, resource, requestedScopes);
+	} catch (err) {
+		if (err instanceof APIError) {
+			return handleRedirect(
+				ctx,
+				formatErrorURL(
+					query.redirect_uri,
+					"invalid_target",
+					err?.message ?? err.body?.message ?? "invalid_resource",
```
gustavovalverde: better-call's `APIError` sources `this.message` from `body.message`, and `checkResource` throws with only `error_description` and `error`, so this chain always collapses to the `"invalid_resource"` fallback regardless of which rule failed. Rework `checkResource` to return a result (audience on success, tagged error on failure) and build `formatErrorURL` here with a literal description, matching the pattern the other handler-level redirects in this file already use.

packages/oauth-provider/src/authorize.ts
```diff
@@ -628,6 +649,28 @@ export async function authorizeEndpoint(
...
+			return redirectWithPromptNoneError(
+				ctx,
+				opts,
+				query,
+				"consent_required",
+				"End-User consent is required",
+			);
+		}
+		return redirectWithPromptCode(ctx, opts, "consent");
+	}
```
gustavovalverde: `consentedResources &&` skips the re-prompt whenever prior consent stored no resources. A user who consented before this feature existed, or whose client never asked for a resource, can then have audience-bound tokens issued for any allow-listed resource without a consent screen. Re-prompt whenever requested resources aren't a subset of the stored consent's resources, including the empty-stored case:

```ts
const requested = Array.isArray(resource) ? resource : resource ? [resource] : [];
const consented = consent?.resources ?? [];
if (requested.some((r) => !consented.includes(r))) { /* prompt */ }
```

packages/oauth-provider/src/authorize.ts
```diff
@@ -628,6 +649,28 @@ export async function authorizeEndpoint(
...
 	}
 
+	// Consent should be given for that resource if different than original consent
+	const consentedResources = consent?.resources;
+	const requestedResources =
+		typeof resource === "string" ? [resource] : resource;
+	if (
+		consentedResources &&
+		(!requestedResources ||
+			requestedResources.length === 0 ||
```
gustavovalverde: Unreachable: the zod schema on /authorize uses `.min(1)` on the array union, so an empty array is rejected at parse time. Drop.

packages/oauth-provider/src/introspect.ts
```diff
@@ -221,6 +221,7 @@ async function validateOpaqueAccessToken(
...
```
gustavovalverde: `resources` is now persisted on opaque tokens and surfaced to `customAccessTokenClaims`, but the introspection envelope at line 235 below doesn't include `aud`. Resource servers use `/oauth2/introspect` to verify audience binding, so without `aud` they can't. Add:

```ts
aud: accessToken.resources?.length === 1
    ? accessToken.resources[0]
    : accessToken.resources,
```

Mirrors the single-vs-array shape `createJwtAccessToken` uses.

packages/oauth-provider/src/types/index.ts
```diff
@@ -921,6 +927,7 @@ export interface VerificationValue {
 	query: OAuthAuthorizationQuery;
 	sessionId: string;
 	userId: string;
+	resource?: string | string[];
```
gustavovalverde: This field is never written (`redirectWithAuthorizationCode` doesn't set it) and never read (consumers go through `query.resource`). Remove before it invites divergent writes.

packages/oauth-provider/src/types/index.ts
```diff
@@ -1112,7 +1123,7 @@ export interface OAuthRefreshToken<
 	Scopes extends readonly Scope[] = InternallySupportedScopes[],
 > {
 	token: string;
-	sessionId: string;
+	sessionId?: string;
```
gustavovalverde: Unrelated to resource indicators; the schema already has `sessionId: { required: false }`, so this is a pre-existing type/schema mismatch. Split into its own PR with its own changeset so the history for each change is legible.

packages/oauth-provider/src/types/index.ts
```diff
@@ -562,7 +562,9 @@ export interface OAuthOptions<
 		referenceId?: string;
 		/** Scopes granted for this token */
 		scopes: Scopes;
-		/** The resource requesting. Provided by the token endpoint. */
+		/** The resources requested. */
+		resources?: string[];
+		/** The resource requesting. Provided by the token endpoint. @deprecated */
 		resource?: string;
```
gustavovalverde: `resource?: string` no longer matches the runtime value: `ctx.body.resource` is `string | string[] | undefined`, and `token.ts` passes it through unchanged. Callers still using the deprecated field silently receive an array where they expect a string. Drop the field and update in-tree callers; the changeset already announces the deprecation, and we're on `next`.

packages/oauth-provider/src/token.ts
```diff
@@ -375,12 +383,12 @@ async function createRefreshToken(
  * Checks the resource parameter, if provided,
  * and returns a valid audience based on the request
  */
-async function checkResource(
+export async function checkResource(
```
gustavovalverde: Four responsibilities in one: scalar-to-array normalization, openid-scope augmentation with `/oauth2/userinfo`, allow-list validation, per-JWT-claim single-vs-array flattening. Split:

```ts
normalizeResources(resource: string | string[] | undefined): string[] | undefined
deriveAudience(resources: string[], scopes: string[], baseURL: string): string[]
assertResourcesAllowed(resources: string[], allowed: Set<string>): void
```

Return `string[]` from all three and flatten once in `createJwtAccessToken`. Today the flatten is done in two places and they can drift.

docs/content/docs/plugins/oauth-provider.mdx
```diff
@@ -494,6 +495,9 @@ The token endpoint supports the following client authentication methods:
 * **`private_key_jwt`** — Client authenticates with a signed JWT assertion ([RFC 7523](https://datatracker.ietf.org/doc/html/rfc7523))
 * **`none`** — Public client (PKCE required)
 
+Important Notes:
+- Use of the `resource` parameter results in a JWT access token (with JWT plugin enabled).
```
gustavovalverde: Contradicts the `disableJwtPlugin: true` section further down. Cover both paths: with the JWT plugin enabled, `resource` produces a JWT access token with `aud`; with `disableJwtPlugin: true` the token stays opaque but is bound to `resources` and surfaces them via `/oauth2/introspect` and `customAccessTokenClaims`. The second half depends on the opaque-introspection `aud` fix.

packages/oauth-provider/src/token.test.ts
```diff
@@ -360,6 +360,52 @@ describe("oauth token - authorization_code", async () => {
 		expect(accessToken.payload.exp).toBe(tokens.data?.expires_at);
 		expect(accessToken.payload.scope).toBe(scopes.join(" "));
 	});
+
+	it("should fall back to authorized resource when token request omits resource", async ({
```
gustavovalverde, line 364: One happy path covered here. Cases to add, all with the existing `createAuthUrl` / `validateAuthCode` / `createRefreshAccessTokenRequest` helpers:

- /authorize: invalid resource URL (fragment, dangerous scheme) emits `error=invalid_target` on the RP redirect with a non-empty `error_description`.
- /authorize: multiple `resource` values persist onto `OAuthConsent.resources`.
- /authorize: `prompt=none` with a resource not covered by prior consent returns `consent_required`.
- /authorize: prior consent without `resources`, new request with `resource`, re-prompts.
- /token authorization_code: resource is a proper subset of authorized, access token `aud` narrows, refresh token keeps the full set.
- /token authorization_code: resource is a superset, emits `invalid_target`.
- /token authorization_code: resource disjoint, emits `invalid_target`.
- /token client_credentials: `resource` produces JWT with `aud`; disallowed resource emits `invalid_target`.
- /token refresh: chained narrowing. authorize([a,b]) → refresh(a) → refresh() returns [a,b]. The §2.2 full-original-grant guard.
- /token refresh: body resource not in refresh token's resources emits `invalid_target`.
- /introspect: opaque access token response includes `aud`.
- /introspect JWT + opaque: `customAccessTokenClaims` receives `resources`.
- zod: `SafeUrlSchema` rejects `https://api.example.com#frag`.

gustavovalverde (changes requested):
Remaining tasks before merging: 

1. Standardize terminology by replacing `validAudiences` with `allowedResources` (deprecated alias included) and using `resources` across the API, reserving `aud` solely for the JWT claim.
2. Implement per-client allow-listing in accordance with RFC 8707 §2.1, by adding `resources?: string[]` to `SchemaClient` and ensuring `checkResource` intersects with `opts.validAudiences`; also update `oauthClient.fields` accordingly.
3. Expand `requestUriResolver`'s return type from `Promise<Record<string, string> | null>` to `Promise<Record<string, string | string[]> | null>` to support multi-valued `resource` entries per RFC 9126.

packages/oauth-provider/src/authorize.ts
dvanmali: Using previous suggestion

packages/oauth-provider/src/types/index.ts
dvanmali: It should be added to the verification value, something got removed. Will reintroduce to `redirectWithAuthorizationCode`

packages/oauth-provider/src/types/index.ts
dvanmali: Ok, was unsure how this would've been handled

dvanmali:
@gustavovalverde done

packages/oauth-provider/src/types/index.ts
dvanmali: Moved to #9324

packages/oauth-provider/src/types/index.ts
```diff
@@ -562,8 +562,8 @@ export interface OAuthOptions<
 		referenceId?: string;
 		/** Scopes granted for this token */
 		scopes: Scopes;
```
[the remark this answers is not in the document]
dvanmali: Prior review mentioned to remove this entirely so I did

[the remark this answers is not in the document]
gustavovalverde: @dvanmali but the changesets indicates it's deprecated https://github.com/better-auth/better-auth/pull/7855/changes#diff-224cbef66a78ae0b798ab93471c6b28c883dc2a681b79bc4e4c2cf95a8b29801R15-R17 (which is different from full removal)

dvanmali:
@gustavovalverde bump

bytaesu:
Hi @dvanmali 😁
Gustavo is currently busy due to personal reasons. I'll let him know to check this PR!

dvanmali:
Sounds good I'll wait. Thanks for letting me know. Could you merge the other PR, I tagged you in though. It'll assist for this PR. @bytaesu

dvanmali:
@gustavovalverde @bytaesu

PR #7860, opened 2026-02-08
fix(core): coerce string where values to match field types in adapter

bytaesu:
> [!NOTE]
> You can see this issue by using `filterValue` in `authClient.admin.listUsers`
> https://www.better-auth.com/docs/plugins/admin#list-users 

---
## Summary by cubic
Coerces string where values in the DB adapter to match schema types (boolean/number), including arrays, and ignores empty strings. This makes filters from HTTP query params work consistently across adapters.

- **Bug Fixes**
  - Convert "true"/"false" and numeric strings (and arrays) to typed values based on schema before queries.
  - Guard against empty strings when coercing numbers.
  - Preserve boolean-to-1/0 mapping when the adapter doesn’t support booleans; added tests for boolean, number, and number array "in" coercion in findMany.

<sup>Written for commit a62f5a289eeb17a813e09a23b141dd8425e33e87. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/db/db.test.ts
+++ b/packages/better-auth/src/db/db.test.ts
@@ -100,16 +100,20 @@ describe("db", async () => {
 	it("should coerce string where values to match field types", async () => {
 		// HTTP query params arrive as strings.
 		// The adapter should coerce values to match the field's schema type.
-		const { db } = await getTestInstance({
-			user: {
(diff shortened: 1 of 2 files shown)


bytaesu:
Boolean arrays are not coerced because ["true", "false"] is pointless 😅

PR #7870, opened 2026-02-09
fix(sso): only call `provisionUser` for new users

### LINKED ISSUE #7857: SSO provisionUser is called on every sign-in, not just for new users

### To Reproduce

1. Configure the SSO plugin with a `provisionUser` callback:
   ```ts
   sso({
     provisionUser: async ({ user, userInfo, token, provider }) => {
       console.log("provisionUser called for", user.email);
     },
   })
   ```
2. Sign in via SSO (OIDC or SAML) with a new user — `provisionUser` is called (expected).
3. Sign in again with the same user — `provisionUser` is called again (unexpected).

**Current behavior:** `provisionUser` is called on **every** SSO sign-in, regardless of whether the user already exists. This can be confirmed in `packages/sso/src/routes/sso.ts` at lines 1668, 2241, and 2700 — all three call sites invoke `provisionUser` unconditionally after `handleOAuthUserInfo` returns.

**Expected behavior:** According to the [documentation](https://www.better-auth.com/docs/plugins/sso): *"When a user is authenticated, if the user does not exist, the user will be provisioned using the `provisionUser` function."* — `provisionUser` should only be called when the user is newly created.

`handleOAuthUserInfo` (in `packages/better-auth/src/oauth2/link-account.ts`) already returns an `isRegister` flag that is `true` only when the user didn't previously exist. This flag is used for redirect logic (line 1698) but is not checked before calling `provisionUser`.

The fix would be to gate `provisionUser` on `isRegister`, e.g.:
```ts
if (options?.provisionUser && linked.isRegister) {
```

1.4.17

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.6.0: Wed Nov  5 21:33:58 PST 2025; root:xnu-11417.140.69.705.2~1/RELEASE_ARM64_T6000",
    "release": "24.6.0",
    "cpuCount": 10,
    "cpuModel": "Apple M1 Pro",
    "totalMemory": "32.00 GB",
    "freeMemory": "0.21 GB"
  },
  "node": {
    "version": "v22.22.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.4"
  },
  "frameworks": [
    {
      "name": "hono",
      "version": "4.6.12"
    }
  ],
  "databases": [
    {
      "name": "postgres",
      "version": "3.4.7"
    },
    {
      "name": "drizzle",
      "version": "0.44.2"
    }
  ],
  "betterAuth": {
    "version": "1.4.17",
    "config": null
  }
}
```

Backend

```typescript

  plugins: [
    sso({
      provisionUser: async ({ user, userInfo, token, provider }) => {
        // This is called on every sign-in, not just for new users
      },
    }),
  ],
});
```

### Additional context

There is a contradiction in the docs: the main description says `provisionUser` is only for new users, but the "Best Practices" section says *"Make sure your provisioning functions can be safely run multiple times"* — implying it may be called repeatedly. If the every-sign-in behavior is intentional, the initial description and the function name are misleading and the docs should be updated. If it's a bug, the three call sites in `packages/sso/src/routes/sso.ts` need to be gated on `isRegister`.

bytaesu:
- Closes #7857

---
## Summary by cubic
Call provisionUser only when a user is first created via SSO, not on every sign-in. This avoids duplicate provisioning and side effects on repeat logins.

- **Bug Fixes**
  - Gate provisionUser behind isRegister in callbackSSO, callbackSSOSAML, and acsEndpoint.
  - Add test to verify it runs on first sign-in and not on subsequent sign-ins.

<sup>Written for commit 3e1952c50e99c2af08050551a3b773e206093977. Summary will update on new commits.</sup>

### DIFF
--- a/packages/sso/src/oidc.test.ts
+++ b/packages/sso/src/oidc.test.ts
@@ -3,7 +3,7 @@ import { createAuthClient } from "better-auth/client";
 import { organization } from "better-auth/plugins";
 import { getTestInstance } from "better-auth/test";
 import { OAuth2Server } from "oauth2-mock-server";
-import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
 import { sso } from ".";
 import { ssoClient } from "./client";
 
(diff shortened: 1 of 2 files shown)


bytaesu:
@Paola3stefania Could you update the SSO docs consistently and push the commits to this PR? 🙏

PR #7885, opened 2026-02-10
fix: safely coerce date values from DB in OAuth provider plugin

### LINKED ISSUE #7819: OAuth Provider plugin assumes `Date` objects from DB without validation with Drizzle

### To Reproduce

1. Set up better-auth with the Drizzle adapter + SQLite + `oauthProvider` plugin
2. Define the Drizzle schema for `oauth_access_token` using `text("expires_at")` or `integer("expires_at", { mode: "timestamp" })` for date columns
3. Complete an OAuth authorization flow (authorization code grant without `resource` parameter, so opaque tokens are created)
4. The callback to the relying party calls the `/api/auth/oauth2/userinfo` endpoint
5. `validateOpaqueAccessToken` reads the access token from the DB and calls `accessToken.expiresAt.getTime()` — **crashes with `TypeError`**

Both `text()` and `integer(..., { mode: "timestamp" })` are reasonable Drizzle column types for dates — only `mode: "timestamp_ms"` happens to work because it's the only mode that returns `Date` objects with millisecond precision, which is what the plugin assumes without checking.

**Current behavior:**

The oauth-provider plugin calls `.getTime()` directly on values read from the database (at least 8 call sites across `introspect.ts`, `register.ts`, `token.ts`) without any type guard or coercion. This creates a hard dependency on the Drizzle schema using exactly `integer('col', { mode: 'timestamp_ms' })` — any other column type for dates crashes at runtime:

```
TypeError: accessToken.expiresAt.getTime is not a function
```

The Drizzle adapter doesn't set `supportsDates`, so it defaults to `true` — meaning the adapter framework skips Date↔string conversion and trusts the ORM. But whether Drizzle actually returns `Date` objects depends entirely on how the user defined their schema. There's no validation that this implicit contract is met.

**Expected behavior:**

Either:

- (a) The plugin should defensively coerce values to `Date` before calling `.getTime()` (e.g., `new Date(value).getTime()`), or
- (b) The Drizzle adapter should explicitly set `supportsDates` based on the database type (like the Kysely adapter does — it sets `false` for SQLite), or
- (c) A dev-mode warning should detect when date fields return non-Date values

1.5.0-beta.8

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.2.0",
    "cpuModel": "Apple M4 Pro"
  },
  "frameworks": [
    { "name": "next", "version": "16.1.1" },
  ],
  "databases": [
    { "name": "@libsql/client", "version": "0.15.15" },
    { "name": "drizzle", "version": "0.45.1" }
  ],
  "betterAuth": { "version": "1.5.0-beta.8" }
}
```

Backend, Package

```typescript

  database: drizzle(db, { schema }), // SQLite via @libsql/client
  plugins: [
    oauthProvider({
      // ... provider config
    }),
  ],
});
```

### Additional context

There are three contributing factors:

**1. Drizzle adapter doesn't set `supportsDates` for SQLite**

The Kysely adapter correctly sets `supportsDates: false` for SQLite:

```typescript
// kysely-adapter.ts:624-626
supportsDates:
    config?.type === "sqlite" || config?.type === "mssql" || !config?.type
        ? false : true,
```

The Drizzle adapter omits `supportsDates` entirely, inheriting the default `true`. This is technically correct when the schema uses `integer` + `timestamp_ms` (Drizzle returns proper `Date` objects), but creates a brittle implicit contract that nothing validates.

**2. Plugin code calls `.getTime()` without type guards**

At least 8 call sites in the oauth-provider plugin assume Date objects without checking:

- `introspect.ts:239` — `accessToken.expiresAt.getTime()`
- `introspect.ts:240` — `accessToken.createdAt.getTime()`
- `introspect.ts:326` — `refreshToken.expiresAt.getTime()`
- `introspect.ts:327` — `refreshToken.createdAt.getTime()`
- `register.ts:359` — `expiresAt.getTime()`
- `register.ts:362` — `createdAt.getTime()`
- `token.ts:134` — `session.createdAt.getTime()`

themavik:
## Summary

Fixes #7819

**Root cause:** The OAuth provider plugin calls `.getTime()` directly on date values read from the database, assuming they are always `Date` objects. However, database adapters (e.g., Drizzle with `text()` or `integer()` column types) may return dates as strings or numbers.

## Changes

- `packages/oauth-provider/src/introspect.ts`: Wrapped `expiresAt` and `createdAt` with `new Date()` before calling `.getTime()` (4 call sites)
- `packages/oauth-provider/src/register.ts`: Same wrapping (2 call sites)

`new Date()` safely handles:
- `Date` objects (returns the same date)
- ISO strings (`"2024-01-01T00:00:00Z"`)
- Unix timestamps in ms (`1704067200000`)

## Risk Assessment

**Low** - `new Date(dateObj)` is a no-op when input is already a Date.

---
## Summary by cubic
Safely coerce DB date fields in the OAuth provider by using new Date before getTime to prevent TypeErrors with string or number values. Fixes #7819.

- **Bug Fixes**
  - OAuth introspect/register: wrap expiresAt/createdAt with new Date when computing exp/iat; Passkey: delete verification by value (data.id) to ensure cleanup.

- **New Features**
  - Email OTP: add rate limits for request-password-reset, reset-password, and forget-password endpoints; Docs: update DodoPayments with checkoutSession and metered usage, and SSO with a SAML warning and trustEmailVerified option.

<sup>Written for commit 98752d94d9c12c1512610d37e3bf24985321980a. Summary will update on new commits.</sup>


ping-maxwell (approved):

himself65 (approved):

himself65:
Please rebase the code to canary branch

PR #7901, opened 2026-02-10
fix(oauth-provider): defensive Date coercion for Drizzle/SQLite compatibility

### LINKED ISSUE #7819: OAuth Provider plugin assumes `Date` objects from DB without validation with Drizzle

jayy-77:
## Problem

The oauth-provider plugin calls `.getTime()` directly on date values retrieved from the database (`expiresAt`, `createdAt`). When using the Drizzle adapter with SQLite, these values may be returned as ISO strings or Unix timestamps rather than native `Date` objects, causing:

```
TypeError: expiresAt.getTime is not a function
```

This affects token introspection, client registration responses, and id token generation.

## Changes

- **`packages/oauth-provider/src/utils/index.ts`**: Added a `toDate()` utility that defensively coerces `string | number | Date` to a `Date` object. Only calls `new Date()` if the value is not already a `Date` instance.
- **`packages/oauth-provider/src/introspect.ts`**: Wrapped 4 `.getTime()` calls with `toDate()` — opaque access token `expiresAt`/`createdAt` and refresh token `expiresAt`/`createdAt`.
- **`packages/oauth-provider/src/register.ts`**: Wrapped 2 `.getTime()` calls with `toDate()` — client `expiresAt` and `createdAt` in `schemaToOAuth`.
- **`packages/oauth-provider/src/token.ts`**: Wrapped 1 `.getTime()` call with `toDate()` — `session.createdAt` used for id token `auth_time` claim.

## Test plan
- `npx tsc --noEmit` in `packages/oauth-provider` — clean
- `npx tsc --noEmit` in `packages/better-auth` — clean

Fixes #7819

---
## Summary by cubic
Defensively coerce DB date fields to Date objects in oauth-provider to avoid runtime errors with Drizzle/SQLite. This fixes bad exp/iat calculations in introspection, client registration responses, and id token auth_time. Fixes #7819.

- **Bug Fixes**
  - Added toDate(value) utility to normalize string|number|Date to Date.
  - Wrapped getTime() usages in introspect (access/refresh token exp/iat) and register (client expiresAt/createdAt).
  - Wrapped session.createdAt in token.ts when computing id token auth_time.

<sup>Written for commit 2fde5a6678809bb1d873ea5e72d518450af95998. Summary will update on new commits.</sup>

### DIFF
--- a/packages/oauth-provider/src/introspect.ts
+++ b/packages/oauth-provider/src/introspect.ts
@@ -18,6 +18,7 @@ import {
 	getJwtPlugin,
 	getStoredToken,
 	parseClientMetadata,
+	toDate,
 	validateClientCredentials,
 } from "./utils";
 
@@ -54,10 +55,10 @@ async function validateJwtAccessToken(
 			jwksFetch: jwtPluginOptions?.jwks?.remoteUrl
 				? jwtPluginOptions.jwks.remoteUrl
 				: async () => {
-						const jwksRes = await jwtPlugin?.endpoints.getJwks(ctx);
-						// @ts-expect-error response is a JSONWebKeySet but within the response field
-						return jwksRes?.response as JSONWebKeySet | undefined;
-					},
+					const jwksRes = await jwtPlugin?.endpoints.getJwks(ctx);
+					// @ts-expect-error response is a JSONWebKeySet but within the response field
+					return jwksRes?.response as JSONWebKeySet | undefined;
+				},
 			verifyOptions: {
 				audience: opts.validAudiences ?? ctx.context.baseURL,
 				issuer: jwtPluginOptions?.jwt?.issuer ?? ctx.context.baseURL,
@@ -216,11 +217,11 @@ async function validateOpaqueAccessToken(
 	// Add Custom Claims
 	const customClaims = opts.customAccessTokenClaims
 		? await opts.customAccessTokenClaims({
-				user,
-				scopes: accessToken.scopes,
-				referenceId: accessToken?.referenceId,
-				metadata: parseClientMetadata(client?.metadata),
-			})
+			user,
+			scopes: accessToken.scopes,
+			referenceId: accessToken?.referenceId,
+			metadata: parseClientMetadata(client?.metadata),
+		})
 		: {};
 
 	// Return the access token in introspection format
@@ -236,8 +237,8 @@ async function validateOpaqueAccessToken(
 		client_id: accessToken.clientId,
 		sub: user?.id,
 		sid: sessionId,
-		exp: Math.floor(accessToken.expiresAt.getTime() / 1000),
-		iat: Math.floor(accessToken.createdAt.getTime() / 1000),
+		exp: Math.floor(toDate(accessToken.expiresAt).getTime() / 1000),
+		iat: Math.floor(toDate(accessToken.createdAt).getTime() / 1000),
 		scope: accessToken.scopes?.join(" "),
 	} as JWTPayload;
 }
@@ -323,8 +324,8 @@ async function validateRefreshToken(
 		iss: jwtPluginOptions?.jwt?.issuer ?? ctx.context.baseURL,
 		sub: user?.id,
 		sid: sessionId,
-		exp: Math.floor(refreshToken.expiresAt.getTime() / 1000),
-		iat: Math.floor(refreshToken.createdAt.getTime() / 1000),
+		exp: Math.floor(toDate(refreshToken.expiresAt).getTime() / 1000),
+		iat: Math.floor(toDate(refreshToken.createdAt).getTime() / 1000),
 		scope: refreshToken.scopes?.join(" "),
 	} as JWTPayload;
 }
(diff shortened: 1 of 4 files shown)


hammer-ai:
In case it's useful, I'm also running into this issue. But I get it when I add `secondaryStorage`:
```
 secondaryStorage: {
    get: async (key) => await kv.get(`better-auth:${key}`),
      if (ttl) {
        await kv.set(`better-auth:${key}`, value, {ex: ttl})
      } else {
        await kv.set(`better-auth:${key}`, value)
      }
    },
    delete: async (key) => {
      await kv.del(`better-auth:${key}`)
    }
  },
```
I then get:
```
[update-user-profile@POST] TypeError: expiresAt.getTime is not a function
    at async POST (src/app/api/user/update-user-profile/route.ts:63:5)
  61 |     // Use auth.api.updateUser to update user and session together.
  62 |     // This ensures the session reflects the updated values immediately.
> 63 |     await auth.api.updateUser({
     |     ^
  64 |       body: {
  65 |         ...(body.username !== undefined && {username: body.username}),
  66 |         ...(body.bio !== undefined && {bio: body.bio}),
 POST /api/user/update-user-profile 400 in 2688ms
```

ping-maxwell:
Hello, I've opened a PR to solve this issue. Thank you for the effort though!

PR #7943, opened 2026-02-12
feat: add `countOn` option for rate limiter

himself65:
## Summary by cubic
Add a countOn option to the rate limiter to control when requests are counted (all vs error). This lets you count only failed responses to avoid throttling successful flows.

- **New Features**
  - countOn in rate limit config: "all" (default) or "error" (only status >= 400 are counted).
  - Per-path customRules can set countOn.
  - Counting now occurs after response and respects countOn; tests cover both modes.

- **Migration**
  - customRules no longer accept functions; use static rule objects or false per path.

<sup>Written for commit ce704b626b76b446d5da7254da917fe036a56641. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/index.ts
+++ b/packages/better-auth/src/api/index.ts
@@ -303,7 +303,7 @@ export const router = <Option extends BetterAuthOptions>(
 			return currentRequest;
 		},
 		async onResponse(res, req) {
-			await onResponseRateLimit(req, ctx);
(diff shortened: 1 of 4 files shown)


gustavovalverde:
Closing in favor of #9309. #9309 preserves the function-form `customRules`, wraps `onResponse` in `try/finally` so accounting still runs when a plugin throws, and counts against the final plugin-modified response; this PR drops the function form and counts pre-plugin.

PR #7965, opened 2026-02-13
fix(stripe): fall back to `customers.list` when `customers.search` API is unavailable

### LINKED ISSUE #7959: Bug: `@better-auth/stripe` — `customers.search` unavailable in India (and other regions)
## To Reproduce

1. Set up a Stripe account registered in India
2. Configure `@better-auth/stripe` plugin with subscription management
3. Call the `/api/auth/subscription/upgrade` endpoint (or trigger `createCustomerOnSignUp`)
4. The Stripe API call fails with: `search feature unavailable for merchant`

The Stripe Search API (`/v1/customers/search`) is **not available** for merchant accounts in certain regions, including India. See: https://docs.stripe.com/search#supported-countries

The `@better-auth/stripe` plugin uses `customers.search` in 3 places within `dist/index.mjs`:

| Location | Purpose |
|----------|---------|
| `upgradeSubscription` (~line 563) | Find org customer by `metadata["organizationId"]` |
| `upgradeSubscription` (~line 603) | Find user customer by email, excluding org customers |
| `stripe()` database hook (~line 1427) | Find existing customer on signup by email |

All 3 throw `UNABLE_TO_CREATE_CUSTOMER` because the `customers.search` call fails.

**Current:** The plugin unconditionally uses `stripe.customers.search()`, which fails with a Stripe API error for merchants in unsupported regions (India, etc.). This makes the entire subscription flow unusable.

**Expected:** The plugin should fall back to `stripe.customers.list()` with appropriate filtering when `customers.search` is unavailable, or use `customers.list` by default since it works globally. The `customers.list` API supports `email` filtering natively, and metadata filtering can be done client-side on a small result set.

### Suggested fix

Replace `customers.search` with `customers.list`:

**For email-based searches:**
```js
// Before
client.customers.search({
  query: `email:"${email}" AND -metadata["customerType"]:"organization"`,
  limit: 1
}).data[0]

// After
client.customers.list({ email, limit: 100 })
  .data.find(c => c.metadata?.customerType !== "organization")
```

**For metadata-based searches (org):**
```js
// Before
client.customers.search({
  query: `metadata["organizationId"]:"${orgId}"`,
  limit: 1
}).data[0]

// After
client.customers.list({ limit: 100 })
  .data.find(c => c.metadata?.organizationId === orgId)
```

This is a drop-in replacement that works in all Stripe regions.

1.4.18 (also affects `@better-auth/stripe` v1.4.18)

## System info

```
better-auth: 1.4.18
@better-auth/stripe: 1.4.18
Runtime: Node.js 22
```

- Package (`@better-auth/stripe`)

## Auth config (if applicable)

```typescript
import { stripe } from "@better-auth/stripe";

  plugins: [
    stripe({
      stripeSecretKey: process.env.STRIPE_SECRET_KEY,
      stripeWebhookSecret: process.env.STRIPE_WEBHOOK_SECRET,
      createCustomerOnSignUp: true,
      subscription: {
        enabled: true,
        plans: [/* ... */],
      },
    }),
  ],
});
```

## Additional context

- **Workaround:** Patching `dist/index.mjs` locally to replace all 3 `customers.search` calls with `customers.list` equivalents works correctly.
- The Stripe Search API documentation explicitly lists supported countries: https://docs.stripe.com/search#supported-countries
- India is a major market and this blocks all Indian Stripe merchants from using Better Auth's subscription features.
- The `customers.list` API is available globally and supports `email` filtering natively. Metadata filtering requires client-side filtering on the result set, but since these queries use `limit: 1`, the data volume is minimal.

bytaesu:
> [!NOTE]
> https://github.com/stripe/stripe-node#auto-pagination

Since emails aren't directly tied to the organization, there's no server side filter we can use with the list method. Using search is the optimal approach, but Stripe mentions that this API may not be available in some countries.

In those cases, even if there are performance tradeoffs, we'll fall back to list method.

- Closes #7959

---
## Summary by cubic
Add a safe fallback to Stripe customers.list when customers.search is unavailable in some regions, preventing duplicate customers for users and organizations.

- **Bug Fixes**
  - Try customers.search first; on failure, log a warning and scan customers.list (limit 100) to find existing records.
  - For users, match by email and skip organization customers via metadata (customerType !== "organization").
  - For organizations, match by metadata organizationId.
  - Applied in user signup/customer link and upgradeSubscription flows.
  - Tests cover signup and upgrade fallbacks to ensure no duplicate creation.

<sup>Written for commit 023156093929a0755a002a2baf917392c68d6c24. Summary will update on new commits.</sup>

### DIFF
--- a/packages/stripe/test/stripe.test.ts
+++ b/packages/stripe/test/stripe.test.ts
@@ -4133,20 +4133,6 @@ describe("stripe", () => {
 			const fallbackEmail = "fallback-user@example.com";
 			const existingCustomerId = "cus_fallback_123";
 
-			// Simulate Search API failure (unavailable in some regions)
-			mockStripe.customers.search.mockRejectedValueOnce(
-				new Error("search feature unavailable for merchant"),
-			);
-			mockStripe.customers.list.mockReturnValueOnce(
-				mockStripeList([
-					{
-						id: existingCustomerId,
-						email: fallbackEmail,
-						metadata: { customerType: "user" },
-					},
-				]),
-			);
-
 			const testOptions = {
 				...stripeOptions,
 				createCustomerOnSignUp: true,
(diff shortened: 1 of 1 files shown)


bytaesu:
@claude, look into this limitation and determine whether it's unavoidable, or if there's a way to address it without introducing a breaking change.

packages/stripe/src/index.ts
```diff
@@ -304,12 +304,32 @@ export const stripe = <O extends StripeOptions>(options: O) => {
...
+											const result = await client.customers.search({
+												query: `email:"${escapeStripeSearchValue(user.email)}" AND -metadata["${customerMetadata.keys.customerType}"]:"organization"`,
+												limit: 1,
+											});
+											stripeCustomer = result.data[0];
+										} catch {
+											// Search API unavailable in some regions, so fall back to paginated list
+											ctx.context.logger.warn(
+												"Stripe customers.search failed, falling back to customers.list",
+											);
```
Bekacru, line 318: we should only log is on dev

Issue #4151, opened 2025-08-22
Dynamic setting of `baseURL` based on current request


Stadly:
> An app could be running on ten different URLs at the same time. e.g. an app deployed on Vercel might have a couple of custom domain names, deployment id sub domain, and environment sub domain. It's not practical to change the environment variable on every deployment or even change the value of `baseUrl` parameter since `betterAuth()` instance is created once and reused; and the URL might not be known at startup time.
> I would suggest that `baseUrl ` parameter be changed to accept a function that takes a `Request` as parameter and returns a URL, similar to `trustedOrigins` parameter. This will allow developers to derive `baseUrl` from the received request (taking `x-forwarded-host`, etc into account). 

 _Originally posted by @RaeesBhatti in [#1119](https://github.com/better-auth/better-auth/issues/1119#issuecomment-2928697529)_

himself65:
This is a good feature request. Right now, for our website, we use this way to set the baseURL and cookie domain

https://github.com/better-auth/better-auth/blob/d476fbed074253121df0e23414bd059c93274553/demo/nextjs/lib/auth.ts#L52-L68

I'm still thinking the best way to improve this.

gregoiregentil:
I have a similar request for cookie domain. I run multiple domains (app.example1.com + auth.example1.com and app.example2.com + auth.example2.com), plus the better-auth backend is proxied. I need to set the cookie domain to example1.com and example2.com

The code [here](https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/cookies/index.ts#L28) shows that the cookie domain can be derived from baseurl if not set up.

Alternatively, a regex cookie domain would be an idea (basically to strip any subdomain). But best would be a hook for cookie domain option.

Stadly:
I ended up creating an auth instance for each origin on demand like this:

```ts
// auth.server.ts
const authConfig = {
  // ...
} satisfies BetterAuthOptions;

// Better Auth requires baseURL to be configured. If it is not set, the origin
// of the first request will be used, both for that request and for all
// subsequent requests. To support multiple origins, we create a Better Auth
// instance per origin. In the future, Better Auth may support dynamic baseURL.
// Ref: https://github.com/better-auth/better-auth/issues/4151
const authInstances = new Map<string, Auth>();
export const getAuth = (baseURL = "") => {
  if (!authInstances.has(baseURL))
    authInstances.set(baseURL, betterAuth({ ...authConfig, baseURL }));
[...]

PR #8009, opened 2026-02-16
feat(auth): support dynamic baseURL with allowedHosts

Paola3stefania:
Addresses https://github.com/better-auth/better-auth/issues/4151

---
## Summary by cubic
Adds dynamic baseURL resolution with an allowlist to support multi-domain and preview deployments. Resolves the base URL per request, isolates request context, updates JWT/cookies/trusted origins, and adds docs. Addresses better-auth/better-auth#4151.

- **New Features**
  - baseURL now accepts { allowedHosts, fallback?, protocol? } with wildcard support (e.g., *.vercel.app, preview-*.myapp.com, localhost:*); validated to be non-empty.
  - Per-request host/protocol resolution via x-forwarded-host/proto with strict validation; unknown hosts are rejected unless a fallback is set. trustedOrigins auto-expands with all allowedHosts and the fallback (adds http for localhost).
  - Request-scoped context prevents cross-host races; JWT issuer/audience and MCP/OAuth Proxy/OIDC/Passkey/Telemetry now use the resolved origin.
  - Cookies: with crossSubDomainCookies enabled, the cookie domain is derived per request; Secure flag precedence is advanced.useSecureCookies > dynamic protocol (https/http) > static baseURL scheme > NODE_ENV for auto/unset.

- **Bug Fixes**
  - Corrected origin and basePath handling across URL builders (OAuth Proxy error URL, MCP, OIDC, SvelteKit), ensuring consistent behavior for static and dynamic configs.
  - Minor typing and lint cleanups.

<sup>Written for commit 5d1848dd50c1cccd3319618ba5479c79b659ff16. Summary will update on new commits.</sup>


packages/core/src/types/init-options.ts
```diff
@@ -46,6 +46,53 @@ export type GenerateIdFn = (options: {
...
+
+	/**
+	 * Protocol to use when constructing the URL.
+	 * - `"https"`: Always use HTTPS (recommended for production)
+	 * - `"http"`: Always use HTTP (for local development)
+	 * - `"auto"`: Derive from `x-forwarded-proto` header or default to HTTPS
+	 *
+	 * @default "auto"
+	 */
+	protocol?: "http" | "https" | "auto" | undefined;
```
himself65, line 87: Should we just let `string` here? since maybe it's not http(s) protocol, like expo

Paola3stefania: Hummm ... the protocol field here is for the servers base URL (https://myapp.vercel.app) which is always http or https. For Expo, exp:// is handled through trustedOrigins + the Expo plugin... not the base URL

Is there a use case where the server itself would need a non-http protocol?

PR #8083, opened 2026-02-20
fix(email-otp): allow checkVerificationOTP for non-existent users during signup

GoPro16:
`checkVerificationOTP` throws `USER_NOT_FOUND` unconditionally when the user doesn't exist in the database. This breaks signup flows where OTP is verified before account creation (e.g. using `sendVerificationOtp` with `type: "sign-in"` for a new email, then calling `checkVerificationOtp` to validate before proceeding).

Applied the same guard used in `sendVerificationOTP`: allow non-existent users when `type === "sign-in"` and `disableSignUp` is false.

Added regression test for the signup flow scenario. Full email-otp test suite passes (47/47).


himself65:

kvanbere:
Hey, this is already working for me without this PR fix.

You can use `signInEmailOTP` with allow user creation set to true, then you use `updateUser` to set the user name, and `setPassword` to add a password to the account.

GoPro16:
> Hey, this is already working for me without this PR fix.
> 
> You can use `signInEmailOTP` with allow user creation set to true, then you use `updateUser` to set the user name, and `setPassword` to add a password to the account.

@kvanbere 

The flow you describe (`signInEmailOTP` → `updateUser` → `setPassword`) works because `signInEmailOTP` handles user creation internally. This PR fixes a different use case: calling `checkVerificationOtp`
  standalone to validate the OTP *before* user creation (e.g. to gate a signup form). `sendVerificationOtp` already allows non-existent users — this just aligns `checkVerificationOtp` with the same guard.

Issue #8080, opened 2026-02-20
Add `authClient.updateSession()` method


ryami333:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

It's currently possible to specify `additionalFields` on a `session` or `user`, like this:

```ts
export const auth = betterAuth({
  session: {
    selectedOrganisation: {
      type: "string",
      required: false,
    }
  },
  user: {
    favoriteColor: {
      type: "string",
      required: false,
    }
  },
};
```

And, there is a mechanism for _updating_ that value for a `user`, like:

```ts
import { createAuthClient } from "better-auth/react";

const authClient = createAuthClient();

authClient.updateUser({ favoriteColor: "red" });
```

However, it's *not* possible to do the same for a `session`:

```ts
// No such method!
authClient.updateSession({ selectedOrganisation: "foo" });
```

### Describe the solution you'd like

Ideally, I'd like to see the `authClient.updateSession(…)` method added.

### Describe alternatives you've considered

I'm aware that this is potentially dangerous – if you can update a session then you can potentially _invalidate_ it. I would like to suggest two potential mitigations:

1. The method could be called `authClient.dangerousUpdateSession` to reflect this.
2. The method could be called `authClient.updateSessionAdditionalFields`, and its implementation could guard the core session fields against mutation.

### Additional context

Related discussions:

- https://github.com/better-auth/better-auth/discussions/3860
- https://github.com/better-auth/better-auth/discussions/6148

PR #8084, opened 2026-02-20
feat: add `/update-session`

himself65:
Fixes: https://github.com/better-auth/better-auth/issues/8080

---
## Summary by cubic
Add POST /update-session to update custom session fields on the current session and refresh the session cookie. Fixes client update signal so session state refetches after updates; fixes better-auth/better-auth#8080.

- **New Features**
  - Adds updateSession endpoint and client.updateSession; only accepts additional fields and rejects core or input:false fields.
  - Refreshes the session cookie and returns the normalized session; passes action "update" to parseSessionInput, adds AdditionalSessionFieldsInput, and updates OpenAPI/docs/tests.

<sup>Written for commit 24d746d5b6acc3f727d5a318fc893a7792efef9a. Summary will update on new commits.</sup>

### DIFF
--- a/docs/components/ui/sparkles.tsx
+++ b/docs/components/ui/sparkles.tsx
@@ -41,8 +41,7 @@ export const SparklesCore = (props: ParticlesProps) => {
 	const particlesLoaded = async (container?: Container) => {
 		if (container) {
 			console.log(container);
-			// biome-ignore lint/nursery/noFloatingPromises: add error handling is not important
-			controls.start({
+			await controls.start({
 				opacity: 1,
 				transition: {
 					duration: 1,
--- a/docs/content/docs/authentication/email-password.mdx
+++ b/docs/content/docs/authentication/email-password.mdx
@@ -168,12 +168,35 @@ If you enable require email verification, users must verify their email before t
 <Callout>
   This only works if you have sendVerificationEmail implemented and if the user
   is trying to sign in with email and password.
+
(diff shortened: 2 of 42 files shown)


packages/better-auth/src/api/routes/update-session.ts
```diff
@@ -0,0 +1,119 @@
...
+				throw APIError.from(
+					"BAD_REQUEST",
+					BASE_ERROR_CODES.BODY_MUST_BE_AN_OBJECT,
+				);
+			}
+
+			const coreFieldsInBody = Object.keys(body).filter((key) =>
+				CORE_SESSION_FIELDS.has(key),
+			);
+			if (coreFieldsInBody.length > 0) {
```
ping-maxwell: We should use `getAuthSchema` and check if a field has `input: false` or not, this will help with plugin-added fields that should be protected too.

ping-maxwell (changes requested):

Bekacru (approved):

packages/better-auth/src/api/routes/update-session.ts
himself65: We use `parseSessionInput`, I think that's okay?

ryami333:
@himself65 are you aware that `updateSession` still doesn't infer the field types as defined in your `auth.ts`. As a result, you get type errors when you do `authClient.updateSession({ selectedOrganisation: "foo" });`.

PR #8101, opened 2026-02-22
Fix Session ID Missing From Secondary Storage

barreeeiroo:
## Summary

Sessions stored in secondary storage (e.g. Redis) are missing the `id` field, causing any code that reads `session.id` from a cached session to get `undefined`.

This breaks the `@better-auth/oauth-provider` plugin: the authorization code is stored without a `sessionId`, and the token exchange fails with `{"error_description":"session no longer exists","error":"invalid_request"}`.

This bug occurs whenever secondary storage is configured, even when using `storeSessionInDatabase: true`. As long as there is secondary storage, the Redis entry will be missing the `id`.

## Root Cause

In `createSession` (`packages/better-auth/src/db/internal-adapter.ts`), the `data` object built for the new session never includes an `id` — that field is normally generated by the DB adapter's `transformInput` during the `create` call.

In `createWithHooks` (`packages/better-auth/src/db/with-hooks.ts`, lines 48–58), the custom `fn` that writes to secondary storage receives `sessionData` which lacks the `id`:

- When `storeSessionInDatabase` is **false** (default): the DB adapter is never called. `fn` receives `actualData` directly — no `id` is ever generated.
- When `storeSessionInDatabase` is **true**: the DB adapter runs first and returns a record with `id`, but `fn` still stores a session without `id` in secondary storage.

The result: the DB has the full session; secondary storage has a partial one.

Redis entry (note no `id` in `session`):
```json
{
  "session": {
    "ipAddress": "...",
    "userAgent": "...",
    "expiresAt": "...",
    "userId": "usr_xxx",
    "token": "abc123",
    "createdAt": "...",
    "updatedAt": "..."
  },
  "user": { "id": "usr_xxx", "..." }
}
```

## Fix

After `createWithHooks` returns, patch the secondary storage entry to include the `id`:

- If the DB result already contains the `id` (i.e. `storeSessionInDatabase: true`), use that.
- Otherwise (secondary-only mode), fall back to `ctx.generateId({ model: "session" })`.

## Workaround

Until this is released, users can patch the Redis entry via a `databaseHooks.session.create.after` hook:

```typescript
session: {
  create: {
    after: async (session) => {
      if (redisClient && session.id && session.token) {
        try {
          const cached = await redisClient.get(session.token);
          if (cached) {
            const parsed = JSON.parse(cached);
            if (parsed?.session && !parsed.session.id) {
              parsed.session.id = session.id;
              const ttl = await redisClient.ttl(session.token);
              await redisClient.set(
                session.token,
                JSON.stringify(parsed),
                ttl > 0 ? { EX: ttl } : undefined,
              );
            }
          }
        } catch (error) {
          console.error("Failed to patch session id in Redis:", error);
        }
      }
    },
  },
},
```

---
## Summary by cubic
Fixes missing session.id in secondary storage by updating the cached session after creation. Restores OAuth token exchange by ensuring authorization codes include sessionId.

- **Bug Fixes**
  - After createWithHooks, inject session.id into secondary storage; use DB id when available, otherwise generate one.
  - Added tests for secondary-only and database+secondary modes to verify id is stored.
  - Prevents OAuth token exchange failure ("session no longer exists").

<sup>Written for commit 4118debd9fd9d60d860b1bed545818e19cbdfcad. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/db/internal-adapter.ts
+++ b/packages/better-auth/src/db/internal-adapter.ts
@@ -387,6 +387,29 @@ export const createInternalAdapter = (
 						}
 					: undefined,
 			);
+			if (secondaryStorage && res) {
+				const id = res.id || ctx.generateId({ model: "session" });
+				if (id && id !== res.id) {
+					res.id = id;
+				}
+				if (id) {
+					const cached = await secondaryStorage.get(data.token);
+					if (cached) {
+						const parsed = safeJSONParse<{
+							session: Record<string, any>;
+							user: User;
+						}>(cached);
+						if (parsed?.session && !parsed.session.id) {
+							parsed.session.id = id;
+							await secondaryStorage.set(
+								data.token,
+								JSON.stringify(parsed),
+								getTTLSeconds(data.expiresAt),
+							);
+						}
+					}
+				}
+			}
 			return res as Session;
 		},
 		findSession: async (
--- a/packages/better-auth/src/db/secondary-storage.test.ts
+++ b/packages/better-auth/src/db/secondary-storage.test.ts
@@ -36,6 +36,7 @@ describe("secondary storage - get returns JSON string", async () => {
 		});
 		expect(s1.data).toMatchObject({
 			session: {
+				id: expect.any(String),
 				userId: expect.any(String),
 				token: expect.any(String),
 				expiresAt: expect.any(Date),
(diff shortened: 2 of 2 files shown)


himself65 (commented):
If I remembered correctly, id not in session is expected behavior. Let me double check

barreeeiroo:
@himself65 cool, let me know once you sort it out.

Even if it's not expected, I'd challenge that. The reason is that, as of now, the OAuth 2.1 Provider plugin is broken if Secondary Storage is used. I did enable `storeSessionInDatabase: true`, but I was facing the same issue of the Session ID not being present in the `verification` table. I had to add the workaround I mention in the Description to be able to use the Provider along with secondary storage. It throws through this branch, as the Session ID is not in verification data: https://github.com/better-auth/better-auth/blob/f3f8261ba6fd015b834fd66fc544911405c3ba36/packages/oauth-provider/src/token.ts#L739

himself65:
Thanks for the detailed investigation and PR! However, I believe the current behavior where session.id is undefined in secondary storage (when storeSessionInDatabase is not true) is
  intentional.                                                                                                                                                                          
                                                                                                                                                                                      
  Some context: PR #7476 originally added session ID pre-generation for secondary storage, but this was deliberately reverted in #7583, which removed the duplicate ID generation logic
  and added explicit e2e assertions that session.id === undefined in secondary storage (see e2e/smoke/test/redis.spec.ts lines 104 and 232). The only case where session.id is expected
  to exist in secondary storage is when storeSessionInDatabase: true, because the DB adapter generates it.

  This PR would also break those existing e2e tests.

  If the @better-auth/oauth-provider plugin needs a session ID from secondary storage, I think the fix should live in the oauth-provider plugin itself rather than changing the
  secondary storage contract. Could you open an issue on the oauth-provider side instead?

  ---
  Feel free to adjust the tone or add/remove details as you see fit.

barreeeiroo:
Okay, makes sense. I did consider fixing the OAuth 2.1 Provider instead of the logic in Secondary Storage, however it made sense to instead fix Secondary Storage. This way, the object stored matches exactly with the session schema like in Primary Storage.  
Also, if the provider itself is broken, maybe some other providers or plugins are broken too...

I'll take a look later today to the OAuth 2.1 Provider code, and try to get a PR out. Let me know if you have any other concerns, otherwise will close this one.

himself65:

bytaesu:
Hi @barreeeiroo, thanks for working on this!

I think there was a inconsistent behavior and regression between using the database and secondary storage. For example, adding Redis after using the DB could break existing systems. I'll address this in PR #8927.

barreeeiroo:
Hi @bytaesu, I'm not sure if that PR addresses the actual issue I'm bringing up here...

From the comment in the code, I see this:

```
		// Session should always have an id, even with secondary storage only

			// When secondary storage is the only store, the database adapter
			// won't run, so we need to generate an id ourselves.
```

However, my case was a hybrid mode... I was having both database and session, and the OAuth 2.1 Provider is broken because of that. Are you sure that PR will address that particular scenario, or only when only using secondary storage?

PR #8110, opened 2026-02-23
fix(core): avoid user modelName collision with account model

dilame:
## Summary

Fixes a schema/model-name collision when configuring:

- `user.modelName = "account"`
- `account.modelName = "identity"`

Previously, references for `session.userId` and `account.userId` were built from `options.user.modelName`, which caused collisions with the built-in `account` model key and produced incorrect references.

This patch makes those references point to the base user schema key (`"user"`), preserving the logical relation to the user model regardless of `modelName` aliases.

## What changed

- `packages/core/src/db/get-tables.ts`
  - use the base user model key for references in:
    - `session.fields.userId.references.model`
    - `account.fields.userId.references.model`

- `packages/core/src/db/test/get-tables.test.ts`
  - add regression test for:
    - `user.modelName = "account"`
    - `account.modelName = "identity"`
  - assert:
    - custom model names are preserved
    - `session.userId.references.model === "user"`
    - `account.userId.references.model === "user"`

## Reproduction (before fix)

Configure Better Auth with:
- `user.modelName = "account"`
- `account.modelName = "identity"`

Then generate schema: references may target the wrong table due to model-name collision.

## Expected

References that represent "user ownership" should always resolve against the user schema key, not be redirected by `modelName` collisions.

## Validation

Ran:
```bash
pnpm --filter @better-auth/core exec vitest run src/db/test/get-tables.test.ts

---
## Summary by cubic
Fixes model-name collisions when user.modelName overlaps with account. User references now always point to the base "user" schema, and default model key resolution prefers modelName over schema key.

- **Bug Fixes**
  - Always use the base "user" model key for session.userId and account.userId references.
  - Prefer modelName over schema key when resolving default model keys (avoids "account"/"identity" collisions).
  - Preserve custom modelName aliases for user and account.

- **Refactors**
  - Use USER_MODEL_KEY constant as the fallback for user.modelName.

<sup>Written for commit 4ffb14d313b4e5b89faff860b42451122c69b32b. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/src/db/adapter/get-default-model-name.ts
+++ b/packages/core/src/db/adapter/get-default-model-name.ts
@@ -24,11 +24,11 @@ export const initGetDefaultModelName = ({
 		// Thus we'll try the search but without the trailing `s`.
 		if (usePlural && model.charAt(model.length - 1) === "s") {
 			const pluralessModel = model.slice(0, -1);
-			let m = schema[pluralessModel] ? pluralessModel : undefined;
+			let m = Object.entries(schema).find(
+				([_, f]) => f.modelName === pluralessModel,
+			)?.[0];
 			if (!m) {
-				m = Object.entries(schema).find(
-					([_, f]) => f.modelName === pluralessModel,
-				)?.[0];
+				m = schema[pluralessModel] ? pluralessModel : undefined;
 			}
 
 			if (m) {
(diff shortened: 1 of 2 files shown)


dilame:
Closing in favor of a refreshed PR with additional join/model resolution fixes and regression tests.

PR #8113, opened 2026-02-23
fix(core): deterministic model/field resolution for user/account collisions

dilame:
## Summary
- make adapter model resolution deterministic when schema keys collide with custom  values
- prefer exact schema key before alias resolution in 
- resolve fields using explicit model key first, then mapped model alias when required (without global schema guessing)
- update adapter join/where internals to use exact schema keys when already provided
- add regression test for  colliding with OAuth 

## Why
When  is set to  and OAuth  is also present (often remapped to ), adapters can resolve model/field lookups against the wrong table, causing runtime errors like  and invalid joins.

## Validation
- 
 RUN  v4.0.18 /Users/bowzee/pinqwin/better-auth/packages/core

 ✓ src/db/test/model-resolution-collision.test.ts (3 tests) 2ms

 Test Files  1 passed (1)
      Tests  3 passed (3)
   Start at  00:15:36
   Duration  206ms (transform 86ms, setup 0ms, import 105ms, tests 2ms, environment 1ms) ✅
- full  test run currently fails on existing unrelated import issue in  (), reproduced locally without touching that area.

---
## Summary by cubic
Make model and field resolution deterministic when schema keys collide with custom modelName aliases (e.g., user → account alongside OAuth account). Prevents wrong table lookups, broken joins, and runtime errors.

- **Bug Fixes**
  - Prefer exact schema key before alias mapping or plural handling.
  - Resolve fields on the explicit model if it has the field; otherwise fall back to alias, then default.
  - Fix join direction by matching FK references against both model keys and modelName aliases; use exact keys in join/where and FK checks.
  - Add regression test for user/account collisions and OAuth account fields.

<sup>Written for commit d3918e023c6b8046e1b2bf804afb9ceb9da194a9. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/src/db/adapter/factory.ts
+++ b/packages/core/src/db/adapter/factory.ts
@@ -152,6 +152,12 @@ export const createAdapterFactory =
 		});
 		const resolveSchemaModelKey = (model: string) =>
 			schema[model] ? model : getDefaultModelName(model);
+		const referencesModelMatches = (
+			referenceModel: string,
+			targetModelKey: string,
+		) =>
+			referenceModel === targetModelKey ||
+			schema[targetModelKey]?.modelName === referenceModel;
 
 		const getModelName = initGetModelName({
 			usePlural: config.usePlural,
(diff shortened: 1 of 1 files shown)


dilame:
Superseded by a new PR to force a fresh pkg.pr.new build from latest commit.

PR #8117, opened 2026-02-23
fix(drizzle-adapter): resolve db.query key mismatch when config.schema keys differ from schema exports

typedrat:
When users pass singular model names in config.schema (e.g. { user: schema.users }) but drizzle() receives plural schema exports, db.query is keyed by the export names ("users"), not the config.schema keys ("user"). The join code path did db.query[model] directly, which errored with "model not found in the query object" when the keys didn't match.

Add getQueryModel() helper that scans config.schema to find the matching db.query key via table object identity when direct lookup fails.

---
## Summary by cubic
Fixes Drizzle adapter join queries when config.schema uses singular model names but the Drizzle schema exports are plural. Joins now resolve the correct db.query key and return related data as expected.

- **Bug Fixes**
  - Added getQueryModel() to resolve the db.query key by trying the model name, an optional plural (when usePlural is true), and finally matching the table object from config.schema against db._.fullSchema.
  - Updated findOne/findMany join paths to use the resolved key; logs once and falls back only if no match is found.
  - Added an e2e regression test covering singular config keys with plural schema exports to ensure joins return related records.

<sup>Written for commit 5440036c98e66d8ed6fc501653a4429b252c73a6. Summary will update on new commits.</sup>

### DIFF
--- a/e2e/adapter/test/drizzle-adapter/adapter.drizzle.plural-joins.test.ts
+++ b/e2e/adapter/test/drizzle-adapter/adapter.drizzle.plural-joins.test.ts
@@ -1,0 +1,259 @@
+/**
+ * Regression test: db.query[model] lookup fails when the Drizzle schema
+ * object passed to drizzle() has plural export names ("users", "sessions")
+ * but config.schema maps singular Better Auth model names to the tables.
+ *
+ * This is the standard setup when users have a hand-written Drizzle schema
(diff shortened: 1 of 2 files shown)


himself65 (approved):

himself65:
Please sign CLA and I can merge it

typedrat:
Okay, CLA signed.

PR #8124, opened 2026-02-24
fix(stripe): handle empty subscription list in onSubscriptionUpdated webhook

JoDrm:
## Bug

`onSubscriptionUpdated` in the Stripe plugin crashes with `TypeError: Cannot read properties of undefined (reading 'id')` when a `customer.subscription.updated` webhook event is received for a customer that has no subscription records in the database.

## Root Cause

In `packages/stripe/src/hooks.ts`, when the subscription lookup by `stripeSubscriptionId` fails, the code falls back to querying by `stripeCustomerId`. If the result is an empty array (`subs.length === 0`), the `else` branch assigns `subscription = subs[0]!` which is `undefined`. The subsequent `subscription.id` access then crashes.

```typescript
// Before (line ~314)
} else {
    subscription = subs[0]!; // undefined when subs is empty
}

// Later (line ~335)
where: [{ field: "id", value: subscription.id }] // TypeError!
```

## Fix

Add an early return when `subs` is empty, consistent with the existing guard for `subs.length > 1` with no active subscription.

```typescript
if (subs.length === 0) {
    ctx.context.logger.warn(
        `Stripe webhook warning: No subscription found for customerId: ${customerId}, ignoring event`,
    );
    return;
}
```

## How to reproduce

1. Have a Stripe customer with no subscription records in the database (e.g. a customer who only made one-time payments)
2. Trigger a `customer.subscription.updated` webhook event for that customer
3. The webhook handler crashes with `TypeError: Cannot read properties of undefined (reading 'id')`

---
## Summary by cubic
Prevents a crash in the Stripe onSubscriptionUpdated webhook when a customer has no subscriptions in the database. The handler now safely ignores the event and logs a warning instead of throwing a TypeError.

- **Bug Fixes**
  - Return early when no subscriptions are found for a customer, avoiding access to an undefined subscription.
  - Add a clear warning log and ignore the event, matching the existing behavior used for multiple non-active matches.
  - Applies to customer.subscription.updated events in the Stripe plugin.

<sup>Written for commit e8204361a84cbae9077af27d5176f296375d89b0. Summary will update on new commits.</sup>

### DIFF
--- a/packages/stripe/src/hooks.ts
+++ b/packages/stripe/src/hooks.ts
@@ -300,6 +300,12 @@ export async function onSubscriptionUpdated(
 				model: "subscription",
 				where: [{ field: "stripeCustomerId", value: customerId }],
 			});
+			if (subs.length === 0) {
+				ctx.context.logger.warn(
+					`Stripe webhook warning: No subscription found for customerId: ${customerId}, ignoring event`,
+				);
+				return;
+			}
 			if (subs.length > 1) {
 				const activeSub = subs.find((sub: Subscription) =>
 					isActiveOrTrialing(sub),


himself65:
Please rebase the code correctly

PR #8133, opened 2026-02-24
fix: prevent double encoded cookie

### LINKED ISSUE #8127: customSession plugin double-encodes Set-Cookie values during session refresh
> **Note:** This issue was generated with AI assistance (Claude Code).

## Description

The `customSession` plugin's `/get-session` endpoint override double-encodes cookie values when forwarding `Set-Cookie` headers from the internal `getSession()` call. The browser receives a double-encoded cookie after a session refresh, causing **"session not found"** errors on subsequent requests.

## Reproduction

1. Configure `betterAuth` with a `customSession` plugin (e.g., to enrich sessions with permissions).
2. Sign in via `/api/auth/sign-in/email`.
3. Observe the correct `Set-Cookie` header after sign-in:

```http
set-cookie: session_token_v1=BCwy6jS9Fm3FfpOoKlPlzrHxHvojZPtE.rNpUR%2B9TXuJJHbqi2lkVwTjVWi%2FTVU3I1nLlWr5t4MU%3D; Max-Age=604800; Path=/; HttpOnly; SameSite=Lax
```

4. Call `/api/auth/get-session` (this triggers a session refresh).
5. Observe the **double-encoded** `Set-Cookie` header returned by the `/get-session` override:

```http
set-cookie: session_token_v1=BCwy6jS9Fm3FfpOoKlPlzrHxHvojZPtE.rNpUR%252B9TXuJJHbqi2lkVwTjVWi%252FTVU3I1nLlWr5t4MU%253D; Max-Age=3599; Path=/; HttpOnly; SameSite=Lax
```

6. Subsequent requests fail with **"session not found"** because the browser sends the double-encoded cookie (e.g., `%252B`) which does not match the token stored in the database.

> Note: `%2B` is the correct encoding of `+`; `%252B` is `%2B` encoded again (`%` → `%25`).

---

## Root cause

In `src/plugins/custom-session/index.ts` the `/get-session` endpoint override reads `session.headers.getSetCookie()` and re-registers each cookie via `ctx.setCookie(...)`:

```ts
for (const cookieStr of session.headers.getSetCookie())
  parseSetCookieHeader(cookieStr).forEach((attrs, name) => {
    ctx.setCookie(name, attrs.value, {
      maxAge: attrs["max-age"],
      expires: attrs.expires,
      domain: attrs.domain,
      path: attrs.path,
      secure: attrs.secure,
      httpOnly: attrs.httponly,
      sameSite: attrs.samesite,
    });
  });
```

Why this double-encodes:

1. The internal `getSession()` sets the cookie using `setSignedCookie()` → `signCookieValue()` → `encodeURIComponent()` — producing a correctly percent-encoded value (e.g., `%2B` for `+`).
2. `parseSetCookieHeader()` extracts `attrs.value` directly from the raw `Set-Cookie` header string **without decoding**, so the value still contains `%2B`.
3. `ctx.setCookie()` (via `better-call`'s `serializeCookie()`) calls `encodeURIComponent()` again — encoding `%` to `%25` and turning `%2B` into `%252B`.

---

## Environment

* **better-auth:** `1.4.19` (also reproduced in `1.5.0-beta.18`)
* **better-call:** `1.1.8`
* **Framework:** Hono (OpenAPIHono)
* **Runtime:** Bun

---

Oluwatobi-Mustapha:
Fix #8127

Decode forwarded `Set-Cookie` values in `customSession` before calling `ctx.setCookie(...)` so `/get-session` refresh does not double-encode session cookies (`%25`).

- decode parsed cookie value before re-setting cookie
- add regression test for refresh cookie encoding

Tests
- `corepack pnpm --dir packages/better-auth exec vitest run src/plugins/custom-session/custom-session.test.ts --maxWorkers=1`
- `corepack pnpm --dir packages/better-auth exec vitest run src/cookies/cookies.test.ts --maxWorkers=1`

---
## Summary by cubic
Prevents double-encoding of session cookies during get-session refresh in the customSession plugin, fixing #8127. Cookie values are decoded once when parsing Set-Cookie so better-auth.session_token stays unchanged on refresh.

- **Bug Fixes**
  - Centralize decoding in parseSetCookieHeader using a safe tryDecode only when "%" is present; add a unit test for URI-decoded values.
  - Remove redundant decode calls in Next.js, SvelteKit, and TanStack Start integrations.
  - Add a regression test to ensure the session token is identical after refresh and contains no %25.

<sup>Written for commit 1d345729d5493498b7319737a41821d0baea64c8. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/cookies/cookie-utils.ts
+++ b/packages/better-auth/src/cookies/cookie-utils.ts
@@ -1,3 +1,11 @@
+function tryDecode(str: string): string {
+	try {
+		return decodeURIComponent(str);
+	} catch {
+		return str;
+	}
+}
+
 export interface CookieAttributes {
 	value: string;
 	"max-age"?: number | undefined;
@@ -85,7 +93,8 @@ export function parseSetCookieHeader(
 			return;
 		}
 
-		const attrObj: CookieAttributes = { value };
+		const decodedValue = value.includes("%") ? tryDecode(value) : value;
+		const attrObj: CookieAttributes = { value: decodedValue };
 
 		attributes.forEach((attribute) => {
 			const [attrName, ...attrValueParts] = attribute!.split("=");
--- a/packages/better-auth/src/cookies/cookies.test.ts
+++ b/packages/better-auth/src/cookies/cookies.test.ts
@@ -226,6 +226,12 @@ describe("cookie-utils parseSetCookieHeader", () => {
 		);
 	});
 
+	it("decodes URI-encoded cookie values", () => {
+		const header = "token=hello%20world%3Dfoo; Path=/";
+		const map = parseSetCookieHeader(header);
+		expect(map.get("token")?.value).toBe("hello world=foo");
+	});
+
 	it("handles cookie with Expires followed by cookie without Expires", () => {
 		const map = parseSetCookieHeader(
 			"session=xyz; Expires=Mon, 01 Jan 2026 00:00:00 GMT, token=abc",
--- a/packages/better-auth/src/integrations/next-js.ts
+++ b/packages/better-auth/src/integrations/next-js.ts
@@ -100,7 +100,7 @@ export const nextCookies = () => {
 									path: value.path,
 								} as const;
 								try {
-									cookieHelper.set(key, decodeURIComponent(value.value), opts);
+									cookieHelper.set(key, value.value, opts);
 								} catch {
 									// this will fail if the cookie is being set on server component
 								}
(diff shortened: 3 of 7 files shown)


packages/better-auth/src/cookies/cookie-utils.ts
```diff
@@ -1,3 +1,11 @@
+function tryDecode(str: string): string {
+	try {
+		return decodeURIComponent(str);
+	} catch {
+		return str;
+	}
+}
+
```
himself65, line 8: I think we probably try to decode on the top level

himself65 (approved):

Oluwatobi-Mustapha:
Thank you @himself65 🙏🏼

PR #8135, opened 2026-02-24
fix(last-login-method): tracks magic-link auth by default

simonfelding:
last-login method should track magic-link authentication by default. I don't see anywhere in the docs why it shouldnt :)


simonfelding:
@himself65 Hi! Thanks for working on this PR so fast. Can I do anything to help it along? I'm thinking the bot suggestion to write a test for the magic link case might be a good idea, but I'm not entirely sure how to write a test for failed verification. If you have time, maybe you could help me out?

edit: i think I figured something out.

himself65 (approved):

Issue #8125, opened 2026-02-24
SvelteKit integration forces secret definition on build time


megla-hburdow:
### Is this suited for github?

- [ ] Yes, this is suited for github

### To Reproduce

You can use the sveltekit template. The issue is present there:
1. `npx sv create example-project`
2. select better-auth in the stepper
3. remove the .env file
4. `pnpm run build`

### Current vs. Expected behavior

Based on the `auth.ts` file under `src/lib/server/auth.ts`. I would expact, that the secret gets dynamicly set on runtime.

(see "Auth config" section for reference...)
In the Auth config it uses the `import { env } from '$env/dynamic/private';` import [which should get the env var on runitme](https://svelte.dev/docs/kit/$env-dynamic-private). Because of that I would expect, that I can set the env var on runtime and not on build time.

The build seems to already require the `BETTER_AUTH_SECRET` in the build context and fails with the following error if not present:
```log
2026-02-24T12:36:51.160Z WARN [Better Auth]: [better-auth] Base URL could not be determined. Please set a valid base URL using the baseURL config option or the BETTER_AUTH_BASE_URL environment variable. Without this, callbacks and redirects may not work correctly.

node:internal/event_target:1122
  process.nextTick(() => { throw err; });
                           ^
[Error [BetterAuthError]: You are using the default secret. Please set `BETTER_AUTH_SECRET` in your environment variables or pass `secret` in your auth config.]

Node.js v24.13.0
 ELIFECYCLE  Command failed with exit code 1.
```

### What version of Better Auth are you using?

1.4.18

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Pro",
    "release": "10.0.26200",
    "cpuCount": 20,
    "cpuModel": "13th Gen Intel(R) Core(TM) i7-13800H",
    "totalMemory": "31.66 GB",
    "freeMemory": "8.38 GB"
  },
  "node": {
    "version": "v24.13.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "10.30.2"
  },
  "frameworks": [
    {
      "name": "svelte",
      "version": "^5.51.0"
    },
    {
      "name": "@sveltejs/kit",
      "version": "^2.50.2"
    }
  ],
  "databases": [
    {
      "name": "@libsql/client",
      "version": "^0.17.0"
    },
    {
      "name": "drizzle",
      "version": "^0.45.1"
    }
  ],
[...]

Issue #8061, opened 2026-02-19
SvelteKit build fails based on environment variables


rosano:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

run `vite build` for SvelteKit app without setting environment variables or inside a Dockerfile

### Current vs. Expected behavior

Build fails with the following error:

```
2026-02-19T07:31:36.197Z WARN [Better Auth]: [better-auth] Base URL could not be determined. Please set a valid base URL using the baseURL config option or the BETTER_AUTH_BASE_URL environment variable. Without this, callbacks and redirects may not work correctly.

node:internal/event_target:1105
  process.nextTick(() => { throw err; });
                           ^

Emitted 'error' event on Worker instance at:
    at [kOnErrorMessage] (node:internal/worker:333:10)
    at [kOnMessage] (node:internal/worker:344:37)
    at MessagePort.<anonymous> (node:internal/worker:233:57)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:831:20)
    at MessagePort.<anonymous> (node:internal/per_context/messageport:23:28) {
  name: 'BetterAuthError'
}
```

### What version of Better Auth are you using?

1.4.18

### System info

```bash
error: unknown command 'info'


macOS 12.7.6, nodejs 20.19.5
```

### Which area(s) are affected? (Select all that apply)

Package

### Auth config (if applicable)

```typescript

```

### Additional context

Seems like the environment variables should be set at runtime, not build time, but an error gets thrown by these lines when building:

https://github.com/better-auth/better-auth/blob/8fe651400a2cd03c427874beb55f74446ec6a9e6/packages/better-auth/src/context/create-context.ts#L56-L60

AlexsandroBezerra:
Same here. Looks like It is ignoring that's in build time, and runs like in production.

himself65:
Have you provided the secret and the base URL during the prod build?

rosano:
> Have you provided the secret and the base URL during the prod build?

I have not because my app is built as part of a Dockerfile to be deployed as a Node.js server-side app and unless I've mistaken it seems that the secret and base URL should come from runtime environment variables in this case. I think better-auth assumes it's production when building whereas it might fit this case better to determine the environment at runtine.

rosano:
Using the fallback values as described by dosu.

megla-hburdow:
This seems more like a workaround than a fix to me.

Setting fallback values will remove the secret validation also for real production cases.
[...]

Issue #288, opened 2024-10-22
Sveltekit + Svelte 5 error on build


gozilla-paradise:
**Describe the bug**
The app can run in dev but cannot be built.

Here is the following betterauth setup

`src/lib/server/auth/index.ts`

```
import { betterAuth } from 'better-auth';
import { drizzleAdapter } from 'better-auth/adapters/drizzle';
import { db } from '../db';
import { env } from '$env/dynamic/public';
import * as schema from '../db/schema';

export const auth = betterAuth({
	emailAndPassword: {
		enabled: true
	},
	database: drizzleAdapter(db, {
		provider: 'pg',
		schema: {
			...schema
		}
	}),
	baseURL: env.PUBLIC_BETTER_AUTH_URL
});
```

and

`src/routes/backoffice/+layout.server.ts`

```import { auth } from '$lib/server/auth';
import { redirect } from '@sveltejs/kit';
import type { LayoutServerLoad } from './$types';

export const load = (async ({ request }) => {
	const session = await auth.api.getSession({
		headers: request.headers
	});

	if (!session) {
		throw redirect(307, '/backoffice');
	}

	return {};
}) satisfies LayoutServerLoad;
```

**To Reproduce**

1. Have simple better-auth setup with drizzle
2. Add server-side session checking code
3. It is error
![CleanShot 2567-10-22 at 12 44 57@2x](https://github.com/user-attachments/assets/583c516e-50a3-458b-b94d-21d913d8d8b4)

**Expected behavior**
The app can be built properly

**Desktop (please complete the following information):**

- OS: OSX
- Browser Edge

** Software **
- Sveltekit 2.7.2
- Svelte 5.0.5
- Drizzle 0.35.3
- Postgres 3.4.4
- Node 20.16.0

PR #8148, opened 2026-02-25
fix: Add test to avoid throwing errors during SvelteKit build process

simonfelding:
Not very pretty, but it should work. There is no other way to do this.

Also exports a new function to check if we are in a sveltekit build process - may come in handy later.

Solves #8125 #8061 #288 once and for all :)

NOTE: I haven't tested this yet. I don't really see how I can build a good test for this. Open to feedback.

---
## Summary by cubic
Detect SvelteKit builds and treat them as test mode to prevent env errors during build. Exposes a helper to detect SvelteKit build state.

- **Bug Fixes**
  - Detect build via $app/environment.building, make isTest return true during build, and add knip ignoreUnresolved for $app/environment to silence unresolved import warnings.

- **New Features**
  - Export isSvelteKitBuilding for future checks.

<sup>Written for commit 436585c9cc03258ef1530ae08c26a96facdc49ea. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/src/env/env-impl.ts
+++ b/packages/core/src/env/env-impl.ts
@@ -59,7 +59,6 @@ export const isDevelopment = () =>
 /** Detect if we are currently building a SvelteKit project */
 let _isSvelteKitBuilding = false;
 try {
-	/** @lintignore */
 	// @ts-expect-error Will only work if we are in SvelteKit
 	const svelteKitEnv = await import("$app/environment");
 	_isSvelteKitBuilding = svelteKitEnv.building;
--- a/packages/core/src/env/env-impl.ts
+++ b/packages/core/src/env/env-impl.ts
@@ -59,10 +59,10 @@ export const isDevelopment = () =>
 /** Detect if we are currently building a SvelteKit project */
 let _isSvelteKitBuilding = false;
 try {
-	// @ts-ignore Will only work if we are in SvelteKit
+	// @ts-expect-error Will only work if we are in SvelteKit
 	const svelteKitEnv = await import("$app/environment");
 	_isSvelteKitBuilding = svelteKitEnv.building;
-} catch (e) {
+} catch (_e) {
 	// If it fails (e.g., we aren't in SvelteKit), it remains false
 }
 
@@ -71,7 +71,8 @@ export function isSvelteKitBuilding() {
 }
 
 /** Detect if `NODE_ENV` environment variable is `test` */
-export const isTest = () => nodeENV === "test" || toBoolean(env.TEST) || isSvelteKitBuilding();
+export const isTest = () =>
+	nodeENV === "test" || toBoolean(env.TEST) || isSvelteKitBuilding();
 
 /**
  * Get environment variable with fallback
--- a/packages/core/src/env/env-impl.ts
+++ b/packages/core/src/env/env-impl.ts
@@ -56,8 +56,22 @@ export const isProduction = nodeENV === "production";
 export const isDevelopment = () =>
 	nodeENV === "dev" || nodeENV === "development";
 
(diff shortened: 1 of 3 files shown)


simonfelding:
I don't know how to get the linter to skip this :/

packages/core/src/env/env-impl.ts
```diff
@@ -56,8 +56,22 @@ export const isProduction = nodeENV === "production";
 export const isDevelopment = () =>
 	nodeENV === "dev" || nodeENV === "development";
 
+/** Detect if we are currently building a SvelteKit project */
+let _isSvelteKitBuilding = false;
+try {
+	// @ts-ignore Will only work if we are in SvelteKit
+	const svelteKitEnv = await import("$app/environment");
```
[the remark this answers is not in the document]
simonfelding: This is a scammy way to fix the problem.

packages/core/src/env/env-impl.ts
```diff
@@ -56,8 +56,23 @@ export const isProduction = nodeENV === "production";
 export const isDevelopment = () =>
 	nodeENV === "dev" || nodeENV === "development";
 
+/** Detect if we are currently building a SvelteKit project */
```
himself65, line 59: Don't put a top-level await import in the core package

himself65 (changes requested):

packages/core/src/env/env-impl.ts
simonfelding: Ok. Can you say a few words on why? Any tips on where I should put it?

It's not a real module and doesn't add any dependencies. Sveltekit generates it and it exports the "building" boolean. So it will only exist in a sveltekit environment. It's the only way to test if it's running in a sveltekit build environment.

simonfelding:
@himself65 

You said in the review:
>Don't put a top-level await import in the core package

Ok, no problem, but do you have any tips on where I should put it instead?

It's not a real module and doesn't add any dependencies. Sveltekit generates it and it exports the "building" boolean. So it will only exist in a sveltekit environment. It's the only way to test if it's running in a sveltekit build environment.

megla-hburdow:
I think we need a reliable way to detect if the current env uses SvelteKit, but as far as I know, there are no good solutions for this.
There is like some kind of workaround implementation which is used in the Netifly adapter of svelte. This implementation exposes a `SVELTEKIT_FORK: 'true'` env, which **could** be used to identify a SvelteKit project at build time.

In the end I think this approach is the wrong one...
1. We should probably try another one, like doing this `isBuildTime` check in the `{ svelteKitHandler } from 'better-auth/svelte-kit'` and passing down a flag to disable the secret check.
2. Another option could be to question the change that caused this issue in general. Why do we need to throw an Error when the default secret gets used in prod? Could a error log maybe be sufficient here?

simonfelding:
@megla-hburdow I think the solution of just logging a warning instead of throwing an error would be fine. It's still kind of annoying though to even get that during build, but I'm ok with either solution.

I don't think we can add the isBuildTime to better-auth/svelte-kit because I'm pretty sure it would then have to be a dependency for the better-auth base project, which defeats the purpose of seperating them.

But hm, maybe the better-auth/svelte-kit can somehow override the behavior of throwing an error or pass the isBuildTime somehow, but I can't really think of a clean way to do it.

megla-hburdow:
@simonfelding I currently don't have enough time to develop a proper fix. I can put this one on my task list, but not sure when I will have time for that...

Who could we ask for reverting the behavior change of throwing an error in `no secrete production cases`?

typed-sigterm:
Because many other SSR frameworks also suffers from this error, I think the solution is to defer the check to the first request / database op, or just have a `BETTER_AUTH_BYPASS_SECRET_CHECK`

PR #8161, opened 2026-02-26
fix: preserve fresh account cookie set during OAuth callback

### LINKED ISSUE #8159: `setCookieCache` overwrites fresh account cookie with stale request data during OAuth re-login

### Minimal Config

```tsx
betterAuth({
	session: {
		cookieCache: { enabled: true, maxAge: 604800, strategy: "jwe" },
	},
	account: {
		storeStateStrategy: "cookie",
		storeAccountCookie: true,
	},
	plugins: [genericOAuth({ config: [auth0({ ... })] })],
});
```

### Steps

1. Log in via OAuth. Session and account cookies are set. Everything works.
2. Wait for the OAuth access token to expire (or simulate by letting it lapse).
3. The app redirects the user to the sign-in page. The old session and account cookies still exist in the browser.
4. Log in again via OAuth.
5. The callback completes, new session and account cookies are set but the account cookie contains the old tokens, not the new ones from the fresh OAuth exchange.
6. Any API call using the access token from the account cookie fails because the token is stale/expired.
7. Clearing all cookies and logging in again works, because there is no old account cookie for `setCookieCache` to read.

### Current behavior

When a user with an existing (stale) session logs in again (e.g. after being redirected to the sign-in page due to an expired access token), the fresh OAuth tokens from the new login are lost. The account data cookie ends up containing the old/stale tokens from the previous session.

### Expected Behavior

After a successful OAuth login, the account cookie should always contain the tokens from the new login, regardless of whether the browser had an old account cookie.

1.4.18

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "cpuCount": 16,
    "cpuModel": "AMD Ryzen 7 7745HX with Radeon Graphics",
    "totalMemory": "15.22 GB",
    "freeMemory": "4.90 GB"
  },
  "node": {
    "version": "v25.6.1",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "10.28.2"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.1.6"
    },
    {
      "name": "react",
      "version": "19.2.4"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "^1.4.18",
    "config": null
  }
}
```

Client

```typescript
betterAuth({
  secret: env.AUTH0_SECRET || "dummy",
  baseURL: env.AUTH0_BASE_URL || "http://localhost:3000",

  session: {
    cookieCache: {
      enabled: true,
      maxAge: 7 * 24 * 60 * 60,
      strategy: "jwe",
      refreshCache: true,
    },
  },
  account: {
    storeStateStrategy: "cookie",
    storeAccountCookie: true,
  },

  onAPIError: {
    errorURL: "/auth/error",
  },

  advanced: {
    cookiePrefix: "my-auth-app",
  },

  plugins: [
    genericOAuth({
      config: [
        auth0({
          clientId: env.AUTH0_CLIENT_ID || "dummy",
          clientSecret: env.AUTH0_CLIENT_SECRET || "dummy",
          domain: urlToDomain(env.AUTH0_ISSUER_BASE_URL || "auth.myauth.io"),
          redirectURI: getRedirectURL(),
        }),
      ],
    }),

    nextCookies(),
  ],
});
```

### Additional context

_No response_

arunanshub:
## Problem

When `storeAccountCookie` is enabled, the OAuth callback flow sets a fresh account cookie (with new tokens from the provider) in `handleOAuthUserInfo`. Immediately after, `setSessionCookie` calls [`setCookieCache`](https://github.com/better-auth/better-auth/blob/84565ccc2a6bc0ae704d290028691adc7e9a4957/packages/better-auth/src/cookies/index.ts#L147), which reads the account cookie from the incoming request (stale data) and writes it back, overwriting the fresh cookie.

This means users end up with expired OAuth tokens after every re-login unless they fully clear their cookies first.

Root cause (`src/cookies/index.ts` lines 255-261)

https://github.com/better-auth/better-auth/blob/84565ccc2a6bc0ae704d290028691adc7e9a4957/packages/better-auth/src/cookies/index.ts#L255-L261

`getAccountCookie` reads from the incoming request via `ctx.getCookie`/`ctx.headers`. It has no awareness of cookies already set in the current response.

## Fix

Check `ctx.responseHeaders` (the `better-call` response accumulator that `ctx.setCookie` appends to) before deciding whether to refresh the account cookie. If the account cookie was already set in this response, skip the refresh since the fresh data is already there.

Also adds `responseHeaders` attribute to `GenericEndpointContext` in `@better-auth/core`, since this property exists at runtime (from better-call's `createInternalContext`) but was not exposed on the public type.

Fixes #8159

---
## Summary by cubic
Prevents stale request data from overwriting the fresh account cookie during OAuth callbacks, so users keep new tokens after re-login. Fixes #8159.

- **Bug Fixes**
  - In `setCookieCache`, skip account cookie refresh if it was already set in this response (detected via `set-cookie` on `ctx.responseHeaders`). Session refresh behavior is unchanged.
  - Added a re-login test that decrypts the cookie and verifies `getAccessToken` returns the fresh token.

- **Refactors**
  - Exposed optional `responseHeaders?: Headers` on `GenericEndpointContext` and used it for the in-response cookie check.

<sup>Written for commit d0e63632673a8fca8d2ccc078dcd1b20d300359c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/8161?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/better-auth/src/api/routes/account.test.ts
+++ b/packages/better-auth/src/api/routes/account.test.ts
@@ -1661,6 +1661,156 @@ describe("account", async () => {
 		expect(reLoginAccessToken.data?.accessToken).toBe("test");
 	});
 
+	it("should not overwrite fresh account cookie with stale request data on re-login", async () => {
+		const { auth, client, cookieSetter } = await getTestInstance({
+			socialProviders: {
+				google: {
+					clientId: "test",
+					clientSecret: "test",
+					enabled: true,
+				},
+			},
+			account: {
+				storeAccountCookie: true,
+			},
+			session: {
+				cookieCache: {
+					enabled: true,
+					strategy: "jwe",
+				},
+			},
+		});
+
+		const ctx = await auth.$context;
+		const accountDataCookieName = ctx.authCookies.accountData.name;
+
+		const headers = new Headers();
+		email = "stale-cookie-test@test.com";
+
+		// First login
+		const signInRes = await client.signIn.social({
+			provider: "google",
+			callbackURL: "/callback",
+			fetchOptions: {
+				onSuccess: cookieSetter(headers),
+			},
+		});
+
+		const state =
+			signInRes.data && "url" in signInRes.data && signInRes.data.url
+				? new URL(signInRes.data.url).searchParams.get("state") || ""
+				: "";
+
+		await client.$fetch("/callback/google", {
+			query: { state, code: "test" },
+			headers,
+			method: "GET",
+			onError(context) {
+				expect(context.response.status).toBe(302);
+				cookieSetter(headers)({ response: context.response });
+			},
+		});
+
+		// Verify first login tokens
+		const firstAccessToken = await client.getAccessToken(
+			{ providerId: "google" },
+			{ headers },
+		);
+		expect(firstAccessToken.data?.accessToken).toBe("test");
+
+		// Change the mock to return different tokens for the second login.
+		// This lets us distinguish fresh tokens from stale ones.
+		server.use(
+			http.post(
+				"https://oauth2.googleapis.com/token",
+				async () => {
+					const data: GoogleProfile = {
+						email,
+						email_verified: true,
+						name: "First Last",
+						picture:
+							"https://lh3.googleusercontent.com/a-/AOh14GjQ4Z7Vw",
+						exp: 1234567890,
+						sub: "1234567890",
+						iat: 1234567890,
+						aud: "test",
+						azp: "test",
+						nbf: 1234567890,
+						iss: "test",
+						locale: "en",
+						jti: "test",
+						given_name: "First",
+						family_name: "Last",
+					};
+					const testIdToken = await signJWT(data, DEFAULT_SECRET);
+					return HttpResponse.json({
+						access_token: "refreshed_token",
+						refresh_token: "refreshed_refresh",
+						id_token: testIdToken,
+					});
+				},
+				{ once: true },
+			),
+		);
+
+		// Re-login WITHOUT signing out. The old account cookie from the
+		// first login is still in the request headers. This is the scenario
+		// that triggers the bug: setCookieCache re-reads the old cookie from
+		// the request and overwrites the fresh one from handleOAuthUserInfo.
+		const reLoginRes = await client.signIn.social({
+			provider: "google",
+			callbackURL: "/callback",
+			fetchOptions: {
+				onSuccess: cookieSetter(headers),
+			},
+		});
+
+		const reLoginState =
+			reLoginRes.data && "url" in reLoginRes.data && reLoginRes.data.url
+				? new URL(reLoginRes.data.url).searchParams.get("state") || ""
+				: "";
+
+		let freshAccountCookieValue: string | undefined;
+		await client.$fetch("/callback/google", {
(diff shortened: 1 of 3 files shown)


himself65 (commented):
## Code Review

### Overview

This PR fixes a real bug where `setCookieCache` overwrites fresh OAuth tokens with stale request data during re-login. The root cause analysis is thorough and the fix is minimal and correct.

### Correctness

The core fix is correct. `setCookieCache` runs during handler execution, at which point `ctx.context.responseHeaders` is still `undefined` (it gets populated after the handler returns in `to-auth-endpoints.ts:148`). Using `ctx.responseHeaders` (better-call's internal `Headers` accumulator that `ctx.setCookie` writes to) is the right call here.

The test is well-structured — it sets up a realistic two-login scenario with different mock tokens, verifies both the raw cookie and decrypted content, and uses `{ once: true }` on the MSW override.

### Suggestions

1. **Add a comment explaining why `ctx.responseHeaders` is used instead of `ctx.context.responseHeaders`.**
   Every other usage in `packages/better-auth/src` (15+ occurrences) uses `ctx.context.responseHeaders`. This is the only place that uses `ctx.responseHeaders`. The reason is valid — during handler execution `ctx.context.responseHeaders` is `undefined` while `ctx.responseHeaders` already has cookies set via `ctx.setCookie` — but a brief comment would help future maintainers understand why this diverges from the pattern.

2. **Minor: the `setCookieHeader.includes(accountCookieName)` check could theoretically substring-match against a different cookie name.** A slightly more robust check would be `includes(`${accountCookieName}=`)` to anchor against the `=` delimiter. Low risk given better-auth's naming conventions, but easy to tighten.

### Verdict

Looks good overall. The fix is correct, well-tested, and minimal. Just the two minor suggestions above.

arunanshub:
@Bekacru @himself65 gentle ping here.

This PR fixes #8159 and is important for our production rollout. Without it, OAuth re-login can preserve expired tokens because the fresh account cookie set during the callback is overwritten by stale request data.

The patch is intentionally small and tested. If there's any concern with the approach, I'm very happy to revise quickly.

jbiglangawa:
Doubling down on this as well, I'm also experiencing the same issue with a new app supposed to rollout soon so having this fix go in will incredibly help!

jbiglangawa:
Oh actually @arunanshub in cases where your cookies get chunked because they went past 4KB due to cookie size limits you'll have multiple cookies. In my case this happens so I had to remove the "=" because i had two cookies: account_data.0 and account_data.1. Perhaps you can consider this on your PR as well.
image

bytaesu:
Closing this PR since #8159 is already addressed by #9987

bytaesu:
> Closing this PR since #8159 is already addressed by #9987

I only left a comment 🤦‍♂️

PR #8174, opened 2026-02-26
fix(api): respect freshAge in deleteUser

### LINKED ISSUE #8173: deleteUser don't respect freshAge

### To Reproduce

1. Create a project with
```json
"better-auth": "^1.4.19",
"next": "^16.1.6",
```
2. Create acount and try to use deleteUser to delete it after the freshAge expires without providing a password.

Current:
deleteUser doesn't respect freshAge
Expected:
The delete operation should respect the freshAge per the [documentation](https://www.better-auth.com/docs/concepts/users-accounts#authentication-requirements)

1.4.19

### System info

```bash
N/A
```

Backend, Documentation

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

bytaesu:
freshAge was multiplied by 1000 twice

- Closes #8173

### DIFF
--- a/packages/better-auth/src/api/routes/update-user.test.ts
+++ b/packages/better-auth/src/api/routes/update-user.test.ts
@@ -491,6 +491,53 @@ describe("delete user", async () => {
 		});
 	});
 
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/8173
+	 */
+	it("should require password when session is no longer fresh", async () => {
+		const { client, signInWithTestUser, db } = await getTestInstance({
+			user: {
+				deleteUser: {
+					enabled: true,
+				},
+			},
+			session: {
+				freshAge: 1,
+			},
+		});
+
+		const { headers } = await signInWithTestUser();
+		const currentSession = await client.getSession({
+			fetchOptions: {
+				headers,
+			},
+		});
+		const sessionId = currentSession.data?.session.id;
+		expect(sessionId).toBeDefined();
+
+		await db.update({
+			model: "session",
+			where: [
+				{
+					field: "id",
+					value: sessionId!,
+				},
+			],
+			update: {
+				createdAt: new Date(Date.now() - 5_000),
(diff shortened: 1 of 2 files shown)


Laurin-Notemann:
I am curious @bytaesu why wouldn't you use freshSessionMiddleware here? 

I have a feeling that middleware seems a bit useless, as it is only used in two two occasions:
- generatePasskeyRegistrationOptions
- unlinkAccount

and even the freshAge config is only used here in the delete User function.

Too me it seems either redundant with sensitiveSessionMiddleware (although i know the logic is quite different, but it in my opinion everywhere where you would want to use freshSessionMiddleware the sensitiveSessionMiddleware is being used) 

Maybe you can help me understand how and why that is? 

And one thing, kind of a feature request almost, it would make sense to have an endpoint from which you can query if you session is fresh or not, in case someone wants to use the freshAge logic elsewhere in their app, because currently the only way to really use it is to create own logic to actually check aginst the freshness of a session, which seems unnecessary when we are already storing that logic in our better auth session.

bytaesu:
> [!NOTE]
> https://better-auth.com/docs/concepts/users-accounts#authentication-requirements 

Hi @Laurin-Notemann,

`deleteUser` currently also supports non-fresh re-verification paths, such as providing the password.

Because of that, applying `freshSessionMiddleware` at the route level would change the endpoint behavior: a stale but otherwise valid session could be rejected before the handler reaches the password-based branch.

Also, `sensitiveSessionMiddleware` and `freshSessionMiddleware` serve different purposes. `sensitiveSessionMiddleware` bypasses cookie cache and resolves the session from the source of truth, while `freshSessionMiddleware` enforces the `freshAge` check. So I do think the separation of responsibilities makes sense here.

What this PR fixed was the inconsistency in the existing `deleteUser` freshness check. It did not try to restructure the endpoint behavior.

PR #8229, opened 2026-03-01
docs: fix home-page demo video path

ping-maxwell:


ping-maxwell:
> What is the propose of getSrc? Seems like all usage of it have failed

I think it was introduced when we used a proxy system, so we can define the domain in ENV

Issue #8228, opened 2026-03-01
Schema generation removes magic link `verifications` & Magic link sign-in broken since `1.5.0`


wottpal:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

We've upgraded to `1.5.0` and ran a new schema generation which want's to remove the `verifications` table. Even if I bring it back manually and don't drop it from the database (= don't apply migration), magic sign in has stopped working, we get the following error upon magic link sign in:

```
next:dev    | 2026-03-01T05:22:54.919Z ERROR [Better Auth]: BetterAuthError [Error [BetterAuthError]: Model "verification" not found in schema]
next:dev    | # SERVER_ERROR:  [Error [BetterAuthError]: Model "verification" not found in schema]
next:dev    |  GET /api/auth/magic-link/verify?token=6cbc450e42e8850cd5d6b1808faee6c6&callbackURL=http%3A%2F%2Flocalhost%3A3000%2Fapp 500 in 93ms (compile: 6ms, render: 87ms)
```

Also, it seems that it could also be related with `usePlural: true` (because it is searching for `verification` even though we use pluralization). Even if i try to manually give it the `verification` singular like below it doesn't work:

```
    schema: {
      ...schema,
      verification: schema.verifications,
    },
```

<img width="1116" height="662" alt="Image" src="https://github.com/user-attachments/assets/940011c3-2c1a-4bd2-ba25-0d9e6b013993" />

### Current vs. Expected behavior

* I expect the `verifications` table not to be removed.
* I expect magic sign-in's to continue working.

### What version of Better Auth are you using?

1.5.0

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.3.0: Wed Jan 28 20:48:41 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T6041",
    "release": "25.3.0",
    "cpuCount": 16,
    "cpuModel": "Apple M4 Max",
    "totalMemory": "128.00 GB",
    "freeMemory": "68.45 GB"
  },
  "node": {
    "version": "v24.14.0",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.10"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "^16.1.6"
    },
    {
      "name": "react",
      "version": "^19.2.4"
    }
  ],
  "databases": [
    {
      "name": "pg",
[...]

PR #8242, opened 2026-03-01
fix(db): support verification CRUD operations with secondary storage

himself65:
When secondaryStorage is enabled without verification.storeInDatabase, updateVerificationValue and deleteVerificationValue would fail with "Model verification not found in schema" because they always tried to hit the database directly.

Fixes: https://github.com/better-auth/better-auth/issues/8228

### DIFF
--- a/packages/better-auth/src/db/internal-adapter.ts
+++ b/packages/better-auth/src/db/internal-adapter.ts
@@ -1044,6 +1044,7 @@ export const createInternalAdapter = (
 
 			const verification = await createWithHooks(
 				{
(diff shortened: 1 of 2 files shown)


packages/better-auth/src/db/internal-adapter.ts
```diff
@@ -1044,6 +1044,7 @@ export const createInternalAdapter = (
 
 			const verification = await createWithHooks(
 				{
+					id: generateId(),
```
himself65, line 1047: Id is not necessary here

PR #8249, opened 2026-03-01
chore: remove `demo/*` from workspace to prevent flaky install

himself65:
The electron postinstall script intermittently fails on CI with 502 errors when downloading binaries. Since demo apps are not used in CI workflows, remove them from the pnpm workspace to avoid this flaky failure.


himself65:
Will do after https://github.com/better-auth/better-auth/pull/8250

PR #8276, opened 2026-03-02
fix(sso): prefer UserInfo endpoint over ID token and map sub claim correctly

### LINKED ISSUE #8269: [BUG] SSO module looking for non-existent ID claim

### To Reproduce

Using an OIDC server like [Tinyauth](https://github.com/steveiliop56/tinyauth) (which does not include the user-info claims in the ID token), set-up a demo application with SSO and try to login. You will be greeted with the `error=invalid_provider&error_description=missing_user_info` error.

Looking through the code, specifically here:

https://github.com/better-auth/better-auth/blob/425c6abf6eb703ad6a77d5b9dd663e24fa2027d5/packages/sso/src/routes/sso.ts#L1714-L1720

Seems like better-auth is trying to extract some `id` claim from the user-info endpoint. However, according to [Section 5.1](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) of the OIDC specification, the OIDC server is not obligated to return such claim. Normally, better-auth should be looking for the `sub` claim (containing the ID of the user that remains static despite of profile changes) instead of an `id` one.

I am guessing that since most OIDC providers include the claims in the ID token (which is not required according to [Section 2](https://openid.net/specs/openid-connect-core-1_0.html#IDToken)), the user-info extraction logic from the ID token happens first and thus there is no need to check the user-info endpoint which would most likely fail.

1.5.0

### System info

```bash
Not relevant to the issue.
```

Client

```typescript
Not applicable.
```

### Additional context

_No response_

himself65:
## Summary

- The UserInfo endpoint fallback path was checking `userInfo.id`, but standard OIDC providers return `sub` (not `id`) as the subject identifier per [OIDC spec Section 5.1](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims)
- Applied the same claim mapping logic (`sub`→`id`, `picture`→`image`, `email_verified`→`emailVerified`, `extraFields`) to the UserInfo endpoint path that already existed for the ID token path
- Updated the `betterFetch` generic type to `Record<string, unknown>` to accurately reflect raw OIDC UserInfo responses

Fixes #8269

## Test plan

- [ ] Verify SSO login works with an OIDC provider that does not include user claims in the ID token (e.g., Tinyauth)
- [ ] Verify SSO login still works with providers that include claims in the ID token (existing behavior unchanged)
- [ ] All existing SSO tests pass (`pnpm vitest packages/sso/src --run`)

### DIFF
--- a/packages/sso/src/oidc.test.ts
+++ b/packages/sso/src/oidc.test.ts
@@ -742,14 +742,14 @@ describe("provisioning", async (ctx) => {
 			headers,
 		});
 		const member = org?.members.find(
-			(m: any) => m.user.email === "sso-user@localhost:8000.com",
+			(m: any) => m.user.email === "test@localhost.com",
 		);
 		expect(member).toMatchObject({
 			role: "member",
 			user: {
 				id: expect.any(String),
-				name: "Test User",
-				email: "sso-user@localhost:8000.com",
+				name: "OAuth2 Test",
+				email: "test@localhost.com",
 				image: "https://test.com/picture.png",
 			},
 		});
--- a/packages/sso/src/routes/sso.ts
+++ b/packages/sso/src/routes/sso.ts
@@ -1630,7 +1630,43 @@ async function handleOIDCCallback(
 		emailVerified?: boolean;
 		[key: string]: any;
 	} | null = null;
-	if (tokenResponse.idToken) {
+	const mapping = config.mapping || {};
+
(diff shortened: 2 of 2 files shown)


steveiliop56:
@himself65 maybe you should also prioritize the user-info endpoint instead of the ID token since it tends to be more reliable.

PR #8291, opened 2026-03-02
refactor: use verification table instead of jwt for email verification

jslno:
Fixes #6573


himself65 (approved):
I think this is good, but switching to DB-based verification might affect the performance of the memory-based solution.

Also, I concern this could have the timing-attack simulation writing to DB

What do you think

PR #8292, opened 2026-03-02
feat(oauth-provider): pairwise subject identifiers (OIDC Core §8)

### LINKED ISSUE #8288: oauth-provider: pairwise subject identifiers documented in types but not implemented
## To Reproduce

1. Register a client via DCR with `subject_type: "pairwise"` in the request body
2. Complete an authorization code flow for two different clients with the same user
3. Compare the `sub` claim in the id_tokens

**Current:** The `sub` claim is always `user.id` regardless of client. Both clients receive the same `sub`, making the user linkable across RPs. The `subject_type` field from DCR is silently stored in client `metadata` (via the `[key: string]: unknown` index signature) but never read by the runtime.

The types acknowledge pairwise as a concept (`types/oauth.ts:206-213`):

```typescript
/**
 * pairwise: the subject identifier is unique to the client
 * public: the subject identifier is unique to the server
 *
 * only `public` is supported.
 */
subject_types_supported: "public"[];
```

And `sub: user.id` is hardcoded after the custom claims spread in `token.ts`, so `customIdTokenClaims` can't override it either:

```typescript
// token.ts:154-166 — createIdToken
const payload: JWTPayload = {
  ...customClaims,   // hook output
  ...userClaims,     // has sub: user.id
  sub: user.id,      // hardcoded last → overwrites
};
```

**Expected:** When a client registers with `subject_type: "pairwise"`, the `sub` should be a per-client pseudonymous identifier as defined in [OIDC Core Section 8.1](https://openid.net/specs/openid-connect-core-1_0.html#PairwiseAlg), so different RPs cannot correlate users. This is supported by all major OIDC providers and required for OIDC conformance when advertising pairwise support.

1.5.0-beta.18

(`@better-auth/oauth-provider` source checked at latest `main`)

## System info

```json
{
  "system": { "platform": "darwin", "arch": "arm64" },
  "frameworks": [{ "name": "next", "version": "16.1.6" }],
  "databases": [{ "name": "@libsql/client", "version": "0.17.0" }, { "name": "drizzle", "version": "0.45.1" }],
  "betterAuth": { "version": "1.5.0-beta.18" }
}
```

gustavovalverde:
## Summary

Adds opt-in **pairwise subject identifier** support to `@better-auth/oauth-provider`, implementing [OIDC Core Section 8](https://openid.net/specs/openid-connect-core-1_0.html#SubjectIDTypes). When enabled, each relying party receives a unique, unlinkable `sub` for the same user — preventing cross-RP user tracking.

Closes #8288

### How it works

- **New config option**: `pairwiseSecret` — a server-side HMAC-SHA256 key
- **Per-client opt-in**: `subject_type: "pairwise"` during client registration (admin or DCR)
- **Sector identifier**: derived from the client's redirect URI host per OIDC Core §8.1 — same-host clients share the same sector and get the same pairwise sub
- **Scope boundary**: pairwise `sub` appears in id_tokens, `/userinfo`, and introspection. JWT access tokens retain the real `user.id` (needed for user lookup by resource servers)

### Configuration

```typescript
oauthProvider({
  loginPage: "/login",
  consentPage: "/consent",
  pairwiseSecret: process.env.PAIRWISE_SECRET, // 32+ char HMAC key
})
```

Then register clients with `subject_type: "pairwise"`:
```typescript
await auth.api.adminCreateOAuthClient({
  body: {
    redirect_uris: ["https://rp.example.com/callback"],
    subject_type: "pairwise",
    // ...
  },
})
```

### Backward compatible

No `pairwiseSecret` configured = pure `public` behavior, identical to current. All existing tests pass unchanged.

---
## Summary by cubic
Adds opt-in pairwise subject identifiers to @better-auth/oauth-provider per OIDC Core §8 so each client gets a unique, unlinkable sub for the same user. Applied to id_token, /userinfo, and introspection; discovery advertises "pairwise" only when enabled; defaults to public otherwise.

- **New Features**
  - pairwiseSecret (32+ chars, HMAC-SHA256) enables pairwise subs; sector = first redirect_uri host.
  - Client registration accepts subject_type: "public" | "pairwise"; rejected if no server secret, or if redirect_uris use different hosts (until sector_identifier_uri is supported).
  - JWT access tokens keep real user.id; id_token, userinfo, and introspection return pairwise sub.
  - Userinfo/introspection sub resolution optimized to avoid unnecessary DB calls and avoid payload mutation.

- **Migration**
  - Set pairwiseSecret in oauthProvider config (32+ chars).
  - Register or update clients with subject_type: "pairwise" and ensure all redirect_uris share the same host.
  - No changes needed if you don’t enable pairwise.

<sup>Written for commit 832f984a2ad7089dd4f1c8d167e516883be2da69. Summary will update on new commits.</sup>

### DIFF
--- a/packages/oauth-provider/src/introspect.ts
+++ b/packages/oauth-provider/src/introspect.ts
@@ -383,11 +383,12 @@ async function resolveIntrospectionSub(
 	client: SchemaClient<Scope[]>,
 ): Promise<JWTPayload> {
 	if (payload.active && payload.sub) {
-		payload.sub = await resolveSubjectIdentifier(
+		const resolvedSub = await resolveSubjectIdentifier(
 			payload.sub as string,
 			client,
 			opts,
 		);
+		return { ...payload, sub: resolvedSub };
 	}
 	return payload;
 }
--- a/packages/oauth-provider/src/userinfo.ts
+++ b/packages/oauth-provider/src/userinfo.ts
@@ -82,16 +82,18 @@ export async function userInfoEndpoint(
 
 	const baseUserClaims = userNormalClaims(user, scopes ?? []);
 
-	// Resolve pairwise sub if client is configured for it
-	const clientId = (jwt.client_id ?? jwt.azp) as string | undefined;
-	if (clientId) {
-		const client = await getClient(ctx, opts, clientId);
-		if (client) {
-			baseUserClaims.sub = await resolveSubjectIdentifier(
-				user.id,
-				client,
-				opts,
-			);
+	// Resolve pairwise sub if server has pairwise enabled and client is configured for it
+	if (opts.pairwiseSecret) {
+		const clientId = (jwt.client_id ?? jwt.azp) as string | undefined;
+		if (clientId) {
+			const client = await getClient(ctx, opts, clientId);
+			if (client) {
+				baseUserClaims.sub = await resolveSubjectIdentifier(
+					user.id,
+					client,
+					opts,
+				);
+			}
 		}
 	}
 	const additionalInfoUserClaims =


packages/oauth-provider/src/register.ts
```diff
@@ -110,6 +110,26 @@ export async function checkOAuthClient(
 		});
 	}
 
+	// Validate subject_type
+	if (client.subject_type !== undefined) {
```
himself65, line 114: Validate that all redirect_uris share the same host when subject_type: "pairwise" — or reject registration. This is a MUST in the spec.

gustavovalverde: Thanks for the commit, you solved this and other things more cleanly :)

dvanmali:
@gustavovalverde Can you also add functionality for revoke endpoint for this?

gustavovalverde:
> @gustavovalverde Can you also add functionality for revoke endpoint for this?

@dvanmali I might not be grasping the idea here, because RFC 7009 (for token revocation) doesn't have a direct relationship with pairwise subject identifiers. The changes in this PR operate on the `sub` claim (identity presentation); while revocation operates on raw token strings (token lifecycle), not on `sub` claims.

dvanmali:
> I might not be grasping the idea here, because RFC 7009 (for token revocation) doesn't have a direct relationship with pairwise subject identifiers. The changes in this PR operate on the sub claim (identity presentation); while revocation operates on raw token strings (token lifecycle), not on sub claims.

Apologies, you are correct in your implementation. Just the introspect endpoint but that was patched too

dvanmali (approved):

docs/content/docs/plugins/oauth-provider.mdx
```diff
@@ -1368,6 +1368,56 @@ oauthProvider({
...
+- `/oauth2/userinfo` response
+- Token introspection (`/oauth2/introspect`)
+
+JWT access tokens always use the real user ID as `sub`, since resource servers may need to look up users directly.
+
+<Callout type="warn">
+**Limitations:**
+- `sector_identifier_uri` is not yet supported. All `redirect_uris` for a pairwise client must share the same host. Clients with redirect URIs on different hosts will be rejected at registration.
+- `pairwiseSecret` must be at least 32 characters long.
+- Rotating `pairwiseSecret` will change all pairwise `sub` values, breaking existing RP sessions. Treat this secret as permanent once set.
```
dvanmali, line 1418: Suggestion: To prevent this, we could tie the secret to each client instead at the `schema` level (encrypted of course). Allows rotation and configuration determined by each client. It's a different implementation than this PR but is more future proof. Additionally, the secret cannot be returnable to the client (https://github.com/better-auth/better-auth/blob/b73b0440a2c7ae2d0bb331b72087ff3ea2723a65/packages/oauth-provider/src/register.ts#L9).

himself65:
@dvanmali Thank for reviewing

docs/content/docs/plugins/oauth-provider.mdx
gustavovalverde: I feel that if we want to make this future-proof the consideration might be to actually add `sector_identifier_uri` support in this PR (which I can certainly do).

Although from an engineer perspective what you're saying might be correct, the spec defines this `pairwiseSecret` as: **"a salt value that is kept secret by the Provider"**. So this must be treated as a Provider-level secret, not as a per-client secret. Think of it as similar consequences as rotating the `BETTER_AUTH_SECRET`; but the blast radius of the `pairwiseSecret` rotation is minimal, in comparison. The spec **intentionally** designs it this way, because a `salt` is meant to be permanent.

> Additionally, the secret cannot be returnable to the client

This is a non-issue with the current design; unless I'm missing something. The `pairwiseSecret` is a server configuration option (passed to `oauthProvider({...})`), is not a database field. It never touches the client schema, so there's no risk of it appearing in DCR responses. But If it were _per-client_, we'd actually create that leakage risk by storing it alongside client data.

gustavovalverde: @dvanmali also, important context here: multiple RPs with the same `sector_identifier_uri` (not implemented yet) must compute identical `sub` values; which would be impossible with _per-client_ secrets

dvanmali: Good catch! In that case, it would be impossible per-client and `salt` is more protected this way.

No need to add `sector_identifier_uri` unless you wish too, PR is currently complete as described.

PR #8301, opened 2026-03-02
fix(cookies): use lookahead heuristic for splitting Set-Cookie headers

bytaesu:
Related to https://discord.com/channels/1288403910284935179/1477776139027087450


bytaesu:
-> updating comment

PR #8303, opened 2026-03-02
chore: move adapter packages from deps to optional peer deps

himself65:
## Summary
- Move `@better-auth/drizzle-adapter`, `@better-auth/mongo-adapter`, and `@better-auth/prisma-adapter` from `dependencies` to optional `peerDependencies` + `devDependencies`
- These packages are only re-exported for user convenience and are not used in core runtime code
- `@better-auth/kysely-adapter` and `@better-auth/memory-adapter` remain as `dependencies` since they are used directly in runtime paths (`get-migration.ts`, `context/init.ts`, `adapter-base.ts`)

## Test plan
- [x] `pnpm install` succeeds
- [x] Verify existing tests still pass
- [x] Verify builds succeed


leon:
@himself65 this problem is back 1.5.5 includes them in dependencies

https://npmx.dev/package-code/better-auth/v/1.5.5/package.json

PR #7182, opened 2026-01-07
fix(session): skip invalid sessions in list

Paola3stefania:
## Summary by cubic
Make session listing resilient by skipping invalid or missing sessions in secondary storage. Prevents empty results when one bad session is present.

- **Bug Fixes**
  - listSessions now ignores missing, malformed, or corrupt entries instead of clearing the list.
  - findSessions now skips invalid or corrupt session data and returns only valid sessions.
  - Added tests for missing, malformed, corrupt, and all-invalid scenarios.

<sup>Written for commit fcec8e49d6659620975bf21dd7a82d386d936723. Summary will update on new commits.</sup>

PR #8328, opened 2026-03-03
fix(secondary-storage): improve JSON parse error handling

himself65:
## Summary
- Replace raw `JSON.parse` calls in `listSessions` and `findSessions` with `safeJSONParse`, so corrupt or double-parsed values no longer crash the server
- Improve the error message in `safeJSONParse` to guide users when their `secondaryStorage.get()` returns unexpected types (e.g. already-parsed objects instead of JSON strings)
- Add a test covering the common mistake of double-parsing JSON in a custom `secondaryStorage.get()` implementation

## Test plan
- [x] Existing secondary storage tests pass (string return + object return)
- [x] New double-parse scenario test passes end-to-end (sign in, get session, list sessions, revoke)
- [x] Internal adapter tests pass (33 tests including corrupt data scenarios)
- [x] Magic link secondary storage tests pass
- [x] `pnpm typecheck` passes

### DIFF
--- a/packages/better-auth/src/db/internal-adapter.ts
+++ b/packages/better-auth/src/db/internal-adapter.ts
@@ -179,24 +179,18 @@ export const createInternalAdapter = (
 					const data = await secondaryStorage.get(token);
 					if (!data) continue;
 
-					try {
-						const parsed = (
-							typeof data === "string" ? JSON.parse(data) : data
-						) as {
-							session: Session;
-							user: User;
-						};
-						if (!parsed?.session) continue;
+					const parsed = safeJSONParse<{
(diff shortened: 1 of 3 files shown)


gustavovalverde:
Closing in favor of #4076 (merged), #8248 (merged), and #7182 (merged). #8248 already ported every `listSessions` / `findSessions` call site to `safeJSONParse` with full date revival; #7182 covered skip-on-invalid for malformed entries; #4076 broadened the secondary-storage return type. This PR's only net-new addition is a verbose `logger.error` global parse log, which the pull-forward guidance explicitly avoids.

PR #8339, opened 2026-03-03
fix(captcha): breaks email-otp flow

ping-maxwell:
The default endpoints we register captcha on includes `/sign-in/email`, since our if-statement checks if the currently hit endpoint **includes** `/sign-in/email` it means that if the user hits `/sign-in/email-otp` the captcha gets triggered and thus breaks the email-otp flow.

This PR solves this by including a blocked list of endpoints, to not accidentally reach `/email-otp`.
The alternative is to change the logic from `.includes` to use something like `===` or otherwise, which would be a breaking change.
If our users still want to match email-otp, they can include it in the `endpoints` array of the plugin config and it will work.


adam-enition:
I have just stumbled across this exact issue, this would work for me. I was initially thinking maybe adding a way to specifically exclude endpoints like so:


```ts
const endpoints = options.endpoints?.length
	? options.endpoints
	: defaultEndpoints;

const includeEndpoints = endpoints.filter((e) => !e.startsWith("!"));
const excludeEndpoints = endpoints
	.filter((e) => e.startsWith("!"))
	.map((e) => e.slice(1));

const isIncluded = includeEndpoints.some((endpoint) =>
	request.url.includes(endpoint)
);

const isExcluded = excludeEndpoints.some((endpoint) =>
	request.url.includes(endpoint)
);

if (!isIncluded || isExcluded) {
	return undefined;
}
```

usage:


```ts
captcha({
  provider: 'cloudflare-turnstile',
  secretKey: env.CLOUDFLARE_TURNSTILE_SECRET,
  endpoints: ["/sign-up/email", "/sign-in/email", "!/sign-in/email-otp"]
})
```

PR #8340, opened 2026-03-03
feat(api-key): enable custom prefix

ping-maxwell:
## NOTE
This is a breaking change, by setting `enableCustomPrefix` to false by default means any existing app that upgrades that intends to change the prefix on the client will fail.

image

---

Also split the schema to it's own sub-page in docs.

CleanShot 2026-03-03 at 10 29 47@2x

closes: https://github.com/better-auth/better-auth/issues/8801

### DIFF
--- a/.changeset/pr-8340.md
+++ b/.changeset/pr-8340.md
@@ -1,5 +1,5 @@
 ---
-"better-auth": major
+"better-auth": minor
 ---
 
 Enable or disable custom prefix in API Keys


gustavovalverde:
@ping-maxwell just a warning that this changeset is set to `major` and would make the next bump version `2.0.0`, it should be `minor`

ping-maxwell:
> @ping-maxwell just a warning that this changeset is set to `major` and would make the next bump version `2.0.0`, it should be `minor`

hahah I was just working on another PR and just remembered and came back just to see you calling it out! 🤣

thanks for the heads up anyways @gustavovalverde !

Issue #7141, opened 2026-01-05
OAuth Proxy: Failed query: delete from "verification" where "verification"."id"


mimamuh:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

- Try to get the OAuth Proxy running with a **local dev environment** and a **staging environment** on a server. - All environments have their own database (nothing shared), different domains. 
- The application is a tanstack-start application. 
- Use `drizzleAdapter` with `provider: "pg"`  
- Use `advanced.database.generateId: "uuid"`
- Setup the OAuth Proxy
- Use `account.storeStateStrategy === 'cookie'` to make the proxy work at all, as without it, the OAuth Proxy plugin still does not use the 'cookie' strategy even if it should do so automatically - **but this is a different bug**.

The basic configuration of the two environments::
```sh
######## Staging build ########

BASE_URL='https://staging.example.dev' # used for `currentURL` parameter of the OAuth Proxy
PRODUCTION_URL='https://staging.example.dev' # used for `productionURL` parameter of the OAuth Proxy
AUTH_TRUSTED_ORIGINS='["https://*.example.dev", "http://localhost:3001"]'
PROJECT_ENV='staging'

AUTH_SECRET='xxxxx' # same on staging + dev
AUTH_DISCORD_ID='xxxxx' # same on staging + dev
AUTH_DISCORD_SECRET='xxxxx' # same on staging + dev

######## Local development server ########

BASE_URL='http://localhost # used for `currentURL` parameter of the OAuth Proxy
PRODUCTION_URL='https://staging.example.dev # used for `productionURL` parameter of the OAuth Proxy
AUTH_TRUSTED_ORIGINS='["http://localhost:3001"]' 
PROJECT_ENV='development'

AUTH_SECRET='xxxxx' # same on staging + dev
AUTH_DISCORD_ID='xxxxx' # same on staging + dev
AUTH_DISCORD_SECRET='xxxxx' # same on staging + dev
```

### Current vs. Expected behavior

After staring the OAuth flow with the development server, logging in into Discord (the OAuth provider in this example) and getting redirected to the staging server, where the OAuth Proxy plugin should kick it and proxy back to the local development server, it errors with the following error:
```
params: oauth-proxy-XXXXXXXXXX
    at AE.queryWithCache (file:///...22)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async Object.delete (file:///...543)
    at async Object.delete (file:///...8)
[...]

PR #8341, opened 2026-03-03
fix(oauth-proxy): add generic OAuth support to proxy

Diabl0570:
Recreated the PR as per request / suggestion
-----
Coming from the discussion in https://github.com/better-auth/better-auth/issues/7141
and the example repo https://github.com/bytaesu/ba-oauth-proxy-example
, the provided examples do not work with the generic OAuth provider.

This PR updates the oauth-proxy so it should also support generic OAuth providers.

This change still needs to be fully verified. Publishing a deployed package makes it easier to test this in a real deployed environment.

---
## Summary by cubic
Add generic OAuth support to the OAuth proxy and fix redirect URI construction so token exchange no longer fails with invalid_code. PKCE is honored and callbacks work for both /callback/:id and /oauth2/callback/:providerId across environments.

- **Bug Fixes**
  - Handle generic OAuth callbacks in the proxy and routes; use a single callbackRedirectURI for /oauth2/callback/:providerId.
  - Build redirectURI from baseURL or request origin plus the detected callback path so it matches the original authorization redirect.
  - Respect PKCE by passing codeVerifier to getToken/validateAuthorizationCode; unify callback path matching with a small helper.

<sup>Written for commit a6cc0e4603ffda929b46347ed40476ce16219a63. Summary will update on new commits.</sup>


zarif-lh:
Hi!

We are using better auth in our project and we are working with an unsupported OAuth provider so we had to setup the generic OAuth plugin. 

We ran into the same issue you are fixing in this PR and took inspiration from your PR to run a patched version of better auth in our project (until this PR gets merged).

We have run into some additional problems with generic OAuth and the proxy feature. I've used AI to summarize the issues we faced in an easy to read format.

---

### Bug 1: `additionalFields` values are dropped in the proxy flow (new users)

**What's broken:** if you declare `user.additionalFields` on the auth instance and return values for them from `mapProfileToUser`, those values never make it to the database when the OAuth proxy is active. They work fine without the proxy.

**Why:** in `oauth-proxy/index.ts`, the callback before-hook copies user info into the encrypted `payload` with a hardcoded field list:

```ts
const payload = {
  userInfo: {
    id: String(userInfo.id),
    email: userInfo.email,
    name: userInfo.name || "",
    image: userInfo.image,
    emailVerified: userInfo.emailVerified,
  },
  // ...
};
```

Everything `mapProfileToUser` returned beyond these five keys is silently discarded at this boundary. The non-proxy handler in `generic-oauth/routes.ts` doesn't do this rewrite, so direct callbacks work.

**Reproduction:**
1. Use a genericOAuth provider whose `mapProfileToUser` returns a custom field (e.g. `adapterId: '123'`).
2. Declare `additionalFields: { adapterId: { type: 'string' } }` on the auth instance.
3. Configure `oAuthProxy` so the flow is actually exercised (different `productionURL` vs `currentURL`).
4. Sign in and inspect the user row. `adapterId` is missing.
5. Remove the proxy plugin and sign in again. `adapterId` is now persisted.

### Bug 2: `overrideUserInfoOnSignIn` is ignored in the proxy flow (returning users)

**What's broken:** setting `overrideUserInfoOnSignIn: true` on a provider config tells better-auth to refresh an existing user's profile fields on every sign-in. This works without the proxy, but the flag has no effect once the proxy is active — the user record never updates on subsequent logins.

**Why:** the proxy's `/oauth-proxy-callback` endpoint calls `handleOAuthUserInfo({ userInfo, account, callbackURL, disableSignUp })` without passing `overrideUserInfo`. The flag is also not included when the payload is built earlier. So even though `handleOAuthUserInfo` supports the flag (and uses it to branch into `updateUser` for existing accounts), it's always `undefined` in the proxy flow. The non-proxy handler in `generic-oauth/routes.ts` passes `overrideUserInfo: providerConfig.overrideUserInfo` correctly — looks like a simple parity miss.

**Reproduction:**
1. Configure any provider with `overrideUserInfoOnSignIn: true` behind the proxy.
2. Sign in once to create the user row.
3. Change a profile field on the provider side (e.g. change `name`).
4. Sign in again via the proxy. The DB row still shows the original name.
5. Drop the proxy and repeat — the row updates as expected.

---

Deferring on the fix shape since you know the library internals far better than we do. Let me know if you'd like these moved to standalone issues or have a preferred direction.

gustavovalverde:
Closing in favor of #1289 (merged), #9069 (merged), and #7720 (merged). #9069 deleted the stale `/sign-in/oauth2` and `/oauth2/callback/:providerId` routes this PR patches; #1289 already fixed `redirectURI` consistency in the surviving code path. Verify the custom `getToken` PKCE propagation here shipped in #9069's rewrite before fully discarding.

PR #8351, opened 2026-03-04
feat(account): support server-side accountInfo calls without session headers

### LINKED ISSUE #8350: Support server-side accountInfo calls without session headers (follow getAccessToken pattern)

I want to get accountInfo server side so my AI tools + background jobs / agents can use it but currently have no way of doing so.

getAccessToken already supports server-side calls without a session by accepting an optional userId body parameter and only enforcing the session when an HTTP request is present

from docs:
```
await auth.api.getAccessToken({
  body: {
    providerId: "google", // or any other provider id
    accountId: "accountId", // optional, if you want to get the access token for a specific account
    userId: "userId", // optional, if you don't provide headers with authenticated token
  },
  headers: await headers() // headers containing the user's session token
});
```

However, accountInfo uses sessionMiddleware which hard-blocks all calls without a valid session. This means server-side code cannot call accountInfo without reconstructing session headers:

```
// This fails — no way to pass userId, sessionMiddleware rejects it
await auth.api.accountInfo({
  query: { accountId: "..." },
  // no headers available in this context
});
```

Solution: make accountInfo follow same pattern as getAccessToken. Will still be fully backwards-compat.

I can open a PR for this today.

- Passing fake headers: Requires constructing a session token just to satisfy middleware
- Custom endpoint wrapper: What we currently do: a separate function that manually calls getAccessToken (which does support server-side) and then directly calls the provider's userinfo API. 

### Additional context

_No response_

NathanColosimo:
## Summary

- Remove `sessionMiddleware` from `accountInfo` and add manual session resolution following the same pattern as `getAccessToken`
- Add optional `userId` query parameter so server-side callers can use `auth.api.accountInfo()` without needing to construct session headers
- Update docs to show the new `userId` parameter in the server-side example
- Fully backwards compatible
- Tested patching node_modules locally and it worked

Closes #8350 

---
## Summary by cubic
Enable server-side accountInfo calls without session headers by adding an optional userId and aligning auth with getAccessToken. Calls can use a session token or userId and remain backward compatible; docs now show a server-side example.

- **New Features**
  - Added `query.userId` for server-side calls without session headers.
  - Replaced session middleware with manual session resolution; require session or userId and validate the account against the resolved user.

<sup>Written for commit 5617cd4606f747fa6f55cbf39562ba5f94d22771. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/eager-cloths-stay.md
+++ b/.changeset/eager-cloths-stay.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+support server-side accountInfo calls without session headers
--- a/docs/content/docs/concepts/oauth.mdx
+++ b/docs/content/docs/concepts/oauth.mdx
@@ -108,7 +108,10 @@ server-side usage:
 
 ```ts
 await auth.api.accountInfo({
-  query: { accountId: "accountId" },
+  query: {
+    accountId: "accountId",
+    userId: "userId", // optional, if you don't provide headers with authenticated token
+  },
   headers: await headers() // headers containing the user's session token
 });
 ```
--- a/packages/better-auth/src/api/routes/account.ts
+++ b/packages/better-auth/src/api/routes/account.ts
@@ -820,6 +820,13 @@ const accountInfoQuerySchema = z.optional(
 					"The provider given account id for which to get the account info",
 			})
 			.optional(),
+		userId: z
+			.string()
+			.meta({
+				description:
+					"The user ID associated with the account. Optional — only needed for server-side calls without session headers.",
+			})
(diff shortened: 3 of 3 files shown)


gustavovalverde:
Superseded by #9813, rebased on current `main` with a shared token-resolution refactor that also fixes the server-side path: the nested `getAccessToken` call here dropped the resolved `userId`, so `accountInfo` with `userId` and no session returned `UNAUTHORIZED`. Your `userId` approach is carried over and you're co-authored on the new commit. Thanks @NathanColosimo; closing in favor of the new PR.

PR #8364, opened 2026-03-04
fix(sso): allow bare domains for SSO domain verification 

### LINKED ISSUE #8361: SSO domain verification fails when domain is domain and not URL

### To Reproduce

1. `registerSSOProvider` with `domain: <any domain>`
2. `requestDomainVerification` fails with `ERR_INVALID_URL`

`domain` should execute normally with a valid domain

main

### System info

```bash
...
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

The raw domain e.g. `github.com` is put into `new URL` which throws because a domain is not a URL.

https://github.com/better-auth/better-auth/blob/bd980f8c5b2ad9ff13a78d27ecbe27fd92ed879f/packages/sso/src/routes/domain-verification.ts#L245

The tests have a URL in the domain field and therefore don't catch it.

https://github.com/better-auth/better-auth/blob/bd980f8c5b2ad9ff13a78d27ecbe27fd92ed879f/packages/sso/src/domain-verification.test.ts#L130

GautamBytes:
Resolves #8361

### Description
The SSO `requestDomainVerification` and `verifyDomain` endpoints previously threw an `ERR_INVALID_URL` if the SSO provider was registered with a bare domain (e.g., `github.com`) instead of a full URL (e.g., `https://github.com`). This mismatch occurred because the native `URL` constructor requires a protocol scheme.

This PR introduces an internal normalization helper (`getHostnameFromDomainInput`) that prepends a secure scheme (`https://`) to bare domains before parsing them. 

### Changes
- Replaced direct `new URL()` calls with a normalization function in `packages/sso/src/routes/domain-verification.ts`.
- Updated test fixtures to use a bare domain (`hello.com`) as the default to prevent future regressions.
- Added a dedicated test case to ensure full URLs (`http://hello.com:8081`) continue to be parsed correctly, preserving backward compatibility.

---
## Summary by cubic
Fix SSO domain verification to accept bare domains (e.g., github.com) without throwing ERR_INVALID_URL. We normalize inputs by adding https when missing and parse the hostname, keeping full URLs supported.

- **Bug Fixes**
  - Added getHostnameFromDomainInput to normalize domain inputs and safely extract the hostname.
  - Updated tests to default to a bare domain and added a case for full URLs to prevent regressions.

<sup>Written for commit 4f29c609bd6274f2ae6d7d7e5a07363e9f2dc875. Summary will update on new commits.</sup>

### DIFF
--- a/packages/sso/src/domain-verification.test.ts
+++ b/packages/sso/src/domain-verification.test.ts
@@ -122,12 +122,13 @@ describe("Domain verification", async () => {
 		async function registerSSOProvider(
 			headers: Headers,
 			organizationId?: string,
+			domain = "hello.com",
 		) {
 			return auth.api.registerSSOProvider({
 				body: {
 					providerId: "saml-provider-1",
 					issuer: "http://hello.com:8081",
-					domain: "http://hello.com:8081",
+					domain,
 					samlConfig: {
 						entryPoint: "http://idp.com:",
 						cert: "the-cert",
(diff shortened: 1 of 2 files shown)


himself65 (commented):
I think maybe we have a similar URL constructor issue across the repo. Let me see if we can use a library to handle that

himself65:
How about https://github.com/better-auth/better-auth/pull/8369?

GautamBytes:
> How about #8369?

Looks great to me 🙌

GautamBytes:
closing in support of #8369

PR #8373, opened 2026-03-04
docs: add warning to SQLite documentation for bun:sqlite

simonxabris:
I ran into the issue with `bun:sqlite` that if you dont use `bunx --bun` to run the cli commands, the cli fails because it does not recognize the `bun:sqlite` module. This is quite frustrating to figure out so I thought adding it to the docs makes sense.

---
## Summary by cubic
Add a warning callout in the SQLite adapter docs telling Bun users to run CLI commands with bunx --bun (e.g., bunx --bun auth@latest generate) to avoid bun:sqlite type errors.

<sup>Written for commit d471e2f2977c7e7004502a6fc6df1d2cbe8824f5. Summary will update on new commits.</sup>

### DIFF
--- a/docs/content/docs/adapters/sqlite.mdx
+++ b/docs/content/docs/adapters/sqlite.mdx
@@ -52,6 +52,10 @@ node your-app.js
 
 ### Bun Built-in SQLite
 
+<Callout type="warning">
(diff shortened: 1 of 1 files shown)


himself65 (approved):
Looks like related to https://github.com/oven-sh/bun/issues/6738

PR #8395, opened 2026-03-04
docs: add background tasks

jonathansamines:


jonathansamines:
> this is good, another thing I would recommend is adding information about this in the "optimize for performance" guide if you think it makes sense

Definitely, adding it now!

PR #8405, opened 2026-03-05
fix(expo): handle origin override across mutable and immutable requests

### LINKED ISSUE #8404: @better-auth/expo server plugin breaks route matching on Bun due to new Request() cloning

### To Reproduce

Set up Better Auth with the expo() server plugin on a Bun-based server (e.g., Next.js running on Bun)
Send a POST to /api/auth/sign-in/social without an Origin header but with expo-origin: myapp:// (this is exactly what the @better-auth/expo client sends from React Native)
The endpoint returns 404 with an empty body

Current behavior:

The expo server plugin's onRequest hook creates a new Request to copy expo-origin to origin:
```const newHeaders = new Headers(request.headers);
newHeaders.set("origin", expoOrigin);
return { request: new Request(request, { headers: newHeaders }) };```

On Bun, new Request(existingRequest, { headers }) doesn't correctly preserve the URL/body when the original Request comes from Bun.serve(). 

The resulting Request fails route matching in better-call's router, returning a bare 404. 

This affects all Expo mobile social sign-in on Bun-based servers. 

Web sign-in is unaffected because the browser sets the Origin header natively, so the plugin skips its onRequest hook.



1.5.3

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.3.0: Wed Jan 28 20:53:05 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T6020",
    "release": "25.3.0",
    "cpuCount": 12,
    "cpuModel": "Apple M2 Pro",
    "totalMemory": "16.00 GB",
    "freeMemory": "0.15 GB"
  },
  "node": {
    "version": "v25.7.0",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.10"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null
  }
}
```

Client, Package, Backend

```typescript
  baseURL: env.NEXT_PUBLIC_WEB_URL,
  advanced: {
    cookiePrefix: "kompose",
  },
    provider: "pg",
    schema,
  }),
  /** Redis-backed storage for sessions and rate limit counters. */
  secondaryStorage: redisSecondaryStorage,
  /** Rate limiting for auth endpoints (sign-in, token refresh, etc.). */
  rateLimit: {
    enabled: true,
    window: 60,
    max: 100,
    storage: "secondary-storage",
  },
  trustedOrigins: [
    env.NEXT_PUBLIC_WEB_URL,
    "kompose://",
    "exp://",
    "http://localhost:3000",
    "tauri://localhost",
    "https://appleid.apple.com",
  ],
  emailAndPassword: {
    enabled: false,
  },
  account: {
    accountLinking: {
      enabled: true,
      // Support linking additional Google accounts that may use different emails.
      allowDifferentEmails: true,
      trustedProviders: ["google", "apple"],
    },
  },
  socialProviders: {
    google: {
      prompt: "select_account consent",
      accessType: "offline",
      clientId: env.GOOGLE_CLIENT_ID,
      clientSecret: env.GOOGLE_CLIENT_SECRET,
      scope: ["https://www.googleapis.com/auth/calendar"],
    },
    apple: {
      clientId: env.APPLE_CLIENT_ID,
      clientSecret: env.APPLE_CLIENT_SECRET,
      appBundleIdentifier: env.APPLE_APP_BUNDLE_IDENTIFIER,
    },
  },
  logger: {
    level: "warn",
  },
  plugins: [
    expo(),
    nextCookies(),
    lastLoginMethod({
      cookieName: "kompose.last_used_login_method",
      storeInDatabase: false,
    }),
    // Bearer token auth for Tauri desktop. The Tauri webview cannot use
    // cookies cross-origin (WKWebView ITP blocks Set-Cookie), so it
    // authenticates via Authorization header instead.
    bearer({ requireSignature: true }),
    // One-time tokens for cross-context auth (Tauri deep-link OAuth flow).
    oneTimeToken({
      storeToken: "hashed",
    }),
  ],
});
```

## Suggested fix in packages/expo/src/index.ts — lines 47-54

// Before (broken on Bun):
```const newHeaders = new Headers(request.headers);
newHeaders.set("origin", expoOrigin);
return { request: new Request(request, { headers: newHeaders }) };```

// After:
```request.headers.set("origin", expoOrigin);
return { request: request };
```

## This fixed it for me locally:
```import { auth } from "@kompose/auth";
import { toNextJsHandler } from "better-auth/next-js";

const handlers = toNextJsHandler(auth.handler);

export const GET = handlers.GET;
export function POST(request: Request) {
  // Bun's `new Request(req, { headers })` breaks route matching inside
  // better-call's router. The @better-auth/expo server plugin uses that
  // pattern to copy expo-origin → origin, which triggers the bug.
  // Workaround: mutate the header in-place so the expo plugin sees origin
  // already set and skips its Request cloning.
  const expoOrigin = request.headers.get("expo-origin");
  if (!request.headers.get("origin") && expoOrigin) {
    request.headers.set("origin", expoOrigin);
  }
  return handlers.POST(request);
}
```

NathanColosimo:
Closes #8404

Tested patching node_modules locally as well in my project and this fixed it.

Essentially, new Request() breaks with bun sometimes, and this changes the already-existing expo fix that copied headers over to a new Request instead to mutate the headers in place to prevent breaking with Bun runtimes.

---
## Summary by cubic
Update Expo middleware to mutate the origin header directly on the incoming Request to avoid Bun route matching issues. If header mutation isn’t allowed (e.g., Cloudflare Workers), fall back to creating a new Request with updated headers; regression tests confirm we preserve the original Request when mutation works and clone when it fails.

<sup>Written for commit b87fda3e5398ea3bf9ee8da9a7479abec87c1cf5. Summary will update on new commits.</sup>

### DIFF
--- a/.cspell/custom-words.txt
+++ b/.cspell/custom-words.txt
@@ -13,3 +13,5 @@ prefs
 uncompromised
 myapp
 Neue
+CCPA
+CPRA
--- a/demo/electron/pnpm-lock.yaml
+++ b/demo/electron/pnpm-lock.yaml
@@ -1466,6 +1466,10 @@ packages:
     resolution: {integrity: sha512-VWSRii4t0AFm6ixFFmLLx1t7wS1gh+ckoa84aOeapGum0h+EZd1EhEumSB+ZdDLnEPuucsVB9oB7cxJHap6Afg==}
     engines: {node: '>=14.14'}
 
+  fs-extra@11.3.4:
+    resolution: {integrity: sha512-CTXd6rk/M3/ULNQj8FBqBWHYBVYybQ3VPBw0xGKFe3tuH7ytT6ACnvzpIQ3UZtB8yvUKC2cXn1a+x+5EVQLovA==}
+    engines: {node: '>=14.14'}
+
   fs-extra@7.0.1:
     resolution: {integrity: sha512-YJDaCJZEnBmcbw13fvdAM9AwNOJwOzrE4pqMqBq5nFiEqXUqHwlK4B+3pUw6JNvfSPtX05xFHtYy/1ni01eGCw==}
     engines: {node: '>=6 <7 || >=8'}
@@ -2640,7 +2644,7 @@ snapshots:
     dependencies:
       cross-dirname: 0.1.0
       debug: 4.4.3
-      fs-extra: 11.3.3
+      fs-extra: 11.3.4
       minimist: 1.2.8
       postject: 1.0.0-alpha.6
     transitivePeerDependencies:
@@ -3808,6 +3812,13 @@ snapshots:
       graceful-fs: 4.2.11
       jsonfile: 6.2.0
       universalify: 2.0.1
+
+  fs-extra@11.3.4:
+    dependencies:
+      graceful-fs: 4.2.11
+      jsonfile: 6.2.0
+      universalify: 2.0.1
+    optional: true
 
   fs-extra@7.0.1:
     dependencies:
--- a/demo/nextjs/pnpm-lock.yaml
+++ b/demo/nextjs/pnpm-lock.yaml
@@ -263,15 +263,19 @@ packages:
     resolution: {integrity: sha512-cTlrKttbrRHEw3W+0/I609A2Matj5JQaRvfLtEIGZvlN0RaPi+3ANsMeqAyCAVlH/lUIW2tmtBlSMni74lcXeg==}
     engines: {node: '>=12'}
 
-  '@better-auth/core@1.5.0-beta.20':
-    resolution: {integrity: sha512-6RmDZ6kU85u1BR/H6OPjV3Z9LrfnCXKVEXFCjOUy0lgNOLRD2QOL59fgu2rpGc85aDxjDi2ddRPrSfvySGnybg==}
+  '@better-auth/core@1.5.1-beta.3':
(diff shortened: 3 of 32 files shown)


bytaesu:
Hi @NathanColosimo,

I tested on Cloudflare Workers and updated it to the current code. Do you have alternative approach to address this? This PR seems likely to cause a regression on Cloudflare Workers 🤔

- https://github.com/better-auth/better-auth/pull/7774

NathanColosimo:
hmm ok I'll see

NathanColosimo:
Ok, since mutating throws on cloudflare workers, but creating a new request has issues in Bun, I'll update the code to attempt a mutate, and if that fails, attempt a clone. If both fails then the overall request just fails.

bytaesu:
Hi @NathanColosimo,

I tested, this PR version works correctly in `Next.js (Bun)`, `Hono (Bun)`, and `Hono (Cloudflare Workers - prod)`

Interestingly, the previous version worked fine in Hono (Bun), but I was able to reproduce the issue in Next.js (Bun) 🧐

bytaesu:
I'm adding regression test case

Issue #8415, opened 2026-03-05
Mixed ID types in compatible with numeric user ids


OscarCornish:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Have a user table with numeric ids
2. Create rest of tables with UUIDs

generateId:

```
generateId({ model }) {
  if (model === "user") {
    return false;
  }
  return uuidv7();
},
```

This works for creating tables, but when searching them it will throw at the adapter because "userId column expected type Int" (or similar)

### Current vs. Expected behavior

I want to be able to declare for a table:
- Let the db generate your id values
- Those id values will be numeric

But if generateIds is false, then the check for numericIds (generateId === "serial") will not be true

### What version of Better Auth are you using?

1.5.1

### System info

```bash
-
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
  advanced: {
    database: {
      generateId({ model }) {
        if (model === "user") {
          return false;
        }
        return uuidv7();
      },
    },
  },
});
```

PR #8416, opened 2026-03-05
feat(core): Support both mixed and numeric ids

OscarCornish:
Adds advanced.database.useNumberId option to control which models use numeric (integer) IDs on a per-model basis.

  Previously, generateId: "serial" forced all tables to use integer IDs. This was a problem for databases where only some tables use auto-increment integers (e.g. user.id is an integer) while others use
  UUIDs. The string-to-number type coercion was globally tied to generateId === "serial", so there was no way to have mixed ID types.

  Now you can do:

  advanced: {
    database: {
      generateId: ({ model }) => model === "user" ? false : crypto.randomUUID(),
      useNumberId: ["user"],
    },
  }

  This makes only the user table use integer IDs while all other tables use UUIDs — and FK columns like session.userId are correctly coerced to integers when they reference a numeric-ID model.

  Accepts true, false, string[], or (model: string) => boolean. When not set, falls back to legacy behavior (generateId: "serial" applies to all models).
  

---
## Summary by cubic
Adds per-model control for numeric IDs with advanced.database.useNumberId, enabling mixed integer and string/UUID IDs across tables. Updates migrations, adapters, and schema generators to type IDs and FKs per model and coerce FK values correctly.

- **New Features**
  - Added advanced.database.useNumberId (true/false, string[], or (model) => boolean). When unset, legacy generateId: "serial" applies to all models.
  - Migrations choose id/foreignKey column types per model for Postgres, MySQL, MSSQL, and SQLite.
  - Adapters coerce FK values to numbers only when referencing numeric-ID models; output types match the model’s ID type.
  - Drizzle and Prisma generators emit correct id/FK types and only import integer helpers when needed.
  - Telemetry now includes useNumberId.

- **Migration**
  - Optional: set advanced.database.useNumberId to pick which models use integer IDs (e.g., ["user"]).

<sup>Written for commit afe3c7944b554360635bb6c98ed6d1b5a10ebf9f. Summary will update on new commits.</sup>


OscarCornish:
Resolves #8415

packages/core/src/types/init-options.ts
```diff
@@ -330,6 +330,29 @@ export type BetterAuthAdvancedOptions = {
...
+				 * ```ts
+				 * advanced: {
+				 *   database: {
+				 *     generateId: ({ model }) => model === "user" ? false : crypto.randomUUID(),
+				 *     useNumberId: ["user"],
+				 *   },
+				 * }
+				 * ```
+				 */
+				useNumberId?: boolean | string[] | ((model: string) => boolean);
```
himself65, line 355: we have removed the useNumberId

himself65: I think maybe a map is better here

ping-maxwell:
closing as stale

PR #8436, opened 2026-03-06
chore(api-key): clean up api-key code

ping-maxwell:


himself65 (commented):
the `APIError.from()` reformatting (extracting error codes into `const msg` variables) is unnecessary churn — the original multi-line format was already clean and idiomatic. Consider dropping those hunks to keep the diff focused on the meaningful changes.

PR #8437, opened 2026-03-06
fix(client): preserve stale session data on network or server errors

### LINKED ISSUE #8420: Session is set to null on network reconnect (mobile)

### To Reproduce

1. Authenticate successfully on a mobile device (React Native / Expo)
2. Turn off internet (WiFi + cellular)
3. Navigate around the app — session is intact, useSession still returns data ✓
4. Turn internet back on
5. → useSession immediately sets data: null → app redirects to login screen ✗

Bonus step to confirm the token is still valid:
6. Force-quit and reopen the app → session is automatically restored ✓
   (proves the token in SecureStore was never invalidated server-side)

**Current behavior:**

When a mobile device loses internet and reconnects, useSession sets data: null and triggers a logout — even though the session token in SecureStore is still perfectly valid.

**Expected behavior:**

A transient network failure during session revalidation should preserve the previously cached session data. Only an explicit 401 Unauthorized response from the server should invalidate the session and trigger a logout.

1.4.16

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.3.0: Wed Jan 28 20:49:24 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T8132",
    "release": "25.3.0",
    "cpuCount": 10,
    "cpuModel": "Apple M4",
    "totalMemory": "24.00 GB",
    "freeMemory": "0.42 GB"
  },
  "node": {
    "version": "v22.22.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.4"
  },
  "frameworks": [
    {
      "name": "react",
      "version": "19.1.1"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "catalog:",
    "config": null
  }
}
```

Client

```typescript
import { expoClient } from "@better-auth/expo/client";
import {
	adminClient,
	inferAdditionalFields,
	oneTimeTokenClient,
	usernameClient,
	emailOTPClient,
	phoneNumberClient,
	twoFactorClient,
	anonymousClient,
} from "better-auth/client/plugins";

	baseURL: process.env.EXPO_PUBLIC_BASE_URL,
	disableDefaultFetchPlugins: true,
	plugins: [
		expoClient({
			scheme: "test",
			storagePrefix: "test",
			storage: SecureStore,
		}),
		anonymousClient(),
		usernameClient(),
		emailOTPClient(),
		phoneNumberClient(),
		oneTimeTokenClient(),
		twoFactorClient(),
	],
});
```

### Root cause

After digging into the source, I found the issue in `node_modules/better-auth/dist/client/query.mjs`:

```js
async onError(context) {
  const retryAttempts = typeof request.retry === "number" ? request.retry : request.retry?.attempts;
  const retryAttempt = request.retryAttempt || 0;

  if (retryAttempts && retryAttempt < retryAttempts) return;

  // No network error check — data is unconditionally set to null on any failure
  value.set({
    error: context.error,
    data: null, // <-- session wiped on network error
    isPending: false,
    isRefetching: false,
    refetch: value.value.refetch
  });

  await opts?.onError?.(context);
},
```

bytaesu:
> [!NOTE]
> Inspired by the stale-while-revalidate pattern (TanStack Query)
> https://github.com/TanStack/query/blob/main/packages/query-core/src/query.ts#L650-L664

only clear session data on 401 and preserve stale data on non-401 errors.

- Closes #8420

---

### Before

https://github.com/user-attachments/assets/3d0a3dc0-6ab8-45d2-b556-01c214012fa2

### After

https://github.com/user-attachments/assets/39da7b3a-5495-4cc6-b111-79c1ad1ff1d1

### DIFF
--- a/.cspell/auth-terms.txt
+++ b/.cspell/auth-terms.txt
@@ -30,3 +30,4 @@ Agentic
 Timestampz
 Vercel
 rgba
+idtoken
--- a/docs/content/docs/concepts/hooks.mdx
+++ b/docs/content/docs/concepts/hooks.mdx
@@ -250,6 +250,52 @@ This may be useful to use instead of using your db directly to get access to `da
 
 You can use `ctx.context.generateId` to generate Id for various reasons.
 
+#### runInBackground
+
+Schedules a task to run after the response is sent. Use for fire-and-forget operations (cleanup, analytics, rate limit counter updates). Configure the handler in [advanced.backgroundTasks](/docs/reference/options#backgroundtasks).
+
+```ts title="auth.ts"
+import { betterAuth } from "better-auth";
+import { createAuthMiddleware } from "better-auth/api";
+
+export const auth = betterAuth({
+  hooks: {
+    after: createAuthMiddleware(async (ctx) => {
+      if (ctx.path.startsWith("/sign-up")) {
+        const newSession = ctx.context.newSession;
+        if (newSession) {
+          ctx.context.runInBackground(sendAnalyticsEvent(newSession.user.id));
+        }
+      }
+    }),
+  },
+});
+```
+
+#### runInBackgroundOrAwait
+
+Defers the task when a handler is configured, otherwise awaits it. Use for operations that must complete (e.g. sending emails) but benefit from not blocking when a handler exists. Configure the handler in [advanced.backgroundTasks](/docs/reference/options#backgroundtasks).
+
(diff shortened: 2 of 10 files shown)


bytaesu:
-> Let me test on Expo app

PR #8455, opened 2026-03-06
fix: use fallback URL as initial baseURL for dynamic config

### LINKED ISSUE #8447: BaseURL seems to be empty when using dynamic baseURL
### To Reproduce

1. Create Better-Auth Backend
2. Try to integrated oauth authentication provider
3. Open the .well-known/openid-configuration endpoint

`baseURL` is currently empty, while it should be set to `{domain}/{basePath}`.
This causes OAuth consumers to fail, because the URL is not complete.

1.5.4

### System info

```json
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.3.0: Wed Jan 28 20:54:55 PST 2026; root:xnu-12377.91.3~2/RELEASE_ARM64_T6031",
    "release": "25.3.0",
    "cpuCount": 16,
    "cpuModel": "Apple M3 Max",
    "totalMemory": "64.00 GB",
    "freeMemory": "4.03 GB"
  },
  "node": {
    "version": "v25.2.1",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.2"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null,
    "error": "ENOENT: no such file or directory, open 'package.json'"
  }
}
```

Backend, Package

```typescript
  baseURL: {
    allowedHosts: [
      'localhost:*',
    ]
    fallback: 'http://localhost:5173',
  }
});
```

### Additional context

BaseURL empty is in this line (and probably other ones):
https://github.com/better-auth/better-auth/blob/184cf31e765bbacbc3f6180c5455c29803c7d5d9/packages/oauth-provider/src/metadata.ts#L23-L27

The baseURL is set here to an empty string:
https://github.com/better-auth/better-auth/blob/184cf31e765bbacbc3f6180c5455c29803c7d5d9/packages/better-auth/src/context/create-context.ts#L253-L255

Because it is set to undefined here:
https://github.com/better-auth/better-auth/blob/184cf31e765bbacbc3f6180c5455c29803c7d5d9/packages/better-auth/src/context/create-context.ts#L129-L134

himself65:
## Summary

Fixes #8447

- When using dynamic `baseURL` config (with `allowedHosts`), the initial `AuthContext.baseURL` was set to `""` (empty string)
- While per-request resolution in the handler correctly resolved the baseURL, any code accessing `ctx.context.baseURL` outside a request context (e.g. `auth.api.getOpenIdConfig()` or `.well-known/openid-configuration`) got an empty string, producing broken URLs
- Now uses the `fallback` URL (or env vars) to populate the initial baseURL so non-request contexts have a usable value

## Test plan

- [x] Added regression test verifying `ctx.baseURL` uses the fallback URL when dynamic config is used
- [x] All existing dynamic baseURL tests pass
- [x] oauth-provider metadata tests pass
- [x] `pnpm lint`, `pnpm typecheck` pass

### DIFF
--- a/packages/better-auth/src/api/to-auth-endpoints.ts
+++ b/packages/better-auth/src/api/to-auth-endpoints.ts
@@ -23,6 +23,11 @@ import type {
 import { kAPIErrorHeaderSymbol, toResponse } from "better-call";
 import { createDefu } from "defu";
 import { isAPIError } from "../utils/is-api-error";
+import {
+	getOrigin,
+	isDynamicBaseURLConfig,
+	resolveBaseURL,
+} from "../utils/url";
 
 type InternalContext = Partial<
 	InputContext<string, any, any, any, any, any> &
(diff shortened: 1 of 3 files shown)


gustavovalverde:
Closing in favor of #9131 (merged). #9131 closes #8447 (the same upstream issue this PR targets) with a broader `resolveRequestContext` rewrite that handles no-source/no-fallback paths, metadata helper forwarding, proxy header gating, and loopback protocol inference.

PR #8463, opened 2026-03-06
feat(sso): add additionalData support to SSO sign in

### LINKED ISSUE #8450: Pipe additional client-provided data through SSO endpoints/callbacks/plugins

When provisioning users from SSO authentication, I need additional data from the client (not just the IdP) in order to provision them correctly. This could be data like timezone, app settings, etc. The `provisionUser` plugin method receives `user`, `userInfo`, `token`, `provider`, but as far as I can tell, there is no way to pipe additional data from the client to this method. Additionally, the hook context for the callback (PATH=/sso/saml2/callback/:providerId), which could hypothetically be another appropriate place to do this, does not work with `getOauthState` like it does with the OAuth callback (PATH=/callback/:id) when `additionalData` is provided from the client API call.

Please let me know if there is an existing workaround for this or if this should already be functional in 1.5

Please implement a way to pipe additional data from the client SSO auth call (`client.signIn.sso`) to either the `provisionUser` plugin method or the callback hooks.

N/A

### Additional context

Would ideally like to be able to call 

```
 await client.signIn.sso({
		email: email,
		callbackURL: url,
		additionalData: {
			timezone: tz,
			example2: ex,
		},
	});
```

and have in plugins: 

```
sso({
	provisionUser: async ({user, userInfo, token, provider, additionalData}) => {
          ...
	}
}),
```

OR have in hooks: 
```	
const additionalData = await getOAuthState<{timezone, example2}>();
```
work with the `/sso/saml2/callback/:providerId` path

rae-fcm:
closes #8450 

RelayState (SAML) and the OAuth state parameter (OIDC) are already used internally to carry callback URLs and other state parameters through the IdP back to the SP round-trip. `additionalData` exposes the same mechanism to consumers in its most general sense, allowing arbitrary key/value data to be sent with the sign-in request, echoed back by the IdP, and delivered to the provisionUser callback and sso callback hooks without any extra round-trips or storage.

---
## Summary by cubic
Adds client-supplied `additionalData` to SSO sign-in and carries it through OIDC and SAML (including ACS) to `provisionUser` and server hooks via `getSSOState`. Re-fetches the user after provisioning so org role hooks see the updated user.

- **New Features**
  - `signIn.sso` and server `signInSSO` accept `additionalData`; it’s encoded in OAuth state and SAML RelayState and delivered to `provisionUser`.
  - Exported `getSSOState` from `@better-auth/sso` to read request-scoped SSO state (e.g., `callbackURL`, `expiresAt`, `codeVerifier`) plus `additionalData` during OIDC/SAML callbacks.

- **Bug Fixes**
  - Re-fetch the user after `provisionUser` and pass the updated user to org provisioning to prevent stale role assignment (OIDC and SAML, including ACS).

<sup>Written for commit 80922f01a70e9a8139e8d4f47fbd0d319f19396f. Summary will update on new commits.</sup>


packages/sso/src/sso-state.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Available in server-side hooks during both OIDC and SAML SSO callbacks.
+ * Returns `undefined` outside of an SSO callback request.
+ *
+ * @example
+ * const data = await getSSOState<{ referralCode?: string }>();
+ */
+async function getSSOState<
+	T extends Record<string, unknown> = Record<string, unknown>,
+>(): Promise<T | undefined> {
+	return _getSSOState() as Promise<T | undefined>;
```
himself65: I think the type here is incorrect

himself65: maybe you can pass a zod type

rae-fcm: Agreed. This was a little convoluted to begin with. Rewrote to be more parallel with OAuth state flow, which exposes the full state for client discretion to use or not use, and additionalData is just included in that. Extends zod type from `stateDataSchema`. Updated tests and docs in https://github.com/better-auth/better-auth/pull/8463/commits/3f1cbd84ba347445c8ab7453c47b08bc5e860eaa

PR #8465, opened 2026-03-06
chore(better-auth): migrate build from tsdown to rollup

himself65:
## Summary

- Replace `tsdown` with `rollup` + `rollup-plugin-swc3` for the `better-auth` package build
- Use `tsc --emitDeclarationOnly` with a dedicated `tsconfig.build.json` for type declarations
- Add post-build script (`scripts/rename-dts.mjs`) to rename `.d.ts` → `.d.mts`
- Fix type inference in `createAuthEndpoint` by using `const Method` type parameter and replacing `any` with proper types (`Method`, `Use`)
- Minor type fixes in `path-to-object.ts`, `anonymous/index.ts`, and `custom-session/index.ts`

## Test plan

- [ ] Verify `pnpm build` succeeds in `packages/better-auth`
- [ ] Verify generated output in `dist/` has correct `.mjs` and `.d.mts` files
- [ ] Run `pnpm typecheck` to ensure no type regressions
- [ ] Run existing tests to confirm no runtime breakage


Xhale1:
Apologies if this is the wrong place to ask. Why is better-auth migrating from tsdown to rollup? With vite migrating from rollup+esbuild to rolldown, it seems curious to go in what feels like the other direction.

I'm asking to satisfy my own curiosity as someone investigating adopting tsdown.

himself65:
> Apologies if this is the wrong place to ask. Why is better-auth migrating from tsdown to rollup? With vite migrating from rollup+esbuild to rolldown, it seems curious to go in what feels like the other direction.
> 
> I'm asking to satisfy my own curiosity as someone investigating adopting tsdown.

This is just a experimental thing. Because I have seen some d.ts issues that tsdown has more complex type definitions output comparison with rollup + drs plugins.

But I cannot 100% tell whether this is from tsdown or rolldown, or some configuration difference.

and maybe this will not be merged since the better solution is we remove third party types from out code

PR #8475, opened 2026-03-07
fix: update-session should allow `input:false` on server

ping-maxwell:
This PR allows the update-session endpoint to update fields with `input: false` to be updated on the server only. The client will remain prohibited from updating fields with `input: false`

Additionally, it fixes a bug in the admin plugin which allows `impersonatedBy` to be updated. As of now, they cannot perform any additional action when changing this field because we perform validation against their `userId`.


himself65:
i dont think differentiation server/client by checking this is good

PR #8477, opened 2026-03-07
docs(2fa): fix OTP server API method names

### LINKED ISSUE #8473: Documentation says 'send2FaOTP' but function is called sendTwoFactorOTP

### To Reproduce

N/A

See
https://github.com/better-auth/better-auth/blob/7fefc2bbed2b66b2349ca182144b10eff171e345/packages/better-auth/src/plugins/two-factor/otp/index.ts#L387-L405

vs the documentation:

https://better-auth.com/docs/plugins/2fa#sending-otp

v1.5.3

### System info

```bash
N/A
```

Documentation

```typescript

```

### Additional context

See also https://github.com/better-auth/better-auth/discussions/8471 .

GautamBytes:
## Summary

Fixes #8473 

- fix the 2FA OTP docs to use the actual public server API names
- align the OTP endpoint JSDoc with the exported `auth.api` surface
- keep internal helper names unchanged to avoid unnecessary runtime churn

## Details

The 2FA OTP plugin already exposes:
- `auth.api.sendTwoFactorOTP`
- `auth.api.verifyTwoFactorOTP`

The docs page and the inline OTP endpoint JSDoc were still referring to:
- `auth.api.send2FaOTP`
- `auth.api.verifyOTP`

This PR updates both documentation sources so they match the real public API.

---
## Summary by cubic
Fix 2FA OTP docs/JSDoc to use correct API names, clarify client usage, and show header forwarding for server calls; adds APIMethod requireHeaders. Replace send2FaOTP/verifyOTP with auth.api.sendTwoFactorOTP/auth.api.verifyTwoFactorOTP and authClient.twoFactor.sendOtp/verifyOtp; no runtime changes (fixes #8473).

<sup>Written for commit 9b349455e47ab3083b5f05bd4252cb4890b0ecdf. Summary will update on new commits.</sup>

### DIFF
--- a/docs/content/docs/plugins/2fa.mdx
+++ b/docs/content/docs/plugins/2fa.mdx
@@ -159,10 +159,15 @@ const authClient = createAuthClient({
 
 When you call `auth.api.signInEmail` on the server, and the user has 2FA enabled, it will return an object where `twoFactorRedirect` is set to `true`. This behavior isn’t inferred in TypeScript, which can be misleading. You can check using `in` instead to check if `twoFactorRedirect` is set to `true`.
 
+`authClient.twoFactor.*` handles cookies automatically in the browser. If you continue the 2FA flow with `auth.api.*` on the server, you must pass the relevant headers so Better Auth can read the current 2FA state and set the resulting 2FA/session cookies. The generated examples below use `await headers()` for this, but in other frameworks you should pass the equivalent incoming request headers.
+
+If you chain multiple `auth.api.*` calls in the same server flow, make sure you forward the cookies from the previous auth response into the next call.
+
 ```ts title="sign-in.ts"
 import { auth } from "@/lib/auth"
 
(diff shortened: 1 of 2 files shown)


kvanbere:
You may want to add a comment to the docs at some stage that you need to pass ‘headers:’ to these functions as well or it won’t work because they all read/write cookies.

edit: what I mean to say is the code example for these functions doesn’t pass headers and therefore doesn’t work, in Svelte anyway.

GautamBytes:
> You may want to add a comment to the docs at some stage that you need to pass ‘headers:’ to these functions as well or it won’t work because they all read/write cookies.
> 
> edit: what I mean to say is the code example for these functions doesn’t pass headers and therefore doesn’t work, in Svelte anyway.

Great point. I’ve updated the auth.api server examples for 2FA to explicitly include headers and added a note on forwarding request headers/cookies.

PR #8499, opened 2026-03-08
feat(cli): studio command

jslno:


bytaesu:
This isn't currently planned. If needed, we can revisit this PR later.

PR #8516, opened 2026-03-09
fix(api): preserve response metadata for early before hook returns

ssijak:
## Summary
- stop passing request headers into `toResponse()` for early `before` hook returns
- preserve actual response headers and status when `returnHeaders` / `returnStatus` are requested
- add regression coverage for request-header leakage on early `before` responses

## Root Cause
Early `before` hook responses were being converted with `context?.headers`, which are request headers. That could leak request `content-length` and other request metadata into the response.

## Changes
- derive response headers and status from `toResponse(before)`
- return those derived headers/status for the early `before` short-circuit path
- keep existing behavior for non-response return values

## Testing
- `pnpm --filter better-auth exec vitest run src/api/to-auth-endpoints.test.ts`
- `pnpm exec biome check packages/better-auth/src/api/to-auth-endpoints.ts packages/better-auth/src/api/to-auth-endpoints.test.ts`

---
## Summary by cubic
Fix early before hook handling in `better-auth` to preserve response headers and status and prevent leaking request headers. `asResponse`, `returnHeaders`, and `returnStatus` now reflect the actual `Response` from the hook.

- **Bug Fixes**
  - Use `toResponse(before)` to derive headers and status; no longer reuse request headers.
  - For early `before` returns: respect `asResponse`, and return real headers/status for `returnHeaders`/`returnStatus`.
  - Add regression tests to ensure request headers (e.g., `content-length`, `x-request`) don’t leak into responses.

<sup>Written for commit 0cbdf5fb2db245370b8a53b93065cc459677e43f. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/to-auth-endpoints.test.ts
+++ b/packages/better-auth/src/api/to-auth-endpoints.test.ts
@@ -199,6 +199,15 @@ describe("before hook", async () => {
 					return { response: true };
 				},
 			),
+			responseHeaders: createAuthEndpoint(
+				"/response-headers",
+				{
+					method: "POST",
+				},
+				async (c) => {
+					return { response: true };
+				},
+			),
 		};
 
 		const authContext = init({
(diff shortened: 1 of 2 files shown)


gustavovalverde:
#8579 was closed in favor of this one; nothing specific to carry over (the header-leak regression is already asserted here).

PR #3417, opened 2025-07-17
feat(magic-link): allow validated additional data for the send email function

Gittified:
# Magic Link

## Motivation

When working on my own application, I came across the issue that I can currently not provide any "custom" data from the client to the function that eventually sends the magic link email. This had me brainstorming, but eventually I figured that I might as well contribute a solution.

My specific use-case is that when someone attempts to log in, I want to show the country that the request came from. Since I use Next.js server actions to call most of my auth functions, the provided `Request` object is `undefined` because no actual HTTP request takes place. This is frustrating because I am able to add headers and a body on the server API but not actually use them there that way. At first, I considered simply changing `ctx.request` to `ctx` and just providing the whole context, but that solution didn't feel right, is difficult to make type-safe and would be a "breaking change".

## Implementation

- Added the `additionalDataSchema` property to the `MagicLinkOptions` allowing a Zod scheme to be configured, defaults to `Record<string, any>`.
- Added the Zod scheme to the request body under the `additionalData` property.
- Added the `additionalData` property to the function signature of `sendMagicLink()`
- Works on both the server & client API.

## Testing

I do not see the need for new tests here.

## Related

- ~~I've noticed that my code is impacted by #3399 which changes the way `ZodObject` works.~~ Fixed

## Checklist

- Updated documentation
- Types work well
- Code has been tested
- Backward compatibility maintained
    

---

## Summary by cubic
Added support for passing custom additional data to the magic link email function, with optional Zod schema validation for type safety.

- **New Features**
  - Accepts an `additionalData` field in the magic link request body.
  - Allows defining an `additionalDataSchema` for validation and type inference.
  - Passes `additionalData` to the `sendMagicLink` function on both server and client.

PR #8528, opened 2026-03-09
feat(magic-link): add additionalData support for sendMagicLink

himself65:
## Summary

- Added `additionalData` field to the magic link sign-in request body, allowing custom data to be passed from the client to the `sendMagicLink` function
- Added `additionalDataSchema` option accepting a Zod schema for runtime validation and type safety of the additional data
- When no schema is provided, defaults to `Record<string, JsonValue>` (unvalidated)
- Updated documentation with usage examples and configuration details

Ported from #3417 and adapted to the current canary codebase (which has been refactored since the original PR — different imports, extracted body schema, `ctx` parameter instead of `request`).

## Test plan

- [x] All 16 existing magic-link tests pass
- [x] Lint passes
- [x] Verify `additionalData` flows through to `sendMagicLink` callback with and without a custom schema
- [x] Verify type inference works correctly when `additionalDataSchema` is provided

### DIFF
--- a/docs/content/docs/plugins/magic-link.mdx
+++ b/docs/content/docs/plugins/magic-link.mdx
@@ -20,7 +20,7 @@ Magic link or email link is a way to authenticate users without a password. When
     export const auth = betterAuth({
         plugins: [
             magicLink({ // [!code highlight]
-                sendMagicLink: async ({ email, token, url }, ctx) => { // [!code highlight]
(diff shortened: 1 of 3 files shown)


gustavovalverde:
Closing in favor of #8571 (merged). The accepted API is `metadata` (a `Record<string, any>` field on `signIn.magicLink` forwarded to `sendMagicLink`), not `additionalData` as proposed here. #8571 also covered docs and a type-inference test.

Issue #8535, opened 2026-03-10
Expose `verifyApiKey` on `authClient` for server-to-server and hosted Better Auth use cases


heiwen:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

**Summary**

A valid use case for `authClient.apiKey.verify(...)` is when Better Auth is hosted as a separate auth service and another backend needs to verify API keys against it over HTTP.

In that setup, the consuming backend cannot use `auth.api.verifyApiKey(...)` because `auth.api` is only available in-process on the Better Auth server. The only current option is to call `POST /api-key/verify` manually with `fetch()`.

**Why this is a problem**

That means remote backends lose the normal Better Auth client ergonomics:

- no typed client method
- no path abstraction
- no shared request/response typing
- manual error handling around a supported endpoint

This feels like an odd gap because the endpoint already exists and is meant to be callable over HTTP, but there is no equivalent client wrapper.

**Current behavior**

Available today:

- in-process server usage via `auth.api.verifyApiKey(...)`
- direct HTTP usage via `POST /api-key/verify`

Missing today:

- `authClient.apiKey.verify(...)`

**Concrete use case**

We host Better Auth as a standalone service and want another backend service to verify API keys against it.

That second service is not a browser client, but it is also not running in the same process as the Better Auth server, so `auth.api` is not available. A typed `authClient` method would be the natural integration point.

**Proposed API**

```ts
const result = await authClient.apiKey.verify({
  key: apiKey,
  configId: "secret",
  permissions: {
    projects: ["read"],
  },
});
```

**Why this seems reasonable**

- `/api-key/verify` already exists
- `authClient` is also used outside the browser, including server-side HTTP integrations
- this would close the gap between supported HTTP endpoints and exposed client methods
- consumers with hosted Better Auth deployments would no longer need to hand-roll `fetch()` calls

**Notes**

I understand the reasoning for treating API key verification as server-side only. This request is not about encouraging browser-side verification. It is about supporting remote server-to-server usage with a typed client wrapper.

If exposing this on the default browser-oriented client surface is a concern, an alternative could be documenting it as a server-side `authClient` capability from `better-auth/client`, rather than as a browser-facing pattern.

### What version of Better Auth are you using?

1.5

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "unknown",
    "release": "25.2.0",
    "cpuCount": 8,
    "cpuModel": "Apple M2",
[...]

PR #8546, opened 2026-03-11
feat: expose /api-key/verify to clients

heiwen:
Added POST /api-key/verify as a concrete API key route and registered it in the API key client plugin. This enables client-side access to API key verification through authClient.apiKey.verify(...) and supports HTTP-based verification flows that were not previously available.

Fixes: https://github.com/better-auth/better-auth/issues/8535

---
## Summary by cubic
Exposed POST `/api-key/verify` to clients. Apps can now call `authClient.apiKey.verify(...)` to verify API keys over HTTP.

<sup>Written for commit 6e211c4cdd1d70ef508af66303b19d633c2f5956. Summary will update on new commits.</sup>

### DIFF
--- a/packages/api-key/src/client.ts
+++ b/packages/api-key/src/client.ts
@@ -10,6 +10,7 @@ export const apiKeyClient = () => {
 		$InferServerPlugin: {} as ReturnType<typeof apiKey>,
 		pathMethods: {
 			"/api-key/create": "POST",
+			"/api-key/verify": "POST",
 			"/api-key/delete": "POST",
 			"/api-key/delete-all-expired-api-keys": "POST",
 		},
--- a/packages/api-key/src/routes/verify-api-key.ts
+++ b/packages/api-key/src/routes/verify-api-key.ts
@@ -250,6 +250,7 @@ export function verifyApiKey({
 	): Promise<void>;
 }) {
 	return createAuthEndpoint(
+		"/api-key/verify",
 		{
 			method: "POST",
 			body: verifyApiKeyBodySchema,


himself65:
Thanks for the PR! We're closing this because exposing `/api-key/verify` over HTTP introduces security concerns that outweigh the convenience:

1. **Unauthenticated brute-force surface** — The verify endpoint has no session middleware and no auth check (by design — it validates the API key itself). Exposing it over HTTP means anyone can hit it and attempt to guess/brute-force API keys.

2. **Side effects on every call** — Verification is not read-only. It decrements `remaining` usage counts, triggers refill logic, deletes expired/exhausted keys, and updates timestamps. An unauthenticated public endpoint allows any caller to consume API key usage quotas.

3. **Information disclosure** — On failure, the endpoint returns specific error codes (`KEY_DISABLED`, `KEY_EXPIRED`, `USAGE_EXCEEDED`, `RATE_LIMITED`) that reveal key state to unauthenticated callers. On success, it returns the full key metadata (permissions, remaining counts, rate limit state, configId, etc.).

4. **API keys in HTTP request bodies** — While HTTPS protects in transit, keys traveling from client to server over HTTP increases attack surface (browser devtools, proxy logs, client-side error reporting, etc.).

The endpoint was intentionally designed as "path-less" (no first argument to `createAuthEndpoint`), which the `better-call` router skips during HTTP route registration. This keeps verification as a server-side-only operation via `auth.api.verifyApiKey(...)`.

For the server-to-server use case described in #8535, the recommended approach is to call `auth.api.verifyApiKey(...)` in-process, or if the auth server is remote, to create a thin authenticated wrapper endpoint on your auth server that your other services can call.

PR #8572, opened 2026-03-12
feat(magic-link): return url and token from sign-in 

mrgrauel:
## Summary
- restore the public `signIn.magicLink` HTTP response to status-only so the raw token never reaches browser clients
- add an explicit `returnToken` option and a server-only `auth.api.signInMagicLinkServer()` helper for trusted server-side workflows
- update docs and tests with the security warning, opt-in server-only usage, and hashed/custom storage coverage

## Validation
- `pnpm --filter better-auth exec vitest run src/plugins/magic-link/magic-link.test.ts`

## Notes
- `authClient.signIn.magicLink()` and `POST /sign-in/magic-link` still return only `{ status: true }`
- when `returnToken: true` is enabled, `auth.api.signInMagicLinkServer()` returns the generated `url` and raw `token`
- the raw token remains opt-in and server-only; it is never added to the public HTTP/OpenAPI contract

### DIFF
--- a/docs/content/docs/plugins/magic-link.mdx
+++ b/docs/content/docs/plugins/magic-link.mdx
@@ -89,6 +89,23 @@ type signInMagicLink = {
 ```
 </APIMethod>
 
+On success, `signIn.magicLink` returns the generated magic-link payload:
+
+```ts
+const response = await authClient.signIn.magicLink({
+  email: "user@email.com",
+});
+
+response.data;
+// {
+//   status: true,
+//   url: "https://example.com/api/auth/magic-link/verify?token=...",
+//   token: "..."
+// }
+```
+
+The returned `url` and `token` are the same values passed to `sendMagicLink`.
+
 <Callout>
 If the user has not signed up, unless `disableSignUp` is set to `true`, the user will be signed up automatically.
 </Callout>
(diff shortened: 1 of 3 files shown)


himself65:
## Security Concern

Returning the raw magic-link token in the HTTP response body is a significant security issue that should block this PR.

The magic-link security model relies on **out-of-band verification** — the token is only accessible in the user's email inbox, not in the browser context that requested it. Returning it to the requesting client collapses two-channel verification into a single channel.

### Specific risks:

- **Any JavaScript on the page** (third-party scripts, browser extensions, XSS vectors) can read the token from the response and authenticate as the user
- **Network intermediaries** (proxies, CDNs, logging) that inspect response bodies will capture authentication tokens
- **Client-side logging/telemetry** may inadvertently capture and transmit the token

This is equivalent to returning a password reset token in the API response — it defeats the purpose of out-of-band verification.

### If there's a legitimate use case (server-to-server, testing, custom delivery), consider:

1. Gating behind an explicit opt-in option (e.g., `returnToken: true` in `MagicLinkOptions`, default `false`)
2. Only returning the token from the server-side `auth.api.signInMagicLink()` call, not the HTTP endpoint
3. Adding clear documentation warnings about the security implications

himself65:
Closing due to the security concern above. The raw authentication token should not be exposed in HTTP responses by default. Happy to re-open if the approach is revised to gate token exposure behind an explicit opt-in option.

PR #7224, opened 2026-01-09
feat(magic-link): add `generateMagicLink` endpoint

himself65:
## Summary by cubic
Added a POST /generate-magic-link endpoint to create magic link URLs and tokens without sending emails. Also updated magic link verification to return full user and session data, including additional fields.

- **New Features**
  - Added POST /generate-magic-link returning { url, token } with OpenAPI metadata.
  - Extracted shared createMagicLink logic used by both generation and email flows.

- **Bug Fixes**
  - /magic-link/verify now uses parseUserOutput and parseSessionOutput, exposing additional user fields and session when no callbackURL is provided.

<sup>Written for commit d18ce48e7b74ed8db43cce673857ac25779220b9. Summary will update on new commits.</sup>

PR #8586, opened 2026-03-13
feat(magic-link): add server-only generateMagicLink API

mrosberghaus:
## Context
- this PR is the follow-up created from the feedback on #8572
- instead of returning the token from the existing HTTP/client sign-in flow, this version adds a separate server-only `auth.api.generateMagicLink()` API
- a follow-up fix restores declared callback route paths in auth hooks so existing `/callback/:id` matchers keep working

## Summary
- add a separate server-only `auth.api.generateMagicLink()` that returns `{ status, url, token }`
- keep `authClient.signIn.magicLink()` and `auth.api.signInMagicLink()` status-only and still call `sendMagicLink`
- share the issuance logic between the server-only and HTTP flows
- preserve declared endpoint paths like `/callback/:id` while still giving pathless server-only endpoints a string hook path
- document the new server helper and add regression coverage for both server parity and routed hook paths

## Validation
- `pnpm exec vitest run packages/better-auth/src/plugins/magic-link/magic-link.test.ts`
- `pnpm exec vitest run packages/better-auth/src/api/to-auth-endpoints.test.ts`
- `pnpm exec vitest run packages/better-auth/src/social.test.ts`
- `pnpm exec vitest run packages/better-auth/src/plugins/oauth-proxy/oauth-proxy.test.ts` (callback/state regressions pass locally; the Postgres-backed UUID case requires the repo Docker setup)
- `pnpm exec vitest run packages/better-auth/src/api/routes/account.test.ts`
- `pnpm exec vitest run packages/better-auth/src/oauth2/link-account.test.ts`


mrosberghaus:
@himself65 Thank you for your feedback in #8572. It was very helpful and provided clear guidance on how to address your concerns.

I chose a separate generateMagicLink() API instead of a returnToken option because these are two different operations with different side effects. The normal magic-link sign-in flow should continue to send the email and return only status, while the server-only flow should return the raw token and let the caller decide whether, when, and how to send it. Keeping them separate avoids conditional behavior in the existing API, reduces security ambiguity, and makes the flow easier to test and maintain.

mrosberghaus:
@himself65 or @Bekacru can you look into it?

docs/content/docs/plugins/magic-link.mdx
```diff
@@ -97,6 +97,42 @@ type signInMagicLink = {
...
+  return only `{ status: true }`. They do not return the raw token.
+</Callout>
+
+### Generate Magic Link
+
+For trusted server-side workflows, you can generate a magic link without calling
+`sendMagicLink` by using `auth.api.generateMagicLink()`.
+
+```ts
+const response = await auth.api.generateMagicLink({
```
ping-maxwell: can you make this use  &lt;APIMethod&gt; and apply serverOnly?

ping-maxwell (approved):
everything looks good to me, just change the docs code example to use APIMethod component and it should be good to go

docs/content/docs/plugins/magic-link.mdx
mrosberghaus: Updated in 19db10ddb. The Generate Magic Link docs now use `<APIMethod>` with `isServerOnly`, and I also added the matching endpoint annotation in the plugin source comment so the docs and source stay aligned.

mrosberghaus: Thanks for you review. Does 19db10ddbf57a08f301626ab8ac085818c89c8fc satisfies your request?

ping-maxwell (dismissed):

docs/content/docs/plugins/magic-link.mdx
```diff
@@ -97,6 +97,63 @@ type signInMagicLink = {
...
+     * URL to redirect if an error happens on verification.
+     * If only callbackURL is provided but without an `errorCallbackURL` then they will be
+     * redirected to the callbackURL with an `error` query parameter.
+     */
+    errorCallbackURL?: string = "/error"
+}
+```
+</APIMethod>
+
+`generateMagicLink` returns the generated `url` and raw `token`, and leaves
```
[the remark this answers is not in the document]
mrosberghaus: I wouldn't recommend following the AI suggestion, as the important information is only the `url` and `token`.

mrosberghaus:
@himself65 what's your opinion?

mrosberghaus:
@Bekacru can you say something?

packages/better-auth/src/plugins/magic-link/index.ts
```diff
@@ -145,6 +148,22 @@ const magicLinkVerifyQuerySchema = z.object({
...
+	url: string;
+	token: string;
+};
+
+type MagicLinkIssuance = {
+	email: string;
+	url: string;
+	token: string;
+};
+
```
bytaesu: Can we keep these types internal instead of exporting them?

packages/better-auth/src/api/to-auth-endpoints.test.ts
bytaesu, line 1: It would be great to have a test case to verify fallback behavior for pathless endpoints 🧐

docs/content/docs/plugins/magic-link.mdx
```diff
@@ -97,6 +97,63 @@ type signInMagicLink = {
 If the user has not signed up, unless `disableSignUp` is set to `true`, the user will be signed up automatically.
 </Callout>
 
+<Callout type="warn">
+  `signIn.magicLink`, `auth.api.signInMagicLink`, and `POST /sign-in/magic-link`
+  return only `{ status: true }`. They do not return the raw token.
+</Callout>
```
bytaesu: I think we don't need a new callout here. We can either fold this into the one above as a single line, or remove it 🧐

packages/better-auth/src/plugins/magic-link/index.ts
mrosberghaus: Fixed in 8d82ca880268fcf2c267f69a379f049e4a427a1b

packages/better-auth/src/api/to-auth-endpoints.test.ts
mrosberghaus: Added in c931983471b0c8aea64f94e7657b94146ea1de4b

docs/content/docs/plugins/magic-link.mdx
mrosberghaus: Folded in c69cfb4625186a73be0cf96d8a607ee7f58a607f

ping-maxwell:
hey @mrgrauel please add the changeset 🙌

mrosberghaus:
@ping-maxwell done and sorry I didn't check the new contribution guidelines.

mrgrauel:
@ping-maxwell do you need anything else?

mrgrauel:
@ping-maxwell and @Bekacru do you want more changes?

mrgrauel:
@bytaesu you are assigned. Can I assist you?

gustavovalverde:
@mrgrauel, we'll take a look at this once we've gone through the other prioritized issues and PRs. If you need this urgently for your deployment you can use `pnpm patch`

gustavovalverde:
#7224 was closed in favor of this one. Things to consider from that PR:
- `generateMagicLinkBodySchema` (omits `metadata` field)
- `issueMagicLink` shared issuance helper
- `attempt: 0` initialization in the stored verification value

brennanbutler01:
I put together seven additional round-trip tests for this proposal: [supplemental commit](https://github.com/brennanbutler01/better-auth/commit/9e8093e9e5b75fb9af1dabbd70c02b44f6058323) ([patch](https://github.com/brennanbutler01/better-auth/commit/9e8093e9e5b75fb9af1dabbd70c02b44f6058323.patch)). It is based directly on `bb1036e0` and can be cherry-picked if useful.

The tests generate a link with `auth.api.generateMagicLink()`, redeem the returned URL through `auth.handler`, and use the response cookies to verify the authenticated user. They cover plain, hashed, and custom-hashed storage; callback query parameters; rejection of reuse and expiration; a custom authentication path; and no email delivery.

All seven pass with your existing implementation—this is additional coverage, not a bug fix. The change adds only tests and an empty changeset.

Verified on Node 24.13.0 with pnpm 10.30.2: all 34 magic-link tests, 60 authentication-handler tests, package builds, type checking, lint, and spelling passed. The broad suite passed 3,678 tests, with three database-related failures (two missing PostgreSQL role errors and one MongoDB timeout), plus test-startup failures in the memory-adapter and redis-storage packages. Those are outside the changed test file; I have not verified the full suite as green.

Prepared with OpenAI Codex assistance. Happy to adjust the coverage to fit this proposal.

PR #8599, opened 2026-03-13
fix(cookies): use dynamic chunk size

### LINKED ISSUE #8585: Cookies are not being chunked

### To Reproduce

* Better auth config with Entra
* The user data needs to generate a session cookie that is approx 4069 but less than 4093
*

Cookie is too big and not being chunked

1.4.18

### System info

```bash
"system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#202602262342 SMP PREEMPT_DYNAMIC Fri Feb 27 00:13:55 UTC 2026",
    "release": "6.19.4-061904-generic",
    "cpuCount": 32,
    "cpuModel": "AMD Ryzen 9 7950X 16-Core Processor",
    "totalMemory": "29.99 GB",
    "freeMemory": "6.59 GB"
  },
  "node": {
    "version": "v24.14.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.9.0"
  },
  "frameworks": [
    {
      "name": "react",
      "version": "^19.2.4"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.17.2"
    }
  ],
  "betterAuth": {
    "version": "^1.4.18",
    "config": null
  }
}
```

Client

```typescript
return betterAuth({
    logger: {
      level: 'debug',
    },
    account: {
      storeAccountCookie: false, // NOTE: Don't store the account data as the cookie is too big
    },
    session: {
      cookieCache: {
        maxAge: 60 * 60 * 24 * 1, // 1 day
        refreshAge: true,
      },
    },
    secret: config.AUTH_SECRET,
    advanced: {
      cookiePrefix: 'abc-admin',
    },
    socialProviders: {
      microsoft: {
        clientId: config.AUTH_MICROSOFT_ENTRA_ID_ID,
        clientSecret: config.AUTH_MICROSOFT_ENTRA_ID_SECRET,
        tenantId: config.AUTH_MICROSOFT_ENTRA_ID_TENANT_ID,
      },
    },
    plugins: [
      tanstackStartCookies(), // NOTE: Must be last
    ],
  });
```

### Additional context

I'm pretty sure the recent cookie chunking work missed something.

In https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/cookies/index.ts#L234
```
// Check if we need to chunk the cookie (only if it exceeds 4093 bytes)
	if (data.length > 4093) {
```

This doesn't use the more sophisticated logic that was implemented in https://github.com/better-auth/better-auth/pull/6393

/cc @jslno

jslno:
closes #8585

### DIFF
--- a/packages/better-auth/src/cookies/cookie-utils.ts
+++ b/packages/better-auth/src/cookies/cookie-utils.ts
@@ -1,3 +1,4 @@
+import { BetterAuthError } from "@better-auth/core/error";
 import type { CookieOptions } from "better-call";
 
 export const ALLOWED_COOKIE_SIZE = 4096;
@@ -177,7 +178,7 @@ export function setCookieToHeader(headers: Headers) {
 export function estimateEmptyCookieSize(
 	name: string,
 	attributes: CookieOptions,
-): number {
+) {
 	const path = attributes.path ?? "/";
 	let s = `${name}=`;
 	s += `; Path=${path}`;
@@ -204,9 +205,10 @@ export function estimateEmptyCookieSize(
 	return s.length;
 }
 
-export function getMaxCookieValueSize(
-	name: string,
-	attributes: CookieOptions,
-): number {
-	return ALLOWED_COOKIE_SIZE - estimateEmptyCookieSize(name, attributes);
+export function getMaxCookieValueSize(name: string, attributes: CookieOptions) {
+	const size = ALLOWED_COOKIE_SIZE - estimateEmptyCookieSize(name, attributes);
+	if (size <= 0) {
+		throw new BetterAuthError(`Cookie ${name} is too large to be chunked.`);
+	}
+	return size;
 }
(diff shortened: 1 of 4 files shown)


packages/better-auth/src/cookies/cookie-utils.ts
```diff
@@ -1,3 +1,8 @@
+import { BetterAuthError } from "@better-auth/core/error";
+import type { CookieOptions } from "better-call";
+
+export const ALLOWED_COOKIE_SIZE = 4096;
```
narek-haytyan, line 4: The actual cookie max size varies from broser to browser:
https://docs.devexpress.com/AspNet/11912/common-concepts/cookies-support#browser-limitations

So to be on the safe size it's better to use the lowest known value as max size: 4093 from Safari.

BElluu:
Hello. Guys, this is really important thing. I do not have problem in FireFox, but in Chrome we got error, because have malformed response cookie. When will you merge this PR ?

bytaesu:
Closing this in favor of #10088

PR #8608, opened 2026-03-14
fix(sso): read InResponseTo from extract.response instead of extract

### LINKED ISSUE #8607: SAML InResponseTo validation always fails — extract.inResponseTo is undefined, should be extract.response?.inResponseTo

### To Reproduce

1. Configure the SSO plugin with strict InResponseTo validation:
```ts
sso({
  saml: {
    enableInResponseToValidation: true,
    allowIdpInitiated: false,
  },
})
```
2. Register a SAML provider (e.g., Microsoft Entra ID / Azure AD)
3. Initiate an SP-initiated SSO login from your application (not from the IdP portal)
4. Azure AD authenticates the user and sends a SAML Response back to the ACS endpoint with a valid `InResponseTo` attribute
5. Login is rejected with: `SAML IdP-initiated SSO rejected: InResponseTo missing and allowIdpInitiated is false`

**Expected:** SP-initiated SAML login succeeds. The plugin reads `InResponseTo` from the parsed SAML Response, matches it against the stored AuthnRequest ID, and allows the login.

**Actual:** Every SP-initiated login is rejected. The plugin reads `extract.inResponseTo` which is always `undefined`. The `InResponseTo` value from Azure AD is actually present at `extract.response.inResponseTo` — `samlify`'s extractor nests it under the `response` key. Because the plugin sees `undefined`, it misclassifies every SP-initiated response as IdP-initiated and rejects it.

**The flow is SP-initiated.** Azure AD correctly includes `InResponseTo` in its SAML Response. The plugin simply reads the wrong field path, so it can never find the value.

Proof using samlify's extractor directly:
```js
const { extract, loginResponseFields } = require('samlify/build/src/extractor');

const xml = `<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
  xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion"
  ID="_resp1" Version="2.0" IssueInstant="2026-01-01T00:00:00Z"
  Destination="http://localhost/acs" InResponseTo="_req1">
  <saml:Issuer>https://idp.example.com</saml:Issuer>
  <samlp:Status><samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/></samlp:Status>
  <saml:Assertion ID="_a1" Version="2.0" IssueInstant="2026-01-01T00:00:00Z">
    <saml:Issuer>https://idp.example.com</saml:Issuer>
    <saml:Subject><saml:NameID>user@example.com</saml:NameID></saml:Subject>
    <saml:Conditions NotBefore="2026-01-01T00:00:00Z" NotOnOrAfter="2026-01-01T01:00:00Z">
      <saml:AudienceRestriction><saml:Audience>http://localhost</saml:Audience></saml:AudienceRestriction>
    </saml:Conditions>
    <saml:AuthnStatement AuthnInstant="2026-01-01T00:00:00Z" SessionIndex="_s1"/>
  </saml:Assertion>
</samlp:Response>`;

const result = extract(xml, loginResponseFields(null));
console.log(result.inResponseTo);           // undefined  ← what the plugin reads
console.log(result.response.inResponseTo);  // "_req1"    ← where the value actually is
```

The logout response handler in the same file does it correctly:
```js
// handleLogoutResponse — correct
const inResponseTo = extract?.response?.inResponseTo;
```

But both login handlers (callbackSSOSAML and acsEndpoint) do it wrong:
```js
// callbackSSOSAML — incorrect
const inResponseTo = extract.inResponseTo;

// acsEndpoint — incorrect
const inResponseToAcs = extract.inResponseTo;
```

**Suggested fix:**
```diff
- const inResponseTo = extract.inResponseTo;
+ const inResponseTo = extract.response?.inResponseTo;
```

1.5.5

### System info

```bash
**System info:**
- `@better-auth/sso`: 1.5.5
- `samlify`: 2.10.2
- IdP: Microsoft Entra ID (Azure AD) with SAML 2.0
- Node.js 24+
- macOS
```

Package

```typescript

```

### Additional context

Additional context:

This is reproducible locally. The bug likely affects all environments since it's a code path issue, not environment-specific. The feature (enableInResponseToValidation) defaults to false, which is probably why it hasn't been reported — most users never enable strict validation. I verified the extract structure by running samlify's extractor directly against a valid SAML Response XML (reproduction script above).

evanenochs:
Fixes #8607
The SAML login handlers read `extract.inResponseTo`, but samlify nests
the Response attributes under `extract.response`. This causes
`InResponseTo` to always be `undefined`, making
`enableInResponseToValidation` reject every SP-initiated login.
The logout handler in the same file already reads the correct path
(`extract?.response?.inResponseTo`). This PR aligns the two login
handlers to match.

---
## Summary by cubic
Fixes SP-initiated SAML logins failing when `enableInResponseToValidation` is on by reading `InResponseTo` from `extract.response.inResponseTo`. Adds a full SP-initiated validation test and tightens unsolicited-response tests.

- **Bug Fixes**
  - In `packages/sso/src/routes/sso.ts`, `callbackSSOSAML` and `acsEndpoint` now read `extract.response?.inResponseTo`.
  - In `packages/sso/src/types.ts`, extend `SAMLAssertionExtract` with `response.inResponseTo`.
  - In `packages/sso/src/saml.test.ts`, add SP-initiated test that extracts the AuthnRequest ID from the deflated `SAMLRequest` and uses it as `InResponseTo`; update the mock IdP to accept a dynamic `inResponseTo`; unsolicited tests send an empty value; fix a test typecheck.

<sup>Written for commit b09762afeb2aeeb675d9420a88eadfa4febd14a7. Summary will update on new commits.</sup>

### DIFF
--- a/packages/sso/src/saml.test.ts
+++ b/packages/sso/src/saml.test.ts
@@ -2250,7 +2250,7 @@ describe("SAML SSO", async () => {
 		const deflatedBuffer = Buffer.from(samlRequestParam, "base64");
 		const xml = inflateRawSync(deflatedBuffer).toString("utf-8");
 		const idMatch = xml.match(/ID="([^"]+)"/);
-		const authnRequestId = idMatch![1];
+		const authnRequestId = idMatch![1] as string;
 
 		let samlResponse: any;
 		await betterFetch(
--- a/packages/sso/src/saml.test.ts
+++ b/packages/sso/src/saml.test.ts
@@ -1,5 +1,6 @@
 import { randomUUID } from "node:crypto";
 import type { createServer } from "node:http";
+import { inflateRawSync } from "node:zlib";
 import { betterFetch } from "@better-fetch/fetch";
 import { betterAuth } from "better-auth";
 import { memoryAdapter } from "better-auth/adapters/memory";
@@ -313,7 +314,8 @@ const generateRequestID = () => {
 	return "_" + randomUUID();
 };
 const createTemplateCallback =
-	(idp: any, sp: any, email: string) => (template: any) => {
+	(idp: any, sp: any, email: string, inResponseTo?: string) =>
+	(template: any) => {
 		const assertionConsumerServiceUrl =
 			sp.entityMeta.getAssertionConsumerService(
 				saml.Constants.wording.binding.post,
@@ -343,7 +345,7 @@ const createTemplateCallback =
 			SubjectConfirmationDataNotOnOrAfter: fiveMinutesLater.toISOString(),
 			NameIDFormat: selectedNameIDFormat,
 			NameID: email,
-			InResponseTo: "null",
+			InResponseTo: inResponseTo ?? "null",
 			AuthnStatement: "",
 			attrFirstName: "Test",
 			attrLastName: "User",
@@ -400,6 +402,7 @@ const createMockSAMLIdP = (port: number) => {
 		"/api/sso/saml2/idp/post",
 		async (req: ExpressRequest, res: ExpressResponse) => {
 			const emailCase = req.query.emailCase as string;
+			const inResponseTo = req.query.inResponseTo as string | undefined;
 			const emailValue =
 				emailCase === "mixed" ? "TestUser@Example.com" : "test@email.com";
 			const user = {
@@ -412,7 +415,7 @@ const createMockSAMLIdP = (port: number) => {
 				{} as any,
 				saml.Constants.wording.binding.post,
 				user,
-				createTemplateCallback(idp, sp, user.emailAddress),
+				createTemplateCallback(idp, sp, user.emailAddress, inResponseTo),
 			);
 			res.status(200).send({ samlResponse: context, entityEndpoint });
 		},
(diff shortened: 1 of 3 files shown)


packages/sso/src/saml.test.ts
```diff
@@ -1890,11 +1894,14 @@ describe("SAML SSO", async () => {
 		});
 
 		let samlResponse: any;
-		await betterFetch("http://localhost:8081/api/sso/saml2/idp/post", {
-			onSuccess: async (context) => {
-				samlResponse = await context.data;
+		await betterFetch(
+			"http://localhost:8081/api/sso/saml2/idp/post?inResponseTo=",
```
jonathansamines, line 1898: Let's revert the changes to existing tests if not actually required.

gustavovalverde:
Thanks @evanenochs — your InResponseTo path fix and SP-initiated regression test were incorporated into #9055, which ships in v1.7.0-beta.1.

PR #8624, opened 2026-03-16
fix: warn when old CLI is used with better-auth v1.5.x+

### LINKED ISSUE #8622: Old CLI should throw a warning when used with Better Auth v.1.5.x due to undefined behaviour

### To Reproduce

If you use the old CLI with a Better Auth `v1.5.x` project, some things do not work correctly.

As an example, per this thread https://github.com/better-auth/better-auth/issues/8542#issuecomment-4060326229 the CLI was generating a `verifications` table even though KV was configured.

In this case it would have been helpful if the old CLI threw a warning such as "you are using the Better Auth v1.4.x CLI with a Better Auth v1.5.x project, this could lead to unexpected results"

The user should be prompted to use the new `auth` CLI instead.

1.5.x

### System info

```bash
N/A
```

Package

```typescript

```

### Additional context

_No response_

luchersou:
Closes #8622

## What changed

Added a version check in `packages/cli/src/index.ts` that warns the user when the old `@better-auth/cli` is being used with a `better-auth` v1.5.x+ project.

## Why

The old CLI can produce unexpected behavior when used with better-auth v1.5.x or later (e.g. generating a `verifications` table even when KV is configured). The warning guides the user to the new `auth` CLI before any damage is done.

## How

* Resolves the installed `better-auth` entry point via `createRequire` (works with npm, pnpm, Yarn PnP and hoisted monorepos)
* Walks up the directory tree to find the nearest `package.json` with `name === "better-auth"`
* Compares the version against `1.5.0` using `semver` (already a dependency)
* Prints a warning to stderr and suggests `npx auth@latest` if the version is >= 1.5.0
* Silently ignored if `better-auth` is not installed yet

---
## Summary by cubic
Warns when the old `@better-auth/cli` runs in projects using `better-auth` v1.5.x+, and points users to the new `auth` CLI to avoid incorrect codegen.

- **New Features**
  - Resolves `better-auth` from the project CWD (not the CLI install path) using `createRequire`, walks up to find its `package.json`, then checks the version with `semver` (>= 1.5.0).
  - Prints a yellow warning with the CLI and `better-auth` versions, and suggests `npx auth@latest`.
  - Silently skips if `better-auth` isn’t installed.

- **Refactors**
  - Applied Biome formatting, fixed trailing newline, and minor comment cleanup.
  - Separated ANSI escape from the `npx` command to satisfy cspell.

<sup>Written for commit 7e3467f4e5d0097484726db3658f66ce2264450e. Summary will update on new commits.</sup>


himself65 (approved):

gustavovalverde:
Closing in favor of #8637 (merged). #8637 was merged into `v1.4.x` (the correct branch for the old `@better-auth/cli`) with a cleaner implementation: `createRequire(process.cwd() + "/")` to require `better-auth/package.json` directly (no directory-walk), `chalk` for output, plus a 4-case `version-warning.test.ts` covering the scenarios in this PR's manual test plan. Main and next use the new CLI.

PR #8625, opened 2026-03-16
fix(oidc-provider): public client registration

okisdev:
fixes three issues preventing MCP/public clients (`token_endpoint_auth_method: "none"`) from completing OAuth:

- hardcoded `type: "web"` → now `"public"` when auth method is `"none"`
- `||` treats `"none"` as falsy → replaced with `??`
- `definePayload` override loses JWT plugin claims → now merges with base payload, adds `scope`

---
## Summary by cubic
Enables full OIDC dynamic registration and token flows for public clients. Supports `token_endpoint_auth_method: "none"`, requires ID token signing via the `jwt` plugin when `openid` is requested, and enforces client auth on refresh tokens.

- **Bug Fixes**
  - Public registration: set type "public" when auth method is "none", preserve "none" via `??`, skip generating/storing `client_secret`, and omit secret fields (incl. `client_secret_expires_at`) in responses. OpenAPI: `type` enum adds "public", `authenticationScheme` enum lists "client_secret_basic" | "client_secret_post" | "none", `clientSecret` is nullable and not required.
  - ID token: merge the `jwt` plugin’s base payload in `definePayload`, add `scope`. Require the `jwt` plugin for public clients only when `openid` is requested; reject HS256 for them. Allow access-token-only exchanges without the `jwt` plugin when no `openid` scope.
  - Refresh tokens: validate client and reject disabled/unknown. Confidential clients must present a valid `client_secret` (verified against stored secret).

<sup>Written for commit 76667c18e0108a96c333a7f7141db7d674c787c4. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/oidc-provider/index.ts
+++ b/packages/better-auth/src/plugins/oidc-provider/index.ts
@@ -753,6 +753,46 @@ export const oidcProvider = (options: OIDCOptions) => {
 								error: "invalid_grant",
 							});
 						}
+
+						// Validate client credentials for refresh token exchange
+						const client = await getClient(
+							client_id.toString(),
+							trustedClients,
+						);
+						if (!client) {
+							throw new APIError("UNAUTHORIZED", {
+								error_description: "invalid client_id",
+								error: "invalid_client",
+							});
+						}
+						if (client.disabled) {
+							throw new APIError("UNAUTHORIZED", {
+								error_description: "client is disabled",
+								error: "invalid_client",
+							});
+						}
+						if (client.type !== "public") {
+							// Confidential clients must authenticate with client_secret
+							if (!client.clientSecret || !client_secret) {
+								throw new APIError("UNAUTHORIZED", {
+									error_description:
+										"client_secret is required for confidential clients",
+									error: "invalid_client",
+								});
+							}
+							const isValidSecret = await verifyStoredClientSecret(
+								ctx,
+								client.clientSecret,
+								client_secret.toString(),
+							);
+							if (!isValidSecret) {
+								throw new APIError("UNAUTHORIZED", {
+									error_description: "invalid client_secret",
+									error: "invalid_client",
+								});
+							}
+						}
+
 						const accessToken = generateRandomString(32, "a-z", "A-Z");
 						const newRefreshToken = generateRandomString(32, "a-z", "A-Z");
 
@@ -993,7 +1033,7 @@ export const oidcProvider = (options: OIDCOptions) => {
 						Math.floor(Date.now() / 1000) +
 						(opts?.accessTokenExpiresIn ?? DEFAULT_ACCESS_TOKEN_EXPIRES_IN);
 
-					let idToken: string;
+					let idToken: string | undefined;
 
 					// The JWT plugin is enabled, so we use the JWKS keys to sign
 					if (options.useJWTPlugin) {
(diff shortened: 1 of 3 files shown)


packages/better-auth/src/plugins/oidc-provider/index.ts
```diff
@@ -1035,7 +1036,15 @@ export const oidcProvider = (options: OIDCOptions) => {
...
-									definePayload: () => payload,
+									definePayload: async (session) => {
+										const basePayload = jwtPlugin.options?.jwt?.definePayload
+											? await jwtPlugin.options.jwt.definePayload(session)
+											: {};
+										return {
+											...basePayload,
+											...payload,
+										};
+									},
```
gustavovalverde, line 1089: This merge executes only inside the `if (options.useJWTPlugin)` branch (line 996). Under the default config (`useJWTPlugin: false`), the HS256 path at line 1045 uses `payload` directly; `definePayload` is never invoked. Add a test covering this merge behavior with a custom `definePayload` configured in the JWT plugin.

packages/better-auth/src/plugins/oidc-provider/index.ts
```diff
@@ -1408,9 +1417,10 @@ export const oidcProvider = (options: OIDCOptions) => {
 							clientId: clientId,
 							clientSecret: storedClientSecret,
 							redirectUrls: body.redirect_uris.join(","),
-							type: "web",
+							type:
+								body.token_endpoint_auth_method === "none" ? "public" : "web",
```
gustavovalverde, line 1480: The OpenAPI schema at line 1292 constrains `type` to `enum: ["web"]`. Needs to include `"public"`.

packages/better-auth/src/plugins/oidc-provider/index.ts
```diff
@@ -983,6 +983,7 @@ export const oidcProvider = (options: OIDCOptions) => {
 							: undefined,
 						nonce: value.nonce,
 						acr: "urn:mace:incommon:iap:silver", // default to silver - ⚠︎ this should be configurable and should be validated against the client's metadata
+						scope: requestedScopes.join(" "),
```
gustavovalverde, line 1028: Untested. No assertion verifies this claim appears in the decoded ID token.

packages/better-auth/src/plugins/oidc-provider/oidc.test.ts
```diff
@@ -1655,3 +1655,75 @@ describe("oidc-jwt", async () => {
...
+		customFetchImpl,
+	} = await getTestInstance({
+		baseURL: "http://localhost:3000",
+		plugins: [
+			oidcProvider({
+				loginPage: "/login",
+				consentPage: "/oauth2/authorize",
+				requirePKCE: true,
+				allowDynamicClientRegistration: true,
+			}),
```
gustavovalverde, line 1672: `useJWTPlugin: true` is not passed to `oidcProvider()`, so this runs through the HS256 code path. `jwt()` is in the plugins array (line 1673) but unused by the OIDC plugin. These tests verify registration shape only; a full token exchange test for a public client would catch the signing issue.

gustavovalverde (commented):
3 issues, ordered by severity.

**1. ID token is unverifiable for public clients under default configuration**

`useJWTPlugin` defaults to `false` (`types.ts:165`). The default code path signs ID tokens with `client.clientSecret` via HS256 (`index.ts:1045-1049`). Public clients never receive `client_secret` in the registration response, so the relying party has no key to verify the token. Registration and token exchange both succeed, but the resulting ID token is cryptographically unusable by the client.

Two possible fixes: require `useJWTPlugin: true` when public clients are enabled, or fall back to asymmetric signing (RS256/EdDSA via JWKS) for public clients on the default path.

**2. `definePayload` merge is unreachable under default config**

The merge logic (lines 1039-1047) executes inside the `if (options.useJWTPlugin)` branch (line 996). Under the default path, `new SignJWT(payload)` at line 1045 uses `payload` directly; `definePayload` is never called.

**3. Test coverage gaps**

The new tests verify registration response shape only. No test exercises the full token exchange for a public client, the `definePayload` merge, or the `scope` claim in the decoded ID token.

gustavovalverde:
Closing as this plugin is deprecated and this is solved in the new `oauth-provider` plugin

PR #8634, opened 2026-03-16
feat(oauth-provider): compute at_hash in id tokens per OIDC Core §3.1.3.6

gustavovalverde:
## Problem

OIDC Core §3.1.3.6 requires the `at_hash` claim in id_tokens when the token is issued alongside an access token from the authorization endpoint.

## Changes

- Compute `at_hash` server-side inside `createIdToken` — the access token is never exposed to `customIdTokenClaims` callbacks to prevent token leakage
- Sequence id_token creation after access token creation so the access token string is available for hashing
- Add `computeOidcHash()` that selects the correct hash algorithm based on the signing algorithm (EdDSA→SHA-512, RS/ES384→SHA-384, RS/ES512→SHA-512, else SHA-256)
- Add `base64url` import from `jose` for the hash encoding

## References

- [OIDC Core §3.1.3.6 — ID Token `at_hash`](https://openid.net/specs/openid-connect-core-1_0.html#HybridIDToken)

### DIFF
--- a/packages/oauth-provider/src/token.test.ts
+++ b/packages/oauth-provider/src/token.test.ts
@@ -10,7 +10,7 @@ import {
 } from "better-auth/oauth2";
 import { jwt } from "better-auth/plugins/jwt";
 import { getTestInstance } from "better-auth/test";
-import { createLocalJWKSet, decodeJwt, jwtVerify } from "jose";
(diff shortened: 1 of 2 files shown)


gustavovalverde:
Superseded by #9079, which ships in v1.7.0-beta.1.

PR #8636, opened 2026-03-16
feat(oauth-provider): plugin-extensible token endpoint

gustavovalverde:
## Problem

The token endpoint is the convergence point for every feature that touches the OAuth token flow. Grant dispatch, token creation, id_token creation, and grant-specific handlers all live in `token.ts`. When the endpoint lacks extension points, every plugin that adds a grant type must modify the same shared code. Features that are functionally independent (CIBA, token exchange, device authorization) become structurally coupled through edits to the same file and the same type definitions.

Two specific constraints enforce this coupling:

1. **Zod strips unknown fields.** The body schema uses `z.object({...})` which discards any field not declared in the schema. A plugin that needs `auth_req_id` (CIBA), `subject_token` (token exchange), or `client_assertion` (JWT auth) must add its field to the core schema, creating a dependency between the plugin and oauth-provider's source.

2. **`grant_type` is a closed enum.** `z.enum(["authorization_code", "client_credentials", "refresh_token"])` rejects any value outside the three built-in types before the handler runs. The switch statement's default case throws unconditionally. A plugin cannot register a custom grant type without modifying both the Zod schema and the switch statement in token.ts.

The `GrantType` TypeScript type reinforces this: it's a closed union, so code that checks `opts.grantTypes.includes(grantType)` requires a cast when the grant type is a plugin-defined string. OAuth 2.x grant types are extensible URIs by spec (RFC 6749 §8.5); the type system should reflect this rather than fight it.

### Type layer

`GrantType` is now `type GrantType = string`. The alias preserves semantic meaning in signatures (`grantTypes?: GrantType[]` reads better than `grantTypes?: string[]`) without constraining the runtime set. No casts needed anywhere.

### Zod schema

- `grant_type` widened to `z.string().trim().min(1)`
- `.passthrough()` on the token endpoint body, so plugin-specific fields survive Zod parsing without being pre-declared in the core schema
- DCR `grant_types` widened to `z.array(z.string())` per RFC 7591 §2 (the spec defines this field as an array of grant type strings, not a closed set)

### Dispatch

The switch default case looks up `customGrantTypeHandlers` from the endpoint context before throwing `unsupported_grant_type`. Plugins register handlers via `init()`:

```ts
init(ctx) {
  return {
    context: {
      customGrantTypeHandlers: {
        [MY_GRANT_TYPE]: myHandler,
      },
    },
  };
}
```

With these three changes, a plugin that adds a grant type (CIBA, token exchange, device authorization) touches zero lines in oauth-provider. It registers its handler, declares its grant type string, and reads its own fields from the passthrough body.

### Docs

- Updated token endpoint section to mention plugin extensibility
- Updated `grantTypes` config description
- Fixed DCR OpenAPI description (was "Requested authentication method")

## References

- [RFC 6749 §8.5 — Defining New Grant Types](https://datatracker.ietf.org/doc/html/rfc6749#section-8.5)
- [RFC 7591 §2 — Client Metadata (`grant_types`)](https://datatracker.ietf.org/doc/html/rfc7591#section-2)
- [RFC 8693 — OAuth 2.0 Token Exchange](https://datatracker.ietf.org/doc/html/rfc8693)
- [CIBA Core §10 — Token Request](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.10)

### DIFF
--- a/docs/content/docs/plugins/oauth-provider.mdx
+++ b/docs/content/docs/plugins/oauth-provider.mdx
@@ -461,6 +461,10 @@ By default, the token endpoint supports the following built-in grants:
 
 Plugins can register additional grant types (e.g., CIBA, token exchange) via the `customGrantTypeHandlers` context. Unknown fields in the token request body are passed through to plugin handlers.
 
+<Callout>
+Plugins that register custom grant types must also add their grant type string to the `grantTypes` allowlist during `init()`. The CIBA plugin does this automatically. If `grantTypes` is not configured (or set to `undefined`), all grant types are allowed through to dispatch.
+</Callout>
+
 #### Authorization code grant
 
 The authorization code grant enables clients to obtain access user access tokens and optionally refresh tokens (with the "offline_access" scope).
--- a/packages/oauth-provider/src/index.ts
+++ b/packages/oauth-provider/src/index.ts
@@ -6,4 +6,5 @@ export {
 	oidcServerMetadata,
 } from "./metadata";
 export { getOAuthProviderState, oauthProvider } from "./oauth";
+export { type GrantTypeHandler, type TokenResponse } from "./token";
 export type * from "./types";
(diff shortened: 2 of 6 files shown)


gustavovalverde:
Superseded by #8677

PR #8639, opened 2026-03-16
feat(captcha): add endpoint exclusions using `!` prefix

adam-enition:
Endpoints starting with `!` are treated as exclusions when matching captcha-protected routes.

This fixes an issue when the EmailOTP plugin and Captcha plugins are enabled at the same time.

Example:

```ts
endpoints: [
  "/sign-up/email",
  "/sign-in/email",
  "!/sign-in/email-otp"
]
```

Also updates the defaultEndpoints to exclude `/sign-in/email-otp` out of the box.

---
## Summary by cubic
Add `!`-prefixed endpoint exclusions for captcha and default-skip `/sign-in/email-otp` to prevent captcha on Email OTP routes and avoid EmailOTP/Captcha conflicts. Endpoint matching now uses the request `pathname` to ignore query strings.

- **Bug Fixes**
  - Removed incorrect `.url` property usage to fix type errors; matching stays on `new URL(request.url).pathname`.

<sup>Written for commit 7ee42d8c2cbb22606d135f8091acce6a48a7e132. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/captcha/constants.ts
+++ b/packages/better-auth/src/plugins/captcha/constants.ts
@@ -3,6 +3,7 @@ import type { Provider } from "./types";
 export const defaultEndpoints = [
 	"/sign-up/email",
 	"/sign-in/email",
+	"!/sign-in/email-otp",
 	"/request-password-reset",
 ];
 
--- a/packages/better-auth/src/plugins/captcha/index.ts
+++ b/packages/better-auth/src/plugins/captcha/index.ts
@@ -35,11 +35,11 @@ export const captcha = (options: CaptchaOptions) =>
 				const pathname = new URL(request.url).pathname;
 
 				const isIncluded = includeEndpoints.some((endpoint) =>
-					pathname.url.includes(endpoint),
+					pathname.includes(endpoint),
 				);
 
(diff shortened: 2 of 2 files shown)


himself65 (changes requested):
Can you add tests for this?

also why do we add `!` in default endpoints since it won't affect anything

adam-enition:
> Can you add tests for this?
> 
> also why do we add `!` in default endpoints since it won't affect anything

@himself65 

Yeah will do - converted this to a draft till I sort.

The default endpoint change just helps those that have both the EmailOTP and Captcha plugins enabled at the same time. By default these two plugins enabled together mean `/sign-in/email-otp` expects a captcha token to be present because it's matched against the existing `/sign-in/email` endpoint.

gustavovalverde:
Closing in favor of #9597 (with #9596 already merged). #9596 fixed the concrete email-OTP captcha bypass on main; #9597 then replaces the substring-based matching with exact + wildcard matching (`/sign-in/*`, `/sign-in/**`), supersedes the `!`-prefix exclusion syntax here, and hardens duplicate-slash bypass attacks.

PR #8654, opened 2026-03-17
fix(organization): member delete hooks in organization plugin

### LINKED ISSUE #8653: Bug: `databaseHooks.member.delete.after` is never called when removing a member via the organization plugin

## Description

The `databaseHooks` API accepts a `member.delete.after` hook, but it is silently never invoked when a member is removed through the organization plugin's `removeMember` endpoint. No error is thrown — it fails silently.

## Steps to Reproduce

```ts
betterAuth({
  databaseHooks: {
    member: {
      delete: {
        after: async (member) => {
          console.log("member deleted", member); // never logged
        },
      },
    },
  },
  plugins: [organization()],
});
```

Call `authClient.organization.removeMember(...)`. The `after` hook never fires.

## Expected Behavior

`databaseHooks.member.delete.after` fires after a member is deleted, consistent with how `user` and `session` hooks behave.

## Actual Behavior

The hook is never called. No error is thrown.

1.5.3

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:34 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T8112",
    "release": "25.2.0",
    "cpuCount": 8,
    "cpuModel": "Apple M2",
    "totalMemory": "16.00 GB",
    "freeMemory": "0.48 GB"
  },
  "node": {
    "version": "v25.7.0",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.10"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null,
    "error": "ENOENT: no such file or directory, open 'package.json'"
  }
}
```

Backend

```typescript

```

## Potential Root Cause

`plugins/organization/adapter.mjs` → `deleteMember()` calls `adapter.delete()` directly (the raw adapter), bypassing the hook-aware `deleteWithHooks()` wrapper in `db/with-hooks.mjs`. All other hooks that work (e.g. `user.create.after`, `session.create.before`) go through the hooked adapter path.

Relevant line in `adapter.mjs`:

```js
const member = await adapter.delete({   // ← raw adapter, hooks never run
  model: "member",
  where: [{ field: "id", value: memberId }]
});
```

This should use the `deleteWithHooks` equivalent so registered `databaseHooks` are respected consistently.

## Workaround

Use the `organizationHooks.afterRemoveMember` callback on the `organization()` plugin instead, which is correctly wired up and documented:

```ts
organization({
  organizationHooks: {
    afterRemoveMember: async ({ member }) => {
      // your logic here
    },
  },
}),
```
However, this workaround has a notable limitation: the callback receives no database client. If the side effect requires database access (e.g. updating sessions), the only option is to close over an externally constructed client, which is not obvious from the API and forces users to manage their own db connection alongside better-auth's internal one:

```ts
// User must construct and manage their own prisma instance just to use this hook

organization({
  organizationHooks: {
    afterRemoveMember: async ({ member }) => {
      await prisma.sessions.updateMany({ ... }); // no db client provided by better-auth
    },
  },
}),
```

Ideally the hook would receive a `db` or `context` argument similar to how other parts of the better-auth API expose the internal adapter.

GautamBytes:
## Summary

Fix organization member deletion to use the shared database hook pipeline so `databaseHooks.member.delete` runs correctly.

## Changes

- route organization member deletion through `internalAdapter.deleteWithHooks`
- allow delete hook lookup for plugin models like `member`
- preserve team-member cleanup inside the custom delete flow
- return `BAD_REQUEST` when `member.delete.before` blocks deletion
- cover `removeMember`, veto behavior, and `leaveOrganization` with regression tests

Fixes #8653 

---
## Summary by cubic
Fix organization member deletion to reliably run the shared DB hook pipeline, so `member.delete.before/after` fire for both remove and leave flows with veto support. Keeps team cleanup intact and returns a clear error when deletion is blocked.

- **Bug Fixes**
  - Exposed `internalAdapter.deleteWithHooks` on `InternalAdapter` and enabled delete hooks for plugin models like `member`.
  - Hardened `deleteWithHooks`: attempt adapter delete even if lookup fails, pass the correct entity to `delete.after` (from lookup or custom delete), and return `null` on veto/no-op.
  - Updated `deleteMember` to use a hook-aware custom delete that also removes team memberships; ensured hook context includes `internalAdapter` via `runWithAdapter` when no store is present.
  - Added guards in `removeMember` and `leaveOrganization` to return `BAD_REQUEST` when deletion is vetoed or a no-op.
  - Added tests for remove, veto, leave, and verifying `member.delete.after` receives the deleted entity and sees `internalAdapter` in context.

<sup>Written for commit 126ed465526797c762c33bed95aae53718ec62eb. Summary will update on new commits.</sup>


GautamBytes:
Hey @ping-maxwell , just pinging for review!

bytaesu:
The linked issue has already been closed after @ping-maxwell  left a comment, so I'll close this PR as well.

Thanks for working on this, and sorry for the delay 🙏

PR #8708, opened 2026-03-20
fix: prevent revoked sessions from being restored via database fallback

bytaesu:
> [!NOTE]
> ### `preserveSessionInDatabase` is intended for tracking, not authentication.
> 
> docs
> 
> https://better-auth.com/docs/concepts/session-management#preserving-sessions

When both `preserveSessionInDatabase` and `storeSessionInDatabase` are enabled, falling back to the database when a session is missing in secondary storage is not appropriate. It may improve resilience if secondary storage is down, but it goes beyond its intended purpose.


bytaesu:
=> updating

PR #8711, opened 2026-03-20
fix(two-factor): prevent unverified TOTP enrollment from gating sign-in

### LINKED ISSUE #8627: twoFactor plugin skips verification if account is setup to use OTA

### To Reproduce

The `twoFactor` ([two factor authentication](https://better-auth.com/docs/plugins/2fa#enabling-2fa)) plugin has two kinds of tokens that it can generate: `OTA` and `TOTP`.

`OTP` tokens can be sent to the user via SMS or Email, whereas `TOTP` tokens are generated using an authenticator app such as Google Authenticator, Microsoft Authenticator, 1Password, etc.

To enable the `twoFactor` plugin, you must set `twoFactorEnabled` on a user. If you want to use `OTP`, then you actually cannot use `auth.api.enableTwoFactor` because this begins the flow for `TOTP`. In order to enforce `OTP`, you must call `updateUser` directly and set `twoFactorEnabled: true` without generating a `TOTP` for the user (so i.e. `twoFactorEnabled` is set on the user, but no associated row is created in the `two_factor` table with their `TOTP` secret).

Once two factor authentication is enabled for a user:

1. Calling `auth.api.signInEmail` will now return `twoFactorRedirect` instead of an actual session. When this `twoFactorRedirect` is present in the response, the application should redirect to a page that handles MFA. For example:
```ts
		let signInEmailResult;
		try {
			// KEEP: Authenticate with database
			signInEmailResult = await locals.auth.api.signInEmail({
				body: {
					email: email.data,
					password: passwordParam,
					rememberMe: true,
					callbackURL: '/settings'
				},
				headers: request.headers
			});
		} catch (ex) {
			if (ex instanceof APIError) {
				// KEEP: Handle authentication errors thrown as exceptions
				return fail(422, {
					error: ex.name,
					description: ex.message
				});
			} else {
				// KEEP: Fall through
				console.error(ex);
			}
		}

		if (!signInEmailResult) {
			return fail(422, {
				error: 'internal_error',
				description: 'Please try again later'
			});
		}

		// KEEP: Two factor authentication may not be required if device is already trusted
		if (
			!('twoFactorRedirect' in signInEmailResult) ||
			signInEmailResult.twoFactorRedirect !== true
		) {
			throw redirect(
				303,
				signInEmailResult?.redirect && signInEmailResult?.url ? signInEmailResult.url : '/settings'
			);
		}

		// KEEP: Two factor authentication is required:
		// KEEP: Check whether to use OTP or TOTP for authenticating this user
		let twoFactorRecord;
		try {
			const ctx = await locals.auth.$context;

			// KEEP: Lookup user identifier for two factor cookie
			const userId = await getTwoFactorUserId(cookies, ctx);

			if (!userId) {
				return fail(422, {
					error: 'internal_error',
					description: 'Please try again later (2)'
				});
			}

			// KEEP: Look up whether the user has created any TOTP codes
			twoFactorRecord = await ctx.adapter.findOne({
				model: 'twoFactor',
				select: ['id'],
				where: [{ field: 'userId', value: userId }]
			});
		} catch (ex) {
			// KEEP: Send to observability
			console.error(ex);
		}

		// KEEP: Redirect to two factor verification
		const twoFactorMethod = twoFactorRecord ? 'totp' : 'otp';
		throw redirect(303, `/login/mfa/${twoFactorMethod}`);
```

2. If the user is using `OTP` for two factor authentication, then you call `auth.api.sendTwoFactorOTP` and redirect to another page that uses `auth.api.verifyTwoFactorOTP`. When calling `auth.api.verifyTwoFactorOTP` with the correct OTP, then the session becomes valid.

But if this user is using `TOTP` (that is, if they have a row in the `two_factor` table and they've generated a valid `TOTP` secret in the past and verified their `TOTP` generator by supplying a valid code) then using `auth.api.verifyTOTP` directly with a valid code from the authenticator app will cause the session to become valid.

3. If a user is using `OTP`, then once they are logged-in they can call `auth.api.enableTwoFactor` which will generate a `TOTP` secret and link it to their account.

**But here lies the issue:**

The `twoFactor` plugin uses the `twoFactorEnabled` flag on the account as a heuristic to track whether the `TOTP` has been verified during the `TOTP` enrolment flow.

So if you enforce `OTP` verification for an account by setting `twoFactorEnabled: true` on that account, and then allow the user to later add a `TOTP` token to their account, then the very moment that they add a `TOTP` code to their account then it automatically becomes valid without verification. That is, the entire Better Auth verification flow is ignored and the `TOTP` becomes instantly active.

This more or less means that the user would get locked out of their account if they do not add the `TOTP` secret to their authenticator app at that very moment it is shown to them, for example if they close the browser window or change their mind.

1. If the application is enforcing `OTP` on every login by setting `twoFactorEnabled: true` on the user, then this should continue to work.
2. If a user who is already using `OTP` wants to enroll in `TOTP` instead, then they should be able to go through the `TOTP` enrolment flow.
3. A user who goes through the `TOTP` enrolment flow should be able to abandon the enrolment flow midway through (before final verification), and on their next login they should be prompted to enter an `OTP` rather than a `TOTP`, the latter causing them to be locked out of their account.

**Suggested fix:**

Currently the `TOTP` secret becomes verified, active and in-use if `twoFactorEnabled` is set on the user account. This assumption is fundamentally flawed, because the user could be using some other 2FA method (for example `OTP`).

Therefore it seems that whether or not the 2FA method is verified should be a property of the row created in the `two_factor` table.

I propose that the `verified: boolean` becomes a property of the specific `TOTP` that is generated. This is so that `twoFactorEnabled` in the user table only tracks whether 2FA is required to generate an active session, and whether or not a specific `TOTP` secret is verified or not is actually tracked under the `two_factor` table in the row for that specific `TOTP`.

This also enables the correct behaviour in the following two scenarios:
1. A user deletes their `TOTP` and correctly falls back to using `OTP` again, or
2. A user deletes their `TOTP` because it has been accidentally leaked. They create a new TOTP with a new secret, and by deleting the original `TOTP` then they are correctly sent back through the `TOTP` verification onboarding flow rather than having their second `TOTP` immediately marked as active (and potentially locking themselves out).

1.5.4

### System info

```bash
N/A
```

Backend

```typescript

```

### Additional context

(As an aside, you can make `OTA` authentication mandatory for all users using this database hook:)
```ts
		databaseHooks: {
			user: {
				create: {
					before: async (user: any) => {
						return {
							data: {
								...user,
								// KEEP: New users created with two factor authentication enabled by default
								twoFactorEnabled: true
							}
						};
					}
				}
			}
		}
```

See also https://github.com/better-auth/better-auth/issues/5738 and comments on https://github.com/better-auth/better-auth/pull/5739 .

aarmful:
closes #8627

the totp enrollment flow was using `user.twoFactorEnabled` to figure out if someone was setting up totp for the first time or signing in. that works fine normally, but if you already have otp enabled (so `twoFactorEnabled` is already `true`) and then start totp setup, the code thinks you're signing in and skips the verification step entirely. if you close the tab mid-enrollment, the unverified secret is now active and you're locked out.

the fix adds a `verified` column to the `twoFactor` table. new rows start unverified, get marked verified once you enter a valid code, and totp sign-in rejects unverified rows. backup codes and otp are unaffected and work as fallbacks during unfinished enrollment. re-enrollment (calling `enableTwoFactor` when totp is already verified) preserves `verified: true` so users are never locked out while rotating their secret.

adds a migration column `verified` (boolean, default `true`). existing rows automatically get `true` which is correct since they were already verified under the old flow. `enableTwoFactor` explicitly sets new rows to `false`.


himself65 (approved):
LGTM, but this looks a breaking change?

aarmful:
> LGTM, but this looks a breaking change?

good point. we should default to `true` instead of `false` in the schema. existing rows get `true` automatically, new rows from `enableTwoFactor` explicitly set `verified: false`

kvanbere:
What about cleaning up unverified TOTP? I suppose user can only make 1x TOTP, so no need?

Is it a security issue if someone begins the TOTP flow, writes down the secret, abandons the flow, and future calls to TOTP enable use the same secret? In that case should existing TOTP be expired and regenerated on second call to enable?

I don’t have answers just asking questions.

aarmful:
> What about cleaning up unverified TOTP? I suppose user can only make 1x TOTP, so no need?
> 
> Is it a security issue if someone begins the TOTP flow, writes down the secret, abandons the flow, and future calls to TOTP enable use the same secret? In that case should existing TOTP be expired and regenerated on second call to enable?
> 
> I don’t have answers just asking questions.

if a user abandons enrollment and tries again, `enableTwoFactor` overwrites the existing row. and the sign-in flow filters for `verified: true`, so an orphaned unverified row just gets ignored.

`enableTwoFactor` generates a new secret each time it's called, so re-enrolling overwrites the old row with a fresh secret. the abandoned one becomes dead.

the only real risk would be if the old unverified secret somehow got used before being overwritten, but since the sign-in flow filters for `verified: true`, it can't be. so i think we're fine as-is.

kvanbere:
Great, this also addresses a separate issue I was having where deleting an existing TOTP didn’t take you through enrolment when you create another TOTP again. I’m not sure if the secret was being regenerated again in that case or not previously, but now it should be. 👍

kvanbere:
Would love this in next release, this bug is such a blocker for us

packages/better-auth/src/plugins/two-factor/totp/index.ts
```diff
@@ -265,30 +269,42 @@ export const totp2fa = (options?: TOTPOptions | undefined) => {
...
+					model: twoFactorTable,
+					update: { verified: true },
+					where: [{ field: "id", value: twoFactor.id }],
+				});
+				if (!user.twoFactorEnabled) {
+					if (!session.session) {
+						throw APIError.from(
+							"BAD_REQUEST",
+							BASE_ERROR_CODES.FAILED_TO_CREATE_SESSION,
+						);
```
ping-maxwell: Perhaps a better error message is required here?

ping-maxwell (changes requested):
please update documentation, for example it needs to include the new `verified` field in the documentation schema.

aarmful:
> please update documentation, for example it needs to include the new `verified` field in the documentation schema.

sure, will get that done this morning

packages/better-auth/src/plugins/two-factor/totp/index.ts
aarmful: this shouldn't really happen in practice since the endpoint already requires a session. it was carried over from the existing code. happy to improve it but might be better as a separate cleanup.

ping-maxwell (dismissed):

ping-maxwell:
Changing schema does make this a breaking change, so will have to wait until next minor  release

kvanbere:
@ping-maxwell people might be interested to know ETA on next minor?

kvanbere:
@himself65 seems like this needed a minor minor version and would have been good for 1.6?

packages/better-auth/src/plugins/two-factor/totp/index.ts
```diff
@@ -237,14 +238,17 @@ export const totp2fa = (options?: TOTPOptions | undefined) => {
 			}
 			const { session, valid, invalid } = await verifyTwoFactor(ctx);
 			const user = session.user as UserWithTwoFactor;
+			const isSignIn = !session.session;
+			const whereClause: Where[] = [{ field: "userId", value: user.id }];
+			// During sign-in, only consider verified TOTP secrets.
+			// Unverified rows (from abandoned enrollment) must be ignored
+			// so the sign-in flow falls through to OTP instead.
+			if (isSignIn) {
+				whereClause.push({ field: "verified", value: true });
```
gustavovalverde: This query-level filter breaks MongoDB and other document stores. Existing documents won't have the `verified` field because they predate the schema change.

The MongoDB adapter emits a literal `{ verified: true }` match (no `$or`/`$exists` expansion), and the core adapter's read path (`transformOutput`) doesn't inject `defaultValue`.

Instead of filtering at the query level, fetch by `userId` and check in application code:
```ts
if (isSignIn && twoFactor?.verified === false) {
  // unverified enrollment → invisible to sign-in
}
```
This lets absent/null/true all pass (legacy-safe) while only rejecting `verified=false` (new enrollment rows).

Separately: `verifyBackupCode` in `backup-codes/index.ts:326` does the same `findOne` by `userId` without any `verified` guard; and thus backup codes from an abandoned enrollment still work at sign-in.

packages/better-auth/src/plugins/two-factor/totp/index.ts
```diff
@@ -265,30 +269,42 @@ export const totp2fa = (options?: TOTPOptions | undefined) => {
...
-					throw APIError.from(
-						"BAD_REQUEST",
-						BASE_ERROR_CODES.FAILED_TO_CREATE_SESSION,
+			// Enrollment mode: TOTP row exists but hasn't been verified yet.
+			// This covers both fresh TOTP setup (twoFactorEnabled=false) and
+			// adding TOTP to an OTP-only account (twoFactorEnabled=true).
+			if (!twoFactor.verified) {
+				await ctx.context.adapter.update({
+					model: twoFactorTable,
+					update: { verified: true },
```
gustavovalverde: This persists `verified=true` before the user/session operations below, and there's no transaction wrapping the sequence. If `updateUser` or `createSession` fails after this update, the retry will see `verified=true` and skip the entire `!twoFactor.verified` block; leaving the user with a verified secret but `twoFactorEnabled` still `false`.

The old code's gate (`!user.twoFactorEnabled`) was retry-safe because `updateUser` is what flips it. The simplest fix would be to move this update to after the session rotation succeeds, or separate the two concerns:
```ts
if (!twoFactor.verified) {
  // just mark verified — no other side effects
}
if (!user.twoFactorEnabled) {
  // enable 2FA + rotate session (retry-safe gate)
}
// then mark verified last, or keep this order
// but gate user promotion on user state, not row state
```

gustavovalverde (commented):
Two issues need addressing before merging

gustavovalverde (changes requested):

aarmful:
@gustavovalverde holding off on any further changes due to an issue with the `lint:dependencies` script on windows. same goes for changes in #8772 

let me know if you want anything changed at this point. i will take a look once i can commit properly

gustavovalverde:
@aarmful fix for that in the main branch

gustavovalverde:
@aarmful considering that I can edit this PR, I'll make the remaining changes to get it merged.

gustavovalverde (approved):

PR #8721, opened 2026-03-21
fix(credential): improve error  response for missing token and callbackURL

mrmovas:
Hello! This is my very first pull request. If there’s anything I can improve, please let me know. I’d be happy to contribute and be part of this project.

I was working on a project using BetterAuth to implement a password reset feature for users who forgot their passwords. While testing, I forgot to pass the callbackURL and received an `INVALID_TOKEN` error, which did not accurately reflect the actual issue.

## The solution I am suggesting:

The previous implementation checked whether either `token` or `callbackURL` was invalid:
```ts
if (!token || !callbackURL) {
	throw ctx.redirect(
		redirectError(ctx.context, callbackURL, { error: "INVALID_TOKEN" }),
	);
}
```
The problem with this logic is that it returns an `INVALID_TOKEN` error even when the issue is actually a missing `callbackURL`. As a result, the error code can be misleading.

### Changes:
- Updated the validation logic to return more accurate and specific error codes: 
  - `INVALID_TOKEN`
  - `INVALID_CALLBACK_URL`
  - `INVALID_TOKEN_AND_CALLBACK_URL`
  So the error code always reflects the actual failure.
- Added `INVALID_TOKEN_AND_CALLBACK_URL` to `BASE_ERROR_CODES`.
- Replaced the `redirectError` function with a direct `APIError` throw. Returning an HTML redirect response is less useful in this context compared to throwing a `BAD_REQUEST` with a `BASE_ERROR_CODE`, which is more appropriate and consistent with how other endpoints handle invalid input.
- The `redirectError` logic is left commented out in case maintainers prefer to keep the redirect-based behaviour instead of the `APIError` approach.

### Note on `redirectError`
If I’m not mistaken, the `redirectError` function is only used in `password.ts`. This is another reason I’ve commented it out, as I’m suggesting it could potentially be removed if it’s no longer needed.

---
## Summary by cubic
Improves password reset errors when `token` or `callbackURL` is missing and keeps the redirect flow. Also makes `redirectTo` required (string; supports relative paths) and updates the OpenAPI schema.

- **Bug Fixes**
  - Validate `token` and `callbackURL` independently; redirect with `INVALID_TOKEN`, `CALLBACK_URL_REQUIRED`, or `INVALID_TOKEN_AND_CALLBACK_URL`. Added docs and linked them in the error index and sidebar.
  - Require `redirectTo` as a string; removed a duplicate field. OpenAPI schema updated to mark `redirectTo` as required.

<sup>Written for commit 38c600d402ae2c054cbf9ab3ccceed2925d20b3e. Summary will update on new commits.</sup>


mrmovas:
I committed the suggestions mentioned by the copilots in the comments of this pull request.

- I removed the `APIError` throw logic and added the `redirectError` function again to avoid changing behaviour for consumers expecting a redirect flow, as suggested.
- Changed `INVALID_CALLBACK_URL` to `CALLBACK_URL_REQUIRED`
- Formatted so the code matches the project’s established style, as suggested.

I'd also like to mention that "INVALID_TOKEN", as well as the other error codes I added and used from the redirectError function, are used to create a URL in the HTML for reading about the error code:
```html
href="https://better-auth.com/docs/reference/errors/${encodeURIComponent(code)}?askai=${encodeURIComponent(`What does the error code ${code} mean?`)}"
```
However, these codes do not exist in the documentation, and visiting the generated URLs leads to a “not found” page.

ping-maxwell (changes requested):
We have specific error pages in our documentation which will require to be updated with these new changes.

mrmovas:
Added `callback_url_required.mdx` and `invalid_token_and_callback_url.mdx` error documentation files.

Also made `redirectTo` required in the `/request-password-reset` request body. Previously, omitting it would generate a reset URL with an empty `callbackURL` parameter:
```
http://localhost:4000/api/auth/reset-password/OpxPg2ok2Iy3o6qObd2vcNwz?callbackURL=
```

Visiting that URL would redirect the user to the HTML error page:
```
http://localhost:4000/api/auth/error?error=CALLBACK_URL_REQUIRED
```

Making the field required prevents this from happening.

ping-maxwell (changes requested):
Please fix CI.

You're also missing `sidebar_contents.tsx` file needing to include the new error pages.

PR #8725, opened 2026-03-21
feat(passkey): infer authenticator name from AAGUID with override hook

mburumaxwell:
This PR adds AAGUID-based passkey labeling during registration

When a user does not provide a name,  resolve the stored label using the developer override via `getAuthenticatorName` and fallback to the built-in best-effort provider suggestion from known AAGUID values

## Why
Some ecosystems surface passkey provider names from AAGUID metadata. This improves default naming UX without introducing new DB storage concerns.

## References
Passkey AAGUID source: https://github.com/passkeydeveloper/passkey-authenticator-aaguids
ASP.Net Core implementation: https://github.com/dotnet/aspnetcore/pull/65343
Related discussion: https://github.com/better-auth/better-auth/discussions/7912

### DIFF
--- a/packages/passkey/src/authenticator-metadata.ts
+++ b/packages/passkey/src/authenticator-metadata.ts
@@ -1,0 +1,33 @@
+// Known AAGUID values are based on:
+// - https://github.com/passkeydeveloper/passkey-authenticator-aaguids
+const KNOWN_AUTHENTICATOR_PROVIDERS: Record<string, string> = {
+	// Google Password Manager
+	"ea9b8d66-4d01-1d21-3ce4-b6b48cb575d4": "Google Password Manager",
+
+	// Apple Passwords / iCloud Keychain
+	"fbfc3007-154e-4ecc-8c0b-6e020557d7bd": "Apple Passwords",
+	"dd4ec289-e01d-41c9-bb89-70fa845d4bf2": "iCloud Keychain",
+
+	// Dashlane
+	"531126d6-e717-415c-9320-3d9aa6981239": "Dashlane",
+
+	// Microsoft Windows Hello
+	"08987058-cadc-4b81-b6e1-30de50dcbe96": "Windows Hello",
+	"9ddd1817-af5a-4672-a2b9-3e3dd95000a9": "Windows Hello",
+	"6028b017-b1d4-4c02-b4b3-afcdafc96bb2": "Windows Hello",
+
+	// 1Password
+	"bada5566-a7aa-401f-bd96-45619a55120d": "1Password",
+	"b5397571-8af2-4d30-9d48-eeb8eee6e9c6": "1Password",
+
+	// Bitwarden
+	"d548826e-79b4-db40-a3d8-11116f7e8349": "Bitwarden",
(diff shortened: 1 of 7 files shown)


packages/passkey/src/authenticator-metadata.ts
```diff
@@ -0,0 +1,33 @@
...
+	"6028b017-b1d4-4c02-b4b3-afcdafc96bb2": "Windows Hello",
+
+	// 1Password
+	"bada5566-a7aa-401f-bd96-45619a55120d": "1Password",
+	"b5397571-8af2-4d30-9d48-eeb8eee6e9c6": "1Password",
+
+	// Bitwarden
+	"d548826e-79b4-db40-a3d8-11116f7e8349": "Bitwarden",
+	"cc45f64e-52a2-451b-831a-4edd8022a202": "Bitwarden",
+};
```
NuroDev, line 26: Is it worth using the `passkey-authenticator-aaguids` package or adding a build script to fetch all the known aaguid's from [passkeydeveloper/passkey-authenticator-aaguids ](https://github.com/passkeydeveloper/passkey-authenticator-aaguids) instead?
Or is that an overkill list to add by default?

mburumaxwell: Seems like an overkill to me, we don't need the extra functionality it brings or the dependencies it has. The developer can still override the function to do anything they want. I think a periodic build script would be better. Would you want me to add that?

ping-maxwell: I think both are a little overkill, I recommend just keeping a list of comment authenticators, then in the docs show users how they can use a library to resolve this

ping-maxwell (dismissed):
the rest, LGTM

ping-maxwell:
@mburumaxwell if you could fix merge conflicts that would be great 🙌

mburumaxwell:
@ping-maxwell I have updated and it is now ready from my end.

mburumaxwell:
@ping-maxwell I have done another rebase on the latest changes. Do you ming checking if it is good to go?

gustavovalverde:
@mburumaxwell please update the changeset to a `patch` version and retarget to the `main` branch. And do the last copilot fix to get this merged

mburumaxwell:
@gustavovalverde this is done

gustavovalverde:
Thanks for pushing this forward, @mburumaxwell. #9927 takes a different architectural direction, so it supersedes this PR (it carries `Closes #8725` and will close this on merge).

The pivot: since the `aaguid` is already stored on every passkey and already returned by `listPasskeys`, #9927 resolves the provider name at read time through an exported `getAuthenticatorName` helper and a `commonAuthenticatorNames` map, instead of writing a guessed label into the `name` column at registration. That keeps `name` as the user's own label, lets the lookup improve without backfilling existing rows, and folds the server-side override into the existing `registration.afterVerification` hook rather than a new top-level option.

Your provider list carried over, re-checked against the canonical source (one AAGUID resolved to ToothPic rather than Bitwarden, and one 1Password entry was not in the source).

PR #8733, opened 2026-03-22
feat(client): add `hydrateSession` for SSR session hydration

bytaesu:
Since `createAuthClient` is created once at the module level, using `initialData` is a bit tricky.

So I'm thinking of solving this with `hydrateSession` method that runs after initialization.

### DIFF
--- a/docs/content/docs/concepts/client.mdx
+++ b/docs/content/docs/concepts/client.mdx
@@ -459,6 +459,46 @@ if(error?.code){
 }
 ```
 
+### SSR Hydration
+
+With SSR frameworks like Next.js, Nuxt, or SvelteKit, `useSession` fetches the session on the client, causing a loading flash. Use `hydrateSession` to pre-fill the session with server-fetched data so `useSession` returns data immediately:
+
+```tsx title="page.tsx (server component)"
+import { auth } from "@/lib/auth"
+import { headers } from "next/headers"
+import { SessionCard } from "./session-card"
+
+export default async function Page() {
+    const session = await auth.api.getSession({
+        headers: await headers(),
+    })
+    return <SessionCard initialSession={session} />
+}
+```
+
+```tsx title="session-card.tsx (client component)"
+"use client"
+
+import { authClient } from "@/lib/auth-client"
+
+type Props = {
+    initialSession: typeof authClient.$Infer.Session | null
+}
+
+export function SessionCard({ initialSession }: Props) {
+    authClient.hydrateSession(initialSession)
+    const { data, isPending } = authClient.useSession()
+
+    // isPending is immediately false, data is available on first render
+    // A background refetch still happens to ensure freshness
+}
+```
+
+<Callout>
+- `hydrateSession` is idempotent. It only sets data when the session is empty.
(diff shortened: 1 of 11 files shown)


enchart:
Will this functionality ever be implemented? This is one of the only things that's been holding me back from using Better Auth in a lot of my SSR-first SvelteKit projects for at least a year at this point.

bytaesu:
@enchart Alright, I'll finalize this 😁

docs/content/docs/concepts/client.mdx
```diff
@@ -459,6 +459,47 @@ if(error?.code){
...
+    const session = isPending ? initialSession : data
+
+    if (!session) return <p>Not signed in</p>
+    return <pre>{JSON.stringify(session, null, 2)}</pre>
+}
+```
+
+<Callout>
+  * Only the first non-null call to `hydrateSession` takes effect.
+  * Passing `null` is safe. It will be ignored and `useSession` will fetch as usual.
```
ping-maxwell, line 500: This example will briefly restore stale session after sign-out.

After sign-out, this flow occurs:

1. Sign-out clears the atom: `{ data: null, isPending: false }`, component shows data (`null`). Correct.
2. A signal triggers a session refetch. In `onRequest` (query.ts line 97), `isPending` is set to `currentValue.data === null ` -> `true`.
3. The component re-renders: isPending is true, so session = initialSession, stale session is briefly displayed.
4. Fetch completes with 401->  `data: null, isPending: false` -> correctly shows null again.

ping-maxwell: A safer pattern might be 
```ts
const session = isPending && !data ? initialSession : data
```

or simply
```ts
const session = data ?? (isPending ? initialSession : null)
```

ping-maxwell (dismissed):
Just documentation related changes

docs/content/docs/concepts/client.mdx
bytaesu: gated with `isPending && !isRefetching` in commit https://github.com/better-auth/better-auth/pull/8733/commits/29b6f151db7bce9c2d687fa58ac04e822de94eb7

both proposals still match the `{ data: null, isPending: true }` tuple during the post-signout refetch.

packages/better-auth/src/client/session-atom.ts
```diff
@@ -6,10 +6,28 @@ import type { AuthQueryAtom } from "./query";
 import { useAuthQuery } from "./query";
 import { createSessionRefreshManager } from "./session-refresh";
 
-export type SessionAtom = AuthQueryAtom<{
+export type SessionData = {
 	user: User;
 	session: Session;
-}>;
+};
```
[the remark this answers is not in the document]
bytaesu: This is out of scope for this PR

enchart:
I didn't expect it to be done so fast, thank you!

bytaesu:
> I didn't expect it to be done so fast, thank you!

I wanted some time to explore a better API, so we're rolling this out in 1.7-beta.
You'll be able to try it in the beta track 😁

PR #8772, opened 2026-03-25
feat(two-factor): include enabled 2fa methods in sign-in redirect response

### LINKED ISSUE #4101: The twoFactorRedirect does not specify TOTP or OTP

### To Reproduce

My user can enable TOTP then twoFactorRedirect will be true and I have to validate using the verifyTotp method. However, if the user want OTP always through SMS or Email I need to manually use sendOTP and use verifyOtp but how can I now from the login result what the user preference is. Of course I use the sendOTP as a fallback if the user don't have the authenticator app, but what if he don't use authenticator app and always wants the sendOTP to be done. How should we setup this

Able to know TOTP or OTP setup

1.x.x

### System info

```bash
-
```

Client

```typescript
  emailAndPassword: {  
    enabled: true
  },

});
```

### Additional context

_No response_

aarmful:
closes #4101

when a user with 2fa signs in, the redirect response used to only return `{ twoFactorRedirect: true }` with no way to know whether they had totp, otp, or both. frontends had to guess or do hacky workarounds to figure out which verification ui to show.

now the response includes `twoFactorMethods` (e.g. `["totp", "otp"]`), and the `onTwoFactorRedirect` client callback receives it as a context param.

---
## Summary by cubic
Include `twoFactorMethods` in the 2FA sign-in redirect and pass it to `onTwoFactorRedirect` so UIs can show the right flow. Methods are based on verified TOTP status and `otpOptions.sendOTP`, and exclude disabled or unverified providers.

- New Features
  - `twoFactor` in `better-auth` now returns `{ twoFactorRedirect: true, twoFactorMethods: [...] }`: adds `"totp"` only if the user has a verified TOTP secret and TOTP isn’t disabled; adds `"otp"` when `otpOptions.sendOTP` is set.
  - `twoFactorClient` calls `onTwoFactorRedirect({ twoFactorMethods })`. Docs updated; tests cover TOTP/OTP combinations, disabled-TOTP, and unverified-TOTP exclusion.

<sup>Written for commit 35356a2bc87f67c2a8cc9750534768e6c6630258. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/plugins/two-factor/client.ts
+++ b/packages/better-auth/src/plugins/two-factor/client.ts
@@ -18,8 +18,18 @@ export const twoFactorClient = (
 				/**
 				 * a redirect function to call if a user needs to verify
 				 * their two factor
+				 *
+				 * @param context.twoFactorMethods - The list of
+				 * enabled two factor providers (e.g. ["totp", "otp"]).
+				 * Use this to determine which 2FA UI to show.
 				 */
-				onTwoFactorRedirect?: () => void | Promise<void>;
+				onTwoFactorRedirect?: (context: {
+					/**
+					 * The list of enabled two factor providers
+					 * for the user (e.g. ["totp", "otp"]).
+					 */
+					twoFactorMethods?: string[];
+				}) => void | Promise<void>;
 		  }
 		| undefined,
 ) => {
@@ -51,7 +61,9 @@ export const twoFactorClient = (
 					async onSuccess(context) {
 						if (context.data?.twoFactorRedirect) {
 							if (options?.onTwoFactorRedirect) {
-								await options.onTwoFactorRedirect();
+								await options.onTwoFactorRedirect({
+									twoFactorMethods: context.data.twoFactorMethods,
+								});
 								return;
 							}
 
--- a/packages/better-auth/src/plugins/two-factor/index.ts
+++ b/packages/better-auth/src/plugins/two-factor/index.ts
@@ -495,16 +495,17 @@ export const twoFactor = <O extends TwoFactorOptions>(options?: O) => {
 						 * that the user actually has a secret stored.
 						 */
 						if (!options?.totpOptions?.disable) {
-							const userTotpSecret = await ctx.context.adapter.findOne({
-								model: opts.twoFactorTable,
-								where: [
-									{
-										field: "userId",
-										value: data.user.id,
-									},
-								],
-							});
-							if (userTotpSecret) {
+							const userTotpSecret =
+								await ctx.context.adapter.findOne<TwoFactorTable>({
+									model: opts.twoFactorTable,
+									where: [
+										{
+											field: "userId",
+											value: data.user.id,
+										},
+									],
+								});
+							if (userTotpSecret && userTotpSecret.verified !== false) {
 								twoFactorMethods.push("totp");
 							}
 						}
--- a/packages/better-auth/src/plugins/two-factor/index.ts
+++ b/packages/better-auth/src/plugins/two-factor/index.ts
@@ -488,10 +488,35 @@ export const twoFactor = <O extends TwoFactorOptions>(options?: O) => {
 							ctx.context.secret,
 							twoFactorCookie.attributes,
 						);
-						const twoFactorMethods = [
-							!options?.totpOptions?.disable && "totp",
-							!!options?.otpOptions?.sendOTP && "otp",
-						].filter(Boolean) as string[];
+						const twoFactorMethods: string[] = [];
+
+						/**
+						 * totp requires per-user setup, so we check
+						 * that the user actually has a secret stored.
+						 */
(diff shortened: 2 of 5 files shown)


kvanbere:
@aarmful quick question, does this return `totp` if the user does not have any created or verified rows in the `two_factor` table?

aarmful:
> @aarmful quick question, does this return `totp` if the user does not have any created or verified rows in the `two_factor` table?

we query the `twoFactor` table for the user before including `"totp"`. if there's no row, it won't be in the array. so a user with `twoFactorEnabled: true` but no totp secret just gets `["otp"]`, assuming otp is configured.

otp is the other way around, it's server-level so if `sendOTP` is configured any 2fa user can receive a code.

kvanbere:
@armful does this need to be updated again

aarmful:
> @armful does this need to be updated again

i'm on a trip. will have a look when i'm back.

kvanbere:
@himself65 seems like this needed a minor minor version and would have been good for 1.6?

packages/better-auth/src/plugins/two-factor/index.ts
```diff
@@ -434,8 +434,38 @@ export const twoFactor = <O extends TwoFactorOptions>(options?: O) => {
...
+									{
+										field: "userId",
+										value: data.user.id,
+									},
+								],
+							});
+							if (userTotpSecret) {
+								twoFactorMethods.push("totp");
+							}
+						}
```
gustavovalverde, line 511: This equates "row exists" with "TOTP is usable", but `skipVerificationOnEnable` breaks that assumption.

`enableTwoFactor` **always** creates the secret row, even when `skipVerificationOnEnable` is `true`; and in that flow it also sets `twoFactorEnabled: true` before any TOTP verification, so a user who enables 2FA intending to use OTP-only still gets `["totp", "otp"]`; but they never scanned the QR code.

The OTP-only test (~line 1724) simulates this by manually flipping `twoFactorEnabled` without creating a row, but the real `skipVerificationOnEnable` flow *does* create a row, so the test doesn’t catch it.

The `twoFactor` table has no `verified` column, so there’s no way to distinguish "row created during enable" from "row verified via `verifyTOTP`".

Possible fixes:
- Add a `verified` boolean to the `twoFactor` schema (it's the cleanest, but  requires a schema change)
- Only advertise TOTP when `skipVerificationOnEnable` is falsy (it's simpler, but imprecise; some skip-verification users *might* have set up their authenticator anyway)

Either way, add a test: config with `skipVerificationOnEnable: true` + TOTP not disabled + user enables 2FA but never calls `verifyTOTP` → should get `["otp"]` not `["totp", "otp"]`.

packages/better-auth/src/plugins/two-factor/index.ts
```diff
@@ -434,8 +434,38 @@ export const twoFactor = <O extends TwoFactorOptions>(options?: O) => {
 							ctx.context.secret,
 							twoFactorCookie.attributes,
 						);
+						const twoFactorMethods: string[] = [];
```
gustavovalverde, line 491: `string[]` is too loose. The plugin only ever pushes `"totp"` or `"otp"`. A union makes the contract explicit and gives consumers autocomplete:

```suggestion
						const twoFactorMethods: ("totp" | "otp")[] = [];
```

Worth defining `type TwoFactorMethod = "totp" | "otp"` in `types.ts` so both server and client share it.

packages/better-auth/src/plugins/two-factor/client.ts
```diff
@@ -17,8 +17,18 @@ export const twoFactorClient = (
...
+				 * enabled two factor providers (e.g. ["totp", "otp"]).
+				 * Use this to determine which 2FA UI to show.
 				 */
-				onTwoFactorRedirect?: () => void | Promise<void>;
+				onTwoFactorRedirect?: (context: {
+					/**
+					 * The list of enabled two factor providers
+					 * for the user (e.g. ["totp", "otp"]).
+					 */
+					twoFactorMethods?: string[];
```
gustavovalverde, line 31: Same; should match the server:

```suggestion
					twoFactorMethods?: ("totp" | "otp")[];
```

gustavovalverde (changes requested):

packages/better-auth/src/plugins/two-factor/index.ts
aarmful: > - Add a `verified` boolean to the `twoFactor` schema (it's the cleanest, but  requires a schema change)

there's a pr for this already.

i will address your comment later. not home at the moment. thank you!

packages/better-auth/src/plugins/two-factor/index.ts
aarmful: i agree.

gustavovalverde (approved):

PR #8785, opened 2026-03-26
feat(auth-provider): add support for zoho social auth provider

ruban-s:
## Summary by cubic
Adds a Zoho OAuth social provider with multi–data center support and offline access (refresh tokens) by default. Exports the provider from `@better-auth/core`, with docs and tests included.

- **New Features**
  - New `zoho` provider in `@better-auth/core`, configurable via `accountsServer` and `accessType` (defaults to "offline").
  - Defaults to `openid email profile` scopes; reads profile from `id_token` when present, with fallback to Zoho User Info API.
  - Provider added to `socialProviders` and re-exported; docs include setup, data center URLs, and sign-in example; tests cover auth URL and sign-in flow.

<sup>Written for commit f9dc57a8f5d501f71b63b3906713da8ed6b3635c. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/semantic-pull-request.yml
+++ b/.github/workflows/semantic-pull-request.yml
@@ -1,0 +1,54 @@
+name: Semantic Pull Request
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - reopened
+      - edited
+      - synchronize
+
+permissions:
+  pull-requests: write
+
+jobs:
+  validate:
+    name: Validate PR title
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


ping-maxwell:
Hey I know zoho interacts with many social providers but we can't accept every oauth provider as the library will at some point be too bloated with them. These days we intend to only support social providers to their truest meaning (eg Instagram, Facebook, etc)

We appriciate your contribution though! I recommend opening an issue asking if it makes sense to be implemented first next time

PR #8803, opened 2026-03-27
feat: enhance api-key rate limit system

### LINKED ISSUE #6035: API Key rate limiting not behaving as expected

### To Reproduce

We've implemented rate limiting on our app but it's not behaving how we'd expect. 

Because it's based on last_request_at it waits until the number of requests has been reached then will wait the time set and refresh

The [docs](https://www.better-auth.com/docs/plugins/api-key#rate-limiting)  are confusing as it make it appear that every day the user has 10 requests per day.  In reality they have 10 requests but have to wait a day from the last request for it to reset.

We're trying to implement 1000 per hour limit but it doesn't work with the current rate limit behaviour.  How do we get it so that it resets once per hour?

Expected that rate limit resets `rateLimitMax` to amount after `timeWindow` has elapsed from first request for that period.

1.3.7

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#35-Ubuntu SMP PREEMPT_DYNAMIC Sat Oct 11 10:06:31 UTC 2025",
    "release": "6.14.0-35-generic",
    "cpuCount": 2,
    "cpuModel": "Common KVM processor",
    "totalMemory": "7.75 GB",
    "freeMemory": "6.62 GB"
  },
  "node": {
    "version": "v22.21.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.4"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null
  }
}
```

Package

```typescript

```

### Additional context

_No response_

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/6035
## Summary

Fixes API key rate limiting so it matches documented “N requests per time window” expectations ([issue #6035](https://github.com/better-auth/better-auth/issues/6035)).

Previously, limiting used `lastRequest` updated on **every** successful validation, so the effective period slid with activity and `tryAgainIn` was tied to time since the **last** successful request. That differed from a **fixed window** starting at the first request in a period (and from “10 per day” style wording).

### Fixed-window algorithm

- Each limit is enforced over **`[rateLimitWindowStart, rateLimitWindowStart + rateLimitTimeWindow)`**.
- **`rateLimitWindowStart`** is set when a window begins and **does not** advance on each allowed request.
- **`requestCount`** increments within that window; when it reaches **`rateLimitMax`**, further validations return `RATE_LIMITED` until the window ends.
- **`tryAgainIn`** is the time remaining until **`rateLimitWindowStart + rateLimitTimeWindow`** (window end), not until “idle for `timeWindow` since last request.”
- Window rollover uses **`elapsed >= rateLimitTimeWindow`** so the boundary at exactly one window length is handled correctly (replacing the old strict `>` behavior on “time since last request”).

### New / updated data

- **`rateLimitWindowStart`** (`Date | null`) added to the API key model and plugin schema; `null` until the first validation in a new period (or after a reset).
- **`lastRequest`** remains for auditing and is still updated on allowed validations when rate limiting applies.
- **Secondary storage**: serialize/deserialize `rateLimitWindowStart` in the API key JSON cache.

### Update API key behavior

- Changing **`rateLimitEnabled`**, **`rateLimitTimeWindow`**, or **`rateLimitMax`** clears **`rateLimitWindowStart`** and resets **`requestCount`** to `0` so the next validation starts a fresh window under the new settings.

# Breaking change

- **Database**: add a nullable **`rateLimitWindowStart`** column (or equivalent) for the **`apikey`** table to match the plugin schema. Existing rows with `NULL` get a new window on the next successful verification.

## Migration / upgrade notes for users

1. Run your usual Better Auth migration flow so the schema includes **`rateLimitWindowStart`**.
2. Expect **different** rate-limit timing vs the old behavior (by design): limits are per **fixed window from window start**, not per rolling idle period from **`lastRequest`**.


xaviemirmon:
Any ideas when this will be merged? My team is following this issue to bring into our codebase.

ping-maxwell:
cc @bytaesu or @gustavovalverde

xaviemirmon:
Can we get an update in this issue? We can roll our own plugin but would rather keep this native to better-auth

xaviemirmon:
@ping-maxwell congrats on the 1.7 launch. Are there any plans to merge this in now that the conflicts have been resolved?

PR #8805, opened 2026-03-27
feat(sso)!: support multiple IdP signing certificates

OscarCornish:
SAML IdPs publish their signing certificates in metadata, and during a key rotation that metadata briefly lists both the outgoing and incoming PEM. Better Auth previously accepted only one certificate per provider. Any rotation forced an outage window where some users hit signature-verification failures, depending on which key the IdP picked for their response.

Both `samlConfig.cert` and `samlConfig.idpMetadata.cert` now accept `string | string[]`. samlify already supports the array shape, so the work here is the schema, the persisted config, and the sanitizer that returns parsed certs to admins. When both fields are set, `idpMetadata.cert` wins, matching the precedence the runtime callback path already used.

The management endpoints (`getSSOProvider`, `listSSOProviders`, `updateSSOProvider`) now return `samlConfig.certificate` as an array in every case, even when a single cert is configured. That removes the `Array.isArray` branching consumers used to need, at the cost of a one-time read-shape update.

---
## Summary by cubic
Adds rolling-cert support for SAML IdP signing keys, prefers `samlConfig.idpMetadata.cert` over `samlConfig.cert`, and makes management endpoints always return certificates as an array. Registration now fails when no cert source is provided, and encrypted SLO decryption is fixed.

- **New Features**
  - `samlConfig.cert` and `samlConfig.idpMetadata.cert` accept `string | string[]`; `idpMetadata.cert` wins. If using `idpMetadata.metadata` XML, you can omit explicit certs.
  - Registration rejects configs with no cert source (`CERT_SOURCE_MISSING`).
  - Management APIs return `samlConfig.certificate` as an array of parsed certificates (or per‑cert errors), or absent when certs are in `idpMetadata.metadata`; same precedence is used at runtime.
  - Shared schemas for OIDC/SAML centralize validation; removed the inline registration schema and redundant issuer URL checks.

- **Migration**
  - No breaking change to config: both `samlConfig.cert` and `samlConfig.idpMetadata.cert` are supported, and `samlConfig.cert` is optional when using metadata XML.
  - Client change: `samlConfig.certificate` in management responses is now always an array. Update consumers to read an array only.

<sup>Written for commit 5ded117fc69b530898eeba9ec1152fccfd9b2a75. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/8805?utm_source=github">Review in cubic</a></sup>


ping-maxwell (dismissed):

kvanbere:
Suggest you be careful here, there’s a lot of CVEs from last year around parsing and multi certificate handling in SAML, particularly because certificates can be nested in unintuitive ways
* https://github.blog/security/sign-in-as-anyone-bypassing-saml-sso-authentication-with-parser-differentials/ and
* https://nvd.nist.gov/vuln/detail/CVE-2025-47949

OscarCornish:
> Suggest you be careful here, there’s a lot of CVEs from last year around parsing and multi certificate handling in SAML, particularly because certificates can be nested in unintuitive ways
> 
> * https://github.blog/security/sign-in-as-anyone-bypassing-saml-sso-authentication-with-parser-differentials/ and
> * https://nvd.nist.gov/vuln/detail/CVE-2025-47949

Thanks for flagging these - but I don't think this PR is adding to the potential attack surface here

The referenced CVEs relate to improper XML parsing and libraries that extract certificates from incoming SAML responses for verification

In our case, the signing certificates are not extracted from the response (they are in db), This PR just allows one of a pool of cetificates to be used (normally for key rotation) - the code path in samlify that actually parses and verifies SAML responses is unchanged

matsoob:
Checking if there are any update on this?

packages/sso/src/routes/schemas.ts
```diff
@@ -42,14 +42,14 @@ const oidcConfigSchema = z.object({
 
 const samlConfigSchema = z.object({
 	entryPoint: z.string().url().optional(),
-	cert: z.string().optional(),
+	cert: z.union([z.string(), z.array(z.string()).nonempty()]).optional(),
```
gustavovalverde: Schema widened here, but the create-endpoint inline body schema at `packages/sso/src/routes/sso.ts:284` (top-level `cert`) and `:295` (`idpMetadata.cert`) is still `z.string()`. So providers can only be registered with a single cert via the API, then *updated* to add more; which is inconsistent and could be surprising. Same `z.union([z.string(), z.array(z.string()).nonempty()])` should apply there.

packages/sso/src/types.ts
```diff
@@ -50,7 +50,7 @@ export interface SAMLConfig {
 				entityID?: string;
 				entityURL?: string;
 				redirectURL?: string;
-				cert?: string;
+				cert?: string | string[];
```
gustavovalverde, line 78: cubic previously flagged: `idpMetadata.cert` is now `string | string[]` in the type, but the create-endpoint inline schema at `packages/sso/src/routes/sso.ts:295` still validates as `z.string()`. Type and wire-validation diverge.

packages/sso/src/routes/providers.ts
```diff
@@ -142,7 +142,14 @@ function sanitizeProvider(
...
 						try {
-							return parseCertificate(samlConfig.cert);
+							const certs = Array.isArray(samlConfig.cert)
+								? samlConfig.cert
+								: [samlConfig.cert];
+							if (certs.length === 0) {
+								return { error: "No certificates provided" };
+							}
+							const parsed = certs.map((cert) => parseCertificate(cert));
+							return parsed.length === 1 ? parsed[0] : parsed;
```
gustavovalverde: Returning `object | object[]` based on count makes every consumer branch on `Array.isArray`. It is worth to always return an array; it's a cleaner contract, less branching for callers, and the PR is already changing the shape so the cost is the same. If you keep the polymorphic form for byte-on-the-wire reasons, document it on the `getSSOProvider` OpenAPI response.

packages/sso/src/routes/providers.ts
```diff
@@ -142,7 +142,14 @@ function sanitizeProvider(
 					digestAlgorithm: samlConfig.digestAlgorithm,
 					certificate: (() => {
 						try {
-							return parseCertificate(samlConfig.cert);
+							const certs = Array.isArray(samlConfig.cert)
+								? samlConfig.cert
+								: [samlConfig.cert];
+							if (certs.length === 0) {
```
gustavovalverde: This is unreachable from schema-validated input; `nonempty()` covers both arrays in `samlConfigSchema`. Defensible as DB defense-in-depth (a hand-edited row with `cert: []`), but `{ error: "No certificates provided" }` diverges from the surrounding `{ error: "Failed to parse certificate" }` contract. Easier on consumers to `throw` inside the `try` and reuse the existing `catch` so the response shape stays uniform.

packages/sso/src/providers.test.ts
```diff
@@ -755,6 +798,90 @@ describe("SSO provider read endpoints", () => {
 			).toBe("Failed to parse certificate");
 		});
 
+		it("should parse multiple signing certificates", async () => {
```
gustavovalverde: These verify parsing and shape, not rotation. The feature value is "an assertion signed by VALID_CERT_2 verifies because both certs are trusted"; so it's worth at least one test that signs a response with the second cert and checks `parseLoginResponse` accepts it through `createIdP`. Without that, samlify could regress (or the array could fail to thread through `signingCert`) and these tests would stay green. `saml.test.ts` already has the signing building blocks.

packages/sso/src/providers.test.ts
```diff
@@ -10,6 +10,49 @@ import { ssoClient } from "./client";
 const TEST_CERT = `MIIDXTCCAkWgAwIBAgIJAJC1HiIAZAiUMA0Gcm9markup
 temporary cert for testing`;
 
+const VALID_CERT_1 = `-----BEGIN CERTIFICATE-----
```
gustavovalverde: Nit: three large PEM blocks at the top of the file is a lot of scrollback before reaching the tests. Either co-locate these inside the new describe-block, or move `TEST_CERT` plus both `VALID_CERT_*` into a `__fixtures__` section.

.changeset/heavy-planes-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@better-auth/sso": patch
```
gustavovalverde: The `getSSOProvider`/`updateSSOProvider` response now returns `samlConfig.certificate` as `object | object[]`, and inputs accept arrays; both consumer-visible widenings. A one-line note about the response shape would save consumers debugging time.

gustavovalverde (changes requested):
A few gaps and blockers

packages/sso/src/providers.test.ts
OscarCornish: Good point, Added a test to check that response validation uses the correct certificate, regardless of index.

packages/sso/src/providers.test.ts
OscarCornish: Couldn't see any patterns for fixtures in the repo, so I've stuck these in the describe block. If you could point me to an example of fixtures I'd be happy change it

packages/sso/src/routes/providers.ts
OscarCornish: I do agree that always returning an array would be preferable, but it makes this a breaking change - I think my recommendation / preferred solution is to make this change now, and in the next version change consolidate on the array return?

Currently, all existing SAML providers are going to be using just a single signing certificate, as we can't support a second one. So most users will be able to ignore the array case without consequence.

I'm happy to help make this change, either as part of this PR or as part of a later one - It will be a quite a big increase the surface area of the change though, which is why I've chosen not to do it here (and I'd like to get this merged in ASAP, as it is blocking my better-auth migration). If you'd like me to consolidate on the array return here, please let me know.

packages/sso/src/saml.test.ts
```diff
@@ -1156,6 +1178,89 @@ describe("SAML SSO", async () => {
...
+		});
+		let samlResponse: any;
+		await betterFetch(signInResponse?.url as string, {
+			onSuccess: async (context) => {
+				samlResponse = await context.data;
+			},
+		});
+		let redirectLocation = "";
+		await betterFetch(
+			"http://localhost:8081/api/sso/saml2/callback/saml-provider-multi-cert",
```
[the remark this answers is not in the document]
gustavovalverde: cc: @OscarCornish

docs/content/docs/guides/saml-sso-with-okta.mdx
```diff
@@ -1,7 +1,8 @@
----
+***
```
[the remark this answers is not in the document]
gustavovalverde: Resolved by the merge — next's docs are now in place and both `.mdx` files use proper `---` frontmatter delimiters.

packages/sso/src/routes/sso.ts
```diff
@@ -2158,9 +2048,7 @@ export const callbackSSOSAML = (options?: SSOOptions) => {
 			} catch (error) {
 				ctx.context.logger.error("SAML response validation failed", {
 					error,
-					decodedResponse: Buffer.from(SAMLResponse, "base64").toString(
-						"utf-8",
-					),
+					decodedResponse: decodeBase64ToUtf8(SAMLResponse),
```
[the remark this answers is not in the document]
gustavovalverde: Resolved by the merge — the Buffer→Uint8Array change that introduced `decodeBase64ToUtf8`/`atob` was reverted in favor of next's decode path. `grep` confirms no `atob` references remain in `packages/sso/src/routes/sso.ts`.

packages/sso/src/routes/providers.ts
gustavovalverde: Done in badbca568: `sanitizeSigningCerts` now always returns `SanitizedCert[]` (or undefined). OpenAPI descriptions on the management endpoints updated to reflect the array-only shape.

packages/sso/src/routes/providers.ts
gustavovalverde: The polymorphic shape is gone in badbca568 and the empty-array branch with it; the `parsed.length === 1 ? parsed[0] : parsed` line was deleted along with the divergent error string.

packages/sso/src/providers.test.ts
gustavovalverde: Restored in badbca568: parameterized `it.each([['bad-then-good', ...], ['good-then-bad', ...]])` test signs the SAML response with `mockIdpSigningCert` and asserts `auth.api.acsEndpoint` accepts it regardless of cert position in the rotation array. Goes through better-auth's ACS endpoint, not the mock IdP.

packages/sso/src/providers.test.ts
gustavovalverde: Moved the two rotation PEMs into the `sanitization` describe block in badbca568. `TEST_CERT` stays at module scope since it's used across describe blocks.

packages/sso/src/utils.ts
```diff
@@ -1,6 +1,17 @@
...
+ * that work in Node, Bun, Deno, and Cloudflare Workers (no `Buffer`).
+ */
+export function decodeBase64ToUtf8(value: string): string {
+	const binary = atob(value);
+	const bytes = new Uint8Array(binary.length);
+	for (let i = 0; i < binary.length; i++) bytes[i] = binary.charCodeAt(i);
+	return new TextDecoder("utf-8").decode(bytes);
+}
+
 /**
```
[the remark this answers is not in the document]
gustavovalverde: Removed `decodeBase64ToUtf8` in badbca568 — the helper was dead since the Buffer→Uint8Array experiment got reverted during the merge.

docs/content/docs/plugins/sso.mdx
```diff
@@ -967,6 +967,30 @@ You also need to provide Identity Provider (IdP) configuration:
...
+            "-----BEGIN CERTIFICATE-----\n<current>\n-----END CERTIFICATE-----",
+            "-----BEGIN CERTIFICATE-----\n<next>\n-----END CERTIFICATE-----",
+        ],
+    },
+}
+```
+
+`idpMetadata.cert` accepts a single PEM string or an array. The top-level `cert` field on `samlConfig` is also accepted for backwards compatibility; when both are set, `idpMetadata.cert` wins.
+
+If your IdP publishes a metadata XML document, pass it as `idpMetadata.metadata` instead and leave `cert` unset. The certificates come from the document and rotation happens when the IdP republishes it. A registration that supplies neither `cert` nor `idpMetadata.metadata` is rejected at registration with the `SAML_CERT_SOURCE_MISSING` error code: samlify has nothing to verify SAML responses against.
```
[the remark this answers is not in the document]
gustavovalverde: Fixed in badbca568 — docs now reference `CERT_SOURCE_MISSING`.

docs/content/docs/plugins/sso.mdx
```diff
@@ -1,7 +1,8 @@
----
+***
+
 title: Single Sign-On (SSO)
 description: Integrate Single Sign-On (SSO) with your application.
----
+------------------------------------------------------------------
```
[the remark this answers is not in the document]
gustavovalverde: Resolved by the merge — sso.mdx now uses proper `---` frontmatter delimiters.

docs/content/docs/guides/saml-sso-with-okta.mdx
```diff
@@ -1,7 +1,8 @@
...
+
 title: SAML SSO with Okta
 description: A guide to integrating SAML Single Sign-On (SSO) with Better Auth, featuring Okta
----
+----------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
gustavovalverde: Resolved by the merge — saml-sso-with-okta.mdx now uses proper `---` frontmatter delimiters.

.changeset/heavy-planes-remain.md
```diff
@@ -0,0 +1,27 @@
...
+        cert: [currentPem, nextPem],
+    },
+}
+```
+
+Both `samlConfig.cert` and `samlConfig.idpMetadata.cert` accept either a single PEM string or an array. When both are set, `idpMetadata.cert` wins.
+
+**Validation**
+
+Registration now rejects SAML configs that supply no signing-cert source. samlify needs either an `idpMetadata.metadata` XML document (which embeds the certs) or an explicit PEM under `cert` or `idpMetadata.cert`. Configs missing both fail with `SAML_CERT_SOURCE_MISSING`.
```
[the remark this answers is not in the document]
gustavovalverde: Fixed in badbca568 — changeset now references `CERT_SOURCE_MISSING`.

gustavovalverde (approved):

PR #8808, opened 2026-03-28
chore: replace `canary` branch references with `main`

bytaesu:
Release-related CI and scripts were intentionally left as-is. It's better to handle them together when we reorganize the release process.

cc @gustavovalverde


ping-maxwell (dismissed):
Did we remove the canary version thing in our docs?

bytaesu:
@ping-maxwell Oh we also need to remove the branch switcher. I'll update this.

Issue #7043, opened 2025-12-28
Admin client plugin only exposes checkRolePermission method - other methods missing from types


yusoofsh:
### Description

The admin client plugin (`adminClient`) only exposes the `checkRolePermission` method in its TypeScript type definitions. All other admin methods such as `listUsers`, `createUser`, `setRole`, `banUser`, `unbanUser`, `revokeUserSessions`, `impersonateUser`, and `stopImpersonating` are not available on the `auth.admin` object according to TypeScript, even though they should be available based on the plugin's functionality.

### Current Status

**Impact:** Blocking - TypeScript compilation fails when trying to use any admin methods other than `checkRolePermission`.

### Expected Status

All admin plugin methods should be properly typed and accessible on the `auth.admin` object when using the `adminClient` plugin, including:
- `listUsers`
- `createUser`
- `setRole`
- `banUser`
- `unbanUser`
- `revokeUserSessions`
- `impersonateUser`
- `stopImpersonating`

### Environment

- **better-auth version:** 1.4.9
- **Plugin configuration:**
  - Server: `admin()` plugin enabled in `betterAuth` config
  - Client: `adminClient()` plugin enabled in `createAuthClient` config

### Code Example

**Client setup (src/lib/auth/client.ts):**
```typescript
import { adminClient } from "better-auth/client/plugins"
import { createAuthClient } from "better-auth/react"

export const auth = createAuthClient({
  plugins: [adminClient()],
})
```

**Usage attempt (src/routes/dashboard/admin/users.tsx:27):**
```typescript
const data = await auth.admin.listUsers({
  query: {
    limit: 10,
    sortBy: "createdAt",
    sortDirection: "desc",
  },
})
```

TypeScript error: `Property 'listUsers' does not exist on type '{ checkRolePermission: ...`

### Possible Solutions

**Verified Option:**
- The admin plugin server endpoints exist and are properly defined in `packages/better-auth/src/plugins/admin/routes.ts`
- The client plugin type inference may not be correctly mapping the server plugin's endpoints to the client API

**Speculative Option:**
- The `$InferServerPlugin` mechanism in `adminClient` may not be properly inferring all endpoint methods from the server `admin` plugin
- Type generation or path-to-object mapping for admin routes may need adjustment

himself65:
I believe the issue is not a valid concern; we do have a test case that includes all the admin APIs on the client side. And that place you mentioned is not for exposing the type.

[...]

Issue #3015, opened 2025-06-13
Property 'createUser' does not exist


sovetski:
I am using this: `await auth.api.createUser` to create initial users with a custom script on my website, it is like `seeds` but only for users. I am running this script at the beginning of the project, and once the project is live for the first time, to create my initial admin accounts.

This code works

```ts
await auth.api.createUser({
    body: {
        name: "Name",
        email: "blabla",
        password: "blabla",
        role: "blabla",
    },
});
```

But it gives this typescript error:

`Property 'createUser' does not exist on type 'InferAPI<{ ok: { <AsResponse extends boolean = false, ReturnHeaders extends boolean = false>(inputCtx_0?: ({ body?: undefined; } & { method?: "GET" | undefined; } & { query?: Record<string, any> | undefined; } & { params?: Record<string, any> | undefined; } & { ...; } & { ...; } & { ...; } & { ...; }) | undefined): ...'`

And I looked at the documentation, but no where we have any information on how to create users, without using [this admin plugin](https://www.better-auth.com/docs/plugins/admin)

Kinfe123:
`createUser` is part of the admin plugin. The TypeScript error you're encountering happens because the types from the admin plugin aren't being picked up in your setup and thats why the createUser is doesn't exist. so you have to use the admin plugin for that. for seeding you can do write a script which first creates the admin and then creates a users.

jpainam:
I have a slighly similar question, `createUser`, `setUserPassword` all exist in the `admin` plugin but only work from `authClient`
Is it possible to `create/update` from `auth.api`?

![Image](https://github.com/user-attachments/assets/de59c041-8af6-427f-8e42-5dea1179a533)

Kinfe123:
this is valid if you are using admin plugin.

```ts
auth.api.createUser({
  body: {
    email: "test@test.com",
    password: "test",
    name: "Test User",
  }
})
```

jpainam:
> this is valid if you are using admin plugin.
> 
> auth.api.createUser({
>   body: {
>     email: "test@test.com",
>     password: "test",
>     name: "Test User",
>   }
> })

See my screenshot above. Can’t find createUser on the auth.api. But on authClient.admin

I’m using admin plugin both in the client and server

Kinfe123:
can you please share your server auth config ?

jpainam:
`auth.ts`
```tsx
const config = {
    database: prismaAdapter(db, {
      provider: "postgresql",
    }),
   
[...]

PR #8824, opened 2026-03-29
fix(types): preserve plugin endpoint types on server-side auth.api

### LINKED ISSUE #8823: fix(types): server auth.api loses plugin endpoint types in lazy singleton / factory patterns
### Describe the bug

When using `betterAuth()` inside a factory function (lazy singleton pattern), `auth.api` loses TypeScript types for plugin-contributed endpoints. For example, `auth.api.createUser()` from the `admin()` plugin works at runtime but TypeScript reports:

```
Property 'createUser' does not exist on type 'InferAPI<...>'
```

This forces `(auth.api as any).createUser(...)` workarounds, losing type safety for all admin plugin methods (`createUser`, `removeUser`, `listUsers`, `setRole`, `banUser`, etc.).

### Affected area(s)

Types

### To Reproduce

```typescript

// Common pattern: lazy singleton for Nitro SSR / serverless
function createAuth() {
  return betterAuth({
    database: "...",
    plugins: [admin()],
  });
}

type Auth = ReturnType<typeof createAuth>;

// ❌ TypeScript error — but works at runtime
declare const auth: Auth;
auth.api.createUser({ body: { email: "test@test.com", password: "test", name: "Test", role: "user" } });
//        ^^^^^^^^^^  Property 'createUser' does not exist on type 'InferAPI<...>'
```

The same issue occurs when:
- Using `ReturnType<typeof betterAuth>` as a type annotation
- Storing options in a variable before passing to `betterAuth()`
- Combining plugins with different shapes (e.g., `[admin(), tanstackStartCookies()]`)

### Expected behavior

`auth.api.createUser`, `auth.api.removeUser`, `auth.api.listUsers`, etc. should be properly typed regardless of whether `betterAuth()` is called inline or inside a factory function.

### Root cause

The `PluginEndpoint` type in `getEndpoints()` uses `Array<infer T>` which doesn't match `ReadonlyArray`. When `betterAuth()` uses a `const` generic or the caller stores options with `as const`, the plugin array becomes `readonly`, and endpoint type extraction silently fails (falls through to `{}`).

Additionally, all plugin type extraction utilities (`InferPluginTypes`, `InferPluginErrorCodes`, `InferPluginIDs`, `InferPluginContext`, `InferDBFieldsFromPlugins/Input`) only handle `Array<infer T>`, not `ReadonlyArray<infer T>`.

### Better Auth Version

1.5.6 (also affects earlier versions — same type definitions)

### Additional context

- Related: #3015 (same symptom, closed without resolution)
- Related: #7043 (client-side variant)
- PR with fix: coming

jmalmo:
## Summary

Fixes server-side `auth.api` losing plugin-contributed endpoint types (e.g., `createUser`, `removeUser` from `admin()`) when `betterAuth()` is used in a lazy singleton / factory pattern.

**Root cause:** All plugin type extraction utilities use `Array<infer T>` which silently fails on `ReadonlyArray`. Combined with no `const` generic on `betterAuth()`, TypeScript can widen the plugin tuple and erase endpoint types.

**Changes:**
- Accept `readonly BetterAuthPlugin[]` in `BetterAuthOptions.plugins`
- Handle `ReadonlyArray<infer T>` in all server-side plugin type extraction:
  - `PluginEndpoint` (in `getEndpoints`)
  - `InferPluginTypes`, `InferPluginErrorCodes`, `InferPluginIDs`, `InferPluginContext`
  - `InferDBFieldsFromPluginsInput`, `InferDBFieldsFromPlugins` (also updated tuple destructuring to `readonly [infer P, ...infer Rest]`)
  - `InferPluginID`, `InferPluginOptions` (in `context.ts` — fixes `hasPlugin`/`getPlugin` inference)
- Add `const` type parameter to `betterAuth()` (full + minimal) for narrower inference
- Add type regression tests: lazy singleton, mixed plugins, `as const` readonly, and `$context` hasPlugin/getPlugin

**Files changed (9):**
- `packages/core/src/types/init-options.ts` — accept readonly plugins
- `packages/core/src/types/context.ts` — `ReadonlyArray` in `InferPluginID` + `InferPluginOptions`
- `packages/core/src/db/type.ts` — readonly constraint + tuple destructuring in `InferDBFieldsFromPlugins*`
- `packages/better-auth/src/auth/full.ts` — `const` generic
- `packages/better-auth/src/auth/minimal.ts` — `const` generic
- `packages/better-auth/src/api/index.ts` — `ReadonlyArray` in `PluginEndpoint`
- `packages/better-auth/src/types/models.ts` — `ReadonlyArray` in `InferPluginTypes`
- `packages/better-auth/src/types/plugins.ts` — `ReadonlyArray` in 3 utility types
- `packages/better-auth/src/types/types.test.ts` — 4 type regression tests

**Known limitations:**
- **`satisfies BetterAuthOptions` pattern:** If options are contextually typed via `satisfies` before reaching `betterAuth()`, plugins may already be widened to `BetterAuthPlugin[]`. The `const` generic cannot recover types that were already erased. Inline options or `as const` work correctly.
- **Client-side consumers:** `client/types.ts` and `client/plugins/infer-plugin.ts` also use `Array<infer P>`. Those are left out of scope for this server-side fix. Happy to expand or open a follow-up PR for #7043.
- **TS version:** The `const` type parameter requires TypeScript 5.0+. The repo uses TS ^5.9.3. If Better Auth needs to support TS 4.x consumers, this would need a different approach.

Fixes #8823
Refs #3015

### DIFF
--- a/packages/better-auth/src/api/index.ts
+++ b/packages/better-auth/src/api/index.ts
@@ -183,7 +183,7 @@ export function getEndpoints<Option extends BetterAuthOptions>(
 		}, {}) ?? {};
 
 	type PluginEndpoint = UnionToIntersection<
-		Option["plugins"] extends Array<infer T>
+		Option["plugins"] extends Array<infer T> | ReadonlyArray<infer T>
 			? T extends BetterAuthPlugin
 				? T extends {
 						endpoints: infer E;
--- a/packages/better-auth/src/auth/full.ts
+++ b/packages/better-auth/src/auth/full.ts
@@ -24,7 +24,7 @@ import { createBetterAuth } from "./base";
  *	  database: drizzleAdapter(db, { provider: "pg" }),
  * });
  */
-export const betterAuth = <Options extends BetterAuthOptions>(
+export const betterAuth = <const Options extends BetterAuthOptions>(
 	options: Options & {},
 ): Auth<Options> => {
 	return createBetterAuth(options, init);
--- a/packages/better-auth/src/auth/minimal.ts
+++ b/packages/better-auth/src/auth/minimal.ts
@@ -8,7 +8,7 @@ export type { BetterAuthOptions };
 /**
  * Better Auth initializer for minimal mode (without Kysely)
  */
-export const betterAuth = <Options extends BetterAuthOptions>(
+export const betterAuth = <const Options extends BetterAuthOptions>(
 	options: Options & {},
 ): Auth<Options> => {
 	return createBetterAuth(options, initMinimal);
--- a/packages/better-auth/src/types/models.ts
+++ b/packages/better-auth/src/types/models.ts
@@ -13,16 +13,17 @@ export type AdditionalSessionFieldsInput<Options extends BetterAuthOptions> =
 	InferDBFieldsFromPluginsInput<"session", Options["plugins"]> &
 		InferDBFieldsFromOptionsInput<Options["session"]>;
 
-export type InferPluginTypes<O extends BetterAuthOptions> =
-	O["plugins"] extends Array<infer P>
-		? UnionToIntersection<
-				P extends BetterAuthPlugin
-					? P["$Infer"] extends Record<string, any>
-						? P["$Infer"]
-						: {}
+export type InferPluginTypes<O extends BetterAuthOptions> = O["plugins"] extends
+	| Array<infer P>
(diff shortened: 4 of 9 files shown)


bytaesu:
Hi @jmalmo,

Closing this in favor of #9466

PR #8839, opened 2026-03-30
fix(credential): apply enumeration protection when autoSignIn is false

### LINKED ISSUE #8832: Email Enumeration Protection Not Applied When autoSignIn is false and requireEmailVerification is false or unset

### To Reproduce

1. Create and configure `betterAuth` on server with `emailAndPassword` configured using `autoSignIn: false` and either `requireEmailVerification: false` or left unset.
2. Sign up with an existing user.
3. API Returns a 422.

Documentation states that email enumeration protection is enabled if `requireEmailVerification: true` is set OR `autoSignIn: false` is set. https://better-auth.com/docs/authentication/email-password#email-enumeration-protection

API Returns 422, I expected a 200 with a synthetic user.

1.5.6

### System info

```bash
Doesn't seem necessary to include this - see additional context.
```

Backend, Documentation

```typescript
  emailAndPassword: {  
    enabled: true,
    autoSignIn: false,
    requireEmailVerification: false, // or unset
  },
});
```

### Additional context

It looks like this is probably the affected line of code. It is only checking for `requireEmailVerification` in this block where a user is found, rather than also considering `autoSignIn`. I've linked this for v1.5.6, but it looks like it's the same line in `main` currently.

https://github.com/better-auth/better-auth/blob/v1.5.6/packages/better-auth/src/api/routes/sign-up.ts#L251

dipan-ck:
## What does this PR do?

Fixes email enumeration protection not being applied when `autoSignIn` 
is set to `false` and `requireEmailVerification` is not set.

## Why?

The documentation states that email enumeration protection is enabled 
when either `requireEmailVerification: true` OR `autoSignIn: false` is 
set. However the code only checked `requireEmailVerification`, ignoring 
`autoSignIn`.

This caused the endpoint to return a 422 instead of a synthetic 200 
response when a duplicate email was used — leaking whether an email 
is registered.

## Changes

- Updated `shouldReturnGenericDuplicateResponse` in `sign-up.ts` to 
  also check `autoSignIn === false`

## Related Issue

Fixes #8832

---
## Summary by cubic
Apply email enumeration protection when `emailAndPassword.autoSignIn` is false in `better-auth`. Duplicate sign-ups return a synthetic user (`token: null`) and trigger `onExistingUserSignUp`, and new sign-ups skip auto sign-in (`token: null`) even without `requireEmailVerification`, aligning with the docs.

<sup>Written for commit d05a52d6a6f997d840d1d01c34ba0b45a4593368. Summary will update on new commits.</sup>


ping-maxwell (dismissed):

ping-maxwell:
please add changeset @dipan-ck

ping-maxwell (approved):

PR #8856, opened 2026-03-31
docs: rework AI resources documentation

ping-maxwell:


landing/components/sidebar-content.tsx
```diff
@@ -2718,6 +2702,111 @@ C0.7,239.6,62.1,0.5,62.2,0.4c0,0,54,13.8,119.9,30.8S302.1,62,302.2,62c0.2,0,0.2,
 			},
 		],
 	},
+	{
+		title: "AI Resources",
```
bytaesu, line 2730: Should we match this with the URL?

-> AI Tools

ping-maxwell: Will fix.

PR #8877, opened 2026-04-01
feat(email): add sendOldEmailVerification option to skip old email confirmation

### LINKED ISSUE #3742: Configurable Email-Address Change Verification Flow

ruban-s:
Closes #3742

  ## Summary

  When users change their email, the current flow sends a confirmation to the old email before verifying the new one. Users who lost access to their old email get stuck and have to contact support.

  This adds `sendOldEmailVerification` option under `emailVerification` config. When set to `false`, the verification email is sent directly to the new email address, skipping the old email
  confirmation step.

  - Password or fresh session is required when old email verification is disabled to compensate for the reduced security
  - Default behavior (`sendOldEmailVerification: true`) is unchanged
  - Adds `afterEmailChange` hook and `sendChangeEmailNotification` handler for post-change lifecycle events

  ## Usage

  ```ts
    user: {
      changeEmail: { enabled: true },
    },
    emailVerification: {
      sendOldEmailVerification: false,
      sendVerificationEmail: async ({ user, url }) => {
        // sends directly to the new email
      },
    },
  })

  Test plan

  - Skip old email with correct password — verification sent to new email
  - Wrong password rejected with INVALID_PASSWORD
  - Fresh session without password — allowed
  - Stale session without password — rejected with SESSION_EXPIRED
  - Default two-step flow preserved when option is not set
  - No email existence leak on duplicate email (timing attack prevention)
  - afterEmailChange and sendChangeEmailNotification hooks fire after verification

---
## Summary by cubic
Adds an option to verify the new email directly and skip old-email confirmation, unblocking users who lost access to their old address while keeping safeguards in place. Aligns with Linear #3742.

- **New Features**
  - `emailVerification.sendOldEmailVerification` (default `true`). When `false`, send verification to the new address and skip the old-email confirmation step (and `sendChangeEmailConfirmation`).
  - Security: require a correct `password` or a fresh session when skipping; wrong password → `INVALID_PASSWORD`, stale session → `PASSWORD_OR_FRESH_SESSION_REQUIRED`.
  - API: `changeEmail` accepts optional `password`; default two-step flow unchanged when the option is not set; OpenAPI updated with the `password` field; no email-existence leak.
  - Hook: `user.changeEmail.afterEmailChange` runs after the new email is verified and applied with `oldEmail` and `newEmail` (called from `verifyEmail`); docs updated with examples and a security callout.

<sup>Written for commit 3d028c457370b5bbd1d1dcecc1f725c174c25030. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/8877?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>


packages/better-auth/src/api/routes/update-user.ts
```diff
@@ -825,7 +832,46 @@ export const changeEmail = createAuthEndpoint(
...
+					ctx.context.session.session.createdAt,
+				).getTime();
+				const freshAge = ctx.context.sessionConfig.freshAge * 1000;
+				if (Date.now() - createdAt >= freshAge) {
+					throw APIError.from("BAD_REQUEST", BASE_ERROR_CODES.SESSION_EXPIRED);
+				}
+			}
+		}
+
+		if (canSendConfirmation && !skipOldEmailVerification) {
```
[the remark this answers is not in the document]
ruban-s: This is a false positive. When sendOldEmailVerification is false, you'd always have sendVerificationEmail configured — that's the whole point. So canSendVerification is truthy and the early guard at line 756 lets it
   through just fine. And the existing-email check on line 767 still returns { status: true } either way, so there's no enumeration leak.

docs/content/docs/concepts/users-accounts.mdx
```diff
@@ -83,6 +83,91 @@ export const auth = betterAuth({
 })
 ```
 
+#### Skipping Current Email Verification
+
+If your users may no longer have access to their old email address, you can skip the confirmation step on the current email by setting `sendOldEmailVerification` to `false`. The verification email will be sent directly to the new email address instead.
```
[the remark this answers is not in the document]
ruban-s: Not a conflict — the existing section documents the default flow (old email gets the confirmation), and the new section is specifically about opting out of that with sendOldEmailVerification: false. They're         
  describing two different configurations, not contradicting each other.

packages/core/src/types/init-options.ts
```diff
@@ -800,6 +812,39 @@ export type BetterAuthOptions = {
 					 * @default false
 					 */
 					updateEmailWithoutVerification?: boolean;
+					/**
```
ping-maxwell, line 983: perhaps both of these should be combined into one, they're effectively called together so there isn't much difference.
Just call it like what you have now `afterEmailChange` or `onEmailChange` and merge the JSDoc

packages/better-auth/src/api/routes/update-user.ts
```diff
@@ -677,6 +677,13 @@ export const changeEmail = createAuthEndpoint(
 					description: "The URL to redirect to after email verification",
 				})
 				.optional(),
+			password: z
+				.string()
+				.meta({
+					description:
+						"The user's password. Required when sendOldEmailVerification is disabled and session is not fresh",
+				})
+				.optional(),
```
ping-maxwell, line 691: make sure to find the documentation about the change-email endpoint and add this field into the docs of it if not done already

packages/better-auth/src/api/routes/update-user.ts
```diff
@@ -825,7 +832,46 @@ export const changeEmail = createAuthEndpoint(
...
+				if (!verify) {
+					throw APIError.from("BAD_REQUEST", BASE_ERROR_CODES.INVALID_PASSWORD);
+				}
+			} else if (ctx.context.sessionConfig.freshAge !== 0) {
+				const createdAt = new Date(
+					ctx.context.session.session.createdAt,
+				).getTime();
+				const freshAge = ctx.context.sessionConfig.freshAge * 1000;
+				if (Date.now() - createdAt >= freshAge) {
+					throw APIError.from("BAD_REQUEST", BASE_ERROR_CODES.SESSION_EXPIRED);
```
ping-maxwell: I would choose another error code or add a new one, a non-fresh session doesn't mean it's expired

packages/better-auth/src/api/routes/update-user.ts
```diff
@@ -825,7 +832,46 @@ export const changeEmail = createAuthEndpoint(
...
+				const accounts = await ctx.context.internalAdapter.findAccounts(
+					ctx.context.session.user.id,
+				);
+				const account = accounts.find(
+					(account) => account.providerId === "credential" && account.password,
+				);
+				if (!account || !account.password) {
+					throw APIError.from(
+						"BAD_REQUEST",
+						BASE_ERROR_CODES.CREDENTIAL_ACCOUNT_NOT_FOUND,
```
ping-maxwell: change this to INVALID_PASSWORD to prevent account enumeration

ping-maxwell (changes requested):

packages/better-auth/src/api/routes/update-user.ts
ruban-s: can i use SESSION_NOT_FRESH error code  that already defined in codes.ts

ruban-s: or i can create new error code named PASSWORD_OR_FRESH_SESSION_REQUIRED.

ruban-s: i created an new error code named PASSWORD_OR_FRESH_SESSION_REQUIRED.

ruban-s (commented):
All changes are completed @ping-maxwell

ping-maxwell (changes requested):
Changes look good, just missing changeset changlogs

packages/better-auth/src/api/routes/update-user.ts
```diff
@@ -825,7 +832,46 @@ export const changeEmail = createAuthEndpoint(
...
+				if (Date.now() - createdAt >= freshAge) {
+					throw APIError.from(
+						"BAD_REQUEST",
+						BASE_ERROR_CODES.PASSWORD_OR_FRESH_SESSION_REQUIRED,
+					);
+				}
+			}
+		}
+
+		if (canSendConfirmation && !skipOldEmailVerification) {
```
ping-maxwell, line 879: when `skipOldEmailVerification` is `true`, the code only validates the password/fresh session and then falls through. There's no code path that actually sends the verification email to the new email address when skipping the old email verification.

My recommendation is to add explicit code path for when skipOldEmailVerification is true AND email is verified to send verification to the new email directly.

ping-maxwell (changes requested):

packages/better-auth/src/api/routes/update-user.ts
ruban-s: The code path does exist — when `skipOldEmailVerification` is true, the condition at line 874 (`canSendConfirmation && !skipOldEmailVerification`) evaluates to false, so it falls through to line 910 which sends verification to the **new** email via `sendVerificationEmail`.

Added a JSDoc comment above that block to make the intent explicit so reviewers don't have to trace the fall-through.

PR #8913, opened 2026-04-02
fix(drizzle): disable boolean and dates for sqlite

### LINKED ISSUE #8882: [Bug]: Drizzle adapter + SQLite fails on INSERT — missing supportsBooleans/supportsDates flags
## Describe the bug

When using the Drizzle adapter with `provider: "sqlite"`, any INSERT that includes a boolean or Date value fails with:

```
TypeError: SQLite3 can only bind numbers, strings, bigints, buffers, and null
```

This affects basic operations like sign-up and sign-in since they insert `emailVerified: false` and date fields.

## Reproduction

```ts
import { drizzle } from "drizzle-orm/better-sqlite3";
import Database from "better-sqlite3";
import * as schema from "./schema";

const sqlite = new Database("test.db");
const db = drizzle(sqlite);

    provider: "sqlite",
    schema,
  }),
});
```

Attempting to sign up or sign in throws the TypeError above.

## Root cause

In `packages/drizzle-adapter/src/drizzle-adapter.ts` (line ~849), the adapter config sets `supportsJSON` and `supportsArrays` to `false` for non-pg providers, but does not set `supportsBooleans` or `supportsDates`:

```ts
supportsUUIDs: config.provider === "pg" ? true : false,
supportsJSON: config.provider === "pg" ? true : false,
supportsArrays: config.provider === "pg" ? true : false,
```

The core adapter factory already handles the conversions when these flags are `false` — `true`/`false` → `1`/`0` and `Date` → ISO string — but the Drizzle adapter never sets the flags for SQLite.

## Suggested fix

Add two lines alongside the existing flags:

```ts
supportsBooleans: config.provider !== "sqlite",
supportsDates: config.provider !== "sqlite",
```

## Related

- #1460 — same underlying issue, closed with maintainer suggesting the `supportsBooleans`/`supportsDates` approach
- The Kysely adapter (raw `better-sqlite3` path) handles this correctly via its own transformation layer

## Environment

- `better-auth`: 1.5.6
- `@better-auth/drizzle-adapter`: 1.5.6
- `drizzle-orm`: 0.44.2
- `better-sqlite3`: 11.10.0
- Node.js: 22

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/8882

### DIFF
--- a/packages/drizzle-adapter/src/drizzle-adapter.ts
+++ b/packages/drizzle-adapter/src/drizzle-adapter.ts
@@ -851,6 +851,8 @@ export const drizzleAdapter = (db: DB, config: DrizzleAdapterConfig) => {
 				config.provider === "pg" // even though mysql also supports it, mysql requires to pass stringified json anyway.
 					? true
 					: false,
+			supportsBooleans: config.provider !== "sqlite",
+			supportsDates: config.provider !== "sqlite",
 			supportsArrays: config.provider === "pg" ? true : false,
 			customTransformOutput: ({ data, fieldAttributes }) => {
 				// not all providers support dates


Bekacru:
Is this still valid @ping-maxwell if not lets close it

Issue #8906, opened 2026-04-02
Link-social callback should look up existing accounts by providerId


skirdey:
### What happened?

In `better-auth@1.5.6`, the social account link callback does a provider-agnostic lookup:

```js
const existingAccount = await c.context.internalAdapter.findAccount(String(userInfo.id));
```

That lives in `dist/api/routes/callback.mjs` in the `if (link) { ... }` branch.

The internal adapter already exposes a provider-aware helper:

```js
findAccountByProviderId(accountId, providerId)
```

Using `findAccount(accountId)` here means the link flow can match an existing account row from the wrong provider if two providers produce the same external subject/account id string. In that case the callback can update the wrong linked account row instead of the provider-specific one.

### Expected behavior

The link-social callback should look up the existing linked account by both `accountId` and `providerId`.

### Proposed fix

Replace:

```js
const existingAccount = await c.context.internalAdapter.findAccount(String(userInfo.id));
```

with:

```js
const existingAccount = await c.context.internalAdapter.findAccountByProviderId(String(userInfo.id), provider.id);
```

### Additional context

We verified this is still present in the latest npm release (`better-auth@1.5.6`) and are carrying the above one-line patch downstream to avoid provider-agnostic updates during account linking.

If helpful, I can open a PR with the exact change.

jaydeep-pipaliya:
Hi, I'd like to work on this — could you assign it to me? I've already reproduced the bug and have a fix with a regression test ready.

PR #8917, opened 2026-04-02
fix(auth): use provider-specific lookup in link-social callbak

SandipGyawali:
- replaced findAccount() with findAccountByProviderId() in Oauth
- prevents two providers return the same external accountiD

Issue: #8906

---
## Summary by cubic
Switch the OAuth link-social callback in `better-auth` to use provider-scoped account lookup to prevent ID collisions across providers. Replace `findAccount(...)` with `findAccountByProviderId(id, provider.id)` to ensure the correct user is linked; addresses #8906.

<sup>Written for commit fbcfc9c728cc62cffe72ec38e92d87010b8303e0. Summary will update on new commits.</sup>


bytaesu (changes requested):
Could you add regression test code?

gustavovalverde:
Solved in #8983

PR #8925, opened 2026-04-02
[codex] Fix provider-aware social link lookup

### LINKED ISSUE #8906: Link-social callback should look up existing accounts by providerId

skirdey:
## Summary

- use a provider-aware lookup in the social link callback so matching is scoped by both `accountId` and `providerId`
- add a focused regression test that links Google and GitHub accounts sharing the same external id and asserts both linked rows are preserved

## Root Cause

The link callback was calling `findAccount(String(userInfo.id))`, which searches only by external account id. When two different providers return the same external id, the callback could update the wrong linked account row instead of creating or updating the provider-specific row.

## Impact

- prevents cross-provider account collisions during social account linking
- keeps provider-specific linked accounts intact when external ids overlap

## Validation

- `pnpm build`
- `pnpm --filter better-auth test --run src/api/routes/callback-link-account.test.ts`

Closes #8906

---
## Summary by cubic
Fixes social account linking by looking up linked accounts by both external id and provider, preventing cross-provider collisions. Adds a regression test to ensure Google and GitHub accounts with the same external id are stored as separate links.

- **Bug Fixes**
  - Replaced `findAccount` with `findAccountByProviderId(accountId, providerId)` in `packages/better-auth/src/api/routes/callback.ts`.
  - Added `callback-link-account.test.ts` to verify both Google and GitHub links are preserved when sharing the same external id.

<sup>Written for commit d0ceaf932505e4fff529476a485a678c13091c90. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/callback-link-account.test.ts
+++ b/packages/better-auth/src/api/routes/callback-link-account.test.ts
@@ -1,0 +1,247 @@
+import type {
+	GithubProfile,
+	GoogleProfile,
+} from "@better-auth/core/social-providers";
+import { HttpResponse, http } from "msw";
+import { setupServer } from "msw/node";
+import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
+import { parseSetCookieHeader } from "../../cookies";
+import { signJWT } from "../../crypto";
+import { getTestInstance } from "../../test-utils/test-instance";
+import { DEFAULT_SECRET } from "../../utils/constants";
+
+const sharedAccountId = "shared-provider-account-id";
+const sharedEmail = "test@test.com";
+
+const githubProfile: GithubProfile = {
+	login: "test-user",
+	id: sharedAccountId,
+	node_id: "node-id",
+	avatar_url: "https://avatars.githubusercontent.com/u/12345",
+	gravatar_id: "",
+	url: "https://api.github.com/users/test-user",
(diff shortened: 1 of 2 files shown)


bytaesu:
Duplicate #8917

PR #8953, opened 2026-04-04
fix(expo): read cached session data from SecureStore on app startup

### LINKED ISSUE #8952: Session data written to SecureStore is never read back — causes login screen flash on app restart

### Reproduction

1. Create an Expo app with `@better-auth/expo` and `expo-secure-store`
2. Sign in (any method — email/password, social, etc.)
3. Confirm session works and `_session_data` is written to SecureStore
4. Close the app completely (kill from app switcher)
5. Reopen the app
6. Observe `useSession()` returns `{ data: null, isPending: true }` for 1-2 seconds while the network request to `/get-session` completes

**Current behavior:**

The Expo client writes session data to SecureStore after every successful `/get-session` response:

```ts
// packages/expo/src/client.ts — onSuccess hook
if (context.request.url.toString().includes("/get-session") && !opts?.disableCache) {
  const data = context.data;
  storage.setItem(localCacheName, JSON.stringify(data)); // ← WRITE
}
```

However, `localCacheName` (`${storagePrefix}_session_data`) is **never read back** from storage. The only reads from SecureStore are for `cookieName` (`${storagePrefix}_cookie`), which holds the cookie jar — not the session data.

This means the session atom always initializes as `{ data: null, isPending: true }` on every app start, regardless of whether valid cached session data exists in SecureStore. Apps that use `useSession()` for auth-based routing (e.g., with Expo Router's `Stack.Protected`) will briefly flash the login screen until the `/get-session` network request completes.

**Expected behavior:**

On app startup (non-web), the client should read `localCacheName` from SecureStore and use it as the initial/optimistic value for the session atom while the `/get-session` network request is in flight. This is the standard stale-while-revalidate pattern — show cached data immediately, then replace it once the fresh response arrives.

A minimal fix could look like:

```ts
// In getActions or during plugin initialization:
if (!isWeb && !opts?.disableCache) {
  const cached = storage.getItem(localCacheName); // SecureStore.getItem is synchronous on native
  if (cached && cached !== "{}") {
    const parsed = safeJSONParse(cached);
    if (parsed?.session?.id && parsed?.user?.id) {
      store.atoms.session?.set({
        data: parsed,
        error: null,
        isPending: true, // still pending — network will confirm
      });
    }
  }
}
```

This would eliminate the login screen flash for returning users while still validating the session server-side.

1.5.6

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.4.0: Thu Mar 19 19:30:44 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T6000",
    "release": "25.4.0",
    "cpuCount": 10,
    "cpuModel": "Apple M1 Max",
    "totalMemory": "32.00 GB",
    "freeMemory": "1.50 GB"
  },
  "node": {
    "version": "v24.14.1",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.11"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null
  }
}
```

Client

```typescript

```

### Additional context

**Environment information**
```
@better-auth/expo: 1.5.6
better-auth: 1.5.6
expo: ~53.0.11
expo-secure-store: ~15.0.2
React Native: 0.79.2
Platform: iOS & Android
```

**Workaround:** We currently work around this by reading the SecureStore cache ourselves at module load time and using it as a fallback in the auth guard:

```ts
// session-cache.ts
import { Platform } from "react-native";

const CACHE_KEY = `${storagePrefix}_session_data`;

export function getCachedSession() {
  if (Platform.OS === "web") return null;
  try {
    const raw = SecureStore.getItem(CACHE_KEY);
    if (!raw || raw === "{}") return null;
    const parsed = JSON.parse(raw);
    if (parsed?.session?.id && parsed?.user?.id) return parsed;
    return null;
  } catch {
    return null;
  }
}
```

```ts
// _layout.tsx — auth guard
const cachedSession = getCachedSession();
// ...
const effectiveSession = session ?? cachedSession;
const hasSession = Boolean(effectiveSession?.user?.id) || isPending;
```

This works but is fragile because it depends on the internal cache key format (`${storagePrefix}_session_data`). Ideally the library itself would handle this.

terijaki:
Closes #8952

## Problem

The Expo client writes session data to SecureStore (`${storagePrefix}_session_data`) after every successful `/get-session` response, but never reads it back on startup. This causes the session atom to always initialize as `{ data: null, isPending: true }`, resulting in a login screen flash for returning users until the `/get-session` network request completes.

## Solution

Read the cached session data from SecureStore in `getActions` and hydrate the session atom with it as an optimistic value (stale-while-revalidate). The cached data is validated before use:

- Must contain valid `session.id` and `user.id`
- Must not be expired (checks `session.expiresAt`)
- Respects `disableCache` option
- Skipped on web (SecureStore is native-only)

The atom is set with `isPending: true` so the `/get-session` network request still runs and replaces the cached data with the fresh server response.

## Tests

Added 4 test cases covering:
- Hydration from valid cached session data
- No hydration when cache is empty
- No hydration when cached session is expired
- No hydration when `disableCache` is `true`

---
## Summary by cubic
Hydrates the Expo session on app startup by reading cached data from `SecureStore`, preventing the login screen flash for returning users. Uses the cache optimistically and replaces it with the next `/get-session` response.

- **Bug Fixes**
  - Validate cache: require `session.id` and `user.id`; treat missing/invalid/expired `session.expiresAt` as expired.
  - Preserve additional user/session fields through the cache round-trip.
  - Respect `disableCache`; skip on web.
  - Set `isPending: true` for stale-while-revalidate.
  - Tests for valid, empty, expired/missing/invalid-`expiresAt`, disabled-cache, and additional-fields cases.

- **Refactors**
  - Reuse `Session` and `User` types from `@better-auth/core/db`.

<sup>Written for commit 0cb0cea4eec98a05663e5e2d4a299b6efe3f7b75. Summary will update on new commits.</sup>


bytaesu:
Hi @terijaki,

I'll push some changes!

bytaesu (approved):
@terijaki Thanks for working on this!

I tested with actual Expo app locally, and it looks good to me 😁

PR #8971, opened 2026-04-05
fix(mysql): handle race conditions for concurrent inserts

### LINKED ISSUE #8867: MySQL inserted id
Noticed the following code that relies on finding the inserted row based on the largest id, which could fail in a high load scenario when another new row is inserted before the select finishes. I believe the inserted id is returned when using MySQL via the `insertId` member on the response, so this could be improved.

https://github.com/better-auth/better-auth/blob/85e37693a2766c7560c50d7ab675adfba8c681a2/packages/kysely-adapter/src/kysely-adapter.ts#L113-L144

howlettt:
Use the returned ID from MySQL when available to avoid race conditions for concurrent inserts.
Took the test from the closed PR https://github.com/better-auth/better-auth/pull/6634

Fixes #8867

---
## Summary by cubic
Fixes MySQL concurrent insert race conditions by re-reading the inserted row via its returned ID in `@better-auth/kysely-adapter` and `@better-auth/drizzle-adapter`. Adds a test that creates 20 users in parallel and checks each returned row matches the request.

- **Bug Fixes**
  - `@better-auth/kysely-adapter` (MySQL): use `executeTakeFirst()` to read `insertId`; if found, select the row by `id`; otherwise fall back to where-based logic.
  - `@better-auth/drizzle-adapter` (MySQL): use `$returningId()` when available; after `execute()`, if an `id` is returned, select by `id`; otherwise fall back.

<sup>Written for commit f2b393d1985e017e4ee9145b486d73429d3fbb91. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/8971?utm_source=github">Review in cubic</a></sup>


ping-maxwell (dismissed):

ping-maxwell:
awesome work @howlettt !

ping-maxwell:
please also add changeset to the pr 🙌

howlettt:
@ping-maxwell I added a changeset, let me know if anything else is missing

PR #8978, opened 2026-04-06
fix(types): guard type utilities against any poisoning

bytaesu:
Improves type safety when `any` is involved. Previously, mixing in an `any` plugin would poison the other client response types. After this change, valid plugin types are preserved even if an `any` plugin is present.


bytaesu:
=> reworking

PR #8994, opened 2026-04-06
feat: add skew protection header to auth fetches

brookemosby:
Pin Better Auth client-side custom fetch requests to the active deployment during rolling releases by adding x-deployment-id when skew protection is enabled. This aligns the client with Vercel skew-protection environment semantics and prevents auth endpoint drift during NextAuth-to-BetterAuth migrations.

---
## Summary by cubic
Pin Better Auth client fetches to the active Vercel deployment during rolling releases by adding the `x-deployment-id` header when skew protection is enabled. This prevents auth endpoint drift during NextAuth-to-BetterAuth migrations.

- **Bug Fixes**
  - Added a default `vercelSkewProtectionPlugin` to the `better-auth` client. Resolves the deployment ID from `globalThis.NEXT_DEPLOYMENT_ID`, `data-dpl-id`, or `VERCEL_DEPLOYMENT_ID`/`NEXT_DEPLOYMENT_ID`, and sets `x-deployment-id` unless already present. Runs only when default fetch plugins are enabled.
  - Gating and DX: opt out with `disableSkewProtection` or `disableDefaultFetchPlugins`; respects `VERCEL_SKEW_PROTECTION_ENABLED`. Exported `vercelSkewProtectionPlugin` and `resolveDeploymentIdForSkewProtection`, documented options in `BetterAuthClientOptions`, and added tests for resolution, header behavior, gating, and env cleanup.

<sup>Written for commit 0572114c02c59a3a5592a76af7d61aed4450374e. Summary will update on new commits.</sup>


ping-maxwell:
I think it would be better if it's a lot more configurable, where only specific routes are pinned to a deployment ID instead of having all routes being pinned to a deployment ID and client plugins can define specific routes to customize the behavoiour as well.

The issue right now is that it's possible that a fix is made that could be critical; however, due to overall auth paths being tied to a deployment ID, it could introduce potential security vulnerabilities. For lighter paths that are less particularly vulnerable, it could have a stronger duration of deployment ID tied to it, but other endpoints that are very critical and are unlikely to have API changes should not have a deployment ID tied to them.

PR #9022, opened 2026-04-07
feat(sso): expose organizationId in providersLimit function

### LINKED ISSUE #7750: Expose organizationId in SSO providersLimit function

When using the SSO plugin with organizations, the `providersLimit` function only receives the `user` object. This makes it impossible to enforce organization-specific constraints, such as allowing SSO provider registration only for organizations on an enterprise plan.

Currently, if a user is a member of multiple organizations (e.g., one on a free plan and one on an enterprise plan), there's no way to differentiate which organization they're registering the SSO provider for within the `providersLimit` function. The `organizationId` is validated later in the flow but isn't accessible when the limit check occurs.

Enhance the `providersLimit` function signature to include the `organizationId`:

```typescript
// Current
providersLimit?: number | ((user: User) => Awaitable<number>);

// Proposed
providersLimit?: number | ((data: {
  user: User;
  organizationId?: string;
}) => Awaitable<number>);
```

This would enable plan-based validation:

```typescript
sso({
  providersLimit: async ({ user, organizationId }) => {
    if (!organizationId) return 0; // Require organization for SSO
    
    const context = getContext();
    const org = await context.adapter.findOne({
      model: "organization",
      where: [{ field: "id", value: organizationId }]
    });
    
    return org?.plan === "enterprise" ? 10 : 0;
  }
})
```

Custom endpoint wrapper that replaces the built-in registration endpoint. This requires significant code duplication and maintenance.

### Additional context

The organizationId is already available in `ctx.body.organizationId` at the point where `providersLimit` is called. The membership validation happens shortly after. Moving the limit check to after the membership validation, or passing the organizationId to the limit function, would enable this use case without requiring workarounds.

This enhancement would maintain backward compatibility since `organizationId` would be optional.

ping-maxwell:
## Summary
- Changed the `providersLimit` function signature from `(user: User) => number` to `({ user, organizationId }) => number`, allowing organization-specific SSO provider limits
- The `organizationId` from the request body is now passed through when registering an SSO provider, enabling plan-based validation per organization

## Tests
- [x] Added regression test verifying `organizationId` is passed to `providersLimit`
- [x] Existing `providersLimit` tests pass with updated signature
- [x] Type check passes Closes #7750


opswiz:
Yes, much needed, In addition to the plan like checks, its very common to limit 1 SSO provider per org.

Bekacru:
@ping-maxwell is

Bekacru:
Seems like it's still valid? if so, lets fix conflict and merge @ping-maxwell

PR #9023, opened 2026-04-07
fix: endpoint instrumentation to always use route template

jonathansamines:


bytaesu:
@jonathansamines Would this be closer to a `fix: ` than `chore: ` ?

ping-maxwell (approved):

PR #9050, opened 2026-04-09
fix(oauth): prevent crash when APIError has empty message in databaseHooks

### LINKED ISSUE #8676: TypeError: session is null when throwing APIError without message in databaseHooks (i18n + OAuth conflict)
When throwing an `APIError` inside `databaseHooks.user.create.before` during OAuth social sign-in, if you omit the `message` property and rely on the i18n plugin to inject the translation based on `code`, the server crashes with a 500:
 
```
<-- POST /api/auth/sign-in/social
2026-03-18T16:16:15.681Z ERROR [Better Auth]: APIError
2026-03-18T16:16:15.682Z ERROR [Better Auth]: TypeError 123 | async function setSessionCookie(ctx, session, dontRememberMe, overrides) {
124 |   const dontRememberMeCookie = await ctx.getSignedCookie(ctx.context.authCookies.dontRememberToken.name, ctx.context.secret);
125 |   dontRememberMe = dontRememberMe !== void 0 ? dontRememberMe : !!dontRememberMeCookie;
126 |   const options = ctx.context.authCookies.sessionToken.attributes;
127 |   const maxAge = dontRememberMe ? void 0 : ctx.context.sessionConfig.expiresIn;
128 |   await ctx.setSignedCookie(ctx.context.authCookies.sessionToken.name, session.session.token, ctx.context.secret, {
                                                                            ^
TypeError: null is not an object (evaluating 'session.session')
      at setSessionCookie (lovpli/server/node_modules/.pnpm/better-auth@1.5.5_@prisma+client@7.5.0_prisma@7.5.0_@types+react@19.2.14_react-dom@19.2_be99b374499587e8f178a6de2ba3112a/node_modules/better-auth/dist/cookies/index.mjs:128:71)

# SERVER_ERROR:  123 | async function setSessionCookie(ctx, session, dontRememberMe, overrides) {
124 |   const dontRememberMeCookie = await ctx.getSignedCookie(ctx.context.authCookies.dontRememberToken.name, ctx.context.secret);
125 |   dontRememberMe = dontRememberMe !== void 0 ? dontRememberMe : !!dontRememberMeCookie;
126 |   const options = ctx.context.authCookies.sessionToken.attributes;
127 |   const maxAge = dontRememberMe ? void 0 : ctx.context.sessionConfig.expiresIn;
128 |   await ctx.setSignedCookie(ctx.context.authCookies.sessionToken.name, session.session.token, ctx.context.secret, {
                                                                            ^
TypeError: null is not an object (evaluating 'session.session')
      at setSessionCookie (lovpli/server/node_modules/.pnpm/better-auth@1.5.5_@prisma+client@7.5.0_prisma@7.5.0_@types+react@19.2.14_react-dom@19.2_be99b374499587e8f178a6de2ba3112a/node_modules/better-auth/dist/cookies/index.mjs:128:71)

--> POST /api/auth/sign-in/social 500 256ms
```
 
If you include `message` explicitly, it no longer crashes, but the returned code is `OAUTH_LINK_ERROR` instead of `INVALID_ADDITIONAL_DATA`. ```// see below client response```

The same applies to `authClient.signUp.email`, throwing an `APIError` without `message` works correctly there, the i18n plugin picks up the translation and returns the correct `INVALID_ADDITIONAL_DATA` code without any issues.
 
## Steps to reproduce
 
Here's the auth config:
 
```ts
import { i18n } from "@better-auth/i18n";
import { APIError } from "better-auth/api";
 
  plugins: [
    i18n({
      defaultLocale: "en",
      translations: {
        en: { INVALID_ADDITIONAL_DATA: "Invalid additional data." },
      },
    }),
  ],
  databaseHooks: {
    user: {
      create: {
        before: async (user, ctx) => {
          if (ctx?.path.startsWith("/sign-in/social")) {
            const additionalData = schema.safeParse(ctx?.body?.additionalData);
            if (additionalData.error) {
              // ❌ crashes — no message, i18n should provide it
                code: "INVALID_ADDITIONAL_DATA",
              }); // in client:  🚀 ~ onSubmit ~ data: null {"status": 500, "statusText": ""}
 
              // ✅ works fine
              // throw new APIError("BAD_REQUEST", {
              //   code: "INVALID_ADDITIONAL_DATA",
              //   message: "Invalid additional data",
              // });
             // in the client this is the response code  ~ onSubmit ~ data: null {"code": "OAUTH_LINK_ERROR", "message": "Invalid additional data", "status": 401, "statusText": ""}, 
//not INVALID_ADDITIONAL_DATA
            }
          }
          return { data: user };
        },
      },
    },
  },
  socialProviders: { google: { /* ... */ } },
});
```
 
And the client call:
 
```ts
  provider: "google",
  additionalData: { /* invalid data */ },
});
```
 
## Expected behavior
 
The i18n plugin should pick up the `code` and inject the translated message, returning:
 
```json
{
  "code": "INVALID_ADDITIONAL_DATA",
  "message": "Invalid additional data."
}
```
 
## What actually happens
 
The request crashes before the error response is ever sent. The server tries to set a session cookie even though the error should have short-circuited the flow, then blows up because there's no session.
 
Worth noting: the same hook works correctly with email/password sign-up and with the idToken flow. It's specific to social OAuth + missing `message`.
 
## Environment
 
- Better Auth: 1.5.5
- i18n plugin: latest
- Database: Prisma + PostgreSQL
- Runtime: Bun 1.2.x

jaydeep-pipaliya:
## What does this PR do?

Fixes #8676

When an `APIError` is thrown in `databaseHooks.user.create.before` during OAuth social sign-in **without a `message` property** (e.g., relying on the i18n plugin to inject translation based on `code`), the server crashes with:

```
TypeError: null is not an object (evaluating 'session.session')
  at setSessionCookie
```

## Root cause

The crash path:

```
1. databaseHooks.user.create.before throws new APIError("FORBIDDEN")  // no message, just code
2. createOAuthUser catches it in link-account.ts
3. Returns { error: e.message, data: null }    // e.message is "" (empty string)
4. callback.ts checks: if (result.error)       // "" is falsy → check passes through!
5. const { session, user } = result.data!      // result.data is null → CRASH
```

The core issue: `APIError.message` can be an empty string when the user omits the `message` parameter and relies on `code` for i18n translation. Empty string `""` is falsy in JavaScript, so `if (result.error)` doesn't catch it.

## Fix

**`link-account.ts`** — fallback chain for error message:
```typescript
// Before
error: e.message,

// After  
error: e.message || e.body?.code || "unable to create user",
```

**`callback.ts`** — defense in depth, also check for missing data:
```typescript
// Before
if (result.error) { ... }
const { session, user } = result.data!;

// After
if (result.error || !result.data) { ... }
const { session, user } = result.data;
```

## Why both fixes?

1. **`link-account.ts`**: Ensures `handleOAuthUserInfo` always returns a truthy error string so callers can rely on the error check
2. **`callback.ts`**: Removes the unsafe `!` non-null assertion and adds a `!result.data` guard — if any future code path returns `{ error: null, data: null }`, the callback won't crash

---
## Summary by cubic
Prevents a crash during OAuth sign-in when `databaseHooks.user.create.before` throws an `APIError` with an empty message. Always redirects with a safe, normalized error code and avoids null data dereferences.

- **Bug Fixes**
  - `packages/better-auth/src/oauth2/link-account.ts`: on `APIError`, return `e.body?.code || e.message || "unable to create user"`.
  - `packages/better-auth/src/api/routes/callback.ts` and `packages/better-auth/src/plugins/generic-oauth/routes.ts`: add `sanitizeErrorCode` to normalize error strings for `/error`; guard `!result.data`; default to `unable_to_create_user`; remove non-null assertions.
  - `packages/better-auth/src/oauth2/link-account.test.ts`: add regression test to ensure redirect includes a normalized error code and no crash.

<sup>Written for commit 7c7db81a408578572cc9fd7fe53be5ad8d76268c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9050?utm_source=github">Review in cubic</a>


ping-maxwell (dismissed):

ping-maxwell:
Hey @jaydeep-pipaliya could you please add changeset? thanks

jaydeep-pipaliya:
Added the changeset! Thanks for the review @ping-maxwell

jaydeep-pipaliya:
Hi @ping-maxwell, the changeset is already included in the PR — see `.changeset/fix-oauth-empty-message-crash.md`. Let me know if you need anything else!

ping-maxwell (dismissed):

jaydeep-pipaliya:
Hi @ping-maxwell — just a gentle ping! The changeset was added on Apr 9 (see `.changeset/fix-oauth-empty-message-crash.md`). Let me know if there's anything else needed for this to move forward.

gustavovalverde:
@jaydeep-pipaliya, there are pending Copilot comments; address those if valid.

jaydeep-pipaliya:
Thanks @gustavovalverde — I checked the Copilot review. It only left a summary comment (pull request overview) with no inline code suggestions or actionable items. The review just describes what the PR does, no changes requested. Let me know if I'm missing something!

ping-maxwell:
I think both remaining copilot suggestions can be actionable.

jaydeep-pipaliya:
Thanks @ping-maxwell — addressed all 4 Copilot suggestions:

1. **Error code priority** — already using `e.body?.code || e.message` (from earlier commit)
2. **Normalize error strings** — replaced `.split(' ').join('_')` with `replace(/[^A-Za-z0-9_-]/g, '_')` to match the `/error` endpoint's `^[A-Za-z0-9_-]+$` validation. Dots, colons, and other special chars are now stripped.
3. **Guard generic-oauth too** — applied the same `result.error || !result.data` guard + error normalization to `plugins/generic-oauth/routes.ts` and removed the `result.data!` non-null assertion
4. **Regression test** — added test that throws `APIError('FORBIDDEN', { body: { code: 'USER_BANNED' } })` from `databaseHooks.user.create.before`, verifies the callback redirects with a non-empty error code instead of crashing

ping-maxwell (dismissed):

jaydeep-pipaliya:
Hey @ping-maxwell — just checking in on this one. Anything blocking the merge? Happy to address any additional feedback.

ping-maxwell:
There's merge conflicts.

jaydeep-pipaliya:
Resolved — rebased onto current main, integrated the new GHSA-g38m-r43w-p2q7 `requireLocalEmailVerified` opt-out test alongside the existing APIError-empty-message test. Both blocks pass locally.

jaydeep-pipaliya:
Rebased onto latest `main` and addressed the remaining Copilot suggestions:

- Prefer `e.body?.code` over `e.message` in `handleOAuthUserInfo` so the redirected `error` query param stays a stable machine code instead of a localized human string.
- Added a `sanitizeErrorCode` helper in `callback.ts` that strips characters disallowed by the default `/error` validator (`^[A-Za-z0-9_-]+$`) — handles trailing dots, punctuation, etc., instead of only collapsing spaces.
- Applied the same `!result.data` guard and `sanitizeErrorCode` normalization to the matching `handleOAuthUserInfo` consumer in `plugins/generic-oauth/routes.ts` so the crash pattern is fixed there too.
- Regression test `oauth2 - APIError with empty message in databaseHooks` covers the original repro and asserts the redirect doesn't end up as `error=&` or `error=UNKNOWN`.

jaydeep-pipaliya:
Resolved — merged latest `main` (callback.ts and generic-oauth/routes.ts both moved). Kept the `!result.data` guard and `sanitizeErrorCode` from this branch and adopted upstream's 3-arg `redirectOnError(ctx, resolvedErrorURL, code)` signature in both call sites. Tests pass locally.

PR #9078, opened 2026-04-09
fix(client): prevent `isMounted` race condition causing many rps

### LINKED ISSUE #9077: Hundreds of concurrent get-session requests per second

### Reproduction

While reviewing production logs we noticed that multiple user IPs were each generating an extraordinary number of requests to `/api/auth/get-session`. We are talking about hundreds of requests per second, in a simultaneous burst, from a single browser tab.

After ruling out browser extensions, service workers, localStorage events, and network-level retries, we believe the root cause is a race condition in `query.ts`. We've done our best to trace the full chain, but any corrections are very welcome.

---

The storm requires two conditions to be simultaneously true:

1. **A `useSession()` consumer whose React fiber is being repeatedly recreated.** The most common cause is a `<Suspense>` boundary with a suspended sibling (any [`loading.tsx`](https://nextjs.org/docs/app/building-your-application/routing/loading-ui-and-streaming) creates one implicitly around the page, and explicit `<Suspense>` boundaries have the same effect). In our production app, we also observed this happening via Next.js's internal `LoadingBoundary` between route segments, even without `loading.tsx`, though this requires enough page complexity to generate chunked RSC streaming and was difficult to isolate in a minimal repro.

2. **A slow `get-session` response relative to the number of retries.** If the server responds in < 50 ms and there are only a handful of retries, the first `setTimeout(0)` callback resolves, `isMounted` becomes `true`, and subsequent callbacks skip the fetch. The storm only manifests when the response is slow enough that `isMounted` is still `false` when all queued callbacks drain.

Since the natural retries from RSC streaming require real app complexity to trigger in isolation, the repro below uses a `setInterval` to simulate them. The mechanism is the same: rapid Suspense boundary retries -> fresh fibers -> `store.get()` with `lc=0` -> `onMount` -> `setTimeout(0)` -> fetch.

**`page.tsx`**

```tsx
"use client";

import { lazy, Suspense, useEffect, useState, type ComponentType } from "react";
import { authClient } from "@/lib/auth-client";

function SessionWatcher() {
  authClient.useSession();
  return null;
}

const SlowChunk = lazy<ComponentType>(
  () => new Promise((r) => setTimeout(() => r({ default: () => null }), 30000)),
);

function StormDriver() {
  const [, setTick] = useState(0);

  useEffect(() => {
    const id = setInterval(() => setTick((t) => t + 1), 15);
    return () => clearInterval(id);
  }, []);

  return (
    <Suspense fallback={null}>
      <SessionWatcher />
      <SlowChunk />
    </Suspense>
  );
}

export default function Page() {
  return (
    <main style={{ padding: 40, fontFamily: "sans-serif" }}>
      <h2>How to reproduce</h2>
      <p>
        Open DevTools → Network → filter <code>get-session</code>, then{" "}
        <strong>reload this page</strong>.
      </p>
      <p>
        <strong>Expected:</strong> 1 request.
        <br />
        <strong>Actual:</strong> 1 000+ requests in under 5 seconds.
      </p>
      <StormDriver />
    </main>
  );
}
```

**`src/app/api/auth/[...all]/route.ts`**

```typescript
import { toNextJsHandler } from "better-auth/next-js";

const { GET: _GET, POST } = toNextJsHandler(auth);

export async function GET(req: Request) {
  if (req.url.includes("get-session")) {
    await new Promise<void>((r) => setTimeout(r, 30000));
  }
  return _GET(req);
}

export { POST };
```

Current behavior: A single page load fires hundreds of concurrent get-session requests from one tab.

Expected behavior: One get-session request per mount cycle.

1.6.2

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.4.0: Thu Mar 19 19:31:09 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T8132",
    "release": "25.4.0",
    "cpuCount": 10,
    "cpuModel": "Apple M4",
    "totalMemory": "16.00 GB",
    "freeMemory": "0.10 GB"
  },
  "node": {
    "version": "v25.2.1",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "10.23.0"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "^16.2.3"
    },
    {
      "name": "react",
      "version": "^19.2.5"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "1.6.2",
    "config": {
      "emailAndPassword": {
        "enabled": true
      },
      "secret": "[REDACTED]",
      "trustedOrigins": [
        "http://localhost:3000"
      ]
    }
  }
}
```

Client

```typescript

```

#### The `isMounted` guard appears to have a timing gap

**`packages/better-auth/src/client/query.ts`**

```typescript
onMount(value, () => {
  const timeoutId = setTimeout(async () => {
    if (!isMounted) {
      await fn();       // fetch starts here
      isMounted = true; // only flipped after the fetch resolves
    }
  }, 0);
  // ...
});
```

If multiple `setTimeout(0)` callbacks are queued before the first fetch resolves, each one would enter the `if (!isMounted)` branch and start its own fetch, since `isMounted` hasn't been set to `true` yet.

#### How multiple `setTimeout(0)` callbacks end up queued

The nanostores React binding (`useStore`) calls `store.get()` inside `useRef` during render:

```js
function useStore(store) {
  let snapshotRef = useRef(store.get()); // fires onMount if lc === 0
  // ...
}
```

When the atom's listener count (`lc`) is 0, `store.get()` creates a temporary subscription (`lc: 0 → 1 → 0`) and fires all `onMount` callbacks synchronously during the render phase, before the commit. The real subscription (via `store.listen`) would only be registered at commit time by `useSyncExternalStore`.

When a component inside a `<Suspense>` boundary suspends, React discards the work-in-progress fiber for siblings inside that boundary, including any component that called `useSession`. Because there was no commit, the listener count returns to 0.

On the next retry of the boundary, `useRef` reinitializes, `store.get()` sees `lc = 0`, `onMount` fires again, and another `setTimeout(0)` is queued.

#### What causes the rapid retries in production?

In our production app (Next.js App Router), we instrumented the session atom and observed **420 Suspense retries** and **852 total re-renders** of the component calling `useSession()` on a single hard refresh, without any explicit `<Suspense>` boundary or `loading.tsx` in our code.

We traced the retries to Next.js's internal `LoadingBoundary`, a Suspense boundary the framework creates between route segments (layout <-> page). When the page segment is complex enough (e.g. heavy component trees), the RSC payload streams in multiple chunks, and each chunk can trigger a retry of the boundary, re-rendering the layout's client components with fresh fibers.

That said, we're not 100% sure this is the only factor. There could be other React internals or framework behaviors contributing to the retry frequency that we haven't identified.

### Related

- #4609 — Same symptom (hundreds of `get-session` calls per page load). PR #4669 introduced the `isMounted` guard but placed it **after** `await fn()` rather than before, which from what we can tell leaves the race condition intact.

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/9077


bytaesu:
@ping-maxwell Would you like me to add a regression test for this?

bytaesu:
I think renaming the internal var from `isMounted` to `isInitialized` could make the behavior clearer. That said, since it's just an internal var, the current naming is also fine!

ping-maxwell:
@bytaesu done!

packages/better-auth/src/client/query.ts
```diff
@@ -120,25 +120,27 @@ export const useAuthQuery = <T>(
...
-						if (!isMounted) {
+						if (!isInitialized) {
+							// Must set to `true` immediately; see https://github.com/better-auth/better-auth/issues/9077
+							isInitialized = true;
 							await fn();
-							isMounted = true;
 						}
 					}, 0);
 					return () => {
+						isInitialized = false;
```
bytaesu: isInitialized doesn't need to be reset during cleanup, since re-mounts already refetch via the `if (isInitialized)` branch. Resetting it would reintroduce the race condition this PR is meant to fix.

ping-maxwell: Oh I added that because of https://github.com/better-auth/better-auth/issues/9077#issuecomment-4217332571

PR #9088, opened 2026-04-10
fix(oauth-provider): sort query params before signing to prevent CDN reordering failures

### LINKED ISSUE #8858: OAuth provider: HMAC signature verification fails when CDN/proxy reorders query parameters
## Description

The OAuth provider plugin's `signParams` and `verifyOAuthQueryParams` functions compute HMAC signatures over `URLSearchParams.toString()`, which preserves insertion order. If any intermediary (CDN, reverse proxy, edge layer) reorders URL query parameters in 302 redirect `Location` headers, the signature verification fails with `invalid_signature`.

This is a real-world issue on **Vercel** (their edge layer reorders query params in redirects — see [vercel/next.js#39017](https://github.com/vercel/next.js/issues/39017)), but could affect any deployment behind a CDN that normalizes URLs.

## Steps to reproduce

1. Deploy a SvelteKit app with `@better-auth/oauth-provider` on Vercel with Deployment Protection enabled
2. Start an OAuth authorization flow (`/api/auth/oauth2/authorize`)
3. The authorize endpoint signs query params and redirects to `/signin?...&exp=X&sig=Y`
4. Vercel's edge layer reorders the query params in the `Location` header
5. The browser receives the URL with params in a different order
6. On sign-in, `oauthProviderClient` sends the reordered params as `oauth_query`
7. `verifyOAuthQueryParams` computes HMAC over the reordered string → **mismatch** → `invalid_signature`

Additionally, the client-side `parseSignedQuery` iterates params and stops at `sig`, assuming it's the last parameter. If `sig` gets reordered to the middle, the query is truncated.

## Root cause

In `packages/oauth-provider/src/authorize.ts` (`signParams`):
```ts
const params = new URLSearchParams(ctx.query);
params.set("exp", String(exp));
const signature = await makeSignature(params.toString(), ctx.context.secret);
```

In `packages/oauth-provider/src/utils/index.ts` (`verifyOAuthQueryParams`):
```ts
queryParams.delete("sig");
const verifySig = await makeSignature(queryParams.toString(), secret);
```

`URLSearchParams.toString()` preserves insertion order. URLs with the same parameters in different orders are semantically equivalent per RFC 3986, so HMAC signatures should be order-independent.

## Proposed fix

Call `URLSearchParams.sort()` before computing the HMAC in both `signParams` and `verifyOAuthQueryParams`. This ensures deterministic parameter ordering regardless of how they arrive. This is the same approach used by AWS Signature V4.

```diff
// In signParams:
  const params = new URLSearchParams(ctx.query);
  params.set("exp", String(exp));
+ params.sort();
  const signature = await makeSignature(params.toString(), ctx.context.secret);

// In verifyOAuthQueryParams:
  queryParams.delete("sig");
+ queryParams.sort();
  const verifySig = await makeSignature(queryParams.toString(), secret);
```

This is a non-breaking change — existing signatures would need to be regenerated on the next authorization flow (they expire in 10 minutes anyway via the `exp` field).

## Environment

- `@better-auth/oauth-provider@1.5.6`
- SvelteKit on Vercel (serverless)
- Vercel Deployment Protection enabled

lawrence3699:
## Summary

Fixes #8858

The OAuth provider plugin signs query parameters with HMAC but the signature depends on parameter insertion order (`URLSearchParams.toString()` preserves order). When a CDN or reverse proxy reorders query params in redirect `Location` headers — which Vercel's edge layer does — the signature verification fails with `invalid_signature` and the OAuth flow breaks.

## Changes

- **`authorize.ts`** (`signParams`): sort params before computing signature
- **`utils/index.ts`** (`verifyOAuthQueryParams`): sort params before verification
- **`client.ts`** (`parseSignedQuery`): stop truncating at `sig` — the old loop assumed `sig` was always last, which breaks when params are reordered or sorted

This follows the same approach used by AWS Signature V4 (canonical query string sorting).

## Before / After

| | Before | After |
|---|---|---|
| CDN reorders params | `invalid_signature`, OAuth flow fails | Verification succeeds — params sorted before HMAC |
| `sig` not last in URL | `parseSignedQuery` truncates valid params | All params preserved |

## Test plan

Added two regression tests in `authorize.test.ts`:
1. Sign params, reverse their order (simulating CDN), verify still passes
2. Sign params, tamper with a value, verify correctly rejects

All existing tests pass:
- `authorize.test.ts`: 20/20
- `oauth.test.ts`: 48/48
- `oauthClient/endpoints.test.ts`: 10/10
- `pnpm lint:types`: all 39 packages pass

---
## Summary by cubic
Sorts OAuth query params before HMAC signing and verification in `@better-auth/oauth-provider` to make signatures order-independent when CDNs reorder query strings. Also updates client parsing so `sig` can appear anywhere.

- **Bug Fixes**
  - Sort params in `authorize.ts` and `utils/index.ts` before computing/verifying signatures to prevent `invalid_signature` after CDN reordering.
  - Update `client.ts` to return the full query string instead of truncating at `sig`.
  - Add tests for reordered params (passes) and tampered params (rejects).

<sup>Written for commit 772cee765aeeaab2ccc25ef519c20dc44062d2db. Summary will update on new commits.</sup>


packages/oauth-provider/src/authorize.ts
```diff
@@ -627,6 +627,8 @@ async function signParams(
 	);
 	params.set("exp", String(exp));
 
+	// Sort params so the HMAC is order-independent (CDN/proxy may reorder query params)
```
gustavovalverde, line 630: Nit: drop this comment. `params.sort()` reads for itself and the repo convention avoids explanatory comments where the code is self-evident.

packages/oauth-provider/src/utils/index.ts
```diff
@@ -144,6 +144,8 @@ export async function verifyOAuthQueryParams(
 	const sig = queryParams.get("sig");
 	const exp = Number(queryParams.get("exp"));
 	queryParams.delete("sig");
+	// Sort to match signing order — CDN/proxy may reorder query params
```
gustavovalverde, line 147: Same nit: the symmetric `params.sort()` on the signing side above makes this self-evident. Safe to drop.

gustavovalverde (approved):
Approach is correct

Address the copilot comments. And also two small comment nits inline; otherwise LGTM

.changeset/sort-hmac-params.md
```diff
@@ -0,0 +1,5 @@
+---
+"@better-auth/oauth-provider": patch
+---
+
+fix(oauth-provider): sort query params before signing/verification to prevent CDN reordering failures
```
ping-maxwell, line 5: When this patch is deployed, any OAuth flow that was initiated before deployment (i.e. params signed without sorting) will fail verification, because the new `verifyOAuthQueryParams` sorts before checking the HMAC. Affected sessions will get `invalid_signature` until their exp expires (default: 600 seconds / 10 minutes).

I think this will be fine for a patch release, but we should mention this transient incompatibility so operators know to expect a blip after deploy

ping-maxwell (changes requested):

lawrence3699:
Closing this because the PR has requested changes, a merge conflict, and a deployment compatibility concern that I am not resolving immediately.

PR #9093, opened 2026-04-10
feat(core): add user soft delete with anonymization

mcorbelli:
# PR Summary — Soft Delete Support for Users

Adds opt-in **soft delete** support for users. Instead of hard-deleting the user row, Better Auth anonymizes the record and sets a `deletedAt` timestamp — preserving referential integrity with related tables (posts, orders, audit logs, etc.) while fully revoking access.

---

## How to Enable

```ts title="auth.ts"
  user: {
    deleteUser: {
      enabled: true,
      softDelete: true, // new
    },
  },
});
```

---

## Behaviour

- All three deletion paths (self-delete immediate, self-delete with email confirmation, admin `removeUser`) route through `internalAdapter.deleteUser` and respect the flag — no per-path changes needed.

- Anonymization overwrites email, name, and image with safe placeholder values.

  **Default strategy:**
  - email → `deleted-{userId}@deleted.invalid` (unique per user, RFC-reserved TLD)
  - name → `"Deleted User"`
  - image → `null`

- A custom `anonymizeUser` callback can be provided to override the defaults.

- Sessions and credential/OAuth accounts are always removed regardless of soft/hard delete.

- Sign-in is blocked for soft-deleted users via a `session.create.before` database hook — returns `403 FORBIDDEN` with code `USER_DELETED`.

---

## Admin Plugin Additions

- New `POST /admin/restore-user` endpoint — clears `deletedAt`, re-enabling sign-in.
  - Requires the `user:restore` permission (admin role only).

- `listUsers` gains an `includeDeleted` boolean query parameter (default: `false`)
  - Soft-deleted users are excluded unless explicitly requested.

---

## Changes

| Area | What changed |
|------|-------------|
| packages/core | `deletedAt` added to `userSchema`; `softDelete` + `anonymizeUser` options added to `user.deleteUser`; `USER_DELETED` error code added |
| packages/better-auth | Soft delete logic in `internalAdapter.deleteUser`; sign-in block hook in `helpers.ts` |
| packages/better-auth/plugins/admin | `deletedAt` schema field; "restore" permission; `YOU_ARE_NOT_ALLOWED_TO_RESTORE_USERS` error code; `restoreUser` endpoint; `includeDeleted` query param in `listUsers` |
| docs | Soft delete section added to `users-accounts.mdx` |
| tests | 7 new tests covering all soft delete scenarios (79/79 passing) |

---

## Notes

- Restoring a user clears `deletedAt` only — original PII (email, name, image) is gone and cannot be recovered.
- Admins should re-onboard the user if needed.

- The `deletedAt` column is added to the user table via the standard schema migration:

```bash
pnpm dlx @better-auth/cli migrate
```

---
## Summary by cubic
Adds opt‑in user soft delete with anonymization and a `deletedAt` timestamp to preserve data links while revoking access and blocking sign‑in. Requires a schema migration to add `deletedAt`.

- **New Features**
  - `user.deleteUser.softDelete: true` anonymizes email/name/image, sets `deletedAt`, and removes sessions plus OAuth/credential accounts.
  - Optional `anonymizeUser` callback to customize anonymization; reserved fields are ignored.
  - Sign‑in blocked for soft‑deleted users (403 `USER_DELETED`).
  - Admin: `POST /admin/restore-user` clears `deletedAt` (requires `user:restore`); does not restore linked accounts—user must be re‑onboarded to sign in.
  - Admin: `listUsers` accepts `includeDeleted` to include soft‑deleted users.

- **Bug Fixes**
  - Correctly parse `includeDeleted` when sent as `"false"` in query params.
  - Sign‑in block runs without request context to prevent programmatic session creation bypass.
  - Add `deletedAt` to core user table schema when soft delete is enabled, even without the admin plugin.

<sup>Written for commit 0dc1c6b53a78262ff9bdc3e8dc3b86bd20f7e9e2. Summary will update on new commits.</sup>


bytaesu:
We won't be making changes to the core schema at the moment. Also, large PRs without proper discussion won’t be reviewed to maintain existing system stability. Please check the contribution guide.

https://github.com/better-auth/better-auth/blob/main/CONTRIBUTING.md

mcorbelli:
Thanks for the feedback — totally understand. I've opened a feature request issue to start the discussion properly before moving forward with the PR.
 
 Just to give a quick overview of the intent: the feature adds soft-delete support to the admin plugin, allowing admins to mark users as deleted (via a deletedAt timestamp) rather than permanently removing them. This is a common requirement in production apps for audit trails, GDPR right-to-erasure workflows, and account recovery.
 
 The core schema change (deletedAt on the user table) is the only touch point outside the admin plugin itself — everything else (endpoints, anonymization, sign-in block, restore) lives entirely within packages/better-auth/src/plugins/admin.  I'm happy to discuss whether deletedAt should live in core or be plugin-scoped, and to scope down the PR based on your feedback.
 
 Looking forward to the discussion on the [issue](https://github.com/better-auth/better-auth/issues/9099). Thanks again for your time! @bytaesu

PR #9103, opened 2026-04-10
feat: warn when cookie integration plugins are not last in plugins[]

### LINKED ISSUE #8911: Add guardrails for cookie integration plugin order to prevent silent missing Set-Cookie

Yes.

When using Better Auth, cookie integration plugins (like `nextCookies`, `tanstackStartCookies`, etc.) must be placed last in the `plugins[]` array.

If they are not last, any cookies added by plugins that run after them are not forwarded to the framework cookie store. This leads to silent failures like:
- session not persisting
- 2FA breaking (INVALID_*_COOKIES)
- null sessions or redirect loops

This behavior is not obvious and no warning/error is shown, making it hard to debug.

Add a guardrail during `betterAuth()` initialization:

- Detect cookie integration plugins (by `plugin.id`)
- If they are not the last plugin:
  - show a clear warning (or throw in dev mode)

Example message:
> Cookie integration plugins should be last. Otherwise some Set-Cookie headers may be silently dropped.

This would prevent common misconfiguration and improve DX.

A more robust solution would be to make execution order independent by ensuring cookie plugins always run last internally (e.g., via hook finalization or priority system).

But a warning-based guardrail is simpler and a good first step.

### Additional context

This issue has come up multiple times where the fix was simply "move the cookie plugin to the end".

Adding a guardrail would prevent this recurring confusion and improve reliability of core auth flows.

Happy to open a PR if this approach sounds good 👍

h30s:
Problem: Cookie integration plugins (nextCookies, tanstackStartCookies, sveltekitCookies) must be last in plugins[]. If they're not, cookies from subsequent plugins are silently dropped — causing missing sessions, broken 2FA, and redirect loops with no error shown.

Fix: Added a logger.warn() in createAuthContext that detects when a cookie integration plugin is not in the final position and prints a clear message with the fix.

Closes #8911

---
## Summary by cubic
Add a guardrail that warns when cookie integration plugins aren’t last in `plugins[]`. This prevents silent cookie drops that cause missing sessions, broken 2FA, or redirect loops.

- **Bug Fixes**
  - In `createAuthContext`, log a `logger.warn` if any of `next-cookies`, `tanstack-start-cookies`, `tanstack-start-cookies-solid`, or `sveltekit-cookies` appear before the last plugin. The warning lists all misplaced IDs and explains the fix.
  - Added tests for misordered vs. correct order, single-plugin setups, and cases where one or multiple known cookie plugins are misplaced.

<sup>Written for commit 432ca87eebbfa87608015bb60f6798b747d11fce. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/cookie-plugin-guardrail.md
+++ b/.changeset/cookie-plugin-guardrail.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+warn when cookie integration plugins are not last in plugins[]
--- a/packages/better-auth/src/context/create-context.test.ts
+++ b/packages/better-auth/src/context/create-context.test.ts
@@ -1938,10 +1938,7 @@ describe("cookie integration plugin order guardrail", () => {
 		const log = vi.fn();
 		await initBase({
 			logger: { level: "warn", log } as any,
-			plugins: [
-				{ id: "next-cookies" },
-				{ id: "some-other-plugin" },
-			],
+			plugins: [{ id: "next-cookies" }, { id: "some-other-plugin" }],
 		});
 		expect(log).toHaveBeenCalledWith(
 			"warn",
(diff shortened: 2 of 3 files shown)


ping-maxwell:
I don't think this is the right approache to the problem, we need a more abstract solution to this, especially when Better Auth supports plugins outside of in-house ones.

h30s:
> I don't think this is the right approache to the problem, we need a more abstract solution to this, especially when Better Auth supports plugins outside of in-house ones.

that makes sense. I agree this approach is too specific. Could you guide me on the more abstract direction you'd prefer?

PR #9104, opened 2026-04-10
chore(adapters): add shared coverage for empty update where conditions

MuzzaiyyanHussain:
Summary

Adds shared adapter test coverage for handling empty where conditions in the update method.

Changes
Added a shared adapter suite test in packages/test-utils/src/adapter/suites/basic.ts
Verifies adapters return null when update is called with an empty where condition
Removed the Prisma-specific duplicate test
Why

The behavior for empty where conditions was previously only covered in a Prisma-specific test.
Moving this coverage into the shared adapter suite ensures consistent behavior validation across all adapters and helps prevent regressions.

Notes
No runtime logic changes
Test coverage improvement only


ping-maxwell (changes requested):
missing changeset, also add this directly `basic.ts` for all adapters to return the same result, instead of just testing prisma.

MuzzaiyyanHussain:
Thanks for the suggestion!

Just to confirm , when you mentioned adding this directly in basic.ts for all adapters, do you mean updating the shared logic in the core adapter (e.g. factory.ts / base adapter), or the test suite in packages/test-utils/src/adapter/suites/basic.ts?

I want to make sure I’m applying the change in the correct place for consistent behavior across all adapters.

ping-maxwell:
> Thanks for the suggestion!
> 
> Just to confirm , when you mentioned adding this directly in basic.ts for all adapters, do you mean updating the shared logic in the core adapter (e.g. factory.ts / base adapter), or the test suite in packages/test-utils/src/adapter/suites/basic.ts?
> 
> I want to make sure I’m applying the change in the correct place for consistent behavior across all adapters.

Yeah, update the `packages/test-utils/src/adapter/suites/basic.ts`

MuzzaiyyanHussain:
Added the test in basic.ts

ping-maxwell (approved):

PR #9111, opened 2026-04-10
chore(core): update `@opentelemetry/api` dep declaration to be optional

### LINKED ISSUE #8765: CLI is missing dependency on @opentelemetry/api

### To Reproduce

This command that previously worked in `v1.5.0` should work:
```sh
yarn dlx auth@latest generate --config src/lib/server/auth/index.cli.ts --output src/lib/server/db/auth-schema.ts -y
```

In PR 8711 based off canary, it currently does not:
```sh
node:internal/modules/package_json_reader:268
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@opentelemetry/api' imported from /private/var/folders/hy/d7x_09b11h31rvkqs02d6xg80000gn/T/xfs-a58b2fdf/dlx-14379/node_modules/@better-auth/core/dist/instrumentation/tracer.mjs
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:268:9)
    at packageResolve (node:internal/modules/esm/resolve:768:81)
    at moduleResolve (node:internal/modules/esm/resolve:858:18)
    at defaultResolve (node:internal/modules/esm/resolve:990:11)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:768:20)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:745:38)
    at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:318:38)
    at #link (node:internal/modules/esm/module_job:208:49) {
  code: 'ERR_MODULE_NOT_FOUND'
}

Node.js v24.8.0
```

This is because `@opentelemetry/api` is set as a peer dependency, but it's not actually optional, even if you pass `BETTER_AUTH_TELEMETRY=0` in `.env`.

To prove this is the case, the following command that explicitly adds `@opentelemetry/api` to the sandbox works fine:
```sh
yarn dlx -p auth@latest -p @opentelemetry/api auth generate --config src/lib/server/auth/index.cli.ts --output src/lib/server/db/auth-schema.ts -y
```

Note there is currently no need to add the other adapters manually here.

N/A

https://pkg.pr.new/@better-auth/core@8711

### System info

```bash
N/A
```

Other

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

jonathansamines:
- Closes #8765


packages/core/src/instrumentation/api.ts
```diff
@@ -0,0 +1,52 @@
...
+	end(): void {}
+	setAttribute(_key: string, _value: unknown): void {}
+	setStatus(_status: unknown): void {}
+	recordException(_exception: unknown): void {}
+}
+
+const noopSpanInstance = new NoopSpan() as Span;
+
+class NoopTracer {
+	startActiveSpan<F extends (span: Span) => unknown>(
```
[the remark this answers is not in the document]
jonathansamines: Not an issue with the current public API.

packages/core/src/instrumentation/api.ts
```diff
@@ -0,0 +1,59 @@
...
+	typeof import("@opentelemetry/api"),
+	"trace" | "SpanStatusCode"
+>;
+
+class NoopSpan {
+	end(): void {}
+	setAttribute(_key: string, _value: unknown): void {}
+	setStatus(_status: unknown): void {}
+	recordException(_exception: unknown): void {}
+}
```
bytaesu: We don't typically use classes in our codebase. Are there alternative approaches?

packages/core/src/instrumentation/instrumentation.test.ts
```diff
@@ -1,23 +1,66 @@
...
 import { withSpan } from ".";
+import { getOpenTelemetryAPI } from "./api";
 import { ATTR_DB_COLLECTION_NAME, ATTR_DB_OPERATION_NAME } from "./attributes";
 
+const spanWaitDefaults = { timeout: 10_000, interval: 5 } as const;
+
+async function primeLazyOpenTelemetryForTests(): Promise<void> {
+	getOpenTelemetryAPI();
+	await import("@opentelemetry/api");
+	await vi.waitUntil(
```
bytaesu, line 26: The tests work around this, but the first request could lose all spans until the dynamic import resolves.

jonathansamines: Yeah, that's a tradeoff with lazy-loading OpenTelemetry like this. To be honest, I don't think this is a huge deal, usually, the SDK would have loaded the OpenTelemetry API to memory by the time application code is even reached, so I'd expect data-loss to be minimal to non-existent.

packages/core/src/instrumentation/api.ts
jonathansamines: Yeah, can use factory functions, but to be honest, not sure what the problem is with classes, specially for internal implementation details like this. Should I switch to factories?

bytaesu: could you refactor this? 🙏
(for consistency with the existing codebase)

packages/core/src/instrumentation/instrumentation.test.ts
bytaesu: I've thought about it, but I don't have a clear solution right now 🤔

packages/core/src/instrumentation/api.ts
jonathansamines: Yes! just updated the implementation to avoid classes.

packages/core/src/instrumentation/instrumentation.test.ts
jonathansamines: Yeah, my original implementation used TLA, but that seemed to break some existing setups (it broke e2e:smoke) that we explicitly want to support. I think TLA is probably the only solution that would be able to guarantee data-loss, as module loading would still happen at module evaluation time, and not at runtime.

I did search for similar implementations, and they all rely on TLA in some capacity, either at runtime or build time (e.g some precompile step). 

For what is worth, I think our current solution is good enough, I'd say let's ship this to unblock users having these issues, and we can always adjust the implementation to something more complex and/or reliable if we hear strong feedback that our current solution is not working.

bytaesu (approved):
Thank you @jonathansamines 😁

PR #9119, opened 2026-04-11
docs(test-utils): clarify production usage

### LINKED ISSUE #9101: docs: clarify if testUtils plugin is fine to unconditionally include in prod

### Reproduction

N/A

its not clear if it must only be used in testing context, or if its safe to just always add to the plugin list unconditionally

obviously you shouldn't use the test helpers IN prod (e.g. for signup) - but the docs are not clear on wether its safe to even load it unconditionally and in prod

N/A

### System info

```bash
N/A
```

Documentation

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

GautamBytes:
## Summary
- clarify that `testUtils` does not expose public routes, but does add privileged helpers on `ctx.test`
- recommend using a separate test-only auth instance such as `auth.test.ts` instead of a production auth config
- document the TypeScript inference tradeoff around conditional plugin arrays
- align the main docs, 1.5 release blog snippet, and plugin JSDoc example

Closes #9101

---
## Summary by cubic
Clarify production guidance for `testUtils`: it registers no public routes but exposes privileged `ctx.test` helpers; recommend using a test-only auth instance (e.g., `auth.test.ts`) to avoid shipping helpers and keep TypeScript inference stable. Update docs and examples (plugin docs with a “Can I include this in production?” note, 1.5 blog) to use `testAuth` and explain the conditional `plugins` inference caveat; add a patch changeset for `better-auth`.

<sup>Written for commit 9b3aae1a82b298ff49c67dab64767574b8c13237. Summary will update on new commits.</sup>


GautamBytes:
@ping-maxwell , ready for review!

ping-maxwell (approved):

PR #9150, opened 2026-04-13
feat(api-key): add option to disable inline expired key cleanup

### LINKED ISSUE #9141: api-key: Add option to disable inline `deleteAllExpiredApiKeys` cleanup

### Problem

The `api-key` plugin runs `deleteAllExpiredApiKeys()` on every route handler (create, verify, list, update, delete). For applications that generate high volumes of keys (e.g. short-lived tokens), this causes serious database performance issues:

- The DELETE query scans the entire table if no index exists on `expires_at`
- Multiple concurrent requests bypass the 10-second in-memory throttle (race condition on `lastChecked` timestamp)
- Foreign key cascades (`ON DELETE ...`) on referencing tables compound the cost

Add a configuration option to disable the inline cleanup, allowing users to manage expired key cleanup externally (e.g. via cron calling the existing `deleteAllExpiredApiKeys` endpoint), or avoiding it entirely.

```ts
apiKey({
  keyExpiration: {
    autoCleanup: false, // default: true
  },
})
```

GautamBytes:
## Description

This PR introduces a configuration option to disable the automatic, inline cleanup of expired API keys. 

Currently, the `api-key` plugin runs `deleteAllExpiredApiKeys()` on every route handler (`create`, `verify`, `list`, `update`, `delete`). For applications that generate high volumes of short-lived tokens, this causes significant database performance bottlenecks (full table scans, race conditions bypassing the in-memory throttle, and cascading deletes).

By adding `keyExpiration.autoCleanup` (defaulting to `true` for backward compatibility), developers can opt out of this inline overhead and manage cleanup externally (e.g., via a background worker or cron job hitting the manual endpoint).

Closes #9141

## Changes Made
- Added `autoCleanup?: boolean` to `ApiKeyConfigurationOptions` under the `keyExpiration` object.
- Wrapped all inline `deleteAllExpiredApiKeys()` triggers in `create`, `get`, `list`, `update`, `delete`, and `verify` routes with the new flag.
- Updated the session-mocking `before` hook to correctly evaluate the resolved configuration before triggering cleanup.
- Ensured seamless integration with `deferUpdates` for serverless environments.
- Added comprehensive unit/integration tests in `api-key.test.ts` to verify background tasks and standard endpoints respect the flag.
- Documented the new option and external cron recommendation in `index.mdx` and `reference.mdx`.

---
## Summary by cubic
Adds `keyExpiration.autoCleanup` to `@better-auth/api-key` to let apps disable inline expired-key cleanup, reducing DB work on hot paths and enabling external cleanup via cron/worker. Also updates the `better-auth` client to refetch the session after `/change-password` and `/revoke-other-sessions`.

- **New Features**
  - New `keyExpiration.autoCleanup` (default `true`) gates inline `deleteAllExpiredApiKeys()` across `create`, `get`, `list`, `update`, `delete`, `verify`, and API key session lookups. `list` runs cleanup only when all matched configs enable `autoCleanup`; `verify` queues cleanup only when both `autoCleanup` and `deferUpdates` are `true`. Use `auth.api.deleteAllExpiredApiKeys()` for external cleanup.

- **Bug Fixes**
  - `better-auth` client now refetches session after `/change-password` and `/revoke-other-sessions`.

<sup>Written for commit 9014e37bb1bad2fa3e4a853867fbc8d13eba6f08. Summary will update on new commits.</sup>


ping-maxwell (approved):

.changeset/rare-rocks-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+"@better-auth/api-key": patch
```
ping-maxwell: this should be `minor`

ping-maxwell (changes requested):

GautamBytes:
@ping-maxwell done!

packages/api-key/src/routes/verify-api-key.ts
```diff
@@ -298,15 +298,18 @@ export function verifyApiKey({
...
+					const cleanupTask = deleteAllExpiredApiKeys(ctx.context).catch(
+						(err) => {
 							ctx.context.logger.error(
 								"Failed to delete expired API keys:",
 								err,
 							);
-						}),
+						},
 					);
+					if (opts.deferUpdates) {
```
ping-maxwell: this `if` statement is not needed, the outer `if` statement already checks if `opts.deferUpdates` is true.

packages/api-key/src/routes/list-api-keys.ts
```diff
@@ -366,7 +369,13 @@ export function listApiKeys({
 				paginatedApiKeys = paginatedApiKeys.slice(0, limit);
 			}
 
-			deleteAllExpiredApiKeys(ctx.context);
+			const shouldAutoCleanup =
+				cleanupConfigs.length > 0 &&
+				cleanupConfigs.every((config) => config.keyExpiration.autoCleanup);
+
```
ping-maxwell, line 374: When no `configId` is provided, the cleanup decision requires **every** matching config to have autoCleanup: true.

This means if a user has two configs, one with `autoCleanup: true` and one with `autoCleanup: false`, calling `listApiKeys()` without a `configId` will never run cleanup, even though one config explicitly wants it. This feels like a foot-gun. 🤔

ping-maxwell (changes requested):

GautamBytes:
Fixed the `verify-api-key` one and removed the extra inner `if`.

For `listApiKeys`, I didn’t switch it to `some(...)`. Since the cleanup runs globally, that would make a no-`configId` list call clean up keys even when one matching config has `autoCleanup: false`, which felt wrong for this option.

ping-maxwell (approved):

PR #9187, opened 2026-04-14
fix(api-key): run secondary-storage api-key executions in parallel

ping-maxwell:


bytaesu:
I'm checking this PR 🤓

bytaesu:
> [!NOTE]
> https://github.com/better-auth/better-auth/pull/9187/commits/c960bde8abe14e734c8eb1e0585d055d9856278c
>
> No breaking change to public API or storage contract. Cache state in fallback mode changes (invalidate-on-write instead of Read->Modify->Write), which is the mechanism  behind the race fix.

bytaesu:
@cubic-dev-ai review

bytaesu:
> [!NOTE]
> `mapConcurrent` -> I'll follow up with another PR to extract it into a shared utility and see if it can be used elsewhere.

bytaesu (approved):

PR #9191, opened 2026-04-14
docs(cimd): recommend removal of `allowUnauthenticatedClientRegistration` for client deduplication

dvanmali:
Add a recommendation to remove `allowUnauthenticatedClientRegistration` in most cases (eg MCP cases) which provides URL identification and deduplication of client ids for the same client.

Format fix for TypeTable to match other mdx files (`lint fix` formats it better this way).

---
## Summary by cubic
Recommend disabling `allowUnauthenticatedClientRegistration` when using `@better-auth/cimd` (especially for MCP-only auth) so clients are identified by the `client_id` URL and duplicate registrations are avoided; keep it only if you need a fallback.

Docs: added a CIMD info callout, clarified that “unauthenticated” means “without an active session,” directed MCP users to CIMD, switched related callouts to info, and standardized the CIMD Options `TypeTable` in MDX.

<sup>Written for commit c3e1ef5e089864c9c921b1db1b58c7bc4197d66e. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/cimd-plugin.md
+++ b/.changeset/cimd-plugin.md
@@ -57,4 +57,4 @@ The `checkOAuthClient` and `oauthToSchema` helpers are now exported for plugins 
 
 `jwks_uri` validation now accepts a same-origin URL when the `client_id` itself is an HTTPS URL, since URL-based discovery flows verify the origin through the `client_id` itself.
 
-Docs recommendation to remove the oauth-provider configuration `allowUnauthenticatedClientRegistration` while using CIMD.
+Documentation now recommends removing the oauth-provider configuration `allowUnauthenticatedClientRegistration` when using CIMD.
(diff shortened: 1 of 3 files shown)


gustavovalverde:
@dvanmali This recommendation feels too broad as written.

Removing `allowUnauthenticatedClientRegistration` makes sense when CIMD is the only public-client mechanism, but that flag still controls whether `/oauth2/register` accepts public clients without a session. In a mixed deployment that supports both CIMD URL clients and ordinary unauthenticated DCR, following this guidance would break the non-CIMD flow.

Can you scope this to MCP/CIMD-only setups, or explicitly call out that tradeoff?

dvanmali:
@gustavovalverde  Updated, hope this wording is better

gustavovalverde (approved):

PR #9193, opened 2026-04-14
fix(org): make invitation acceptance idempotent

### LINKED ISSUE #9154: Race condition in acceptInvitation creates duplicate member records

### To Reproduce

1. Call `acceptInvitation` twice concurrently with the same `invitationId` (this naturally happens in React Strict Mode during development, or under real network retries/double-clicks in production)
2. Both calls succeed and create two separate `member` records for the same `(organizationId, userId)` pair

**Current behavior:**

The `acceptInvitation` endpoint in `crud-invites.ts` has a classic TOCTOU (time-of-check-time-of-use) race condition:

1. **Line 561–568**: It reads the invitation and checks `status === "pending"` 
2. **Line 629–632**: It updates the invitation status to `"accepted"`
3. **Line 694–699**: It calls `adapter.createMember()`

Two concurrent requests can both pass the `status === "pending"` check (step 1) before either has updated the status (step 2). Both then proceed to update the invitation and create separate member records, resulting in **duplicate members** for the same user in the same organization.

Additionally, there is no unique constraint on `(organizationId, userId)` in the `member` model, so the database does not catch these duplicates either.

**Expected behavior:**

- `acceptInvitation` should be idempotent — calling it twice for the same invitation should not create duplicate members.
- At minimum, one of the following should prevent duplicates:
  1. An atomic check-and-update on the invitation (e.g., `UPDATE ... WHERE status = 'pending'` and check affected rows) so only one concurrent request proceeds
  2. A unique constraint on `(organizationId, userId)` in the `member` table
  3. An existence check before `createMember` (e.g., check if a member with this `userId` + `organizationId` already exists)

### Suggested fix

The most robust fix would combine two things:

1. **Atomic status update**: Change `updateInvitation` in `adapter.ts` to conditionally update only if the current status is `"pending"`:
   ```ts
   // adapter.ts - updateInvitation
   const invitation = await adapter.update({
     model: "invitation",
     where: [
       { field: "id", value: data.invitationId },
       { field: "status", value: "pending" }, // Only update if still pending
     ],
     update: { status: data.status },
   });
   // If invitation is null, another request already accepted/rejected it
   ```

2. **Guard before member creation**: Add an existence check or use an upsert:
   ```ts
   // crud-invites.ts - before createMember
   const existingMember = await adapter.findMemberByOrgId({
     userId: session.user.id,
     organizationId: invitation.organizationId,
   });
   if (existingMember) {
     return ctx.json({ invitation: acceptedI, member: existingMember });
   }
   ```

Backend

### Additional context

This was discovered in a Next.js app where React Strict Mode double-mounts components in development, causing the `useEffect` that calls `acceptInvitation` to fire twice concurrently. While the client-side can be guarded with a ref, the server should be resilient to concurrent calls regardless.

The same pattern (non-atomic read-then-write without constraints) was previously identified in the JWK plugin (#5663) and invitation resend logic (#3507).

GautamBytes:
## Summary

- prevent duplicate organization members when `acceptInvitation` is called more than once for the same invitation
- make repeated accepts reuse the existing member once the invitation is already accepted
- guard the invitation state transition with a conditional update and restore team/session state on replay paths

## Notes

- Closes #9154
- Adds a patch changeset for `better-auth`
- No breaking changes

---
## Summary by cubic
Make organization invitation acceptance idempotent and race-safe to prevent duplicate members. Replayed or concurrent accepts now reuse the existing member, set the active org, and restore the team in the session.

- **Bug Fixes**
  - Guard state change with expectedStatus "pending"; if already "accepted" (even if expired), return the existing member via a short retry, set active org, and restore team when applicable.
  - Prevent duplicate members on concurrent accepts by checking for an existing member before creating.
  - Add `invitationAcceptance` options to `OrganizationOptions` to tune replay lookups: `memberLookupAttempts` and `memberLookupDelayMs` (defaults 20 and 25ms).
  - Update `kysely-adapter` SQLite dialect to report affected row counts for non-returning mutations and only return `insertId` for INSERT/REPLACE.
  - Expand tests for retries, expiry replay, and concurrent accepts; add a patch changeset for `better-auth`.

<sup>Written for commit d86ccab7c59d6a37a152367ba44dbfc6121680be. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9193?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


GautamBytes:
@gustavovalverde can you review this whenever feasible?

Bekacru:
Hey @GautamBytes thanks for the PR, lets fix conflicts and also lets make attempt and other constant variables configurable instead!

@gustavovalverde @bytaesu we should get it merged

bytaesu:
cc @ping-maxwell

GautamBytes:
> Hey @GautamBytes thanks for the PR, lets fix conflicts and also lets make attempt and other constant variables configurable instead!

Done!

PR #9217, opened 2026-04-16
docs: add discord username example

xp4u1:
I updated the Discord auth docs to add an optional username flow: it explains that if you want to store the Discord username you should map it from the provider profile with `mapProfileToUser`, add username to `user.additionalFields`, and infer additional fields in `auth-client.ts` so the field is typed on the client, with a link to the TypeScript docs for that inference step.

Currently, the needed steps are spread across different docs, and this example puts the server mapping and additional fields in one clear snippet.

---
## Summary by cubic
Added an optional Discord username example in the auth docs to make storing usernames straightforward. It shows mapping `profile.username` via `mapProfileToUser`, defining `user.additionalFields.username`, and how to infer the field on the client in `auth-client.ts` (with a link to the TypeScript docs).

<sup>Written for commit b7b42cc923b6cd00a7b447fdf1fefa3d7bcce3ae. Summary will update on new commits.</sup>

### DIFF
--- a/docs/content/docs/authentication/discord.mdx
+++ b/docs/content/docs/authentication/discord.mdx
@@ -74,3 +74,35 @@ export const auth = betterAuth({
 ```
 
 **Note:** The `permissions` parameter only works when the `bot` scope is included in your OAuth2 scopes. Read more about [Discord bot permissions](https://discord.com/developers/docs/topics/permissions).
+
+### Usernames (Optional)
+
+To include the Discord username in your user model, add it to `additionalFields` and map it from the provider profile.
+
+```ts title="auth.ts"
+import { betterAuth } from "better-auth"
+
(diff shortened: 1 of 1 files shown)


ping-maxwell:
I don't think it makes sense to just have an example for Discord, there are many other providers which might have similar use-case.
Our docs can navigate people to finding out it's possible to do this, but maybe it's not straight forward enough.
We need a more generalized example to show-case this.

bytaesu (commented):
I agree with @ping-maxwell's point. This is just one use case, and trying to cover every possible usage would make the documentation less concise and clear.

I'll close this PR.

bytaesu:
+) We already cover this use-case here https://better-auth.com/docs/concepts/oauth#mapprofiletouser

PR #9227, opened 2026-04-17
feat(core): add `mapConcurrent` bounded-concurrency utility

bytaesu:
## Follow-up candidates

Other unbounded fan-out sites that could adopt `mapConcurrent` later.
Not included here to keep this PR scoped to the utility introduction:

### Noteworthy (same shape as the api-key migration)
- https://github.com/better-auth/better-auth/blob/main/packages/core/src/db/adapter/factory.ts#L1225  
  — parallel `transformOutput` over `findMany` results

- https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/db/internal-adapter.ts#L68  
  — per-session secondary-storage reads

### Size-dependent
- https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/api/routes/session.ts#L887  
  — parallel session deletions on revoke-all

- https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/organization/adapter.ts#L360  
  — parallel `deleteMany` over teams

- https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/custom-session/index.ts#L82  
  — user callback per session

### DIFF
--- a/.changeset/bounded-map-concurrent.md
+++ b/.changeset/bounded-map-concurrent.md
@@ -2,4 +2,4 @@
 "@better-auth/core": patch
 ---
 
-chore(core): add `mapConcurrent` bounded-concurrency utility at `@better-auth/core/utils/async`
+feat(core): add `mapConcurrent` bounded-concurrency utility at `@better-auth/core/utils/async`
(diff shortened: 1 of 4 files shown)


bytaesu:
> [!NOTE]
> Since this is extracted from and improves the utility in api-key (`better-auth`), I only added a changeset for `core`.

PR #9235, opened 2026-04-18
fix(cookies): preserve `partitioned` attribute on set-cookie round-trip

### LINKED ISSUE #9231: `customSession` plugin drops `Partitioned` cookie attribute on session refresh
### Reproduction

1. Set `defaultCookieAttributes: { partitioned: true, sameSite: "none", secure: true }` and use `customSession` plugin
2. Sign in (cookie is correctly `Partitioned`)
3. Wait for `updateAge` to elapse, call `/get-session` -- session refresh re-sets cookie **without** `Partitioned`
4. Browser now has two same-name cookies (one partitioned, one not)
5. Sign out -- `expireCookie` correctly clears the partitioned cookie but the unpartitioned one survives (or vice versa)
6. Sign in again -- new partitioned cookie is set, but the ghost unpartitioned cookie from step 3 still exists
7. `/get-session` receives both cookies, picks the wrong one, returns `null`

Root cause in `custom-session/index.ts` L112-124 -- the `ctx.setCookie` reconstruction explicitly maps known attributes but omits `partitioned`:

```typescript
ctx.setCookie(name, attrs.value, {
    maxAge: attrs["max-age"],
    expires: attrs.expires,
    domain: attrs.domain,
    path: attrs.path,
    secure: attrs.secure,
    httpOnly: attrs.httponly,
    sameSite: attrs.samesite,
    // partitioned not forwarded
});
```

`parseSetCookieHeader` correctly captures `partitioned` via the default switch case, but the reconstruction ignores it.

Session breaks after refresh + re-login cycle when using partitioned cookies. All cookie attributes -- including `Partitioned` -- should be preserved when forwarding Set-Cookie headers.

1.4.22 (also confirmed unfixed on `main` and v1.6.5)

### System info

N/A (code-level omission, not environment-specific)

Backend

```typescript
    advanced: {
        defaultCookieAttributes: {
            partitioned: true,
            sameSite: "none",
            secure: true,
            httpOnly: true,
        },
        crossSubDomainCookies: { enabled: true },
    },
    plugins: [
        customSession(async (session) => {
            return { ...session /* custom fields */ };
        }),
    ],
});
```

### Additional context

Introduced in v1.4.19 (PR #7879) which changed cookie forwarding from raw `ctx.setHeader("set-cookie", ...)` to a parse-and-reconstruct approach via `parseSetCookieHeader` + `ctx.setCookie`. The v1.4.5 raw pass-through approach did not have this issue.

Fix: either spread remaining parsed attributes (`...rest`) into the `ctx.setCookie` call, or revert to raw Set-Cookie header forwarding via `getSetCookie()`.

bytaesu:
- Closes #9231


packages/better-auth/src/cookies/cookie-utils.ts
```diff
@@ -14,10 +14,24 @@ export interface CookieAttributes {
 	path?: string | undefined;
```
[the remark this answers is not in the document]
bytaesu: no

packages/better-auth/src/integrations/svelte-kit.ts
```diff
@@ -80,14 +80,10 @@ export const sveltekitCookies = (
 
 							for (const [name, { value, ...ops }] of parsed) {
 								try {
+									const cookieOptions = toCookieOptions({ value, ...ops });
```
ping-maxwell: we destructure `{ value, ...ops }` but then resemble it at `toCookieOptions({ value, ...ops });` - seems uneeded?

ping-maxwell (dismissed):
the rest LGTM

packages/better-auth/src/integrations/svelte-kit.ts
bytaesu: Nice 👍

PR #9240, opened 2026-04-18
feat(username): add immutable username option

adrianmxb:
Add a new configuration option `immutableUsername` to the username plugin that prevents users from updating their username after it has been set. This allows for use cases where usernames should be selected once and remain immutable.

### Changes
- **Error Codes**: Added `USERNAME_IS_IMMUTABLE` error code for when username updates are blocked
- **Configuration**: Added `immutableUsername?: boolean` option to `UsernameOptions` (defaults to `false`)
- **Validation Logic**: Added middleware hook that checks for username immutability during user updates
- **Tests**:
  - Sign-up with immutable usernames enabled
  - Blocking username updates to different values
  - Allowing updates to the same username value (no-op)
  - Allowing displayUsername updates even when username is immutable

### Behavior
- When `immutableUsername: true`, users cannot change their username after initial creation
- Users can still update other profile fields like `displayUsername`
- Setting the same username value is allowed
- Backward compatible - defaults to `false` to maintain existing behavior

---
## Summary by cubic
Add an `immutableUsername` option to the `better-auth` `username` plugin to lock a user’s username after it’s first set. Users can still set it once if unset and update `displayUsername`; default behavior stays the same.

- **New Features**
  - Added `immutableUsername?: boolean` to `UsernameOptions` (default `false`).
  - Enforced immutability on `/update-user`: allow first set if unset, allow same-value updates, block different-value changes; check runs before availability lookup.
  - Added `USERNAME_IS_IMMUTABLE`, tests for sign-up/blocked/same-value/`displayUsername` updates, and docs with an “Immutable Username” section.

<sup>Written for commit 6ec6cf2b12de327715c026e10d588128276a9c90. Summary will update on new commits.</sup>


packages/better-auth/src/plugins/username/index.ts
```diff
@@ -555,6 +562,19 @@ export const username = (options?: UsernameOptions | undefined) => {
...
+								const hasUsername = !!session?.user.username;
+								const usernamesDiffer =
+									session?.user.username !== normalizedUsername;
+								if (hasUsername && usernamesDiffer) {
+									throw APIError.from(
+										"BAD_REQUEST",
+										ERROR_CODES.USERNAME_IS_IMMUTABLE,
+									);
+								}
+							}
```
ping-maxwell: Immutability check should come before the uniqueness DB query

ping-maxwell: Also there is a redundant `getSessionFromCtx` call, earlier code already has it in an `if` statement, pull that out and use it for both `if` statements.

ping-maxwell (changes requested):
Please add documentation.

packages/better-auth/src/plugins/username/index.ts
adrianmxb: Thanks for the review, moved the check up.

>Also there is a redundant getSessionFromCtx call, earlier code already has it in an if statement, pull that out and use it for both if statements. 

I kept this separated in each of the if-blocks on purpose as I think it keeps things a bit cleaner, given that the hook itself runs for both `update-user` and `sign-up`, but peaceful with the change, moved the session out and conditionally set it for `update-user`.

ping-maxwell (approved):

PR #9244, opened 2026-04-19
fix(oauth-provider): userinfo Authorization from ctx.headers for auth.api

### LINKED ISSUE #8806: auth.api.oauth2UserInfo returns body: { error_description: "request not found", error: "invalid_request" },

### To Reproduce

1. Call auth.api.oauth2UserInfo

Currently it returns:
`
[APIError] {
  status: "UNAUTHORIZED",
  body: { error_description: "request not found", error: "invalid_request" },
  headers: {},
  statusCode: 401
}
`

Expected behavoir: return the user information like when calling via fetch("/oauth2/userinfo")

1.5.4

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 10 IoT Enterprise LTSC 2021",
    "release": "10.0.19044",
    "cpuCount": 28,
    "cpuModel": "Intel(R) Core(TM) i7-14700K",       
    "totalMemory": "31.77 GB",
    "freeMemory": "15.38 GB"
  },
}
```

Backend

```typescript
const auth = betterAuth({
  disabledPaths: [
    "/token",
  ],
  plugins: [
    jwt(),
    oauthProvider({ 
      loginPage: "/sign-in", 
      consentPage: "/consent", 
      // ...other options
    }) 
  ],
});
```

### Additional context

_No response_

TanishValesha:
## Summary
`oauth2UserInfo` read `Authorization` only from `ctx.request.headers`. Direct `auth.api.oauth2UserInfo({ headers })` calls often have no `ctx.request`, which returned `invalid_request` / "request not found"-style behavior.

Resolve the header source as `ctx.request?.headers ?? ctx.headers` so programmatic calls match `GET /oauth2/userinfo` over HTTP.

## Test plan
- [x] `pnpm exec vitest run src/userinfo.test.ts` (from `packages/oauth-provider`)
- [x] `pnpm typecheck` (repo root)

Fixes #8806

---
## Summary by cubic
In `@better-auth/oauth-provider`, the OAuth2 userinfo endpoint now reads the `Authorization` header from `ctx.headers`, so direct `auth.api.oauth2UserInfo({ headers })` matches `GET /oauth2/userinfo` and no longer requires `ctx.request` (fixes #8806). When the header is missing, it returns 401 `invalid_request` ("authorization header not found"); tests cover both cases.

<sup>Written for commit e8ac8e7ed5de56eb0d556a168f05359c1de754ce. Summary will update on new commits.</sup>


TanishValesha:
Hi @gustavovalverde  , just a gentle reminder to take a look at this PR when you get a chance. Thanks!

packages/oauth-provider/src/userinfo.ts
```diff
@@ -37,14 +37,15 @@ export async function userInfoEndpoint(
...
+	if (!headerBag) {
 		throw new APIError("UNAUTHORIZED", {
-			error_description: "request not found",
+			error_description: "authorization header not found",
 			error: "invalid_request",
 		});
 	}
 
-	const authorization = ctx.request.headers.get("authorization");
+	const authorization = headerBag.get("authorization");
```
gustavovalverde: The `headerBag` intermediate and the `if (!headerBag)` guard aren't needed. `ctx.headers?.get("authorization")` handles both HTTP and `auth.api.*` paths directly; that is the form used in `packages/api-key/src/index.ts:141`, `packages/scim/src/middlewares.ts:14`, and even elsewhere in this package at `authorize.ts:55`. The existing `!token?.length` check a few lines below already covers the missing-header case, so the guard is redundant.

gustavovalverde (dismissed):
Fix works, but can be simplified to the canonical header-reading pattern used elsewhere in this codebase. One inline comment.

Separate follow-up (not this PR): `extractClientCredentials` in `utils/index.ts` has the same `ctx.request?.headers...` pattern without a fallback. Worth a small PR of its own.

packages/oauth-provider/src/userinfo.ts
TanishValesha: @gustavovalverde Thanks for the review! 
I've simplified the implementation to use `ctx.headers?.get("authorization")` and removed the redundant guard as suggested.

Also noted the point about `extractClientCredentials` - I’ll open a separate PR for that.

PR #9246, opened 2026-04-19
fix(core): use package imports map for internal async_hooks reference

sunnymodi:
Internal modules under `packages/core/src/context/` imported the package's own subpath via `@better-auth/core/async_hooks`. Self- referential subpath imports fail to resolve under some install layouts (e.g. Bun's content-addressed cache in production install mode used by Vercel's Bun runtime), surfacing as

    TypeError: Requested module is not instantiated yet

or

    Cannot find module '@better-auth/core/async_hooks'

Switch the internal references to `#async_hooks` and add a matching `imports` map to `packages/core/package.json`. The map preserves the existing edge/browser/default condition split, so published consumers (still resolving the public `./async_hooks` subpath) are unaffected.

---
## Summary by cubic
Switch internal async hooks imports to `#async_hooks` and add an `imports` map in `@better-auth/core`. This fixes self-referential subpath errors in Bun production installs (e.g., Vercel) without changing the public `./async_hooks` export.

- **Bug Fixes**
  - Updated `endpoint-context.ts`, `request-state.ts`, `transaction.ts`, and `global.ts` to import from `#async_hooks`.
  - Added `imports.#async_hooks` to `packages/core/package.json` mirroring edge/browser/default conditions (incl. types and `dev-source`); public `./async_hooks` remains unchanged.

<sup>Written for commit 0c072d3852d5d8f5590f712e65a355f1e898c3d2. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/internal-imports-async-hooks.md
+++ b/.changeset/internal-imports-async-hooks.md
@@ -1,0 +1,7 @@
+---
+"@better-auth/core": patch
+---
+
+fix(core): use package imports map for internal async_hooks reference
+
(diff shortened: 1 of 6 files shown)


ping-maxwell (changes requested):
`tsdown.config.ts` `neverBundle` is stale and not updated

```ts
		neverBundle: ["@better-auth/core/async_hooks"],
```

Issue #4751, opened 2025-09-18
[Prisma Adapter] Organization plugin ignores advanced.database.generateId = false parameter


mdivani:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

set generateId to false, add organization plugin and try creating organization with custom Id from organizationHook - beforeCreateOrganization

### Current vs. Expected behavior

I need to provide custom ids for some tables, so I have disabled Id generation from config, es per the docs

```
{
 ....
  advanced: {
    database: {
      generateId: false,
    },
  },
```

still when trying to add custom Id from organization hooks

```
        beforeCreateOrganization: async ({ organization, user }) => {

          const employer = await employeeClient.createEmployer({
            name: organization.name || '',
          });

          return {
            data: {
              ...organization,
              id: employer.id,
            },
          };
        },
```

I get the error:

`[Prisma Adapter] - You are trying to create a record with an id. This is not allowed as we handle id generation for you, unless you pass in the `forceAllowId` parameter. The id will be ignored.`

Id is not actually created since config does work, but id I'm passing is also ignored and I end up with prisma error for missing id.

Also would try forceAllowId, but no idea where to put that parameter.

### What version of Better Auth are you using?

1.3.11

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 23.5.0: Wed May  1 20:16:51 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T8103",
    "release": "23.5.0",
    "cpuCount": 8,
    "cpuModel": "Apple M1",
    "totalMemory": "8.00 GB",
    "freeMemory": "0.15 GB"
  },
  "node": {
    "version": "v22.17.1",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.2"
  },
  "frameworks": [
    {
      "name": "express",
      "version": "5.1.0"
    }
  ],
  "databases": [
    {
      "name": "@prisma/client",
      "version": "6.12.0"
    }
  ],
  "betterAuth": {
    "version": "1.3.11",
    "config": {
      "appName": "Flighter Group",
      "baseURL": "https://api.dev.flightergroup.com",
      "user": {
        "additionalFields": {
          "resetPassword": {
            "type": "boolean",
            "required": false
          }
        }
      },
      "databaseHooks": {
        "user": {
          "create": {}
        },
        "session": {
          "create": {}
        }
[...]

PR #9253, opened 2026-04-20
fix(organization): allow passing id through `beforeCreateTeam` and `beforeCreateInvitation`

baptisteArno:
## Summary

Adds `forceAllowId: true` to `adapter.createTeam` and `adapter.createInvitation`, mirroring the fix in #4765 for `createOrganization`. Ids returned from the corresponding hooks (`beforeCreateTeam`, `beforeCreateInvitation`) now survive the DB insert instead of being silently replaced by a generated one.

## Motivation

Symmetry with #4765. Useful for optimistic-UI clients (e.g. [TanStack DB](https://tanstack.com/db)) that want a client-generated id to round-trip through the server so React keys stay stable across the optimistic insert → refetch cycle. Today this works for organizations but not for teams or invitations.

## Changes

- `adapter.createTeam` → pass `forceAllowId: true`
- `adapter.createInvitation` → pass `forceAllowId: true`
- Regression tests for both hooks returning a custom `id`


packages/better-auth/src/plugins/organization/adapter.ts
ping-maxwell: remove the `omit`

ping-maxwell: remove `omit`

packages/better-auth/src/plugins/organization/routes/crud-org.test.ts
```diff
@@ -445,6 +445,80 @@ describe("organization hooks", async () => {
 		expect(afterCreateTeam).toHaveBeenCalled();
 	});
 
+	it("should allow passing id through `beforeCreateTeam`", async () => {
```
ping-maxwell, line 451: please add comments above the `it()` with JSdoc to include a `@see <issue-url>`

ping-maxwell (changes requested):

packages/better-auth/src/plugins/organization/routes/crud-org.test.ts
baptisteArno: Which issue URL are you referring to?

The original issue seems to be: https://github.com/better-auth/better-auth/issues/4751, but I find that it's not very helpful as a context for a comment here.

Let me know

ping-maxwell: you can use https://github.com/better-auth/better-auth/issues/9288
just opened it for tracking purposes.

baptisteArno:
rtr @ping-maxwell, thank you!

ping-maxwell (approved):

PR #9263, opened 2026-04-21
feat(api-key): add per-key IP allowlist

hgaddipati1118:
## Summary

Adds an optional `ipAllowlist` field on API keys. When set to a non-empty array of IPs or CIDR ranges (IPv4 + IPv6), verification rejects requests whose client IP isn't in the list with a new `IP_NOT_ALLOWED` error. A rejected request does not consume `remaining`, `requestCount`, or update `lastRequest` — a leaked key can't burn quota from a wrong-IP attacker.

Prior art: AWS IAM `aws:SourceIp`, Stripe restricted keys, GCP API keys, GitHub fine-grained PATs, Unkey all ship this. No prior issue/PR in this repo that I could find.

## Design

**Schema.** `ipAllowlist` is stored as a JSON-stringified `string[]` on the `apikey` table — mirrors exactly how `permissions` is stored. One new nullable column, additive migration, existing keys unaffected.

**Enforcement lives in `validateApiKey`.** Placed *after* the permissions check and *before* any quota/rate-limit mutation, so a denied IP never touches counters. Putting it in the shared function (not just the `/api-key/verify` route) means the `enableSessionForAPIKeys` session-hook path is also covered — there's a regression test for that.

**Client IP resolution:**
1. Optional `clientIp` field on the verify body (for callers behind a gateway — the request hitting `/api-key/verify` may be the gateway itself).
2. Falls back to deriving from `ctx.request` using the existing `advanced.ipAddress.ipAddressHeaders` config.
3. A dedicated `getClientIpExact` helper is used so matching is *exact* — reusing `getIp` would let a configured `ipv6Subnet` silently collapse IPv6 addresses to a /64 and widen matches. **Fails closed** if the key has an allowlist but no IP can be resolved.

**CIDR matcher** is pure, stdlib-only (BigInt), inlined into the existing `packages/api-key/src/utils.ts`. Handles IPv4, IPv6, and IPv4-mapped-IPv6 against v4 CIDRs via `normalizeIP`. No new files, no new dependencies.

**Input validation.** `create` and `update` validate each entry with `isValidCidr` and reject empty arrays (ambiguous: no-op vs deny-all). Use `ipAllowlist: null` on update to clear. `ipAllowlist` is a server-only property, same as `permissions`.

## What changed

- `schema.ts` — new `ipAllowlist` field on `apikey`
- `types.ts` — `ipAllowlist?: string[] \| null` on `ApiKey`
- `utils.ts` — `parseCidr`, `isValidCidr`, `matchesAnyCidr`, `getClientIpExact`
- `error-codes.ts` — `IP_NOT_ALLOWED`, `INVALID_IP_ALLOWLIST_ENTRY`, `EMPTY_IP_ALLOWLIST`
- `routes/create-api-key.ts`, `routes/update-api-key.ts` — accept + validate + persist + server-only guard
- `routes/verify-api-key.ts` — check in `validateApiKey`, new optional `clientIp` on body
- `api-key.test.ts` — 9 new tests (see below)
- `docs/content/docs/plugins/api-key/index.mdx` — usage + trusted-proxy + gateway caveats + cloud-egress note
- `.changeset/api-key-ip-allowlist.md` — minor bump for `@better-auth/api-key`

No existing tests modified.

## Test plan

- [x] `pnpm --filter=@better-auth/api-key exec vitest src/api-key.test.ts --run` — 175/175 pass (9 new + 166 unchanged)
- [x] `pnpm --filter=@better-auth/api-key exec vitest src/org-api-key.test.ts --run` — 10/10 pass
- [x] `pnpm typecheck` — clean across the repo

New tests cover:
- Create with allowlist round-trips through response
- Invalid CIDR rejected at create (`INVALID_IP_ALLOWLIST_ENTRY`)
- Empty array rejected at create (`EMPTY_IP_ALLOWLIST`)
- Verify with matching `clientIp` succeeds
- Verify with non-matching `clientIp` returns `IP_NOT_ALLOWED` **and** `remaining` / `requestCount` / `lastRequest` are unchanged
- Exact IPv6 match (guards the `ipv6Subnet` footgun)
- v4 CIDR against v4-mapped-IPv6 client IP
- `ipAllowlist: null` on update clears the restriction
- `enableSessionForAPIKeys` session-hook path rejects non-allowlisted IPs (regression guard)

## Why not use `customAPIKeyValidator`?

It can't access per-key state (you'd have to reach into `ctx.context.adapter` yourself), can't compose with rate-limit/quota semantics correctly, and doesn't serialize with the key — so it's not a real substitute for a first-class field.

---
## Summary by cubic
Adds per-key IP allowlists to `@better-auth/api-key` to restrict keys by IPs or CIDRs. Verification denies non-allowed IPs without consuming quota; includes a nullable `ipAllowlist` column (existing keys unaffected).

- **New Features**
  - `ipAllowlist?: string[]` on API keys (IPv4/IPv6, single IPs or CIDRs). Server-only.
  - Enforced in `validateApiKey` after permissions and before counters; returns `IP_NOT_ALLOWED`.
  - `/api-key/verify` accepts `clientIp`; otherwise resolved via configured headers. Exact IP resolver prevents widened IPv6 matches. Fails closed if unresolved.
  - Create/update validate entries (`isValidCidr`); use `ipAllowlist: null` to clear.
  - New errors: `INVALID_IP_ALLOWLIST_ENTRY`, `EMPTY_IP_ALLOWLIST`, `IP_NOT_ALLOWED`. Applies to `enableSessionForAPIKeys`.

- **Bug Fixes**
  - Harden CIDR parsing to reject malformed prefixes (e.g. trailing slash, non-digit masks).
  - Include `ipAllowlist` in 200 response schemas for create/update.
  - Run secondary-storage API-key operations in parallel for faster listing.

<sup>Written for commit 86736815b00744023235dee1b8a836b4a33f8332. Summary will update on new commits.</sup>


bytaesu:
This PR is difficult to review in its current state and has had no follow-up for a long time. For a feature this large, we should first discuss the direction and whether we want to proceed in an issue.

Please check: https://github.com/better-auth/better-auth?tab=contributing-ov-file#pull-request-guidelines

bytaesu:
Feel free to open an issue or mention me if you have further questions!

Issue #9198, opened 2026-04-15
authorize endpoint doesn't verify MFA state when issuing auth codes


mbochneak:
## Summary

The MCP plugin's `authorizeMCPOAuth` function (in `src/plugins/mcp/authorize.ts`) issues OAuth authorization codes as soon as a session exists, without checking whether multi-factor authentication (TOTP/two-factor) has been completed. This means a user who has logged in with a password or social provider but hasn't yet verified their TOTP token will receive a valid auth code that can be exchanged for access tokens.

## Expected Behavior

When the two-factor plugin is enabled and the user has MFA configured, `authorizeMCPOAuth` should not issue an authorization code until the TOTP step is verified — the same way a normal login flow gates access on MFA completion.

## Actual Behavior

`authorizeMCPOAuth` checks only for session existence:

```js
const session = await getSessionFromCtx(ctx);
if (!session) {
    // redirect to login page
}
// ... immediately generates auth code and redirects
```

There is no check for `twoFactorVerified`, `totpVerifiedAt`, or any MFA-related session state. The auth code is generated at line 80 and the redirect happens at line 110 without any MFA gate.

## Impact

This is a security concern for applications that rely on MFA — particularly in compliance-sensitive environments (e.g., CJIS, HIPAA). The OAuth MCP flow effectively bypasses the MFA requirement, issuing tokens that authenticate as the user without the second factor.

## Current Workaround

Server-side applications can check for a TOTP-verified session at MCP request time (after the OAuth flow completes), rejecting requests where MFA wasn't completed:

```ts
const verifiedSession = await prisma.session.findFirst({
  where: {
    userId: mcpSession.userId,
    totpVerifiedAt: { not: null },
    expiresAt: { gt: new Date() },
  },
});
if (!verifiedSession) {
  res.status(403).json({ error: 'totp_required' });
  return;
}
```

This works but is an application-level workaround — the auth code should not be issued in the first place.

## Suggested Fix

In `authorizeMCPOAuth`, after confirming the session exists, check whether the user has two-factor enabled and whether the session has completed TOTP verification. If MFA is required but not completed, redirect to the login/MFA page instead of issuing the auth code.

## Environment

- better-auth version: latest (as of April 2026)
- Plugins: `mcp()`, `twoFactor()`, `organization()`
- MFA provider: TOTP (authenticator app)

Thank you for the excellent work on better-auth and the MCP plugin!

bytaesu:
[...]

Issue #7607, opened 2026-01-25
session_data cookie not regenerating after expiry when using twoFactor plugin


Dnori96:
### Is this suited for github?

- [x] Yes, this is suited for github

### Little description of the bug

When using `cookieCache` with the `twoFactor` plugin, the `session_data` cookie is not regenerated after it expires. The `session_token` cookie persists, but once `session_data` expires (after `maxAge`), the user is logged out because the session cannot be restored.

Additionally, disabling `cookieCache` entirely causes TOTP verification to fail with "invalid two factor cookie" error, suggesting the `twoFactor` plugin depends on `cookieCache` to store the pending 2FA state.

### To reproduce

1. Configure better-auth with `cookieCache` and `twoFactor` plugin
2. Sign in with a user that has 2FA enabled
3. Complete TOTP verification successfully
4. User is redirected to a protected page, both `session_token` and `session_data` cookies are set
5. Wait for `cookieCache.maxAge` to expire (e.g., depending of the maxAge in auth config)
6. Refresh the page or navigate
7. `session_data` cookie disappears and is never regenerated
8. User is logged out despite `session_token` still existing

### Current vs. Expected behavior

After `session_data` expires, calling `getSession()` should:
1. Validate the `session_token` against the database
2. Regenerate the `session_data` cookie with fresh cached data

### What version of Better Auth are you using?

1.4.17

### System info

```bash
"node": {
    "version": "v22.20.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.7.0"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "^16.0.7"
    },
    {
      "name": "react",
      "version": "^19.2.3"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.16.3"
    },
    {
      "name": "@prisma/client",
      "version": "^7.3.0"
    }
  ],
  "betterAuth": {
    "version": "1.4.17",
    "config": null,
  }
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth";
import { prismaAdapter } from "better-auth/adapters/prisma";
import { prisma } from "./db/prisma";
import { nextCookies } from "better-auth/next-js";
import { twoFactor } from "better-auth/plugins";

export const auth = betterAuth({
  database: prismaAdapter(prisma, {
    provider: "postgresql",
  }),
  emailAndPassword: {
[...]

Issue #6022, opened 2025-11-16
lastLoginMethod not set when logging in with SIWE method


proof-llc:
### Is this suited for github?

- [ ] Yes, this is suited for github

### To Reproduce

1. Login with SIWE
2. Check Last Login Method
3. Result is empty string

### Current vs. Expected behavior

We get empty string instead of something like "SIWE"

### What version of Better Auth are you using?

1.4.0 beta

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:40 PDT 2025;",
    "release": "24.6.0",
    "cpuCount": 16,
    "cpuModel": "redacted",
    "totalMemory": "128.00 GB",
    "freeMemory": "20.62 GB"
  },
  "node": {
    "version": "v24.7.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.5.1"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "0.1.0",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
  lastLoginMethod({
    storeInDatabase: true,
  })
});
```

rovertrack:
u must declare it inside the plugins array!

```
import { betterAuth } from "better-auth"
import { lastLoginMethod } from "better-auth/plugins"

export const auth = betterAuth({
    plugins: [
        lastLoginMethod({
            storeInDatabase: true
        })
    ]
})
```

proof-llc:
I did that. but it appears to set the DB when logging in with other ways but not SIWE.

proof-llc:
Hi, it is still relevant

gustavovalverde:
Closing as this was fixed by #6027, merged on 2025-11-16 and present on the current `main` branch. The last-login-method plugin now recognizes SIWE and updates the stored value on session creation.

Issue #4491, opened 2025-09-06
`dontRememberMe` does not work when cookie cache is enabled


widavies:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Enable cookie cache
2. Sign in with `rememberMe: false`

<img width="1360" height="202" alt="Image" src="https://github.com/user-attachments/assets/e62cee0f-4b5c-4e4d-b73a-20e1ee27ab3c" />

### Current vs. Expected behavior

When `rememberMe: false`, closing the tab should sign the user out, but it does not.

### What version of Better Auth are you using?

1.3.8

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Home",
    "release": "10.0.26100",
    "cpuCount": 32,
    "cpuModel": "AMD Ryzen 9 9950X 16-Core Processor            ",
    "totalMemory": "61.65 GB",
    "freeMemory": "26.73 GB"
  },
  "node": {
    "version": "v24.4.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.4.2"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "^15.5.2"
    },
    {
      "name": "react",
      "version": "^19.1.1"
    },
    {
      "name": "hono",
      "version": "^4.9.6"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "0.1.0",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
  session: {
    cookieCache: {
      enabled: true,
      maxAge: 5 * 60, // Cache duration in seconds
    },
  },

});
```

### Additional context

I believe this bug is caused because the cookie expiration for the `session_data` cookie is _not_ set to `Session`

<img width="1360" height="202" alt="Image" src="https://github.com/user-attachments/assets/d0426659-f813-42f5-a841-79c085ac4809" />

widavies:
[...]

PR #9278, opened 2026-04-21
refactor(two-factor)!: rewrite sign-in challenges around paused attempts

### LINKED ISSUE #9132: twoFactor endpoints use cookie-cached session; generateBackupCodes fails with TWO_FACTOR_NOT_ENABLED after verifyTotp

### Reproduction

Config (`session.cookieCache.enabled: true`, `twoFactor` plugin, credential-provider user):

```ts
betterAuth({
  session: { cookieCache: { enabled: true, maxAge: 5 * 60 } },
  plugins: [twoFactor({ issuer: "App", allowPasswordless: true })],
  // ...
});
```

Steps:

1. User signs in via email/password. The `session_data` cookie is set with `user.twoFactorEnabled=false`.
2. User calls `authClient.twoFactor.enable({ password })`. The `enable` endpoint does **not** call `setSessionCookie` (unless `skipVerificationOnEnable: true`), so the cookie still holds `twoFactorEnabled=false`.
3. Client navigates to a TOTP-verification page and calls `authClient.twoFactor.verifyTotp({ code })`. Server flips `twoFactorEnabled=true` in the DB and calls `setSessionCookie` with the refreshed user. Response includes `Set-Cookie: better-auth.session_data=…`.
4. Client navigates to `/backup-codes` via `window.location.assign` (or any mechanism that triggers a new top-level request).
5. Client calls `authClient.twoFactor.generateBackupCodes({ password })`.
6. Server middleware chain: `sessionMiddleware → getSessionFromCtx` returns the cached session object (cookie-cache path, `cookieRefreshCache` not forced off). `ctx.context.session.user.twoFactorEnabled` is whatever the cookie held on that request.

In the browser, we observed step 6 reading `twoFactorEnabled=false` and `generateBackupCodes` throwing `TWO_FACTOR_NOT_ENABLED`, even though the DB row has been `true` since step 3. Happens reliably on the end-to-end flow under Playwright/Chromium; works in `two-factor.test.ts` only because the test harness manually mirrors every `Set-Cookie` into the next request via `sessionSetter(headers)`.

**Current**: `generateBackupCodes` (and any other twoFactor endpoint that mutates state based on `user.twoFactorEnabled` / `verified`) uses `sessionMiddleware`, which prefers the `session_data` cookie cache. Immediately after enrollment, the cookie can be stale relative to the DB, so the endpoint throws `TWO_FACTOR_NOT_ENABLED`. The only workaround today is client-side: manually expire the cookie and force a DB read.

```ts
// Workaround
invalidateSessionDataCache();
await authClient.getSession({ query: { disableCookieCache: true } });
const result = await authClient.twoFactor.generateBackupCodes({ password });
```

**Expected**: twoFactor endpoints that gate on `user.twoFactorEnabled` should read fresh state from the DB, not from the cookie cache. The library already exports `sensitiveSessionMiddleware` (`packages/better-auth/src/api/routes/session.ts:586`) with the exact contract we need — its own JSDoc says "This should be used for sensitive operations like password changes, account deletion, etc. to ensure that revoked sessions cannot be used even if they're still cached in cookies."

Swapping `sessionMiddleware → sensitiveSessionMiddleware` on `generateBackupCodes` (and arguably `twoFactor.disable`, `verifyBackupCode`, `verifyTotp` when not a sign-in) would close this cleanly without forcing every caller to hand-roll a cache-bust.

1.6.0

### System info

```bash
Node: v24.12.0
pnpm: 10.33.0
OS: Darwin 25.2.0
Framework: Next.js 16.1.7, React 19.2.4
Adapter: Drizzle (libSQL / Turso)
better-auth: 1.6.0
```

- Backend
- Package

```typescript
import { twoFactor } from "better-auth/plugins";

  session: {
    expiresIn: 60 * 60 * 24 * 7,
    updateAge: 60 * 60 * 24,
    cookieCache: { enabled: true, maxAge: 5 * 60 },
  },
  plugins: [
    twoFactor({ issuer: "App", allowPasswordless: true }),
  ],
});
```

### Additional context

- `twoFactor.enable` also skips `setSessionCookie` when `skipVerificationOnEnable` is falsy (the default), which is the first place the cookie and DB can drift. Whether that needs fixing is separate, since `verifyTotp` does refresh the cookie afterwards; the more pressing issue is that the subsequent call relies on cookie-cached state.
- Relevant source pointers:
  - `packages/better-auth/src/plugins/two-factor/backup-codes/index.ts:409` — `use: [sessionMiddleware]`
  - `packages/better-auth/src/plugins/two-factor/backup-codes/index.ts:446` — `if (!user.twoFactorEnabled) throw TWO_FACTOR_NOT_ENABLED`
  - `packages/better-auth/src/api/routes/session.ts:586` — `sensitiveSessionMiddleware`
- Happy to send a PR that switches the gating endpoints to `sensitiveSessionMiddleware` if the maintainers agree with that direction.

gustavovalverde:
## Summary

This PR rewrites Better Auth's sign-in and 2FA flow around a paused sign-in model and a method-first two-factor architecture.

Instead of creating a session first and then trying to challenge or clean it up in an after-hook, sign-in now resolves to either a finalized session or a paused challenge, and the session is only published once the challenge completes.

## Why

The previous matcher-based enforcement was the wrong center of gravity:

- it could silently miss new sign-in surfaces
- it created sessions before policy ran and then tried to clean them up afterwards
- it mixed method state with user-level 2FA enablement, which led to OTP/TOTP inconsistencies and lockout-prone flows
- it relied on cookie-cached session state for DB-authoritative 2FA configuration reads

This rewrite makes the paused attempt the source of truth, keeps the factor chain on the session, and leaves a safer pattern for other plugins and future auth flows to copy.

## What changed

- replace the legacy `twoFactorRedirect` / `challengeId` shape with the generic sign-in challenge envelope and persisted `signInAttempt`
- refactor the 2FA plugin around methods, explicit verification state, recovery codes, and shared verification resolution
- make paused sign-in attempts atomic and identity-aware, and add per-attempt verification limits and lockout
- record `session.amr` and project AMR into OIDC `id_token`; collapse `last-login-method` onto that source of truth
- move enforcement to sign-in-time challenge resolution and add `twoFactor({ enforcement: { decide } })` plus the nested `trustDevice` policy surface
- move sensitive 2FA configuration reads onto DB-authoritative sessions to fix stale cookie-cache failures
- update docs, demo flows, adapter/integration coverage, and end-to-end tests around the new model

## Breaking surface

This is a breaking rewrite of the 2FA contract.

- paused sign-ins now use the challenge envelope: `{ kind: "challenge", challenge: { kind, attemptId, methods } }`
- `trustDeviceMaxAge` is replaced by `trustDevice.{ maxAge, requireReverificationFor }`
- `session.amr` becomes the canonical factor chain, and `last-login-method` becomes a thin projection over it

## Fixed and superseded work

Closes #9132.

This PR also addresses the root cause behind #8627, replaces the matcher-based enforcement from #9122 that was later reverted in #9205, and supersedes the narrower follow-up in #9203 with the model-level rewrite.

## Validation

- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`
- `pnpm knip`
- `pnpm knip --production`
- `demo/nextjs` manual and Playwright validation for sign-up, enrollment, sign-in challenge, OTP verification, and recovery-code fallback

---
## Summary by cubic
Refactored sign-in and 2FA around a paused-attempt model with a standard challenge envelope and a single resolver/finalizer that only commits sessions after completion. Sessions now record `amr` and project to OIDC across all sign-in paths (including API keys), improving verification, cookies, and redirects.

- **Bug Fixes**
  - Hardened paused attempts: atomic consume to prevent concurrent completion, cross-user checks when a session exists, and expiry re-check at consume time.
  - Made recovery-code replacement atomic to avoid partial or empty sets.
  - Guarded One Tap two-factor fallback redirects with a safe-URL check.
  - OAuth proxy now catches session-hook `APIErrors` from `resolveSignInWithRedirect` and redirects to the provider’s error URL (e.g., bans), not 500.
  - Locked `acr_values_supported`/`amr_values_supported` in OIDC discovery and MCP metadata to prevent clobbering.
  - `appendSignInChallengeToURL` now replaces an existing `challenge` param and keeps details out of the URL; client proxy uses the canonical `isSignInChallenge` guard; `node:sqlite` treats CTE writes as writes.

- **Migration**
  - Replace legacy `twoFactorRedirect`/`challengeId` with the challenge envelope and handle via `isSignInChallenge*` and `appendSignInChallengeToURL`.
  - Schema: add required `session.amr`; adopt the new `signInAttempt` shape; remove `user.lastLoginMethod` and `SignInAttempt.loginMethod`.
  - Cookies/options renamed: `dont_remember` → `session_only`; `dontRememberMe` → `rememberMe`; `trustDeviceMaxAge` → `trustDevice.{ maxAge, requireReverificationFor }`; `trust_device` → `trusted_device` (default TTL now 7 days).
  - 2FA APIs are method-based; “backup codes” are now “recovery codes”; verification modes are `finalize` and `session`. Update flows that assumed a session existed before 2FA.

<sup>Written for commit d3c97722cd09333496e68f7d8480b16caf5cdd18. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9278?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


gustavovalverde:
Full scope map of what this PR fixes and supersedes, for reviewer visibility.

**Fixed (open issues):**
- Fixes #9132: sensitive 2FA endpoints read DB instead of stale cookie cache.
- Fixes #7607: `signInAttempt` row replaces `session_data` as paused-state store; 2FA survives cache expiry.
- Fixes #9198: `session.amr` is canonical, so token issuers can gate auth codes and project `amr` into `id_token`.
- Fixes #6022: AMR is written from every sign-in path, including SIWE.
- Fixes #4491: `rememberMe` polarity flipped and propagated through cookie-cache refresh.

**Superseded (open PRs):**
- Closes #9122: matcher-based enforcement replaced by paused-attempt model.
- Closes #9203: narrower identity-aware guard absorbed into the rewrite.
- Closes #7015: `trustDevice`/state options superseded by nested `trustDevice` rework and method-first model.

**Structurally addressed (already closed, kept for context):**
- #8627: method-first verification removes the OTP/TOTP enablement conflict.
- #4101: `challenge.methods` exposes TOTP/OTP in the response envelope.
- #4340: paused-attempt model applies uniformly to SSO.
- #2728: `trustDevice` cookie is refreshed with explicit `maxAge` (7-day default).
- #1564: `rememberMe` survives 2FA completion.
- #6077, #2524: session replacement is atomic; twofactor cookies travel via `FinalizedSignIn.commit`.
- #585: challenge envelope is a real response shape with `attemptId` on the body.
- #2631, #4905: resolver split returns the correct error per mode (finalize vs session).
- #4626, #4627, #5970, #4696, #6128: `lastLoginMethod` reads `session.amr[0]` and writes only after finalization.
- #348: no session is published until the challenge completes.
- #3112, #3987, #7127: `rememberMe` propagation and type consistency.
- #5739: OTP enablement, already superseded by #9057, which this PR carries forward.
- #9205: revert of #9122; this PR is the proper follow-up.

gustavovalverde:
@cubic-dev-ai review

Issue #5314, opened 2025-10-15
OIDC and SSO plugins fail to bundle in Convex environment (Node APIs used inside HTTP routes)


Wizzlesky124:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1.Follow the Better Auth Convex integration guide
2.Configure createAuth to include he oidcProvider or the sso plugin:

```
import { createClient, type GenericCtx } from "@convex-dev/better-auth";
import { convex } from "@convex-dev/better-auth/plugins";
import { betterAuth } from "better-auth";
import { oidcProvider, jwt } from "better-auth/plugins";
import { sso } from "@better-auth/sso";

export const createAuth = (ctx: GenericCtx<any>) =>
  betterAuth({
    database: createClient<any>().adapter(ctx),
    plugins: [
      convex(),
      jwt(),
      oidcProvider({ ... }),
      sso({ ... })
    ]
  });
```

4. Register the routes in http.ts:

```
import { httpRouter } from "convex/server";
import { authComponent, createAuth } from "./auth";

const http = httpRouter();
authComponent.registerRoutes(http, createAuth, { cors: true });
export default http;
```

### Current vs. Expected behavior

Current behavior:
Convex build fails with errors like:

The package "constants" wasn't found on the file system but is built into node.
It looks like you are using Node APIs from a file without the "use node" directive.
"use node" directive is not allowed for http.ts.

Even when "use node" is added to the file where createAuth is defined, Convex cannot bundle the OIDC and SSO plugins because HTTP routes cannot depend on Node APIs.

Expected behavior:
The OIDC and SSO plugins should be compatible with Convex’s bundling, or there should be a documented Node-only alternative pattern for using them (e.g. via actions).

### What version of Better Auth are you using?

1.3.27

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 10 Pro",
    "release": "10.0.19045",
    "cpuCount": 4,
    "cpuModel": "Intel(R) Core(TM) i5-7600K CPU @ 3.80GHz",
    "totalMemory": "15.93 GB",
    "freeMemory": "3.07 GB"
  },
  "node": {
    "version": "v22.14.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.2.0"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "1.3.27",
    "config": null
  }
}
```

[...]

PR #9281, opened 2026-04-21
fix(core): serve noop `./instrumentation` on browser and edge conditions

ramonclaudio:
Apply the `./async_hooks` conditional-export shape to `./instrumentation` so runtimes where the dynamic `import("@opentelemetry/api")` in `packages/core/src/instrumentation/api.ts` can't resolve via standard ECMAScript dynamic-import semantics get a noop entry that doesn't depend on runtime import resolution at all.

Convex is the reproducer. Its V8 isolate rejects bare specifiers at resolve time via `deno_core::resolve_import` and throws synchronously from the `import()` expression instead of rejecting the returned promise. `dynamic_import_callback` in `get-convex/convex-backend` `crates/isolate/src/request_scope.rs` returns `None` on resolver error, which V8 treats as a pending exception. The `.catch()` in `getOpenTelemetryAPI` never runs, and every `withSpan` call through `packages/better-auth/src/api/to-auth-endpoints.ts` and `packages/better-auth/src/db/with-hooks.ts` surfaces an uncaught error:

```
[Better Auth]: Error: Relative import path "@opentelemetry/api" not prefixed
with / or ./ or ../ from "convex:/_deps/ZJDHZ5TX.js"
```

Reproducible whenever `@convex-dev/better-auth` resolves `better-auth` to 1.6.x, which happens today for any user upgrading past the currently-published `>=1.5.0 <1.6.0` peer range or adopting the pending 1.6 migration at get-convex/better-auth#323. The breaking pattern landed in #9111, which shipped in v1.6.6. `@opentelemetry/api` itself is runtime-agnostic and ships a noop proxy when no SDK is registered, so the issue isn't about OTel support in the browser or at the edge. The issue is that the dynamic-import-with-catch pattern in `./api` isn't portable to runtimes that throw synchronously from unresolvable bare specifiers.

The fix is the same shape already used for `./async_hooks`: conditional exports that hand strict-bundler runtimes a module with no runtime import probe. Convex's esbuild bundler activates the `browser` condition automatically because it runs with `platform: "browser"` (`get-convex/convex-js` `src/cli/lib/config.ts:637`), so matching on `browser` covers Convex without a vendor-specific condition key. The `browser` route also gives pure-browser bundles an eager noop, which matches the noop-by-default behavior `@opentelemetry/api` already has when no SDK is registered.

```diff
   "./instrumentation": {
     "dev-source": "./src/instrumentation/index.ts",
     "types": "./dist/instrumentation/index.d.mts",
+    "node": "./dist/instrumentation/index.mjs",
+    "deno": "./dist/instrumentation/index.mjs",
+    "bun": "./dist/instrumentation/index.mjs",
+    "edge": "./dist/instrumentation/pure.index.mjs",
+    "workerd": "./dist/instrumentation/index.mjs",
+    "browser": "./dist/instrumentation/pure.index.mjs",
     "default": "./dist/instrumentation/index.mjs"
   }
```

- `packages/core/src/instrumentation/noop.ts`: extracted the noop OpenTelemetry API factories and the `noopOpenTelemetryAPI` singleton so `./api`'s in-process fallback has a single source of truth
- `packages/core/src/instrumentation/api.ts`: imports `noopOpenTelemetryAPI` from `./noop`, keeps the dynamic-import-with-catch logic
- `packages/core/src/instrumentation/pure.index.ts`: passthrough `withSpan` plus the `./attributes` re-export. Public surface is identical to `./index` so consumers see the same shape on every export condition
- `packages/core/package.json`: six conditions added on `./instrumentation` mirroring `./async_hooks`
- `packages/core/tsdown.config.ts`: new entry wired into the build
- `packages/core/src/instrumentation/pure.test.ts`: sync/async return, throw/reject propagation, absence of `@opentelemetry/api` at runtime, surface-equality against `./index`; `@see` tagged to #8765
- `packages/core/src/instrumentation/noop.test.ts`: `SpanStatusCode` enum, `trace.getTracer`/`getActiveSpan` shape, span-mutator safety, 3-arity `startActiveSpan` overload regression; `@see` tagged to #8765

Forward-compatible with #9274: the shared noop already includes `getActiveSpan`/`updateName` on its noop shapes so this PR and #9274 rebase cleanly on each other regardless of merge order.

Reproduced on a real app. Before, every `/api/auth/*` request returned 500 with the error above. After, the Convex function bundle resolves `./instrumentation` to `pure.index.mjs` through the `browser` condition, the noop `withSpan` passes `fn()` through, and requests round-trip normally.

`pnpm build`, `pnpm lint`, `pnpm lint:dependencies`, `pnpm lint:packages`, `pnpm lint:spell`, `pnpm format:check`, `pnpm lint:types`, `pnpm typecheck`, `pnpm typecheck:dist` all clean. `pnpm -C packages/core test` 199/199 across 14 files, with `pure.test.ts` (6 tests) and `noop.test.ts` (4 tests) alongside the 11 existing `instrumentation.test.ts` tests. `pnpm --filter better-auth exec vitest run src/instrumentation.db.test.ts src/instrumentation.endpoint.test.ts` 15/15. `grep opentelemetry packages/core/dist/instrumentation/pure.index.mjs` returns 0.

Related: #8765 (optional peer dep declaration, fix via #9111 in 1.6.6), #5314 (Convex bundling class of issues), #8458 (prior `use conditional exports to replace dynamic import hacks` in telemetry).

### DIFF
--- a/packages/core/src/instrumentation/pure.index.ts
+++ b/packages/core/src/instrumentation/pure.index.ts
@@ -1,30 +1,16 @@
-import type { OpenTelemetryAPI } from "./noop";
-import { noopOpenTelemetryAPI } from "./noop";
-
 /**
  * Noop variant of `./instrumentation` for runtimes where the dynamic
- * `import("@opentelemetry/api")` in `./api` can't resolve via the normal
- * ECMAScript dynamic-import semantics.
+ * `import("@opentelemetry/api")` in `./api` throws synchronously instead of
+ * rejecting its returned promise. Convex's V8 isolate is the reproducer: bare
+ * specifiers are rejected at resolve time in `get-convex/convex-backend`
+ * `crates/isolate/src/request_scope.rs`, so the `.catch()` in
+ * `getOpenTelemetryAPI` never runs and every `withSpan` call surfaces an
+ * uncaught error.
  *
- * Convex's V8 isolate rejects bare specifiers at resolve time and throws
- * synchronously from the `import()` expression itself rather than rejecting
- * the returned promise (see `get-convex/convex-backend`
- * `crates/isolate/src/request_scope.rs` `dynamic_import_callback`, which
- * returns `None` on resolver error; V8 treats that as a pending exception).
- * The `.catch()` in `getOpenTelemetryAPI` never runs and every `withSpan`
- * call surfaces an uncaught error.
- *
- * This module is wired through `package.json` conditional exports on the
- * `./instrumentation` subpath, matching the existing `./async_hooks` pattern.
- * The noop primitives live in `./noop` and are shared with `./api` so the
- * two entries can't drift.
+ * Public surface must stay identical to `./index` (enforced by `pure.test.ts`).
  */
 
 export * from "./attributes";
-
-export function getOpenTelemetryAPI(): OpenTelemetryAPI {
-	return noopOpenTelemetryAPI;
-}
 
 export function withSpan<T>(
 	name: string,
--- a/packages/core/src/instrumentation/pure.index.ts
+++ b/packages/core/src/instrumentation/pure.index.ts
@@ -1,73 +1,26 @@
-import type { Span, Tracer } from "@opentelemetry/api";
+import type { OpenTelemetryAPI } from "./noop";
+import { noopOpenTelemetryAPI } from "./noop";
 
 /**
  * Noop variant of `./instrumentation` for runtimes where the dynamic
- * `import("@opentelemetry/api")` in `./api` can't be relied on.
+ * `import("@opentelemetry/api")` in `./api` can't resolve via the normal
+ * ECMAScript dynamic-import semantics.
  *
- * Browsers and edge bundlers (`browser`, `edge` package.json conditions)
- * don't typically run a global `TracerProvider`, and some isolate runtimes
- * reject bare specifiers at resolve time and throw synchronously from the
- * `import()` expression itself rather than rejecting the returned promise.
- * In those cases the `.catch()` in `getOpenTelemetryAPI` never runs and
- * every call through `withSpan` surfaces an uncaught error.
+ * Convex's V8 isolate rejects bare specifiers at resolve time and throws
+ * synchronously from the `import()` expression itself rather than rejecting
+ * the returned promise (see `get-convex/convex-backend`
+ * `crates/isolate/src/request_scope.rs` `dynamic_import_callback`, which
+ * returns `None` on resolver error; V8 treats that as a pending exception).
+ * The `.catch()` in `getOpenTelemetryAPI` never runs and every `withSpan`
+ * call surfaces an uncaught error.
  *
- * This module mirrors the public shape of `./index` without touching
- * `@opentelemetry/api` at all. Wired through `package.json` conditional
- * exports on the `./instrumentation` subpath, matching the existing
- * `./async_hooks` pattern.
+ * This module is wired through `package.json` conditional exports on the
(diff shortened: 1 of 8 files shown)


jonathansamines:
>  Same trade-off applies here. OpenTelemetry is a server-side concern, browsers and edge bundles don't run a TracerProvider, and some isolate runtimes don't tolerate unresolved bare specifiers in dynamic imports.

I don't think this is true, `@opentelemetry/api` is runtime agnostic, it is designed to work regardless of the environment with an explicit design goal of decoupling the implementation from it (e.g different runtimes provide the implementation via the SDK). Furthermore, `@opentelemetry/api` is designed to be no-op by default, so unless the app or runtime brings the SDK, there is no implementation at all and acts as a simple proxy. This sounds more like a convex-specific issue.

Having said that, if we already support this sort of convex-specific workaround, these changes should be fine. So, will take a look at them!

packages/core/src/instrumentation/pure.index.ts
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+export * from "./attributes";
+
+type OpenTelemetryAPI = Pick<
+	typeof import("@opentelemetry/api"),
+	"trace" | "SpanStatusCode"
+>;
+
+function createNoopSpan(): Span {
```
jonathansamines: I think we already have this exact as part of `api.ts`. It's probably worth looking into creating a shared no-op module which is reused in both `api.ts` and here, to avoid multiple potentially diverging implementations.

ramonclaudio:
You're right @jonathansamines.`@opentelemetry/api` is runtime-agnostic and ships a noop proxy when no SDK is registered, the issue is the dynamic-import-with-catch pattern in `./api` hitting Convex's V8 isolate. The isolate throws synchronously from the `import()` expression for unresolvable bare specifiers instead of rejecting the returned promise, so the `.catch()` never runs.

`browser`/`edge` still work as the hooks because Convex's esbuild activates `browser` by default (`platform: "browser"` in `get-convex/convex-js` `src/cli/lib/config.ts:637`) and because it keeps shape parity with `./async_hooks`.

packages/core/src/instrumentation/pure.index.ts
ramonclaudio: Done. Extracted `createNoopSpan`/`createNoopTracer`/`createNoopTraceAPI`/`createNoopOpenTelemetryAPI` and the `noopOpenTelemetryAPI` singleton into `packages/core/src/instrumentation/noop.ts`. `./api.ts` and `./pure.index.ts` both import from it now, so the two entries can't drift.

packages/core/src/instrumentation/pure.index.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * This module is wired through `package.json` conditional exports on the
+ * `./instrumentation` subpath, matching the existing `./async_hooks` pattern.
+ * The noop primitives live in `./noop` and are shared with `./api` so the
+ * two entries can't drift.
+ */
+
+export * from "./attributes";
+
+export function getOpenTelemetryAPI(): OpenTelemetryAPI {
```
jonathansamines: `getOpenTelemetryAPI` is not currently part of the public API of the `/instrumentation` entrypoint. This could lead to differences between consumers depending on the export condition that gets activated, so I'd suggest we remove it for consistency with the existing entrypoint.

ramonclaudio: Done.

jonathansamines (approved):

ping-maxwell (approved):

PR #9283, opened 2026-04-21
feat: use typecheck scripts to verify packages

blakeembrey:
In https://github.com/better-auth/better-auth/pull/8702, a `typecheck:dist` script was added, but I think it might be better to rely on the existing `typecheck` scripts in each package instead.

The alternative would be to use the root `tsconfig.json` and reference every package and builds them in one pass using `tsc --build`, but I opted for this approach as it's less of a structural change to the project where nested scripts seem to be the norm. In https://github.com/better-auth/better-auth/pull/8701 it was believed the the `customConditions` was the cause of the missed type issues but it's because the root `tsconfig.json` never included that package to check. 

Having less places to configure the "packages" list seemed like a win, but the downside is they require a `typecheck` to exist.

I also collapsed the `tsconfig.test.json` files into the root versions for less configuration and keeping them separate doesn't appear to add anything to the packages using it. Every package gets a `tsconfig.build.json` which is only used to remove `customConditions` for type checking. They could be collapsed into a single `tsconfig.json` for every package, but it _may_ be useful in the future to have a separate `build` anyway for things like separating test configuration globals from the output build.

---
## Summary by cubic
Switches CI and the monorepo to per‑package typecheck scripts using `tsc --build`. Covers `docs`, demos, e2e suites and fixtures, updates tsconfig references and test imports, and removes the empty `e2e/adapter` parent.

- **Refactors**
  - Add `tsconfig.build.json` across packages; switch `typecheck` scripts to `tsc --build`; update `references` to `.build.json`; remove per‑package `tsconfig.test.json`. `docs` now runs `next typegen`, `fumadocs-mdx`, then `tsc`.
  - CI runs `pnpm -r typecheck`; remove root `typecheck`/`typecheck:dist` scripts and `tsconfig.dist-check.json`; update `knip.jsonc` to ignore the `typecheck` binary.
  - E2E/tests: per‑suite tsconfigs and `typecheck` scripts for adapter suites, `e2e/integration` (incl. `test-utils`), and smoke fixtures; switch tests to public imports (`better-auth`, `@better-auth/api-key`); add assertions for OAuth redirect URL, adjust Prisma client typing, and update API key tests to use `referenceId`.

<sup>Written for commit fc68ecacb887ac7a55abe93d0d8437290b5220de. Summary will update on new commits.</sup>


gustavovalverde:
Closing in favor of #9325. #9325's body explicitly states `Supersedes #9283`; same author, same goal, and it goes further by replacing `tsdown` with `tsc` as the compiler. The typecheck script coverage from this PR is fully absorbed.

PR #4951, opened 2025-09-27
fix: Generic OAuth and SSO do not work with IDP initiated flows.

mhornbacher:
Modifies `parseState` to safely access `c.body`, which, as the rest of the file suggests, can be undefined. (For context, see https://github.com/better-auth/better-auth/pull/491.)

Additionally, this pull request introduces support for OAuth providers that don’t require a `state` parameter, enhancing compatibility with providers that initiate the OAuth flow themselves (such as Clever). The changes make the `state` parameter optional and update the logic to handle cases where it’s not provided. For context [here is the clever documentation](https://dev.clever.com/docs/oauth-implementation#initiating-logins) and I'm sure there are many other provider initiated login flows out there.
    

---

## Summary by cubic
Fixes a crash in generic OAuth when the request body is undefined and adds support for providers that don’t use the state parameter (e.g., Clever). Improves compatibility and stability across more OAuth providers.

- **Bug Fixes**
  - Safely access c.body in parseState using optional chaining to avoid undefined errors.

- **New Features**
  - Added stateOptional to GenericOAuthConfig to allow skipping state validation for providers that don’t send state.
  - When stateOptional is enabled and no state is provided, bypass parseState and use safe defaults for the OAuth flow.

PR #9299, opened 2026-04-22
feat(generic-oauth,sso): support IDP-initiated flows via secure bounce

gustavovalverde:
Adds `allowIdpInitiated` to `GenericOAuthConfig` and the SSO `OIDCConfig`. When a stateless callback arrives at `/oauth2/callback/:providerId` or `/sso/callback/:providerId` and the resolved provider has `allowIdpInitiated: true`, the IDP-issued code is discarded and the OAuth flow is restarted server-side: a fresh `state` and PKCE verifier are generated, and the user is redirected to the provider's authorize endpoint. The provider's live IDP session completes the second hop automatically, so CSRF protection and PKCE stay in effect throughout.

The flag matches the existing `allowIdpInitiated` on the SAML SSO config for naming parity. Defaults to `false`.

Also guards `parseState` against `c.body` being undefined on GET callbacks, which previously crashed with a TypeError before the existing `please_restart_the_process` redirect could run.

This supersedes and implements the use case in PR #4951 without weakening CSRF or nullifying PKCE. The original approach there bypassed state validation entirely and introduced a silent PKCE downgrade; the bounce keeps both protections intact. Clever's own documentation recommends this pattern: "ignore the incoming Clever-initiated login and restart the authentication flow from your own site so you can include `state` yourself" (https://dev.clever.com/docs/oauth-implementation#initiating-logins).

Original work and motivation by @mhornbacher.

Closes #4951
Closes #9293

---
## Summary by cubic
Adds support for IdP‑initiated OAuth/OIDC via a secure “bounce” in `better-auth` and `@better-auth/sso`. Stateless callbacks are now supported without weakening CSRF or PKCE.

- **New Features**
  - Add `allowIdpInitiated` to `GenericOAuthConfig` and SSO `OIDCConfig` (default: false; matches SAML). On stateless callbacks to `/oauth2/callback/:providerId` or `/sso/callback/:providerId`, ignore the IdP code and restart server‑side with fresh `state` and PKCE, then redirect to the provider’s authorize endpoint (resolved via discovery when available).
  - Providers without the flag still redirect to the existing error. If `state` is present, normal validation runs (no bounce).
  - Updated docs and tests for bounce and strict paths.

- **Bug Fixes**
  - Guard `parseState` against undefined request bodies on GET callbacks to prevent a TypeError.

<sup>Written for commit dbbf039ccb647fab8c6be43341c2d50a5a08e20d. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/oauth-idp-initiated-bounce.md
+++ b/.changeset/oauth-idp-initiated-bounce.md
@@ -1,0 +1,6 @@
+---
+"better-auth": patch
+"@better-auth/sso": patch
+---
+
+Add `allowIdpInitiated` to `GenericOAuthConfig` and SSO `OIDCConfig` to support providers that initiate OAuth without a `state` parameter (e.g. Clever). When enabled, stateless callbacks restart the OAuth flow server-side with fresh state and PKCE, preserving CSRF protection. Also hardens `parseState` against undefined request bodies on GET callbacks.
--- a/docs/content/docs/plugins/generic-oauth.mdx
+++ b/docs/content/docs/plugins/generic-oauth.mdx
@@ -513,6 +513,37 @@ getUserInfo: async (tokens) => {
 }
 ```
 
+### IDP-Initiated Flows
+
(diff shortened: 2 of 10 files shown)


gustavovalverde:
Superseded by #9301 (targets `next` instead). The `next`-branch architecture is a cleaner fit for this feature — single bounce site in the shared `/callback/:id` handler rather than a TODO-flagged relocation on sync. Closing.

PR #9317, opened 2026-04-22
fix: skip stateless defaults when secondaryStorage is configured

ping-maxwell:
Fixes a spurious refreshCache warning emitted when secondaryStorage is configured with cookieCache: { enabled: false }.


Bekacru:
Seems like should be merged cc @ping-maxwell @bytaesu

PR #9327, opened 2026-04-23
fix: apply `disableBody: true` on common no-body POST endpoints

### LINKED ISSUE #9295: POST /sign-out with Content-Type: application/json + empty or malformed body returns HTTP 500 instead of 400
### Summary

When a client POSTs to `/api/auth/sign-out` with header `Content-Type: application/json` and a body that is empty or not valid JSON, Better Auth throws `JSON.parse` inside the handler and surfaces HTTP **500** to the caller. The expected behavior is **400 Bad Request** (or a no-body pass-through equivalent to `{}`), since client-supplied payload shape should not cause a server crash class.

### Environment

- `better-auth@1.6.4` (also reproducible on `1.6.3` per changelog inspection — no body-parse-related changes between the two versions)
- Next.js 16 App Router, `toNextJsHandler(auth)`
- Node 22

### Reproduction

Fresh Better Auth install, default `toNextJsHandler(auth)` wiring, no custom middleware.

```bash
# With empty body
curl -X POST -v https://YOUR-HOST/api/auth/sign-out \
  -H 'Content-Type: application/json' \
  -H 'Origin: https://YOUR-HOST' \
  --data ''
# → HTTP 500, empty body

# With malformed body
curl -X POST -v https://YOUR-HOST/api/auth/sign-out \
  -H 'Content-Type: application/json' \
  -H 'Origin: https://YOUR-HOST' \
  --data 'not json'
# → HTTP 500, empty body

# With valid `{}` body (for contrast)
curl -X POST -v https://YOUR-HOST/api/auth/sign-out \
  -H 'Content-Type: application/json' \
  -H 'Origin: https://YOUR-HOST' \
  --data '{}'
# → HTTP 200, body {"success":true}
```

The session-cookie state does not affect the outcome: no cookie + `{}` body returns a clean 200 as well. The trigger is specifically the CT-declares-JSON + body-fails-to-parse combination.

### Expected behavior

A malformed or empty JSON body is client input and should return **HTTP 400** (or be tolerated as an implicit `{}` given that `/sign-out` has no required body fields). Returning 500 causes two downstream issues:

1. **Alerting noise** — every `.catch()`-less client that double-sends sign-out (double-click, back-button after sign-out, mobile client with stale cookie) produces a Sentry / APM error instead of a clean no-op.
2. **Test-harness workaround spread** — teardown code in E2E suites ends up wrapping sign-out in `.catch(() => {})`, which hides real regressions. Several of our suites had acquired this pattern silently before we traced it back here.

### Suggested fix

Catch the `JSON.parse` at the framework boundary in the sign-out handler (and any other POST path that reads a JSON body without requiring its presence). Either:

- Return 400 with a descriptive error body, OR
- Treat empty / unparseable body as equivalent to `{}` for endpoints whose schema has all-optional fields.

### Local workaround (to be removed when upstream fixes)

We've wrapped the route handler to substitute `{}` on the sign-out path when CT:json + empty/malformed body are present. Narrow scope — only the sign-out path, passes through everything else so legitimate 400-class JSON errors elsewhere remain visible:

```ts
// apps/web/src/app/(api)/api/auth/[...all]/route.ts (extract)
export async function POST(req: Request) {
  const url = new URL(req.url);
  const isSignOutPath = url.pathname.endsWith('/sign-out');
  const isJsonBody = (req.headers.get('content-type') ?? '')
    .toLowerCase()
    .includes('application/json');

  if (!isSignOutPath || !isJsonBody) {
    return stockHandler.POST(req);
  }

  const raw = await req.clone().text();
  const needsGuard =
    raw.trim().length === 0 ||
    (() => { try { JSON.parse(raw); return false; } catch { return true; } })();

  if (!needsGuard) return stockHandler.POST(req);

  return stockHandler.POST(
    new Request(req.url, { method: 'POST', headers: req.headers, body: '{}' }),
  );
}
```

Happy to open a PR if a maintainer would find that useful.

### Related

- #5860 — `JSON.parse(undefined)` with `dynamicAccessControl` (closed Nov 2025, semantically adjacent)
- #5573 — 403 `MISSING_OR_NULL_ORIGIN` on sign-out (different symptom, different path)

Nothing matched the CT:json + empty-body → 500 shape directly on a pre-file issue search.

isaack1198-ux:
## Summary

Closes #9295.

The 5 no-body POST endpoints currently return **HTTP 500** when a client sends `Content-Type: application/json` with an empty or malformed JSON body — the framework attempts `JSON.parse` before the handler runs, and the parse exception propagates unchecked.

Applying `disableBody: true` (already supported by better-call per `EndpointBaseOptions`) tells the framework to skip body parsing on endpoints whose contract accepts no payload. After this change, a request with `Content-Type: application/json` + empty/malformed body returns **HTTP 200** (the contractual no-op).

## Affected endpoints

- `/sign-out` — `packages/better-auth/src/api/routes/sign-out.ts`
- `/revoke-sessions` — `packages/better-auth/src/api/routes/session.ts`
- `/revoke-other-sessions` — `packages/better-auth/src/api/routes/session.ts`
- `/sign-in/anonymous` — `packages/better-auth/src/plugins/anonymous/index.ts`
- `/delete-anonymous-user` — `packages/better-auth/src/plugins/anonymous/index.ts`

## Why these five

Each is documented as taking no request body (their handlers never read `ctx.body` / their OpenAPI `requestBody` is absent). Picked from the set surfaced in #9295 + a grep across `packages/better-auth/src/` for POST endpoints with no body schema + no `ctx.body` access.

## Reproduction (pre-fix)

```bash
curl -X POST https://YOUR-HOST/api/auth/sign-out \
  -H 'Content-Type: application/json' \
  -H 'Origin: https://YOUR-HOST' \
  --data ''
# → HTTP 500
```

## After this change

```bash
# same request → HTTP 200 {"success":true}
```

Other shapes unchanged:
- no body + no `Content-Type` → same as today (framework no-ops)
- valid `{}` body → same as today (returned 200 already)

## Notes

- No test additions — happy to follow up with ones if maintainers prefer. The `disableBody` option path is exercised by other endpoints in the codebase already.
- Happy to split the anonymous-plugin changes out if you'd prefer a narrower PR scope.

Original bug report: #9295

---
## Summary by cubic
Disable body parsing for five POST endpoints that accept no payload to prevent 500s on empty/malformed JSON requests. Fixes #9295; these requests now return 200 (no-op).

- **Bug Fixes**
  - Set `disableBody: true` on:
    - POST /sign-out
    - POST /revoke-sessions
    - POST /revoke-other-sessions
    - POST /sign-in/anonymous
    - POST /delete-anonymous-user

<sup>Written for commit a8b4c38e5724987061951aa3c366c6391e0d1f64. Summary will update on new commits.</sup>

### DIFF
--- a/packages/better-auth/src/api/routes/session.ts
+++ b/packages/better-auth/src/api/routes/session.ts
@@ -783,6 +783,7 @@ export const revokeSessions = createAuthEndpoint(
 	"/revoke-sessions",
 	{
 		method: "POST",
+		disableBody: true,
 		use: [sensitiveSessionMiddleware],
 		requireHeaders: true,
 		metadata: {
@@ -838,6 +839,7 @@ export const revokeOtherSessions = createAuthEndpoint(
 	"/revoke-other-sessions",
 	{
 		method: "POST",
+		disableBody: true,
 		requireHeaders: true,
 		use: [sensitiveSessionMiddleware],
 		metadata: {
--- a/packages/better-auth/src/api/routes/sign-out.ts
+++ b/packages/better-auth/src/api/routes/sign-out.ts
@@ -5,6 +5,7 @@ export const signOut = createAuthEndpoint(
 	"/sign-out",
 	{
 		method: "POST",
+		disableBody: true,
 		operationId: "signOut",
 		requireHeaders: true,
 		metadata: {
--- a/packages/better-auth/src/plugins/anonymous/index.ts
+++ b/packages/better-auth/src/plugins/anonymous/index.ts
@@ -65,6 +65,7 @@ export const anonymous = (options?: AnonymousOptions | undefined) => {
 				"/sign-in/anonymous",
 				{
 					method: "POST",
+					disableBody: true,
 					metadata: {
 						openapi: {
 							description: "Sign in anonymously",
@@ -145,6 +146,7 @@ export const anonymous = (options?: AnonymousOptions | undefined) => {
 				"/delete-anonymous-user",
 				{
 					method: "POST",
+					disableBody: true,
 					use: [sensitiveSessionMiddleware],
 					metadata: {
 						openapi: {


ping-maxwell (approved):

bytaesu:
This issue was fixed by https://github.com/better-auth/better-call/pull/130

Closing this PR.

PR #9337, opened 2026-04-23
fix: reference ID comparison to handle string types

devxrithm:
Fix reference ID comparison to handle string types 
https://github.com/better-auth/better-auth/pull/9337#top

### DIFF
--- a/packages/api-key/src/routes/list-api-keys.ts
+++ b/packages/api-key/src/routes/list-api-keys.ts
@@ -344,7 +344,7 @@ export function listApiKeys({
 				const referencesType = keyConfig?.references ?? "user";
 				return (
 					referencesType === expectedReferencesType &&
-					key.referenceId === referenceId
+					String(key.referenceId) === String(referenceId)
 				);
 			});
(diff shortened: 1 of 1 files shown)


packages/api-key/src/routes/list-api-keys.ts
```diff
@@ -344,7 +344,7 @@ export function listApiKeys({
 				const referencesType = keyConfig?.references ?? "user";
 				return (
 					referencesType === expectedReferencesType &&
-					key.referenceId === referenceId
+					String(key.referenceId) === String(referenceId)
```
bytaesu, line 347: > https://github.com/better-auth/better-auth/blob/main/packages/api-key/src/schema.ts#L41-L46

Q. `referenceId` is already defined as a string in the database type. Is there any case where this guard would be necessary? 🤔

bytaesu:
Hi @devxrithm,

Thanks for working on this. However, since the issue reporter has provided more detailed context, I'll continue this in PR #9338 🙏

PR #9338, opened 2026-04-23
fix(api-key): coerce referenceId to string before strict comparison in list route

### LINKED ISSUE #9336: apiKey.list silently returns empty when adapter returns non-string referenceId (strict-equal post-filter type asymmetry)

### Reproduction

Any setup where `apikey.referenceId` is stored/returned by the adapter as a non-string type. Most common trigger: Postgres with `advanced.database.generateId: "serial"` so `user.id` is an integer, and the `apikey.referenceId` column is also integer. Also reproduces with MongoDB ObjectId.

Steps:
1. Configure the api-key plugin with any adapter where `referenceId` is returned as a non-string.
2. `await auth.api.createApiKey({ body: { name: "test" }, headers })` using a logged-in session — succeeds, row written, HTTP 200.
3. Verify in the DB directly that the row exists with the expected `referenceId` (e.g. `reference_id = 1`).
4. `await auth.api.listApiKeys({ headers })` with the same session.
5. Observe `{ apiKeys: [], total: 0 }`. No error. No log.

**Current:** `listApiKeys` silently returns empty when `key.referenceId` and `session.user.id` aren't both JS strings. Root cause at `packages/api-key/src/routes/list-api-keys.ts:347`:

```ts
key.referenceId === referenceId
```

- `referenceId` (right) is `session.user.id`, which the core adapter factory always stringifies on output (`packages/core/src/db/adapter/factory.ts` — the `originalKey === "id" || field.references?.field === "id"` branch calls `String(newValue)`).
- `key.referenceId` (left) is whatever the adapter returned. The api-key schema declares `referenceId: { type: "string" }` with **no `references` entry**, so the factory's output transform never coerces it. If the backend returns a number or ObjectId, strict `===` evaluates false and the row is filtered out.

The upstream `adapter.findMany` in the same route returns the row correctly — DBs typically match int columns against string `"1"` via implicit coercion. The row is dropped *after* the adapter call, in the JS post-filter.

**Expected:** keys belonging to the session user are returned regardless of the adapter's storage representation.

Suggested fixes (pick one):
1. `String(key.referenceId) === String(referenceId)` at line 347. Lowest risk, purely additive coercion.
2. Declare `referenceId` in the api-key schema with `references: { model: "user", field: "id" }` so the core factory's output-stringify path applies symmetrically. Structurally cleaner but changes semantics across create/update/where, and may be wrong given `referenceId` can point at an organization — not only a user.

Related: #7892 (closed) — different site (`verifyApiKey` mutating ObjectId in MongoDB), same class of type-asymmetry issue. This report is independent — manifests on a fresh key with no verification having occurred.

`@better-auth/api-key@1.6.7` in the reporting project. Bug confirmed present in `@better-auth/api-key@1.6.8` (latest stable) and `@better-auth/api-key@1.7.0-beta.2`.

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.2.0: Tue Nov 18 21:03:25 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T8122",
    "release": "25.2.0",
    "cpuCount": 8,
    "cpuModel": "Apple M3",
    "totalMemory": "24.00 GB",
    "freeMemory": "0.32 GB"
  },
  "node": { "version": "v24.14.0", "env": "development" },
  "packageManager": { "name": "npm", "version": "11.9.0" },
  "frameworks": [
    { "name": "next", "version": "^16.2.4" },
    { "name": "react", "version": "^19.2.5" }
  ],
  "databases": [
    { "name": "pg", "version": "^8.20.0" },
    { "name": "@neondatabase/serverless", "version": "^1.1.0" }
  ],
  "betterAuth": { "version": "^1.6.7", "config": null }
}
```

(`config: null` because Better Auth is configured via a Payload CMS plugin rather than at a path the CLI scans. Effective config shape below.)

Backend, Package

```typescript
import { apiKey } from "@better-auth/api-key"
import { payloadAdapter } from "@delmaredigital/payload-better-auth"

  database: payloadAdapter({ payloadClient: payload }),
  advanced: {
    database: { generateId: "serial" }, // users.id is Postgres SERIAL (integer)
  },
  plugins: [
    apiKey(), // default config; no custom schema overrides
  ],
})
```

### Additional context

- Postgres (Neon) via `pg` + `@neondatabase/serverless`.
- `apikey.reference_id` column is `integer` in the live DB.
- Independently verified by calling `payload.find({ collection: "apikeys", where: { referenceId: { equals: 1 } } })` (and `{ equals: "1" }`) — both return the row. The adapter's `findMany` returns the row to `listApiKeys`; the strict `===` on line 347 is where it's dropped.
- Happy to open a PR with fix (1) + a regression test if useful.

allandelmare:
## Summary

Fix `apiKey.list` silently returning `{ apiKeys: [], total: 0 }` when the storage adapter returns `referenceId` as a non-string type.

`packages/api-key/src/routes/list-api-keys.ts` post-filters the adapter results with strict `===`:

```ts
key.referenceId === referenceId
```

- `referenceId` (right) is `session.user.id`, which the core adapter factory always stringifies on output (`originalKey === "id" || field.references?.field === "id"` → `String(newValue)`).
- `key.referenceId` (left) is whatever the adapter returned. The api-key schema declares `referenceId: { type: "string" }` with **no `references` entry**, so the factory's output transform never coerces it. Adapters that store/return `referenceId` as a non-string (Postgres integer columns, MongoDB ObjectId, etc.) make the strict `===` evaluate false and the row is silently filtered out.

Values are now coerced with `String()` on both sides before comparison. Purely additive — same-type comparisons remain equivalent, and no storage representation changes.

## Why not fix the schema instead?

Adding `references: { model: "user", field: "id" }` to the api-key schema would let the core factory stringify `referenceId` on output symmetrically, but that would break the plugin's dual user/organization reference design (confirmed in the triage on #9336) — `referenceId` can point at either a user or an organization, not strictly `user.id`. String coercion at the comparison site is the safe, targeted fix.

## Changes

- `packages/api-key/src/routes/list-api-keys.ts`: `String(key.referenceId) === String(referenceId)` at the post-filter, with a comment pointing to this issue.
- `packages/api-key/src/api-key.test.ts`: regression test that uses the secondary-storage path (which exercises the post-filter directly, since KV keys aren't pre-filtered by any adapter `findMany`). The test stores a key with a `Number` `referenceId` whose `String()` matches `session.user.id` and asserts `listApiKeys` still returns it. Confirmed to fail without the fix and pass with it.
- `.changeset/…`: patch bump for `@better-auth/api-key`.

## Test plan

- [x] `pnpm vitest run` in `packages/api-key` — **177/177 pass**
- [x] `pnpm typecheck` in `packages/api-key` — clean
- [x] `pnpm biome check packages/api-key/src/routes/list-api-keys.ts packages/api-key/src/api-key.test.ts` — clean
- [x] Regression test fails on unpatched `main`, passes with fix

Closes #9336

---
## Summary by cubic
Fixes `apiKey.list` returning empty results when adapters store `referenceId` as non-strings (e.g., Postgres ints, MongoDB ObjectIds). The post-filter now compares stringified IDs so keys are returned consistently.

- **Bug Fixes**
  - Added a regression test covering numeric `referenceId` via secondary storage.
  - Patch bump for `@better-auth/api-key`.

<sup>Written for commit 85f9f5f8da1f97e085ad45d158ef97b2760bbb67. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/fix-api-key-list-string-coerce-referenceid.md
+++ b/.changeset/fix-api-key-list-string-coerce-referenceid.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/api-key": patch
+---
+
+Fix `apiKey.list` silently returning empty when the storage adapter returns `referenceId` as a non-string type (e.g. Postgres integer columns, MongoDB ObjectId). The list route's post-filter used strict `===` to compare `key.referenceId` against the stringified `session.user.id`, which dropped every row when the adapter returned a type other than `string`. Values are now coerced with `String()` on both sides before comparison.
--- a/packages/api-key/src/api-key.test.ts
+++ b/packages/api-key/src/api-key.test.ts
@@ -4641,3 +4641,77 @@ describe("api-key", async () => {
 		});
 	});
 });
+
+// Regression for https://github.com/better-auth/better-auth/issues/9336
+//
+// listApiKeys post-filters the adapter results with strict `===`:
+//
+//   key.referenceId === referenceId
+//
+// `referenceId` on the right is derived from `session.user.id`, which the core
+// adapter factory always stringifies on output. `key.referenceId` on the left
+// is whatever the adapter returned — the api-key schema declares it as a plain
+// `string` field with no `references` entry, so the factory's output transform
+// never coerces it. Adapters that store/return `referenceId` as a non-string
+// (e.g. Postgres integer columns, MongoDB ObjectId) therefore made the strict
+// equality silently fail for every row.
+describe("api-key (regression #9336) — list tolerates non-string referenceId", async () => {
+	// Secondary-storage mode exercises the list endpoint's post-filter
+	// directly: keys are returned from the KV as-is, without any pre-filter
+	// strict-equality gating. This mirrors how real SQL/BSON adapters behave —
+	// their findMany tolerates int-vs-string via implicit coercion, so the
+	// mismatched types only bite at the JS-level post-filter inside the route.
+	const store = new Map<string, string>();
+	const secondaryStorage: SecondaryStorage = {
+		set(key, value) {
+			store.set(key, value as string);
+		},
+		get(key) {
+			return store.get(key) || null;
+		},
+		delete(key) {
+			store.delete(key);
+		},
+	};
+
+	const { auth, signInWithTestUser } = await getTestInstance(
+		{
+			secondaryStorage,
+			plugins: [apiKey({ storage: "secondary-storage" })],
+			advanced: {
+				// SERIAL ids — user.id is numeric in the DB. The core factory
+				// still stringifies it on output, so session.user.id is the
+				// string "1" at the plugin boundary. This mirrors the original
+				// report's Postgres setup.
+				database: { generateId: "serial" },
+			},
+		},
+		{
+			clientOptions: { plugins: [apiKeyClient()] },
+		},
+	);
+	const { headers, user } = await signInWithTestUser();
+
+	it("returns API keys whose stored referenceId is a Number (e.g. Postgres integer column)", async () => {
+		const created = await auth.api.createApiKey({ body: {}, headers });
+
+		// Rewrite the stored key so referenceId is a JS Number that String()-
+		// equals session.user.id. The list route's post-filter runs
+		// `key.referenceId === referenceId` and `referenceId` is always a
+		// string because the core factory stringifies id outputs. Without the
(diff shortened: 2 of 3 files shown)


bytaesu:
I think this doesn't represent a valid test case, since it intentionally converts the value to a number mid-flow.

If you have further questions, feel free to mention me 😁

PR #9345, opened 2026-04-24
fix(api): preserve current session on change-password with revokeOtherSessions

### LINKED ISSUE #6881: Password Change Issue
Note: Organization plugin is used.

1. Try to change password with the following code below
```
  const onSubmit = async (data: ChangePasswordInput) => {
    await authClient.changePassword(
      {
        currentPassword: data.currentPassword,
        newPassword: data.newPassword,
        revokeOtherSessions: true,
      },
      {
        onSuccess() {
          toast.success("Password changed successfully");
          form.reset();
        },
        onError({ error }) {
          toast.error(error.message || "Password changed failed");
        },
      }
    );
  };
```
2.You will be logged out of the session and when you try to login again the active organization is not available in session table.

ramonclaudio:
`auth.api.changePassword({ revokeOtherSessions: true, ... })` silently rotates the caller's own session, leaving any cached JWT (Convex queries, OIDC consumers) pointing at the deleted session id for ~500 to 1500 ms. The JSDoc on `revokeOtherSessions` reads "revoke all sessions that are not the current one logged in by the user", so the behavior contradicts its own contract.

This PR adds `internalAdapter.deleteOtherSessions(userId, exceptToken)` and routes both `/change-password`'s `revokeOtherSessions: true` branch and `/revoke-other-sessions` through it. `deleteSessions` is unchanged.

# Test plan

- `pnpm typecheck` clean
- `pnpm vitest internal-adapter.test.ts -t deleteOtherSessions`: 4/4 pass
- `pnpm vitest update-user.test.ts -t "revoke other sessions while preserving"`: 1/1 pass

Refs #6881

### DIFF
--- a/packages/better-auth/src/api/routes/update-user.ts
+++ b/packages/better-auth/src/api/routes/update-user.ts
@@ -187,9 +187,8 @@ export const changePassword = createAuthEndpoint(
 									properties: {
 										token: {
 											type: "string",
-											nullable: true, // Only present if revokeOtherSessions is true
-											description:
-												"New session token if other sessions were revoked",
+											nullable: true,
+											description: "Always null; current session preserved.",
 										},
 										user: {
 											type: "object",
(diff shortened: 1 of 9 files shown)


gustavovalverde:
@ramonclaudio, can you make this PR body more concise, please? Right now, it's way too extensive to quickly review and understand what the issue was and the solution.

ramonclaudio:
Sorry @gustavovalverde, cleaning it up as we speak. Essentially, changing your password currently causes it to silently replace your current session with a new one, leaving any cached JWTs pointing at the deleted session until the client catches up. Now it leaves your session alone and only revokes the other devices, matching the /revoke-other-sessions endpoint's semantics, the parameter's own JSDoc, and the token: null sentinel that sign-up already uses when no session is minted. This PR just brings /change-password in line with the rest of the library.

packages/better-auth/src/api/routes/update-user.ts
```diff
@@ -286,28 +286,25 @@ export const changePassword = createAuthEndpoint(
...
 			});
-			token = newSession.token;
+			const otherSessions = activeSessions.filter(
+				(session) => session.token !== ctx.context.session.session.token,
+			);
+			await Promise.all(
+				otherSessions.map((session) =>
+					ctx.context.internalAdapter.deleteSession(session.token),
+				),
+			);
```
ping-maxwell: Is it not possible to have a new internalAdapter function which deletes all sessions except one with a specific token?
otherwise I think we should be doing that instead

ping-maxwell (changes requested):

packages/better-auth/src/api/routes/update-user.ts
ramonclaudio: Done. Added `deleteOtherSessions(userId, exceptToken)` to `internalAdapter` and routed both `/change-password` and `/revoke-other-sessions` through it. Existing `deleteSessions` is unchanged.

ping-maxwell (dismissed):

PR #9349, opened 2026-04-24
fix(organization): re-export field types to prevent TS2742 with additionalFields

### LINKED ISSUE #7039: TypeScript inference fails with `organizationClient` when `additionalFields` is used

### To Reproduce

1.	Create a client using createAuthClient() including organizationClient() plugin.
2.	Set the following in tsconfig.json:
```json
{
  "declaration": true,
  "moduleResolution": "bundler"
}
```
3.	Use the following client configuration:
```ts
  baseURL: import.meta.env.VITE_API_URL,
  basePath: "/api/user/auth",
  plugins: [
    adminClient(),
    organizationClient({
      teams: { enabled: true },
      schema: {
        organization: {
          additionalFields: {
            publicId: {
              type: "string",
              input: false,
            },
          },
        },
      },
    }),
  ],
});
```

Current:
TypeScript throws the following error: `The inferred type of authClient cannot be named without a reference to ../../node_modules/better-auth/dist/index-BZSqJoCN.mjs. This is likely not portable. A type annotation is necessary. (ts 2742)`

Expected:
TypeScript should correctly infer the type of authClient without requiring a manual type annotation.

1.4.9

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:45 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6030",
    "release": "25.2.0",
    "cpuCount": 11,
    "cpuModel": "Apple M3 Pro",
    "totalMemory": "18.00 GB",
    "freeMemory": "0.19 GB"
  },
  "node": {
    "version": "v22.20.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "10.22.0"
  },
  "frameworks": [
    {
      "name": "react",
      "version": "^19.2.0"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null,
    "error": "fetch failed"
  }
}
```

Client

```typescript
  baseURL: import.meta.env.VITE_API_URL,
  basePath: "/api/user/auth",
  plugins: [
    adminClient(),
    organizationClient({
      teams: {
        enabled: true,
      },
      schema: {
        organization: {
          additionalFields: {
            publicId: {
              type: "string",
              input: false,
            },
          },
        },
      },
    }),
  ],
});
```

### Additional context

When using `organizationClient()` with the `additionalFields` option, TypeScript fails to infer the type of authClient under `declaration: true` and `moduleResolution: "bundler"`.

This is similar to the previously reported issue #6565, but now it only occurs when `additionalFields` are specified in the plugin schema.

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/7039

## Summary

Fixes a TS2742 error when using `organizationClient` with `additionalFields` under `declaration: true` + `moduleResolution: "bundler"`.

## Problem

When `organizationClient` is called with `additionalFields` in the schema, the return type includes `InferOrganization<CO>` which expands to use `FieldAttributeToObject<RemoveFieldsWithReturnedFalse<Field>>` from `src/db/field.ts`. During the build, these types are emitted into `dist/db/field.d.mts` — a declaration chunk that is not accessible through any public export path in the package.

When a consumer uses this configuration:

```ts
  plugins: [
    organizationClient({
      schema: {
        organization: {
          additionalFields: {
            publicId: { type: "string", input: false },
          },
        },
      },
    }),
  ],
});
```

TypeScript cannot name the inferred type of `authClient` through any public module path, resulting in:

```
error TS2742: The inferred type of 'authClient' cannot be named without a reference to
'../node_modules/better-auth/dist/db/field.mjs'. This is likely not portable.
A type annotation is necessary.
```

Without `additionalFields`, the same code compiles cleanly.

## Fix

Re-export `FieldAttributeToObject` and `RemoveFieldsWithReturnedFalse` from the `better-auth/client/plugins` barrel (`src/client/plugins/index.ts`). This makes these types accessible through the public export path so TypeScript can reference them when emitting declarations.


bytaesu (approved):
Is there a way to address the root cause so we don't have to fix these one-by-one?

gustavovalverde:
> Is there a way to address the root cause so we don't have to fix these one-by-one?

+1

PR #9351, opened 2026-04-25
fix: skip getAccessToken updateAccount in stateless mode

### LINKED ISSUE #7703: on stateless setup, calling `getAccessToken` does not refresh the access token when needed

daniele-20tab:
## Summary

Fixes #7703. Supersedes #7705 (inactive since Jan 30, 2026).

In stateless setups (no `database` configured), `getAccessToken` unconditionally calls `internalAdapter.updateAccount` against the in-memory fallback adapter. This is wasted work and, in multi-instance deployments where the cookie-resolved `account.id` does not exist in the current instance's memory, can mask the refreshed tokens behind a generic `BAD_REQUEST`.

This PR adopts the form @bytaesu suggested in their review on #7705 — \`if (account.id && ctx.context.options.database)\` — instead of the try/catch in the original PR. It does **not** include the \`accessTokenExpiresAt\`-missing change from #7705 (per @bytaesu's request, that belongs in a separate PR).

Co-authored with @biniam0 to acknowledge the original work on #7705.

## Changes

- \`packages/better-auth/src/api/routes/account.ts\`: skip \`internalAdapter.updateAccount\` in \`getAccessToken\` when \`options.database\` is undefined. The account cookie is still refreshed via \`setAccountCookie\` when \`storeAccountCookie\` is enabled.
- \`packages/better-auth/src/api/routes/account.test.ts\`: add \`should skip internalAdapter.updateAccount during getAccessToken auto-refresh when no database is configured\` — verifies the spy on \`updateAccount\` is not called and refreshed tokens are still returned.
- \`.changeset/fix-stateless-getaccesstoken-skip-update.md\`: \`patch\` changeset.

## Verification

- On \`main\` without this fix: new test fails (\`updateAccount\` called once)
- With this fix: new test passes; full \`account.test.ts\` 24/24 passing
- No \`any\`, no implicit-\`any\`, no TS errors in the new test (verified with \`tsc --noEmit -p .\`)

## Out of scope

- The \`accessTokenExpiresAt\`-missing case from #7705. Per @bytaesu's review comment, that change is better handled in a separate PR.

---
## Summary by cubic
Skip `internalAdapter.updateAccount` in `getAccessToken` when no `database` is configured to avoid stateless multi-instance `BAD_REQUEST`s; tokens still return and the account cookie is refreshed when `storeAccountCookie` is enabled.

- **Bug Fixes**
  - Gate update behind `account.id && ctx.context.options.database`.
  - Add test to ensure no update call and successful token refresh in stateless mode.

<sup>Written for commit 4d9638c22eb1a144ac5d98719bdffec98733465a. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/fix-stateless-getaccesstoken-skip-update.md
+++ b/.changeset/fix-stateless-getaccesstoken-skip-update.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+fix: skip `internalAdapter.updateAccount` during `getAccessToken` auto-refresh when no `database` is configured. Previously, refreshed tokens triggered a redundant call to the in-memory fallback adapter; with no real DB to update, the call is now skipped and the refreshed tokens are still returned and persisted in the account cookie when `storeAccountCookie` is enabled. Fixes #7703.
--- a/packages/better-auth/src/api/routes/account.test.ts
+++ b/packages/better-auth/src/api/routes/account.test.ts
@@ -1,5 +1,6 @@
 import { BASE_ERROR_CODES } from "@better-auth/core/error";
 import type { GoogleProfile } from "@better-auth/core/social-providers";
+import type { ErrorContext } from "@better-fetch/fetch";
 import { HttpResponse, http } from "msw";
 import { setupServer } from "msw/node";
 import type { MockInstance } from "vitest";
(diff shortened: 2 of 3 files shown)


daniele-20tab:
@bytaesu when you have a moment — this PR implements the change you suggested in your review on #7705 (`if (account.id && ctx.context.options.database)` instead of try/catch), with a regression test and a changeset. All 27 CI checks are green; the 2 Vercel "Authorization required" failures are the standard fork-deploy gate and need a maintainer to authorize.

I co-authored the commit with @biniam0 to acknowledge the original work on #7705. Happy to amend if you want any changes.

bytaesu:
It seems like this might not be addressing the root cause and is only fixing things at a surface level, so I'll take a closer look.

gustavovalverde:
Superseded by #9799.

Digging into #7703, the stateless `updateAccount` call isn't the root cause: when a provider omits `expires_in`, no expiry is stored at all, so `getAccessToken` never reaches the refresh path (with or without a database). #9799 addresses that by letting a `genericOAuth` provider declare `accessTokenExpiresIn`, which synthesizes the expiry at sign-in and on refresh when the provider omits it. Thanks for surfacing the problem.

PR #9353, opened 2026-04-25
feat(google): add option to disable granted scopes

### LINKED ISSUE #9326: feat: add a option to disable include_granted_scopes param on google sso

include_granted_scopes: "true" ([source](https://github.com/better-auth/better-auth/blob/main/packages/core/src/social-providers/google.ts#L96)) is currently hardcoded in the Google provider and cannot be removed through configuration options.

The include_granted_scopes parameter causes Google to return all previously granted scopes in addition to the newly requested ones. This can be a problem when a app wants to explicitly control exactly which scopes are active for each session.

Add a option to disable include_granted_scopes param

```typescript
const auth = betterAuth({
  socialProviders: {
    google: {
      ...
      disableIncludeGrantedScopes: true, // or includeGrantedScopes: false
    },
  }
}
```

_No response_

### Additional context

_No response_

GautamBytes:
## Summary

Adds a Google provider option, `includeGrantedScopes`, to control whether Better Auth sends `include_granted_scopes=true` in Google authorization requests.

By default, behavior is unchanged. Users can now set `includeGrantedScopes: false` to omit the parameter when they want each OAuth flow to request only the scopes explicitly provided for that flow.

## Changes

- Add `includeGrantedScopes?: boolean` to `GoogleOptions`
- Keep `include_granted_scopes=true` by default
- Omit `include_granted_scopes` when `includeGrantedScopes` is `false`
- Add Google provider tests for default and disabled behavior
- Document the new option
- Add a patch changeset

---
## Summary by cubic
Adds an `includeGrantedScopes` option to the Google provider to control Google’s `include_granted_scopes` behavior during OAuth. Default stays the same; set `includeGrantedScopes: false` to request only the scopes you pass for that flow.

- New Features
  - Added `includeGrantedScopes?: boolean` to `GoogleOptions` (defaults to true).
  - Omits `include_granted_scopes` in the auth URL when disabled; only sends the requested scope(s).
  - Updated docs, added tests for both modes, and included a patch changeset in `better-auth`.

<sup>Written for commit 66bb590e302b01387d6b959fad2a1d364125e17e. Summary will update on new commits.</sup>

### DIFF
--- a/packages/core/src/social-providers/google.ts
+++ b/packages/core/src/social-providers/google.ts
@@ -51,6 +51,11 @@ export interface GoogleOptions extends ProviderOptions<GoogleProfile> {
 	 * The hosted domain of the user
 	 */
 	hd?: string | undefined;
+	/**
+	 * Include previously granted scopes in the authorization request.
+	 * Defaults to true.
+	 */
+	includeGrantedScopes?: boolean | undefined;
 }
 
 export const google = (options: GoogleOptions) => {
@@ -92,9 +97,12 @@ export const google = (options: GoogleOptions) => {
 				display: display || options.display,
 				loginHint,
 				hd: options.hd,
-				additionalParams: {
-					include_granted_scopes: "true",
-				},
+				additionalParams:
+					options.includeGrantedScopes === false
+						? undefined
+						: {
+								include_granted_scopes: "true",
+							},
 			});
 			return url;
 		},
--- a/.changeset/google-include-granted-scopes.md
+++ b/.changeset/google-include-granted-scopes.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Add a Google provider option to omit `include_granted_scopes` from authorization requests.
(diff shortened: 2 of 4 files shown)


packages/core/src/social-providers/google.ts
```diff
@@ -92,9 +97,12 @@ export const google = (options: GoogleOptions) => {
...
 				hd: options.hd,
-				additionalParams: {
-					include_granted_scopes: "true",
-				},
+				additionalParams:
+					options.includeGrantedScopes === false
+						? undefined
+						: {
+								include_granted_scopes: "true",
+							},
```
bytaesu, line 105: > For Google: Return all granted scopes to current key, so that signIn does not overwrite previously granted scopes in db

The default for this parm is `false` on Google side, but it was intentionally set to `true` in PR #3013.

If we make this configurable, it could introduce unexpected regressions for some users. What do you think? 🤔

bytaesu:
I'm looking into whether this issue can be resolved in a different way.

packages/core/src/social-providers/google.ts
GautamBytes: This only adds an explicit opt-out for apps that intentionally want per-flow scope isolation, should i add a stronger docs note about the tradeoff?

bytaesu:
Thanks for working on this @GautamBytes

I'll continue this work in PR https://github.com/better-auth/better-auth/pull/9383 and add you as a co-author 😁

PR #9363, opened 2026-04-26
fix: skip identical cookie writes in `nextCookies` after-hook

### LINKED ISSUE #8464: nextjs cookie plugin causes router refresh
## Note: This is opened for tracking purposes.

Discord: https://discord.com/channels/1288403910284935179/1476968169762586747

### To Reproduce

calling cookies.set in a server actions invalidates client router cache and forces a refetch. This refetches our pages and causes a continuous “loop” forever refetching the page

current:
router refresh due to cookie set
expected:
no router refreshes

1.5.4

### System info

```bash
not applicable
```

Client

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

bytaesu:
#9059 closed #8464 by removing the cookie probe, but the same `cookies().set()` invalidation still fires from the after-hook on session refresh.

## Limitation

The `session_data` cookie used by `cookieCache.enabled: true` embeds timestamps in its signed value, so its raw bytes differ on every refresh and the idempotency check cannot catch it.

## Path to root fix
- **Upstream**: vercel/next.js#84313 adds `cookies().set(name, value, { revalidate: false })`. Once shipped, this patch becomes a small replacement that drops the idempotency check entirely and applies uniformly to `session_token` and `session_data`. Authored by @cramforce, currently an inactive draft.

### DIFF
--- a/.changeset/fix-nextcookies-idempotent-writes.md
+++ b/.changeset/fix-nextcookies-idempotent-writes.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Skip identical cookie writes in `nextCookies()` after-hook to avoid router cache invalidation when the session token value is unchanged
--- a/packages/better-auth/src/integrations/next-js.test.ts
+++ b/packages/better-auth/src/integrations/next-js.test.ts
@@ -160,4 +160,62 @@ describe("next-js integration", () => {
 
 		expect(session).not.toBeNull();
 	});
+
+	/**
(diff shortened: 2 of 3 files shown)


JamBalaya56562:
Thanks for digging into this. The underlying issue #8464 was resolved by #9059 (header-based RSC detection in `nextCookies`), so this may no longer be needed — though if you're still seeing duplicate cookie writes on top of that fix, let me know.

PR #9369, opened 2026-04-27
fix: incorrect email casing across one-tap, email-otp & email-verification

ping-maxwell:
Our email verification flow didn't handle proper email normalization, this PR resolves such.


packages/better-auth/src/db/internal-adapter.ts
bytaesu, line 951: `data`: missing case-sensitive

bytaesu: `data`: missing case-sensitive

packages/better-auth/src/api/routes/email-verification.ts
bytaesu: `updateTo`: this is also an email

bytaesu (commented):
This could also be applied to the `one-tap` and `email-otp` plugins, though it might be redundant since it probably goes through the internal adapter logic. That said, the calc overhead is minimal, so it might still be worth explicitly handling case-insensitivity there as well 🧐

bytaesu (approved):

Issue #9371, opened 2026-04-28
`npx auth generate --output src/lib -y` throws an `EISDIR` error


TheMikeyRoss:
### Is this suited for github?

- [x] Yes, this is suited for github

### Reproduction

1. Run the command `npx auth generate --output src/lib -y`
2. See the following error:
```
node:internal/fs/promises:636
  return new FileHandle(await PromisePrototypeThen(
                        ^

Error: EISDIR: illegal operation on a directory, open 'X:\development\Project\src\lib'
    at async open (node:internal/fs/promises:636:25)
    at async Object.writeFile (node:internal/fs/promises:1205:14)
    at async Command.generateAction (file:///C:/Users/Mikey/AppData/Local/npm-cache/_npx/cec068fea869db4d/node_modules/auth/dist/index.mjs:1819:26) {
  errno: -4068,
  code: 'EISDIR',
  syscall: 'open',
  path: 'X:\\development\\Project\\src\\lib'
}

Node.js v22.22.1
```

### Current vs. Expected behavior

If I run `npx auth generate -y` it successfully generates the `auth-schema.ts` in the root of my project. But when I specify the output directory I get the error mentioned above.

### What version of Better Auth are you using?

`1.6.4` (And the CLI is 0.8.0)

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 10 Pro",
    "release": "10.0.19045",
    "cpuCount": 12,
    "cpuModel": "Intel(R) Core(TM) i7-6850K CPU @ 3.60GHz",
    "totalMemory": "63.90 GB",
    "freeMemory": "32.67 GB"
  },
  "node": {
    "version": "v22.22.1",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.4"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.2.3"
    },
    {
      "name": "react",
      "version": "^19.2.5"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.20.0"
    },
    {
      "name": "postgres",
      "version": "^3.4.9"
    },
    {
      "name": "drizzle",
      "version": "^0.45.2"
    }
  ],
  "betterAuth": {
    "version": "^1.6.4",
    "config": {
      "appName": "Project",
      "baseURL": "http://localhost:1406",
      "socialProviders": {
        "google": {
          "clientId": "[REDACTED]",
[...]

PR #9372, opened 2026-04-28
docs: update cli.mdx to include an example for `--output` flag

TheMikeyRoss:
added an example in the cli docs to clarify how to use the `--output` flag to prevent someone running into a similar problem to what I ran into in https://github.com/better-auth/better-auth/issues/9371

---
## Summary by cubic
Add an inline example for the `--output` flag in the CLI docs to show how to set the destination path. Example: `npx auth@latest generate --output src/lib/auth-schema.ts`.

<sup>Written for commit d3c6930ec2d5b93e36ca5a515b030ae8d8d816a3. Summary will update on new commits.</sup>


bytaesu (changes requested):
Hi @TheMikeyRoss, I've checked the situation in the issue.

Instead of showing a specific example command, how about incorporating into the explanation that a file path needs to be provided? 🤔

TheMikeyRoss:
> Hi @TheMikeyRoss, I've checked the situation in the issue.
> 
> Instead of showing a specific example command, how about incorporating into the explanation that a file path needs to be provided? 🤔

How about:

`--output` — Defines where the generated schema is saved. While it defaults to `prisma/schema.prisma` for Prisma, `schema.ts` for Drizzle, or `schema.sql` for Kysely, you can provide a custom path like `src/lib/auth-schema.ts` directly after the flag.

ping-maxwell (approved):

gustavovalverde:
Closing in favor of #9564. This PR is a 1-line docs workaround; #9564 fixes the actual EISDIR crash in `packages/cli/src/commands/generate.ts` by stat-checking the resolved path and appending `auth-schema.ts` when `--output` points to an existing directory.

PR #9382, opened 2026-04-28
fix(oauth): preserve account.scope across re-auth and refresh

bytaesu:
`account.scope` is now treated as the user's accumulated granted-scope
set rather than the latest token's `scope` claim. linkSocial unions
newly granted scopes into the stored value, while sign-in
re-authentication and refresh-token responses leave it untouched.

Generalizes the Google-specific `include_granted_scopes=true` band-aid
from #3013 into a provider-agnostic invariant grounded in RFC 6749 §1.5.

---

### DIFF
--- a/.changeset/quiet-pandas-merge.md
+++ b/.changeset/quiet-pandas-merge.md
@@ -1,0 +1,6 @@
+---
+"better-auth": patch
+"@better-auth/core": patch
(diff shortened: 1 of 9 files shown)


gustavovalverde:
Part of #9825 in the `next` branch

PR #9398, opened 2026-04-29
fix(sso): use findSAMLProvider in spMetadata so defaultSSO works

### LINKED ISSUE #9397: [SSO] /sso/saml2/sp/metadata returns 404 for defaultSSO providers

### Reproduction

```ts

const auth = betterAuth({
  database: memoryAdapter({
    user: [], session: [], verification: [], account: [], ssoProvider: [],
  }),
  plugins: [
    sso({
      defaultSSO: [
        {
          domain: "example.com",
          providerId: "default-saml",
          samlConfig: {
            issuer: "http://idp.example.com",
            entryPoint: "http://idp.example.com/sso",
            cert: "<idp-cert>",
            callbackUrl: "http://localhost:3000/sso/saml2/sp/acs/default-saml",
            spMetadata: { metadata: "<sp-metadata-xml>" },
          },
        },
      ],
    }),
  ],
});

await auth.api.spMetadata({ query: { providerId: "default-saml" } });
// → APIError "NOT_FOUND": "No provider found for the given providerId"
```

**Current:** `/sso/saml2/sp/metadata?providerId=<id>` returns the SP metadata XML for SAML providers stored in the database, but throws `NOT_FOUND` for providers configured via the `defaultSSO` option — even though every other SAML endpoint in the plugin (`signInSSO`, the `/callback/:providerId` ACS handler, `signOut`) honours `defaultSSO` via the shared `findSAMLProvider` helper.

The `spMetadata` handler is the only SAML endpoint that calls `ctx.context.adapter.findOne({ model: "ssoProvider", ... })` directly instead of going through `findSAMLProvider`, so the in-memory `defaultSSO` array is never consulted.

**Expected**: the endpoint returns the SP metadata XML for `defaultSSO` providers, matching the rest of the SAML surface.

1.6.9 (current `@better-auth/sso`); also reproduces on `canary` (`packages/sso/src/routes/sso.ts` lines ~97-141).

### System info

```bash
N/A — bug is in plugin source, reproduces on any host.
```

Backend, Package

See reproduction above.

### Additional context

The `findSAMLProvider` helper (already imported in `packages/sso/src/routes/sso.ts`) handles `defaultSSO` resolution before falling back to the adapter — every other SAML route uses it. 

Happy to send a PR.

Craga89:
## Description

`/sso/saml2/sp/metadata?providerId=<id>` is the only SAML endpoint in the plugin that calls `ctx.context.adapter.findOne` directly instead of going through the shared `findSAMLProvider` helper. As a result, providers configured via the `defaultSSO` option (which are not persisted to the `ssoProvider` table) cause the endpoint to throw `NOT_FOUND`, even though `signInSSO`, the SAML callback handler, and `signOut` all resolve them correctly.

This swaps the inline adapter lookup for `findSAMLProvider`, which already handles `defaultSSO` matches before falling back to the adapter. `safeJsonParse` is dropped at the call site because `findSAMLProvider` returns `samlConfig` as a parsed object on both code paths.

Closes #9397

## Changes

- `packages/sso/src/routes/sso.ts` — replace `adapter.findOne` with `findSAMLProvider`; drop now-unused `safeJsonParse` call at the spMetadata site (helper is already imported in this file).
- `packages/sso/src/saml.test.ts` — add an `spMetadata` test inside the existing `"SAML SSO with defaultSSO array"` describe block. Without the source change this test throws `NOT_FOUND`; with it the test passes.
- `.changeset/sso-default-metadata.md` — patch changeset for `@better-auth/sso`.

## Breaking changes

None. The endpoint contract is unchanged for DB-backed providers; this only adds `defaultSSO` resolution.

## Tests

- `pnpm --filter @better-auth/sso typecheck` — clean.
- `pnpm --filter @better-auth/sso vitest run src/saml.test.ts` — 109 tests pass (existing 108 + the new `defaultSSO` spMetadata test).

### DIFF
--- a/.changeset/sso-default-metadata.md
+++ b/.changeset/sso-default-metadata.md
@@ -1,0 +1,11 @@
+---
+"@better-auth/sso": patch
+---
+
+fix(sso): use findSAMLProvider in spMetadata so defaultSSO providers resolve
+
+`/sso/saml2/sp/metadata` was the only SAML endpoint that called `adapter.findOne`
+directly, so providers configured via `defaultSSO` (which are not persisted to the
+database) caused it to throw `NOT_FOUND`. The endpoint now uses the shared
+`findSAMLProvider` helper, matching `signInSSO`, the SAML callback handler, and
+`signOut`.
--- a/packages/sso/src/routes/sso.ts
+++ b/packages/sso/src/routes/sso.ts
@@ -108,25 +108,18 @@ export const spMetadata = (options?: SSOOptions) => {
 			},
 		},
 		async (ctx) => {
-			const provider = await ctx.context.adapter.findOne<{
-				id: string;
-				samlConfig: string;
-			}>({
-				model: "ssoProvider",
-				where: [
-					{
-						field: "providerId",
-						value: ctx.query.providerId,
-					},
-				],
-			});
+			const provider = await findSAMLProvider(
+				ctx.query.providerId,
+				options,
+				ctx.context.adapter,
+			);
 			if (!provider) {
 				throw new APIError("NOT_FOUND", {
 					message: "No provider found for the given providerId",
 				});
 			}
 
-			const parsedSamlConfig = safeJsonParse<SAMLConfig>(provider.samlConfig);
+			const parsedSamlConfig = provider.samlConfig;
 			if (!parsedSamlConfig) {
 				throw new APIError("BAD_REQUEST", {
 					message: "Invalid SAML configuration",
--- a/packages/sso/src/saml.test.ts
+++ b/packages/sso/src/saml.test.ts
@@ -658,6 +658,17 @@ describe("SAML SSO with defaultSSO array", async () => {
 			redirect: true,
 		});
 	});
+
+	it("should fetch sp metadata for a defaultSSO provider", async () => {
+		const spMetadataRes = await auth.api.spMetadata({
+			query: {
+				providerId: "default-saml",
+			},
+		});
+		const spMetadataResValue = await spMetadataRes.text();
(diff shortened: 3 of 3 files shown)


Craga89:
Following up on @dosubot's [reply on the linked issue](https://github.com/better-auth/better-auth/issues/9397#issuecomment-4345161162) about the `signInSSO` and OIDC callback handlers — I had a look at both before submitting this PR.

Both already check `defaultSSO` _before_ falling back to `adapter.findOne`, so they don't have the same `NOT_FOUND` gap as `spMetadata`:

- [`signInSSO`](https://github.com/better-auth/better-auth/blob/c8ec03d26effa0c72729b27afc2db8308d416c9a/packages/sso/src/routes/sso.ts#L1043-L1070) — defaultSSO branch is hit first, DB is the fallback at [L1077+](https://github.com/better-auth/better-auth/blob/c8ec03d26effa0c72729b27afc2db8308d416c9a/packages/sso/src/routes/sso.ts#L1077-L1130).
- [OIDC callback](https://github.com/better-auth/better-auth/blob/c8ec03d26effa0c72729b27afc2db8308d416c9a/packages/sso/src/routes/sso.ts#L1390-L1404) — same pattern, DB fallback at [L1406+](https://github.com/better-auth/better-auth/blob/c8ec03d26effa0c72729b27afc2db8308d416c9a/packages/sso/src/routes/sso.ts#L1406-L1426).

Keeping this PR focused on the `spMetadata` fix. Happy to open a separate one to refactor those handlers onto the shared helper if maintainers think the inline duplication is worth tidying.

Issue #9186, opened 2026-04-14
tracking: admin plugin: super-admin scope, impersonation, permission math


gustavovalverde:
## Summary
This tracker covers the admin-plugin work that sits outside dynamic-role lifecycle changes. The open sub-issues focus on cross-org super-admin scope, impersonation behavior, team-role assignment, permission union math, role helpers, schema alignment, client cache invalidation, and role data on user list responses. One typing symptom at the invite endpoint belongs here because it follows the broader admin endpoint pattern.

## Root cause
The admin plugin models same-org administration more directly than cross-org administration. Super-admin authority is not expressed as one explicit scope. Impersonation is exposed as an API request without one documented contract or one aligned schema for `impersonatedBy`. Team-role assignment is missing. Permission checks do not always union multiple assigned roles correctly. Access-control overrides cannot fully replace built-in behavior. Client role state can also remain stale after sign-out, and some admin endpoints do not expose a consistent dynamic-role TypeScript surface or return org role data in user-shaped responses.

## Scope
**In:** platform admin vs org admin authority, impersonation contract and schema, team-role assignment, permission union math, access-control override, client-side role-cache invalidation, dynamic-role TS surface on admin endpoints, role surfacing on user list responses.

**Out:** dynamic-role creation/rename lifecycle (covered by #4557 rescoped tracker). Session cookie cache invalidation (cross-reference to `tracking: session cookie cache`).

## Resolution criteria
- Super-admin authority is expressed and enforced across orgs.
- Impersonation has a documented contract; `impersonatedBy` types align with the id strategy.
- Team-role assignment has a stable API.
- `hasPermission` returns the correct union across multiple role assignments.
- A user-shaped role helper exists.
- Built-in access-control rules are overridable.
- The client invalidates the role cache on sign-out without a manual refetch.
- Dynamic-role operations typecheck at every endpoint that accepts a role input, including invite.
- User list responses carry org role where applicable.

## Related
- Sibling: `tracking: admin/org dynamic-role lifecycle` (#4557 rescoped).
- Cross-reference: `tracking: session cookie cache` for the role-cache invalidation path.
- Prior closed context: #7315 (endpoint/docs mismatch).

Issue #8807, opened 2026-03-27
bug(email-otp): change-email doesn't trigger session signal


SeriousMonk:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1.  Enable emailOTP plugin with changeEmail enabled.
2. Follow the documentation steps to change the email using otp https://better-auth.com/docs/plugins/email-otp#usage-1
3. Have a component using the useSession() hook and displaying the user email
4. The component with the email does not update after the email is successfully changed

### Current vs. Expected behavior

When following the procedure to change the email using an OTP described here https://better-auth.com/docs/plugins/email-otp#usage-1 the session is not revalidated after ```authClient.emailOtp.changeEmail``` method is done. 

Therefore any components using the ```useSession()``` hook do not update.

It appears the same thing was happening with the emailOtp email-verification and was fixed in this [PR](https://github.com/better-auth/better-auth/pull/5219)

### What version of Better Auth are you using?

1.5.5

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Pro",
    "release": "10.0.26200",
    "cpuCount": 20,
    "cpuModel": "12th Gen Intel(R) Core(TM) i7-12700K",
    "totalMemory": "31.80 GB",
    "freeMemory": "11.96 GB"
  },
  "node": {
    "version": "v24.5.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.5.1"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Client

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  plugins: [
    emailOTP({
      changeEmail: { enabled: true },
    })
  ]
});
```

PR #9402, opened 2026-04-29
fix(admin): revalidate useSession after impersonation

### LINKED ISSUE #318: reactive values becomes outdated when `useUserSession` reactive values changes
While continuing my development from PR commit #295, I come across the same issue of outdated session.

In one such case I am triggering user impersonation and it is updating user value in the store but it is not updating user elsewhere in my app where I used `useUserSession` composable. I believe this is the same issue that we were facing with middleware, where even after sign out the value was not being reflected in middleware. 

I think we should have a shared composable state instead of separate reactive value initialization wherever these composable is used. The current pattern won't only produce unlikely errors in the module but also in user application, like the above mentioned case.

### LINKED ISSUE #9401: bug(admin): impersonateUser does not revalidate useSession

### To Reproduce

Using the Better Auth vanilla client with the admin plugin:

1. Sign in as an admin user.
2. Read the current session from `authClient.useSession()`.
3. Call `authClient.admin.impersonateUser({ userId })`.
4. Read the session again from:
   - the reactive `useSession()` store
   - a fresh `authClient.getSession()` request

Current behavior:

- the server session changes successfully
- a fresh `getSession()` call returns the impersonated user
- the reactive `useSession()` store still points at the previous admin user until a manual `refetch()` is called

Expected behavior:

- `authClient.admin.impersonateUser(...)` should invalidate the client session store the same way other session-mutating routes do, so `useSession()` updates immediately to the impersonated user

1.6.9

```bash
Node 25
vanilla Better Auth client
Express server
```

Client

### Additional context

This appears to be the same class of bug as previous missing session-signal invalidation fixes, for example:

- #1286
- #8807
- #5741
- #927
- #7877
- PR #5219

The current admin client registers `"/admin/stop-impersonating"` as a mutating route, but it does not register `"/admin/impersonate-user"` for session invalidation.

onmax:
Closes #9401.
Closes https://github.com/nuxt-modules/better-auth/issues/318

This adds the missing session invalidation metadata for the admin impersonation routes. `admin.impersonateUser()` already updates the server session, but the admin client was not marking `/admin/impersonate-user` as a mutating route and was not emitting `$sessionSignal`, so `useSession()` stayed stale until a manual refetch.

The fix follows the same route-signal pattern used in #5219 instead of introducing a new invalidation mechanism.

Repros:
- Bug: https://github.com/onmax/repros/tree/repro/better-auth-9401/better-auth-9401
- Fix: https://github.com/onmax/repros/tree/repro/better-auth-9401/better-auth-9401-fix

Related context:
- #9186
- #1286
- #8807
- #5219

---
## Summary by cubic
Fixes stale `useSession` after admin impersonation. The admin client now revalidates the session on start/stop impersonation so the UI updates without a manual refetch (closes #9401).

- **Bug Fixes**
  - Marked `/admin/impersonate-user` and `/admin/stop-impersonating` as mutating in `pathMethods` and emit `$sessionSignal` via `atomListeners` to revalidate `useSession`.
  - Added a regression test that waits for `useSession` hydration and avoids fake timers; added a patch changeset for `better-auth`.

<sup>Written for commit 26a15fcf451d04470c1fc9f109b7a3ad43477c3e. Summary will update on new commits.</sup>


ping-maxwell (approved):
nice work

PR #9410, opened 2026-04-30
fix(organization): listMembers fails with "User not found" for orgs with 100+ members

### LINKED ISSUE #9407: organizations `listMember` fails with 100+ users with `User not found for member`

### Reproduction

Have more than 100 users in an org e.g. 200. 
Set the membershipLimit to 500
Call listMembers

Fails to return users

main

### System info

```bash
main
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

Members are compared to users and only the members query sets the limit explicitly to memershipLimit.

https://github.com/better-auth/better-auth/blob/6cc30e79504143cb5e3481d99431fb6538437305/packages/better-auth/src/plugins/organization/adapter.ts#L144-L189

Fetching users fetches without explicit limit. Copying the limit clause from above makes everything work.

https://github.com/better-auth/better-auth/blob/6cc30e79504143cb5e3481d99431fb6538437305/packages/better-auth/src/plugins/organization/adapter.ts#L190-L199

TanishValesha:
Fixes #9407

- [x] pnpm typecheck (repo root)

`listMembers` was throwing `"Unexpected error: User not found for member"` for organizations with more than 100 members.

## Root Cause

`listMembers` in `packages/better-auth/src/plugins/organization/adapter.ts` runs two sequential queries:

1. **Fetch members** — correctly limited by `data.limit || membershipLimit || 100`
2. **Fetch users** for those members — had **no `limit`** at all

The Prisma adapter's `findMany` defaults to `take: limit || 100` when no limit is provided, silently capping the user results at 100. Any member beyond index 100 had no matching user in the result, causing the error.

The Kysely and Drizzle adapters happen not to apply a default cap, which is why the bug was Prisma-specific.

## Fix

Added the same `limit` expression to the users `findMany` call that was already on the members query, consistent with how `findFullOrganization` handles the same pattern (fixed previously in #4724):

```diff
  const users = await adapter.findMany<User>({
    model: "user",
    where: [{ field: "id", value: members[0].map((m) => m.userId), operator: "in" }],
+   limit:
+     data.limit ||
+     (typeof options?.membershipLimit === "number"
+       ? options.membershipLimit
+       : 100) ||
+     100,
  });

### DIFF
--- a/.changeset/strong-meals-hang.md
+++ b/.changeset/strong-meals-hang.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Fix listMembers failing with "User not found" error when an organization has more than 100 members
--- a/packages/better-auth/src/plugins/organization/adapter.ts
+++ b/packages/better-auth/src/plugins/organization/adapter.ts
@@ -196,6 +196,12 @@ export const getOrgAdapter = <O extends OrganizationOptions>(
 						operator: "in",
 					},
 				],
+				limit:
+					data.limit ||
+					(typeof options?.membershipLimit === "number"
+						? options.membershipLimit
+						: 100) ||
+					100,
 			});
 			return {
 				members: members[0].map((member) => {


gustavovalverde:
Closing in favor of #9411. #9411's `members[0].length` fix is strictly tighter than the `data.limit || membershipLimit || 100` expression here (it cannot underflow when the caller passes a small `limit`), and #9411 includes the 120-member regression test.

PR #9411, opened 2026-04-30
fix(organization): resolve listMembers missing users for orgs with >100 members

### LINKED ISSUE #9407: organizations `listMember` fails with 100+ users with `User not found for member`

MuzzaiyyanHussain:
Fixes #9407

Ensure `listMembers` returns all users when organizations have >100 members by applying the same limit to the users query. Adds a regression test to prevent this issue from reoccurring.


gustavovalverde:
#9410 was closed in favor of this one. Things to consider from that PR:
- Carry over `.changeset/strong-meals-hang.md` from #9410 before merging (#9411 currently lacks the changeset).

bytaesu:
The linked issue was fixed by #10342

Closing this PR.

PR #9418, opened 2026-05-01
feat: add JWT client assertion OAuth2 support

ItalyPaleAle:
This is based on RFC 7523

It allows configuring an OAuth2 provider with a `clientAssertionProvider` instead of a `clientSecret`, so omitting long-lived credentials.

This PR is currently not concerned with what providers support (or will support) client assertions. It is also not concerned with _how_ the assertion is obtained: could be a Kubernetes token, a token from a cloud platform, etc.

## Example: Microsoft Entra ID with a certificate-signed JWT assertion

This is the canonical `private_key_jwt` flow: instead of storing a client secret, you register a certificate (public key) with Entra ID and sign a JWT with the corresponding private key.

```ts
import {
  genericOAuth,
  microsoftEntraId,
} from "better-auth/plugins/generic-oauth";
import { SignJWT, importPKCS8 } from "jose";
import { randomUUID } from "node:crypto";

const tenantId = process.env.MS_TENANT_ID!;
const clientId = process.env.MS_CLIENT_ID!;

// Load the private key once at startup
const privateKey = await importPKCS8(process.env.MS_PRIVATE_KEY_PEM!, "RS256");

// x5t#S256 thumbprint header that Entra expects to identify the certificate
const x5tS256 = process.env.MS_CERT_THUMBPRINT_SHA256!;

async function getEntraClientAssertion(): Promise<string> {
  const now = Math.floor(Date.now() / 1000);
  return new SignJWT({})
    .setProtectedHeader({ alg: "RS256", typ: "JWT", "x5t#S256": x5tS256 })
    .setIssuer(clientId)
    .setSubject(clientId)
    .setAudience(
      `https://login.microsoftonline.com/${tenantId}/oauth2/v2.0/token`,
    )
    .setJti(randomUUID())
    .setIssuedAt(now)
    .setExpirationTime(now + 5 * 60) // short-lived: 5 minutes
    .sign(privateKey);
}

  plugins: [
    genericOAuth({
      config: [
        microsoftEntraId({
          clientId,
          tenantId,
          // No clientSecret — authenticate with a JWT assertion instead
          clientAssertionProvider: getEntraClientAssertion,
        }),
      ],
    }),
  ],
});
```

## Example: generic OAuth2 provider, assertion from a cloud provider (Vercel)

For an application running on Vercel and authenticating with a generic OAuth2 provider (e.g. Pocket ID), you can now configure better-auth with:

```ts
import { getVercelOidcToken } from '@vercel/oidc'

genericOAuth({
  config: [
    {
      providerId: "my-idp",
      discoveryUrl: "https://idp.example.com/.well-known/openid-configuration",
      clientId: process.env.IDP_CLIENT_ID!,
      // Replaces clientSecret
      clientAssertionProvider: async (): Promise<string> => {
        return getVercelOidcToken()
      },
      pkce: true,
    },
  ],
});
```

You then just need to configure your application in the IdP to accept federation with these values:

- Issuer: `https://oidc.vercel.com/<vercel-team>`
- Audience: `https://vercel.com/<vercel-team>`
- Subject: `owner:<vercel-team>:project:<project-name>:environment:production`


ItalyPaleAle:
Closed in favor of #9625, targeting the next branch

PR #9431, opened 2026-05-01
feat: make `Auth` instance fetchable

pi0:
This PR adds `auth.fetch` as alias to `auth.handler`.

I am trying to figure out super minimal or ideally zero config support for better-auth via `auth.ts` [for Nitro](https://github.com/nitrojs/nitro/pull/3942/). Ideally something like this is all one will have to do:

```ts
// auth.ts
export default betterAuth({ /* ... */ })
```

Today, modern http frameworks (H3, Hono, Elysia) and platforms (Vercel, Cloudflare, Deno) and Runtimes (Bun, Deno and srvx) support fetchable default exports. `export default app` / `export default { fetch }` works out of the box.

Nitro v3 too supports this as a universal convention. By this change, a simple `auth.ts` in any Nitro app can be elegantly used for both better-auth CLI migrator and route handler for `/api/auth/**`.

## Alternatives

Today one can make a `api/auth/[..].ts` that basically reexports `auth.handler` in a wrapper. It adds an unncessary extra file.

Anther alternative could be `export const auth` + `export const fetch = auth.handler` but this pattern is really not supported (both runtimes and nitro expect a default export with `.fetch` attached to) and is also less intuative. 

Relavant fixes: https://github.com/better-auth/better-auth/pull/9477

---
## Summary by cubic
Adds `auth.fetch` as an alias to `auth.handler`, making `Auth` instances directly fetchable for zero‑config default exports in Nitro v3 and other fetch-based runtimes.

- **New Features**
  - Added `auth.fetch(request)`, identical to `auth.handler(request)`, for platforms expecting a default `fetch`.
  - Exposed `Auth.fetch` in types; release as a minor update in `better-auth`.

<sup>Written for commit e8d8f266bce00fdf4203bcda94edbb0cf6e4d490. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/hungry-drinks-read.md
+++ b/.changeset/hungry-drinks-read.md
@@ -1,0 +1,5 @@
+---
+"better-auth": minor
+---
+
+feat: make `Auth` instance fetchable
--- a/packages/better-auth/src/auth/base.ts
+++ b/packages/better-auth/src/auth/base.ts
@@ -26,59 +26,58 @@ export const createBetterAuth = <Options extends BetterAuthOptions>(
 		}
 		return acc;
 	}, {});
-	return {
-		handler: async (request: Request) => {
-			const ctx = await authContext;
-			const basePath = ctx.options.basePath || "/api/auth";
+	const handler = async (request: Request) => {
+		const ctx = await authContext;
+		const basePath = ctx.options.basePath || "/api/auth";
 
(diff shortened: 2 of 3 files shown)


Bekacru (approved):

bytaesu:
Thanks for working on this, @pi0 😁

PR #9445, opened 2026-05-04
feat(sso): support additionalFields on ssoProvider

### LINKED ISSUE #9396: Support additionalFields on sso_provider in SSO plugin

I want to add a display name to each sso provider such that I can let users choose between multiple registered SSO providers per organization.

I expected to be able to define additionalFields through a schema config just like for `user`, `organization`.

Using the providerId is a poor solution a) due to formatting and url-compatibility issues and b) because the display name must only be unique per organization not globally.

A sidecar table with metadata is possible but updates don't run in the same transaction and the solution requires custom endpoints for SSO related apis.

### Additional context

_No response_

GautamBytes:
## Summary
- Adds `additionalFields` support to `ssoProvider`, matching the pattern used by `user`, `organization`, and other core tables
- Fields are validated/parsed on create and update, and filtered on read

## Changes
- Types: `SSOProviderAdditionalFields`, `InferSSOProvider`, `SSOProviderSchema`, `SSOOptions.schema`
- Runtime helpers (`additional-fields.ts`): parse, filter, and Zod schema generation
- Routes: register, list, get, and update endpoints handle additional fields
- Client: schema option for type inference
- Tests: required validation, store/return, update, and type inference

Closes #9396

---
## Summary by cubic
Adds support for custom `schema.ssoProvider.additionalFields` in `@better-auth/sso`, with validation, storage, filtering, and end-to-end type inference. Hardens provider field handling to prevent collisions and disallow unsafe input, enabling provider-specific metadata for #9396.

- **New Features**
  - Schema: `additionalFields` with `required`, `input`, `defaultValue`, `returned`, and `fieldName`; can override `modelName` and per-field names via `schema.ssoProvider.fields` (incl. `domainVerified`).
  - Runtime: helpers to generate Zod schemas and parse/filter additional fields on create/update; output filtering via `filterOutputFields`.
  - Endpoints: `register`/`update` parse and persist additional fields; `register`/`get`/`list`/`update` return only fields with `returned !== false`.
  - Types/Client: `SSOProviderSchema`, `InferSSOProvider`, `SSOProviderAdditionalFieldsInput`; plugin exposes `$Infer.SSOProvider`; `ssoClient({ schema })` enables type inference. Docs add an example.

- **Bug Fixes**
  - Reject additional field keys that collide with built-in or response-only keys, and fields whose `fieldName` maps to built-in columns.
  - Block `input: false` fields on register/update and require at least one valid change on update; sanitize all responses to include only allowed fields.

<sup>Written for commit 8e4f4d41fbcb0710d80ca4f7b388af5acd1e0957. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9445?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


GautamBytes:
@ping-maxwell  @bytaesu ready for review!

packages/sso/src/index.ts
```diff
@@ -253,52 +272,88 @@ export function sso<O extends SSOOptions>(
...
+								domainVerified: {
+									type: "boolean",
+									required: false,
+									fieldName:
+										options?.schema?.ssoProvider?.fields?.domainVerified ??
+										"domainVerified",
+								},
+							}
 						: {}),
+					...(options?.schema?.ssoProvider?.additionalFields ?? {}),
```
gustavovalverde, line 422: Custom fields are merged after the built-in `ssoProvider` fields, so a config can define `additionalFields.providerId`, `domain`, `issuer`, etc. and replace the core DB metadata, including `providerId.unique`.

Reject additional-field keys that collide with built-in `ssoProvider` fields and keep the built-ins authoritative.

packages/sso/src/providers.test.ts
```diff
@@ -1116,6 +1121,142 @@ describe("SSO provider read endpoints", () => {
...
+			});
+
+			const registered = (await registerSAMLProvider(
+				headers,
+				"okta",
+				undefined,
+				{
+					displayName: "Okta",
+					internalCode: "secret",
+					source: "client",
```
gustavovalverde: This locks in silently ignoring an `input: false` field supplied by the caller: `source: "client"` later becomes the default `"system"`. The existing additional-fields contract says `input: false` prevents callers from passing a value, so this should reject the request, or the SSO docs need to explicitly document the silent-ignore behavior.

packages/sso/src/additional-fields.ts
```diff
@@ -0,0 +1,52 @@
+import type { DBFieldAttribute } from "@better-auth/core/db";
```
gustavovalverde: This new top-level helper is a table-specific wrapper used only by the SSO route files. Since `routes/schemas.ts` now owns the register/update request schemas after rebasing, this split creates a pattern future plugin code will copy without a real boundary. 

Fold request-schema construction into `routes/schemas.ts` and keep response filtering next to the provider sanitizer/presenter.

gustavovalverde (changes requested):

GautamBytes:
@gustavovalverde done with the changes!

PR #9463, opened 2026-05-06
fix(device-authorization): make schema option optional in Zod schema

### LINKED ISSUE #9422: deviceAuthorization() throws ZodError on import when no schema option is passed
### Bug

Importing the `deviceAuthorization` plugin and calling it with no `schema` option crashes at module load with a `ZodError`:

```
ZodError: [
  {
    "code": "invalid_type",
    "expected": "nonoptional",
    "path": ["schema"],
    "message": "Invalid input: expected nonoptional, received undefined"
  }
]
```

### Repro

```ts
import { deviceAuthorization } from "better-auth/plugins";

deviceAuthorization({ verificationUri: "https://example.com/device" });
// ZodError thrown at module evaluation
```

- `better-auth`: 1.4.22 (also reproduces on 1.6.9, the latest at time of filing)
- `zod`: 4.4.1 (peerDependency on better-auth: `^4.3.5`)
- Runtime: Bun 1.3.9 / Node 22

### Root cause

`packages/better-auth/src/plugins/device-authorization/index.ts:122`:

```ts
schema: z.custom<InferOptionSchema<typeof schema>>(() => true),
```

Under Zod v4, `z.custom()` is **non-optional by default**, so `deviceAuthorizationOptionsSchema.parse(options)` rejects `undefined` for the `schema` key. Every consumer that omits `schema` (which is the documented default) crashes at import time.

The plugin already treats `schema` as optional internally — the next lines call `mergeSchema(schema, options?.schema)` with optional chaining.

### Suggested fix

Add `.optional()`:

```diff
-  schema: z.custom<InferOptionSchema<typeof schema>>(() => true),
+  schema: z.custom<InferOptionSchema<typeof schema>>(() => true).optional(),
```

Happy to open a PR with this change plus a regression test if useful.

### Workaround

Pass an empty object explicitly: `deviceAuthorization({ schema: {}, ... })`. `mergeSchema` treats `{}` as a no-op rename map.

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/9422
- Add `.optional()` to the `schema` field in `deviceAuthorizationOptionsSchema` to align with all other `z.custom()` fields in the same schema
- Add regression tests to ensure the schema option remains optional


bytaesu:
Closing this PR since #9422 is already addressed by https://github.com/better-auth/better-auth/pull/9939

PR #9467, opened 2026-05-06
fix(core): getSessionCookie fails when Cookie header uses ";" without…

### LINKED ISSUE #9465: `getSessionCookie` fails when `Cookie` header uses `";"` without space as separator (AWS API Gateway / AWS Lambda Web Adapter)

### Reproduction

`getSessionCookie` returns `null` even when the session cookie is present in the request, when the app is deployed on **AWS Lambda behind API Gateway (v2) using the AWS Lambda Web Adapter**.

The root cause is that `parseCookies` splits the `Cookie` header on `"; "` (semicolon + space), but AWS Lambda Web Adapter forwards the header with cookies separated by `";"` (semicolon only, no space). This causes the entire cookie string to be treated as a single unparseable token.

## Steps to reproduce

1. Deploy a Next.js + better-auth application to AWS Lambda behind API Gateway (or any proxy that strips spaces from cookie separators).
2. Sign in — cookies are set correctly in the browser (`__Secure-better-auth.session_token`, etc.).
3. Make a subsequent request. The `Cookie` header arrives at Lambda as:
```
   __Secure-better-auth.account_data.0=eyJ...;__Secure-better-auth.session_token=qlTt...
```
   Note: `";"` with **no space**.
4. Call `getSessionCookie(req)` in middleware/proxy — it returns `null`.

## Evidence

Logging the raw `Cookie` header on the request in the `proxy.ts` (proxy/middleware) in NextJS 16:

```typescript
import { getSessionCookie } from 'better-auth/cookies';
// ...

export default function proxy(req: NextRequest): NextResponse {
  // ...

  const cookies = getSessionCookie(req);
  const isAuthenticated = !!cookies;
  const cookieNames = req.cookies.getAll().map((c) => c.name);
  const cookieHeader = req.headers.get('cookie');
  const cookieHeaderSize = cookieHeader?.length ?? 0;
  log.debug(
    {
      account,
      accountPath,
      isAuthenticated,
      cookieNames,
      cookieHeader,
      cookieHeaderSize,
    },
    'proxy: routing',
  );

  // ...
}
```

resulted in logs containing the following information:

```json
{
  // ...
  "cookieNames": [
    "__Secure-better-auth.account_data.0",
    "__Secure-better-auth.account_data.1",
    "__Secure-better-auth.session_token",
    "__Secure-better-auth.session_data"
  ],
  "cookieHeader": "__Secure-better-auth.account_data.0=eyJ...;__Secure-better-auth.account_data.1=t1UC...;__Secure-better-auth.session_token=qlTt7H9x...;__Secure-better-auth.session_data=eyJhbG...",
  "isAuthenticated": false
}
```

`req.cookies.getAll()` correctly enumerates all four cookies (Next.js likely uses a lenient parser). `getSessionCookie` returns `null` due to `parseCookies` spliting on `"; "`; see below.

## Root cause

The internal `parseCookies` function at [`packages/better-auth/src/cookies/index.ts#L359`](https://github.com/better-auth/better-auth/blob/a597ee01ed4e6d85aba5ee9f15100acc578390d9/packages/better-auth/src/cookies/index.ts#L359):

```ts
function parseCookies(cookieHeader) {
    const cookies = cookieHeader.split("; ");
    // ...
}
```

splits on the literal string `"; "`. When API Gateway forwards the `Cookie` header to Lambda via the AWS Lambda Web Adapter, the cookies are joined with `";"` only — no trailing space — so the entire header value is treated as a single token and `parsedCookie.get(...)` finds nothing.

## Current behavior

`getSessionCookie` returns `null` even when the session cookie is present in the request, when the app is deployed on **AWS Lambda behind API Gateway**. The root cause is that `parseCookies` splits the `Cookie` header on `"; "` (semicolon + space), but API Gateway forwards the header with cookies separated by `";"` (semicolon only, no space). This causes the entire cookie string to be treated as a single unparseable token.

## Expected behaviour

`getSessionCookie` should find the session cookie regardless of whether the `Cookie` header separator is `"; "` or `";"` even if this is not the strictest implemenation of the RFC (see additional information).

1.6.9

### System info

```bash
{
  "node": {
    "version": "v24.14.1",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.13.0"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.2.4"
    },
    {
      "name": "react",
      "version": "19.2.5"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "^1.6.9",
    "config": {
      "baseURL": "https://canterly.com",
      "secret": "[REDACTED]",
      "session": {
        "cookieCache": {
          "enabled": true,
          "strategy": "jwt",
          "maxAge": 86400,
          "refreshCache": true
        }
      },
      "socialProviders": {
        "cognito": {
          "clientId": "[REDACTED]",
          "clientSecret": "[REDACTED]",
          "region": "[REDACTED]",
          "userPoolId": "[REDACTED]",
          "domain": "[REDACTED]"
        }
      }
    }
  }
}
```

Backend

Please note that I stripped the `system` portion above which was my local system and provides no information about the Lambda environment this application runs on. This is more relevant:

- **better-auth**: `v1.2.9` (bug present on current HEAD as of `a597ee0`)
- **Framework**: Next.js 16 (App Router)
- **Deployment**: AWS Lambda (Node.js runtime) behind AWS API Gateway v2 (HTTP API)
- **Architecture**: SSL terminates at API Gateway; Lambda receives plain HTTP internally. The `cookie` header is reconstructed by the AWS Lambda Web Adapter before forwarding to Lambda, and the space after `";"` is stripped in this process.

## RFC 6265 context

RFC 6265 [Section 4.2.1](https://www.rfc-editor.org/rfc/rfc6265.html) defines the canonical grammar as:

```
cookie-header = "Cookie:" OWS cookie-string OWS
cookie-string = cookie-pair *( ";" SP cookie-pair )
```

The current implementation is RFC-compliant. That said, RFC 6265 itself uses a more permissive approach in the example algorithms it provides in section 5.2 to illustrate how user-agents should parse Set-Cookie header. It seems also that Next.js own implementation is more tolerant and likely splits in `/;\s*/`. (I have not verified this, at this point, but trust it to be true given my tests and evidences.)

sbougerel:
… space as separator.

`getSessionCookie` returns null even when the session cookie is present in the request, when the app is deployed on AWS Lambda behind API Gateway (v2).

The root cause is that `parseCookies` splits the Cookie header on `"; "` (semicolon + space), but API Gateway forwards the header with cookies separated by `";"` (semicolon only, no space). This causes the entire cookie string to be treated as a single unparseable token.

- Closes: https://github.com/better-auth/better-auth/issues/9465

---
## Summary by cubic
Fix session cookie parsing in `better-auth` when the Cookie header uses ";" without a space, which caused `getSessionCookie` to return null in some gateways. `parseCookies` now splits on `/; ?/` and a test covers semicolon-only separators.

<sup>Written for commit 50ecfb09c639581786d731b96666971faae13bc9. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/silent-meals-work.md
+++ b/.changeset/silent-meals-work.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+getSessionCookie fails when Cookie header uses ";" without space as separator
--- a/packages/better-auth/src/cookies/cookies.test.ts
+++ b/packages/better-auth/src/cookies/cookies.test.ts
@@ -1437,6 +1437,20 @@ describe("parse cookies", () => {
 			"session-data.signature=",
 		);
 	});
+
+	it("should parse cookies separated only by a semi-colon", () => {
+		const cookieHeader =
+			"better-auth.session_token=session-token.signature;better-auth.session_data=session-data.signature";
+
+		const parsedCookies = parseCookies(cookieHeader);
+
+		expect(parsedCookies.get("better-auth.session_token")).toBe(
+			"session-token.signature",
+		);
+		expect(parsedCookies.get("better-auth.session_data")).toBe(
+			"session-data.signature",
+		);
+	});
 });
 
 describe("expireCookie", () => {
--- a/packages/better-auth/src/cookies/index.ts
+++ b/packages/better-auth/src/cookies/index.ts
@@ -357,7 +357,7 @@ export function deleteSessionCookie(
 }
 
 export function parseCookies(cookieHeader: string) {
-	const cookies = cookieHeader.split("; ");
+	const cookies = cookieHeader.split(/; ?/);
 	const cookieMap = new Map<string, string>();
 
 	cookies.forEach((cookie) => {


bytaesu:
Closing this in favor of https://github.com/better-auth/better-auth/pull/9543

PR #9037, opened 2026-04-08
feat: implement SSO plugin organization isolation and security hardening

surendarcs:
## 🛡️ Objective: Secure SSO Organization Isolation

This PR implements comprehensive security hardening for organization isolation when using the SSO plugin. It ensures that any user session authenticated via a specific SSO provider/organization is strictly restricted to that organization's data and management endpoints, preventing cross-tenant access.

### 🛠️ Key Changes

- **Centralized Isolation Guard**: Implemented `checkSSOIsolation` to intercept and block any management requests (Teams, Invites, Members, etc.) that target a different organization than the one the user is SSO-scoped to.
- **SSO Plugin Hardening**: Injected isolation checks into all critical SSO routes:
  - `registerSSOProvider`
  - `getSSOProvider`, `updateSSOProvider`, `deleteSSOProvider`
  - `requestDomainVerification` and `verifyDomain`
- **Smart Resource Filtering**: Patched `listOrganizations` and `listUserTeams` to automatically filter results, ensuring users in SSO sessions only see their authorized organization and its resources.
- **Session Lifecycle Integration**: Updated the SAML callback logic to automatically inject and persist the `ssoOrganizationId` in the session upon successful sign-in.

### ✅ Verification Results

I have implemented and verified these changes using a new E2E test suite: `packages/sso/src/sso-isolation.test.ts`.

**Test Results Summary:**
- **Step 1: listOrganizations Filtering**: Confirmed SSO sessions only see 1 authorized organization instead of all memberships.
- **Step 2: Team Management**: Confirmed `403 Forbidden` when attempting to update teams in another organization.
- **Step 3: SSO Provider Isolation**: Confirmed `403 Forbidden` when fetching providers from different organizations.
- **Step 4: Domain Verification**: Confirmed isolation for domain ownership verification.

All tests passed successfully.

---
## Summary by cubic
Adds strict SSO organization isolation with a closed-by-default guard across organization and SSO management routes to prevent cross-tenant access. SSO sessions are now scoped via new session fields, and endpoint instrumentation is clarified for cleaner traces.

- New Features
  - Introduced `checkSSOIsolation` in `better-auth` and `@better-auth/sso`; enforced on org routes (roles, invites, members, org, teams) and SSO routes (provider CRUD, registration, domain verification). When a session has `ssoOrganizationId`, cross-org requests are denied.
[...]

Issue #9013, opened 2026-04-07
User logged in via SSO of one Org able to access other.


opswiz:
### Is this suited for github?

- [x] Yes, this is suited for github

### Reproduction

1. Create Two Orgs
2. Configure different SSO providers in the org
3. Login to the app using one IDP
4. You can access both the organizations

### Current vs. Expected behavior

Currently there is an issue, Lets say user belong to multiple organization. Where each organization have different SSO configured. If user logs in to the application via SSO from the IDP he gets redirected back to the app and a session is created, by default he is able to access both the organization even for the one which he did not authentication with SSO.

isn't this bad? shouldn't he be able to access only the org which he SSO'ed into? Am missing something or missing any configuration?

### What version of Better Auth are you using?

1.5.6

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.4.0: Thu Mar 19 19:33:09 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T8112",
    "release": "25.4.0",
    "cpuCount": 8,
    "cpuModel": "Apple M2",
    "totalMemory": "8.00 GB",
    "freeMemory": "0.11 GB"
  },
  "node": {
    "version": "v24.14.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.9.0"
  },
  "frameworks": [
    {
      "name": "react",
      "version": "19.2.4"
    },
    {
      "name": "hono",
      "version": "^4.12.11"
    }
  ],
  "databases": [
    {
      "name": "@prisma/client",
      "version": "^7.6.0"
    }
  ],
  "betterAuth": {
    "version": "^1.5.6",
    "config": {
      "secret": "[REDACTED]",
      "baseURL": "http://localhost:5150",
      "session": {
        "expiresIn": 172800,
        "updateAge": 86400,
        "cookieCache": {
          "enabled": true,
          "maxAge": 300
        }
      },
      "advanced": {
        "trustedProxyHeaders": true,
        "defaultCookieAttributes": {
          "sameSite": "Strict"
        }
      },
      "account": {
        "accountLinking": {
          "allowDifferentEmails": false,
          "allowUnlinkingAll": true
        }
      },
      "emailVerification": {
        "autoSignInAfterVerification": true
      },
      "emailAndPassword": {
        "enabled": false
      },
      "databaseHooks": {
        "user": {
          "create": {}
        }
[...]

PR #9471, opened 2026-05-06
fix(sso): filter listSSOProviders by activeOrganizationId

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/9031
- When `activeOrganizationId` is set in the session, `listSSOProviders` now only returns providers from that organization
- Previously, the endpoint returned providers from ALL organizations where the user is an admin, which leaked SSO provider metadata across organizations
- When no active organization is set, the previous behavior is preserved (returning all providers from all orgs where user is admin)


ping-maxwell:
cc @Paola3stefania

gustavovalverde:
#9037 was closed in favor of this one (with #9220 already merged); nothing specific to carry over (the broad isolation policy was declined in #9013).

PR #9473, opened 2026-05-06
fix(stripe): lock library-owned Checkout Session fields against `getCheckoutSessionParams`

bytaesu:
- Following up from #9474 

`getCheckoutSessionParams` is designed to allow extending checkout session params beyond what the library manages internally. Allowing it to override internally-owned checkout session fields could lead to unexpected bugs.

While this can be seen as a behavior change (-> beta track), the docs already present it as a way to add custom fields, and this aligns with the intended purpose of `getCheckoutSessionParams`, so this is being addressed as a stable track.

+) Mutually exclude `customer` / `customer_email` so the unused side cannot leak through.

---

### DIFF
--- a/.changeset/fix-stripe-lock-checkout-fields.md
+++ b/.changeset/fix-stripe-lock-checkout-fields.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/stripe": patch
+---
+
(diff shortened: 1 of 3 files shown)


bytaesu:
-> will re-open this

Issue #9469, opened 2026-05-06
Username SignIn does not redirect like Email does


chdanielmueller:
### Is this suited for github?

- [x] Yes, this is suited for github

### Reproduction

The signIn Function for email does redirect to the callbackURL after successful sign in.
```typescript
await authClient.signIn.email({
  email,
  password,
  rememberMe,
  callbackURL: '/'
})
```

The signIn Function for username does NOT redirect to the callbackURL.
```typescript
await authClient.signIn.username({
  username,
  password,
  rememberMe,
  callbackURL: '/'
})
```

### Current vs. Expected behavior

I expect the two signIn Functions to behave in the same way.

### What version of Better Auth are you using?

1.6.9

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "arm64",
    "version": "#1 SMP Fri Apr 17 14:56:37 UTC 2026",
    "release": "6.12.76-linuxkit",
    "cpuCount": 12,
    "cpuModel": "unknown",
    "totalMemory": "23.43 GB",
    "freeMemory": "15.94 GB"
  },
  "node": {
    "version": "v24.15.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.12.1"
  },
  "frameworks": [
    {
      "name": "express",
      "version": "5.2.1"
    }
  ],
  "databases": [
    {
      "name": "mongodb",
      "version": "7.2.0"
    }
  ],
  "betterAuth": {
    "version": "1.6.9",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Client

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
import { username } from "better-auth/plugins";
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
  plugins: [
    username(),
  ],
});
```

PR #9475, opened 2026-05-06
fix(username): respect callbackURL on sign-in

jaydeep-pipaliya:
`/sign-in/email` honours `callbackURL` — it sets a `Location` header and returns `{ redirect, url }`. `/sign-in/username` accepts the same field on the body but ignores it: the handler just returns `{ token, user }`, so `authClient.signIn.username({ ..., callbackURL: '/' })` silently does nothing while the email equivalent redirects.

Reported in #9469.

This mirrors the email-and-password flow in the username sign-in handler: set `Location` when `callbackURL` is present, and add `redirect`/`url` to the JSON response. The OpenAPI metadata and the body field description are updated to match.

Two regression tests added in `username.test.ts` covering the with- and without-`callbackURL` paths. Existing 33 tests still pass; full username suite is 35.

---
## Summary by cubic
Make `/sign-in/username` honor `callbackURL` on sign-in, matching the email flow for consistent client redirects.

- **Bug Fixes**
  - Set `Location` when `callbackURL` is provided and return `redirect` and `url` in JSON.
  - Updated OpenAPI schema and body field description; added regression tests for with/without `callbackURL`; included a changeset.

<sup>Written for commit 773b4fb2cdc7e677a103d3f7d02c5a76c2e759b6. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/fix-username-callback-url-redirect.md
+++ b/.changeset/fix-username-callback-url-redirect.md
@@ -1,0 +1,11 @@
+---
+"better-auth": patch
+---
+
+fix(username): respect callbackURL on `/sign-in/username`
+
+The endpoint accepted a `callbackURL` body field but ignored it, so
+`authClient.signIn.username({ ..., callbackURL })` silently did nothing
+while `authClient.signIn.email` redirected as expected. The handler now
+sets a `Location` header when `callbackURL` is provided and returns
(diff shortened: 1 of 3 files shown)


ping-maxwell (changes requested):
Looks good, just missing changeset.

ping-maxwell (approved):

Issue #4454, opened 2025-09-05
Allow an admin to set a user password


maelp:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

Right now, `auth.api.setPassword` requires the current user session to be given, in order to allow the user to change his password.

But there are cases (like seeding user in test database, or providing an ad-hoc mechanism to set initial passwords by the admin) where we would want to have a `setPassword({userId, newPassword})` (only accessible in the backend)

### Describe the solution you'd like

Allow setPassword to be called with a `userId` as parameter (and not using the user in the current session — which might not exist when running from an automated script)

### Describe alternatives you've considered

Re-implement the creation of the corresponding account, but this is error-prone

ping-maxwell:
we have auth.api.setUserPassword which does this

PR #9482, opened 2026-05-07
fix(admin): create credential account in setUserPassword when missing

### LINKED ISSUE #3553: Create an account on setUserPassword if account does not exist
Hi, I’m migrating users to better-auth and want them to reset their passwords. However, `auth.api.setUserPassword` fails if the user doesn’t already have a `credential` account in the `Account` table.

New users work fine with `signUp` since it creates both the user and the account.

Can setUserPassword fallback to creating a credential account if one doesn’t exist?

bytaesu:
Mirror the existing branch in resetPassword(https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/api/routes/password.ts#L300-L311):

Look up the accounts, update if a credential account exists, otherwise create one (Upsert)

- Closes #3553
Related #4454

### DIFF
--- a/.changeset/admin-set-password-creates-credential.md
+++ b/.changeset/admin-set-password-creates-credential.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+`admin.setUserPassword` now creates a credential account when the target user does not have one, matching the behavior of `resetPassword`. Previously the call returned `status: true` without doing anything for users without an existing credential account (e.g., social-only or magic-link signups), so admins migrating users from another auth system or assigning an initial password to a social-only user can now do so directly without poking the `account` table.
--- a/packages/better-auth/src/plugins/admin/routes.ts
+++ b/packages/better-auth/src/plugins/admin/routes.ts
@@ -1717,8 +1717,28 @@ export const setUserPassword = (opts: AdminOptions) =>
 				ctx.context.logger.error("Password is too long");
 				throw APIError.from("BAD_REQUEST", BASE_ERROR_CODES.PASSWORD_TOO_LONG);
 			}
+			const user = await ctx.context.internalAdapter.findUserById(userId);
+			if (!user) {
+				throw APIError.from("NOT_FOUND", BASE_ERROR_CODES.USER_NOT_FOUND);
+			}
 			const hashedPassword = await ctx.context.password.hash(newPassword);
-			await ctx.context.internalAdapter.updatePassword(userId, hashedPassword);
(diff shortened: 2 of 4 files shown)


.changeset/admin-set-password-creates-credential.md
```diff
@@ -0,0 +1,5 @@
+---
+"better-auth": minor
```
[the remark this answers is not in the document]
bytaesu: I think creating a new account is a significant behavior change that should be clearly called out in the release note blog, so I'm going to keep this as a `minor` bump. The previous behavior was inconvenient, but not harmful.

packages/better-auth/src/plugins/admin/routes.ts
```diff
@@ -1550,8 +1550,28 @@ export const setUserPassword = (opts: AdminOptions) =>
...
+			const user = await ctx.context.internalAdapter.findUserById(userId);
+			if (!user) {
+				throw APIError.from("NOT_FOUND", BASE_ERROR_CODES.USER_NOT_FOUND);
+			}
 			const hashedPassword = await ctx.context.password.hash(newPassword);
-			await ctx.context.internalAdapter.updatePassword(userId, hashedPassword);
+			const accounts = await ctx.context.internalAdapter.findAccounts(userId);
+			const credentialAccount = accounts.find(
+				(account) => account.providerId === "credential",
+			);
```
ping-maxwell, line 1728: What do you think of the idea of create a new internalAdapter function purely to fetch an account based on `providerId` - rather than fetching the entire list?

I know for the most part every user has a very small list of accounts but I feel like it's a more efficient pattern we can use

ping-maxwell (dismissed):

packages/better-auth/src/plugins/admin/routes.ts
bytaesu: > https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/db/internal-adapter.ts#L1015-L1032

Maybe we can already use this one `findAccountByProviderId`

Just from looking through it, I feel like we need to clean up our internal adapter a bit 🧐

ping-maxwell: makes sense.

R-udren:
Any progress?

R-udren:
I would like to complete this feature. Do I have to create new PR? @ping-maxwell

bytaesu:
Hi @R-udren, let me finalize this !

bytaesu:
@cubic-dev-ai review

Issue #6766, opened 2025-12-15
Update Drizzle ORM Adapter to support the new Drizzle ORM query syntax (drizzle-orm v1.0.0)


ghassenbenghorbal:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

I migrated to the new drizzle-orm@beta but now I can't use better-auth because the drizzle orm adapter does not support the new syntax

```javascript9 |    const createCustomAdapter = (db$1) => ({ getFieldName, options }) => {
10 |            function getSchema(model) {
11 |                    const schema = config.schema || db$1._.fullSchema;
12 |                    if (!schema) throw new BetterAuthError("Drizzle adapter failed to initialize. Schema not found. Please provide a schema object in the adapter options object.");
13 |                    const schemaModel = schema[model];
14 |                    if (!schemaModel) throw new BetterAuthError(`[# Drizzle Adapter]: The model "${model}" was not found in the schema object. Please pass the schema directly to the adapter options.`);
                                    ^
BetterAuthError: [# Drizzle Adapter]: The model "user" was not found in the schema object. Please pass the schema directly to the adapter options.
 cause: undefined,
```

### Describe the solution you'd like

Update Drizzle ORM Adapter to support drizzle-orm@beta (v1.0)

### Describe alternatives you've considered

None

CBell045:
I was having this issue as well. Passing my schema and relations to my drizzle adapter seems to have fixed it: 

```
import "dotenv/config";
import * as schema from "./schema";
import { relations } from "./relations";
import { drizzle } from "drizzle-orm/bun-sql";

export const db = drizzle({
  connection: process.env.DATABASE_URL!,
  casing: "snake_case",
  schema,
  relations,
});
```

matulef:
Even if you pass the schema, I think there is still a compatibility issue with the new drizzle 1.0 syntax. It occurs if you have joins turned on. The error I get is `Unknown relational filter field: "decoder"`. If I comment out experimental joins in my auth config, e.g. 
```
// experimental: { joins: true }
```
then it works because it doesn't rely on the drizzle-defined relations. But ideally, we'd be able to use the new drizzle and get the benefit of joins in better-auth too. 

I think this can be done in a backwards compatible way by replacing db.query with db._query if it's defined. See here: https://orm.drizzle.team/docs/relations-v1-v2

ping-maxwell:
[...]

Issue #5931, opened 2025-11-12
Testing new adapter - test if `count` is implemented


realtebo:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

implementing a new adapter I can sucesfully past tests having

```
count: async ({
          model,
          where,
        }: {
          model: string;
          where?: Required<Where>[];
        }): Promise<number> => {
          console.log("count model:", model);
          console.log("count where:", where);
          // ...

          return -1;
        },
```

### Describe the solution you'd like

 add a new test 

### Describe alternatives you've considered

remove need for count if un needed

realtebo:
I am manually running 

```
 npx vitest run ./better-auth-jwt-adapter.test.ts --bail 1
```

where the file contains, as per your doc

```
import { describe, afterAll } from "vitest";
import { runAdapterTest } from "better-auth/adapters/test";
import { BetterAuthJwtAdapter } from "./better-auth-jwt-adapter";

describe("My Adapter Tests", async () => {
  afterAll(async () => {
    // Run DB cleanup here...
  });
  const adapter = BetterAuthJwtAdapter({
    debugLogs: {
      // If your adapter config allows passing in debug logs, then pass this here.
      isRunningAdapterTests: true, // This is our super secret flag to let us know to only log debug logs if a test fails.
    },
  });
  await runAdapterTest({
    getAdapter: async (betterAuthOptions = {}) => {
      return adapter(betterAuthOptions);
    },
  });
});

```

@better-auth-agent all test passed even if my `count` function is not implemented

```
 Test Files  1 passed (1)
      Tests  25 passed | 2 skipped (27)
   Start at  16:38:23
   Duration  206ms (transform 39ms, setup 0ms, collect 94ms, tests 20ms, environment 0ms, prepare 2ms)
```

This is due the fact, dear my @better-auth-agent , that no test is currently implemented for the `count` method of an adapter

realtebo:
@better-auth-agent

Please, add a custom test case for the count method that verifies its expected behavior (for example, returning the correct number based on provided filters). This way, any future changes to count will be caught.

Add this to your project, not to mine

realtebo:
@better-auth-agent

 add label 'test'

GautamBytes:
working on it !!

GautamBytes:
Hey, thanks for reporting this! 

[...]

PR #6913, opened 2025-12-20
feat: support drizzle relations v2

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/6766, https://github.com/orgs/better-auth/projects/6/views/2?pane=issue&itemId=168869122&issue=better-auth%7Cbetter-auth%7C5931

### Notes
This PR introduces a new package: `drizzle-adapter` as `@better-auth/drizzle-adapter` which will need to be published on npm.
This PR also introduces `better-auth/adapters/test-adapter` as an export to the better-auth adapter test utility and all of it's tests. The reason I didn't export those functions right under `better-auth/adapter` was to avoid issues related to those functions needing the `vitest` import which could be bundled into other code if exported purely under `/adapter`.

---
## Summary by cubic
Adds first-class support for Drizzle Relations v2 via a new @better-auth/drizzle-adapter/relations-v2 adapter and updates the CLI to generate v2 schema and relations. Enables join queries using Drizzle’s v2 relations without changing your database.

- New Features
  - New relations-v2 adapter for pg, mysql, and sqlite with optional transactions, camelCase, pluralization, and debug logs.
  - Adapter-powered schema generation (createSchema) that outputs both schema and defineRelations; the CLI now prefers the adapter’s generator when available and falls back to built-ins.

- Migration
  - Install: npm install @better-auth/drizzle-adapter
  - Update: import { drizzleAdapter } from "@better-auth/drizzle-adapter/relations-v2" and set provider ("sqlite" | "pg" | "mysql").
  - Generate: npx @better-auth/cli generate
  - Drizzle init: pass both schema and relations from the generated file to drizzle()

<sup>Written for commit 627eccb7cfe89e4cd8172c5401513d4fb6a5efbb. Summary will update on new commits.</sup>

PR #9489, opened 2026-05-07
feat(drizzle-adapter): support Drizzle Relations v2

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/6766, https://github.com/orgs/better-auth/projects/6/views/2?pane=issue&itemId=168869122&issue=better-auth%7Cbetter-auth%7C5931

supersedes https://github.com/better-auth/better-auth/pull/6913


luxaritas:
Would it be a good idea to bump the version to rc?

makl11:
Hi, first of all, thanks for your work!

I looked over the related issue and tested the changes in my project.

The generated schema still uses `defineRelations` instead of `defineRelationsPart`. I did a simple find-and-replace of `defineRelations` to `defineRelationsPart` locally without issues.

Also, I think the many-to-one relations should use the non-pluralized name, similar to how one-to-one relations are handled, i.e:
```typescript
sessions: {
  user: r.one.users({
    from: r.sessions.userId,
    to: r.users.id,
  }),
},
```
instead of
```typescript
sessions: {
  users: r.one.users({
    from: r.sessions.userId,
    to: r.users.id,
  }),
},
```

packages/drizzle-adapter/src/relations-v2/generate-drizzle-schema.ts
```diff
@@ -0,0 +1,651 @@
...
+function generateImport({
+	databaseType,
+	tables,
+	options,
+}: {
+	databaseType: "sqlite" | "mysql" | "pg";
+	tables: BetterAuthDBSchema;
+	options: BetterAuthOptions;
+}) {
+	const rootImports: string[] = ["defineRelations"];
```
makl11:
```suggestion
	const rootImports: string[] = ["defineRelationsPart"];
```

packages/drizzle-adapter/src/relations-v2/generate-drizzle-schema.ts
```diff
@@ -0,0 +1,651 @@
...
+
+			if (relationDefs.length > 0) {
+				relationsEntries.push(
+					`  ${modelName}: {\n${relationDefs.join(",\n")}\n  }`,
+				);
+			}
+		}
+
+		if (relationsEntries.length > 0) {
+			relationsString = `\n\nexport const relations = defineRelations(${schemaObject}, (r) => ({\n${relationsEntries.join(",\n")}\n}));\n`;
```
makl11:
```suggestion
			relationsString = `\n\nexport const relations = defineRelationsPart(${schemaObject}, (r) => ({\n${relationsEntries.join(",\n")}\n}));\n`;
```

ping-maxwell:
Thanks for your suggestion @makl11 I've switched to `defineRelationsPart` now.

KhalidAlansary:
> closes https://github.com/better-auth/better-auth/issues/6766, https://github.com/orgs/better-auth/projects/6/views/2?pane=issue&itemId=168869122&issue=better-auth%7Cbetter-auth%7C5931

merging this pr won't close #5931 . If this closes it, use "closes #6766, #closes #5931" in the pr description

D4ario0:
Is this planned to be included for the next beta release or the next major update?

crazen3:
> Is this planned to be included for the next beta release or the next major update?

@D4ario0 based on https://github.com/better-auth/better-auth/issues/6766#issuecomment-4464154668, maybe in 1.7.1+ or 1.8? Definitely not a major update though, since this isn't even a breaking change

typedrat:
The `deleteMany` implementation doesn't correctly work with Drizzle when using postgres-js as the underlying library. Every successful DELETE without RETURNING reports 0 rows affected, which causes any caller that depends on the count — most notably internalAdapter.consumeOne in better-auth/dist/db/internal-adapter.mjs — to return null despite the row having been deleted.

```ts
async deleteMany({ model, where }) {
  const schemaModel = getSchema(model);
  const clause = convertWhereClause(where, model);
  const res = await db.delete(schemaModel).where(...clause);
  let count = 0;
  if (res && "rowCount" in res) count = res.rowCount;          // node-postgres
  else if (Array.isArray(res)) count = res.length;             // matches postgres-js, but for the wrong reason
  else if (res && ("affectedRows" in res || "rowsAffected" in res || "changes" in res)) count = res.affectedRows ?? res.rowsAffected ?? res.changes;
  ...
  return count;
}
```

The problem is the second branch. postgres-js's Result class extends Array, so `Array.isArray` is true. For DELETE / UPDATE / INSERT without `RETURNING`, the result has `length === 0` and the actual affected-row count lives in `res.count`. The `Array.isArray` branch wins, `count = res.length = 0`, and `consumeOne` reports the operation as having affected 0 rows even when it deleted a row.

For us this surfaced as broken OAuth code exchange (every authorization-code grant returned `invalid_grant`), but anything else that goes through the core `consumeOne` fallback on postgres-js is similarly affected — for example, any in-band verification-token consumption (email verification, password reset, OTPs, magic links) where you depend on the returned row, not just its absence.

typedrat:
After patching the previous bug, I discovered that the relations-v2 `convertWhereClause` doesn't map `{ value: null }` to `isNull(field)` — it falls through to `eq(field, w.value)` → `field = NULL`, which is always NULL (and therefore false-y) in SQL and never matches any rows. The legacy adapter handles this correctly, but the relations-v2 rewrite dropped the branch in all three sub-clauses (single, AND, and OR) and never emits `isNull`/`isNotNull` at all. 

Surfaces as `invalid_grant` on every refresh-token rotation, but also breaks `/oauth2/revoke` and the device-authorization plugin's "link pending code to user" step — anywhere better-auth uses `value: null` to mean "WHERE field IS NULL".

remorses:
Found a bug in the drizzle adapter's `convertWhereClause`: `eq(column, null)` generates `column = NULL` in SQL, which is never true (SQL null semantics). The correct form is `column IS NULL`. Same applies to `ne(column, null)` vs `IS NOT NULL`.

This breaks the device authorization plugin because the claim step in `deviceVerify` uses `{ field: "userId", operator: "eq", value: null }` in a WHERE clause. The UPDATE matches zero rows silently, so device codes are never claimed and approve always fails with "Device code has not been claimed by a verifying session."

Fix: https://github.com/remorses/better-auth-drizzle-adapter/commit/b8a6413

marr:
@ping-maxwell I made a PR to your branch here that should fix the linting issue. Is this work still in progress? At a cursory glance, it seems there are some conditionals creeping in that may indicate the agents are getting confused.

packages/drizzle-adapter/src/relations-v2/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+	sql,
+} from "drizzle-orm";
+import {
+	insensitiveEq,
+	insensitiveIlike,
+	insensitiveInArray,
+	insensitiveNe,
+	insensitiveNotInArray,
+} from "../query-builders";
+import { generateDrizzleSchema } from "./generate-drizzle-schema";
```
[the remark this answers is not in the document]
ping-maxwell: Fixed in 019c4424a: `generateDrizzleSchema` is now lazy-loaded via dynamic import inside `createSchema`, so Node-only modules are only evaluated when schema generation runs.

e2e/adapter/test/drizzle-v2-relations-adapter/generate-schema.ts
```diff
@@ -0,0 +1,102 @@
...
+	const thisCount = generationCount;
+	const i = async (x: string) => {
+		// Clear the Node.js module cache for the generated schema file to ensure fresh import
+		try {
+			const resolvedPath =
+				require?.resolve?.(x) ||
+				(import.meta && new URL(x, import.meta.url).pathname);
+			if (resolvedPath && typeof resolvedPath === "string" && require?.cache) {
+				delete require.cache[resolvedPath];
+			}
```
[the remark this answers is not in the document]
ping-maxwell: Fixed in 019c4424a: replaced bare `require` references with `typeof require !== "undefined"` guards before calling `require.resolve` / accessing `require.cache`.

e2e/adapter/test/drizzle-v2-relations-adapter/generate-schema.ts
```diff
@@ -0,0 +1,102 @@
...
+	});
+
+	await fs.writeFile(
+		join(
+			import.meta.dirname,
+			`/.tmp/generated-${dialect}-schema-${thisCount}.ts`,
+		),
+		code || "",
+		"utf-8",
+	);
```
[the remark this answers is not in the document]
ping-maxwell: Fixed in 019c4424a: changed `join(import.meta.dirname, "/.tmp")` to `join(import.meta.dirname, ".tmp")` so paths stay relative to the test directory.

e2e/adapter/test/prisma-adapter/generate-prisma-schema.ts
```diff
@@ -12,7 +12,18 @@ export async function generatePrismaSchema(
...
-	const i = async (x: string) => await import(x);
+	const i = async (x: string) => {
+		// Clear the Node.js module cache for the generated schema file to ensure fresh import
+		try {
+			const resolvedPath =
+				require?.resolve?.(x) ||
+				(import.meta && new URL(x, import.meta.url).pathname);
+			if (resolvedPath && typeof resolvedPath === "string" && require?.cache) {
+				delete require.cache[resolvedPath];
+			}
```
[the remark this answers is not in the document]
ping-maxwell: Fixed in 019c4424a: same `typeof require !== "undefined"` guard applied to the prisma e2e helper.

packages/drizzle-adapter/src/relations-v2/generate-drizzle-schema.ts
```diff
@@ -0,0 +1,667 @@
...
+					mysql: field.bigint
+						? `bigint('${name}', { mode: 'number' })`
+						: `int('${name}')`,
+				},
+				date: {
+					sqlite: `integer('${name}', { mode: 'timestamp_ms' })`,
+					pg: `timestamp('${name}')`,
+					mysql: `timestamp('${name}', { fsp: 3 })`,
+				},
+				"number[]": {
```
gustavovalverde, line 164: This MySQL branch is still generating invalid Drizzle RC3 code for array and JSON fields. `text(name, { mode: "json" })` is not a MySQL text-column option in `drizzle-orm@1.0.0-rc.3`, and `json(name, { mode: "json" })` is not a valid MySQL JSON builder signature either. That means a MySQL project with `additionalFields` using `string[]`, `number[]`, or `json` will generate schema code that fails even though the default MySQL e2e suite passes.

The fix should make the generated MySQL columns match Drizzle's real RC3 API, likely `json(name)` for arrays/JSON, and align the adapter flags at the same time. Right now `supportsArrays` was set to true to avoid double stringifying arrays, but `supportsJSON` remains false; Drizzle's MySQL `json()` already stringifies driver values, so Better Auth will pre-stringify JSON and Drizzle will stringify the string again.

gustavovalverde (dismissed):

.changeset/four-turkeys-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"@better-auth/drizzle-adapter": minor
+---
+
+feat: support drizzle relations v2
```
bytaesu: As I mentioned before, could you rewrite the changeset from a user-facing perspective?

https://github.com/better-auth/better-auth/blob/main/CONTRIBUTING.md#changesets

docs/content/docs/adapters/drizzle.mdx
```diff
@@ -174,6 +174,72 @@ export const auth = betterAuth({
...
+		schema,
+	}),
+	//... the rest of your config
+});
+```
+
+Then regenerate your schema using the Better Auth CLI:
+
+```package-install
+npx @better-auth/cli@latest generate
```
dotnize:
```suggestion
npx auth@latest generate
```

bytaesu: 👍

bytaesu:
@greptileai review

bytaesu:
-> https://github.com/better-auth/better-auth/pull/9489/commits/9261d7c873021cfdc85b7c70b5db738b9c8d3c45

I will follow-up on the `main` branch as well.

packages/drizzle-adapter/src/relations-v2/index.ts
```diff
@@ -0,0 +1,1092 @@
...
+									select?.length && select.length > 0
+										? select.reduce(
+												(acc, field) => {
+													acc[getFieldName({ model, field })] = true;
+													return acc;
+												},
+												{} as Record<string, boolean>,
+											)
+										: undefined,
+								limit: limit ?? 100,
```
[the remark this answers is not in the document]
bytaesu: @greptileai The adapter never receives an undefined limit. The core factory injects unsafeLimit ?? defaultFindManyLimit ?? 100 before every findMany, so both the join and non-join paths cap identically. The ?? 100 here is just a redundant fallback, not a silent divergence.

bytaesu:
https://github.com/better-auth/better-auth/pull/9489/commits/7e9469c9b5b11978bac9857868995d45402751e4 is related to #10048

bytaesu:
-> https://github.com/better-auth/better-auth/pull/9489/commits/4be62ca27b1fb2c936027ef0db8a219ba1b4c7a2

I will follow-up on the `main` branch as well.

bytaesu (approved):
> [!NOTE]
> We're planning to release `v1.7-rc` soon. Since the public API for this feature is unlikely to change, we are going to merge Max's great work first and continue polishing the internal implementation afterward 😁

PR #9492, opened 2026-05-08
chore: drop changeset for cli refactor (#9477)

bytaesu:
Follow-up from #9477

The behavior described by that changeset was already possible before, so at this point the PR seems to be purely a refactor.


ping-maxwell (approved):

pi0:
before exporting both named and default was broken

Issue #9488, opened 2026-05-07
Error `email_doesn't_match` seems out of place


itschip:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

Seeing all other errors following a strict error_something_something, it's odd to suddenly see a ' in the error, which honestly makes it a tad more annoying to have in the URL :P

Open for a PR to change it to just `email_does_not_match`?

### Describe the solution you'd like

Change error to `email_does_not_match`.

ping-maxwell:
This is already fixed in our `next` branch, it will arrive next minor.

PR #9499, opened 2026-05-08
fix: rename `email_doesn't_match` error to `email_does_not_match`

ping-maxwell:
The error code email_doesn't_match contains an apostrophe which is inconsistent with other error codes that follow a strict snake_case format (e.g., unable_to_link_account, account_already_linked_to_different_user).

https://github.com/better-auth/better-auth/issues/9488

This is a potential breaking change to users who's logic relies on `email_doesn't_match` error codes, so I've placed this in `minor`.


ping-maxwell:
Turns out it's already implemented.

PR #9501, opened 2026-05-08
chore(changeset): restore changeset for cli named+default fix (#9477)

bytaesu:
Restoring the changeset I dropped in #9492. I had misread it as an internal refactor. The original wording listed shapes that were already supported, which threw me off. The real fix was that `export const auth = ...; export default auth;` configs now work under auto-detect (previously failed with no `--config`). Bumped to `minor` to match the sibling PR #9431.

- Also relate to https://github.com/better-auth/better-auth/pull/9500

@pi0 looks good? 😁

### DIFF
--- a/.changeset/cli-named-default-auto-detect.md
+++ b/.changeset/cli-named-default-auto-detect.md
@@ -1,0 +1,5 @@
+---
+"auth": minor
+---
+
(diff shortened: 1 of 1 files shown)


pi0:
Look cool thanks! Sorry my initial changeset wasn’t clear enough.

PR #9505, opened 2026-05-08
fix(api-key): return 429 instead of 401 when API key is rate limited

### LINKED ISSUE #9504: API key rate limiting returns 401 instead of 429
## Bug

When an API key exceeds its rate limit, the server responds with **401 Unauthorized** instead of **429 Too Many Requests**.

### Root Cause

In [`packages/api-key/src/routes/verify-api-key.ts`](https://github.com/better-auth/better-auth/blob/main/packages/api-key/src/routes/verify-api-key.ts), the rate limit check at line ~160 throws:

```typescript
    message: message ?? undefined,
    code: "RATE_LIMITED" as const,
    details: { tryAgainIn },
});
```

The error type `"UNAUTHORIZED"` maps to HTTP 401. It should be `"TOO_MANY_REQUESTS"` (HTTP 429), consistent with the usage-exceeded checks on lines 132 and 151 which correctly use `"TOO_MANY_REQUESTS"`.

### Impact

- Clients and API gateways cannot distinguish between an authentication failure and a rate limit violation based on status code alone.
- Retry logic based on 429 status codes will not trigger.
- The `error.code` field is `"RATE_LIMITED"` (correct), but the HTTP status contradicts it.

### Steps to Reproduce

1. Configure the API key plugin with a low rate limit (e.g., `maxRequests: 2`)
2. Create an API key
3. Make 3+ requests using the API key via the `x-api-key` header
4. Observe the 3rd request returns **HTTP 401** instead of **HTTP 429**

### Expected Behavior

Rate-limited requests should return **HTTP 429 Too Many Requests**.

### Suggested Fix

Change `"UNAUTHORIZED"` to `"TOO_MANY_REQUESTS"` in the rate limit check in `validateApiKey()`.

I have a PR ready with the fix and a regression test.

Kvizas:
## Summary

- Changed `APIError("UNAUTHORIZED")` to `APIError("TOO_MANY_REQUESTS")` in the per-key rate limit check in `validateApiKey()`, so rate-limited API key requests return **HTTP 429** instead of **HTTP 401**.
- Added a regression test that verifies the 429 status code when an API key exceeds its rate limit via the `before` hook (using `x-api-key` header + `getSession`).

## Root Cause

In `packages/api-key/src/routes/verify-api-key.ts`, the rate limit check threw `APIError("UNAUTHORIZED")` (HTTP 401) while the usage-exceeded checks in the same function correctly used `"TOO_MANY_REQUESTS"` (HTTP 429). The rate limit path was missed.

Closes #9504

## Test plan

- [x] Added regression test: `should return 429 when API key rate limit is exceeded via before hook`
- [x] `pnpm typecheck` passes
- [x] Existing rate limit tests still pass (requires DB setup)

---
## Summary by cubic
Return HTTP 429 instead of 401 when an API key is rate limited, fixing incorrect status codes in the per-key rate limit path. This makes throttling distinguishable from auth failures.

- **Bug Fixes**
  - In `validateApiKey()`, throw `APIError("TOO_MANY_REQUESTS")` for per-key rate limits (was `UNAUTHORIZED`).
  - Regression test: uses `x-api-key` with `getSession`, allows two requests, asserts 429 on the third; fixed the test’s `@see` reference.
  - Added changeset to release a patch for `@better-auth/api-key`.

<sup>Written for commit 7ce69f6546c55d0f0f58b08f821f3f517b0c1409. Summary will update on new commits.</sup>


bytaesu (approved):
@Kvizas Thanks for working on this 😁

PR #9506, opened 2026-05-08
feat: introduce `registerBeforeSend` and remove must-be-last constraint on cookie plugins

bytaesu:
_wip_


gustavovalverde:
Closing in favor of #9537 and #9538. #9537 implements a native `try...finally` block calling cookie hooks unconditionally; #9538 then migrates all four cookie integrations (`nextCookies`, `sveltekitCookies`, `tanstackStartCookies` React and Solid) to `hooks.finally` and deletes `cookie-plugin-guard.ts` entirely, making plugin order irrelevant. The `registerBeforeSend` approach here is order-dependent.

Issue #7896, opened 2026-02-10
Drizzle adapter mssql support


JohnFScha:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

I'm currently working on a project that heavily relies on SQL Server (the infra is built entirely on Microsoft's vendor lock-in).

Not having the chance to use drizzle in my node projects creates a serious overhead to roll out my own auth.

### Describe the solution you'd like

Drizzle ORM has realesed a beta 1.0.0 with long awaited SQL Server support.

It would be nice to have this translated to the adapter, so as to be able to deploy this architecture from windows server + IIS 

### Describe alternatives you've considered

I can use sqlite on this windows server instance, but I loose a lot of native types that come in handy in diverse scenarios.

ping-maxwell:
hmm? this should work out of the box.

Are you running into some error?

JohnFScha:
It doesn't, instantiating the better-auth client on the backend gives a type error on the provider:

```ts
export const auth = betterAuth({
  database: drizzleAdapter(db, {
    provider: "mssql", // Type '"msqql"' is not assignable to type '"pg" | "mysql" | "sqlite"'
  }),
 ...
});
```

chriscors:
Any updates on this? Have the same blocker. I can use the Kysely adapter as a workaround for auth, but would be best to have all drizzle

chriscors:
Took a swing at this — opened #9510 with `provider: "mssql"` support in `@better-auth/drizzle-adapter`.

A few notes on the approach for anyone watching the issue:

- The PR widens the `drizzle-orm` peer dep to `^0.45.2 || ^1.0.0-rc` so existing users aren't disrupted. MSSQL itself requires `drizzle-orm@^1.0.0-rc` (where `mssql-core` and `node-mssql` live), and you opt in by passing `provider: "mssql"` and bringing your own v1 install.
- Drizzle's MSSQL builder uses `.output()` rather than `.returning()`, and `.output()` has to be chained *before* `.values()` on inserts. To avoid restructuring the adapter's builder construction, MSSQL reuses the existing MySQL "execute, then re-select the affected row" path — same approach the Kysely MSSQL adapter already uses in this repo.
[...]

PR #9510, opened 2026-05-08
feat(drizzle-adapter): add MSSQL provider support

chriscors:
## Summary

Adds `provider: "mssql"` to `@better-auth/drizzle-adapter`, addressing #7896. Scoped to the adapter package; CLI schema generator and an e2e MSSQL test will follow in a separate PR.

## What changed

- **Provider union** (`src/drizzle-adapter.ts`, `src/query-builders.ts`): `"mssql"` is now accepted alongside `"pg" | "mysql" | "sqlite"`.
- **`withReturning`**: MSSQL falls into the same path as MySQL — execute the write, then re-select the affected row. Drizzle's MSSQL builder uses `.output()` (not `.returning()`) and requires `.output()` chained *before* `.values()` on inserts, so reusing the MySQL fallback keeps the existing builder shape working without a structural rewrite. This mirrors how Better Auth's Kysely MSSQL adapter handles it.
- **Numeric-id fallback**: `LAST_INSERT_ID()` becomes `SCOPE_IDENTITY()` when `provider === "mssql"`.
- **Peer dep**: widened to `"^0.45.2 || ^1.0.0-rc"` so users on drizzle v1 (where `mssql-core` / `node-mssql` live) install without peer warnings. Existing 0.45 users are unaffected — the adapter doesn't import any v1-only symbols.
- **Docs**: `docs/adapters/drizzle.mdx` and `installation.mdx` mention `"mssql"` in provider examples.
- **Changeset**: minor bump for `@better-auth/drizzle-adapter`.

## Why a re-select instead of OUTPUT

Drizzle's MSSQL `.output()` is on the *insert builder* (before `.values()`), not on the resulting promise:

```ts
db.insert(table).output().values({...})  // returns rows
db.insert(table).values({...})           // returns metadata only
```

The adapter currently constructs `db.insert(table).values(values)` and then asks `withReturning` to extract the row. Restructuring the `create()` path to thread provider awareness into builder construction would be a larger change for marginal benefit; the re-select approach is consistent with how MySQL is already handled and is what the Kysely MSSQL adapter does in this repo.

## Out of scope (follow-up PR)

- `packages/cli/src/generators/drizzle.ts`: emit Drizzle MSSQL schemas from `npx better-auth generate`. Until then, MSSQL users must hand-write their `mssqlTable` schema and pass it to the adapter — which is fully supported by this PR.
- `e2e/adapter/test/drizzle-adapter/adapter.drizzle.mssql.test.ts`: depends on the CLI generator above.

## Test plan

- [x] `pnpm --filter @better-auth/drizzle-adapter typecheck`
- [x] `pnpm --filter @better-auth/drizzle-adapter lint:package`
- [x] `pnpm --filter @better-auth/drizzle-adapter test --run` — 7/7 pass (added 2 mssql-specific tests)
- [x] Drizzle SQLite e2e regression check: 441/441 pass
- [ ] Drizzle Postgres / MySQL e2e regression check (recommend running in CI)
- [ ] Manual smoke against a real MSSQL instance (will be covered by the follow-up e2e test)

---
## Summary by cubic
Adds MSSQL support to `@better-auth/drizzle-adapter` for Microsoft SQL Server. Uses execute-then-reselect instead of `.returning()`/MSSQL `.output()`; CLI schema generation for MSSQL is not yet supported.

- **Refactors**
  - Switched MSSQL selects to `TOP`/`OFFSET-FETCH` with a default `ORDER BY id` when needed, added small helpers for single-row reselects, and removed an unreachable last-insert-id fallback.

- **Dependencies**
  - Widened `drizzle-orm` peer to `^0.45.2 || ^1.0.0-rc` and updated the lockfile to match.

<sup>Written for commit 0813ff9ac7b67e072d957d1f75d8bad92b903f25. Summary will update on new commits.</sup>


chriscors:
Hit a small bug on the MSSQL provider path while smoke-testing this PR downstream — `deleteMany` returns the wrong shape and trips the "not a number" guard.

### Repro

Running the adapter with `provider: "mssql"` (drizzle-orm `1.0.0-rc.2` + the `mssql` driver) against SQL Server 2019. Better-auth's internal verification-token cleanup triggers it on every sign-in attempt:

```
ERROR [Better Auth]: [Drizzle Adapter] The result of the deleteMany operation is not a number. This is likely a bug in the adapter. Please report this issue to the Better Auth team. {
  res: { recordsets: [], output: {}, rowsAffected: [ 4 ] },
  model: 'verification',
  where: [
    { operator: 'lt', connector: 'AND', field: 'expiresAt', value: '2026-05-11T20:24:49.464Z', mode: 'sensitive' }
  ]
}
```

The DELETE itself executed correctly (4 rows removed). It's only the count-unwrap that fails, so it's noisy-but-non-blocking — `deleteMany` returns the unwrapped array (`[4]`) instead of a number to the caller.

### Root cause

In `packages/drizzle-adapter/src/drizzle-adapter.ts`, the existing unwrap (line 869) handles pg/mysql/sqlite shapes:

```ts
if (res && "rowCount" in res) count = res.rowCount;
else if (Array.isArray(res)) count = res.length;
else if (res && ("affectedRows" in res || "rowsAffected" in res || "changes" in res))
  count = res.affectedRows ?? res.rowsAffected ?? res.changes;
```

The `mssql` driver returns `IResult<T>` where **`rowsAffected` is `number[]`** (one entry per statement in a batch), not a number. The `??` chain happily binds `count` to the array, then the downstream `typeof count !== "number"` check fires the warning, and the return value to better-auth is `[4]` instead of `4`.

### Suggested fix

Treat `rowsAffected` as a possibly-array and sum:

```ts
else if (res && ("affectedRows" in res || "rowsAffected" in res || "changes" in res)) {
  const raw = res.affectedRows ?? res.rowsAffected ?? res.changes;
  count = Array.isArray(raw)
    ? raw.reduce((sum, n) => sum + n, 0)  // MSSQL: number[] per-statement counts
    : raw;
}
```

This is the same class of "MSSQL returns a different shape" issue as the `.top()` vs `.limit()` change earlier in this PR — same one-line guard pattern works.

gustavovalverde:
Closing in favor of #9512. #9512 stacks on this PR with the same adapter changes (`drizzle-adapter.ts`, `deleteMany` `rowsAffected` fix, `TOP`/`OFFSET-FETCH` pagination) plus the CLI generator and snapshot tests.

PR #9519, opened 2026-05-09
fix(email): enforce single-use for change-email verification tokens

### LINKED ISSUE #9479: Email verification token can be reused to repeatedly send verification email

### Reproduction

Steps to reproduce:
1. Initiate email change with **sendChangeEmailConfirmation** configured.
2. Receive verification link
3. Reuse the same link multiple times (refresh or re-request).
4. Each request continues to return success and sending emails again on new email.

### Current behavior
The same token can be reused multiple times, and each request returns { success: true } and may trigger side effects (e.g. sending emails again). There is no mechanism to mark the token as used or prevent repeated usage until expiration.
### Expected behavior
The verification token should be single-use. After the first successful verification request, the token should be invalidated or marked as used. Any subsequent requests with the same token should be rejected with an appropriate error (e.g. token already used) and should not trigger any side effects such as sending emails again.

1.5.6

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Pro",
    "release": "10.0.26200",
    "cpuCount": 12,
    "cpuModel": "AMD Ryzen 5 5600H with Radeon Graphics         ",
    "totalMemory": "13.86 GB",
    "freeMemory": "4.57 GB"
  },
  "node": {
    "version": "v20.19.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.4"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "^16.2.1"
    },
    {
      "name": "react",
      "version": "19.2.3"
    }
  ],
  "databases": [
    {
      "name": "mysql2",
      "version": "^3.16.2"
    },
    {
      "name": "drizzle",
      "version": "^0.45.1"
    }
  ],
  "betterAuth": {
    "version": "Unknown",
    "config": null,
  }
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

Vishesh-Verma-07:
fix(email): enforce single-use for change-email verification tokens
Problem
Closes #9479
Email change verification tokens could be reused indefinitely until expiry. Clicking the same confirmation or verification link multiple times would return { success: true } each time and re-trigger side effects like sending emails again.
Root Cause
Email change tokens are stateless JWTs — they're validated cryptographically against the server secret, but no record of them exists in the database. This means there's nothing to delete after first use, unlike password reset tokens which are stored in the verification table and deleted immediately after consumption with deleteVerificationByIdentifier.
Fix
After a change-email-confirmation or change-email-verification token is consumed for the first time, a tombstone record is written to the existing verification table using a SHA-256 hash of the raw JWT string as the identifier. On any subsequent request with the same token, the tombstone is found and the request is rejected with TOKEN_ALREADY_USED before any side effects occur.
Key design decisions:

The tombstone check runs before findUserByEmail — so reuse is caught even after the email has already been updated in the DB (which would otherwise produce a misleading USER_NOT_FOUND error)
The tombstone is written before sending emails or updating the user record, so concurrent requests with the same token are also blocked
The tombstone's expiresAt matches the token's own expiresIn config, so no manual cleanup is needed — it self-expires alongside the token
No schema changes — reuses the existing verification table present in every Better Auth deployment

Changes

packages/core/src/error/codes.ts — Added TOKEN_ALREADY_USED error code
packages/better-auth/src/api/routes/email-verification.ts — Added getTokenIdentifier() helper (SHA-256 → base64url) and single-use enforcement logic in verifyEmail
packages/better-auth/src/api/routes/email-verification.test.ts — Added regression tests covering reuse of confirmation tokens, reuse of verification tokens, and absence of side-effect emails on reuse

Testing
Three new regression tests added under Email Change Token Single-Use Enforcement:

Reusing a change-email-confirmation token is rejected with TOKEN_ALREADY_USED
Reusing a change-email-verification token is rejected with TOKEN_ALREADY_USED
No additional emails are sent when a token is reused

### DIFF
--- a/.changeset/curly-meals-wonder.md
+++ b/.changeset/curly-meals-wonder.md
@@ -1,0 +1,6 @@
+---
+"better-auth": major
+"@better-auth/core": major
+---
+
+Email change confirmation and verification tokens are now single-use. Replaying the same verification link is rejected with TOKEN_ALREADY_USED instead of silently succeeding and re-sending emails.
--- a/.changeset/curly-meals-wonder.md
+++ b/.changeset/curly-meals-wonder.md
@@ -1,6 +1,6 @@
 ---
-"better-auth": major
-"@better-auth/core": major
+"better-auth": patch
+"@better-auth/core": patch
 ---
 
 Email change confirmation and verification tokens are now single-use. Replaying the same verification link is rejected with TOKEN_ALREADY_USED instead of silently succeeding and re-sending emails.
--- a/packages/better-auth/src/api/routes/email-verification.test.ts
+++ b/packages/better-auth/src/api/routes/email-verification.test.ts
@@ -684,3 +684,198 @@ describe("Email Verification Secondary Storage", async () => {
 		expect(result.status).toBe(true);
 	});
 });
+
+describe("Email Change Token Single-Use Enforcement", async () => {
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/9479
+	 */
+
+	it("should reject a reused change-email-confirmation token", async () => {
+		let capturedToken: string;
+		const { client, auth, signInWithTestUser, cookieSetter, db, testUser } =
+			await getTestInstance({
+				emailAndPassword: { enabled: true },
+				emailVerification: {
+					async sendVerificationEmail({ token: t }) {
+						capturedToken = t;
+					},
+				},
+				user: {
+					changeEmail: {
+						enabled: true,
+						async sendChangeEmailConfirmation(data) {
+							capturedToken = data.token;
+						},
+					},
+				},
+			});
+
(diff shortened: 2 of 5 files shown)


Vishesh-Verma-07:
Hi @bytaesu / @ping-maxwell — the auto-retarget bot moved this to next because my changeset incorrectly said major. I've now fixed it to patch (this is a bug fix, no breaking changes). Could you please retarget this PR back to main? Thanks!

ping-maxwell (changes requested):
We should probably use a "store-and-delete" approach instead of a tombstone method.
It follows existing patterns of password reset and other flows.

Vishesh-Verma-07:
@ping-maxwell  i have the made the changes in the approach have a look to it

PR #9522, opened 2026-05-09
fix(api-key): return 429 when API key is rate limited

### LINKED ISSUE #9504: API key rate limiting returns 401 instead of 429

MuzzaiyyanHussain:
## Summary

Fixes an incorrect HTTP status code returned when an API key exceeds its rate limit.

Previously, rate-limited API key requests returned `401 Unauthorized` even though the API key was valid. This changes the response to return `429 Too Many Requests` instead.

## Changes

* Replaced `UNAUTHORIZED` with `TOO_MANY_REQUESTS` in the API key rate-limit validation logic
* Added a regression test to verify rate-limited requests return the expected behavior

## Why

Returning `401` for rate limiting makes it difficult for clients and API gateways to distinguish between:

* invalid authentication
* temporary request throttling

Using `429` aligns with standard HTTP semantics and enables proper retry handling for clients.

## Impact

* Correct HTTP status code for rate-limited API key requests
* Improves compatibility with retry middleware and API gateways
* No breaking changes

Closes #9504 

---
## Summary by cubic
Return HTTP 429 for rate-limited API keys instead of 401 to improve client handling of throttling. This clarifies valid-key throttling and enables proper retry/backoff.

- **Bug Fixes**
  - Throw `APIError("TOO_MANY_REQUESTS")` with `code: "RATE_LIMITED"` when the rate limit is exceeded in `packages/api-key/src/routes/verify-api-key.ts`.
  - Added a regression test in `packages/api-key/src/api-key.test.ts` to assert valid responses before the threshold and `RATE_LIMITED` after.

<sup>Written for commit c04e8ae6c21d423f5a2c89a848e69f9954656d3d. Summary will update on new commits.</sup>


bytaesu:
Duplicate of #9505

PR #9547, opened 2026-05-10
feat(cli): add create-admin command

### LINKED ISSUE #9546: Create Initial Admin user through CLI

Currently, there is no way to create an initial admin in better-auth. 
To create an initial admin user, we have to write an seeding script or put the keys in the admin plugin. 

I would like to be able to create users through the cli. 

Currently, to create an admin via API, you need to be authenticated as an admin.

A CLI command would bypass authentication requirements entirely, allowing us to create the first admin before any user exists in the system.

`npx auth create-admin --email admin@example.com --password "secure-password" --role "admin"`

_No response_

### Additional context

_No response_

GautamBytes:
## Summary
- Add `auth create-admin` for bootstrapping an initial admin user
- Use the configured Better Auth instance and existing admin `createUser` API path
- Prompt before creating admins when users already exist
- Prefer masked password prompt in docs examples

Closes #9546

---
## Summary by cubic
Adds `auth create-admin` to bootstrap the first admin user through your configured `better-auth` instance. Uses the Admin plugin’s `createUser` API so passwords are hashed and hooks run, with improved errors and confirmation prompts.

- **New Features**
  - New `auth create-admin` command; requires a Better Auth config, a persistent database, and the `admin()` plugin.
  - Prompts for missing email/password; supports `--email`, `--password`, `--name`, `--role`, `--data`, `--no-email-verified`, `--force`, `-y/--yes` (`--y` deprecated), `--config`, and `--cwd`. Marks the admin email as verified by default.
  - Confirms when users already exist (`--force`/`--yes` skip). Registers the command in the CLI and updates docs/tests.

- **Bug Fixes**
  - Clearer errors when the working directory is invalid, config/database is missing, the `admin()` plugin is not enabled, the email is invalid, or checking existing users fails (includes the underlying error).

<sup>Written for commit 0dcd5099315f237de75dc5dcff7c4fc593db62e1. Summary will update on new commits.</sup>

### DIFF
--- a/packages/cli/src/commands/create-admin.ts
+++ b/packages/cli/src/commands/create-admin.ts
@@ -1,0 +1,237 @@
+import { existsSync } from "node:fs";
+import path from "node:path";
+import { APIError } from "@better-auth/core/error";
+import { betterAuth } from "better-auth";
+import chalk from "chalk";
+import { Command } from "commander";
+import prompts from "prompts";
+import * as z from "zod";
+import { getConfig } from "../utils/get-config";
+
+type CreateUserApi = (input: {
+	body: {
+		email: string;
+		password: string;
+		name: string;
+		role: string;
+		data?: Record<string, unknown> | undefined;
+	};
+}) => Promise<{
+	user?: {
+		id?: string | number | undefined;
+		email?: string | undefined;
+		role?: string | undefined;
+	};
+}>;
+
+function exitWithError(message: string): never {
+	console.error(chalk.red(`Error: ${message}`));
+	process.exit(1);
+	throw new Error(message);
+}
+
+function parseData(data: string | undefined) {
+	if (!data) return undefined;
+	try {
+		const parsed = JSON.parse(data);
+		if (!parsed || typeof parsed !== "object" || Array.isArray(parsed)) {
+			exitWithError("--data must be a JSON object.");
+		}
+		return parsed as Record<string, unknown>;
+	} catch (error) {
+		if (error instanceof SyntaxError) {
+			exitWithError("--data must be valid JSON.");
+		}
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/create-admin.ts
```diff
@@ -0,0 +1,234 @@
...
+		id?: string | number | undefined;
+		email?: string | undefined;
+		role?: string | undefined;
+	};
+}>;
+
+function exitWithError(message: string): never {
+	console.error(chalk.red(`Error: ${message}`));
+	process.exit(1);
+	throw new Error(message);
```
ping-maxwell: probably don't need the `new Error` if we're going to terminate it, it won't reach the error.
Unless you want to move that line above the process.exit.

packages/cli/src/commands/create-admin.ts
```diff
@@ -0,0 +1,234 @@
...
+		| CreateUserApi
+		| undefined;
+
+	if (typeof createUser !== "function") {
+		exitWithError(
+			"The admin plugin is required. Add `admin()` to your Better Auth plugins before running this command.",
+		);
+	}
+
+	if (!options.name) exitWithError("Admin name is required.");
```
ping-maxwell: the zod schema already sets a default name so I don't think this can be reached.

ping-maxwell: Just a nit, likely fine to keep anyway, won't make much difference.

ping-maxwell (changes requested):

GautamBytes:
@ping-maxwell done with the changes!

ping-maxwell (approved):

PR #9557, opened 2026-05-11
docs: document account storeStateStrategy

### LINKED ISSUE #9556: Docs don't have a description for `account.storeStateStrategy`

### Reproduction

The options documentation mentions `account.storeStateStrategy` (in [advanced](https://better-auth.com/docs/reference/options#advanced)) - but the [account section](https://better-auth.com/docs/reference/options#account) doesn't actually have a description of it.

I expected a description of `account.storeStateStrategy` in the [account section](https://better-auth.com/docs/reference/options#account).
I observed there wasn't one.

1.6.2

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Enterprise",
    "release": "10.0.26100",
    "cpuCount": 16,
    "cpuModel": "AMD Ryzen 7 7840U w/ Radeon  780M Graphics     ",
    "totalMemory": "31.28 GB",
    "freeMemory": "6.90 GB"
  },
  "node": {
    "version": "v24.15.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.11.1"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.2.4"
    },
    {
      "name": "react",
      "version": "19.2.3"
    },
    {
      "name": "express",
      "version": "^5.2.1"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "^1.6.2",
    "config": {
      "onAPIError": {
        "errorURL": "/auth/error"
      },
      "secondaryStorage": {},
      "session": {
        "cookieCache": {
          "enabled": false,
          "refreshCache": false
        }
      },
      "user": {
        "additionalFields": {
          "audience": {
            "type": "string",
            "required": false
          }
        }
      },
      "account": {
        "storeAccountCookie": false,
        "storeStateStrategy": "database"
      },
      "socialProviders": {
        "microsoft": {
          "clientSecret": "[REDACTED]",
          "tenantId": "organizations",
          "disableProfilePhoto": true
        }
      },
      "plugins": [
        {
          "name": "custom-session",
          "config": {
            "id": "custom-session",
            "version": "1.6.5",
            "hooks": {
              "after": [
                {}
              ]
            },
            "endpoints": {},
            "$Infer": {
              "Session": {}
            }
          }
        }
      ]
    }
  }
}
```

Documentation

```typescript
  account: {
    storeAccountCookie: false,
    storeStateStrategy: "database",
  },
});
```

### Additional context

_No response_

GautamBytes:
## What Changed

Added documentation for `account.storeStateStrategy` in the Options reference account section.

The option was already referenced under `advanced`, but the account section did not explain what it does or which values are supported.

Closes #9556

## Details

- Added `storeStateStrategy` to the `account` example.
- Documented `"cookie"` and `"database"` strategies.
- Clarified the default behavior:
  - `"database"` when a database is configured
  - `"cookie"` for database-less/stateless setups

---
## Summary by cubic
Documented the `account.storeStateStrategy` option in the Options reference and added it to the account example. Explains `cookie` (store state in an encrypted cookie, no DB write) vs `database` (store in verification storage, set and validate a signed state cookie), defaults (`database` with a DB, `cookie` otherwise), and how to use `secondaryStorage` by setting `storeStateStrategy` to `database` to persist state there instead of in the cookie.

<sup>Written for commit 12cdc1c306382b9d9300084f3c7d6d8803093672. Summary will update on new commits.</sup>

### DIFF
--- a/docs/content/docs/reference/options.mdx
+++ b/docs/content/docs/reference/options.mdx
@@ -474,7 +474,7 @@ export const auth = betterAuth({
 			userId: "user_id"
 		},
 		encryptOAuthTokens: true, // Encrypt OAuth tokens before storing them in the database
-		storeStateStrategy: "database", // Store OAuth state in the database instead of an encrypted cookie
+		storeStateStrategy: "database", // Store OAuth state payload in verification storage
 		storeAccountCookie: true, // Store account data after OAuth flow in a cookie (useful for database-less flows)
 		accountLinking: {
 			enabled: true,
@@ -502,7 +502,7 @@ will be updated on sign in with the latest data from the provider.
 Controls where OAuth state data is stored during the authentication flow.
 
 * `"cookie"`: Store the state payload in an encrypted cookie. This keeps OAuth state stateless and avoids a database write during the start of the flow.
-* `"database"`: Store the state payload in the verification storage/table and validate it with the signed state cookie returned from the OAuth callback.
+* `"database"`: Store the state payload in the verification storage/table, set a signed state cookie when starting the flow, and validate the stored state against the signed cookie sent by the browser on the OAuth callback request.
 
 Defaults to `"database"` when a database is configured and `"cookie"` for database-less/stateless setups.
 
--- a/docs/content/docs/reference/options.mdx
+++ b/docs/content/docs/reference/options.mdx
@@ -504,7 +504,7 @@ Controls where OAuth state data is stored during the authentication flow.
 * `"cookie"`: Store the state payload in an encrypted cookie. This keeps OAuth state stateless and avoids a database write during the start of the flow.
 * `"database"`: Store the state payload in the verification storage/table, set a signed state cookie when starting the flow, and validate the stored state against the signed cookie sent by the browser on the OAuth callback request.
 
-Defaults to `"database"` when a database is configured and `"cookie"` for database-less/stateless setups.
+Defaults to `"database"` when a database is configured and `"cookie"` for database-less/stateless setups. If you only provide `secondaryStorage`, set this to `"database"` to store OAuth state there instead of in the cookie.
 
 ### `storeAccountCookie`
 
--- a/docs/content/docs/reference/options.mdx
+++ b/docs/content/docs/reference/options.mdx
@@ -474,6 +474,7 @@ export const auth = betterAuth({
 			userId: "user_id"
 		},
 		encryptOAuthTokens: true, // Encrypt OAuth tokens before storing them in the database
+		storeStateStrategy: "database", // Store OAuth state in the database instead of an encrypted cookie
 		storeAccountCookie: true, // Store account data after OAuth flow in a cookie (useful for database-less flows)
 		accountLinking: {
 			enabled: true,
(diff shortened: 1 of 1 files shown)


IanGraingerGMSL:
Thanks for this! 🙏 

I don't know if its interesting, but we're only using `secondaryStorage` - we didn't need a 'full' database to use this functionality. That might be worth mentioning, especially because without adding this, we were getting a 413 from AWS CloudFront when hosting our app using better-auth and nextjs because the cookies were making the requests too large. With this setting it worked 🙂 

So I guess we have an example config that works with nextjs + better-auth + opennext + dynamodb for secondaryStorage...

GautamBytes:
> Thanks for this! 🙏
> 
> I don't know if its interesting, but we're only using `secondaryStorage` - we didn't need a 'full' database to use this functionality. That might be worth mentioning, especially because without adding this, we were getting a 413 from AWS CloudFront when hosting our app using better-auth and nextjs because the cookies were making the requests too large. With this setting it worked 🙂
> 
> So I guess we have an example config that works with nextjs + better-auth + opennext + dynamodb for secondaryStorage...



> Thanks for this! 🙏
> 
> I don't know if its interesting, but we're only using `secondaryStorage` - we didn't need a 'full' database to use this functionality. That might be worth mentioning, especially because without adding this, we were getting a 413 from AWS CloudFront when hosting our app using better-auth and nextjs because the cookies were making the requests too large. With this setting it worked 🙂
> 
> So I guess we have an example config that works with nextjs + better-auth + opennext + dynamodb for secondaryStorage...

Good call, thanks! Added a short note that `secondaryStorage` can be used here without a full DB.

docs/content/docs/reference/options.mdx
```diff
@@ -504,7 +504,7 @@ Controls where OAuth state data is stored during the authentication flow.
 * `"cookie"`: Store the state payload in an encrypted cookie. This keeps OAuth state stateless and avoids a database write during the start of the flow.
 * `"database"`: Store the state payload in the verification storage/table, set a signed state cookie when starting the flow, and validate the stored state against the signed cookie sent by the browser on the OAuth callback request.
 
-Defaults to `"database"` when a database is configured and `"cookie"` for database-less/stateless setups.
+Defaults to `"database"` when a database is configured and `"cookie"` for database-less/stateless setups. If you only provide `secondaryStorage`, set this to `"database"` to store OAuth state there instead of in the cookie.
```
IanGraingerGMSL, line 507: suggestion: I think this could be clearer about 'database' vs 'secondaryStorage'.

I'm assuming (please check!) the default is `"database"` when a 'full' (what's the standard name for a 'non-secondary' database? 'main'? 'full'?) database is configued - but `"cookie"` when it isn't (regardless whether its a 'full' or 'secondaryStorage' database).

So perhaps just clarifying:
```suggestion:
When using only `secondaryStorage`, this will default to `"cookie"`. Set this to `"database"` to store OAuth state there.
```
Actually - I think this points to something confusing. I can't really see why you'd want to use "cookie" here now you can use `secondaryStorage`. I think the default when you have a 'full' database or `secondaryStorage` should _already_ be `"database"`? I found it confusing that I had to set this value - I expected this to just be _how it worked_.

ping-maxwell (approved):

docs/content/docs/reference/options.mdx
ping-maxwell: this could arguably come down to a code-level issue rather than purely a docs clarification, I recommend opening an issue.

PR #9570, opened 2026-05-12
docs: add per-ip rate limit to enterprise contact form

bytaesu:


bytaesu:
@copilot @cubic-dev-ai

> https://github.com/better-auth/better-auth/blob/main/docs/lib/ai-chat/rate-limit.ts

Since this follows the same logic, I just dismissed the comments. This is only meant to be a minimal safeguard. The refactoring cost later wouldn't be high either, so there's no need to overcomplicate it upfront.

PR #9593, opened 2026-05-12
fix(generic-oauth): use absolute redirect_uri with allowedHosts dynamic baseURL

thaoms:
## Summary

- When `baseURL` is configured with `allowedHosts` (dynamic config), `genericOAuth`'s social sign-in path sent a **relative** `redirect_uri` (e.g. `/oauth2/callback/provider`) to OAuth providers like Auth0 — causing the flow to fail.
- Root cause: the `init` closure captures `ctx.baseURL` at startup. For dynamic `allowedHosts` config, `ctx.baseURL` is `""` because the URL is resolved per-request into a cloned context; the original `ctx` is never mutated.
- Fix: compose the `redirect_uri` the same way the working endpoint handlers do — `new URL(data.redirectURI).origin` (per-request origin from the social sign-in call) + `ctx.options.basePath` (normalised at startup by `createAuthContext`). This mirrors `ctx.context.baseURL` used in route handlers without needing string-matching on `/callback/`.

The bug only affected the **social sign-in path** (`/sign-in/social` → `createAuthorizationURL` in the `init` closure). The dedicated `signIn.oauth2` endpoint at `/sign-in/oauth2` was already correct as it uses `ctx.context.baseURL` (per-request).

## Test Plan

- [x] Added a regression test in `generic-oauth.test.ts`: creates a `getTestInstance` with `baseURL: { allowedHosts: ["localhost:3000"] }`, calls `/sign-in/social`, and asserts the `redirect_uri` in the returned authorization URL is absolute (`/^https?:\/\//`)
- [x] All 61 tests in `generic-oauth.test.ts` pass
- [x] `pnpm typecheck` passes

### DIFF
--- a/.changeset/generic-oauth-absolute-redirect-uri-allowedhosts.md
+++ b/.changeset/generic-oauth-absolute-redirect-uri-allowedhosts.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Fix `genericOAuth` plugin producing a relative `redirect_uri` when using `allowedHosts` dynamic baseURL config. The `init` closure captured `ctx.baseURL` (empty string at startup for dynamic config) instead of deriving the base from the per-request resolved URL available in `data.redirectURI`.
--- a/packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
@@ -2513,3 +2513,54 @@ describe("oauth2", async () => {
 		});
 	});
 });
+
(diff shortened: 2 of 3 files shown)


gustavovalverde:
Superseded by #9783. The new PR redoes this architecturally: every `OAuthProvider` declares a `callbackPath: string` and the framework composes `redirect_uri = ctx.context.baseURL + provider.callbackPath` per request, which removes the init-time `ctx.baseURL` read across the OAuth surface. Your diagnosis and regression test shaped the new test coverage; you're credited as co-author on the lead commit.

gustavovalverde:
Closed by https://github.com/better-auth/better-auth/pull/9825 (as part of a bigger and different refactor)

PR #9597, opened 2026-05-12
feat(captcha): support exact and wildcard endpoint matching

bytaesu:
For archival purposes.


gustavovalverde:
#8639 was closed in favor of this one (with #9596 already merged); nothing specific to carry over.

PR #9600, opened 2026-05-12
fix(oauth-provider): return NOT_FOUND when consent update references a missing client

bytaesu:
updateConsentEndpoint re-checked !consent after the client lookup instead of !client, so a missing client never triggered the NOT_FOUND branch. The endpoint could then proceed against stale state. Check !client and return NOT_FOUND.


bytaesu:
@copilot review

PR #9603, opened 2026-05-12
fix(access): reject empty action lists and continue "OR" evaluation on unknown resources

bytaesu:


packages/better-auth/src/plugins/access/access.ts
bytaesu, line 14: It feels like the `role()` function doesn't have a clean separation of concerns. Part of it is the inconsistent return shape, but the logic itself also isn't very flat, which makes it harder to read.

bytaesu, line 12: It feels a bit unclean that the success and failure cases have different shapes. Also, if we model error as `string | null`, then success becomes unnecessary in the first place.

bytaesu:
@cursoragent

Open "Task" type issue based on https://github.com/better-auth/better-auth/pull/9603#discussion_r3230440367 https://github.com/better-auth/better-auth/pull/9603#discussion_r3230449824. Maybe you need to open general parent issue and put these into sub-issues.

PR #9618, opened 2026-05-14
feat(auth): add routeInputs option for custom request body fields

### LINKED ISSUE #9612: Extend existing endpoints with custom input options

Currently, Better Auth allows adding custom input fields to `signUp.email()` through schema fields using `input: true`, but there’s no equivalent way to extend `signIn.email()` or other routes with custom request properties.

In my plugin, I need to add custom inputs to both `signIn.email()` and `signUp.email()`. This is not possible with the current state of Better Auth.

I’d like to see a more generic and extensible system for extending API route inputs.

Instead of relying on schema fields with `input: true`, Better Auth could expose something like an `extendRoutes` option that allows adding custom input definitions to any route.

Example:

```ts
extendRoutes: {
  "/sign-in/email": {
    customField: {
      type: "string",
      required: true
    }
  },

  "/sign-up/email": {
    inviteCode: {
      type: "string"
    }
  }
}
```

These custom properties could then be accessed from hooks/middleware, like we can already do.

One alternative would be extending the existing `input: true` behavior to also work on `signIn.email()`.

However, this approach feels too tied to the schema system and does not provide enough flexibility.

### Additional context

I think this would also improve the developer experience significantly, because currently custom route inputs feel somewhat indirect and limited by the schema abstraction.

GautamBytes:
## Summary

Adds `routeInputs` as a generic way to declare validated request body fields for auth routes from both top-level auth options and plugins.

- Supports custom request inputs on routes like `/sign-in/email` and `/sign-up/email`
- Validates route inputs before hooks and handlers so values are available on `ctx.body`
- Rejects duplicate route input fields and collisions with built-in endpoint body fields
- Adds client and server type inference for top-level and plugin-declared route inputs
- Reflects custom route inputs in OpenAPI request body schemas
- Documents the new option and plugin authoring flow

Closes #9612.

## Notes

Route inputs are request-only fields. They are not persisted to user/session/account records unless separately configured through schema additional fields.

---
## Summary by cubic
Adds `routeInputs` to declare and validate custom request body fields on auth routes like `/sign-in/email` and `/sign-up/email`. Inputs are validated before hooks/handlers, available on `ctx.body`, fully typed on server and client, and included in OpenAPI.

- New Features
  - Configure `routeInputs` in top-level options or plugins.
  - Validates inputs and rejects duplicates or collisions with existing fields.
  - Client/server infer custom input types (including plugin-provided inputs).
  - OpenAPI request bodies include custom inputs with required/optional and defaults.

- Bug Fixes
  - Apply .optional before .default and include function defaults in type inference.

<sup>Written for commit 948ab412e84390fcd5ee44134f0612c611eecb9b. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/custom-route-inputs.md
+++ b/.changeset/custom-route-inputs.md
@@ -1,0 +1,6 @@
+---
+"better-auth": minor
+"@better-auth/core": minor
+---
+
+Add routeInputs for validating custom request body fields on auth routes.
--- a/docs/content/docs/authentication/email-password.mdx
+++ b/docs/content/docs/authentication/email-password.mdx
@@ -97,7 +97,7 @@ To sign a user in, you can use the `signIn.email` function provided by the clien
 </APIMethod>
 
 <Callout>
-  These are the default properties for the sign in email endpoint, however it's possible that with [additional fields](/docs/concepts/typescript#additional-fields) or special plugins you can pass different properties to the endpoint.
+  These are the default properties for the sign in email endpoint. You can add request-only fields with [`routeInputs`](/docs/reference/options#routeinputs), and plugins can also add route inputs that are available in hooks.
 </Callout>
 
 ### Sign Out
--- a/docs/content/docs/concepts/plugins.mdx
+++ b/docs/content/docs/concepts/plugins.mdx
@@ -236,6 +236,40 @@ This will add an `age` field to the `user` table and all `user` returning endpoi
   Don't store sensitive information in the `user` or `session` table. Create a new table if you need to store sensitive information.
 </Callout>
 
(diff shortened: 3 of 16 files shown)


0-Sandy:
@bytaesu When you have time, could you review this? Thanks

0-Sandy:
@bytaesu Sorry to bother you, but could you please take a look at this when you have a chance? Thanks again

aleksola (approved):

aleksola (approved):

0-Sandy:
@Bekacru Could you review this when you have a chance? Thanks

PR #9624, opened 2026-05-14
fix: `consumeVerificationValue` returns null for expired rows

bytaesu:
It's being handled separately at every call. If that's the case, I think it would be better to package it directly into the method itself.

The user-facing change is the error message text, so I don't think this should be a `minor` release. Let's go with `patch`.

### DIFF
--- a/.changeset/consume-verification-rejects-expired.md
+++ b/.changeset/consume-verification-rejects-expired.md
@@ -1,0 +1,6 @@
+---
+"better-auth": patch
+"@better-auth/oauth-provider": patch
+---
+
+Expired magic-link tokens and OAuth authorization codes are now reliably rejected. Magic-link verify redirects to `?error=INVALID_TOKEN` for expired tokens (was `?error=EXPIRED_TOKEN`). The OIDC, MCP, and `@better-auth/oauth-provider` `/token` endpoints return `error_description: "invalid code"` for expired codes (was `"code expired"`). The OAuth `error` value stays `invalid_grant`.
--- a/packages/better-auth/src/db/internal-adapter.ts
+++ b/packages/better-auth/src/db/internal-adapter.ts
@@ -1259,30 +1259,34 @@ export const createInternalAdapter = (
 					? [storedIdentifier, identifier]
 					: [storedIdentifier];
 
+			// After a JSON round-trip `expiresAt` arrives as a string, so coerce
+			// it back to a valid `Date` to match what the DB adapter returns.
+			const hydrateCachedVerification = (raw: unknown): Verification | null => {
+				if (!raw) return null;
+				const candidate =
+					typeof raw === "string"
+						? safeJSONParse<Verification>(raw)
+						: typeof raw === "object"
+							? (raw as Verification)
+							: null;
+				if (!candidate) return null;
+				const expiresAt = new Date(candidate.expiresAt);
+				if (!Number.isFinite(expiresAt.getTime())) return null;
+				return { ...candidate, expiresAt };
+			};
+
+			let consumed: Verification | null = null;
+
 			if (secondaryStorage && !options.verification?.storeInDatabase) {
-				const parseCachedVerification = (raw: unknown) => {
-					if (!raw) return null;
-					// Secondary storage wrappers can return either a JSON string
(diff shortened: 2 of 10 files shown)


.changeset/consume-verification-rejects-expired.md
```diff
@@ -0,0 +1,6 @@
+---
+"better-auth": patch
+"@better-auth/oauth-provider": patch
+---
+
+Expired magic-link tokens and OAuth authorization codes are now reliably rejected. Magic-link verify redirects to `?error=INVALID_TOKEN` for expired tokens (was `?error=EXPIRED_TOKEN`). The OIDC, MCP, and `@better-auth/oauth-provider` `/token` endpoints return `error_description: "invalid code"` for expired codes (was `"code expired"`). The OAuth `error` value stays `invalid_grant`.
```
gustavovalverde, line 6: For `@better-auth/oauth-provider`, expired authorization codes now fall through to the existing `!verification` branch, which still returns `error_description: "Invalid code"` with a capital `I`; this changeset says the package returns `"invalid code"`. Since the user-facing change is the error text, please either preserve that capitalization here or distinguish the package-specific response.

packages/better-auth/src/db/internal-adapter.ts
```diff
@@ -1300,67 +1304,67 @@ export const createInternalAdapter = (
...
+				if (consumed && secondaryStorage) {
+					await Promise.all(
+						identifiersToTry.map((stored) =>
+							secondaryStorage.delete(`verification:${stored}`),
+						),
+					);
+				}
 			}
 
+			// Single expiry gate. A row past its `expiresAt` is treated as already
```
gustavovalverde, line 1365: With this new expiry gate, `consumeVerificationValue` now treats expired rows as invalid, but the JSDoc above still says expiry is intentionally not checked here so callers can keep their own expiry handling. That contradiction is on a security-sensitive single-use credential path and can mislead future changes back into duplicating or removing the centralized expiry logic.

bytaesu: Fixed https://github.com/better-auth/better-auth/pull/9624/commits/bd792be6147beb06bc361c0bdafe5cc218dcad72

.changeset/consume-verification-rejects-expired.md
bytaesu: Fixed https://github.com/better-auth/better-auth/pull/9624/commits/dd513382cdc28e8796426598f7cf1416fca9a242

packages/better-auth/src/db/internal-adapter.ts
```diff
@@ -1300,67 +1304,67 @@ export const createInternalAdapter = (
...
+			// invalid, so callers can rely on a non-null return meaning "valid".
+			if (!consumed || consumed.expiresAt < new Date()) return null;
```
ping-maxwell, line 1367: super small nit: should probably use `<=` instead of just `<`, this is fine in practice but `<=` would be semantically cleaner since "at expiry time" means "expired". The old call-site code also used `<`, so this is consistent - just noting it.

ping-maxwell (dismissed):
I believe the `one-time-token` plugin also requires this logic, should that be included in this PR also?

packages/better-auth/src/db/internal-adapter.ts
bytaesu: > I believe the one-time-token plugin also requires this logic, should that be included in this PR also?

I think this is better iterated on in follow-up PRs. This PR worked on the existing call sites of `consumeVerificationValue`, and if there are any missing call sites, that should be considered a separate scope.

bytaesu: > super small nit: should probably use `<=` instead of just `<`, this is fine in practice but `<=` would be semantically cleaner since "at expiry time" means "expired". The old call-site code also used `<`, so this is consistent - just noting it.

I think we need a rule for whether something being expired should map to `<` or `<=`. I'll handle it as a follow-up task. Good point by the way 😁

gustavovalverde:
#8412 was closed in favor of this PR and #9622; nothing specific to carry over.

PR #9626, opened 2026-05-14
feat(sso): add callback for validated SSO responses

jdnvn:
## The problem

Some applications let their users create externally accessible applications or pages that should be protected by the user’s configured SSO provider. The people authenticating to those user-created apps are not necessarily users of the host application itself, so provisioning a better auth user and issuing a session is not desired.

Currently, the SSO plugin always continues from a validated provider response into better auth user provisioning/session creation. There is no hook to take over the callback after OIDC/SAML validation but before that finalization step, so applications that need this behavior have to reimplement the SSO flow from scratch.

## Proposed solution

- Add `onSSOAuthenticated` to the SSO plugin, called after OIDC or SAML provider responses are validated and user info is extracted
- Allow the hook to return a `Response` to take over the callback before Better Auth creates or links a user and sets a session cookie
- Keep existing SSO behavior unchanged when the hook returns nothing

---
## Summary by cubic
Adds `onSSOAuthenticated` to the SSO plugin so apps can handle validated OIDC/SAML callbacks before `better-auth` creates or links a user and sets a session. The hook receives a sanitized callback URL and can return a `Response` to take over; if nothing is returned, the existing flow runs as before.

- New Features
  - New `onSSOAuthenticated` hook in `@better-auth/sso` for OIDC and SAML flows. Receives: `protocol`, `userInfo`, `token` (OIDC only), `provider`, `callbackURL` (sanitized), `request`.
  - New types: `SSOAuthenticatedUserInfo` and `SSOAuthenticatedContext`, exported from `@better-auth/sso`.
  - Updated docs with a usage example and option description.
  - Tests added to verify redirect takeover and context payload.

<sup>Written for commit ceae329a2af89087c359cd3fb0dcd345a7a0cd0e. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/silver-sso-callbacks.md
+++ b/.changeset/silver-sso-callbacks.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+feat(sso): add callback for handling validated SSO responses
--- a/.changeset/silver-sso-callbacks.md
+++ b/.changeset/silver-sso-callbacks.md
@@ -2,4 +2,4 @@
 "better-auth": patch
 ---
 
-feat(sso): add callback for handling validated SSO responses
+Add an SSO callback hook for handling validated provider responses before Better Auth creates a user session.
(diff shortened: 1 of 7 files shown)


jdnvn:
hey @gustavovalverde , sorry to bother but do you mind reviewing this? also happy to open an issue for it and close this if the implementation doesn't feel right

PR #9629, opened 2026-05-15
fix(cookies): reject Set-Cookie values outside the RFC 6265 octet set

bytaesu:
> [!NOTE]
> Reference:
https://github.com/jshttp/cookie/blob/e8db32e04f7aa5affff00724c8a28dd0d17fc397/src/index.ts#L529-L537 

parseSetCookieHeader decoded percent-encoded values without checking that the result still fit the cookie-octet shape, so an input like name=foo%3Bbar parsed to foo;bar and would split into multiple cookies or attributes once the value flowed back into a Cookie header. Validate the decoded value against the same regex parseCookies already uses and drop violators on parse, so both header directions agree on the value shape.


bytaesu:
will open a new PR

PR #9630, opened 2026-05-15
fix(organization): wrap delete cascades in a transaction

bytaesu:
deleteOrganization and removeMember cascades now run inside runWithTransaction so a mid-cascade failure rolls back instead of leaving orphan rows. removeMember's per-team cleanup runs as a single deleteMany with an in clause to stay portable across transaction backends that reject parallel ops on the same session (MongoDB).


bytaesu:
This is an org plugin fix
cc @ping-maxwell

PR #9631, opened 2026-05-15
fix(cookies): percent-encode values on Cookie header serialize

bytaesu:
`setRequestCookie` and `applySetCookies` previously dropped values containing `;`, `"`, `\`, or other bytes outside the bare `cookie-octet` range. As a result, flows storing structured values (e.g. the OIDC provider plugin `oidc_login_prompt` JSON cookie) silently lost data.

Switch to encode-on-wire semantics:
- Keep values in semantic form internally
- `encodeURIComponent` at the wire boundary
- Decode during parsing

---

`bearer` now passes the decoded token to `setRequestCookie` to align with the new contract. Otherwise, signed cookies become double-encoded and fail HMAC verification.

### DIFF
--- a/.changeset/cookie-serialize-validate.md
+++ b/.changeset/cookie-serialize-validate.md
@@ -1,0 +1,6 @@
+---
+"better-auth": patch
+"@better-auth/electron": patch
+---
+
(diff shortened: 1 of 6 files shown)


bytaesu:
@cubic-dev-ai review

PR #8817, opened 2026-03-28
fix(session): preserve real session expiry during stateless cache refresh

cyphercodes:
## Description

Fixes #8770

When `cookieCache.refreshCache` is enabled (stateless mode), the get-session endpoint was incorrectly overwriting `session.session.expiresAt` with `now + cookieCache.maxAge` instead of preserving the real session expiry.

### Root Cause

In the stateless cache refresh path, `refreshedSession` was overriding `expiresAt` with the cookie cache TTL:

```typescript
const newExpiresAt = getDate(cookieMaxAge, "sec");
const refreshedSession = {
  session: {
    ...session.session,
    expiresAt: newExpiresAt, // ← overwrites real session expiry with cache TTL
  },
  ...
};
```

### Fix

Remove the `expiresAt: newExpiresAt` override. The spread `...session.session` already carries the real `expiresAt`, and `setCookieCache` sets the JWE cookie expiration independently from `authCookies.sessionData.attributes.maxAge`.

### Changes
- Removed `newExpiresAt` variable
- Removed `expiresAt: newExpiresAt` from `refreshedSession`

---
## Summary by cubic
Preserves real session expiry when `session.cookieCache.refreshCache` is enabled by removing the wrong `expiresAt` override in `getSession`, so long‑lived sessions aren’t cut to the cache TTL. Extends stateless‑mode tests to verify cookie rotation (`better-auth.session_data`), stable token (`better-auth.session_token`), and unchanged expiry; updates CI; fixes #8770.

<sup>Written for commit 1021ba6e50d6a2efbe16921cf6f699da30d09a27. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/8817?utm_source=github">Review in cubic</a></sup>

PR #9646, opened 2026-05-16
fix(session): preserve session.expiresAt across stateless refreshCache cycles

### LINKED ISSUE #8770: session.expiresAt overwritten with cookieCache.maxAge during stateless refreshCache renewal
## Describe the bug

When `cookieCache.refreshCache` is enabled (stateless mode — no `database` configured), the `get-session` endpoint overwrites `session.session.expiresAt` with `now + cookieCache.maxAge` instead of preserving the real session expiry. Every `getSession` call during a cache refresh returns a corrupted `session.expiresAt` (e.g. 10 minutes from now instead of 7 days).

## Root Cause

In `src/api/routes/session.ts`, the stateless cache refresh path creates a `refreshedSession` that overrides `expiresAt` with the cookie cache TTL:

```ts
const newExpiresAt = getDate(ctx.context.options.session?.cookieCache?.maxAge || 300, "sec");
const refreshedSession = {
  session: {
    ...session.session,
    expiresAt: newExpiresAt, // ← overwrites real session expiry (e.g. 7 days) with cache TTL (e.g. 10 min)
  },
  user: session.user,
  updatedAt: Date.now(),
};
await setCookieCache(ctx, refreshedSession, false);
return ctx.json({ session: parsedRefreshedSession, ... });
//                          ↑ expiresAt = now + 10min, not the real 7-day session expiry
```

`setCookieCache` already derives the JWE cookie TTL independently from `authCookies.sessionData.attributes.maxAge` — `newExpiresAt` serves no other purpose here and only corrupts `session.session.expiresAt`.

## Steps to Reproduce

1- Configure better-auth without a `database` (stateless mode — memory adapter used automatically)
2- Enable `cookieCache` with `refreshCache: true`:
```ts
const auth = betterAuth({
  session: {
    cookieCache: {
      enabled: true,
      strategy: 'jwe',
      maxAge: 60,
      refreshCache: true,
    },
  },
});
```
3- Sign in - observer `session.expiresAt` is correct (7 days from now)
4- Wait for `updateAge` to pass
5- Call `getSession` again - the refresh path triggers
6- `session.expiresAt` is now `now + 60s` instead of `now + 7 days`

## Expected behavior

`session.expiresAt` always reflects `session.expiresIn` (default 7 days). The cookie cache TTL is internal and must not leak into the public session object.

## Actual behavior

After the first refresh cycle, `session.expiresAt` equals `now + cookieCache.maxAge`. The corrupted value is both stored in the new JWE cookie (causing the cache to prematurely expire based on `maxAge` rather than real session lifetime) and returned in the JSON response.

## Fix

Remove `expiresAt: newExpiresAt` from `refreshedSession`. The spread `...session.session` already carries the real `expiresAt`, and `setCookieCache` sets the JWE `exp` independently:

```ts
// Before
const newExpiresAt = getDate(ctx.context.options.session?.cookieCache?.maxAge || 300, "sec");
const refreshedSession = {
  session: { ...session.session, expiresAt: newExpiresAt }, // remove override
  ...
};

// After
const refreshedSession = {
  session: { ...session.session }, // preserves real expiresAt
  ...
};
```

jaydeep-pipaliya:
## Summary

The `/get-session` stateless refresh path overwrote `session.session.expiresAt` with `now + cookieCache.maxAge` (the cache TTL, e.g. 5 minutes) instead of preserving the real session expiry from `session.expiresIn` (e.g. 7 days). The corrupted value was both returned in the JSON response and embedded in the next JWE cookie, so subsequent refreshes treated the session as expiring with the cache TTL.

`setCookieCache` already derives the cookie's own TTL from `authCookies.sessionData.attributes.maxAge`, so the override served no purpose other than corrupting the public expiry.

## Root cause

In `packages/better-auth/src/api/routes/session.ts`, the stateless refresh branch built `refreshedSession` with `expiresAt: newExpiresAt` where `newExpiresAt = getDate(cookieCache.maxAge, "sec")`. That value then flowed into:
- the JWE cookie payload (capping next refresh at the cache TTL)
- the JSON response (consumers seeing a 5-minute expiry instead of 7 days)

## Fix

Drop the `expiresAt` override. `...session.session` already carries the real expiry, and `setCookieCache` derives its own TTL from `sessionData.attributes.maxAge`.

## Test plan

- [x] New regression test: `should preserve session.expiresAt across stateless refreshCache cycles` (`session-api.test.ts`)
  - Configures `expiresIn: 7 days` + `cookieCache.maxAge: 5 min` + `refreshCache.updateAge: 60s`
  - Signs in, captures `firstSession.session.expiresAt`
  - Advances time past the refresh threshold
  - Asserts the refreshed `session.expiresAt` matches the original (not the cache TTL)
- [x] Test fails on `main` (confirmed) and passes after the fix
- [x] All 7 existing `refreshCache` tests still pass

Closes #8770

---
## Summary by cubic
Fixes stateless `/get-session` refresh in `better-auth` to preserve the real `session.expiresAt` (e.g., 7 days) instead of overwriting it with `cookieCache.maxAge`. Responses and JWE cookies now report the correct expiry, while the cache TTL remains managed by `setCookieCache`.

<sup>Written for commit c4f8fc25da1c9cdb7be3d0d8552e384819fc6a9b. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/9646?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/fix-stateless-cache-expires-at-corruption.md
+++ b/.changeset/fix-stateless-cache-expires-at-corruption.md
@@ -1,0 +1,15 @@
+---
+"better-auth": patch
+---
+
+fix(session): preserve real `session.expiresAt` across stateless `refreshCache` cycles
+
+When `cookieCache.refreshCache` was enabled without a database, the stateless
+refresh path on `/get-session` overwrote `session.expiresAt` with
+`now + cookieCache.maxAge`. The corrupted value was both returned in the
+response and embedded in the next JWE cookie, so subsequent refreshes treated
+the session as expiring with the cache TTL (e.g. 5 minutes) instead of the
+real `session.expiresIn` (e.g. 7 days). `setCookieCache` already derives the
+cookie's own TTL from `authCookies.sessionData.attributes.maxAge`, so the
+override served no purpose other than to corrupt the public expiry — it is now
+removed.
--- a/packages/better-auth/src/api/routes/session-api.test.ts
+++ b/packages/better-auth/src/api/routes/session-api.test.ts
@@ -1206,6 +1206,75 @@ describe("cookie cache refreshCache", async () => {
 	});
 
 	/**
+	 * @see https://github.com/better-auth/better-auth/issues/8770
+	 */
+	it("should preserve session.expiresAt across stateless refreshCache cycles", async () => {
+		const sessionExpiresIn = 60 * 60 * 24 * 7; // 7 days
+		const cookieCacheMaxAge = 300; // 5 minutes
+		const { client, testUser, cookieSetter } = await getTestInstance({
+			database: undefined as any,
+			session: {
+				expiresIn: sessionExpiresIn,
+				cookieCache: {
+					enabled: true,
+					strategy: "jwe",
+					maxAge: cookieCacheMaxAge,
+					refreshCache: {
+						updateAge: 60,
+					},
+				},
+			},
+		});
+
+		const headers = new Headers();
+		await client.signIn.email(
+			{
+				email: testUser.email,
+				password: testUser.password,
+			},
+			{
+				onSuccess: cookieSetter(headers),
+			},
+		);
+
+		const firstSession = await client.getSession({
+			fetchOptions: {
+				headers,
+				onSuccess: cookieSetter(headers),
+			},
+		});
+		const originalExpiresAt = new Date(
+			firstSession.data!.session.expiresAt,
+		).getTime();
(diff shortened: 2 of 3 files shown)


gustavovalverde:
Closing in favor of #8817. Both PRs apply an identical production fix to `session.ts`; #8817 is older and the active vehicle for issue #8770, and its regression test covers more ground (deleted DB session for true stateless mode, `session_data` rotation, `session_token` preservation).

PR #9651, opened 2026-05-16
fix(client): prevent duplicate session refetch on focus

### LINKED ISSUE #9613: Duplicate /get-session requests on window focus

### To Reproduce

1. Create a React app using `better-auth/react` and `authClient.useSession()` with default client session options.
2. Log in so a valid session exists.
3. Switch away from the tab, then make the tab visible again.
4. Observe the network requests to `/api/auth/get-session`.

On each visibility/focus refetch, two requests are made to `/api/auth/get-session`.

Expected: a single session refetch should happen for one focus/visibility event.

From stack traces, the duplicate appears to come from two independent paths:

1. `session-refresh` handles the visibility event and calls `$fetch("/get-session")` directly.
2. After that request resolves, `session-refresh` updates the session atom and toggles `$sessionSignal`, which notifies the `useAuthQuery` subscriber and triggers a second `/get-session` fetch.

This matches the root cause described in https://github.com/better-auth/better-auth/pull/8760, which looks like a possible solution by unifying session fetching under the session atom instead of having both the direct refresh path and the signal-driven query path.

1.6.10

React app using `better-auth/react`. Observed locally in browser devtools on `localhost`, with default `sessionOptions.refetchOnWindowFocus` behavior.

Client

```ts

  plugins: [/* app plugins */],
});
```

### Additional context

Related broader issue: https://github.com/better-auth/better-auth/issues/8910

Possible fix: https://github.com/better-auth/better-auth/pull/8760

The duplicate requests are visible from stack traces like:

```txt
visibilityHandler
setFocused
triggerRefetch
fetchSessionWithRefresh
betterFetch -> /get-session
```

and then:

```txt
sessionSignal.set
nanostores notify
useAuthQuery/query.mjs
betterFetch -> /get-session
```

yb175:
## Summary

Adds an explicit refresh marker flow for session refresh deduplication.

The refresh manager now emits a dedicated refresh marker before triggering the normal session signal. `useAuthQuery` consumes that marker to skip the immediate self-triggered refetch exactly once, while preserving existing invalidation and synchronization behavior.

The implementation keeps the behavior explicitly scoped to session refresh wiring instead of relying on implicit multi-atom ordering.

Closes #9613

## Changes

* added `$sessionRefreshMarker` for explicit refresh context
* added one-shot skip handling in `useAuthQuery`
* scoped deduplication through explicit `refreshMarkerAtom` config
* added regression coverage for duplicate focus refetches

---
## Summary by cubic
Prevents duplicate get-session requests when the tab regains focus by adding a refresh marker and a one-shot skip in the client. This removes the extra refetch while keeping session sync behavior unchanged.

- **Bug Fixes**
  - Added `$sessionRefreshMarker` and `refreshMarkerAtom` config to `useAuthQuery` to skip the next self-triggered refetch.
  - `createSessionRefreshManager` now emits the marker before toggling `sessionSignal` to dedupe the focus-triggered bounce.
  - Wired the marker into `getSessionAtom`, added a regression test to ensure one request per focus event, and included a changeset for a patch release.

<sup>Written for commit 012cef97172caab2630ea61696ddcf013cb2394f. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/9651?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/happy-apes-hug.md
+++ b/.changeset/happy-apes-hug.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Prevent duplicate session refetches triggered by focus events in the client.
--- a/packages/better-auth/src/client/query.ts
+++ b/packages/better-auth/src/client/query.ts
@@ -17,6 +17,10 @@ export type AuthQueryAtom<T> = PreinitializedWritableAtom<{
 	) => Promise<void>;
 }>;
 
+export interface UseAuthQueryConfig {
+	refreshMarkerAtom?: PreinitializedWritableAtom<any> | null;
+}
+
 export const useAuthQuery = <T>(
 	initializedAtom:
 		| PreinitializedWritableAtom<any>
@@ -33,6 +37,7 @@ export const useAuthQuery = <T>(
 				| ClientFetchOption
 		  )
 		| undefined,
+	config?: UseAuthQueryConfig,
 ) => {
 	const value: AuthQueryAtom<T> = atom({
 		data: null,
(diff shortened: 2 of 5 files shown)


yb175:
Hi @ping-maxwell , Just following up on this PR whenever you get a chance to review it. Thanks! 😁

PR #9664, opened 2026-05-18
fix: scope provider account lookups to current user

### LINKED ISSUE #9502: `internalAdapter`: fix buggy account/session helpers
- [x] deleteAccount(accountId) queries using the PK id field, not the accountId column. The parameter name is misleading, and this can either silently no-op or delete the wrong row if a PK happens to match.  
- [ ] findAccount(accountId) returns a non-deterministic first match. accountId itself is not schema-unique; the actual unique tuple is (userId, providerId, accountId).  
- [ ] The polymorphic behavior of deleteSessions(string | string[]) is also dangerous. Passing a single string is treated as a userId, which can lead to unexpected behavior when the caller intended to delete a single session token.

kgarg2468:
## Summary
- scope `/account-info` accountId lookups to the authenticated user's accounts unless a providerId is explicitly supplied
- use provider-scoped account lookup for Google One Tap so provider account IDs from other providers cannot block linking
- add regressions for colliding provider account IDs in account-info and One Tap flows

Fixes #9502.

## Testing
- `pnpm --filter better-auth exec vitest run src/api/routes/account.test.ts src/plugins/one-tap/one-tap.test.ts`

---
## Summary by cubic
Scope provider account lookups to the current user and provider to prevent collisions and ambiguous matches. Fixes Google One Tap linking when another provider shares the same account ID and adds `providerId` support to `/account-info`.

- **Bug Fixes**
  - `/account-info`: If only `accountId` is provided, search the authenticated user's accounts; if multiple matches exist, return not found and require `providerId`. Support provider-scoped lookup via `providerId` (`findAccountByProviderId`).
  - Google One Tap: Use provider-scoped lookup (`findAccountByProviderId` with `"google"`) so other providers with the same ID don't block linking.
  - Tests: Added regressions for cross-user and same-user provider ID collisions in `/account-info`, and for One Tap collisions.

<sup>Written for commit 3bbe38d894d3b81087bad66e41b12adbdd3b5f94. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/9664?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/provider-scoped-account-lookups.md
+++ b/.changeset/provider-scoped-account-lookups.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Avoid ambiguous provider account lookups in account info and Google One Tap flows when different providers issue the same account ID.
--- a/packages/better-auth/src/api/routes/account.test.ts
+++ b/packages/better-auth/src/api/routes/account.test.ts
@@ -227,6 +227,74 @@ describe("account", async () => {
 		});
 	});
 
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/9502
+	 */
+	it("should resolve account info from the current user's accounts when provider account IDs collide", async () => {
+		const { auth, signInWithTestUser, client } = await getTestInstance({
+			socialProviders: {
+				google: {
+					clientId: "test",
+					clientSecret: "test",
+					enabled: true,
+				},
+			},
+			account: {
+				accountLinking: {
+					allowDifferentEmails: true,
+				},
+				encryptOAuthTokens: true,
+			},
+		});
+		const ctx = await auth.$context;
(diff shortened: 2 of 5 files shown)


kgarg2468:
Thanks for pointing out the ambiguous no-provider case. I pushed 3bbe38d89 so account-info only falls back to providerless lookup when exactly one current-user account matches the provider-issued accountId; otherwise it returns ACCOUNT_NOT_FOUND and callers can pass providerId to disambiguate. Added a regression covering same-user cross-provider accountId collisions.\n\nValidation run:\n- `pnpm exec biome check packages/better-auth/src/api/routes/account.ts packages/better-auth/src/api/routes/account.test.ts`\n- `pnpm --filter better-auth exec vitest run src/api/routes/account.test.ts -t "providerId when a current user's provider account IDs collide"`

PR #9666, opened 2026-05-18
fix: support DPoP authorization scheme tokens

### LINKED ISSUE #8656: oauth-provider: userinfo, introspection, revocation, and MCP endpoints reject DPoP Authorization scheme (RFC 9449)

### To Reproduce

1. Configure `oauthProvider` with `tokenBinding` that returns `{ tokenType: "DPoP", cnf: { jkt } }` (i.e., any DPoP token binding implementation)
2. A client exchanges an authorization code at the token endpoint with a `DPoP` proof header
3. The token endpoint correctly issues a DPoP-bound access token — the JWT payload contains `cnf: { jkt: "<thumbprint>" }`
4. The client sends the DPoP-bound token to the userinfo endpoint using the correct scheme per [RFC 9449 §7.1](https://datatracker.ietf.org/doc/html/rfc9449#section-7.1): `Authorization: DPoP <token>` + `DPoP: <proof>`
5. The userinfo endpoint returns **400 invalid_request**

**Current:** The `userInfoEndpoint` only strips the `Bearer ` prefix from the Authorization header:

```js
// oauth-provider, userInfoEndpoint
const token = typeof authorization === "string" && authorization?.startsWith("Bearer ")
  ? authorization?.replace("Bearer ", "")
  : authorization;
```

When a client sends `Authorization: DPoP eyJ...`, the variable `token` becomes the literal string `"DPoP eyJ..."` (including the scheme prefix). This is passed to `validateAccessToken`, which fails JWT parsing and throws a 400.

The same `Bearer`-only extraction pattern exists in 3 other locations in the oauth-provider:
- **`mcpHandler`** — same Bearer-only extraction
- **Introspection endpoint** — same Bearer-only extraction
- **Revocation endpoint** — same Bearer-only extraction

**Expected:** When the `tokenBinding` option is configured and the token endpoint issues DPoP-bound tokens (`cnf.jkt` present via the binding result), resource endpoints should accept the `DPoP` Authorization scheme per [RFC 9449 §7.1](https://datatracker.ietf.org/doc/html/rfc9449#section-7.1):

1. Parse both `DPoP <token>` and `Bearer <token>` Authorization schemes
2. When DPoP scheme: extract the token, validate the DPoP proof from the `DPoP` header (method, URL, JWK thumbprint matches `cnf.jkt`)
3. When Bearer scheme with a DPoP-bound token: behavior depends on policy (reject for strict enforcement, or accept for permissive mode)

The token endpoint already accepts DPoP proofs and correctly binds tokens via the `tokenBinding` callback — but its sibling endpoints in the same package can't consume those tokens with the correct Authorization scheme.

1.5.5 (also reproduces on @better-auth/oauth-provider 1.5.1-beta.3)

### System info

```bash
{
  "system": { "platform": "darwin", "arch": "arm64" },
  "packageManager": { "name": "pnpm", "version": "10.12.1" },
  "frameworks": [{ "name": "next", "version": "16.1.6" }],
  "databases": [{ "name": "drizzle", "version": "0.45.1" }],
  "betterAuth": { "version": "1.5.5" }
}
```

Backend, Package

```typescript

// Any tokenBinding that returns DPoP-bound tokens triggers this bug
const dpopTokenBinding = async (input) => {
  // ... validate DPoP proof, compute JWK thumbprint ...
  return {
    tokenType: "DPoP",
    cnf: { jkt: thumbprint },
    responseHeaders: { "DPoP-Nonce": nonce },
  };
};

  plugins: [
    oauthProvider({
      tokenBinding: dpopTokenBinding,
      // Token endpoint correctly issues DPoP-bound tokens (cnf.jkt),
      // but userinfo/introspection/revocation can't consume them
      // when the client uses the DPoP Authorization scheme.
    }),
  ],
});
```

### Additional context

**Suggested fix:** A shared `extractAccessToken(request, opts?)` utility that handles both schemes:

```typescript
function extractAccessToken(request: Request): { token: string; scheme: "bearer" | "dpop" } | null {
  const authorization = request.headers.get("authorization");
  if (!authorization) return null;
  if (authorization.startsWith("Bearer ")) return { token: authorization.slice(7), scheme: "bearer" };
  if (authorization.startsWith("DPoP "))   return { token: authorization.slice(5), scheme: "dpop" };
  return null;
}
```

When `scheme === "dpop"`, the endpoint would validate the DPoP proof (checking `htm`, `htu`, JWK thumbprint against `cnf.jkt`) before proceeding. This could be wired through the existing `tokenBinding` callback or a new `accessTokenValidator` option on `oauthProvider`.

**Workaround:** We intercept the userinfo request before it reaches better-auth, validate the DPoP proof, and rewrite `Authorization: DPoP <token>` to `Authorization: Bearer <token>`:

```typescript
export async function rewriteDpopForUserinfo(request: Request): Promise<Request> {
  const authorization = request.headers.get("authorization");
  if (!authorization?.startsWith("DPoP ")) return request;

  const token = authorization.slice(5);
  const tokenPayload = decodeJwt(token);
  await validateDpopProof({ request, tokenPayload }); // custom validation

  const headers = new Headers(request.headers);
  headers.set("authorization", `Bearer ${token}`);
  return new Request(request.url, { method: request.method, headers });
}
```

kgarg2468:
## Summary
- centralize Authorization header token extraction for Bearer and DPoP schemes
- apply DPoP token parsing to introspection, revocation, userinfo, and MCP resource flows
- add coverage for DPoP and backward-compatible raw token handling

Fixes #8656.

## Testing
- `pnpm --filter @better-auth/oauth-provider exec vitest run src/utils/authorization.test.ts src/introspect.test.ts src/revoke.test.ts src/userinfo.test.ts src/mcp.test.ts`

---
## Summary by cubic
Accept DPoP authorization scheme tokens across OAuth endpoints and trim Authorization header values by centralizing token parsing. Fixes #8656 and preserves raw token handling for backward compatibility.

- **Bug Fixes**
  - Extract tokens from Bearer or DPoP schemes (case-insensitive) and trim whitespace via `getAuthorizationToken`.
  - Apply to introspection, revocation, `userinfo`, and MCP middleware.
  - Add tests for DPoP, raw tokens, and trimmed/case-insensitive values; add `dpop` to spelling; publish patch for `@better-auth/oauth-provider`.
  - No API changes; existing Bearer and raw token calls still work.

<sup>Written for commit b2b7e096e3b6ac957029f49dde0a0a925b941f84. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/9666?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/oauth-dpop-authorization-scheme.md
+++ b/.changeset/oauth-dpop-authorization-scheme.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/oauth-provider": patch
+---
+
+Accept DPoP authorization scheme values anywhere OAuth provider endpoints parse access tokens.
--- a/.cspell/auth-terms.txt
+++ b/.cspell/auth-terms.txt
@@ -1,6 +1,7 @@
 totp
 otps
 pkce
+dpop
 unban
 siwe
 scim
--- a/packages/oauth-provider/src/introspect.test.ts
+++ b/packages/oauth-provider/src/introspect.test.ts
@@ -205,6 +205,31 @@ describe("oauth introspect", async () => {
 		});
 	});
 
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/8656
+	 */
+	it("should accept DPoP authorization scheme for access token introspection", async () => {
+		const tokens = await getTokens();
+		const introspection = await client.oauth2.introspect(
+			{
+				client_id: oauthClient?.client_id,
+				client_secret: oauthClient?.client_secret,
+				token: `DPoP ${tokens.data?.access_token!}`,
+				token_type_hint: "access_token",
+			},
+			{
+				headers: {
+					accept: "application/json",
+					"content-type": "application/x-www-form-urlencoded",
+				},
+			},
+		);
+		expect(introspection.data).toMatchObject({
+			active: true,
+			client_id: oauthClient?.client_id,
+		});
+	});
+
 	it("should pass with token_type_hint access_token and sent opaque access_token", async () => {
 		const tokens = await getTokens();
 		const introspection = await client.oauth2.introspect(
--- a/packages/oauth-provider/src/introspect.ts
+++ b/packages/oauth-provider/src/introspect.ts
@@ -14,6 +14,7 @@ import type {
 } from "./types";
 import {
 	basicToClientCredentials,
+	getAuthorizationToken,
 	getClient,
 	getJwtPlugin,
 	getStoredToken,
@@ -424,9 +425,7 @@ export async function introspectEndpoint(
 	}
 
 	// Check token
-	if (token && typeof token === "string" && token.startsWith("Bearer ")) {
-		token = token.replace("Bearer ", "");
-	}
+	token = getAuthorizationToken(token) ?? "";
 	if (!token?.length) {
 		throw new APIError("BAD_REQUEST", {
 			error_description: "missing a required token for introspection",
--- a/packages/oauth-provider/src/mcp.ts
+++ b/packages/oauth-provider/src/mcp.ts
@@ -3,6 +3,7 @@ import { verifyAccessToken } from "better-auth/oauth2";
 import { APIError } from "better-call";
 import type { JWTPayload } from "jose";
 import type { Awaitable } from "./types/helpers";
+import { getAuthorizationToken } from "./utils";
 
 /**
  * A request middleware handler that checks and responds with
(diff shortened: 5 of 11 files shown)


kgarg2468:
Thanks for catching this. I pushed b2b7e096e so DPoP/Bearer extraction trims the matched token instead of returning trailing whitespace, and added a regression assertion for padded DPoP headers.\n\nValidation run:\n- `pnpm exec biome check packages/oauth-provider/src/utils/index.ts packages/oauth-provider/src/utils/authorization.test.ts`\n- `pnpm --filter @better-auth/oauth-provider exec vitest run src/utils/authorization.test.ts`

gustavovalverde:
Closing for an architectural redo. Parsing the `DPoP` scheme without verifying the proof or binding the access token via `cnf.jkt` would advertise [RFC 9449](https://datatracker.ietf.org/doc/html/rfc9449) support through `dpop_bound_access_tokens_required` while resource endpoints accept unbound tokens. #8656 stays open; the fix will land as a shared seam: token-endpoint issuance, proof verification, and Bearer-replay policy across `userinfo`, `introspect`, `revoke`, and `mcp`.

Issue #5378, opened 2025-10-17
Allow excluding user fields from cookie cache to prevent oversized cookies


gottfrois:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Configure BetterAuth with Microsoft OAuth provider and cookie cache enabled:

```ts
  export const auth = betterAuth({
    session: {
      cookieCache: {
        enabled: true,
        maxAge: 300
      }
    },
    socialProviders: {
      microsoft: {
        clientId: env.MICROSOFT_CLIENT_ID,
        clientSecret: env.MICROSOFT_CLIENT_SECRET,
        tenantId: env.MICROSOFT_TENANT_ID,
      }
    }
  });
```

2. Sign in using Microsoft OAuth
3. Microsoft returns a base64-encoded profile picture in the user.image field (often 10-50KB+)
4. BetterAuth attempts to cache the entire user object (including the large image) in the session cookie
5. The cookie exceeds the 4KB browser limit, causing a 500 error on GET /api/auth/callback/microsoft

### Current Behavior

When `session.cookieCache.enabled` is `true`, the entire user object (including the image field) is serialized into the cookie. There is no way to exclude specific core user fields from being cached. The `returned: false` property in `additionalFields` only works for custom fields, not for built-in fields like `user.image`.

### Expected Behavior

I should be able to configure which user fields are excluded from cookie cache to prevent oversized cookies. 

For example:
```ts
  session: {
    cookieCache: {
      enabled: true,
      maxAge: 300,
      excludeUserFields: ["image"] // Don't cache these user fields in cookies
    }
  }
```

This would allow cookie caching for performance benefits while avoiding the cookie size limit issue with large fields like profile images.

### Current Workaround

The only workaround is to disable cookie cache entirely with `enabled: false`, which eliminates the performance benefits of cookie caching.

### What version of Better Auth are you using?

1.3.11

### System info

```bash
npx @better-auth/cli info --json
error: unknown command 'info'
```

### Which area(s) are affected? (Select all that apply)

Backend

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth";
import { drizzleAdapter } from "better-auth/adapters/drizzle";

export const auth = betterAuth({
  appName: "Lia",
  secret: env.BETTER_AUTH_SECRET,
  session: {
    cookieCache: {
      // Currently disabled due to cookie size issues with Microsoft SSO
      // profile pictures encoded in base64 being too large
      enabled: false,
[...]

PR #9687, opened 2026-05-19
fix(oauth): strip inline data: URIs from OAuth profile image

### LINKED ISSUE #8338: JWT Payload Overflow: ERR_RESPONSE_HEADERS_TOO_BIG due to Base64 Encoded Images in User Claims

### To Reproduce

- Configure an auth provider that returns a large Base64 encoded image string in the user's profile data (claims).
- Attempt to sign in via this provider.
- Upon successful authentication and redirection back to the application, the browser will fail to load the page.
- Check the browser console/network tab to see the ERR_RESPONSE_HEADERS_TOO_BIG error.

Image

Current Behavior: The current behavior allows for large Base64 encoded images to be sent from an OAuth provider. These large images are stored as a Base64 string directly within the user's data. This can potentially cause a system failure, such as the ERR_RESPONSE_HEADERS_TOO_BIG error, if the user data becomes too large for headers.

Expected Behavior: The expected behavior would automatically handle and sanitize large Base64 image strings. Better Auth should ideally include a built-in feature or plugin that intercepts any user.image claim coming in as a Base64 string. It would then offload the storage of this image to an external provider (like S3 or Cloudinary), and replace the Base64 string with the public URL of the uploaded image before the user profile is persisted or the session is created. This would maintain a small session payload and prevent ERR_RESPONSE_HEADERS_TOO_BIG.

Image

1.5.2

### System info

```bash
/
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

nikhilgupta58:
### The bug

When an OAuth provider returns a profile with `image` as an inline `data:image/...;base64,...` URI, Better-Auth persists that entire blob into `user.image` and bakes it into the session cookie cache. The browser rejects the response with `ERR_RESPONSE_HEADERS_TOO_BIG` and the OAuth callback fails silently.

The reproducible source is Better-Auth's own Microsoft provider at `packages/core/src/social-providers/microsoft-entra-id.ts:261`, which fetches Graph `/me/photo/$value` and converts the binary to `data:image/jpeg;base64, <120 KB>`. Auth.js takes the same shortcut but defaults to a tiny 48×48 photo "to avoid running out of space in case the session is saved as a JWT" — Better-Auth does not.

The bloat also leaks into JWT payloads when the `jwt` plugin is enabled, because the default payload is `ctx.context.session!.user` (`packages/better-auth/src/plugins/jwt/sign.ts:155-156`), so a cookie-only fix would be incomplete.

### The fix

A single sanitizer in `handleOAuthUserInfo` (the central funnel that every OAuth code path lands in, including SSO/OIDC via `packages/sso/src/routes/sso.ts:1657` and SAML via `packages/sso/src/routes/saml-pipeline.ts:431`). It strips inline `data:` URIs from `userInfo.image` (case-insensitive per RFC 2397) and emits a warn-level log. OIDC Core §5.1 defines `picture` as a URL, so a `data:` URI is non-standard; the prefix is a near-zero-false-positive signal — legitimate signed CDN URLs and Unicode names never start with it.

```diff
+function sanitizeOAuthProfileImage<T extends { image?: string | null }>(
+  userInfo: T,
+  options: BetterAuthOptions,
+  log: { warn: (msg: string) => void },
+): T {
+  if (options.account?.allowInlineProfileImage) return userInfo;
+  const image = userInfo.image;
+  if (typeof image !== "string" || !image.toLowerCase().startsWith("data:")) {
+    return userInfo;
+  }
+  log.warn("Stripped inline data: URI from OAuth profile image ...");
+  return { ...userInfo, image: null };
+}

 export async function handleOAuthUserInfo(c, opts) {
-  const { userInfo, account, callbackURL, disableSignUp, overrideUserInfo } = opts;
+  const { account, callbackURL, disableSignUp, overrideUserInfo } = opts;
+  const userInfo = sanitizeOAuthProfileImage(
+    opts.userInfo, c.context.options, c.context.logger,
+  );
   // rest unchanged
```

New opt-out: `account.allowInlineProfileImage: boolean` (default `false`), following the existing `account.accountLinking.allowDifferentEmails` / `allowUnlinkingAll` naming convention.

### Why here and not the cookie serializer

A cookie-only fix would leave the DB row bloated AND leak the bloat into JWT tokens (the `jwt` plugin signs the full `session.user` by default per `plugins/jwt/sign.ts:155-156`). The central sanitizer at `handleOAuthUserInfo` is the lowest-blast-radius location that covers all three downstream sinks (DB row, cookie cache, JWT). It also covers every provider (built-in social, generic-oauth, ID-token sign-in, oauth-proxy, SSO/OIDC) with one edit; sanitizing inside each provider's `getUserInfo` would mean 30+ edits and miss future providers.

### `mapProfileToUser` still gets the original

`mapProfileToUser` runs inside each provider's `getUserInfo`, BEFORE `handleOAuthUserInfo`. Users who already wired up the documented workaround (capture the data URI in `mapProfileToUser` and hoist it to a CDN) are unaffected — they still see the original value and can transform it.

### `databaseHooks.user.create.before` users — silent behavior change

Users relying on `databaseHooks.user.create.before` to receive the data URI will now see `null` instead. Two recovery paths:

1. Set `account.allowInlineProfileImage: true` to preserve the data URI all the way into the hook (same behavior as before this PR).
2. Move the CDN-upload logic from `databaseHooks` into `mapProfileToUser` (preferred — runs before sanitization).

### Test plan

Eight regression tests added in `packages/better-auth/src/oauth2/link-account.test.ts` under a new `describe("oauth2 - data: URI profile image sanitization (#8338)", ...)` block, each pinned with `@see https://github.com/better-auth/better-auth/issues/8338`:

1. strips a data: URI from OAuth profile image by default (the core bug — fails on `main`, passes after fix)
2. preserves the data: URI when `account.allowInlineProfileImage: true`
3. invokes `mapProfileToUser` with the original data URI before sanitization (compatibility guard)
4. case-insensitive: strips `data:`, `Data:`, and `DATA:` per RFC 2397
5. does not modify a normal `https:` image URL (false-positive guard)
6. passes `null` (not the data URI) to `databaseHooks.user.create.before` after stripping (documents the behavior change)
7. strips on the `overrideUserInfo: true` update path (covers existing-user-relink scenarios)
8. asserts every Set-Cookie on the OAuth callback stays under `ALLOWED_COOKIE_SIZE` (4096 bytes per `packages/better-auth/src/cookies/session-store.ts:11`) — the bug-class regression guard

**Before fix on `main`:**

```
 RUN  v4.1.5

 × strips a data: URI from OAuth profile image by default
 × invokes mapProfileToUser with the original data: URI before sanitization
 × strips uppercase Data: and DATA: URIs (case-insensitive per RFC 2397)
 × passes null (not the data: URI) to databaseHooks.user.create.before after stripping

 Test Files  1 failed (1)
      Tests  4 failed | 4 passed | 18 skipped (26)
```

**After fix:**

```
 RUN  v4.1.5

 Test Files  1 passed (1)
      Tests  8 passed | 18 skipped (26)
```

**Full `link-account.test.ts`:** `26 passed (26)`
**Full `packages/better-auth/src/oauth2/`:** `39 passed (39)`
**Full `packages/better-auth/src/social.test.ts`:** `49 passed (49)`
**Full `packages/better-auth/src/plugins/generic-oauth/`:** `60 passed (60)`
**Full `packages/sso/` (covers `sso.ts:1657` + `saml-pipeline.ts:431`):** `362 passed | 2 todo`
**`tsc --project tsconfig.json --noEmit`:** clean
**Biome + cspell pre-commit hooks:** clean

### Migration

DB rows with bloated `user.image` from previous OAuth callbacks are overwritten on the next callback. To clean immediately:

```sql
UPDATE "user" SET "image" = NULL WHERE "image" LIKE 'data:%';
```

This SQL is also in the changeset.

### Risk

- Behavior change is narrow: `user.image` becomes `null` when a provider returns a `data:` URI. The warn log makes the change discoverable in dev.
- `mapProfileToUser` users: unaffected.
- `databaseHooks.user.create.before` users with a CDN-upload pipeline: silent behavior change. Documented opt-out (`allowInlineProfileImage: true`) and a recommended migration to `mapProfileToUser`.
- No risk to subscription flow / non-OAuth flow / SAML pipeline (SAML never sets `image`).
- Forward compatible: future Stripe-like providers that emit `data:` URIs are protected automatically.

### Out of scope (explicit, to head off scope creep)

- Generic byte-size cap on profile fields (false-positive risk on legitimate Unicode names, signed CDN URLs, Entra group claims; different design space).
- Changes to `microsoft-entra-id.ts` to default the photo size to 48×48 like Auth.js does (reasonable follow-up, orthogonal).
- `excludeUserFields` cookie-cache option from #5378 (different ask, superset of problems, larger design risk).
- Auto-upload to CDN (belongs in user `mapProfileToUser` callback).

### Open questions

1. **Naming:** `account.allowInlineProfileImage` matches the existing `account.accountLinking.allow*` convention. Alternative if maintainers prefer is `account.stripInlineProfileImage: false` (inverted polarity, opt-in default). Happy to flip in review.
2. **Default behavior:** if you'd rather ship as opt-in (default behavior unchanged, sanitizer only runs when explicit), I'll flip immediately — the warn log and recovery mechanism are valuable either way. Just need a steer.

---
## Summary by cubic
Strips inline `data:` URIs from OAuth profile images to prevent oversized cookies/JWTs and silent OAuth callback failures. Adds an opt-out via `account.allowInlineProfileImage`. Closes #8338.

- **Bug Fixes**
  - Sanitizes `user.image` to null when it starts with `data:` (case-insensitive) in the central `handleOAuthUserInfo` path.
  - Prevents cookie bloat and large JWTs across all providers and SSO/OIDC; logs a warning when stripping.
  - Keeps original value available to `mapProfileToUser` (runs before sanitization).
  - New option: `account.allowInlineProfileImage: true` to preserve inline images if you handle CDN upload downstream.
  - Adds targeted tests to cover default strip, opt-out, case-insensitivity, hook behavior, and Set-Cookie size. 

- **Migration**
  - Optional cleanup for old rows: run `UPDATE "user" SET "image" = NULL WHERE "image" LIKE 'data:%';` or let future callbacks overwrite.

<sup>Written for commit 379bcd83b573ff4212b1490055c2fd9b3534841c. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/9687?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/strip-inline-data-uri-oauth-profile-image.md
+++ b/.changeset/strip-inline-data-uri-oauth-profile-image.md
@@ -1,0 +1,17 @@
+---
+"better-auth": patch
+---
+
+fix(oauth): strip inline `data:` URIs from OAuth profile images to prevent `ERR_RESPONSE_HEADERS_TOO_BIG`. Providers like Microsoft Entra ID (via the library's own Graph `/me/photo/$value` fetch) and some Keycloak/OIDC setups embed base64 images as `data:image/...;base64,...` in the `picture` claim. That value cascaded into the session cookie cache and into JWT payloads (the `jwt` plugin signs the full user object by default), silently failing the OAuth callback in browsers because Set-Cookie blew past the per-cookie 4 KB limit and the response past the per-header 16 KB limit (Vercel / nginx default).
+
+`handleOAuthUserInfo` now strips inline `data:` URIs (case-insensitive per RFC 2397) from `user.image` and emits a warn-level log. OIDC Core §5.1 defines `picture` as a URL, so a `data:` URI is non-standard; the prefix is a near-zero-false-positive signal (legitimate signed CDN URLs and Unicode names never start with it).
+
+Opt out with `account.allowInlineProfileImage: true` if you have a `mapProfileToUser` callback or `databaseHooks.user.create.before` hook that hoists the inline image to a CDN. `mapProfileToUser` runs before the sanitizer, so it can still capture the original data URI even when this option is left at the default.
+
+Existing DB rows with bloated `user.image` will be overwritten on the next OAuth callback. To clean immediately:
+
+```sql
+UPDATE "user" SET "image" = NULL WHERE "image" LIKE 'data:%';
+```
+
+Closes #8338.
--- a/packages/better-auth/src/oauth2/link-account.test.ts
+++ b/packages/better-auth/src/oauth2/link-account.test.ts
@@ -1438,3 +1438,300 @@ describe("oauth2 - accountLinking.requireLocalEmailVerified: false opt-out", asy
 		expect(promoted?.emailVerified).toBe(true);
 	});
 });
+
+/**
+ * @see https://github.com/better-auth/better-auth/issues/8338
+ *
+ * Microsoft Entra ID and certain Keycloak/OIDC configurations return a
+ * `picture` claim that is an inline `data:image/...;base64,...` URI
+ * instead of a URL. When that value reaches `user.image` it bloats the
+ * session cookie cache past Chromium's per-cookie 4 KB limit and the
+ * per-header 16 KB limit (Vercel / nginx default), failing the OAuth
+ * callback with `ERR_RESPONSE_HEADERS_TOO_BIG`. It also bloats JWT
+ * payloads when the `jwt` plugin is enabled. These tests cover the
+ * sanitizer that strips inline data URIs by default, with an opt-out
+ * for users who hoist to a CDN downstream.
+ */
+describe("oauth2 - data: URI profile image sanitization (#8338)", async () => {
+	const DATA_URI_PROFILE_IMAGE = `data:image/jpeg;base64,${"x".repeat(120_000)}`;
+
+	async function setupInstance(opts?: {
+		allowInlineProfileImage?: boolean;
+		mapProfileToUser?: (profile: GoogleProfile) => Record<string, unknown>;
+		captureCreateBeforeImage?: { value: unknown };
+	}) {
+		const { auth, client, cookieSetter } = await getTestInstance({
+			socialProviders: {
+				google: {
+					clientId: "test",
+					clientSecret: "test",
+					enabled: true,
+					mapProfileToUser: opts?.mapProfileToUser,
+				},
+			},
+			account: {
+				allowInlineProfileImage: opts?.allowInlineProfileImage,
+			},
+			...(opts?.captureCreateBeforeImage
+				? {
+						databaseHooks: {
+							user: {
+								create: {
+									before: async (user) => {
+										opts.captureCreateBeforeImage!.value = user.image;
+										return { data: user };
+									},
+								},
+							},
+						},
+					}
+				: {}),
+		});
+		const ctx = await auth.$context;
+		return { auth, client, ctx, cookieSetter };
+	}
+
+	async function driveOAuthCallback(
+		client: Awaited<ReturnType<typeof setupInstance>>["client"],
+		cookieSetter: Awaited<ReturnType<typeof setupInstance>>["cookieSetter"],
+		profileOverrides: Partial<GoogleProfile>,
+	) {
+		server.use(
+			http.post("https://oauth2.googleapis.com/token", async () => {
+				const profile: GoogleProfile = {
+					email: "8338-test@example.com",
+					email_verified: true,
+					name: "Test User",
+					picture: "https://example.com/photo.jpg",
+					exp: 1234567890,
+					sub: "google_oauth_sub_8338",
+					iat: 1234567890,
+					aud: "test",
+					azp: "test",
+					nbf: 1234567890,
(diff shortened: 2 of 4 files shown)


ping-maxwell:
Not the right solution to the issue in this case.

nikhilgupta58:
Totally fair call, @ping-maxwell. Not bloating the public API surface for what's effectively a single-provider workaround makes sense, and `mapProfileToUser` is the cleaner escape hatch. Happy to help with the Microsoft Entra docs update if useful (a short snippet showing the `mapProfileToUser` pattern that hoists the data URI to a CDN, plus a note about `disableProfilePhoto: true` as the at-source opt-out). Otherwise I'll move on. Thanks for the quick triage.

PR #9699, opened 2026-05-20
fix(admin): route banned OAuth users to errorCallbackURL

bytaesu:
Related to https://discord.com/channels/1288403910284935179/1504897654587654224

The admin plugin `session.create.before` hook was sniffing OAuth callback paths and redirecting directly to the auth server's `baseURL`, bypassing the OAuth state's `errorCallbackURL` routing and breaking multi-domain deployments.

The query string changes from `?error=banned` to `?error=BANNED_USER`, matching `ADMIN_ERROR_CODES.BANNED_USER`. Kept as `patch` since the lowercase was an ad-hoc literal, not documented API.

### DIFF
--- a/.changeset/admin-banned-oauth-error-callback.md
+++ b/.changeset/admin-banned-oauth-error-callback.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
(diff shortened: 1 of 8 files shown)


bytaesu:
-> will open a new pr

PR #9702, opened 2026-05-20
fix: redirect hook rejections to errorCallbackURL across auth callback flows

bytaesu:
Related to https://discord.com/channels/1288403910284935179/1504897654587654224

The admin plugin's `session.create.before` hook was sniffing two OAuth callback paths (`/callback`, `/oauth2/callback`) and redirecting directly to the auth server's `baseURL`, bypassing the OAuth state's `errorCallbackURL` and breaking multi-domain deployments. Other browser-redirect callbacks (oauth-proxy, SSO OIDC, SSO SAML) didn't match the sniff and returned JSON 4xx instead, also broken UX.

The hook now throws a plain `APIError`. All five browser-redirect callbacks catch it and route through `errorCallbackURL`, while the  id-token JSON sign-in bubbles it as before. The query value changes from `?error=banned` to `?error=BANNED_USER`, matching `ADMIN_ERROR_CODES.BANNED_USER`. Kept as `patch` since the lowercase was an ad-hoc literal, not documented API.

### DIFF
--- a/packages/better-auth/src/api/routes/callback.ts
+++ b/packages/better-auth/src/api/routes/callback.ts
@@ -268,6 +268,10 @@ export const callbackOAuth = createAuthEndpoint(
 			overrideUserInfo: provider.options?.overrideUserInfoOnSignIn,
 		});
 		if (result.error) {
+			if (result.errorCode) {
+				c.context.logger.error(result.errorCode);
+				return redirectOnError(result.errorCode, result.error);
+			}
 			c.context.logger.error(result.error.split(" ").join("_"));
 			return redirectOnError(result.error.split(" ").join("_"));
 		}
--- a/packages/better-auth/src/api/routes/callback.ts
+++ b/packages/better-auth/src/api/routes/callback.ts
@@ -9,6 +9,7 @@ import { handleOAuthUserInfo } from "../../oauth2/link-account";
 import { parseState } from "../../oauth2/state";
 import { setTokenUtil } from "../../oauth2/utils";
 import { HIDE_METADATA } from "../../utils/hide-metadata";
+import { isAPIError } from "../../utils/is-api-error";
 
 const schema = z.object({
 	code: z.string().optional(),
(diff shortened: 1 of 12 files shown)


packages/better-auth/src/plugins/admin/admin.ts
```diff
@@ -113,19 +113,6 @@ export const admin = <O extends AdminOptions>(options?: O | undefined) => {
...
-												ctx.context.options.onAPIError?.errorURL ||
-												`${ctx.context.baseURL}/error`;
-											throw ctx.redirect(
-												`${redirectURI}?error=banned&error_description=${opts.bannedUserMessage}`,
-											);
-										}
-
 										throw APIError.from("FORBIDDEN", {
 											message: opts.bannedUserMessage,
 											code: "BANNED_USER",
```
bytaesu, line 118: Follow-up:

This shouldn't be hardcoded. It should be imported from the existing error-code file.

packages/better-auth/src/api/routes/callback.ts
```diff
@@ -253,20 +254,28 @@ export const callbackOAuth = createAuthEndpoint(
...
+					(provider.disableImplicitSignUp && !requestSignUp) ||
+					provider.options?.disableSignUp,
+				overrideUserInfo: provider.options?.overrideUserInfoOnSignIn,
+			});
+		} catch (e) {
+			if (isAPIError(e) && e.body?.code) {
+				return redirectOnError(e.body.code, e.body.message);
+			}
+			throw e;
+		}
```
bytaesu, line 278: follow-up:

this catch only exists because hooks can't express "reject with reason" via the current `databaseHooks` contract. Extending it to a `{ data, error }` shape where hooks express rejection via the `error` field would remove this idiom across the OAuth callbacks, the split/join transform below, and the `error: e.message` asymmetry in link-account.ts. It would also cover non-OAuth browser flows (magic-link, email-verification, electron) where admin throws still leak as JSON.

packages/sso/src/routes/sso.ts
```diff
@@ -1654,31 +1655,43 @@ async function handleOIDCCallback(
...
+			const baseURL = errorURL || callbackURL;
+			const params = new URLSearchParams({ error: e.body.code });
+			if (e.body.message) params.set("error_description", e.body.message);
+			const sep = baseURL.includes("?") ? "&" : "?";
+			throw ctx.redirect(`${baseURL}${sep}${params.toString()}`);
+		}
+		throw e;
+	}
 	if (linked.error) {
 		throw ctx.redirect(`${errorURL || callbackURL}?error=${linked.error}`);
```
bytaesu, line 1696: follow-up:

looks like URL encoding is needed

PR #9703, opened 2026-05-20
fix(core): convert jose verification failures to UNAUTHORIZED APIError

### LINKED ISSUE #9654: verifyAccessToken passes jose verification errors through as raw Errors instead of APIError(UNAUTHORIZED)

### Reproduction

`verifyAccessToken` (`better-auth/oauth2`, defined in `@better-auth/core/src/oauth2/verify.ts`) catches errors from `verifyJwsAccessToken` and translates `JWTExpired` / `JWTInvalid` into `APIError("UNAUTHORIZED")`. **Every other error class — including the jose verification failures that are the normal failure mode for an invalid bearer token — falls through `else throw error` as a raw `Error`.** Callers downstream then have to either know jose's error taxonomy or surface these as 500s.

#### Minimal direct repro (no MCP needed)

```ts
import { verifyAccessToken } from "better-auth/oauth2";

// JWT with valid signature but `aud` that doesn't match `verifyOptions.audience`
await verifyAccessToken(tokenWithWrongAud, {
  jwksUrl: "https://auth.example.com/jwks",
  verifyOptions: {
    issuer: ["https://auth.example.com"],
    audience: "https://mcp.example.com/v1",
  },
});
```

Observed: throws raw `JWTClaimValidationFailed` from jose.
Expected: throws `APIError("UNAUTHORIZED")`, matching the existing typed-error contract for `JWTExpired` / `JWTInvalid`.

The same fall-through applies to `JWTClaimValidationFailed`, `JWSSignatureVerificationFailed`, `JWKSNoMatchingKey`. (`JWTExpired` and `JWTInvalid` already get translated; `TypeError` / `JWSInvalid` are intentionally swallowed as an opaque-token fallback.)

#### Downstream symptom in `mcpHandler` (the visible failure)

The most visible consequence is `mcpHandler` in `@better-auth/oauth-provider`. Setup:

```ts
import { mcpHandler } from "@better-auth/oauth-provider";

const issuer = "https://auth.example.com";
const audience = "https://mcp.example.com/v1";

app.all("/v1*", (request) =>
  mcpHandler(
    {
      jwksUrl: `${issuer}/jwks`,
      verifyOptions: { issuer: [issuer], audience },
    },
    handler,
  )(request),
);
```

Send a request with the wrong-audience token:

```
GET /v1/anything
Authorization: Bearer <token-with-wrong-aud>
Accept: application/json, text/event-stream
```

`mcpHandler` catches the raw jose error in `handleMcpErrors`, doesn't match `isAPIError`, falls through to `throw error`, and the surrounding framework surfaces it as `500 Internal Server Error` with no `WWW-Authenticate` header. RFC 9728's OAuth recovery path is broken.

**Additional case:** a JWT whose header has no `kid` triggers `throw new Error("Missing jwt kid")` from `getJwks` (`verify.ts:85`) and 500s through the same path. This is a Better-Auth-local precondition — jose itself is fine with `kid`-less JWTs per RFC 7515 §4.1.4 — but it presents to callers identically to the jose-error cases: an untyped `Error` they have no way to handle.

**Current**: `verifyAccessToken` throws raw jose `Error` instances for most verification-failure modes (wrong audience, bad signature, unknown key, missing `kid`). Callers that route by `APIError` (including `mcpHandler` via `handleMcpErrors`) don't recognize them, so they surface as `500 Internal Server Error` with no `WWW-Authenticate` header.

**Expected**: `verifyAccessToken` throws `APIError("UNAUTHORIZED")` for all client-token-shape failures, matching the existing typed-error contract for `JWTExpired` / `JWTInvalid`. Infrastructure failures (`JWKSTimeout`, `JWKSInvalid`, `JWKSMultipleMatchingKeys`) remain as raw jose errors so they continue to surface as 500s. Downstream callers like `mcpHandler` then naturally produce `401 Unauthorized` with the `WWW-Authenticate: Bearer resource_metadata="…", scope="…"` challenge that `handleMcpErrors` already builds for `APIError UNAUTHORIZED`.

#### Why this matters

1. **Typed error contracts at the public API boundary.** `verifyAccessToken` is exported from `better-auth/oauth2` as a public API. Callers shouldn't need to know about jose's error class taxonomy (`JWTClaimValidationFailed`, `JWSSignatureVerificationFailed`, `JWKSNoMatchingKey`, …) to handle the most common failure mode — an invalid bearer token. The existing partial translation for `JWTExpired` / `JWTInvalid` already commits to this contract; the rest is a completion of that work.
2. **Breaks the RFC 9728 OAuth recovery path in `mcpHandler`.** OAuth-aware MCP clients rely on the `WWW-Authenticate: Bearer resource_metadata="…"` challenge to re-discover the protected resource metadata and re-run their OAuth flow. A `500` drops that header entirely, so the client has no path back to a valid token.
3. **Misclassifies telemetry.** Real client auth failures (wrong token, expired token, key rotation gap) show up as server errors in Sentry / Datadog / structured logs, drowning out genuine server-side incidents and making alerts noisy.
4. **Is reachable from normal user behavior.** In a multi-resource OAuth deployment (a single auth server fronting multiple protected resources with different audiences), any user who copies a token intended for resource A into a client talking to resource B will hit this code path. The wrong-audience case is the _normal_ failure mode for cross-resource token confusion.

#### Root cause

The fix layer is `verifyAccessToken` in `@better-auth/core/src/oauth2/verify.ts`, which already implements partial translation:

```ts
try {
  payload = await verifyJwsAccessToken(token, { ... });
} catch (error) {
  if (error instanceof Error) {
    if (error.name === "TypeError" || error.name === "JWSInvalid") {
      // likely an opaque token (continue)
    } else if (error.name === "JWTExpired") {
      throw new APIError("UNAUTHORIZED", { message: "token expired" });
    } else if (error.name === "JWTInvalid") {
      throw new APIError("UNAUTHORIZED", { message: "token invalid" });
    } else {
      throw error;   // ← everything else falls through here
    }
  }
}
```

Every other jose error class — `JWTClaimValidationFailed`, `JWSSignatureVerificationFailed`, `JWKSNoMatchingKey` — falls through `else throw error`. So does the plain `Error("Missing jwt kid")` thrown by `getJwks` at `verify.ts:85` (a Better-Auth-local precondition, not a jose error).

`mcpHandler` then surfaces these as 500 via `handleMcpErrors`, which only normalises `APIError("UNAUTHORIZED")` into a 401 + `WWW-Authenticate`.

Affected error classes (verified against `jose@^6`, which `@better-auth/oauth-provider@1.6.9` depends on):

**Client auth failures (should be 401):**
- `JWTClaimValidationFailed` — wrong `aud`, `iss`, `nbf`, etc.
- `JWTExpired`
- `JWTInvalid`
- `JWSSignatureVerificationFailed`
- `JWKSNoMatchingKey` — token signed by a key the server doesn't trust (key rotation gap)
- `JWSInvalid` — currently swallowed as "likely an opaque token"; ends up as 401 via the function's terminal `"no token payload"` check, so behaviour is correct but for the wrong reason

**Server-side failures (should remain 500):**
- `JWKSTimeout` — JWKS endpoint unreachable
- `JWKSInvalid` — JWKS endpoint returned a malformed document
- `JWKSMultipleMatchingKeys` — ambiguous key selection

**Better-Auth-local precondition (should be 401):**
- `getJwks` `throw new Error("Missing jwt kid")` — a `kid`-less JWT is spec-valid (RFC 7515 §4.1.4) but unverifiable in Better Auth's `kid`→JWKS lookup flow

#### Suggested fix

Rather than maintaining an explicit allowlist of every jose error name that means "client gave us a bad token," **invert the rule**: anything that's a jose error and isn't on the small list of infrastructure failures is an auth failure. This is more stable across jose versions (new client-side error classes get treated correctly by default) and shorter to maintain.

```ts
import { APIError } from "better-call";
import { errors as joseErrors } from "jose";

// Small, stable set: jose errors that mean "server problem," not "bad token."
const JOSE_INFRA_CODES = new Set<string>([
  joseErrors.JWKSTimeout.code,             // ERR_JWKS_TIMEOUT
  joseErrors.JWKSInvalid.code,             // ERR_JWKS_INVALID
  joseErrors.JWKSMultipleMatchingKeys.code,// ERR_JWKS_MULTIPLE_MATCHING_KEYS
]);

try {
  payload = await verifyJwsAccessToken(token, { ... });
} catch (error) {
  // Preserve opaque-token fallback: structurally non-JWT tokens may be opaque
  // tokens that should fall through to remote introspection.
  if (error instanceof Error && (error.name === "TypeError" || error.name === "JWSInvalid")) {
    // payload stays undefined; falls through to remote verify or the terminal
    // "no token payload" 401 check.
  } else if (error instanceof joseErrors.JOSEError) {
    if (JOSE_INFRA_CODES.has(error.code)) {
      throw error;   // genuine 500
    }
    // Generic message — do not leak jose's claim-mismatch detail to
    // unauthenticated clients (could otherwise reveal accepted audiences).
    throw new APIError("UNAUTHORIZED", { message: "invalid access token" });
  } else if (error instanceof Error) {
    throw error;
  } else {
    throw new Error(error);
  }
}
```

Notes on this approach:

- **Match by `.code`, not `.name`.** Each jose error has a stable `.code` (`ERR_JWT_*`, `ERR_JWS_*`, `ERR_JWKS_*`) that's part of jose's API surface; `.name` is the JS class name, more convention than contract.
- **`instanceof jose.JOSEError` is the discriminator.** It cleanly excludes random framework / `TypeError` / non-jose throws without listing them.
- **Future-proof.** New jose error classes for token verification (e.g. a hypothetical `JWTAudienceMismatch` split out from `JWTClaimValidationFailed`) automatically fall into the 401 bucket. New infra error classes would need to be added to the deny-list, but those are far rarer.
- **The body deliberately doesn't surface jose's detailed message** (e.g. `unexpected "aud" claim value`) so the server's accepted audiences aren't leaked.
- **Preserves the existing opaque-token fallback** for `TypeError` / `JWSInvalid` — structurally non-JWT tokens still fall through to remote introspection (if configured) or the terminal `"no token payload"` 401 check.

**Second fix — at the precondition throw in `getJwks` (`verify.ts:85`):**

The `kid`-less plain `Error("Missing jwt kid")` is a Better-Auth-local invariant violation, not a jose error, so the rule above can't classify it. It should throw the same `APIError("UNAUTHORIZED")` directly at the throw site:

```ts
if (!jwtHeaders.kid) {
  throw new APIError("UNAUTHORIZED", { message: "invalid access token" });
}
```

After both fixes, `handleMcpErrors`'s existing `isAPIError(error) && error.status === "UNAUTHORIZED"` branch handles the response shape — no changes needed in `@better-auth/oauth-provider`. The upstream fix belongs at the throw site for clean error semantics across all callers of `verifyAccessToken` (not just `mcpHandler`).

#### Behavioral change scope

This is intentionally a behavioral change for **every caller of `verifyAccessToken`**, not just `mcpHandler`. `verifyAccessToken` is re-exported as a public API from `better-auth/oauth2`, so the proposed change affects what any caller sees in the error channel:

| Error                                                       | Before                                       | After                                              |
| ----------------------------------------------------------- | -------------------------------------------- | -------------------------------------------------- |
| `JWTExpired`                                                | `APIError UNAUTHORIZED` (`"token expired"`)  | `APIError UNAUTHORIZED` (`"invalid access token"`) |
| `JWTInvalid`                                                | `APIError UNAUTHORIZED` (`"token invalid"`)  | `APIError UNAUTHORIZED` (`"invalid access token"`) |
| `JWTClaimValidationFailed`                                  | raw jose error (uncaught → 500 at caller)    | `APIError UNAUTHORIZED` (`"invalid access token"`) |
| `JWSSignatureVerificationFailed`                            | raw jose error (uncaught → 500 at caller)    | `APIError UNAUTHORIZED` (`"invalid access token"`) |
| `JWKSNoMatchingKey`                                         | raw jose error (uncaught → 500 at caller)    | `APIError UNAUTHORIZED` (`"invalid access token"`) |
| `getJwks` `Error("Missing jwt kid")`                        | plain `Error` (uncaught → 500 at caller)     | `APIError UNAUTHORIZED` (`"invalid access token"`) |
| `JWKSTimeout` / `JWKSInvalid` / `JWKSMultipleMatchingKeys`  | raw jose error                               | raw jose error (unchanged — server infra failures) |
| `TypeError` / `JWSInvalid` (opaque-token fallback)          | swallowed                                    | swallowed (unchanged)                              |

Two notes for maintainers evaluating this:

1. **Error message becomes generic.** Callers that branch on the message text of `JWTExpired` / `JWTInvalid` (currently `"token expired"` / `"token invalid"`) will see `"invalid access token"` instead. The unification is intentional — leaking jose's claim-mismatch detail (e.g. `unexpected "aud" claim value`) to unauthenticated clients reveals accepted audiences. Callers that branch on the typed `APIError` status (`UNAUTHORIZED`) keep working. If preserving the distinct messages is preferred, they can be kept as separate branches in the same translation block — the inversion is what matters.
2. **No more raw jose errors at the API boundary.** Callers that wanted to distinguish claim-mismatch from signature failure for their own logs would lose that distinction. The right escape hatch (if needed) is to log the underlying jose error inside `verifyAccessToken` before translating — not to expose jose error classes through a public API surface.

`@better-auth/oauth-provider@1.6.9` (also originally reproduced on `1.4.18`; the relevant `mcpHandler` / `handleMcpErrors` path is unchanged between those versions)

### System info

```json
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:12 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6050",
    "release": "25.5.0",
    "cpuCount": 18,
    "cpuModel": "Apple M5 Max",
    "totalMemory": "128.00 GB",
    "freeMemory": "0.19 GB"
  },
  "node": {
    "version": "v24.14.1",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.13"
  },
  "frameworks": null,
  "databases": [
    {
      "name": "drizzle",
      "version": "catalog:drizzle-deps"
    }
  ],
  "betterAuth": {
    "version": "1.6.9",
    "config": null
  }
}
```

> Note: `npx auth info` reported `betterAuth.version: "Unknown"` because the CLI couldn't read the auth config export. The resolved version is `better-auth@1.6.9` / `@better-auth/oauth-provider@1.6.9` from the workspace lockfile — filled in manually above.

- Backend
- Package (`@better-auth/oauth-provider`, `@better-auth/core`)

MCP server side (where the 500 occurs):

```typescript
import { mcpHandler } from "@better-auth/oauth-provider";

const issuer = "https://auth.example.com";
const audience = "https://mcp.example.com/v1";

app.all("/v1*", (request, handler) =>
  mcpHandler(
    {
      jwksUrl: `${issuer}/jwks`,
      verifyOptions: { issuer: [issuer], audience },
    },
    handler,
  )(request),
);
```

Auth server side (mints tokens with a different audience — the cross-resource confusion case):

```typescript

  plugins: [
    oauthProvider({
      // …
      validAudiences: ["https://api.example.com"], // different from MCP audience above
    }),
  ],
});
```

### Additional context

- Verified against `@better-auth/oauth-provider@1.6.9` and `jose@^6.1.3` (the version `oauth-provider@1.6.9` pulls in).
- The two throw sites are independent: the inverted-allowlist fix in `verifyAccessToken` covers the jose-error cases (`JWTClaimValidationFailed`, `JWTExpired`, etc.); the `getJwks` change covers `Missing jwt kid`. A `kid`-less JWT bypasses jose entirely (it fails Better Auth's precondition before `jwtVerify` is called), so it cannot be classified by the jose-error rule.

umeshchandra-rao:
Fixes #9654.

This PR ensures that `verifyAccessToken` translates all jose client-token-shape verification failures (such as a wrong audience or missing `kid`) into an `APIError("UNAUTHORIZED")` rather than letting them fall through as unhandled raw `Error`s. 

- Known server-side infrastructure failures (`JWKSTimeout`, `JWKSInvalid`, `JWKSMultipleMatchingKeys`) are intentionally kept as raw exceptions to correctly surface as 500s.
- Existing opaque-token fallback logic remains intact.

Testing:
- Checked error logic structure against `#9654` suggested fixes.

---
## Summary by cubic
Standardizes access token verification errors by converting `jose` verification failures in `verifyAccessToken` into `APIError("UNAUTHORIZED")`, returning consistent 401s. JWKS infrastructure failures still bubble up as 500s.

- **Bug Fixes**
  - Map all non-infra `jose` `JOSEError`s to `APIError("UNAUTHORIZED", { message: "invalid access token" })`.
  - Keep JWKS infra errors (`JWKSTimeout`, `JWKSInvalid`, `JWKSMultipleMatchingKeys`) as raw to surface 500s.
  - Treat missing `kid` as unauthorized instead of a generic error.
  - Preserve opaque-token fallback for `TypeError`/`JWSInvalid`.

<sup>Written for commit 6106f4dc792c2073393e5320064b73cfb012f659. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/9703?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/core/src/oauth2/verify.ts
+++ b/packages/core/src/oauth2/verify.ts
@@ -7,6 +7,7 @@ import type {
 	ProtectedHeaderParameters,
 } from "jose";
 import {
+	errors as joseErrors,
 	createLocalJWKSet,
 	decodeProtectedHeader,
 	jwtVerify,
@@ -15,6 +16,13 @@ import {
 import { logger } from "../env";
 
 /** Last fetched jwks used locally in getJwks @internal */
+// Small, stable set: jose errors that mean "server problem," not "bad token."
+const JOSE_INFRA_CODES = new Set<string>([
+	joseErrors.JWKSTimeout?.code || "ERR_JWKS_TIMEOUT",
+	joseErrors.JWKSInvalid?.code || "ERR_JWKS_INVALID",
+	joseErrors.JWKSMultipleMatchingKeys?.code || "ERR_JWKS_MULTIPLE_MATCHING_KEYS",
+]);
+
 let jwks: JSONWebKeySet | undefined;
 
 export interface VerifyAccessTokenRemote {
@@ -82,7 +90,9 @@ export async function getJwks(
 		throw new Error(error as unknown as string);
 	}
 
-	if (!jwtHeaders.kid) throw new Error("Missing jwt kid");
+	if (!jwtHeaders.kid) {
+		throw new APIError("UNAUTHORIZED", { message: "invalid access token" });
+	}
 
 	// Fetch jwks if not set or has a different kid than the one stored
 	if (!jwks || !jwks.keys.find((jwk) => jwk.kid === jwtHeaders.kid)) {
@@ -134,20 +144,20 @@ export async function verifyAccessToken(
 				verifyOptions: opts.verifyOptions,
 			});
 		} catch (error) {
-			if (error instanceof Error) {
-				if (error.name === "TypeError" || error.name === "JWSInvalid") {
-					// likely an opaque token (continue)
-				} else if (error.name === "JWTExpired") {
-					throw new APIError("UNAUTHORIZED", {
-						message: "token expired",
-					});
-				} else if (error.name === "JWTInvalid") {
-					throw new APIError("UNAUTHORIZED", {
-						message: "token invalid",
-					});
-				} else {
+			if (
+				error instanceof Error &&
+				(error.name === "TypeError" || error.name === "JWSInvalid")
+			) {
+				// likely an opaque token (continue)
+			} else if (error instanceof joseErrors.JOSEError) {
+				if (JOSE_INFRA_CODES.has(error.code)) {
 					throw error;
 				}
+				throw new APIError("UNAUTHORIZED", {
+					message: "invalid access token",
+				});
+			} else if (error instanceof Error) {
+				throw error;
 			} else {
 				throw new Error(error as unknown as string);
 			}


gustavovalverde:
Closing this in favor of #9655, which has merged and closes #9654. #9655 used the tested changeset vehicle and incorporated the important behavior from this PR: missing JWT `kid` maps to unauthorized, while JWKS infrastructure failures continue to surface as raw JOSE errors. Thanks for the contribution.

PR #9709, opened 2026-05-20
fix(next-js): forward multiple set-cookie headers in nextCookies

### LINKED ISSUE #9705: nextCookies after hook silently drops all cookies — Headers.get("set-cookie") returns null for multiple Set-Cookie headers

### Reproduction

Steps to reproduce:

1. Configure auth with `nextCookies()`, `cookieCache` enabled, and `storeAccountCookie: true`:

```ts
  session: {
    cookieCache: {
      enabled: true,
      strategy: "jwe",
      maxAge: 600,
    },
  },
  account: {
    storeAccountCookie: true,
  },
  plugins: [
    genericOAuth({ ... }),
    nextCookies(), // last in the array
  ],
});
```

2. Call any `auth.api.*` method server-side that refreshes cookies (e.g. `auth.api.getSession()` or `auth.api.getAccessToken()`) from a Server Component, Server Action, or middleware.
3. Observe that no `Set-Cookie` header is sent to the browser — `session_data` and `account_data` are never updated on the client.

**Expected:** The `nextCookies()` after hook reads all `Set-Cookie` values from the response and forwards them to the browser via `cookies().set()` from `next/headers`.

**Actual:** The hook calls `returned.get("set-cookie")` which returns `null` in Node.js when there are multiple `Set-Cookie` headers (the Fetch spec explicitly [excludes `set-cookie` from comma-joining](https://fetch.spec.whatwg.org/#concept-header-list-get)). The `if (!setCookies) return` guard then exits silently — **no cookie is ever forwarded to the browser**.

The relevant code in `dist/integrations/next-js.mjs`:

```js
const setCookies = returned?.get("set-cookie"); // always null when > 1 Set-Cookie header
if (!setCookies) return;                        // always exits here
const parsed = parseSetCookieHeader(setCookies);
```

The correct API is `Headers.getSetCookie()` which returns `string[]` — one entry per cookie:

```diff
-const setCookies = returned?.get("set-cookie");
-if (!setCookies) return;
-const parsed = parseSetCookieHeader(setCookies);
+const setCookiesList = (typeof returned?.getSetCookie === "function"
+    ? returned.getSetCookie()
+    : [returned?.get("set-cookie")]
+).filter(Boolean);
+if (!setCookiesList.length) return;
+const parsed = new Map(
+    setCookiesList.flatMap((c) => Array.from(parseSetCookieHeader(c).entries()))
+);
```

**Practical impact:** any server-side auth call that produces a token refresh (e.g. after the OAuth access token expires) silently fails to update `account_data` on the browser. If the OAuth provider has refresh token rotation enabled, the browser's next `getAccessToken` call uses the now-invalidated refresh token and receives a 401.

1.6.11

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#23~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Apr 14 16:11:48 UTC 2",
    "release": "6.17.0-23-generic",
    "cpuCount": 22,
    "cpuModel": "Intel(R) Core(TM) Ultra 7 155H",
    "totalMemory": "30.81 GB",
    "freeMemory": "9.48 GB"
  },
  "node": {
    "version": "v24.15.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "11.1.2"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null
  }
}
```

Backend

```typescript
  session: {
    cookieCache: {
      enabled: true,
      strategy: "jwe",
      maxAge: 600,
      refreshCache: { updateAge: 120 },
    },
  },
  account: {
    storeAccountCookie: true,
  },
  plugins: [
    genericOAuth({ ... }),
    nextCookies(),
  ],
});
```

### Additional context

- Reproducible locally and in production (Node.js runtime, not Edge).
- The bug is present in all versions that include the `nextCookies` after hook.
- Related issues with the same symptom but no identified root cause: #2434, #4389, #7008.
- `Headers.getSetCookie()` is available since Node.js 18.14 / undici 5.x and is the [WHATWG-recommended way](https://fetch.spec.whatwg.org/#dom-headers-getsetcookie) to read multiple `Set-Cookie` values.

GautamBytes:
## Summary

- fixes `nextCookies()` so server-side auth API calls forward every returned `Set-Cookie` header
- reads cookies with `Headers.getSetCookie()` when available, with the existing split-header helper as a fallback
- adds a regression test for the case where `Headers.get("set-cookie")` is unavailable while multiple cookies are present
- includes a patch changeset for `better-auth`

## Root cause

`nextCookies()` only read `responseHeaders.get("set-cookie")`. In runtimes where multiple `Set-Cookie` headers are exposed through `getSetCookie()` instead, the hook could see no cookies and silently skip forwarding them through Next.js `cookies().set()`.

Fixes #9705.

---
## Summary by cubic
Fixes `nextCookies()` to forward all `Set-Cookie` headers from server-side auth responses in Next.js, preventing dropped multi-cookie updates (e.g., session + cache).

- **Bug Fixes**
  - Use `Headers.getSetCookie()` when available; fallback to `splitSetCookieHeader(returned.get('set-cookie') || '')`.
  - Forward each cookie to `cookies().set()` with parsed options.
  - Add a regression test for the `getSetCookie` path when `get('set-cookie')` is null, plus a patch changeset for `better-auth`.

<sup>Written for commit 2defccffd6935facee7aac3e671c0fed581a6ab3. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/9709?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/next-cookies-set-cookie-list.md
+++ b/.changeset/next-cookies-set-cookie-list.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+`nextCookies()` now forwards every `Set-Cookie` header from server-side auth API calls. Apps that refresh multiple cookies, such as session and account cache cookies, no longer drop those updates in Next.js server contexts.
--- a/packages/better-auth/src/integrations/next-js.test.ts
+++ b/packages/better-auth/src/integrations/next-js.test.ts
@@ -1,3 +1,5 @@
+import type { BetterAuthPlugin } from "@better-auth/core";
+import { createAuthEndpoint } from "@better-auth/core/api";
 import { afterEach, describe, expect, it, vi } from "vitest";
 
 describe("next-js integration", () => {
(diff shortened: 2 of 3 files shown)


bytaesu:
This might have been fixed by recent `better-call` changes. I'll look into it.

bytaesu:
Closing this in favor of #10013

PR #9734, opened 2026-05-23
fix(cimd): tolerate unsupported grant_types/response_types in metadata documents

olorin-ghzlab:
### Summary

CIMD (Client ID Metadata Document) validation rejects the **entire** client when its `grant_types` or `response_types` contains any value the server doesn't support. This breaks VS Code's MCP OAuth flow against any Better Auth server.

VS Code sends `client_id=https://vscode.dev/oauth/client-metadata.json`, whose document declares:

```json
"grant_types": ["authorization_code", "refresh_token", "urn:ietf:params:oauth:grant-type:device_code"]
```

`packages/cimd/src/validate-metadata-document.ts` validates with a strict subset check, so the `device_code` entry makes the whole client fail with:

```
{"error":"invalid_client","error_description":"grant_types must be a subset of [\"authorization_code\", \"refresh_token\"]"}
```

### Why this is the right fix

[RFC 7591 §2.1](https://datatracker.ietf.org/doc/html/rfc7591#section-2.1) and the [CIMD draft](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/) say the authorization server should **ignore** grant/response types it doesn't support, not reject the client. A client only needs to be able to use *some* of its declared grants.

### Changes

- **`validate-metadata-document.ts`**: require only that `authorization_code` (and `code` for `response_types`) is present; tolerate unsupported entries. `ALLOWED_GRANT_TYPES`/`ALLOWED_RESPONSE_TYPES` are now exported.
- **`client-store.ts`** (`toOAuthClientBody`): filter `grant_types`/`response_types` down to the supported sets before persistence, so unsupported values (e.g. `device_code`) never reach the DB. The token endpoint already gates `grant_type` independently, so this grants **no** new capability.
- Regression tests: VS Code's metadata (with `device_code`) is accepted; `["code","token"]` accepted; existing rejection cases (`client_credentials`, `token`) still reject because the required type is absent.

### Testing

`pnpm --filter @better-auth/cimd test` → **71 passed**. `@better-auth/cimd` builds clean.

### AI disclosure

Drafted with Claude per the project's AI policy; I've reviewed each diff, the spec citations, and the test output.

---
## Summary by cubic
Tolerate unsupported `grant_types` and `response_types` in CIMD so clients like VS Code are accepted. Require `authorization_code` and `code`, and strip unsupported values before persisting.

- **Bug Fixes**
  - Relaxed `@better-auth/cimd` validation to ignore unknown types per RFC 7591/CIMD; only require `authorization_code` and `code`.
  - Exported `ALLOWED_GRANT_TYPES`/`ALLOWED_RESPONSE_TYPES` and filter in `client-store.ts` so unsupported entries never hit the DB.
  - Added regression tests for VS Code’s metadata and mixed response types.

<sup>Written for commit f168f01e7aae113f079a95d798796521cb17ec90. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/9734?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/cimd-tolerate-unsupported-grant-types.md
+++ b/.changeset/cimd-tolerate-unsupported-grant-types.md
@@ -1,0 +1,26 @@
+---
+"@better-auth/cimd": patch
+---
+
+fix(cimd): tolerate unsupported grant_types/response_types in metadata documents
+
+Client ID Metadata Document validation rejected the entire client when its
+`grant_types` or `response_types` contained any value the server does not
+support. VS Code's published metadata document
+(`https://vscode.dev/oauth/client-metadata.json`) declares
+`urn:ietf:params:oauth:grant-type:device_code` alongside `authorization_code`
+and `refresh_token`, so its MCP OAuth flow failed against any better-auth
+server with:
+
+```
+{"error":"invalid_client","error_description":"grant_types must be a subset of [\"authorization_code\", \"refresh_token\"]"}
+```
+
+Per [RFC 7591 §2.1](https://datatracker.ietf.org/doc/html/rfc7591#section-2.1)
+and the [CIMD draft](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/),
+the authorization server should ignore grant/response types it does not support
+rather than rejecting the client. Validation now only requires that
+`authorization_code` (and `code` for `response_types`) is present, and the
+unsupported entries are filtered out before the client is persisted, so they
+never reach the database or grant any capability. The token endpoint already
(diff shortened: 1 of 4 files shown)


patrikduksin:
+1 on that

bytaesu (commented):
Hi,

This case was superseded by #10731 and #11010. Relaxing `response_types` remains a separate proposal and is not required for this PR reported case.

bytaesu:
Closing this PR.

Feel free to let me know if you have further questions!

PR #9736, opened 2026-05-23
fix(organization): refetch active organization on session-changing client paths

### LINKED ISSUE #9710: useActiveOrganization returns stale null after sign-in when initial active org is set via the documented session.create.before hook
## Context

The org plugin docs recommend setting an initial active organization for a user via a `databaseHooks.session.create.before` hook:

> [Set Initial Active Organization with Database Hooks](https://www.better-auth.com/docs/plugins/organization#set-the-active-organization)
>
> ```ts
> databaseHooks: {
>   session: {
>     create: {
>       before: async (session) => {
>         const organization = await getInitialOrganization(session.userId)
>         return { data: { ...session, activeOrganizationId: organization?.id } }
>       },
>     },
>   },
> }
> ```

This is the right pattern for any multi-tenant SaaS where the user is **always** working inside an organization (Linear, Vercel, Notion, Stripe, Slack, etc. all behave this way — there's no "no active workspace" state in the UI). The user is dropped straight into a workspace after sign-in, with the switcher already populated.

The server side of this works correctly: the session row in the DB is born with `activeOrganizationId` set, and `/organization/get-full-organization` returns the right org.

The client side does not pick it up.

## Reproduction

1. Use better-auth with the `organization` plugin.
2. Configure a `databaseHooks.session.create.before` hook that populates `activeOrganizationId` (per the documented recipe linked above).
3. Have a user that is a member of at least one organization.
4. From a fresh browser state, sign in via `authClient.signIn.email({ email, password })`.
5. Render a component that calls `authClient.useActiveOrganization()`.

**Observed:** `useActiveOrganization()` returns `data: null` after sign-in. The DB session row has `activeOrganizationId` correctly populated. `/organization/get-full-organization` would return the right org if called.

**Workaround:** Hard-refresh the browser. The hook then returns the correct value.

Minimal client snippet:

```tsx
function WorkspaceSwitcher() {
  const { data: org } = authClient.useActiveOrganization()
  return <div>{org?.name ?? "Select"}</div> // stays "Select" until hard refresh
}
```

**Current:** After sign-in succeeds with a session whose `activeOrganizationId` is non-null on the server, `useActiveOrganization` continues to return its initial (pre-session) cached value, typically `null`. The hook never refetches because the org plugin's `$activeOrgSignal` is only invalidated on `/organization/*` paths.

**Expected:** When the session is invalidated (sign-in/sign-up/verify-email — events the core already flips `$sessionSignal` on), `useActiveOrganization` should also refetch, so it reflects the active org of the newly-created session.

## Root cause

The core `$sessionSignal` matcher flips on every auth-state transition (`packages/better-auth/src/client/config.ts`):

```js
matcher(path) {
  return path === "/sign-out" || path === "/update-user"
      || path === "/update-session" || path === "/sign-up/email"
      || path === "/sign-in/email" || path === "/delete-user"
      || path === "/verify-email" || /* ... */;
}
```

The org plugin's `$activeOrgSignal` matcher (`packages/better-auth/src/plugins/organization/client.ts`) only flips on `/sign-out` or paths starting with `/organization/`:

```js
{
  matcher(path) {
    return path === "/sign-out" || path.startsWith("/organization");
  },
  signal: "$activeOrgSignal"
},
```

So sign-in flips `$sessionSignal` (good — `useSession` updates), but `$activeOrgSignal` stays put, so `useActiveOrganization` keeps its stale value even though the session it's keyed off has changed.

This is most visibly broken when `activeOrganizationId` is set server-side (e.g. via the documented `session.create.before` hook), because in that flow the **client never calls `/organization/set-active`** — there's no path that flips `$activeOrgSignal`.

It also breaks subtly in a second scenario regardless of design: signing out on device A and then signing in on device B with a session row that already has `activeOrganizationId` set on the server. `useSession` picks up the new session on device B but `useActiveOrganization` doesn't refetch.

## Proposed fix

Expand the org plugin's `$activeOrgSignal` matcher to also flip on session-changing paths (mirror the core `$sessionSignal` list):

```diff
 {
   matcher(path) {
-    return path === "/sign-out" || path.startsWith("/organization");
+    return (
+      path === "/sign-out" ||
+      path === "/sign-in/email" ||
+      path === "/sign-up/email" ||
+      path === "/verify-email" ||
+      path === "/update-session" ||
+      path === "/delete-user" ||
+      path.startsWith("/organization")
+    );
   },
   signal: "$activeOrgSignal"
 },
```

A similar adjustment is likely warranted for `$activeMemberSignal` and `$activeMemberRoleSignal` (see #7981).

## Workaround for anyone hitting this today

Subscribe to `$sessionSignal` and forward changes to `$activeOrgSignal` at the auth-client level:

```ts
// auth-client.ts
  plugins: [organizationClient(), /* ... */],
})

if (typeof window !== "undefined") {
  let first = true
  authClient.$store.atoms.$sessionSignal.subscribe(() => {
    if (first) {
      first = false
      return
    }
    authClient.$store.notify("$activeOrgSignal")
  })
}
```

This makes `useActiveOrganization` refetch whenever the session changes — the behavior the org plugin should arguably have by default for apps following the documented "set initial active org" pattern.

1.6.11

## System info

```json
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "release": "25.3.0"
  },
  "node": { "version": "v24.5.0", "env": "development" },
  "packageManager": { "name": "npm", "version": "11.13.0" },
  "frameworks": [{ "name": "react", "version": "^19.2.3" }],
  "databases": [
    { "name": "pg", "version": "^8.17.1" },
    { "name": "kysely", "version": "^0.28.9" }
  ],
  "betterAuth": { "version": "^1.6.11" }
}
```

Client, Documentation

## Related

- #3608 — Server-side auth actions don't trigger `useSession` updates (same root cause, different hook)
- #3837 — `setActiveOrganization` doesn't update other react hooks (sibling: `$activeOrgSignal` flips but `$sessionSignal` doesn't)
- #7981 — `useActiveMember` does not refetch when active organization changes (sibling: `$activeMemberSignal` has the same gap)
- #8910 — Replace custom `useAuthQuery` with nanoquery (the wholesale fix)

spokodev:
Closes #9710.

## Problem

The org plugin's `\$activeOrgSignal` matcher in
[`packages/better-auth/src/plugins/organization/client.ts`](packages/better-auth/src/plugins/organization/client.ts)
only flips on `/sign-out` or paths starting with `/organization`. The
core `\$sessionSignal` matcher flips on every auth-state transition
(`/sign-in/email`, `/sign-up/email`, `/verify-email`, `/sign-out`,
`/update-session`, `/delete-user`, …).

So when a new session is established with `activeOrganizationId`
populated server-side via the [documented `databaseHooks.session.create.before` recipe](https://www.better-auth.com/docs/plugins/organization#set-the-active-organization),
the client never hits `/organization/set-active`. `useActiveOrganization()`
keeps its pre-session cached value (typically `null`) until a hard
refresh. Same shape of bug also fires on sign-in on a new device when
the server-side session already has `activeOrganizationId` set.

Issue body has a more detailed walk-through of the symptom and the
root-cause analysis.

## Fix

Expand `\$activeOrgSignal`'s matcher to also flip on the session-changing
paths the core `\$sessionSignal` matcher already lists:

\`\`\`diff
 {
   matcher(path) {
-    return path === "/sign-out" || path.startsWith("/organization");
+    return (
+      path === "/sign-out" ||
+      path === "/sign-in/email" ||
+      path === "/sign-up/email" ||
+      path === "/verify-email" ||
+      path === "/update-session" ||
+      path === "/delete-user" ||
+      path.startsWith("/organization")
+    );
   },
   signal: "\$activeOrgSignal"
 }
\`\`\`

`/update-user`, `/change-email`, `/change-password`, and `/revoke-*` are
intentionally not added, since they don't open a new session with a
potentially different `activeOrganizationId`.

## Tests

[`packages/better-auth/src/plugins/organization/client.test.ts`](packages/better-auth/src/plugins/organization/client.test.ts):
two new unit tests inspired by [`packages/better-auth/src/client/proxy.test.ts`](packages/better-auth/src/client/proxy.test.ts).

- **positive** — `\$activeOrgSignal` matcher returns `true` for every
  session-change path, `/sign-out`, and existing `/organization/*`
  coverage.
- **negative** — matcher returns `false` for read-only paths
  (`/get-session`, `/list-sessions`), user-attribute updates
  (`/update-user`, `/change-password`, `/change-email`), and arbitrary
  unrelated paths, so it doesn't over-trigger refetches.

### Verification

\`\`\`
$ pnpm vitest run src/plugins/organization/client.test.ts
 Test Files  1 passed (1)
      Tests  4 passed (4)

$ pnpm vitest run src/plugins/organization/
 Test Files  9 passed (9)
      Tests  208 passed | 1 skipped | 1 todo (210)

$ pnpm typecheck
$ tsc --build

---
## Summary by cubic
Expand the `\$activeOrgSignal` matcher to refetch the active organization on all session-changing client paths. Fixes stale `useActiveOrganization()` values after sign-in or new-session events.

- **Bug Fixes**
  - Mirror core `\$sessionSignal` paths: `/sign-in/email`, `/sign-up/email`, `/verify-email`, `/update-session`, `/delete-user`, and `/sign-out`, plus existing `/organization/*`, so the active-org query updates without a hard refresh.
  - Exclude non-session paths (`/update-user`, `/change-email`, `/change-password`, `revoke-*`) to avoid extra refetches; added unit tests for positive and negative cases.

<sup>Written for commit e3c44a14a093fecd154f45363d947a8ae1ce4f10. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/9736?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/better-auth/src/plugins/organization/client.test.ts
+++ b/packages/better-auth/src/plugins/organization/client.test.ts
@@ -1,4 +1,4 @@
-import { describe, it } from "vitest";
+import { describe, expect, it } from "vitest";
 import { betterAuth } from "../../auth/full";
 import { createAuthClient } from "../../client";
 import { inferOrgAdditionalFields, organizationClient } from "./client";
@@ -68,5 +68,62 @@ describe("organization", () => {
 			//@ts-expect-error - this field is not available
 			unavailableField: "123", //this should be not allowed
 		});
+	});
+});
+
+describe("organizationClient atomListeners", () => {
+	const getActiveOrgMatcher = () => {
+		const plugin = organizationClient();
+		const listener = plugin.atomListeners?.find(
+			(l) => l.signal === "$activeOrgSignal",
+		);
+		if (!listener) throw new Error("$activeOrgSignal listener not found");
+		return listener.matcher;
+	};
+
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/9710
+	 */
+	it("$activeOrgSignal matcher flips on session-changing paths", () => {
+		const matcher = getActiveOrgMatcher();
+
+		// Session-changing paths — the new session may carry a different
+		// activeOrganizationId set server-side via the documented
+		// `databaseHooks.session.create.before` hook, so the active-org query
+		// must refetch to avoid returning a stale `null`.
+		expect(matcher("/sign-in/email")).toBe(true);
+		expect(matcher("/sign-up/email")).toBe(true);
+		expect(matcher("/verify-email")).toBe(true);
+		expect(matcher("/update-session")).toBe(true);
+		expect(matcher("/delete-user")).toBe(true);
+
+		// Existing behaviour — still flips on /sign-out and all /organization paths.
+		expect(matcher("/sign-out")).toBe(true);
+		expect(matcher("/organization/set-active")).toBe(true);
+		expect(matcher("/organization/create")).toBe(true);
+		expect(matcher("/organization/anything")).toBe(true);
+	});
+
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/9710
+	 *
+	 * Guard against over-triggering: non-session, non-organization paths must
+	 * not flip $activeOrgSignal — otherwise unrelated calls would force the
+	 * active-org query to refetch on every request.
+	 */
+	it("$activeOrgSignal matcher does not flip on unrelated paths", () => {
+		const matcher = getActiveOrgMatcher();
+
+		// Read-only session reads — must not trigger a refetch.
+		expect(matcher("/get-session")).toBe(false);
+		expect(matcher("/list-sessions")).toBe(false);
+
+		// User updates that don't change the active session/org — must not flip.
+		expect(matcher("/update-user")).toBe(false);
+		expect(matcher("/change-password")).toBe(false);
+		expect(matcher("/change-email")).toBe(false);
+
+		// Arbitrary unrelated path.
+		expect(matcher("/some/other/path")).toBe(false);
 	});
 });
--- a/packages/better-auth/src/plugins/organization/client.ts
+++ b/packages/better-auth/src/plugins/organization/client.ts
@@ -248,7 +248,15 @@ export const organizationClient = <CO extends OrganizationClientOptions>(
 			},
 			{
 				matcher(path) {
-					return path === "/sign-out" || path.startsWith("/organization");
+					return (
+						path === "/sign-out" ||
+						path === "/sign-in/email" ||
+						path === "/sign-up/email" ||
+						path === "/verify-email" ||
+						path === "/update-session" ||
+						path === "/delete-user" ||
+						path.startsWith("/organization")
+					);
 				},
 				signal: "$activeOrgSignal",
 			},


spokodev:
Quick check-in on this since 2026-05-23. Could you take a quick look when convenient? Ready to address review feedback. Thanks!

PR #9747, opened 2026-05-25
feat(admin): support multi-field list users search

### LINKED ISSUE #5990: Admin list users — multi-field search support

In the current `/admin/list-users` endpoint, the search functionality only allows searching in a *single* field using `searchField`.

When working with custom user models that include additional fields (e.g., `firstName`, `lastName`, `username`, etc.), or when trying to build richer admin UIs, it's limiting to only be able to search in one field at a time.

This causes frustration when building search bars that should match across multiple fields simultaneously (email + name, or email + firstName + lastName, etc.).

Add support for **multi-field search** through two new optional query parameters:

#### **1. `searchInFields?: string[]`**

* When provided, the admin API should search across all specified fields.
* Only applied when `searchField` is *not* set.
* Example:

  ```ts
  searchInFields: ["email", "name", "firstName"]
  ```

#### **2. `searchMode?: "OR" | "AND"`**

* Determines how `searchInFields` are combined.
* `"OR"` (default): match if *any* field contains the search value
* `"AND"`: match only if *all* fields contain the search value

#### Example request

```
/admin/list-users?searchValue=john&searchInFields=email,firstName,lastName&searchMode=OR
```

This would return users where **email OR firstName OR lastName** contains `"john"`.

#### Expected benefits

* Much more powerful admin search functionality
* Works automatically with additional fields defined using `inferAdditionalFields`
* More intuitive search experience for admin dashboards
* Backward compatible

* **Client-side filtering**: Not scalable; requires fetching all users (impossible with pagination).
* **Combining multiple requests** (one per field): Inefficient, requires deduplication and merging client-side.
* **Joining fields into a single computed DB column**: Complicated and undocumented; still doesn't solve AND vs OR logic.

None of these alternatives provide the simplicity or clean API shape of adding multi-field search support directly in the backend.

### Additional context

This feature would integrate cleanly into the existing `listUsers` API and remain fully backward-compatible.

The implementation is straightforward:

* Extend Zod schema in the admin plugin
* Build multiple `where` conditions when `searchInFields` is present
* Support OR or AND logic in the internal adapter

haoxiangliew:
Closes https://github.com/better-auth/better-auth/issues/5990

Adds `POST /admin/list-users` support for multi-field admin user searches while preserving the existing `GET` query-parameter path.

- Accepts a `search` array with adapter-style conditions so callers can search across fields like `email`, `name`, and custom user fields such as `username`.
- Keeps pagination, sorting, and filter inputs shared between query parameters and the POST body.
- Lets the admin client infer POST when body fields are passed to `admin.listUsers`, instead of forcing the route to GET.
- Fixes client input inference for endpoints that have both a body and optional query parameters.
- Updates OpenAPI generation to unwrap optional query/body schemas and refreshes the admin plugin docs.
- Adds regression coverage for client and server multi-field search, custom username search, and the updated type inference.

No breaking changes.

---
## Summary by cubic
Adds POST-based multi-field admin user search to `/admin/list-users` while keeping GET. Supports AND/OR, case modes, built-in and schema-mapped custom fields, and validates fields; query and body are optional and can be combined.

- **New Features**
  - Add POST `/admin/list-users` with `search[]` (AND/OR, case modes). Accepts string/number/boolean and array values. Supports schema-configured user field names across search/filter/sort.
  - Admin client auto-uses POST when body fields are passed to `admin.listUsers`. Docs add multi-field examples and `trustedOrigins` guidance.

- **Bug Fixes**
  - Validate search/filter/sort fields and return 400 with `VALIDATION_ERROR`. Only validate `filterField` when `filterValue` is provided.
  - Improve client input/type inference: optional query stays optional with a body; required keys enforced; catch-all query records aren’t required. OpenAPI generator unwraps optional query/body schemas, includes optional query params, and marks request bodies optional.

<sup>Written for commit 1af3b2e9005bba3394acccdfdedab3638338b0a6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9747?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ping-maxwell (changes requested):
We should validate the fields they provide against the known schema, or at minimum catching and handling database adapter errors gracefully. Technically only admins should be allowed to run this, so this is just defense in depth. Also add unit tests for this too.

packages/better-auth/src/plugins/admin/routes.ts
```diff
@@ -686,14 +788,17 @@ export const listUsers = (opts: AdminOptions) =>
...
+					offset,
 				});
+			} catch (error) {
+				if (error instanceof APIError) {
+					throw error;
+				}
+				throw APIError.from(
+					"BAD_REQUEST",
+					ADMIN_ERROR_CODES.INVALID_LIST_USERS_FIELD,
+				);
```
ping-maxwell: Since invalid fields are already prevalidated, remaining failures could be other adapter/db bugs, we shouldn't use this error code.

packages/better-auth/src/plugins/admin/routes.ts
```diff
@@ -648,32 +731,51 @@ export const listUsers = (opts: AdminOptions) =>
...
+			const allowedFields = getListUsersAllowedFields(ctx.context.options);
+
+			if (ctx.body?.search) {
+				for (const search of ctx.body.search) {
+					validateListUsersField(search.field, allowedFields);
+				}
+			}
+
+			validateListUsersField(input.filterField, allowedFields);
+			validateListUsersField(input.sortBy, allowedFields);
```
ping-maxwell: this is a small behaviour regression; this runs before `input.filterValue !== undefined`. That means a request with `filterField=notAUserField` but no `filterValue` now returns 400, even though the field is otherwise ignored.

ping-maxwell (changes requested):

PR #9758, opened 2026-05-26
fix(core): handle Twitch token refreshing: Handle arrays of scopes during token refresh

### LINKED ISSUE #9076: auth.api.getAccessToken: token refreshes fail with Twitch provider

### Reproduction

Reproduced using fresh SvelteKit project using `sv create` CLI with Drizzle and Better-Auth options selected with GitHub OAuth demo. DB started with podman compose, auth schema migrations ran according to sv CLI instructions post project creation.

Manually modified the demo to use Twitch OAuth with multiple scopes defined. Updated better-auth to latest version (1.6.2) to ensure issue still persists.

Defined a SvelteKit page and `+page.server.ts` route in `src/routes/demo/better-auth/refresh` containing the call:
```typescript
const tokens = await auth.api.getAccessToken({
    body: {
        providerId: "twitch",
        userId: String(user.id)
    }
})
```

In the database, manually expire the access token by setting the `accessTokenExpiresAt` timestamp to yesterday.

Navigate to the `/demo/better-auth/refresh` page while logged in using Twitch oauth to find that the request fails with a SvelteKit Internal Error (code 500)

Expected: A page or API route featuring an auth.api.getAccessToken() call will successfully return a new Twitch access token, even if the token must be refreshed via an internal call to Twitch's refresh endpoint

Current: auth.api.getAccessToken() will error out the page load or API route if the token must be refreshed, because the procedure to do so fails. Catching and logging the error returned from getAccessToken returns a generic
```
[APIError: Failed to get a valid access token] {
  status: 'BAD_REQUEST',
  body: {
    message: 'Failed to get a valid access token',
    code: 'FAILED_TO_GET_ACCESS_TOKEN'
  },
  headers: {},
  statusCode: 400
}
```

Returning an access token that does NOT need to be refreshed works as expected.

1.6.2

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#1 SMP PREEMPT_DYNAMIC Thu Apr  2 16:55:52 UTC 2026",
    "release": "6.19.11-200.fc43.x86_64",
    "cpuCount": 16,
    "cpuModel": "AMD Ryzen 7 7700 8-Core Processor",
    "totalMemory": "30.22 GB",
    "freeMemory": "19.03 GB"
  },
  "node": {
    "version": "v22.22.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "10.31.0"
  },
  "frameworks": [
    {
      "name": "svelte",
      "version": "^5.55.2"
    },
    {
      "name": "@sveltejs/kit",
      "version": "^2.57.0"
    }
  ],
  "databases": [
    {
      "name": "postgres",
      "version": "^3.4.9"
    },
    {
      "name": "drizzle",
      "version": "^0.45.2"
    }
  ],
  "betterAuth": {
    "version": "~1.6.2",
    "config": {
      "baseURL": "http://localhost:5173",
      "secret": "[REDACTED]",
      "emailAndPassword": {
        "enabled": true
      },
      "socialProviders": {
        "twitch": {
          "clientId": "[REDACTED]",
          "clientSecret": "[REDACTED]",
          "scope": [
            "user:read:email",
            "moderator:read:followers"
          ]
        }
      },
      "plugins": [
        {
          "name": "sveltekit-cookies",
          "config": {
            "id": "sveltekit-cookies",
            "version": "1.6.2",
            "hooks": {
              "after": [
                {}
              ]
            }
          }
        }
      ]
    }
  }
}
```

Backend

```typescript
import { betterAuth } from 'better-auth/minimal';
import { drizzleAdapter } from 'better-auth/adapters/drizzle';
import { sveltekitCookies } from 'better-auth/svelte-kit';
import { env } from '$env/dynamic/private';
import { env as envPub } from '$env/dynamic/public'
import { getRequestEvent } from '$app/server';
import { db } from '$lib/server/db';

	baseURL: env.ORIGIN,
	secret: env.BETTER_AUTH_SECRET,
	database: drizzleAdapter(db, { provider: 'pg' }),
	socialProviders: {
		twitch: {
            clientId: env.TWITCH_CLIENT_ID,
            clientSecret: env.TWITCH_CLIENT_SECRET,
            scope: ["user:read:email","moderator:read:followers"],        
        }
	},
	plugins: [
		sveltekitCookies(getRequestEvent) // make sure this is the last plugin in the array
	]
});
```

### Additional context

Through console.log()s and tracing function errors through .catch methods, I believe this issue is related to Better Auth mishandling scopes returned by Twitch's token refresh endpoint.

The error occurs in the /get-access-token endpoint defined in `packages/better-auth/src/api/routes/account.ts` attempts to call provider.refreshAccessToken() at line 560. 

Via error catching, that function call pointed to `packages/core/src/oauth2/refresh-access-token.ts` where the failure happens at the end of this block of code:
```typescript
const { data, error } = await betterFetch<{
		access_token: string;
		refresh_token?: string | undefined;
		expires_in?: number | undefined;
		refresh_token_expires_in?: number | undefined;
		token_type?: string | undefined;
		scope?: string | undefined;
		id_token?: string | undefined;
	}>(tokenEndpoint, {
		method: "POST",
		body,
		headers,
	});
	if (error) {
		throw error;
	}
	const tokens: OAuth2Tokens = {
		accessToken: data.access_token,
		refreshToken: data.refresh_token,
		tokenType: data.token_type,
		scopes: data.scope?.split(" "),
		idToken: data.id_token,
	};
```

Placing a `console.log(data.scope)` directly above the attempted assignment of `tokens` indicates the issue: The data coming back from Twitch's token refresh endpoint already formats the scopes field into a comma-delimited array, therefore the `.split(" ")` method cannot be run on it. The BetterFetch call at the top of the code block suggests expectation that scope is being returned as a string (perhaps other providers do this)

According to examples on [Twitch's documentation on refreshing access tokens](https://dev.twitch.tv/docs/authentication/refresh-tokens/), their refresh endpoint returns a string for single-scope cases and an array of strings for multi-scope cases.

I hope this information is sufficient to track down the issue.

alexlitty:
Some providers, like Twitch, are sending oauth scopes as an array and not as a space-delimited string.

We handle this correctly in the initial token grant, but it looks like the refresh logic has drifted. It still expects only a string here, so refreshing tokens fails for these providers.

Resolves #9076.

I considered either fixing it inline or calling the same `getOAuth2Tokens` parser used for the initial grant.

As much as I'd like to prevent future drift, I inlined the fix for now: Using the initial parser does change the response slightly (I see it adds `raw` at the very least; not sure how sensitive downstream would be to that).

---
## Summary by cubic
Fix OAuth2 token refresh in `@better-auth/core` to handle scope arrays, not just space-delimited strings. Restores working refresh for Twitch and similar providers.

- **Bug Fixes**
  - Accept `scope` as `string | string[]` and parse it correctly during refresh.
  - Added tests for both array and space-delimited scope responses.

<sup>Written for commit dcad35fce24df206f0d992a3b7aad870a43b3b00. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/9758?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/big-moles-thank.md
+++ b/.changeset/big-moles-thank.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/core": patch
+---
+
+Fixes token refreshing for providers sending scopes as an array. Fixes token refreshing for Twitch.
--- a/packages/core/src/oauth2/refresh-access-token.test.ts
+++ b/packages/core/src/oauth2/refresh-access-token.test.ts
@@ -87,4 +87,49 @@ describe("refreshAccessToken", () => {
 
 		expect(tokens.refreshTokenExpiresAt).toBeUndefined();
 	});
+
+	it("should handle scope as an array", async () => {
+		mockedBetterFetch.mockResolvedValueOnce({
+			data: {
+				access_token: "new-access-token",
+				refresh_token: "new-refresh-token",
+				expires_in: 3600,
+				token_type: "Bearer",
+				scope: ["channel:read:subscriptions", "channel:manage:polls"],
+			},
+			error: null,
+		});
+
+		const tokens = await refreshAccessToken({
+			refreshToken: "old-refresh-token",
+			options: { clientId: "test-client", clientSecret: "test-secret" },
+			tokenEndpoint: "https://example.com/token",
+		});
+
+		expect(tokens.scopes).toEqual([
+			"channel:read:subscriptions",
+			"channel:manage:polls",
+		]);
+	});
+
+	it("should handle scope as a space-delimited string", async () => {
+		mockedBetterFetch.mockResolvedValueOnce({
+			data: {
+				access_token: "new-access-token",
+				refresh_token: "new-refresh-token",
+				expires_in: 3600,
+				token_type: "Bearer",
+				scope: "read write profile",
+			},
+			error: null,
+		});
+
+		const tokens = await refreshAccessToken({
+			refreshToken: "old-refresh-token",
+			options: { clientId: "test-client", clientSecret: "test-secret" },
+			tokenEndpoint: "https://example.com/token",
+		});
+
+		expect(tokens.scopes).toEqual(["read", "write", "profile"]);
+	});
 });
--- a/packages/core/src/oauth2/refresh-access-token.ts
+++ b/packages/core/src/oauth2/refresh-access-token.ts
@@ -121,7 +121,7 @@ export async function refreshAccessToken({
 		expires_in?: number | undefined;
 		refresh_token_expires_in?: number | undefined;
 		token_type?: string | undefined;
-		scope?: string | undefined;
+		scope?: string | string[] | undefined;
 		id_token?: string | undefined;
 	}>(tokenEndpoint, {
 		method: "POST",
@@ -135,7 +135,7 @@ export async function refreshAccessToken({
 		accessToken: data.access_token,
 		refreshToken: data.refresh_token,
 		tokenType: data.token_type,
-		scopes: data.scope?.split(" "),
+		scopes: typeof data.scope === "string" ? data.scope.split(" ") : data.scope,
 		idToken: data.id_token,
 	};
 


gustavovalverde:
Fixed as part of #9825

PR #9766, opened 2026-05-27
feat(phone-number): add server-side OTP consumption API

### LINKED ISSUE #9754: Expose a server-side API to verify or consume phone OTP without user/session mutation

The `phone-number` plugin currently exposes `sendPhoneNumberOTP`, which can create a verification record for a phone number without requiring an existing user.

However, the matching verification flow is only available through `verifyPhoneNumber`, which is a higher-level auth lifecycle endpoint. After validating the OTP, it continues into user/session mutation:

- update an existing user's `phoneNumberVerified`
- optionally create a user via `signUpOnVerification`
- create a session unless `disableSession` is set
- call `callbackOnVerification`

This makes it difficult to compose custom signup/linking flows where the application wants to verify that a phone OTP is valid first, then decide what to do next.

For example, in a custom signup flow I may want to:

1. send an OTP using Better Auth's built-in phone OTP mechanism
2. consume/verify the OTP
3. validate the rest of the signup payload
4. other app-specific flow

Currently, if `signUpOnVerification` is disabled and no user exists for the phone number, `verifyPhoneNumber` can successfully consume the OTP but then fail later because no user exists. That means the OTP verification and the user/session mutation are tightly coupled. This coupling also makes it harder to preserve operation consistency in application-level flows.

A lower-level server-side way to consume or verify a phone OTP without mutating user or session state may solve this.

Possible API shape:

```ts
await auth.api.consumePhoneNumberOTP({
  body: {
    phoneNumber: "+8613800000000",
    code: "123456",
  },
})
```
or by adding an option to current API:

```ts
await auth.api.verifyPhoneNumber({
  body: {
    phoneNumber: "+8613800000000",
    code: "123456",
    consumeOnly: true,
  },
})
```

I considered using `verifyOTP`, but it does not solve this use case. `verifyOTP` only customizes whether the OTP is valid. It still runs inside `verifyPhoneNumber`, so the endpoint still continues into the user/session mutation flow afterward.

I also considered implementing a fully custom OTP provider outside Better Auth. That works, but it feels heavy for this use case because the phone-number plugin already provides the needed primitives for issuing OTPs, storing verification values, expiry, attempts, and rate limiting. The missing piece is the matching server-side verification/consume primitive.

Enabling `signUpOnVerification` and updating the created user afterward is possible, but it forces user creation before the full signup payload has been validated. If any later step fails, this can leave dirty state, such as an incomplete temporary user, requiring additional cleanup or rollback logic.

### Additional context

I understand that `verifyPhoneNumber` is intentionally tied to the auth lifecycle, and I am not proposing to change its current behavior.

The request is to expose a lower-level server-side primitive that matches `sendPhoneNumberOTP`:

- `sendPhoneNumberOTP`: issue OTP for a phone number
- proposed API: consume/verify OTP for a phone number
- `verifyPhoneNumber`: high-level endpoint that verifies phone number and mutates user/session

If this direction fits with Better Auth's design, I would be happy to work on a PR for this.

GautamBytes:
## Summary

- add a server-only `auth.api.consumePhoneNumberOTP` primitive for custom phone verification flows
- share phone OTP validation and consumption behavior with `verifyPhoneNumber` without user or session mutation
- document the server-only composition and concurrency contract

## Why

`sendPhoneNumberOTP` can issue a verification code without an existing user, but the matching verification API currently continues into user and session mutation. This prevents applications from consuming a phone OTP inside their own signup or account-linking flow without triggering Better Auth lifecycle state changes.

The new virtual server API verifies and consumes the OTP without registering an HTTP route, creating a session, updating a user, or invoking `callbackOnVerification`.

Closes #9754

---
## Summary by cubic
Add a server-only `auth.api.consumePhoneNumberOTP` to verify and consume phone OTPs without creating users or sessions. No HTTP route or client method is exposed.

- New Features
  - Adds `auth.api.consumePhoneNumberOTP({ body: { phoneNumber, code } })` for server use only.
  - Shares validation, expiry, attempt limits, and deletion behavior with `verifyPhoneNumber`; supports custom `verifyOTP` and removes internal records on success.
  - Hidden endpoint (no route, no `authClient` method); wrap any public usage with rate limiting.

- Refactors
  - Extracts shared verification/consumption logic used by both `consumePhoneNumberOTP` and `verifyPhoneNumber`.
  - Hardens attempt handling (normalizes corrupt counters, honors `allowedAttempts: 0`).
  - Updates docs and tests.

<sup>Written for commit ed66ff967ea9a25b154e1a33525902eefd28002f. Summary will update on new commits. <a href="https://cubic.dev/pr/better-auth/better-auth/pull/9766?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/consume-phone-number-otp.md
+++ b/.changeset/consume-phone-number-otp.md
@@ -1,0 +1,5 @@
+---
+"better-auth": minor
+---
+
+Add a server-only `auth.api.consumePhoneNumberOTP` API for custom phone OTP flows that need to verify and consume a code without creating or updating users or sessions.
--- a/docs/content/docs/plugins/phone-number.mdx
+++ b/docs/content/docs/plugins/phone-number.mdx
@@ -110,6 +110,33 @@ After the OTP is sent, users can verify their phone number by providing the code
   ```
 </APIMethod>
 
+### Consume an OTP on the Server
+
+For custom sign-up or account-linking flows, consume an OTP without creating or
+updating a user or session. This server-only API is exposed through `auth.api`;
+it does not register an HTTP route or generate an `authClient` method.
+
+```ts title="auth.ts"
+const result = await auth.api.consumePhoneNumberOTP({
+    body: {
+        phoneNumber: "+1234567890",
+        code: "123456"
+    }
+})
+```
+
+This method uses the same validation and consumption behavior as
+`verifyPhoneNumber`; it does not introduce a stronger concurrency guarantee.
+If your server flow can receive parallel redemption attempts and requires
+strict single-use acceptance, configure [`verifyOTP`](#verifyotp) with a
+provider that atomically consumes accepted codes.
+
+Call this method inside the server-side flow that uses the verification result;
+it does not return a session or reusable proof of verification. Validate any
+inputs that do not depend on successful OTP verification first. If you expose
+this capability through your own public endpoint, apply rate limiting and
+appropriate abuse protection there.
+
 ### Allow Sign-Up with Phone Number
 
 To allow users to sign up using their phone number, you can pass `signUpOnVerification` option to your plugin configuration. It requires you to pass `getTempEmail` function to generate a temporary email for the user.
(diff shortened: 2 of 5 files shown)


GautamBytes:

PR #9774, opened 2026-05-27
feat(passkey): support allowCredentials override on signIn.passkey

### LINKED ISSUE #9772: Feature request: expose `allowCredentials` override on `signIn.passkey()`

We'd like to present returning users with a passkey hint of the passkey they used last time.

# Feature request: expose `allowCredentials` override on `signIn.passkey()`

## Problem

`authClient.signIn.passkey()` currently has no way to scope the WebAuthn picker via `allowCredentials`. The plugin fetches `generate-authenticate-options` from the server, then immediately passes the result to `startAuthentication()` with no caller-side hook to modify the options between those two steps.

When a relying party already knows which credential the user should sign in with (e.g., a "Continue with Passkey" screen for a returning user where the client previously stored the credential ID via `returnWebAuthnResponse`), there is no supported way to narrow the OS credential picker. The user sees every passkey their password manager has saved for the relying party, even though the app knows exactly which one is wanted.

## Proposed solution

Add an optional `allowCredentials` parameter to `signIn.passkey()`:

```ts
await authClient.signIn.passkey({
  allowCredentials: [{ id: 'base64url-credential-id', type: 'public-key' }],
});
```

When provided, the plugin merges the value into the options returned by `generate-authenticate-options` before invoking `startAuthentication`. When omitted, behavior is unchanged.

## Why this is safe

`allowCredentials` is a client-side hint for the authenticator/OS picker only. It does not affect the security model of authentication:

- The server's `/passkey/verify-authentication` endpoint verifies the assertion signature against the stored public key for whichever credential ID the authenticator returns.
- A malicious client passing an attacker-chosen `allowCredentials` list still cannot forge a valid assertion for a credential it doesn't possess.
- The challenge is server-issued and bound to the session/cookie, so replay is prevented independently.

This is the same reasoning the WebAuthn spec uses for allowing `allowCredentials` to be set freely on the client side of a `navigator.credentials.get()` call. Pairs naturally with the existing `returnWebAuthnResponse: true` flag for "store the ID, then narrow the next sign-in to it" flows.

## Alternatives considered

- **`transformOptions` callback.** A generic `(options) => options` hook on `signIn.passkey()` would cover this case and any future need to manipulate other option fields. More flexible but with a broader surface; the direct `allowCredentials` option is simpler and covers the common case.
- **Server-side filtering via a custom `before` hook.** Possible today by patching `/passkey/generate-authenticate-options` to accept an identifier param and emit a scoped `allowCredentials` list. Heavier-weight than necessary because the filter doesn't need server help — the client already knows the credential ID. Also requires a server round-trip change for what is conceptually a UI hint.
- **Bypassing `signIn.passkey()` entirely.** What we ended up doing as a workaround (see Additional context). Re-implements the plugin's options fetch, ceremony call, verify POST, and session signal. Works but duplicates plumbing the plugin already owns.

## How it would benefit the project

This is a common need for any relying party that supports "last used method" prefill or per-account sign-in. Right now every such RP either lives with a noisy PM picker or implements the same bypass we did. A first-class option keeps consumers on the plugin's happy path and avoids divergent in-the-wild reimplementations of the ceremony.

## Compatibility / branch targeting

Strictly additive. Default behavior unchanged when the new option is not passed. No migration required. Should be eligible for `main` (patch or minor changeset) rather than `next`.

## Follow-up

Happy to open the PR once scope and API shape are agreed on. Comfortable doing the tests (unit-level `getTestInstance()` covers the happy path + omitted-option path; e2e via the existing passkey suite).

## Additional context

For reference, the bypass we ended up with — what a one-line `allowCredentials: [...]` option would replace:

```ts
const optsRes = await fetch(
  '/api/auth/passkey/generate-authenticate-options',
  { method: 'GET', credentials: 'include' },
);
const serverOptions = (await optsRes.json()) as PublicKeyCredentialRequestOptionsJSON;
const optionsJSON = {
  ...serverOptions,
  allowCredentials: [{ id: scopedCredentialId, type: 'public-key' }],
};
const assertion = await startAuthentication({ optionsJSON });
const verifyRes = await fetch(
  '/api/auth/passkey/verify-authentication',
  {
    method: 'POST',
    credentials: 'include',
    headers: { 'Content-Type': 'application/json' },
    body: JSON.stringify({ response: assertion }),
  },
);
authClient.$store.notify('$sessionSignal');
```

Note that the registration side already exposes `excludeCredentials` semantics server-side. The sign-in side simply lacks the symmetric `allowCredentials` knob.

Bypassing `signIn.passkey()` entirely:

1. Fetch `/api/auth/passkey/generate-authenticate-options` manually.
2. Mutate `allowCredentials` on the parsed JSON.
3. Call `@simplewebauthn/browser`'s `startAuthentication()` directly.
4. POST the assertion to `/api/auth/passkey/verify-authentication` manually.
5. Re-implement the session signaling that `signIn.passkey()` would have done (`$store.notify('$sessionSignal')`).

### Additional context

_No response_

albert-sway:
## What

`authClient.signIn.passkey()` accepts an optional `allowCredentials` array. When provided, the plugin overrides the server-issued `allowCredentials` on the options returned by `/passkey/generate-authenticate-options` before invoking `startAuthentication()`. When omitted, behavior is unchanged.

## Why

Lets relying parties scope the OS credential picker on "Continue with Passkey" / per-account sign-in screens where the credential ID is already known (e.g. previously captured via `returnWebAuthnResponse: true`). Closes the symmetric gap to registration-side `excludeCredentials`. Safe because `allowCredentials` is a client-side picker hint — server still verifies the assertion against stored public keys and the challenge stays session-bound.

## Tests

`packages/passkey/src/client.test.ts` — added override path + omitted-option regression. All passkey client tests pass.

## Breaking changes

None. Strictly additive.

Closes #9772

---
## Summary by cubic
Add optional `allowCredentials` to `authClient.signIn.passkey()` so apps can override the server-issued list and narrow the OS credential picker (pairs well with `returnWebAuthnResponse: true`). Omit this option for resident‑key/discoverable flows; this is a client-side hint only—server verification and the session‑bound challenge are unchanged.

<sup>Written for commit bed89c4843e78aa5bbc93dc8a60149c390c529de. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9774?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ping-maxwell (dismissed):

ping-maxwell:
> [!note]
> given it's additive change, it can stay in `main` not `next`

ping-maxwell (approved):

gustavovalverde:
@ping-maxwell I'll make some changes before merging

ping-maxwell:
> @ping-maxwell I'll make some changes before merging

sounds good!

PR #9787, opened 2026-05-28
fix(client): preserve null in useSession().data type with throw:true

### LINKED ISSUE #9781: `useSession().value.data` typed as non-nullable when `fetchOptions.throw = true`

### Reproduction

Minimal repro repo: https://github.com/NinjaKinshasa/better-auth-usesession-typing-repro

```ts

const client = createAuthClient({ fetchOptions: { throw: true } });
const session = client.useSession();

console.log(session.value.data);  // → null
console.log(session.value.error); // → null

// But TS only typed one of them correctly:
//   session.value.error : BetterFetchError | null    ✓
//   session.value.data  : { user; session }          ✗ missing `| null`

// @ts-expect-error — should accept null (data IS null right now)
const _: typeof session.value.data = null;
```

**Current:** with `fetchOptions: { throw: true }`, `useSession().value.data` is typed as `SessionData` (no `| null`). But the underlying atom (`dist/client/query.mjs`, `useAuthQuery`) initialises with `data: null` and resets to `null` on 401 — runtime is `T | null`.

**Expected:** `data` should always be `SessionData | null`, regardless of `fetchOptions.throw`.

The conditional at `dist/client/vue/index.d.mts:19-31` (and React/Svelte/Solid/vanilla equivalents) drops the `null` branch — see Fix.

1.6.11

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.6.0: Mon Jul 14 11:28:30 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6030",
    "release": "24.6.0",
    "cpuCount": 12,
    "cpuModel": "Apple M3 Pro",
    "totalMemory": "36.00 GB",
    "freeMemory": "0.23 GB"
  },
  "node": {
    "version": "v22.15.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.2"
  },
  "frameworks": [
    {
      "name": "vue",
      "version": "^3.5.0"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "1.6.11",
    "config": null
  }
}
```

Client

```typescript

```

## Fix

Decouple `useSession`'s `data` type from `getSession`'s return — always union with `null`:

```diff
--- a/dist/client/vue/index.d.mts
+++ b/dist/client/vue/index.d.mts
@@ -16,7 +16,7 @@
 declare function createAuthClient<Option extends BetterAuthClientOptions>(options?: Option | undefined): UnionToIntersection<InferResolvedHooks<Option>> & InferClientAPI<Option> & InferActions<Option> & {
   useSession: {
     (): DeepReadonly<Ref<{
-      data: InferClientAPI<Option> extends {
+      data: (InferClientAPI<Option> extends {
         getSession: () => Promise<infer Res>;
       } ? Res extends {
         data: null;
@@ -28,7 +28,7 @@
       } | {
         data: infer S;
         error: null;
-      } ? S : Res extends Record<string, any> ? Res : never : never;
+      } ? S : Res extends Record<string, any> ? Res : never : never) | null;
       isPending: boolean;
```

ping-maxwell:
Fixes useSession().data being typed as non-nullable (SessionData) instead of SessionData | null when fetchOptions: { throw: true } is set on Vue, Svelte, Solid, and Vanilla clients.

closes https://github.com/better-auth/better-auth/issues/9781


packages/better-auth/src/client/client.test.ts
bytaesu, line 1: How about wrapping these in a single `describe` block and grouping them under one `@see` reference?

.changeset/seven-sides-report.md
bytaesu, line 1: > https://github.com/better-auth/better-auth/blob/main/CONTRIBUTING.md#changesets

Since this description will be in the changelog, I think it's better to describe the actual user-facing behavior change rather than the internal implementation details.

bytaesu: Something like:

"~~ was previously behaving incorrectly as ~~, and now it behaves as ~~." 🤔

hartbit:
@ping-maxwell Do you have an update on this fix? Thanks in advance 🙇

ping-maxwell:
> @ping-maxwell Do you have an update on this fix? Thanks in advance 🙇

Will update!

PR #9790, opened 2026-05-28
chore: exclude `@better-auth/*`, `@better-fetch/fetch` and `better-call` from minimumReleaseAge

bytaesu:


bytaesu:
I'll add `better-call` and `@better-fetch/fetch` as well

PR #9793, opened 2026-05-29
feat(oauth-provider): add getSubject hook for per-(user, workspace) OIDC sub

Paola3stefania:
## Summary

Adds an optional `getSubject` hook to the `oauth-provider` plugin so a multi-tenant
issuer can return a per-`(user, workspace)` OIDC `sub` while keeping a single login
and session per human.
- **Default unchanged:** with no hook, `sub` is the raw `user.id` (existing tokens
  stay byte-for-byte identical).
- **Presentation layer only:** when set, `getSubject` rewrites the `sub` on the
  id token, `/userinfo`, and `/introspect`. The access token's *internal* `sub`
  stays the raw `user.id` because `/userinfo` uses it as the DB lookup key.
- **Composes with pairwise:** `getSubject` provides the *base* subject; pairwise
  hashing still applies on top when the client opts in.
- **Stateless JWT access tokens:** the already-resolved subject (same value the
  client gets as the id token `sub`) is embedded as an internal, URI-namespaced
  claim and consumed + stripped at `/userinfo` and `/introspect`. The raw
  `referenceId` never leaves the server.


Paola3stefania:
COmments on bot reviews: That grant persists no access-token or refresh-token rows, so affectedClientIds is empty and prepareBackchannelLogoutPlan returns null before any subject is resolved — zero Logout Tokens are dispatched, so no incorrect subject is ever sent

PR #9802, opened 2026-05-30
fix(admin): guard unbanUser and setRole against missing users

### LINKED ISSUE #9800: admin.unbanUser leaks Prisma P2025 and returns HTTP 500 for unknown user ids

### Reproduction

Steps to reproduce:

1. Configure Better Auth with:
   - Prisma adapter
   - admin plugin
   - email/password enabled

2. Create and authenticate an admin user.

3. Call the admin endpoint with a nonexistent user id:

```http
POST /api/auth/admin/unban-user
Content-Type: application/json

{
  "userId": "nonexistent-user-id"
}
```

4. Observe the server response and logs.

Minimal E2E reproduction:
```ts
const response = await request(httpServer)
  .post('/api/auth/admin/unban-user')
  .set('Cookie', adminCookie)
  .send({ userId: 'nonexistent-user-id' })

console.log(response.statusCode)
```

Current behavior:

The endpoint returns:

```http
500 Internal Server Error
```

and the server logs show an uncaught Prisma error:
```
PrismaClientKnownRequestError:
Invalid `prisma.user.update()` invocation:

An operation failed because it depends on one or more records that were required but not found.

code: 'P2025'
```

The stack trace points to the Prisma adapter and the Better Auth admin plugin route implementation.

Expected behavior:

Better Auth should intercept Prisma P2025 errors generated by nonexistent records and convert them into a controlled application-level response such as:

```http
404 Not Found
```
or
```http
400 Bad Request
```

instead of leaking the raw database exception as HTTP 500.

1.6.11

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "cpuCount": 12,
    "cpuModel": "12th Gen Intel(R) Core(TM) i5-12450H",
    "totalMemory": "7.76 GB",
    "freeMemory": "1.93 GB"
  },
  "node": {
    "version": "v24.14.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.9.0"
  },
  "frameworks": [
    {
      "name": "fastify",
      "version": "^5.8.5"
    }
  ],
  "databases": [
    {
      "name": "@prisma/client",
      "version": "^7.8.0"
    }
  ],
  "betterAuth": {
    "version": "^1.6.11",
    "config": null
  }
}
```

Backend, Package

```typescript
// relevant config below:
import { admin } from 'better-auth/plugins'
import { prismaAdapter } from '@better-auth/prisma-adapter'
import { PrismaClient } from "@/generated/prisma/client";

    provider: 'sqlite',
  }),
  emailAndPassword: {
    enabled: true,
  },
  plugins: [admin()],
})
```

### Additional context

I reproduced this consistently in local E2E tests using NestJS + Fastify + Prisma 7.

The issue appears to happen because the Prisma adapter allows `P2025` exceptions to escape directly from internal `update()` operations used by the admin plugin.

This may also affect other admin mutation endpoints that internally rely on Prisma `update()` without existence guards.

I also verified that this is not caused by malformed input:
- authentication succeeds
- authorization succeeds
- request body shape is valid

The failure only occurs when the target user does not exist.

spokodev:
Closes #9800.

## What was wrong

`unbanUser` and `setRole` in `packages/better-auth/src/plugins/admin/routes.ts` called `internalAdapter.updateUser` without checking that the target user existed. When the caller passed an unknown id the underlying database error (for example Prisma's `P2025`) bubbled all the way up to the HTTP layer as a generic 500 with a raw Prisma stack in the logs.

The sibling `banUser` endpoint already does this check (it was added in #4649), so the three admin user-update endpoints were inconsistent.

## The fix

Mirror the `banUser` guard in both endpoints: look the user up via `findUserById` first, and throw a clean `NOT_FOUND` with `BASE_ERROR_CODES.USER_NOT_FOUND` when no row is returned.

```ts
const foundUser = await ctx.context.internalAdapter.findUserById(
  ctx.body.userId,
);

if (!foundUser) {
  throw APIError.from("NOT_FOUND", BASE_ERROR_CODES.USER_NOT_FOUND);
}
```

For `setRole` the guard runs after the existing role-validation step so that invalid role inputs still surface as `BAD_REQUEST` rather than `NOT_FOUND`. For `unbanUser` the guard is the first thing after the permission check, matching `banUser` exactly.

## Tests

Added three regression tests under `packages/better-auth/src/plugins/admin/admin.test.ts`, each linked via `@see` to #9800:

- `unbanUser` returns 404 with `USER_NOT_FOUND` for a nonexistent user id (was 500 on `main`)
- `setRole` returns 404 with `USER_NOT_FOUND` for a nonexistent user id (was 500 on `main`)
- `banUser` returns 404 with `USER_NOT_FOUND` for a nonexistent user id (was already passing; included so the three endpoints stay in lockstep)

Verification:

- `pnpm vitest run src/plugins/admin/admin.test.ts`: 77 / 77 passing (74 existing + 3 new)
- RED baseline: reverting `routes.ts` to `main` makes the two new tests fail with 500 / Prisma `P2025`. Applying the fix turns them green; the `banUser` test stays green on both, which is the regression contract.
- Lefthook ran biome + spell on commit; both clean
- Changeset added at `.changeset/admin-routes-user-not-found-guards.md` (`better-auth`: patch)

## Out of scope

Other admin endpoints that read a single user (`getUser`, `listUserSessions`, `removeUser`, etc.) were not touched. `getUser` already has the same guard. The rest fall into either delete-style flows (which can be argued either way) or list-style flows (where 200 + empty is the right shape), so this PR is intentionally limited to the three endpoints that update a user record.

---
## Summary by cubic
Fixes `unbanUser` and `setRole` in `better-auth` to return a clean 404 (`USER_NOT_FOUND`) when the user id doesn’t exist, instead of a 500 from the DB. Aligns these endpoints with `banUser`. Closes #9800.

- **Bug Fixes**
  - Check existence with findUserById before update; throw NOT_FOUND (`USER_NOT_FOUND`) when missing.
  - In setRole, run the guard after role validation to keep invalid roles as BAD_REQUEST.
  - Added regression tests for `unbanUser`, `setRole`, and `banUser`.

<sup>Written for commit 145642fa6ff1977d6426455def1349f8cce42304. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9802?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/admin-routes-user-not-found-guards.md
+++ b/.changeset/admin-routes-user-not-found-guards.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+The admin plugin's `unbanUser` and `setRole` endpoints used to call `internalAdapter.updateUser` without checking that the target user existed, so when the caller passed an unknown id the underlying database error (for example Prisma's `P2025`) bubbled up as a generic HTTP 500. Both endpoints now mirror the existing guard in `banUser`: look the user up via `findUserById`, and throw a clean `NOT_FOUND` (`USER_NOT_FOUND`) when no row is returned. Closes #9800.
--- a/packages/better-auth/src/plugins/admin/admin.test.ts
+++ b/packages/better-auth/src/plugins/admin/admin.test.ts
@@ -707,6 +707,58 @@ describe("Admin plugin", async () => {
 		expect(res.error?.status).toBe(403);
 	});
 
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/9800
+	 */
+	it("should return 404 from unbanUser when the target user does not exist", async () => {
+		const res = await client.admin.unbanUser(
+			{
+				userId: "nonexistent-user-id",
+			},
+			{
+				headers: adminHeaders,
+			},
+		);
+		expect(res.error?.status).toBe(404);
+		expect(res.error?.code).toBe("USER_NOT_FOUND");
+	});
+
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/9800
+	 */
+	it("should return 404 from setRole when the target user does not exist", async () => {
+		const res = await client.admin.setRole(
+			{
+				userId: "nonexistent-user-id",
+				role: "admin",
+			},
+			{
+				headers: adminHeaders,
+			},
+		);
+		expect(res.error?.status).toBe(404);
+		expect(res.error?.code).toBe("USER_NOT_FOUND");
+	});
+
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/9800
+	 *
+	 * Regression coverage for the existing banUser guard so the three admin
+	 * endpoints (banUser, unbanUser, setRole) stay consistent.
+	 */
+	it("should return 404 from banUser when the target user does not exist", async () => {
+		const res = await client.admin.banUser(
+			{
+				userId: "nonexistent-user-id",
+			},
+			{
+				headers: adminHeaders,
+			},
+		);
+		expect(res.error?.status).toBe(404);
+		expect(res.error?.code).toBe("USER_NOT_FOUND");
+	});
+
 	it("should allow admin to list user sessions", async () => {
 		const res = await client.admin.listUserSessions(
 			{
--- a/packages/better-auth/src/plugins/admin/routes.ts
+++ b/packages/better-auth/src/plugins/admin/routes.ts
@@ -153,6 +153,15 @@ export const setRole = <O extends AdminOptions>(opts: O) =>
 					}
 				}
 			}
+
+			const foundUser = await ctx.context.internalAdapter.findUserById(
+				ctx.body.userId,
+			);
+
+			if (!foundUser) {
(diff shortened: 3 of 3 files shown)


JamBalaya56562:
Thanks for the fix. #9875 was merged and handles the same case (returning `USER_NOT_FOUND` before update for `unbanUser`/`setRole`, plus `adminUpdateUser`), closing #9800. This one looks superseded and can be closed.

spokodev:
Thanks for the heads up. #9875 covers the same case with broader scope (adminUpdateUser too). Closing this in favor of the merged fix.

PR #9806, opened 2026-05-30
fix(cookies): prefer `__Secure-` cookie in getSessionCookie

bytaesu:


packages/better-auth/src/cookies/index.ts
```diff
@@ -443,12 +443,13 @@ export const getSessionCookie = (
...
+	// Prefer __Secure- (HTTPS-only) over a non-secure leftover.
 	const getCookie = (name: string) =>
-		parsedCookie.get(name) ||
-		parsedCookie.get(`${SECURE_COOKIE_PREFIX}${name}`);
+		parsedCookie.get(`${SECURE_COOKIE_PREFIX}${name}`) ??
+		parsedCookie.get(name);
 
 	const sessionToken =
-		getCookie(`${cookiePrefix}.${cookieName}`) ||
+		getCookie(`${cookiePrefix}.${cookieName}`) ??
```
bytaesu:
```suggestion
		getCookie(`${cookiePrefix}.${cookieName}`) ||
```

PR #9107, opened 2026-04-10
fix(cli): add `$app/environment` stub for SvelteKit CLI compatibility

bytaesu:
Related to https://github.com/better-auth/better-auth/issues/8933#issuecomment-4225087144

Issue #8933, opened 2026-04-03
[Bug] auth generate error in sveltekit app


bennajah:
### Is this suited for github?

- [x] Yes, this is suited for github

### Reproduction

1. clone this repo https://github.com/bennajah/svelte-better-auth
2. run auth generate
3. see the error

### Current vs. Expected behavior

getting this error
```zsh
bennajah@MacBook-Pro my-app % bun auth:schema     
$ auth generate --config src/lib/server/auth.ts --output src/lib/server/db/auth.schema.ts --yes
Couldn't read your auth config. Error: Cannot find module '/Users/bennajah/Development/projects/src/lib/server/db'
Require stack:
- /Users/bennajah/Development/projects/my-app/src/lib/server/auth.ts
    at Module._resolveFilename (node:internal/modules/cjs/loader:1456:15)
    at require.resolve (node:internal/modules/helpers:163:19)
    at jitiResolve (/Users/bennajah/Development/projects/my-app/node_modules/jiti/dist/jiti.cjs:1:148703)
    at jitiRequire (/Users/bennajah/Development/projects/my-app/node_modules/jiti/dist/jiti.cjs:1:150290)
    at import (/Users/bennajah/Development/projects/my-app/node_modules/jiti/dist/jiti.cjs:1:158307)
    at /Users/bennajah/Development/projects/my-app/src/lib/server/auth.ts:6:17
    at async Function.import (/Users/bennajah/Development/projects/my-app/node_modules/jiti/dist/jiti.cjs:1:158301)
    at async resolveConfig (file:///Users/bennajah/Development/projects/my-app/node_modules/c12/dist/index.mjs:318:20)
    at async loadConfig (file:///Users/bennajah/Development/projects/my-app/node_modules/c12/dist/index.mjs:147:22)
    at async getConfig (file:///Users/bennajah/Development/projects/my-app/node_modules/auth/dist/index.mjs:1591:23) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/bennajah/Development/projects/my-app/src/lib/server/auth.ts'
  ]
}
error: script "auth:schema" exited with code 1
```

### What version of Better Auth are you using?

1.5.6

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.4.0: Thu Mar 19 19:33:43 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T8142",
[...]

PR #9834, opened 2026-05-31
fix(cli): resolve SvelteKit, Vite asset, and Cloudflare virtual-module imports

bytaesu:
> [!NOTE]
> Mirroring framework virtual modules outside their bundler can't be perfect. Only running the real bundler (Vite) would be exact. This PR improves the existing mirror rather than replacing that approach. 

Fixes the remaining cases in #8933
Supersedes #9107


bytaesu:
Hey @DarthGigi, sorry for the delay!! Could you test this version in your environment? 🧐

```bash
npx https://pkg.pr.new/auth@9834
```

DarthGigi:
> Hey @DarthGigi, sorry for the delay!! Could you test this version in your environment? 🧐
> 
> ```shell
> npx https://pkg.pr.new/auth@9834
> ```

Hey @bytaesu , no problem at all! I just tested it and it works perfectly, the CLI no longer crashes and generates the schema successfully. Thanks so much!

bytaesu:
@DarthGigi Thanks for letting me know 😁

I'll review this once more and get it into the next release!

DarthGigi:
@bytaesu Hey, the svelte team recently posted this PR on Reddit asking users for their opinion: https://github.com/sveltejs/kit/pull/15934

I think the important parts for the better-auth cli to keep working are these:

> With those variables defined, we can import them via two new modules — `$app/env/private` and `$app/env/public`. As with existing server-only modules, `$app/env/privat` cannot be imported into code that can run in the client.

> The $app/environment module is aliased as `$app/env`, to make everything feel a bit more cohesive. The goal is to remove `$app/environment` and the four `$env/*` modules (and the associated configuration) in SvelteKit 3.

bytaesu:
@DarthGigi I'm merging this with your suggestion applied. Thanks for letting me know.

PR #9835, opened 2026-05-31
refactor(core): prevent accidental HTTP exposure of server-only endpoints

gustavovalverde:
Several endpoints are callable only through `auth.api.*` from trusted server code and were never meant to be HTTP routes. The organization `addMember`, two-factor backup-code retrieval, and JWT signing are among them. These endpoints signaled server-only status only by omitting the path from the definition. That signal is invisible in the source and has no margin for error. Adding a path string, through a refactor or a copy-paste, would silently turn one of them into a public, unauthenticated route.

This change makes server-only status explicit and enforced. A new `createAuthEndpoint.serverOnly` helper sets `metadata.SERVER_ONLY`, so better-call's router skips the endpoint whether or not it carries a path. The two checks together are defense in depth: an accidentally added path no longer exposes the endpoint. All 10 path-less endpoints now use the helper, and a test fails the build if any server-only endpoint becomes reachable over HTTP or loses its marker.

Behavior is unchanged. These endpoints were already off the HTTP router, so the OpenAPI schema and the existing tests are untouched, and `auth.api.*` callers see no difference. The `addMember` doc comment that read like an HTTP route now states the endpoint is server-only.

---
## Summary by cubic
Make server-only endpoints explicit and safe against accidental HTTP exposure. Adds `createAuthEndpoint.serverOnly`, migrates all server-only endpoints, and validates behavior with real-router tests; runtime behavior is unchanged.

- **Refactors**
  - Added `createAuthEndpoint.serverOnly` to set `metadata.SERVER_ONLY` so the router skips an endpoint even if a path is later added.
  - Migrated server-only endpoints: organization `addMember`; two-factor `viewBackupCodes`; TOTP `generateTOTP`; JWT `signJWT`/`verifyJWT`; email-OTP `createVerificationOTP`/`getVerificationOTP`; API-key `verifyApiKey`/`deleteAllExpiredApiKeys`; `setPassword`.
  - Reworked tests to probe the real router: path-less endpoints have no route; a pathed endpoint marked `SERVER_ONLY` returns 404; `POST /organization/add-member` and `POST /two-factor/view-backup-codes` are unrouted (404) while `remove-member` still returns 401; removed a dead advisory reference.

<sup>Written for commit 41aa9951036cc798fa0d6647b848a66e0a409cf1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9835?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/better-auth/src/api/server-only-endpoints.test.ts
```diff
@@ -0,0 +1,137 @@
...
+type RegisteredEndpoint = {
+	path?: string;
+	options?: {
+		method?: string | string[];
+		metadata?: { SERVER_ONLY?: boolean };
+	};
+};
+
+/** Mirrors better-call's router gate (`!endpoint.path || SERVER_ONLY` ⇒ skip). */
+function isHttpReachable(endpoint: RegisteredEndpoint): boolean {
```
bytaesu: It seems like we should test whether it's actually reachable, rather than mirroring it.

.changeset/server-only-endpoint-hardening.md
bytaesu, line 1: Changesets should be based on user-facing changes, not internal implementation refactors.

packages/better-auth/src/api/server-only-endpoints.test.ts
bytaesu: - 🤔 For example: #9843

.changeset/server-only-endpoint-hardening.md
gustavovalverde: Do you mean with this that I should remove the changeset?

Bekacru:
This is a good one. We should merge it

bytaesu:
- https://github.com/better-auth/better-auth/issues/9994

PR #9842, opened 2026-05-31
fix(organization): allow null logo on create and update

### LINKED ISSUE #9829: Allow updateOrganization to clear logo with null

Yes. This is an API/schema mismatch in the organization plugin.

### To Reproduce

Call `auth.api.updateOrganization` and try to clear an organization logo:

```ts
await auth.api.updateOrganization({
  body: {
    organizationId,
    data: {
      logo: null,
    },
  },
  headers,
});
```

The API type for `data.logo` is currently `string | undefined`, and the endpoint schema uses `z.string().optional()`, so `null` cannot be passed through the API. Passing `undefined` is not equivalent because it leaves the field unset/unchanged in adapters such as Drizzle. Passing an empty string is also not equivalent to clearing a nullable DB column.

### Current behavior

`updateOrganization` accepts `logo?: string`, which allows setting a logo but does not provide a clean way to clear it back to `null`.

### Expected behavior

`updateOrganization` should allow `logo: null` for nullable organization logos, matching the organization schema where `logo` is nullable/nullish.

For example, the update body could allow:

```ts
data: {
  logo?: string | null
}
```

and persist `null` so callers can clear an organization logo without bypassing Better Auth.

### Additional context

In our app we currently have to call `updateOrganization` for permission checks/name updates, then manually update the organization row with `logo: input.logo || null` to support clearing the logo. It would be cleaner and safer if this could go through `updateOrganization` directly.

### Better Auth Version

latest

### Environment

- Runtime: Bun
- Database adapter: Drizzle

bytaesu:
- Closes https://github.com/better-auth/better-auth/issues/9829
- Companion PR https://github.com/better-auth/better-auth/pull/9841

### DIFF
--- a/.changeset/org-update-clear-logo.md
+++ b/.changeset/org-update-clear-logo.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+You can now clear an organization's logo by passing `logo: null` to `createOrganization` and `updateOrganization`. Previously only a string was accepted, so an existing logo could not be removed.
--- a/packages/better-auth/src/plugins/organization/organization.test.ts
+++ b/packages/better-auth/src/plugins/organization/organization.test.ts
@@ -2611,7 +2611,7 @@ describe("Additional Fields", async () => {
 		expectTypeOf<Params>().toEqualTypeOf<{
 			name: string;
 			slug: string;
-			logo?: string | undefined;
+			logo?: string | null | undefined;
 			userId?: string | undefined;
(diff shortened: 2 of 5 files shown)


bytaesu:
This is a PR for the org plugin!

cc @ping-maxwell

ping-maxwell:
let's merge both this & companion first, then I'll apply to my org rewrite.

ping-maxwell (approved):

Issue #9767, opened 2026-05-27
Module not found: Can't resolve '@better-auth/core/utils/string'


renekahr:
### Is this suited for github?

- [x] Yes, this is suited for github

### Reproduction

installing 1.6.11 doesn't install anything inside /core/utils???

"better-auth": "1.6.11"
"@better-auth/cli": "1.4.21"

<img width="1724" height="888" alt="Image" src="https://github.com/user-attachments/assets/59c81e0a-c091-4022-81ac-b3b9602493aa" />

### Current vs. Expected behavior

.

### What version of Better Auth are you using?

1.6.11

### Which area(s) are affected? (Select all that apply)

Package

bytaesu:
> [!NOTE]
> https://better-auth.com/docs/concepts/cli 

Hi @renekahr,

Could you try using `auth` instead of `@better-auth/cli`? 😁

ping-maxwell:
Closing as stale.

PR #9848, opened 2026-05-31
fix(kysely-adapter): drop fragile `@better-auth/core/utils/string` subpath import

ping-maxwell:
Fixes [#9767](http://github.com/better-auth/better-auth/issues/9767).

The kysely adapter's compiled bundle imports `@better-auth/core/utils/string`. Projects that keep a stale `@better-auth/cli@1.4.x` (the last version published under the old name, before it was renamed to auth) end up installing two `@better-auth/core`versions side by side:

`@better-auth/core@1.6.x`: the version this adapter is built against
`@better-auth/core@1.4.x`: pinned as a regular dependency by the legacy `@better-auth/cli@1.4.x`
npm (and pnpm with node-linker=hoisted) hoist the older `@better-auth/core@1.4.x` to the top of node_modules. Resolving `@better-auth/core` from inside the kysely adapter (a peer-dep consumer) walks up the tree and finds that older core first. Its exports map only declares `./utils` - the `./utils/*` wildcard was added in 1.5+

### DIFF
--- a/.changeset/kysely-adapter-inline-capitalize.md
+++ b/.changeset/kysely-adapter-inline-capitalize.md
@@ -1,0 +1,9 @@
+---
+"@better-auth/kysely-adapter": patch
+---
+
+Drop the `@better-auth/core/utils/string` subpath import from the kysely adapter and inline the tiny `capitalizeFirstLetter` helper instead.
+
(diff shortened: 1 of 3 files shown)


bytaesu (changes requested):
I don't think this is the right solution. It looks like the user was still using the legacy CLI. Changing our internal implementation to work around that seems odd.

PR #9849, opened 2026-05-31
refactor: unify AuthorizeResponse to { error: string | null }

### LINKED ISSUE #9675: Unify `AuthorizeResponse` return shape in access plugin
## Parent

Part of #9674

## Problem

The `AuthorizeResponse` type in `packages/better-auth/src/plugins/access/access.ts` currently uses a discriminated union with different shapes for success and failure:

```typescript
export type AuthorizeResponse =
  | { success: false; error: string }
  | { success: true; error?: never | undefined };
```

This means the success and failure cases have inconsistent shapes. Consumers must check `success` before accessing `error`, and the two branches carry different fields.

## Proposed solution

Model `error` as `string | null`:

```typescript
export type AuthorizeResponse = { error: string | null };
```

With this shape, `success` becomes redundant — a `null` error means success, and a non-null error means failure. This simplifies both the internal return sites in `role()` and all external consumers.

### Considerations

- This is a **breaking change** for any code that checks `result.success`. Migration path should be documented.
- All return sites in `role()` need to be updated.
- All consumers of `authorize()` across the codebase and plugins need to be audited and updated.

## Origin

https://github.com/better-auth/better-auth/pull/9603#discussion_r3230449824

parshipcy:
## Summary
- Fixes #9675 
- `role().authorize()` used to return different shapes depending on pass or fail: success had `{ success: true }`, failure had `{ success: false, error: "..." }`. That made the type awkward, we had to check success before we could safely use error. So, now it always returns the same shape: `{ error: string | null }`. If `error` is `null`, the check passed. If `error` is a string, it failed and that string explains why.

## What changed
- Updated `AuthorizeResponse` and all return paths in the access plugin
- Updated internal callers in organization, admin, and api-key plugins
- Updated access plugin tests

## Breaking change
If we call `role().authorize()` directly and use `result.success`, switch to `result.error === null` (allowed) or `result.error !== null` (denied).
Not affected: `hasPermission`, `userHasPermission`, and `checkRolePermission` - those still use `{ success: boolean }` or a plain boolean.

## Testing
- `access.test.ts` - direct `authorize()` behavior
- Rebuild `better-auth`, then vitest on admin/org/api-key permission suites
- Smoke typecheck fixtures (org + admin plugins compile)

---
## Summary by cubic
Unifies the access check result to a single shape: `AuthorizeResponse` now returns `{ error: string | null }` instead of a success/failure union in `better-auth`. This simplifies checks and removes the need to branch on `success`.

- **Migration**
  - Replace `result.success` checks with `result.error === null` (allowed) or `result.error !== null` (denied).
  - No changes needed if you use helpers like `hasPermission`, `userHasPermission`, or `checkRolePermission` (their public return shapes are unchanged).

<sup>Written for commit 9536860b7408ddabf43a870789a997bb3abe62d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9849?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/better-auth/src/plugins/access/access.ts
+++ b/packages/better-auth/src/plugins/access/access.ts
@@ -7,9 +7,7 @@ import type {
 	Statements,
 } from "./types";
 
-export type AuthorizeResponse =
-	| { success: false; error: string }
-	| { success: true; error?: never | undefined };
+export type AuthorizeResponse = { error: string | null };
 
 export function role<
 	const TRoleStatements extends Statements,
@@ -30,7 +28,6 @@ export function role<
 				if (!allowedActions) {
 					if (connector === "AND") {
 						return {
-							success: false,
 							error: `You are not allowed to access resource: ${requestedResource}`,
 						};
 					}
@@ -68,22 +65,18 @@ export function role<
 					}
 				}
 				if (success && connector === "OR") {
-					return { success };
+					return { error: null };
 				}
 				if (!success && connector === "AND") {
 					return {
-						success: false,
 						error: `unauthorized to access resource "${requestedResource}"`,
 					};
 				}
 			}
 			if (success) {
-				return {
-					success,
-				};
+				return { error: null };
 			}
 			return {
-				success: false,
 				error: "Not authorized",
 			};
 		},
--- a/.changeset/unify-authorize-response.md
+++ b/.changeset/unify-authorize-response.md
@@ -1,0 +1,5 @@
+---
+"better-auth": minor
+---
+
+`AuthorizeResponse` from `role().authorize()` is now `{ error: string | null }` instead of a discriminated union with `success`. A `null` error means authorized; a non-null error means denied. Replace `result.success` checks with `result.error === null` (or `result.error !== null` for failures).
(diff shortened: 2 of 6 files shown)


Bekacru (approved):

Bekacru:
cc @gustavovalverde this looks good to me

packages/better-auth/src/plugins/access/access.ts
```diff
@@ -7,9 +7,7 @@ import type {
 	Statements,
 } from "./types";
 
-export type AuthorizeResponse =
-	| { success: false; error: string }
-	| { success: true; error?: never | undefined };
```
bytaesu, line 12: We have a few patterns like this in the codebase, but they feel like an anti-pattern to me. I'd prefer to standardize on shapes like `{ data, error }`, `{ error }`, or `void`.

If we're using HTTP status codes and error handling correctly, an explicit `success` return value doesn't really add any meaningful signal, since success is already implied by the absence of an error.

bytaesu: - Related task: https://github.com/better-auth/better-auth/issues/9674

PR #9851, opened 2026-05-31
fix(next-js): restore RSC detection when Next.js middleware is present

### LINKED ISSUE #9776: nextCookies before-hook can't detect RSC context when Next.js middleware is present (routing headers stripped)

### Reproduction

**Background**

Following up on Bekacru's guidance in issue #2115 (*"most likely due to the server not being able to set the cookie... if you're only calling `auth.api.getSession` on react server component or if you're not using `nextCookies` plugin"*) — adding `nextCookies()` did resolve our main symptom of session cookie not refreshing in an RSC-heavy App Router app. Server-Action-triggered refreshes now propagate the cookie correctly.

While instrumenting the plugin's `before` hook to verify the RSC-skip optimization was engaging on read-only navigation, we found a separate, narrower issue: the hook's RSC detection silently fails when Next.js middleware (`proxy.ts` in Next 16 / `middleware.ts` in earlier) is in the request chain.

**Steps to reproduce**

1. Set up `nextCookies()` in the plugins array per the standard guide, with it placed last.

2. Add a minimal Next.js middleware — even one that does nothing meaningful:

```ts
// proxy.ts (Next.js 16) or middleware.ts (Next.js ≤15)
import { NextResponse } from "next/server";
import type { NextRequest } from "next/server";

export function proxy(request: NextRequest) {
  const response = NextResponse.next();
  response.headers.set("x-pathname", request.nextUrl.pathname);
  return response;
}

export const config = {
  matcher: ["/((?!api|_next/static|_next/image|favicon.ico).*)"],
};
```

3. Add a diagnostic plugin **before** `nextCookies()` that mirrors the same `before`-hook header check:

```ts
{
  id: "DIAG-rsc-check",
  hooks: {
    before: [{
      matcher: (ctx) => (ctx as { path?: string }).path === "/get-session",
      handler: createAuthMiddleware(async () => {
        try {
          const { headers } = await import("next/headers");
          const h = await headers();
          const keys = Array.from(h.keys());
          console.log("[DIAG] rsc:", JSON.stringify(h.get("RSC")));
          console.log("[DIAG] next-action:", JSON.stringify(h.get("next-action")));
          console.log("[DIAG] header count:", keys.length);
          console.log("[DIAG] header keys:", JSON.stringify(keys));
        } catch (e) {
          console.log("[DIAG] headers() THREW:", String(e));
        }
      }),
    }],
  },
} satisfies BetterAuthPlugin,
```

4. Sign in fresh, then click a `<Link>` to navigate between two RSC pages.

5. Observe the dev server console.

**Expected**

With `nextCookies()` last in the plugins array, the plugin's `before` hook on `/get-session` should detect RSC navigation (`RSC: 1` header present, no `next-action` header) and call `setShouldSkipSessionRefresh(true)` so the refresh skips. This prevents DB drift in pure-RSC contexts where `cookies().set()` cannot be called anyway (Next.js architectural rule), keeping DB and browser cookie consistently frozen during read-only navigation until a Server Action or Route Handler call advances both.

**Actual**

With `proxy.ts` middleware present, Next.js's internal routing headers (`rsc`, `next-action`, `next-url`, `next-router-state-tree`) are stripped from the request before reaching `headers()` in downstream code called via `auth.api.*`. The hook's `headers().get("RSC")` returns `null`, so it never calls `setShouldSkipSessionRefresh`, and refresh runs on every RSC nav. Browser cookie still cannot propagate (RSC architectural rule), but the DB session row's `expiresAt` and `updatedAt` advance on every RSC nav unnecessarily.

**Empirical capture**

DevTools Network tab — request headers actually sent by the browser:
```
GET /some-page?_rsc=AbCdEf HTTP/1.1
rsc: 1
next-router-state-tree: <encoded tree>
next-url: /previous-page
cookie: <session cookie>
referer: http://localhost:3000/previous-page
sec-fetch-dest: empty
sec-fetch-mode: cors
... etc
```

`headers()` inside the BetterAuth `before` hook (after middleware ran):
```
[DIAG] rsc: "null"
[DIAG] next-action: "null"
[DIAG] header count: 21
[DIAG] header keys: [
  "host","connection","sec-ch-ua-platform","x-nextjs-request-id",
  "sec-ch-ua","sec-ch-ua-mobile","user-agent","x-nextjs-html-request-id",
  "accept","sec-fetch-site","sec-fetch-mode","sec-fetch-dest","referer",
  "accept-encoding","accept-language","cookie",
  "x-forwarded-host","x-forwarded-port","x-forwarded-proto","x-forwarded-for",
  "x-pathname"
]
```

`rsc`, `next-router-state-tree`, `next-url` all stripped. `x-forwarded-*` and the custom `x-pathname` from our middleware added.

**Empirical capture of DB drift consequence**

With test config `expiresIn: 4h`, `updateAge: 1min` (lowered from production 30m for capture speed):

| Action | DB expiresAt advanced? | Cookie advanced? |
|---|---|---|
| Sign in | (baseline) | (baseline) |
| Click `<Link>` to another page (RSC nav) | ✅ yes | ❌ no |
| Click any save button (Server Action) | ✅ yes | ✅ yes |
| Hard reload (Cmd+R) | ✅ yes | ❌ no |

The Server-Action `after`-hook path works as designed — cookie propagates via `cookies().set()`. The RSC nav and hard-reload paths both still write to the DB session row even though no cookie propagation is possible (Next.js doesn't allow `cookies().set()` from RSC render contexts). This DB drift would have been prevented if the `before` hook's RSC detection had engaged.

### Root cause hypothesis

Next.js's middleware layer consumes its internal routing headers (`rsc`, `next-*`) for routing decisions and does not forward them to downstream `headers()` callers. The headers ARE present in the browser-sent request and ARE visible inside the middleware function itself, but not to anything that runs after `NextResponse.next()`. This is observable regardless of whether the middleware modifies request headers — even a minimal `NextResponse.next()` triggers the stripping.

Since middleware is extremely common in production Next.js apps, the `nextCookies` plugin's RSC-skip optimization silently fails to engage for most real-world apps using it.

### Impact

- The `after` hook for cookie propagation still works correctly in Server Action / Route Handler contexts (verified empirically — cookie advances)
- The `before` hook's skip-refresh optimization never engages in RSC nav for apps with middleware
- Net behaviour for users: same DB write pattern as without the plugin in RSC contexts (DB advances every RSC nav), harmless to end users but inefficient. Server Actions and Route Handlers work correctly.

### Suggested directions

1. Document the middleware interaction in the `nextCookies` plugin docs
2. Add a runtime warning if `headers()` returns no `rsc`-related headers AND `next/headers` was importable (suggests middleware filtering)
3. Consider alternative detection mechanisms that don't depend on the `rsc` request header — for example, an explicit context flag set when the SDK is called from RSC, or a probe of whether `cookies().set()` would throw (though the existing comment in the source notes router-cache concerns with that approach)
4. Alternatively, document that apps with middleware must explicitly forward the internal headers in their middleware via `NextResponse.next({ request: { headers } })` — if this works

Happy to share more detail or test a fix branch if useful.

1.6.11

### System info

```
- better-auth: 1.6.11
- @better-auth/api-key: 1.6.11
- @better-auth/passkey: 1.6.11
- @better-auth/prisma-adapter: 1.6.11
- next: 16.2.6
- node: macOS, recent LTS
- App Router, ~24/26 authenticated pages are RSCs
```

shubhiscoding:
Closes #9776

## Problem

The `nextCookies()` plugin detects RSC navigation by checking for the
`RSC: 1` request header, and calls `setShouldSkipSessionRefresh(true)`
to avoid a DB write that can't be reflected back in cookies (RSC context
is read-only for cookies).

When any Next.js middleware is present — even a bare `NextResponse.next()`
— Next.js strips its own internal routing headers (`RSC`, `next-action`,
etc.) during the middleware → app router transition. By the time the
plugin's `before` hook runs, `headersStore.get("RSC")` is always `null`,
so the skip-refresh optimization never activates for these apps.

Result: every RSC navigation causes a wasted DB write and a growing
expiry drift between the DB session record and the browser cookie.
No functional harm to users, but unnecessary load.

## Solution

Two-part fix:

**1. `nextCookiesMiddleware()` export** — a helper users call in their
`middleware.ts` that copies `RSC: 1` to `x-better-auth-is-rsc: 1` (and
`next-action` to `x-better-auth-is-server-action: 1`) *before* Next.js
strips the originals. Custom `x-*` headers survive the middleware → app
transition.

**2. Updated detection in the `before` hook** — checks both the native
headers (no-middleware path) and the custom forwarded headers
(middleware path). Fully backwards-compatible.

## Usage

```ts
// middleware.ts
import { NextResponse } from "next/server";
import type { NextRequest } from "next/server";
import { nextCookiesMiddleware } from "better-auth/next-js";

export function middleware(request: NextRequest) {
  return NextResponse.next(nextCookiesMiddleware(request));
}
```

---
## Summary by cubic
Restores RSC detection in `nextCookies()` when Next.js middleware is present to prevent unnecessary session refresh DB writes. Adds `nextCookiesMiddleware` to forward RSC context headers through middleware.

- **Bug Fixes**
  - Exported `nextCookiesMiddleware` from `better-auth/next-js` to map `RSC`/`next-action` to `x-better-auth-*` before Next.js strips them.
  - Updated `nextCookies()` to read both native and forwarded headers, skipping refresh on RSC (but not during server actions).
  - Usage: `return NextResponse.next(nextCookiesMiddleware(request))` in `middleware.ts`.
  - Added docs, tests, and a patch changeset.

<sup>Written for commit bb5114ade0fb8eadf13a69a05b21e361cd841094. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9851?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/major-grapes-beg.md
+++ b/.changeset/major-grapes-beg.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+`nextCookies()` plugin now correctly skips session refresh during RSC navigation when Next.js middleware is present. Previously, middleware caused Next.js to strip the `RSC` header before the plugin could read it, resulting in unnecessary database writes. Use the new `nextCookiesMiddleware` helper in your middleware to forward RSC context headers.
--- a/docs/content/docs/integrations/next.mdx
+++ b/docs/content/docs/integrations/next.mdx
@@ -122,6 +122,38 @@ const signIn = async () => {
 }
 ```
 
+### Usage with Next.js Middleware
+
+When your app includes Next.js middleware (even a minimal pass-through), Next.js strips internal routing headers (`RSC`, `next-action`, etc.) before forwarding requests to the app router. This prevents `nextCookies()` from detecting RSC navigation, causing unnecessary session refreshes that cannot be reflected back in cookies.
+
+To fix this, use the `nextCookiesMiddleware` helper in your middleware to forward the necessary context headers:
+
+```ts title="middleware.ts"
+import { NextResponse } from "next/server";
+import type { NextRequest } from "next/server";
+import { nextCookiesMiddleware } from "better-auth/next-js";
+
+export function middleware(request: NextRequest) {
+    // ... your middleware logic
+    return NextResponse.next(nextCookiesMiddleware(request));
+}
+```
+
+If you need to combine it with other response options, spread the result:
+
+```ts title="middleware.ts"
+import { NextResponse } from "next/server";
+import type { NextRequest } from "next/server";
+import { nextCookiesMiddleware } from "better-auth/next-js";
+
+export function middleware(request: NextRequest) {
+    return NextResponse.next({
+        ...nextCookiesMiddleware(request),
+        // other NextResponse.next() options...
+    });
+}
+```
+
 ## Auth Protection
 
 In Next.js proxy/middleware, it's recommended to only check for the existence of a session cookie to handle redirection to avoid blocking requests by making API or database calls.
--- a/packages/better-auth/src/integrations/next-js.test.ts
+++ b/packages/better-auth/src/integrations/next-js.test.ts
@@ -160,4 +160,70 @@ describe("next-js integration", () => {
 
 		expect(session).not.toBeNull();
 	});
+
+	/**
+	 * When Next.js middleware is present (even a pass-through `NextResponse.next()`),
+	 * it strips internal routing headers (RSC, next-action, etc.) before forwarding
+	 * to the app router. The nextCookiesMiddleware() helper maps them to custom
+	 * x-better-auth-* headers that survive the middleware→app transition.
+	 *
+	 * @see https://github.com/better-auth/better-auth/issues/9776
+	 */
+	it("should skip refresh when x-better-auth-is-rsc header is set (middleware scenario)", async () => {
+		// Simulates a request that went through nextCookiesMiddleware():
+		// RSC header was stripped by Next.js but x-better-auth-is-rsc was forwarded.
+		const { cookiesMock, headersMock, session } =
+			await getSessionWithNextHeaders({
+				"x-better-auth-is-rsc": "1",
+			});
+
+		expect(headersMock).toHaveBeenCalledTimes(1);
+		expect(cookiesMock).not.toHaveBeenCalled();
+		expect(session).not.toBeNull();
+		expect(session?.needsRefresh).toBe(false);
+	});
+
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/9776
+	 */
+	it("should not skip refresh when x-better-auth-is-server-action is set (middleware + server action scenario)", async () => {
+		const { session } = await getSessionWithNextHeaders({
+			"x-better-auth-is-rsc": "1",
+			"x-better-auth-is-server-action": "1",
+		});
+
+		expect(session).not.toBeNull();
+		expect(session?.needsRefresh).toBe(true);
+	});
+
+	describe("nextCookiesMiddleware", () => {
+		it("should map RSC header to x-better-auth-is-rsc", async () => {
+			const { nextCookiesMiddleware } = await import("./next-js");
+			const request = { headers: new Headers({ RSC: "1" }) };
+			const result = nextCookiesMiddleware(request);
+			expect(result.request.headers.get("x-better-auth-is-rsc")).toBe("1");
+		});
+
+		it("should map next-action header to x-better-auth-is-server-action", async () => {
+			const { nextCookiesMiddleware } = await import("./next-js");
+			const request = {
+				headers: new Headers({ RSC: "1", "next-action": "abc123" }),
+			};
+			const result = nextCookiesMiddleware(request);
+			expect(result.request.headers.get("x-better-auth-is-rsc")).toBe("1");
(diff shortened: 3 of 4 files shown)


bytaesu (changes requested):
> [!NOTE]
> - https://nextjs.org/docs/app/api-reference/file-conventions/proxy#rsc-requests-and-rewrites
> - https://github.com/vercel/next.js/blob/58e8c0b4457f8f7cb737158efc543f3708f3e6e3/packages/next/src/client/components/app-router-headers.ts#L25-L31

Actually, the `"rsc"` header isn't accessible to consumers. So we need to improve this part of the existing logic, and this PR won't work either.

My previous PR description for reference: https://github.com/better-auth/better-auth/pull/7763

bytaesu:
Closing this as the proposed direction doesn't seem right. And we want to address this through a plugin-based approach instead of middleware.

Issue #6613, opened 2025-12-08
Upgrade from 1.3 to 1.4: Error runWithRequestState on cloudflare workers


tnkuehne:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Follow the setup guide for SvelteKit https://www.better-auth.com/docs/integrations/svelte-kit. This worked perfectly deployed on cloudflare workers (node compat enabled `"compatibility_date": "2025-06-12"` and `"compatibility_flags": ["nodejs_compat"]` with version 1.3
2. Upgrade to 1.4
3. Call `await auth.api.signInSSO`

```ts
const res = await auth.api.signInSSO({
    body: {
        email,
        callbackURL: redirectTo,
    },
});
```

Throws error
```console
No request state found. Please make sure you are calling this function within a `runWithRequestState` callback.
    at getCurrentRequestState (file:///home/timo/ZedProjects/enterprise/node_modules/.pnpm/@better-auth+core@1.4.5_@better-auth+utils@0.3.0_@better-fetch+fetch@1.1.18_better-call_c97e445e802b315319ad04ddb5da9384/node_modules/@better-auth/core/dist/context-DgQ9XGBl.mjs:41:20)
    at async set (file:///home/timo/ZedProjects/enterprise/node_modules/.pnpm/@better-auth+core@1.4.5_@better-auth+utils@0.3.0_@better-fetch+fetch@1.1.18_better-call_c97e445e802b315319ad04ddb5da9384/node_modules/@better-auth/core/dist/context-DgQ9XGBl.mjs:63:5)
    at async generateState (file:///home/timo/ZedProjects/enterprise/node_modules/.pnpm/better-auth@1.4.5_@sveltejs+kit@2.49.1_@opentelemetry+api@1.9.0_@sveltejs+vite-plugin-s_2117d6321a543e35ae24707332b8d2bc/node_modules/better-auth/dist/utils-db7gNqd-.mjs:28:2)
    at null.<anonymous> (async file:///home/timo/ZedProjects/enterprise/.wrangler/tmp/dev-OZuMhX/_worker.js:116425:26)
    at async internalHandler (file:///home/timo/ZedProjects/enterprise/node_modules/.pnpm/better-call@1.1.4_zod@4.1.13/node_modules/better-call/src/endpoint.ts:503:20)
[...]

PR #9862, opened 2026-06-01
fix(core): dedup request-state AsyncLocalStorage init to fix intermittent "No request state found"

OrangeManLi:
## Summary

`@better-auth/core`'s `ensureAsyncStorage()` has an initialization race that can intermittently throw:

```
No request state found. Please make sure you are calling this function within a `runWithRequestState` callback.
```

It does check → `await getAsyncLocalStorage()` → assign, with **no in-flight dedup**:

```ts
if (!betterAuthGlobal.context.requestStateAsyncStorage) {
  const AsyncLocalStorage = await getAsyncLocalStorage(); // yields
  betterAuthGlobal.context.requestStateAsyncStorage = new AsyncLocalStorage(); // last writer wins
}
```

When several requests hit a cold isolate concurrently, they all observe the singleton as unset, all `await` (the shared `node:async_hooks` import promise resolves on a microtask, so they interleave), and each constructs its own `AsyncLocalStorage`; the last assignment overwrites the rest. `runWithRequestState().run()` then executes on instance A while a nested `getCurrentRequestState()` resolves the overwritten instance B → `getStore()` is `undefined` → the error above. Since `to-auth-endpoints` wraps every endpoint in `runWithRequestState`, any server-side `auth.api.*` call (e.g. `getSession()`) is exposed.

We hit this on **Cloudflare Workers** (`nodejs_compat`, Next.js/OpenNext) — intermittent, clustered around cold start / fresh deploys, surfaced from server-side `getSession()`. It is **not** the Dual Module Hazard (single `@better-auth/core` copy; `betterAuthGlobal` is on `globalThis`).

Related report: #6613.

## Fix

Memoize the in-flight init promise so concurrent first-callers share a single `AsyncLocalStorage`, and assign idempotently (`??=`) so the global singleton stays stable even if multiple BetterAuth copies race here:

```ts
let asyncStorageInit: Promise<AsyncLocalStorage<RequestStateWeakMap>> | null = null;

const ensureAsyncStorage = async () => {
  const betterAuthGlobal = __getBetterAuthGlobal();
  const existing = betterAuthGlobal.context.requestStateAsyncStorage;
  if (existing) return existing as AsyncLocalStorage<RequestStateWeakMap>;
  if (!asyncStorageInit) {
    asyncStorageInit = getAsyncLocalStorage().then((AsyncLocalStorage) => {
      betterAuthGlobal.context.requestStateAsyncStorage ??= new AsyncLocalStorage<RequestStateWeakMap>();
      return betterAuthGlobal.context.requestStateAsyncStorage as AsyncLocalStorage<RequestStateWeakMap>;
    });
  }
  return asyncStorageInit;
};
```

## Test

Added a regression test to `request-state.test.ts` that resets the module (`vi.resetModules()`) and the global singleton to force a cold start, then races 32 concurrent first-callers through `runWithRequestState` + nested `getCurrentRequestState`. **Verified: fails on the current `main` implementation (the racy losers throw "No request state found"), passes with this change.** Existing request-state tests still pass.

## Changeset

`@better-auth/core` patch included.

---
## Summary by cubic
Deduplicates request-state `AsyncLocalStorage` init in `@better-auth/core` to stop intermittent “No request state found” errors on serverless cold starts. Concurrent first calls now converge on one instance, keeping the global stable even if multiple BetterAuth copies load.

- **Bug Fixes**
  - Memoized `ensureAsyncStorage()` with a shared init promise and idempotent `??=` so concurrent first callers share one `AsyncLocalStorage`, including under dual-module races.
  - Added a regression test that resets module/global state and races 32 callers; uses `vi.resetModules()` and avoids `delete` to satisfy Biome.

<sup>Written for commit adffb2b87c073865441dfc438daf6cb3fc8f5c9b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9862?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hsab:
This is a highly anticipated PR especially for those using BA in Cloudflare environments

Vaiune:
we are also experiencing this on cloudflare workers #10300 
when can we expect this to land?

coji:
Heads-up from #10315: there is a second failure mode in this same initialization that this PR's memoization doesn't cover — and the memoized `asyncStorageInit` promise itself inherits it.

On Cloudflare Workers, a promise created while handling a request belongs to that request's IoContext; if the client aborts before it settles, workerd never settles it. So whichever request first drives this lazy init can, when aborted, leave the cached promise (`AsyncLocalStoragePromise`, and with this PR also `asyncStorageInit`) permanently pending — after which every auth API call in the isolate hangs forever before reaching the DB. We hit this in production; diagnosis details in #10315.

Both #10300's race and the abort-poisoning disappear if the lookup short-circuits synchronously before any dynamic import, e.g. in `@better-auth/core/src/async_hooks`:

```ts
const AsyncLocalStoragePromise: Promise<typeof AsyncLocalStorage | null> =
  "AsyncLocalStorage" in globalThis
    ? Promise.resolve((globalThis as any).AsyncLocalStorage)
    : import(/* @vite-ignore */ /* webpackIgnore: true */ "node:async_hooks")
        .then((mod) => mod.AsyncLocalStorage)
        .catch(/* existing fallback */);
```

`nodejs_compat` on Workers exposes `globalThis.AsyncLocalStorage`, so the dynamic-import path (the only abortable step) is never taken there, and `ensureAsyncStorage` resolves in the same tick — which also removes the interleaving window this PR guards against. Happy to test a patch.

coji:
Correction to my comment above: on workerd, `AsyncLocalStorage` is **not** exposed on `globalThis` — `nodejs_compat` only provides it via `import from "node:async_hooks"` (https://developers.cloudflare.com/workers/runtime-apis/nodejs/asynclocalstorage/). So the `"AsyncLocalStorage" in globalThis` short-circuit wouldn't fire there.

The abort-safe variant for Workers is a **static** import instead of the dynamic one — a static `import { AsyncLocalStorage } from "node:async_hooks"` resolves at module evaluation (no request-owned promise to poison). Since not all targets have `node:async_hooks`, that likely means a separate entry/export condition (e.g. a `workerd`/`edge` condition that uses the static import, keeping the dynamic-import fallback for browsers). That, or the timeout-and-retry on the cached promise. Sorry for the noise.

hsab:
If it can be of help, this is our patch to resolve the issue until fixed in upstream

```patch
diff --git a/dist/context/endpoint-context.mjs b/dist/context/endpoint-context.mjs
index 9e2e674cbc12079cde9ccbfbb4ef6208339b4793..05fb6c5e576b1dcde65f6efce8bae2032ab5c992 100644
--- a/dist/context/endpoint-context.mjs
+++ b/dist/context/endpoint-context.mjs
@@ -5,8 +5,14 @@ import { getAsyncLocalStorage } from "@better-auth/core/async_hooks";
 const ensureAsyncStorage = async () => {
 	const betterAuthGlobal = __getBetterAuthGlobal();
 	if (!betterAuthGlobal.context.endpointContextAsyncStorage) {
-		const AsyncLocalStorage = await getAsyncLocalStorage();
-		betterAuthGlobal.context.endpointContextAsyncStorage = new AsyncLocalStorage();
+		// Patched (mirrors better-auth#9862): memoize the in-flight init so
+		// concurrent first callers share one AsyncLocalStorage instead of the
+		// last writer replacing instances other requests are running inside.
+		betterAuthGlobal.context.endpointContextAsyncStoragePromise ??= getAsyncLocalStorage().then((AsyncLocalStorage) => {
+			betterAuthGlobal.context.endpointContextAsyncStorage ??= new AsyncLocalStorage();
+			return betterAuthGlobal.context.endpointContextAsyncStorage;
+		});
+		return betterAuthGlobal.context.endpointContextAsyncStoragePromise;
 	}
 	return betterAuthGlobal.context.endpointContextAsyncStorage;
 };
diff --git a/dist/context/request-state.mjs b/dist/context/request-state.mjs
index 8d84ffa13deca4e220e8f9f90fe57a39fa361e0e..64d70eee932fbf2468c7c5009004a2dc9c069378 100644
--- a/dist/context/request-state.mjs
+++ b/dist/context/request-state.mjs
@@ -5,8 +5,14 @@ import { getAsyncLocalStorage } from "@better-auth/core/async_hooks";
 const ensureAsyncStorage = async () => {
 	const betterAuthGlobal = __getBetterAuthGlobal();
 	if (!betterAuthGlobal.context.requestStateAsyncStorage) {
-		const AsyncLocalStorage = await getAsyncLocalStorage();
-		betterAuthGlobal.context.requestStateAsyncStorage = new AsyncLocalStorage();
+		// Patched (mirrors better-auth#9862): memoize the in-flight init so
+		// concurrent first callers share one AsyncLocalStorage instead of the
+		// last writer replacing instances other requests are running inside.
+		betterAuthGlobal.context.requestStateAsyncStoragePromise ??= getAsyncLocalStorage().then((AsyncLocalStorage) => {
+			betterAuthGlobal.context.requestStateAsyncStorage ??= new AsyncLocalStorage();
+			return betterAuthGlobal.context.requestStateAsyncStorage;
+		});
+		return betterAuthGlobal.context.requestStateAsyncStoragePromise;
 	}
 	return betterAuthGlobal.context.requestStateAsyncStorage;
 };
diff --git a/dist/context/transaction.mjs b/dist/context/transaction.mjs
index 20cc831249eef424662b110a208c2f5f8de846ad..07e0548ad15eb0ad421e8bcfd2805287f513629a 100644
--- a/dist/context/transaction.mjs
+++ b/dist/context/transaction.mjs
@@ -5,8 +5,14 @@ import { getAsyncLocalStorage } from "@better-auth/core/async_hooks";
 const ensureAsyncStorage = async () => {
 	const betterAuthGlobal = __getBetterAuthGlobal();
 	if (!betterAuthGlobal.context.adapterAsyncStorage) {
-		const AsyncLocalStorage = await getAsyncLocalStorage();
-		betterAuthGlobal.context.adapterAsyncStorage = new AsyncLocalStorage();
+		// Patched (mirrors better-auth#9862): memoize the in-flight init so
+		// concurrent first callers share one AsyncLocalStorage instead of the
+		// last writer replacing instances other requests are running inside.
+		betterAuthGlobal.context.adapterAsyncStoragePromise ??= getAsyncLocalStorage().then((AsyncLocalStorage) => {
+			betterAuthGlobal.context.adapterAsyncStorage ??= new AsyncLocalStorage();
+			return betterAuthGlobal.context.adapterAsyncStorage;
+		});
+		return betterAuthGlobal.context.adapterAsyncStoragePromise;
 	}
 	return betterAuthGlobal.context.adapterAsyncStorage;
 };
```

ping-maxwell (dismissed):
lgtm

ping-maxwell (approved):
fixed CI, lgtm

PR #9864, opened 2026-06-01
feat(auth): add user.validateUserInfo provisioning gate

### LINKED ISSUE #9732: Support email validation/whitelisting in stateless social auth (no database)

When using stateless auth with social providers (no database adapter), there's no clean way to validate or whitelist user emails during the OAuth flow.

A dedicated callback (e.g. `onBeforeUserCreated` or `validateUser`) to social provider config that:

1. Receives the OAuth profile (including email)
2. Works without a database adapter
3. Supports returning a custom error/redirect URL on rejection
4. Clearly communicates to the user why sign-in was denied

This would also cover use cases like blocking disposable emails, restricting to specific email addresses, or enforcing domain-based access control — all common needs for apps using stateless social auth.

I considered throwing an error from `mapProfileToUser`, which:

- Results in a generic error with no custom message
- Doesn't allow custom error redirects or user-friendly error pages
- Feels like a workaround rather than an intentional API

### Additional context

_No response_

GautamBytes:
Adds `user.validateUserInfo`, a gate that decides which identities Better Auth admits before it creates a user, links an account, or signs a returning OAuth user back in. Closes #9732.

Coverage comes from a single seam rather than per-flow wiring. Every creation path routes through `internalAdapter.createUser`, which now takes a required `source` argument and runs the gate there, so a new flow cannot silently skip it: a missed path becomes a type error. The earlier per-provider shape had already missed phone number, admin-created users, and SCIM. Account linking writes no user row, so it calls the gate directly with `action: "link-account"`.

OAuth sign-in is the only returning-user case worth re-validating, because it is the only method that re-asserts an external profile each time. The gate runs inside the existing sign-in transaction, before the account write. That gives a policy the current provider identity rather than the stored row, and rolls back a rejected sign-in instead of refreshing tokens for a denied user. Non-OAuth returning sign-ins carry only the unchanged stored row, so they are left to ban controls or `databaseHooks.session.create.before`. The `validateUserInfo` name is kept for now, with a TODO to rename to `validateUser`: "UserInfo" reads as the OIDC term and misleads for the credential, SIWE, and phone methods.


packages/better-auth/src/plugins/generic-oauth/types.ts
```diff
@@ -142,6 +142,27 @@ export interface GenericOAuthConfig<ID extends string = string> {
...
+	 * Runs after `mapProfileToUser` but before user creation. Return `void` to
+	 * allow the sign-in. Return an object with `error` or `errorDescription`
+	 * to reject. Redirect-based OAuth callbacks send the rejection to the error
+	 * URL; API-based flows such as ID-token sign-in return a 401 API error.
+	 *
+	 * The `user` argument contains the mapped user data. The `profile` argument
+	 * contains the raw OAuth user info from the provider, allowing access to
+	 * provider-specific fields for custom validation logic.
+	 */
+	validateUser?:
```
Bekacru: we should rename the option to `validateUserInfo`

Bekacru:
We should most likely consider more global way of validating user info than a per-provider config, that works for all provider + any other signup method including email/password.

GautamBytes:
> We should most likely consider more global way of validating user info than a per-provider config, that works for all provider + any other signup method including email/password.

did the suggested changes, ready for review!

gustavovalverde:
@cubic-dev-ai review

gustavovalverde:
@GautamBytes I made additional changes (maybe way more than I'd have expected)

PR #9873, opened 2026-06-02
feat(passkey): create session during passkey registration

### LINKED ISSUE #9866: Support creating a session when adding a passkey

Registering a passkey does not create a session and thus does not sign in the user. Signing in via the passkey creates a session. This forces a passkey-first (or more precisely passkey-only) onboarding to first prompt the user to add a passkey, and then prompt the user again to sign in via the passkey. Ideally this would be one step, like in Pocket ID. 

A solution would be to add a field in the `authClient.passkey.addPasskey()` options where you can toggle whether a session is also created when registering the passkey. An API like the code below would suffice:

```ts
await authClient.passkey.addPasskey({
	name: 'foo',
	context: 'bar',
	session: true, // alternative key name?: createSession
});
```

The only alternative I know of is the aforementioned method of calling `authClient.signIn.passkey()` after calling `authClient.passkey.addPasskey()`.

### Additional context

[Original question posted on the Discord server.](https://discord.com/channels/1288403910284935179/1288403910284935182/1508092102503039026)

GautamBytes:
Passkey-first registration cannot currently establish a session, which forces users through a second passkey ceremony immediately after onboarding. This adds opt-in session creation after WebAuthn verification while preserving the existing registration behavior by default.

## What changes

- `createSession: true` creates a session, sets its cookie, and returns the `session` and `user` with the registered passkey.
- The client refreshes the passkey state and notifies `$sessionSignal` when registration creates a session.
- `registration.afterVerification` remains the user-provisioning boundary, allowing integrations to create or load the user only after credential verification succeeds.
- Session-enabled persistence uses the adapter transaction boundary. Adapters with transaction support roll back Better Auth writes if session creation fails; adapters without transactions retain the existing sequential behavior.

Fixes #9866

### DIFF
--- a/.changeset/passkey-registration-session.md
+++ b/.changeset/passkey-registration-session.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/passkey": minor
+---
+
+Add `createSession` to passkey registration so passkey-first onboarding can sign the user in immediately after a successful registration.
--- a/.changeset/passkey-registration-session.md
+++ b/.changeset/passkey-registration-session.md
@@ -2,4 +2,4 @@
 "@better-auth/passkey": minor
 ---
 
-Add `createSession` to passkey registration so passkey-first onboarding can sign the user in immediately after a successful registration.
+Add the optional `createSession` setting to passkey registration. When enabled, successful registration signs the user in, sets the session cookie, and returns the session and user with the registered passkey.
--- a/docs/content/docs/plugins/passkey.mdx
+++ b/docs/content/docs/plugins/passkey.mdx
@@ -120,6 +120,7 @@ When using passkey-first flows (`registration.requireSession: false`), pass the 
 await authClient.passkey.addPasskey({
   name: "Primary passkey",
   context: "signed-registration-token",
+  createSession: true,
 })
 ```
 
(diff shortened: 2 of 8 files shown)


GautamBytes:

GautamBytes:
> @ping-maxwell @gustavovalverde ready for review!

jlian:
Thanks for working on this. [We](https://github.com/jlian/wingdex) do passkey-first signup on both web and native iOS, and this looks like it would finally give us the passkeys-first signup we wanted.

One question: since `resolveUser` currently needs to return an existing user before the passkey ceremony starts, does that mean the app still needs to create a temporary user first?

Would it make sense to support creating the user only after the passkey has been verified? Something like: 

1. verify the registration, 
2. create the user, 
3. save the passkey, then 
4. create the session. 

That would avoid leaving temporary users behind when someone cancels the passkey prompt.

Together with `createSession`, I think that would give us a complete passkey signup flow on both web and native without needing an anonymous-session workaround. Not sure if that fits in this PR or makes more sense as a follow-up, but very much +1 on the direction.

GautamBytes:
cc : @ping-maxwell @bytaesu

PR #9890, opened 2026-06-03
feat(oauth-popup): add popup-based OAuth sign-in

bytaesu:
> [!NOTE]
> Experimental plugin 

### Server

```ts
import { oauthPopup, bearer } from "better-auth/plugins";

  // ...
  plugins: [oauthPopup(), bearer()], // <- bearer is required for iframe
});
```

### Client

```ts
import { createAuthClient } from "better-auth/client";
import { oauthPopupClient } from "better-auth/client/plugins";

  // ...
  plugins: [oauthPopupClient()],
});
```

### Sign-in

```ts
const { data, error } = await authClient.signIn.popup({
  provider: "",
  callbackURL: "",
});

if (error) {
 // handle err
}

// session is ready
```

https://github.com/user-attachments/assets/26ce467d-51ef-42f2-b869-4d93a93e2d58


packages/better-auth/src/plugins/oauth-popup/index.ts
```diff
@@ -0,0 +1,360 @@
...
+<head><meta charset="utf-8"><title>Completing sign-in</title></head>
+<body>
+<script type="application/json" id="${OAUTH_POPUP_DATA_ELEMENT_ID}">${inlineJSON(data)}</script>
+<script>${OAUTH_POPUP_COMPLETE_SCRIPT}</script>
+</body>
+</html>`;
+
+	return new Response(html, {
+		status: 200,
+		headers: {
```
ping-maxwell, line 123: add
"Cross-Origin-Opener-Policy": "same-origin-allow-popups",

ping-maxwell (dismissed):

packages/better-auth/src/plugins/oauth-popup/index.ts
cloudycotton: This header is relevant for the `window.closed` warnings, but I found it does not fix the direct-mode failure by itself. Also, it needs to be present on the opener document; adding it only to the popup completion response is too late to establish the opener policy.

I made a zero-config reproduction against the PR preview package: https://github.com/runablehq/better-auth-auth-popup-failure

Observed behavior:

- direct page: popup completes, then `POPUP_SIGN_IN_FAILED`; session is signed out
- cross-origin iframe: popup completes and authenticates through the bearer token
- both opener pages already send `Cross-Origin-Opener-Policy: same-origin-allow-popups`

The concrete issue appears to be the callback cookie replay in the after-hook. The OAuth callback creates `better-auth.session_token` and `better-auth.session_data`, but after replacing the redirect with the completion `Response`, the final HTTP response only contains `better-auth.session_data`. The session token cookie is lost, so the top-level client's subsequent cookie-authenticated `/get-session` is signed out.

I verified a server-side fix with the client unchanged: append every original `Set-Cookie` value directly to the returned completion `Response` (including `session_token`), append the popup-marker expiry there too, and remove the accumulated `set-cookie` headers before assigning `c.context.returned`. After that, direct mode and iframe mode both authenticate successfully.

It would be useful to add an integration assertion that the completion response preserves all callback cookies, especially `session_token`, rather than only testing the posted token.

bytaesu:
> [!NOTE]
> The root cause is in `better-call`. The fix is already in place, but until a release is out, I'm using a workaround for now.

packages/better-auth/src/plugins/oauth-popup/index.ts
```diff
@@ -0,0 +1,371 @@
...
+						}
+
+						// TODO: drop this once the response builder preserves multiple
+						// Set-Cookie headers (e.g. via `getSetCookie`).
+						for (const raw of setCookies) {
+							response.headers.append("set-cookie", raw);
+						}
+						response.headers.append(
+							"set-cookie",
+							`${cookie.name}=; Max-Age=0; Path=${cookie.attributes.path ?? "/"}`,
```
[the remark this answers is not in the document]
bytaesu: see https://github.com/better-auth/better-auth/pull/9890#issuecomment-4640564259

packages/better-auth/src/plugins/oauth-popup/index.ts
bytaesu: Seems related to https://github.com/better-auth/better-call/pull/137

bytaesu: @cloudycotton I confirmed the issue occurs in stateless (DB-less) mode. I also reproduced and verified the exact failure point with a follow-up workaround commit.

It looks like we can fix this cleanly by releasing and adopting the change that's already been merged in `better-call`, so I'll try that approach first.

packages/better-auth/src/plugins/oauth-popup/index.ts
```diff
@@ -0,0 +1,364 @@
...
+									description:
+										new URL(redirectTo, c.context.baseURL).searchParams.get(
+											"error_description",
+										) ?? undefined,
+								},
+							});
+						}
+
+						// Swap the redirect for the completion page; the callback's
+						// session cookies ride along on the response.
```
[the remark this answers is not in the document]
bytaesu: fixed in better-call v1.3.6

packages/better-auth/src/plugins/oauth-popup/oauth-popup.test.ts
```diff
@@ -0,0 +1,239 @@
...
+	OAUTH_POPUP_SCRIPT_CSP_HASH,
+	oauthPopup,
+} from "./index";
+
+describe("oauth popup completion script", () => {
+	// The completion page runs this script inline and pins its hash in the CSP.
+	// If the script body changes, recompute and update OAUTH_POPUP_SCRIPT_CSP_HASH.
+	it("pins the script's sha256 in the response CSP hash", () => {
+		const digest = createHash("sha256")
+			.update(OAUTH_POPUP_COMPLETE_SCRIPT)
```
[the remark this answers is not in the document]
bytaesu: NO

PR #9894, opened 2026-06-04
fix(oauth-provider): run hooks for authorize resume paths

GautamBytes:
## Summary
- run configured auth hooks around internal OAuth authorize resume paths
- preserve normal endpoint-dispatch behavior for the public /oauth2/authorize endpoint
- cover post-login, continue/select-account, and consent resume flows

## Root cause
OAuth-provider resume paths called authorizeEndpoint() directly, bypassing the framework endpoint pipeline that normally runs user-configured hooks.before and hooks.after.

Fixes #9887

---
## Summary by cubic
Runs configured auth hooks around OAuth authorize resume paths and moves the hook pipeline into the core endpoint layer so hooks run for both HTTP requests and internal `auth.api` calls. Keeps `/oauth2/authorize` behavior unchanged while preserving before-hook response headers and cookies across redirects and errors.

- **Bug Fixes**
  - Implemented core hook pipeline in `createAuthEndpoint`; runs `hooks.before`/`hooks.after` once per call for `auth.api.*` and router; preserves response headers from short-circuited before hooks, merges `Set-Cookie`, and propagates APIError headers via `kAPIErrorHeaderSymbol`.
  - Refactored resume flows to call the `oauth2Authorize` endpoint via an internal runner so hooks execute consistently: continue/select-account, created/post-login, consent, and `prompt=login`.
  - Added tests for the endpoint hook pipeline and OAuth resumes; verified before-hook header preservation (no request cookie leaks) and after-hook behaviors (replace redirects, throw APIErrors, return null).

<sup>Written for commit 11a81aea299ca2569e749b7990df3efe0c3bb873. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9894?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #9918, opened 2026-06-06
Make dispatch the single boundary where configured hooks run


gustavovalverde:
## Problem

Configured `hooks.before` / `hooks.after` run at dispatch boundaries: the HTTP router and `auth.api.*`. They do not run when one endpoint calls another as a function, such as `getSessionFromCtx` invoking the `getSession` endpoint, or `sessionMiddleware` resolving a session. The boundary is worth keeping. The alternative, running hooks on every endpoint call, would fire global hooks for internal `/get-session` on most authenticated requests.

Today that boundary is implicit. Nothing documents it, so a contributor could wrap `createAuthEndpoint` to recover a "missing" hook and silently reintroduce the internal-call regression.

## Context

#9894 fixes the immediate symptom (#9887, hooks skipped when authorize resumes after a fresh sign-in) and adds `dispatchAuthEndpoint(endpoint, ctx)`, the sanctioned way for a plugin to re-enter the hook pipeline on purpose. This issue tracks the follow-up work that PR leaves out.

## Scope

- Document the boundary: hooks run on dispatch, and `dispatchAuthEndpoint` is the only supported way to re-run them from inside a handler.
- Audit `oidc-provider` and `mcp` for the same resume-skips-hooks pattern, and adopt `dispatchAuthEndpoint` where it applies.
- Reduce endpoints reused as subroutines. `getSessionFromCtx` calls the `getSession` endpoint to reuse its logic; extracting a shared function would make the boundary structural instead of conventional.

YashChincholi:
Hi @gustavovalverde and Better-Auth Team! 👋

This is an excellent architectural refinement. Relying on implicit boundaries for hook execution makes the plugin ecosystem brittle and prone to regression leaks whenever internal handlers invoke sister endpoints directly as standard subroutines.

I would love to take this on, formalize the dispatch boundary, and refactor the shared subroutines to make the execution flow structurally explicit.

### 🛠️ Proposed Implementation Approach

Following the pattern established by `dispatchAuthEndpoint(endpoint, ctx)`, I plan to tackle the scope across three clear phases:

1. **Deconstruct Endpoint Subroutines**:
   * Audit handlers like `getSessionFromCtx` that currently invoke the full `getSession` endpoint directly. 
[...]

PR #9919, opened 2026-06-06
fix(oauth-provider): run configured hooks when authorize resumes

### LINKED ISSUE #9887: oauth-provider: configured `hooks.before` / `hooks.after` do not run when authorize resumes after a fresh sign-in

### Reproduction

1. Configure an auth instance with `oauthProvider` and a `hooks.before` (e.g. one that logs
   or sets a marker every time it runs).
2. Start the authorize flow while **already signed in** → the hook runs. ✅
3. Start the authorize flow while **signed out** → sign in → authorize resumes → the hook
   does **not** run. ❌

Reproduction repository is here: [kopancek/better-auth-hooks-before-authorize-replay-repro](https://github.com/kopancek/better-auth-hooks-before-authorize-replay-repro)

### Expected
Configured `hooks.before` / `hooks.after` run consistently for every request the auth
instance processes — including when authorize resumes after a fresh sign-in.

### Actual
On the fresh post-login authorize resume, the configured hooks are never called.

When `/oauth2/authorize` resumes after a fresh sign-in, the oauth-provider runs the
authorize step by calling its internal `authorizeEndpoint(...)` function **directly**,
rather than dispatching it through the normal endpoint pipeline. Because the framework
runs the configured `options.hooks.before` / `options.hooks.after` only as part of endpoint
dispatch, those hooks are **not invoked** for this post-login authorize step.

The effect: a hook you register runs for a standalone `/oauth2/authorize` request, but is
silently skipped when authorize resumes immediately after login. Whatever logic you put in
that hook simply doesn't execute on the post-login path.

1.6.14

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.6.0: Mon Jan 19 22:00:55 PST 2026; root:xnu-11417.140.69.708.3~1/RELEASE_ARM64_T6000",
    "release": "24.6.0",
    "cpuCount": 10,
    "cpuModel": "Apple M1 Max",
    "totalMemory": "64.00 GB",
    "freeMemory": "0.21 GB"
  },
  "node": {
    "version": "v24.13.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.2"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "1.6.14",
    "config": null
  }
}
```

Backend

```typescript
import { createAuthMiddleware } from "better-auth/api";
import { jwt } from "better-auth/plugins";
import { memoryAdapter } from "@better-auth/memory-adapter";

  secret: "repro-secret-0000000000000000000000000000",
  database: memoryAdapter({}),            // any adapter; not relevant to the bug

  // The hook under test: runs for a standalone /oauth2/authorize request,
  // but NOT when authorize resumes after a fresh sign-in.
  hooks: {
      console.log("hooks.before ran for:", ctx.path);
    }),
  },

  plugins: [
    jwt(),
    oauthProvider({
      loginPage: "/sign-in",
      consentPage: "/consent",
    }),
  ],
});
```

### Proposed fix
Invoke the configured hooks.before / hooks.after around this resumed authorize step (or route the resume through the normal dispatch pipeline), so registered hooks run the same way they do for a standalone authorize request.

### Related
[#9178](https://github.com/better-auth/better-auth/issues/9178) (tracking — "Authorize does not always resume after a fresh sign-in"). This is specifically about hook-execution consistency on that resume path.

gustavovalverde:
Resumed `/oauth2/authorize` flows (post-login, account selection, consent) skipped the configured hook pipeline because they called the authorize function directly.

#9894 fixed this by running hooks on every endpoint call, which also fired global hooks on internal `getSession` calls (i.e. most authenticated requests). This instead adds one shared `dispatchAuthEndpoint` primitive and routes the resume paths through it explicitly, keeping hooks at one run per dispatch boundary while internal endpoint calls stay hook-free. Also fixes two latent hook-header cases (before-hook short-circuit, after-hook APIError headers).

Supersedes #9894, built on @GautamBytes's diagnosis and implementation. Boundary decision tracked in #9918.

Fixes #9887
Closes #9894

### DIFF
--- a/.changeset/oauth-authorize-resume-hooks.md
+++ b/.changeset/oauth-authorize-resume-hooks.md
@@ -1,0 +1,12 @@
+---
+"better-auth": minor
+"@better-auth/oauth-provider": patch
+---
+
+Run configured hooks when OAuth authorize resumes after sign-in
+
+Configured `hooks.before` / `hooks.after` now run when the OAuth provider resumes `/oauth2/authorize` after a fresh sign-in, account selection, or consent. The resume previously called the authorize function directly and skipped the hook pipeline, so a hook registered on the auth instance ran for a standalone authorize request but not the post-login resume.
+
+The hook pipeline is now a shared `dispatchAuthEndpoint` primitive, exported from `better-auth/api`. A plugin re-runs the configured hooks for one endpoint by dispatching it, rather than calling it as a plain function. Internal endpoint-to-endpoint calls (for example resolving the session through the `getSession` endpoint) still skip hooks, so a single request does not fire them twice.
+
+This also corrects two hook header cases. A `hooks.before` that sets response headers or cookies before short-circuiting now keeps them on the response instead of dropping them. A `hooks.after` that throws an `APIError` now merges the cookies it accumulated with the error's explicit headers, so neither side is lost.
--- a/.changeset/oauth-authorize-resume-hooks.md
+++ b/.changeset/oauth-authorize-resume-hooks.md
@@ -1,12 +1,10 @@
 ---
-"better-auth": minor
+"better-auth": patch
 "@better-auth/oauth-provider": patch
 ---
 
-Run configured hooks when OAuth authorize resumes after sign-in
+Run configured hooks through the whole OAuth sign-in flow
 
-Configured `hooks.before` / `hooks.after` now run when the OAuth provider resumes `/oauth2/authorize` after a fresh sign-in, account selection, or consent. The resume previously called the authorize function directly and skipped the hook pipeline, so a hook registered on the auth instance ran for a standalone authorize request but not the post-login resume.
+`hooks.before` / `hooks.after` configured on the auth instance now run for the OAuth authorization that continues after a user signs in, selects an account, or consents. They were being skipped there.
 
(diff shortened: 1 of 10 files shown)


.changeset/oauth-authorize-resume-hooks.md
```diff
@@ -0,0 +1,12 @@
+---
+"better-auth": patch
+"@better-auth/oauth-provider": patch
+---
+
+Run configured hooks when OAuth authorize resumes after sign-in
+
+Configured `hooks.before` / `hooks.after` now run when the OAuth provider resumes `/oauth2/authorize` after a fresh sign-in, account selection, or consent. The resume previously called the authorize function directly and skipped the hook pipeline, so a hook registered on the auth instance ran for a standalone authorize request but not the post-login resume.
+
+The hook pipeline is now a shared `dispatchAuthEndpoint` primitive, exported from `better-auth/api`. A plugin re-runs the configured hooks for one endpoint by dispatching it, rather than calling it as a plain function. Internal endpoint-to-endpoint calls (for example resolving the session through the `getSession` endpoint) still skip hooks, so a single request does not fire them twice.
```
bytaesu: > Image

I think changesets should focus on user-visible behavior changes rather than describing internal implementation details. They're primarily meant for users to understand what changed and whether it affects them.

gustavovalverde: This should be part of the `CLAUDE.md` or a skill for our codebases

PR #9923, opened 2026-06-07
feat(oauth-provider): add OIDC front-channel logout

### LINKED ISSUE #9922: feat(oauth-provider): OIDC Front-Channel Logout 1.0
## Summary

[OIDC Front-Channel Logout 1.0](https://openid.net/specs/openid-connect-frontchannel-1_0.html) is the browser-based sibling to the back-channel logout shipped in #9304. The back-channel PR explicitly called it out as the next step:

> *Front-channel logout (the sibling spec). The session-delete hook and the split revoke/deliver phases are reusable when we add it.*

This issue tracks the implementation.

## What it does

When a user initiates RP-Initiated Logout at `/oauth2/end-session`, instead of immediately redirecting to `post_logout_redirect_uri`, the OP:

1. Enumerates all clients that hold tokens for the current session (via `oauthAccessToken`)
2. Fetches the `frontchannel_logout_uri` for each registered client
3. Returns an HTML page containing one hidden 0x0 iframe per client URI
4. After all iframes have loaded (or a safety timeout), the browser redirects to `post_logout_redirect_uri`

Each RP's `frontchannel_logout_uri` endpoint is a simple GET that clears the RP's own session cookies and returns 200. No server-to-server call from the OP — the browser does the fan-out.

## Spec anchors

- Section 2 (RP Logout Functionality): RP metadata — `frontchannel_logout_uri`, `frontchannel_logout_session_required`. When `frontchannel_logout_session_required` is true, the OP must append `iss` and `sid` query parameters to the logout URI (if either is included, both MUST be).
- Section 3 (OP Logout Functionality): OP metadata — `frontchannel_logout_supported`, `frontchannel_logout_session_supported` — and the iframe fan-out mechanism ("a dynamically constructed page with HTML `<iframe src=` tags rendering each logged-in RP's logout URI").

## Proposed changes

**Schema** (`schema.ts`):
```ts
frontchannelLogoutUri: z.string().url().optional()
frontchannelLogoutSessionRequired: z.boolean().optional()
```

**Discovery** (`metadata.ts`):
```ts
frontchannel_logout_supported: true,
frontchannel_logout_session_supported: true,
```

**DCR / Admin** (`register.ts`, `oauthClient/index.ts`): accept and store `frontchannel_logout_uri` + `frontchannel_logout_session_required`, with the same SSRF validation (`isPublicRoutableHost`) applied to `backchannel_logout_uri` in #9304.

**`rpInitiatedLogoutEndpoint`** (`logout.ts`): after the session is deleted, enumerate clients that held tokens for that session and return an iframe HTML page instead of an immediate redirect. Append `iss` + `sid` to each URI whose client registered `frontchannel_logout_session_required: true`. The client-enumeration phase from #9304's revoke/deliver split is reusable here; the iframe rendering itself must live in the end-session endpoint since it needs a browser response (unlike the back-channel POST, which can fire from the `session.delete.before` hook on any session end).

## Contrast with back-channel

| | Back-channel (#9304) | Front-channel (this issue) |
|---|---|---|
| Delivery | OP -> RP (server POST) | Browser -> RP (iframe GET) |
| Works behind firewall | Yes | No |
| Cookie clearing | No | Yes (browser context) |
| RP server required | Yes | No (static endpoint) |

Both are valuable — front-channel is simpler for RPs that only need cookie-clearing and don't expose a server-side backchannel endpoint.

## Proof of concept

We've been running this pattern in production as a custom route that overrides the built-in `end-session` handler. The iframe coordination logic (XSS-escaped iframe HTML, `load`/`error` listeners, 3-second timeout fallback) works reliably across Chrome and Safari. Happy to share implementation details or draft a PR if useful.

## Related

- #9304 — back-channel logout (merged, explicitly deferred front-channel)
- #9178 — oauth-provider conformance checklist

shiminshen:
Closes #9922

Implements [OpenID Connect Front-Channel Logout 1.0](https://openid.net/specs/openid-connect-frontchannel-1_0.html) in `@better-auth/oauth-provider` — the browser-based sibling that #9304 explicitly deferred. When the end-user terminates their session with a browser navigation to `/oauth2/end-session`, the OP now responds with a logout page containing one hidden iframe per client that holds tokens on the session and registered a `frontchannel_logout_uri`. The browser fans the logout out by loading each RP's URI so the RP can clear its own session cookies in its own browser context — no server-to-server call involved. Once every iframe settles (or after a 3s safety timeout, so a hung RP cannot stall the user), the page redirects to the validated `post_logout_redirect_uri` when one was provided.

## What the client developer sees

```ts
await auth.api.adminCreateOAuthClient({
  headers,
  body: {
    redirect_uris: [redirectUri],
    enable_end_session: true,
    frontchannel_logout_uri: "https://rp.example.com/logout/frontchannel",
    frontchannel_logout_session_required: true,
  },
});
```

DCR and the client update endpoints accept the same two fields. When `frontchannel_logout_session_required` is `true`, the OP appends `iss` and `sid` query parameters to the iframe URL (§2: if either is included, both MUST be). The OP advertises `frontchannel_logout_supported` and `frontchannel_logout_session_supported` on both discovery endpoints; unlike the back-channel fields these are not gated on the JWT plugin, because front-channel logout signs nothing.

## Behavior preservation

- No front-channel clients on the session → immediate redirect, exactly as before.
- Fetch-style requests (`sec-fetch-mode: cors` or `accept: application/json`) keep the JSON contract — an iframe page is useless to them. Channel selection mirrors `handleRedirect`.
- Back-channel dispatch is untouched: it still fires from `session.delete.before`. Front-channel enumeration runs in the end-session endpoint *before* the session row is deleted, because token rows reference the session with `onDelete: "set null"` and the linkage disappears with the row.

## Rendering and security posture

- Registration validation mirrors `backchannel_logout_uri`: absolute http/https URL, no fragment (including the bare trailing `#` that `url.hash` misses), https on confidential clients with a loopback carve-out (RFC 8252 §7.3), and the same `isPrivateHostname` host policy — applied here because the OP directs every end-user's browser at this URI from its own page.
- Iframe `src` values are HTML-entity escaped; the deferred redirect URL is embedded JSON-encoded with `<` escaped so it cannot break out of the inline script context.
- The page is served with `Cache-Control: no-store` and `Referrer-Policy: no-referrer` — the page URL carries `id_token_hint` and must never leak to RPs via Referer.
- The iframes are plain HTML and fire without JavaScript; a meta-refresh covers the deferred redirect when scripting is disabled.

## Known limitation

Browsers that partition or block third-party cookies (Safari ITP, Chrome's third-party-cookie phase-out) may hide the RP's session cookie inside the iframe — front-channel logout is best-effort by design. The docs say to pair it with back-channel logout when reliable session termination is required.

## Test plan

- `frontchannel-logout.test.ts` (new): iframe-per-client rendering + session termination; `iss`/`sid` appended only for `frontchannel_logout_session_required` clients; validated and unregistered `post_logout_redirect_uri` handling; immediate-redirect preservation when no front-channel client is present; HTML escaping of iframe sources; JSON-contract preservation for fetch-style requests; rendering with `disableJwtPlugin: true`.
- `register.test.ts`: round-trip, fragment / http-on-confidential / private-tunneled-metadata-host rejection, public-client http loopback allowance, and registration allowed with `disableJwtPlugin` (contrast with back-channel).
- `metadata.test.ts`: both discovery documents advertise support; stays `true` with `disableJwtPlugin`.
- Full `@better-auth/oauth-provider` suite passes (437 tests); `pnpm typecheck` clean.

---
## Summary by cubic
Adds OIDC Front-Channel Logout 1.0 to `@better-auth/oauth-provider`, so browser navigations to `/oauth2/end-session` render a logout page with one hidden iframe per opted-in RP. Previously the endpoint always redirected; now it redirects only when no front-channel targets exist or after the iframes settle.

- Opt-in via client metadata: `frontchannel_logout_uri` and `frontchannel_logout_session_required` (adds `iss`/`sid` query params when true).
- Browser navigations get HTML with hidden iframes and a 3s bounded redirect to a validated `post_logout_redirect_uri`; fetch/JSON requests keep the JSON contract.
- `frontchannel_logout_uri` validation: absolute HTTPS only (no HTTP, including loopback), no fragment, no credentials, and non-public hosts are rejected.
- Discovery advertises `frontchannel_logout_supported` and `frontchannel_logout_session_supported`; support is independent of the JWT plugin.
- Security: iframe sources are escaped; responses use `Cache-Control: no-store` and `Referrer-Policy: no-referrer`.

- Migration
  - Clients must register an HTTPS `frontchannel_logout_uri` without fragments or credentials; HTTP URIs (including loopback) are rejected.
  - RPs that require session matching must set `frontchannel_logout_session_required: true` to receive `iss`/`sid`.

<sup>Written for commit b42b7a1287727cea76bc1aa74bb642cce81761cb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9923?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/oauth-provider-frontchannel-logout.md
+++ b/.changeset/oauth-provider-frontchannel-logout.md
@@ -1,0 +1,18 @@
+---
+"@better-auth/oauth-provider": minor
+---
+
+Add OIDC Front-Channel Logout 1.0 support to `@better-auth/oauth-provider`.
+
+When the end-user terminates their session with a browser navigation to `/oauth2/end-session`, the provider now responds with a logout page containing one hidden iframe per OAuth client that holds tokens on the session and has registered a `frontchannel_logout_uri`. The browser fans the logout out by loading each RP's URI so the RP can clear its own session cookies; once every iframe has settled — or after a 3-second safety timeout — the page redirects to the validated `post_logout_redirect_uri` when one was provided. Clients opt in by registering `frontchannel_logout_uri` (and optionally `frontchannel_logout_session_required`, which makes the provider append `iss` and `sid` query parameters) via DCR or the admin client endpoints.
+
+The URI is validated at registration with the same rules as `backchannel_logout_uri` (absolute http/https URL, no fragment, https for confidential clients with a loopback carve-out, non-public hosts rejected), but unlike back-channel logout it works with `disableJwtPlugin: true`, since nothing is signed.
+
+Behavior is unchanged when no front-channel client holds tokens on the session (immediate redirect, as before) and for fetch-style requests to `/oauth2/end-session`, which keep the JSON contract.
+
+Discovery documents at `/.well-known/openid-configuration` and `/.well-known/oauth-authorization-server` now advertise `frontchannel_logout_supported: true` and `frontchannel_logout_session_supported: true`.
+
+Schema changes on `@better-auth/oauth-provider`:
+
+- `oauthClient.frontchannelLogoutUri: string | null`
+- `oauthClient.frontchannelLogoutSessionRequired: boolean`
--- a/docs/content/docs/plugins/oauth-provider.mdx
+++ b/docs/content/docs/plugins/oauth-provider.mdx
@@ -706,6 +706,42 @@ The Logout Token carries the §2.4 claims (`iss`, `aud`, `iat`, `exp`, `jti`, `e
   The OP advertises `backchannel_logout_supported: true` and `backchannel_logout_session_supported: true` on both `.well-known/openid-configuration` and `.well-known/oauth-authorization-server`. RPs use these fields during dynamic client registration to decide whether to register a `backchannel_logout_uri`.
 </Callout>
 
+### Front-Channel Logout
+
+[Front-Channel Logout](https://openid.net/specs/openid-connect-frontchannel-1_0.html) is the browser-based sibling of Back-Channel Logout: when the end-user navigates to `/oauth2/end-session`, the OP responds with a page containing one hidden iframe per opted-in Relying Party, and the user's browser fans the logout out by loading each RP's `frontchannel_logout_uri`. The RP endpoint clears its own session cookies in the browser context — no server-to-server call is involved.
+
+To opt a client in, register a `frontchannel_logout_uri`:
+
+```ts title="admin-create-oauth.ts"
+await auth.api.adminCreateOAuthClient({
+  headers,
+  body: {
+    redirect_uris: [redirectUri],
+    enable_end_session: true,
+    frontchannel_logout_uri: "https://rp.example.com/logout/frontchannel", // [!code highlight]
+    frontchannel_logout_session_required: true, // [!code highlight]
+  }
+});
+```
+
+When `frontchannel_logout_session_required` is `true`, the OP appends `iss` and `sid` query parameters to the iframe URL so the RP can validate the request and terminate the right session.
+
+The `frontchannel_logout_uri` is validated at registration with the same rules as `backchannel_logout_uri`: absolute `http`/`https` URL, no fragment, `https` for confidential clients (loopback `http` such as `http://127.0.0.1:<port>` is allowed for local development), and non-public hosts (private, reserved, tunneled, cloud-metadata) are rejected. Unlike the back-channel URI it has no `jwt` plugin requirement, since nothing is signed.
+
+After the OP terminates the session (and back-channel notifications fire), it renders the iframe page; once every iframe has settled — or after a 3-second safety timeout, so a hung RP cannot stall the user — the page redirects to the validated `post_logout_redirect_uri` when one was provided. When no front-channel client holds tokens on the session, `/oauth2/end-session` redirects immediately, exactly as before.
+
+<Callout type="warn">
+  Browsers that partition or block third-party cookies (Safari ITP, Chrome's third-party-cookie phase-out) may hide the RP's session cookie inside the iframe, which makes front-channel logout best-effort by design. Pair it with Back-Channel Logout when you need reliable session termination.
+</Callout>
+
+<Callout type="info">
+  The iframe page is only rendered for browser navigation requests. Fetch-style requests to `/oauth2/end-session` (e.g. `authClient.oauth2.endSession()`) keep the JSON contract: the session still ends and back-channel notifications still fire, but no front-channel fan-out occurs.
+</Callout>
+
+<Callout type="info">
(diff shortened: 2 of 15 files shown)


shiminshen:
Rebased onto current `next`. This was a bigger move than a conflict resolution, so flagging what changed and one decision I'd like your call on.

`next` has moved a long way since this branch was opened (474 commits), and `#10812` rewrote RP-initiated logout around a confirmation step — `logout.ts` went from 661 to 1033 lines. Rather than resolve the conflicts mechanically I took the current file as the base and re-fitted the feature into the new flow, which turned out cleaner:

- The iframe fan-out now hooks into both termination paths — `completeConfirmedLogout` and the already-confirmed direct path in the endpoint — enumerating participating clients immediately before `deleteLogoutSession`, since token rows reference the session with `onDelete: "set null"`.
- My branch carried its own `escapeHtml` and an ad-hoc navigation check. `next` now has both (`escapeHtml`, `isBrowserNavigation`), so I dropped mine and use yours. `isBrowserNavigation` is stricter than what I had — it wants `sec-fetch-mode: navigate` or an HTML `accept` — which is the more accurate definition, so the page renders for real navigations only and fetch callers keep the JSON contract.
- `clientRegistrationRequestSchema` moved to `types/zod.ts` after this branch was written, so the new fields went there rather than into the inline schema that no longer exists. Same for the exhaustive `OAUTH_CLIENT_RECORD_FIELDS` map in `client-metadata.ts`.

**The decision I'd like your input on:** my original validation allowed `http` on public clients with a loopback carve-out for local development. `next`'s back-channel validation now enforces `https` unconditionally, with no carve-out, so I changed front-channel to match it exactly — absolute URL, no fragment, https only, no credentials, no private or reserved host. Consistency between the two logout URIs seemed clearly right, and the private-host rule matters here for a different reason than back-channel: the OP points every end-user's *browser* at this URI from its own logout page, so it's the browser-side analogue of your SSRF guard rather than the same one.

The cost is that an RP can no longer register an `http://localhost` front-channel URI for local development. If you'd rather keep a loopback carve-out on this one I'll add it back — it's a small change and I have no strong view, I just didn't want to quietly diverge from the policy you set for back-channel.

Verified on the rebased branch: `packages/oauth-provider` typecheck clean, `biome check --error-on-warnings` clean, and the full package suite green at 941/941. Front-channel logout is still unimplemented on `next`, so this remains additive — only back-channel is there today.

PR #9939, opened 2026-06-08
fix(device-authorization): make `schema` option optional under Zod v4

### LINKED ISSUE #9422: deviceAuthorization() throws ZodError on import when no schema option is passed

benpsnyder:
## Summary

`deviceAuthorization()` throws a `ZodError` at construction when called without a `schema` option — which is the common case. This fixes #9422 (closed but unresolved; still reproduces on `better-auth@1.6.14`).

## Root cause

In `deviceAuthorizationOptionsSchema`, the `schema` field is the only `z.custom()` field missing `.optional()`:

```ts
schema: z.custom<InferOptionSchema<typeof schema>>(() => true),
```

Every sibling field (`generateDeviceCode`, `generateUserCode`, `validateClient`, `onDeviceAuthRequest`, `verificationUri`) is `.optional()`. The plugin signature is `options: Partial<DeviceAuthorizationOptions> = {}` and the body reads `options?.schema`, so `schema` is clearly intended to be optional — but `deviceAuthorizationOptionsSchema.parse(options)` enforces it as required.

Under **Zod v4**, a bare `z.custom()` rejects `undefined` (`invalid_type`, `expected: "nonoptional"`), whereas Zod v3 let it through — so this surfaces as a hard regression for anyone resolving Zod v4.

## Reproduction

```ts
import { deviceAuthorization } from 'better-auth/plugins';

deviceAuthorization({ expiresIn: '30m', interval: '5s' });
// ZodError: [{ code: 'invalid_type', expected: 'nonoptional', path: ['schema'],
//   message: 'Invalid input: expected nonoptional, received undefined' }]
```

## Fix

Add `.optional()` to the `schema` field (one line) — matching every other field in the same schema. `mergeSchema(schema, options?.schema)` already handles `undefined`, so behavior is unchanged when `schema` is omitted.

Happy to add a changeset or adjust the commit/PR style if preferred.

---
## Summary by cubic
Prevents crashes in `better-auth` device authorization on Zod v4 when `schema` is omitted by making the `schema` option optional. Includes a changeset to publish this as a patch.

<sup>Written for commit 9b7c235ee8120779a28a028e6e3959090d46f778. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9939?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


aarmful:
running into the same issue currently. merge would be appreciated.

ping-maxwell (approved):

PR #9945, opened 2026-06-08
fix: pg timestamptz generation

### LINKED ISSUE #9920: CLI drizzle generator emits bare timestamp(...) for Postgres; should support timestamptz
## What

`packages/cli/src/generators/drizzle.ts` hardcodes the PG date column as:

\`\`\`ts
date: {
  sqlite: \`integer('\${name}', { mode: 'timestamp_ms' })\`,
  pg: \`timestamp('\${name}')\`,
  mysql: \`timestamp('\${name}', { fsp: 3 })\`,
},
\`\`\`

The PG branch produces `timestamp('${name}')`, which Drizzle maps to Postgres `timestamp without time zone`.

## Why this is wrong for these columns

Every date column emitted by the generator — `createdAt`, `updatedAt`, `expiresAt`, `lastUsedAt`, `accessTokenExpiresAt`, `refreshTokenExpiresAt` — represents an absolute-time event. `timestamp without time zone` is a wall-clock string type: it has no UTC offset, and Postgres reinterprets it at the connection's `TimeZone` setting on read. A session token stored as `expiresAt` from a process in `UTC` will be read back as a wall-clock string and re-interpreted as the reader's session timezone, silently producing a different absolute moment.

`timestamptz` (Drizzle: `timestamp(name, { withTimezone: true })`) is the correct mapping. Postgres internally stores the UTC instant and returns it correctly regardless of session timezone.

## What I expected

The generator emits `timestamp('${name}', { withTimezone: true })` for PG by default, or accepts an option for it.

## Workaround in the meantime

Hand-patch `auth-schema.ts` after every regen to add `{ withTimezone: true }`. Vulnerable to silent drift on the next regen.

## Proposed change

Either:
1. Switch the default PG mapping to `timestamp('${name}', { withTimezone: true })`. Existing users who care about wall-clock semantics for some reason can override per-column.
2. Add a CLI flag / config option (e.g. `dateMode: 'utc' | 'local'`) and default to `'utc'`.

Option 1 is the right default for an auth library: every column in the generated schema represents an absolute moment.

## Repro context

- better-auth `1.4.20`
- `@better-auth/cli generate` against Postgres + Drizzle adapter
- Generated columns observed with `pg_typeof(created_at)` returning `timestamp without time zone`

## Related community workaround

Examples of users patching the generated file by hand after regen, e.g. [this Nuxt + better-auth + drizzle guide](https://www.madebyfabian.com/blog/how-to-setup-better-auth-with-nuxt-3-and-drizzle).

yb175:
## Summary

This PR updates the PostgreSQL Drizzle schema generator to emit timezone-aware timestamps for generated date fields.

Previously, PostgreSQL date columns were generated as:

```ts
timestamp(name)
```

which maps to PostgreSQL `timestamp without time zone`.

This PR changes the mapping to:

```ts
timestamp(name, { mode: "date", withTimezone: true })
```

which maps to PostgreSQL `timestamp with time zone` (`timestamptz`).

### Why?

The generated authentication date fields (`createdAt`, `updatedAt`, `expiresAt`, `lastUsedAt`, `accessTokenExpiresAt`, and `refreshTokenExpiresAt`) represent absolute points in time rather than wall-clock values.

A dedicated integration test was added to reproduce the behavior using multiple runtime timezones (`UTC`, `America/New_York`, and `Asia/Kolkata`). The reproduction demonstrates that timezone-naive timestamps can be interpreted differently across processes running in different timezones, while `timestamptz` preserves the original instant consistently.

### Validation

This change is covered by:

* Generator and snapshot tests verifying PostgreSQL schemas now emit `{ mode: "date", withTimezone: true }`.
* Integration tests validating PostgreSQL column types.
* Cross-timezone roundtrip tests verifying timestamp stability across different runtime timezones.

Closes : #9920 

---
## Summary by cubic
Switches the PostgreSQL Drizzle schema generator to timezone-aware timestamps for all generated date fields to prevent cross-timezone drift. Adds end-to-end tests that verify column types and confirm consistent instants across timezones.

- **Bug Fixes**
  - Map PG dates from `timestamp(name)` to `timestamp(name, { mode: "date", withTimezone: true })` (uses `timestamptz`).
  - Added cross-timezone roundtrip tests (UTC, America/New_York, Asia/Kolkata) and schema type checks, plus a guard test to ensure non-date fields are unchanged.
  - Updated generator tests and snapshots; added changeset for patch releases to `@better-auth/oauth-provider`, `better-auth`, `@better-auth/passkey`, and `auth`.

- **Migration**
  - Regenerate your Drizzle schema and run a migration.
  - If migrating existing columns manually: `ALTER TABLE ... ALTER COLUMN ... TYPE timestamptz USING ...;`

<sup>Written for commit 910e4aae38e248cde4a5d296c17a6f0dffbd72ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9945?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/stale-spies-attack.md
+++ b/.changeset/stale-spies-attack.md
@@ -1,0 +1,8 @@
+---
+"@better-auth/oauth-provider": patch
+"better-auth": patch
+"@better-auth/passkey": patch
+"auth": patch
+---
+
+Updated the PostgreSQL Drizzle generator to generate date columns using timezone-aware `timestamp` columns (`withTimezone: true`). This resolves timezone parsing discrepancies and silent hour drifts when clients in different default timezones read and write date records.
--- a/e2e/adapter/test/drizzle-adapter/adapter.drizzle.timezone.test.ts
+++ b/e2e/adapter/test/drizzle-adapter/adapter.drizzle.timezone.test.ts
@@ -16,8 +16,15 @@ const helperPath = path.join(
 );
 
 function runHelper(tz: string, mode: string): any {
-	const cmd = `TZ=${tz} TEST_MODE=${mode} node ${tsxCliPath} ${helperPath}`;
-	const output = execSync(cmd, { encoding: "utf8" });
+	const cmd = `node ${tsxCliPath} ${helperPath}`;
+	const output = execSync(cmd, {
+		encoding: "utf8",
+		env: {
+			...process.env,
+			TZ: tz,
+			TEST_MODE: mode,
+		},
+	});
 	return JSON.parse(output.trim());
 }
 
--- a/e2e/adapter/test/drizzle-adapter/adapter.drizzle.timezone.test.ts
+++ b/e2e/adapter/test/drizzle-adapter/adapter.drizzle.timezone.test.ts
@@ -1,0 +1,97 @@
+/**
+ * @see https://github.com/better-auth/better-auth/issues/9920
+ */
+// cspell:ignore Kolkata
+import { execSync } from "node:child_process";
+import * as path from "node:path";
+import { describe, expect, it } from "vitest";
+
+const tsxCliPath = path.resolve(
+	import.meta.dirname,
+	"../../../../packages/cli/node_modules/tsx/dist/cli.mjs",
+);
+const helperPath = path.join(
+	import.meta.dirname,
+	"fixtures/timezone-helper.ts",
+);
+
+function runHelper(tz: string, mode: string): any {
+	const cmd = `TZ=${tz} TEST_MODE=${mode} node ${tsxCliPath} ${helperPath}`;
+	const output = execSync(cmd, { encoding: "utf8" });
+	return JSON.parse(output.trim());
+}
+
+describe("PostgreSQL Timezone Behavior & Schema Type Verification", () => {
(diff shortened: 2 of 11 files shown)


bytaesu (changes requested):
Hi @yb175,

This PR affects the core schema. To preserve backward compatibility, we avoid core schema changes in patch and minor releases. We'll keep that issue open and revisit it as part of a future planned change.

If you have a clean solution that maintains backward compatibility, feel free to suggest it 😁

bytaesu:
Thanks for working on this, but I've decided to close this PR because of https://github.com/better-auth/better-auth/pull/9945#pullrequestreview-4454001296

yb175:
Thanks for the review!
Yeah, migration impact was the biggest thing I kept running into too. I'll dig around for a backward-compatible path and drop it in the thread if I find something.
Appreciate it 😁

PR #9947, opened 2026-06-08
fix(sign-in): persist idToken, refreshToken, accessTokenExpiresAt on idToken social sign-in

### LINKED ISSUE #9943: sign-in/social idToken flow does not persist refreshToken / idToken / accessTokenExpiresAt to the account
First of all, thank you for Better Auth — it's been a pleasure to build our Cognito (email/password via SRP) login flow on top of it.

I searched existing issues before opening this and couldn't find an exact match for this specific code path. The closest relatives are the confirmed token-persistence bugs #8191 and #7999, which share the same shape (a field selectively omitted while building the account object). I've tried to include everything needed to investigate so this is low-effort to triage, and I'm very happy to send a PR if the suggested fix looks right to you.

## Summary

When signing in via `authClient.signIn.social({ provider, idToken: { token, accessToken, refreshToken, expiresAt } })`, the `refreshToken`, `idToken`, and `accessTokenExpiresAt` are **not** persisted to the account, even though they are provided by the client and accepted by the endpoint schema. As a result, a subsequent `auth.api.refreshToken({ body: { providerId } })` fails with `REFRESH_TOKEN_NOT_FOUND`.

## Root cause

In `packages/better-auth/src/api/routes/sign-in.ts`, the idToken branch passes only `accessToken` to `handleOAuthUserInfo`'s `account` object (around L309–313 on `main`):

```ts
const data = await handleOAuthUserInfo(c, {
  userInfo: { /* ... */ },
  account: {
    providerId: provider.id,
    accountId: String(userInfo.user.id),
    accessToken: c.body.idToken.accessToken,
    // idToken / refreshToken / accessTokenExpiresAt are not forwarded here
  },
  // ...
});
```

Two things make this look like an oversight rather than intentional stateless behavior:

1. The immediately preceding `getUserInfo` call (around L275–280) **does** receive `refreshToken`, so the value is available in scope:

   ```ts
   const userInfo = await provider.getUserInfo({
     idToken: token,
     accessToken: c.body.idToken.accessToken,
     refreshToken: c.body.idToken.refreshToken, // present here
     user: c.body.idToken.user,
   });
   ```

2. The request body schema accepts `idToken.refreshToken` and `idToken.expiresAt`, and `handleOAuthUserInfo` (`packages/better-auth/src/oauth2/link-account.ts`) reads `account.refreshToken` / `account.idToken` / `account.accessTokenExpiresAt` in all three persistence paths (create / link / sign-in update). So when the caller omits them here, they're stored as `undefined`.

The regular OAuth redirect (callback) flow persists all of these, which is why the asymmetry stands out.

> I also confirmed the `idToken` branch on `main` is identical to `v1.6.13` for these lines (line numbers above are from `main`), so this isn't already fixed on the latest as far as I can tell.

## Reproduction

1. Configure a provider that supports idToken sign-in (we use Cognito), with `account.storeAccountCookie: true`.
2. On the client, obtain `idToken` / `accessToken` / `refreshToken` and call:
   ```ts
     provider: "cognito",
     idToken: { token, accessToken, refreshToken, expiresAt },
   });
   ```
3. Inspect the persisted account (DB row, or the `account_data` cookie) → `refreshToken` is `null`/`undefined`.
4. Call `POST /api/auth/refresh-token` with `{ providerId: "cognito" }` → `REFRESH_TOKEN_NOT_FOUND`.

## Current vs. expected behavior

- **Current:** `refreshToken` / `idToken` / `accessTokenExpiresAt` provided to `signIn.social({ idToken })` are silently dropped at sign-in, so refresh later fails with `REFRESH_TOKEN_NOT_FOUND`.
- **Expected:** the same tokens are persisted to the account on idToken sign-in (consistent with the OAuth callback flow), so `auth.api.refreshToken(...)` / `getAccessToken(...)` can use them.

## Suggested fix

Forward the remaining fields, mirroring the callback flow:

```ts
account: {
  providerId: provider.id,
  accountId: String(userInfo.user.id),
  accessToken: c.body.idToken.accessToken,
  idToken: token,
  refreshToken: c.body.idToken.refreshToken,
  ...(c.body.idToken.expiresAt != null
    ? { accessTokenExpiresAt: new Date(c.body.idToken.expiresAt * 1000) }
    : {}),
},
```

(`idToken.expiresAt` is documented/used as a unix timestamp in seconds, hence `* 1000`.) Happy to open a PR against `main` with a regression test (`@see` this issue) if you'd like.

## Version

`better-auth@1.6.13` (and the relevant lines are identical on `main`).

## System info

```bash
# Environment where the bug was observed
better-auth: 1.6.13
framework: Next.js (App Router) with the nextCookies() plugin
node: 22.x
provider: Cognito (custom email/password via SRP -> idToken sign-in)
```

(I wasn't able to attach `npx auth info --json` output for this report; happy to provide it if helpful.)

## Affected area

Backend (the `sign-in/social` route handler).

## Auth config (sanitized)

```typescript
betterAuth({
  account: {
    storeStateStrategy: "cookie",
    storeAccountCookie: true,
    accountLinking: { enabled: true, trustedProviders: ["cognito"] },
  },
  socialProviders: {
    cognito: {
      clientId: [COGNITO_CLIENT_ID, COGNITO_SRP_CLIENT_ID],
      clientSecret: COGNITO_CLIENT_SECRET,
      domain: COGNITO_DOMAIN,
      region: COGNITO_REGION,
      userPoolId: COGNITO_USER_POOL_ID,
      requireClientSecret: true,
      disableDefaultScope: true,
      scope: ["openid", "profile", "email", "aws.cognito.signin.user.admin"],
    },
  },
  plugins: [nextCookies()],
});
```

## Additional context

- Reproducible locally; not browser-specific.
- Related (not duplicates): #8191 and #7999 are the same class of "token field omitted/overwritten in a persistence path" bug, both confirmed and fixed. #6379 (Cognito `account_data` cookie size) is an adjacent concern but a different root cause.

Thanks again for maintaining this project, and please let me know if a PR would help.

TanishValesha:
Fixes #9943 

## Summary

Fixes a bug where tokens provided to `authClient.signIn.social({ idToken })` were not persisted to the associated account. As a result, refresh flows could fail with `REFRESH_TOKEN_NOT_FOUND`.

Specifically, the following values were not being saved:

* `refreshToken`
* `idToken`
* `expiresAt`

## Root Cause

The ID token sign-in path in `sign-in.ts` only forwarded `accessToken` to `handleOAuthUserInfo(...)`.

However, `handleOAuthUserInfo(...)` is responsible for persisting OAuth credentials and expects additional token fields when available:

* `idToken`
* `refreshToken`
* `accessTokenExpiresAt`

Because these values were omitted, they were treated as `undefined` and never persisted to the `Account` record.

### Implementation

Updated the ID token sign-in flow to forward:

* `idToken`
* `refreshToken`
* `accessTokenExpiresAt`

`expiresAt` is converted from a Unix timestamp (seconds) to a JavaScript `Date` before being passed to `handleOAuthUserInfo(...)`.

## Files Modified

* `sign-in.ts`

---
## Summary by cubic
Persisted `idToken`, `refreshToken`, and `accessTokenExpiresAt` during ID-token social sign-in in `better-auth`, fixing failed refresh flows (e.g., REFRESH_TOKEN_NOT_FOUND).
Improves token storage by forwarding all available fields to the OAuth persistence logic.

- **Bug Fixes**
  - Forward `idToken`, `refreshToken`, and `accessTokenExpiresAt` to the OAuth handler in the ID-token social sign-in path.
  - Convert `expiresAt` (seconds) to a JS `Date` for `accessTokenExpiresAt` when present.

<sup>Written for commit ba5e559f03f7f5585837a83080c91846af499a3d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9947?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/bumpy-tigers-enjoy.md
+++ b/.changeset/bumpy-tigers-enjoy.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Persist `idToken`, `refreshToken`, and `accessTokenExpiresAt` on id-token social sign-in.
--- a/packages/better-auth/src/api/routes/sign-in.ts
+++ b/packages/better-auth/src/api/routes/sign-in.ts
@@ -310,6 +310,15 @@ export const signInSocial = <O extends BetterAuthOptions>() =>
 						providerId: provider.id,
 						accountId: String(userInfo.user.id),
 						accessToken: c.body.idToken.accessToken,
+						idToken: token,
+						refreshToken: c.body.idToken.refreshToken,
+						...(c.body.idToken.expiresAt != null
+							? {
+									accessTokenExpiresAt: new Date(
+										c.body.idToken.expiresAt * 1000,
+									),
+								}
+							: {}),
 					},
 					callbackURL: c.body.callbackURL,
 					disableSignUp:


TanishValesha:
@gustavovalverde please look at the PR whenever possible

dcustodio (commented):
Hi @TanishValesha , I am also having this issue with the missing `refreshToken` so I tried your PR. In my case I also needed to modify the file `packages/better-auth/src/api/routes/account.ts` so that the `/account-info` endpoint also returns the refreshToken. 
Anyway, it would be nice that this PR gets some attention from the mantainers!

PR #9957, opened 2026-06-09
docs(two-factor): document newSession is null during 2FA challenge

Paola3stefania:
## Summary

Follow-up to #9639. 

That PR added `ctx.context.setNewSession(null)` to the
2FA sign-in challenge path. When a 2FA-enabled user signs in via a credential
endpoint (`/sign-in/{email,username,phone-number}`), the pending session is
deleted and `newSession` is reset to `null` — there is no authenticated
session until the second factor is verified.

This is the correct state, but it was a silent behavior change: server-side
`after` hooks that read `ctx.context.newSession` after sign-in previously got
the (now-deleted) credential session and now get `null`. Unguarded consumer
hooks doing `newSession.user.x` throw `Cannot read properties of null
(reading 'user')` while a 2FA challenge is in flight.

All built-in plugins that read `newSession` (multi-session, anonymous, jwt,
mcp, oidc-provider) already null-check, so this is purely documentation +
clarity, no behavior change.

### DIFF
--- a/.changeset/two-factor-newsession-null.md
+++ b/.changeset/two-factor-newsession-null.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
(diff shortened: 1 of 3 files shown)


bytaesu:
@Bekacru @Paola3stefania we can use `skip-changeset` label

bytaesu:
Let me drop changeset with follow-up PR

Paola3stefania:
Ty @bytaesu missed that one !

PR #9974, opened 2026-06-10
fix: address bug findings across packages 

Bekacru:


packages/core/src/social-providers/reddit.ts
bytaesu, line 1: https://better-auth.com/docs/concepts/oauth#handling-providers-without-email

PR #9979, opened 2026-06-10
fix(account): resolve stateless account cookies across instances

### LINKED ISSUE #9978: Stateless mode: getAccessToken returns ACCOUNT_NOT_FOUND when the account cookie's userId differs from the session

### Reproduction

Running Better Auth statelessly (no `database`, no `secondaryStorage`) with `account.storeAccountCookie` and an OAuth provider, across more than one runtime instance (e.g. serverless):

1. Configure auth without a database, with `genericOAuth` and `account: { storeStateStrategy: "cookie", storeAccountCookie: true }` and `session.cookieCache` enabled.
2. Sign in. The OAuth callback creates a user with a freshly generated `user.id` and writes that id to both the session cache and the `account_data` cookie.
3. With no persistent store, another instance handling a later request (or a re-auth) generates a *different* `user.id` for the same provider account.
4. Once the browser holds a session cookie minted by one instance and an `account_data` cookie written by another, `auth.api.getAccessToken({ body: { providerId } })` (and `accountInfo`) throw:

```
APIError: Account not found  { status: "BAD_REQUEST", code: "ACCOUNT_NOT_FOUND" }
```

Minimal local repro: sign the same provider account in on two independently-constructed stateless instances (same secret), then call `getAccessToken` on a third fresh instance with the session cookie from the first + the `account_data` cookie from the second. A failing test reproducing exactly this is included in the linked PR.

Since 1.6.0 (#8786) the account-cookie match in `getValidAccessToken` additionally requires `accountData.userId === resolvedUserId`. In stateless mode the user id isn't stable across instances, so a valid encrypted account cookie gets rejected, `findAccounts` has nothing to fall back to, and the request fails with `Account not found`. `setCookieCache` compounds it by deleting the account cookie on the same mismatch. Before 1.6.0 the cookie was matched by provider id and this worked.

Expected: in a stateless setup the encrypted account cookie is the source of truth, so a valid cookie should resolve the token by provider identity rather than failing on a user id it cannot reliably validate.

1.6.16 (regressed in 1.6.0)

### System info

```
Node 22 · Next.js App Router (serverless) · better-auth 1.6.16
Stateless: no database / no secondaryStorage
genericOAuth provider, account.storeAccountCookie: true, session.cookieCache enabled
```

Backend

```typescript
import { genericOAuth } from "better-auth/plugins/generic-oauth";

  secret: process.env.AUTH_SECRET,
  // no database -> stateless
  session: { cookieCache: { enabled: true, maxAge: 60 * 60 * 24 } },
  account: { storeStateStrategy: "cookie", storeAccountCookie: true },
  plugins: [
    genericOAuth({
      config: [{ providerId: "keycloak", clientId: "...", clientSecret: "...", discoveryUrl: "..." }],
    }),
  ],
});
```

### Additional context

Reproducible against the latest release. A fix with a regression test is proposed in the linked PR.

SferaDev:
Fixes #9978.

Stateless OAuth requests can land on different instances after sign-in. In that setup the session cookie and OAuth account cookie are both valid, but their local user ids can differ because there is no durable account row to bind them.

This keeps the strict user-id binding when a database-backed account store exists. Without a database, account selection is constrained by provider identity and optional provider account id, then reused by access-token, refresh-token, account-info, and session-cache refresh flows.

Secondary storage remains a session store, not an account store, so the store capability helpers keep those decisions separate.


gustavovalverde:
@cubic-dev-ai review

PR #9982, opened 2026-06-10
fix(mcp): honor discovery metadata overrides

### LINKED ISSUE #9961: mcp plugin: OAuth discovery issuer/endpoints/authorization_servers pinned to baseURL with no override — breaks split-origin auth servers
## Summary

The `mcp` plugin's authorization-server metadata (`/.well-known/oauth-authorization-server`) derives its `issuer` and all endpoint URLs from `ctx.context.baseURL`, and the protected-resource metadata hardwires `authorization_servers: [origin]` to that same base. There is no working way to override them, so you cannot host the OAuth authorization server at a different origin than `baseURL`.

This breaks any deployment where the Better Auth routes are served on a different host than the app's `baseURL` — e.g. `@convex-dev/better-auth`, where `/api/auth/*` is served on the Convex `*.convex.site` domain while `baseURL` is the frontend origin. Browser MCP clients (claude.ai, MCP Inspector) then receive discovery pointing at an origin that doesn't serve `/api/auth`, and the OAuth flow can't complete.

## Details

`getMCPProviderMetadata(ctx, options)` (`src/plugins/mcp/index.ts`):

```ts
const issuer = ctx.context.options.baseURL;
const baseURL = ctx.context.baseURL;
return {
  issuer,
  authorization_endpoint: `${baseURL}/mcp/authorize`,
  // …
  ...options?.metadata,   // <-- see below
};
```

The `/.well-known/oauth-authorization-server` handler calls `getMCPProviderMetadata(c, options)` where `options` is the **`mcp` plugin options** object (`{ loginPage, resource, oidcConfig }`). That type has **no top-level `metadata`** — metadata lives under `oidcConfig.metadata`. So `...options?.metadata` is always `undefined`, and there is no override path: `issuer` + endpoints are pinned to `baseURL`.

By contrast, `getMCPProtectedResourceMetadata` *does* read `options?.oidcConfig?.metadata?.jwks_uri` / `scopes_supported` — so the two metadata builders are inconsistent about where they read overrides. And its `authorization_servers: [origin]` is hardwired with no override at all.

## Impact

- AS discovery advertises the wrong (often unreachable) origin when the auth server's public origin ≠ `baseURL`.
- `authorization_servers` in the protected-resource doc can't be pointed at the real AS.
- Net: browser MCP clients can't discover the right endpoints; the OAuth handshake fails.

## Suggested fix

1. In the `/.well-known/oauth-authorization-server` handler, pass `options.oidcConfig` to `getMCPProviderMetadata` (so `...options.metadata` resolves to `oidcConfig.metadata`) — or have `getMCPProviderMetadata` read `options?.oidcConfig?.metadata` for consistency with `getMCPProtectedResourceMetadata`.
2. Make `issuer` and `authorization_servers` overridable (e.g. honor `oidcConfig.metadata.issuer`, and derive `authorization_servers` from it) instead of pinning to `ctx.context.baseURL`'s origin.

## Workaround (for reference)

Serve a corrected `/.well-known/oauth-authorization-server` document from the host (with `issuer`/endpoints at the real auth origin) and point the resource-server's `authorization_servers` at it. Works, but it's a shim around the metadata builder.

Environment: `better-auth@1.4.9`, used via `@convex-dev/better-auth@0.10.10`.

RitwijParmar:
Fixes #9961.

The MCP authorization-server discovery handler was passing the whole MCP plugin options object into getMCPProviderMetadata. That function expects OIDC options, so oidcConfig.metadata overrides were never applied. In split-origin deployments this can advertise the frontend/baseURL instead of the real authorization server.

This PR does three things.

- Passes options.oidcConfig to getMCPProviderMetadata so issuer and endpoint overrides apply.
- Makes protected-resource authorization_servers derive from explicit metadata, then metadata.issuer, then the current origin as fallback.
- Reuses the same metadata object for jwks_uri and scopes_supported in protected-resource discovery.

I added regression tests for:

- MCP authorization-server discovery using oidcConfig.metadata overrides.
- Protected-resource discovery pointing authorization_servers at the configured issuer.
- Explicit protected-resource authorization_servers metadata when deployments need multiple AS entries.

Validation

- bunx @biomejs/biome check packages/better-auth/src/plugins/mcp/index.ts packages/better-auth/src/plugins/mcp/mcp.test.ts passed.
- git diff --check passed.

I tried the focused Vitest command, but bunx vitest fails before tests run because the temporary Rolldown native binding is rejected by macOS code signing in this environment.

---
## Summary by cubic
Fixes MCP discovery to honor OIDC metadata overrides in `better-auth`, ensuring the correct issuer and endpoints are advertised in split-origin deployments. Updates protected-resource discovery to prefer configured authorization servers and reuse OIDC metadata when present.

- **Bug Fixes**
  - Passes `oidcConfig` to `getMCPProviderMetadata` so issuer and endpoint overrides apply.
  - Derives `authorization_servers` from explicit metadata, then `metadata.issuer`, falling back to the current origin.
  - Reuses provided metadata for `jwks_uri` and `scopes_supported` in protected-resource discovery.

<sup>Written for commit c66f81776445ae9aac0bb0172d383e1708c7777f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/9982?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/better-auth/src/plugins/mcp/index.ts
+++ b/packages/better-auth/src/plugins/mcp/index.ts
@@ -113,12 +113,20 @@ export const getMCPProtectedResourceMetadata = (
 ) => {
 	const baseURL = ctx.context.baseURL;
 	const origin = new URL(baseURL).origin;
+	const metadata = options?.oidcConfig?.metadata as
+		| (Partial<OIDCMetadata> & {
+				authorization_servers?: string[];
+		  })
+		| undefined;
+	const authorizationServers =
+		metadata?.authorization_servers ??
+		(metadata?.issuer ? [metadata.issuer] : [origin]);
 
 	return {
 		resource: options?.resource ?? origin,
-		authorization_servers: [origin],
-		jwks_uri: options?.oidcConfig?.metadata?.jwks_uri ?? `${baseURL}/mcp/jwks`,
-		scopes_supported: options?.oidcConfig?.metadata?.scopes_supported ?? [
+		authorization_servers: authorizationServers,
+		jwks_uri: metadata?.jwks_uri ?? `${baseURL}/mcp/jwks`,
+		scopes_supported: metadata?.scopes_supported ?? [
 			"openid",
 			"profile",
 			"email",
@@ -273,7 +281,7 @@ export const mcp = (options: MCPOptions) => {
 				},
 				async (c) => {
 					try {
-						const metadata = getMCPProviderMetadata(c, options);
+						const metadata = getMCPProviderMetadata(c, options.oidcConfig);
 						return c.json(metadata);
 					} catch (e) {
 						console.log(e);
--- a/packages/better-auth/src/plugins/mcp/mcp.test.ts
+++ b/packages/better-auth/src/plugins/mcp/mcp.test.ts
@@ -6,7 +6,7 @@ import { getTestInstance } from "../../test-utils/test-instance";
 import { genericOAuth } from "../generic-oauth";
 import { genericOAuthClient } from "../generic-oauth/client";
 import { jwt } from "../jwt";
-import type { Client } from "../oidc-provider/types";
+import type { Client, OIDCMetadata } from "../oidc-provider/types";
 import { mcp, withMcpAuth } from ".";
 
 // Pre-verifies any user the RP creates via OAuth signup so the existing-user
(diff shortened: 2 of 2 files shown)


bytaesu:
Hi @RitwijParmar,

I just left a comment. Please check https://github.com/better-auth/better-auth/issues/9961#issuecomment-4677612486

PR #9987, opened 2026-06-10
fix: jwks caching, oauth id mapping, team invitations, account cookie, and scim deprovision bugs

bytaesu:
Several auth edge cases could leave cached credentials, provider identity, or organization membership state out of sync with the authoritative row.

This patch keeps each fix at the boundary that owns that state:

- JWKS verification scopes caches by source, treats JWS `kid` as optional, and rate-limits no-`kid` forced refetches so key rotation still gets a retry without turning repeated invalid tokens into JWKS fetches. Function-based JWKS sources can still opt into a stable cache key without retaining per-request closures.
- Generic OAuth keeps non-empty provider ids stable, falls back to OIDC UserInfo `sub` when `id` is empty or null in both callback and provider-wrapper paths, and leaves missing ids absent so `mapProfileToUser` can derive the account id before the callback guard requires a non-empty value.
- Invitation acceptance validates and writes from the claimed invitation row after the pending-to-accepted transition. Team deletion removes deleted team ids from pending invitations inside the same transaction as the team delete.
- Account-cookie refresh skips stale request-cookie sync when the response already contains a fresh account cookie.
- Organization-scoped SCIM deletes keep member hooks around a single transaction that removes the organization membership, team memberships, and SCIM account link.

### DIFF
--- a/packages/better-auth/src/plugins/generic-oauth/routes.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/routes.ts
@@ -25,6 +25,18 @@ import { isAPIError } from "../../utils/is-api-error";
 import { GENERIC_OAUTH_ERROR_CODES } from "./error-codes";
 import type { GenericOAuthOptions } from "./types";
 
+export type GenericOAuthUserInfo = Omit<OAuth2UserInfo, "id"> & {
+	id?: string | number | null | undefined;
+	sub?: string | number | null | undefined;
+	[key: string]: unknown;
+};
+
+function isNonEmptyOAuthId(
+	id: string | number | null | undefined,
+): id is string | number {
+	return id !== undefined && id !== null && id !== "";
+}
+
 const signInWithOAuth2BodySchema = z.object({
 	providerId: z.string().meta({
 		description: "The provider ID for the OAuth provider",
@@ -426,7 +438,7 @@ export const oAuth2Callback = (options: GenericOAuthOptions) =>
 						providerConfig.getUserInfo
 							? await providerConfig.getUserInfo(tokens)
 							: await getUserInfo(tokens, finalUserInfoUrl)
-					) as OAuth2UserInfo | null;
+					) as GenericOAuthUserInfo | null;
 					if (!userInfo) {
 						redirectOnError(ctx, resolvedErrorURL, "user_info_is_missing");
 					}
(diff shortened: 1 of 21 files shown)


packages/scim/src/routes.ts
```diff
@@ -1147,14 +1148,35 @@ export const deleteSCIMUser = (authMiddleware: AuthMiddleware) =>
...
+					await orgOptions?.organizationHooks?.beforeRemoveMember?.({
+						member,
+						user,
+						organization,
 					});
+				}
+
+				await ctx.context.adapter.transaction(async () => {
+					if (member) {
+						await orgAdapter.deleteMember({
```
gustavovalverde: `orgAdapter.deleteMember()` opens its own transaction, so calling it inside the SCIM transaction can commit member/team cleanup even if the account delete later fails. Please make this cleanup use the same transaction boundary.

packages/better-auth/src/plugins/generic-oauth/routes.ts
```diff
@@ -813,18 +813,17 @@ export async function getUserInfo(
...
-	// Without a stable subject id, the account would be stored under an empty
-	// id and could be resolved by a different user on the same provider.
-	if (subjectId === undefined || subjectId === null || subjectId === "") {
-		return null;
-	}
+	const profile = userInfo.data;
+	// Non-empty `id` wins over `sub` to keep stored account ids stable. Empty
+	// is allowed here: `mapProfileToUser` and the callback guard decide later.
+	const subjectId =
+		profile?.id !== undefined && profile.id !== "" ? profile.id : profile?.sub;
```
gustavovalverde: `id: null` should fall back to `sub` like an empty id does. As written, `id: null, sub: "abc"` becomes `id: ""` later and the callback fails with `id_is_missing`.

packages/better-auth/src/plugins/organization/routes/crud-invites.ts
```diff
@@ -732,6 +732,28 @@ export const acceptInvitation = <O extends OrganizationOptions>(options: O) =>
...
 
+			if (
+				ctx.context.orgOptions.teams &&
+				ctx.context.orgOptions.teams.enabled &&
+				"teamId" in invitation &&
+				invitation.teamId
+			) {
+				// Validate teams before consuming the invitation so a failed
+				// check leaves it pending and retryable.
+				for (const teamId of (invitation.teamId as string).split(",")) {
```
gustavovalverde: This validates `invitation.teamId`, but the writes use `acceptedI.teamId` after `updateInvitation`. If those ids change in between, the invite can be consumed before the actual team ids are validated.

gustavovalverde (changes requested):

gustavovalverde (approved):

PR #10005, opened 2026-06-11
feat: soft-delete support

### LINKED ISSUE #9099: feat(admin): user soft delete with anonymization and restore

When an admin deletes a user via `admin.removeUser`, the user record is permanently removed from the database. This creates several problems:

- **Audit trail loss:** any foreign-key references (logs, activity records, etc.) that pointed to that user become dangling or must be cascade-deleted, destroying historical data.
- **No recourse:** once deleted there is no way to recover the user or re-enable access without manually reconstructing the record.
- **GDPR/privacy compliance:** a hard delete satisfies the “right to erasure” but leaves no trace for internal audit requirements. A soft delete with anonymization satisfies both: the record is retained (for referential integrity) but PII is scrubbed.

[PR 9093](https://github.com/better-auth/better-auth/pull/9093)

Add a `softDelete` option to `user.deleteUser` (the existing user self-delete config object) and to the admin plugin’s `removeUser` endpoint.

When `softDelete: true`:

1. **`admin.removeUser`**
   - Instead of deleting the user row, sets `deletedAt = now()` and overwrites PII fields (`name`, `email`, `image`) with anonymous placeholders.
   - Linked accounts and sessions are still removed.
   - A custom `anonymizeUser(user)` callback can override the default anonymization.
   - Reserved system fields (`id`, `createdAt`, `updatedAt`, `emailVerified`, `deletedAt`) are protected and cannot be overwritten by the callback.

2. **Sign-in blocked**
   - A `session.create.before` hook rejects session creation for any user whose `deletedAt` is set, returning `FORBIDDEN / USER_DELETED`.

3. **`admin.listUsers`**
   - Excludes soft-deleted users by default.
   - An `includeDeleted` query param allows admins to include them.

4. **`admin.restoreUser`**
   - New endpoint that clears `deletedAt`, making the user record active again.
   - Because accounts are removed during deletion, the admin must separately set a new password or the user must re-link an OAuth provider to regain sign-in.

5. **Self-delete (`user.deleteUser`)**
   - The same `softDelete` flag applies to the user’s own `deleteUser` flow (immediate delete and email-confirmation-based delete).

6. **DB migration**
   - `deletedAt` is added to the core user table schema (via `getAuthTables`) when `softDelete: true`, so `migrate` / `generate` create the column automatically.

---

## Schema / API Surface

```ts
// auth.ts
  user: {
    deleteUser: {
      enabled: true,
      softDelete: true,
      anonymizeUser: (user) => ({   // optional
        name: "Deleted User",
        email: `deleted-${user.id}@deleted.invalid`,
      }),
    },
  },
  plugins: [admin()],
});

// client
await authClient.admin.removeUser({ userId });   // soft-deletes
await authClient.admin.restoreUser({ userId });  // clears deletedAt
await authClient.admin.listUsers({ query: { includeDeleted: true } });
```

_No response_

### Additional context

- The `deletedAt` field is contributed to the user table both by the admin plugin schema (for admin-plugin users) and conditionally by `getAuthTables` (for non-admin-plugin users who still use `softDelete`).
- `z.coerce.boolean()` is avoided for the `includeDeleted` query param because it mishandles the string `"false"`; a `z.union` with explicit `"true"` / `"false"` transform is used instead.
- All new behaviour is covered by tests:
  - sign-in blocked
  - `listUsers` filtering
  - `restoreUser`
  - custom `anonymizeUser`
  - reserved-field protection
  - permission checks
- Follows the same permission model as existing admin endpoints:
  - `user: ["delete"]` for `removeUser`
  - new `user: ["restore"]` for `restoreUser`

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/9099

### DIFF
--- a/.changeset/odd-plums-relate.md
+++ b/.changeset/odd-plums-relate.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/core": minor
+---
+
+feat: soft-delete support
--- a/docs/content/docs/concepts/database.mdx
+++ b/docs/content/docs/concepts/database.mdx
@@ -1012,6 +1012,10 @@ To add new tables and columns to your database, you have two options:
 
 Both methods ensure your database schema stays up to date with your plugins' requirements.
 
+Plugins may also mark the models they define or extend as soft-deletable.
+In that case the table carries a nullable `deletedAt` column, and deleting through that plugin flags rows as deleted instead of removing them.
+This is controlled by the plugin author, not by your auth config - see [Soft Delete](/docs/concepts/plugins#soft-delete) in the plugin development docs.
+
 ## Experimental Joins
 
 Since Better-Auth version `1.4` we've introduced experimental database joins support.
--- a/docs/content/docs/concepts/plugins.mdx
+++ b/docs/content/docs/concepts/plugins.mdx
@@ -209,6 +209,8 @@ const myPlugin = (opts: PluginOptions) => {
 }
 ```
 
+`softDelete`: if the model should support soft deletion. (default: `false`) See [Soft Delete](#soft-delete) below.
+
 if you add additional fields to a `user` or `session` table, the types will be inferred automatically on `getSession` and `signUpEmail` calls.
 
 ```ts title="plugin.ts"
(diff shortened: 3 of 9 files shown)


ping-maxwell:
@greptile review

PR #10017, opened 2026-06-12
docs: render fallback for MDX components missing from synced beta content

Bekacru:
## Problem

The Vercel docs build on `main` failed while prerendering `/docs/beta/authentication/apple`:

```
Error: Expected component `GenerateAppleJwt` to be defined: you likely forgot to import, pass, or provide it.
```

**Root cause:** `/docs/beta/*` content is synced from the `next` branch at build time (`scripts/sync-beta-content.ts`), while the app code rendering it comes from the deploying branch. #9938 removed the `GenerateAppleJwt` component from `main`'s app code, but `next`'s `apple.mdx` still used `<GenerateAppleJwt />` at deploy time, so the beta page crashed the entire build.

`next` has since been re-synced so today's content no longer triggers it, but the failure mode is structural: it recurs any time a component is removed/renamed on `main` (or added on `next`) while the two branches are out of sync — and it takes down the whole docs deploy.

## Fix

For beta pages only, scan the page's processed markdown for component-like JSX tags (`<PascalCase`) and provide a fallback component for any tag the app doesn't supply. Cross-branch drift now degrades to a visible callout ("This section relies on an interactive component that isn't available in this version of the docs.") plus a build-log warning, instead of a failed deploy.

Stable pages are unchanged: their content and components ship from the same commit, so a missing component there should still fail the build loudly.

## Verification

- Reproduced the exact Vercel failure locally: `main` app code + pre-sync `next` content (with `<GenerateAppleJwt />`) → identical prerender crash
- With this fix, the same build passes; `/docs/beta/authentication/apple` renders the heading with the fallback callout, and `/docs/authentication/apple` (stable) is byte-identical to before
- `tsc --noEmit` passes in `docs/`

### DIFF
--- a/docs/app/docs/[[...slug]]/page.tsx
+++ b/docs/app/docs/[[...slug]]/page.tsx
@@ -23,6 +23,7 @@ import {
 	ForkButton,
 	GenerateSecret,
 } from "@/components/docs/mdx-components";
+import { collectMissingMdxFallbacks } from "@/components/docs/missing-mdx-fallback";
 import { Callout } from "@/components/ui/callout";
 import type { DocsVersion } from "@/lib/docs-versions";
 import {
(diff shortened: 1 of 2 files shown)


gustavovalverde:
This is not required. The fix was done in the `next` branch at the time of sync, and just required a Vercel rebuild

PR #10023, opened 2026-06-12
fix(session): fall back when cache cookie expires

### LINKED ISSUE #10021: Bug: Expired `session_data` cookie cache logs the user out instead of falling back to the database

### Reproduction

1. Set `session.cookieCache: { enabled: true, strategy: "jwe", maxAge: 60 }`.
2. Sign in. The client now has `session_token` (signed pointer to the DB row, lives for `session.expiresIn`) and `session_data` (the JWE cache, `exp` = `maxAge`).
3. Call `GET /api/auth/get-session` with a still-valid `session_token` but a `session_data` whose JWE `exp` has passed. This happens in Production when a frozen or sleeping tab replays the cookie (see Additional context).
4. The endpoint returns `null` and the user is logged out, even though `session_token` is valid and the session in the database hasn't expired.

**Current.** In `src/api/routes/session.ts`, decoding the expired cookie fails (`symmetricDecodeJWT` / `verifyJWT` return `null` because jose throws on `exp`), so `getSession` hits `expireCookie(...); return ctx.json(null)` before it ever reaches the `session_token` DB lookup further down.

**Expected.** An expired cache cookie is a cache miss, not a dead session. It should fall through to the DB lookup, which is exactly what the existing `hasExpired` branch already does when the cookie decodes fine. A 60s cache TTL shouldn't be able to kill a session that's meant to last days (a corrupted or tampered cookie is a different story and should probably still log out).

1.6.17

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "arm64"
  },
  "node": {
    "version": "v24.16.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "11.1.1"
  },
  "betterAuth": {
    "version": "1.6.17"
  }
}
```

Backend

```typescript
  session: {
    cookieCache: { enabled: true, strategy: "jwe", maxAge: 60 },
  },
});
```

### Additional context

We hit this in production. Our users (practitioners) keep a session tab open for hours, often across days, with session re-validation running on a background timer. We started seeing users randomly logged out a few times a day.

Tracking it down took a while because the error is swallowed in practice: the `customSession` plugin (which we also use) wraps `getSession` in `.catch(() => null)`. Once we added diagnostics, every case looked the same: a valid `session_token` and a live DB session, but `getSession` returning `null` once the `session_data` JWE had lapsed.

The trigger is a **suspended / frozen tab**. Normally the browser drops `session_data` when `exp` passes (its `Max-Age` matches), so nothing is sent and the request falls back to the DB. But a frozen tab pauses our re-validation timer while the cookie store persists. On resume it replays a `session_data` cookie sitting just past its `exp`, and the 15s `clockTolerance` is too small to absorb that. Cross-instance clock skew on serverless, where `exp` is stamped by one instance and checked by another, could also produce the same issue.

**Suggested fix:** let an authentic but expired cookie decode (its integrity is already verified before the `exp` check) and fall through to the existing DB lookup instead of returning `null`. This is what we've done on our end, patching the lib locally, and can confirm this resolved the issue.

arnnvv:
Closes #10021.

  ## What changed

  Expired JWT/JWE `session_data` cache cookies were returning `null` before the signed
  `session_token` could be checked against the database.

  This keeps signature/decryption validation in place, but lets the session cache path
  read the payload when jose rejects only the `exp` claim. The existing expiry branch
  then clears the stale cache cookie and falls through to the database lookup.

  ## Tests

  - `pnpm biome check .changeset/fix-expired-cookie-cache-fallback.md packages/better-
  auth/src/api/routes/session-api.test.ts packages/better-auth/src/api/routes/session.ts
  packages/better-auth/src/crypto/jwt.ts --error-on-warnings`
  - `pnpm --filter better-auth typecheck`
  - `pnpm --filter better-auth test src/api/routes/session-api.test.ts src/crypto/secret-
  rotation.test.ts --run`

---
## Summary by cubic
Fixes session retrieval when the `session_data` cookie cache expires. We decode expired JWT/JWE cache cookies with `ignoreExpiration`, clear the stale cookie, and fall back to a DB lookup with the signed `session_token`.

- **Bug Fixes**
  - `verifyJWT` and `symmetricDecodeJWT` accept `{ ignoreExpiration: true }`; return payload only on `JWTExpired` and still verify signature/decryption.
  - Session route uses `{ ignoreExpiration: true }` for both strategies; clears stale cache cookie before DB lookup.
  - Tests cover JWT and JWE paths and assert exactly one DB read.

<sup>Written for commit 7ec640f6c6792d8c147b3398fa04f3ea72c8d7cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10023?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ping-maxwell:
hey I recently merged a fix, closing this.

PR #10045, opened 2026-06-13
feat(oauth-provider): consistent and audience-scoped token introspection

### LINKED ISSUE #8267: oauth2Introspect returns {active: false} when introspecting client differs from token-issuing client

The `/oauth2/introspect` endpoint returns `{active: false}` when the `client_id` used to authenticate the introspection request differs from the `client_id` that issued   
  the token (`azp` for JWTs, `clientId` for opaque tokens). This makes it impossible to implement a standard resource server pattern where a frontend client obtains an
  access token and a separate API/resource server validates it using its own credentials.

  Remove or relax the `azp !== clientId` / `accessToken.clientId !== clientId` checks in `introspect.ts`. RFC 7662 §2.1 only requires the introspecting party to be
  authenticated — it does not require it to be the same client that issued the token. The `azp`/`client_id` field is already returned in the response so callers can inspect
  it themselves.

  Optionally, make strict same-client enforcement opt-in via something like `opts.strictClientIntrospection`.

  - Calling `/oauth2/userinfo` instead — works since it skips the `clientId` check, but requires `openid` scope and is semantically wrong for API token validation.
  - Using the token-issuing client's credentials for introspection — not viable when the resource server doesn't have access to the frontend client's secret.

### Additional context

  The relevant checks in `packages/oauth-provider/src/introspect.ts`:

  ```ts
  // JWT tokens
  if (clientId && jwtPayload.azp !== clientId) {
      return { active: false };
  }

  // Opaque tokens
  if (clientId && accessToken.clientId !== clientId) {
      return { active: false };
  }
  ```

  RFC 7662 reference: https://datatracker.ietf.org/doc/html/rfc7662#section-2.1

gustavovalverde:
Introspecting an opaque access token returned a different, smaller claim set than the equivalent JWT. Each token format built its claims independently, so the two drifted. Opaque introspection dropped per-resource `customClaims` entirely. It also passed a `customAccessTokenClaims` callback's output straight through, without removing reserved RFC 9068 names, so the callback could overwrite server-owned values like `auth_time` or `amr`.

This routes both surfaces through one `resolveAccessTokenClaims` that strips reserved names inside itself, so no call site can skip the step. A JWT and an opaque token now carry the same claims for the same grant. The JWT output stays byte-identical for the same inputs: the precedence is unchanged, with plugin claims, then per-resource claims, then the server's own. Opaque introspection re-derives from stored state on each call. That gives it the same liveness as the JWT path, so a token whose resource was deleted now reads `{ active: false }`.

The second change is who may call introspection. It used to require the client that issued the token. That blocks the common split where a frontend holds the token and a separate API validates it. RFC 7662 §2.1 only asks the caller to authenticate, not to be the issuer. A client may now introspect a token it issued, or a token whose audience resource it is linked to. Unrelated clients still get `{ active: false }`, so introspection gives no probing signal (RFC 7662 §4), and refresh tokens stay issuer-only. Pairwise `sub` is resolved for the issuing client, so a resource server sees the subject the token actually represents.

Closes #8267

---
## Summary by cubic
Unifies access-token claim construction in `@better-auth/oauth-provider` and scopes token introspection to audience. Also renames `ResolvedResourcePolicy.customClaims` to `rawCustomClaims`, fixes opaque-token liveness and a dist typecheck, and clarifies docs on who can introspect and how JWT vs opaque claims behave.

- **New Features**
  - Audience-scoped introspection (RFC 7662): issuer or a resource server linked to any audience resource may introspect; others get `{ active: false }`. Tokens without a resource audience and refresh tokens remain issuer-only. Pairwise `sub` resolves against the issuing client. (Related: #8267)
  - JWT and opaque tokens now return the same enriched claims. Reserved RFC 9068 names are stripped centrally so extensions cannot override server-owned fields.

- **Bug Fixes**
  - Opaque introspection includes per-resource `customClaims` and matches the JWT claim set for the same grant.
  - Opaque tokens whose bound resource was deleted now introspect as inactive; disabling a resource keeps existing tokens valid until expiry.
  - Fix dist typecheck by typing claim-input metadata as `object`.
  - Batch per-resource `customClaims` lookup during introspection to avoid N+1 reads on multi-resource tokens.

<sup>Written for commit 77245a93f56eaf8ac751123f809b75d1c505a172. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10045?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


gustavovalverde:
@cubic-dev-ai review

PR #10058, opened 2026-06-14
fix(drizzle): unwrap mysql DELETE tuple in consumeOne (closes #10054)

### LINKED ISSUE #10054: resetPassword broken since 1.6.17 with drizzle/mysql

### Reproduction

After updating to 1.6.18 the password-reset is not working anymore. I get a token invalid error from the API. Tried 1.6.17 but there i have the same behavior.

What I checked:
- The token is valid (i see the reset-password:<token> in the mysql table)
- Website (docker container) and mariadb (also docker) run in the same time zone, the token is valid, expire time).
- When running the function, i get the INVALID_TOKEN error, but the token has disappeared from the table (assuming consumeVerificationValue did that)

Worked fine in .16, broken since .17

1.6.17

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
  },
  "node": {
    "version": "v26.2.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.16.0"
  },
  "frameworks": [
    {
      "name": "svelte",
      "version": "^5.49.2"
    },
    {
      "name": "@sveltejs/kit",
      "version": "^2.50.2"
    }
  ],
  "databases": [
    {
      "name": "mysql2",
      "version": "^3.14.0"
    },
    {
      "name": "drizzle",
      "version": "^0.45.1"
    }
  ],
  "betterAuth": {
    "version": "1.6.16",
    "config": {
      "baseURL": {
        "allowedHosts": [
          "172.19.0.4"
        ],
        "protocol": "auto",
        "fallback": "http://172.19.0.4:5173"
      },
      "secret": "[REDACTED]",
      "emailAndPassword": {
        "enabled": true,
        "requireEmailVerification": true,
        "password": {}
      },
      "emailVerification": {
        "sendOnSignUp": true,
        "sendOnSignIn": true,
        "autoSignInAfterVerification": true
      },
      "plugins": [
        {
          "name": "sveltekit-cookies",
          "config": {
            "id": "sveltekit-cookies",
            "version": "1.6.18",
            "hooks": {
              "after": [
                {}
              ]
            }
          }
        }
      ],
      "user": {
        "additionalFields": {
          "surname": {
            "type": "string",
            "required": true
          },
          "postalCode": {
            "type": "string",
            "required": true
          },
          "houseNumber": {
            "type": "string",
            "required": true
          }
        }
      }
    }
  }
}
```

Backend

```typescript
import { betterAuth } from 'better-auth/minimal';
  ...
});
```

### Additional context

_No response_

tsushanth:
Closes #10054. Regression introduced by #10003 (fdef997, shipped in v1.6.17) when the MySQL path for `consumeOne` was added.

The new MySQL path does a `SELECT ... FOR UPDATE` followed by a `DELETE`, then reads the affected-row count off the raw driver result to decide whether to return the claimed row. The problem: drizzle-orm/mysql2 returns `MySqlRawQueryResult`, a `[ResultSetHeader, FieldPacket[]]` tuple. Reading `.affectedRows` directly off the tuple yields `undefined`, so the `count = rowsAffected ?? affectedRows ?? changes ?? 0` chain falls through to `0`. `consumeOne` then returns `null` even though the DELETE removed the row.

Downstream, `internalAdapter.consumeVerificationValue` returns null, and `routes/password.ts:298-300` throws `INVALID_TOKEN`. From the reporter's perspective the password-reset token is valid, the row is in the table, and yet calling `resetPassword` errors with INVALID_TOKEN — and after the call, the row is gone. That matches the report exactly.

The codebase already had a `getAffectedRowCount` / `hasDriverRowCount` / `readDriverRowCount` set of helpers for the same shape problem (in `updateMany` / `deleteMany`); the MySQL `consumeOne` path was just written before they were generalized. This PR unwraps the tuple and routes through `hasDriverRowCount` / `readDriverRowCount` so all three driver shapes (mysql2 tuple, drizzle's `{rowsAffected}`, SQLite-style `{changes}`) are handled consistently.

## Test

Added a `describe("consumeOne MySQL affected-row count")` block in `packages/drizzle-adapter/src/drizzle-adapter.test.ts`, mirroring the existing `updateMany` table:

- parametrized over the three driver shapes (`[{ affectedRows: 1 }]` tuple, `{ rowsAffected: 1 }`, `{ changes: 1 }`); each asserts the claimed row is returned, not null
- a fourth case asserts a null `SELECT FOR UPDATE` correctly returns null without touching the delete branch

The tuple case fails against `main` with `expected null not to be null` — the same symptom as the user's password-reset flow. With the fix all four pass; the rest of the suite (16 cases) is untouched.

```
$ pnpm --filter @better-auth/drizzle-adapter exec vitest run -t "consumeOne MySQL"
Tests  4 passed | 16 skipped (20)
```

---
## Summary by cubic
Fixes a MySQL regression in `consumeOne` that returned null after a successful DELETE, causing INVALID_TOKEN during password resets. Unwraps the `mysql2` result tuple and uses shared row-count helpers to correctly return the claimed row. Closes #10054.

- **Bug Fixes**
  - Unwrap `[ResultSetHeader, FieldPacket[]]` from `drizzle-orm/mysql2` and use `hasDriverRowCount`/`readDriverRowCount` to read affected rows across `{ rowsAffected }`, `{ changes }`, and the MySQL tuple.
  - Adds tests for the three result shapes and the null-select path; all pass.

<sup>Written for commit a5f7c05924fb8417dcdef25064fbee54614da9b6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10058?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/content/docs/authentication/google.mdx
+++ b/docs/content/docs/authentication/google.mdx
@@ -161,6 +161,14 @@ socialProviders: {
 
 Set `hd: "*"` to allow any Google Workspace hosted domain. Tokens with no `hd` claim are rejected whenever `hd` is configured.
 
+<Callout type="warn">
+  A custom `getUserInfo` callback replaces Google's built-in callback-path `hd`
+  check. Validate the claim from a trusted provider response and return `null`
+  when it is missing or does not match. Direct ID-token sign-in follows the
+  provider's ID-token verification path, while Google One Tap enforces the
+  configured `hd` separately.
+</Callout>
+
 ### Always ask to select an account
 
 If you want to always ask the user to select an account, you pass the `prompt` parameter to the provider, setting it to `select_account`.
--- a/docs/content/docs/concepts/database.mdx
+++ b/docs/content/docs/concepts/database.mdx
@@ -606,7 +606,8 @@ To add custom fields, use the `additionalFields` property in the `user` or `sess
 * `type`: The data type of the field (e.g., "string", "number", "boolean").
 * `required`: A boolean indicating if the field is mandatory.
 * `defaultValue`: The default value for the field (note: this only applies in the JavaScript layer; in the database, the field will be optional).
-* `input`: This determines whether a value can be provided when creating a new record (default: `true`). If there are additional fields, like `role`, that should not be provided by the user during signup, you can set this to `false`.
+* `input`: Whether Better Auth accepts the field when creating or updating a record (default: `true`). For user fields, this includes values from API input and `mapProfileToUser`. Set this to `false` for server-owned fields such as `role`.
+* `returned`: Whether Better Auth includes the stored field in response bodies (default: `true`). This does not affect whether input can write the field.
 
 Here's an example of how to extend the user schema with additional fields:
 
@@ -646,7 +647,7 @@ const res = await auth.api.signUpEmail({
 });
 
 //user object
-res.user.role; // > "admin"
+res.user.role; // > "user"
 res.user.lang; // > "fr"
 ```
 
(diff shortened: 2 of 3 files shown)


tsushanth:
Rebased onto current main.

tsushanth:
Rebased onto current main.

PR #10059, opened 2026-06-14
fix(device-authorization)!: add lookup indexes

### LINKED ISSUE #10025: Documentation regarding indexes for Device Authorization
The [Device Authorization](https://better-auth.com/docs/plugins/device-authorization) plugin has two non unique columns deviceCode and userCode which are used for searching the database for the device authorization flow but there seems to be no documentation regarding whether indices can be added here safely for optimization. Any guidance would be appreciated.

GautamBytes:
Device Authorization resolves `deviceCode` and `userCode` during token, verification, approval, and denial requests, but generated schemas leave both lookup paths unindexed. This adds portable table-level indexes through the shared database-index contract so schema generation and migration use the same definition.

## What changes

- **Schema contract**: `deviceCode` and `userCode` receive non-unique lookup indexes with portable, bounded names.
- **Migration behavior**: existing schemas receive missing indexes when their columns satisfy the target database's index-key limits.
- **Generated schemas**: SQL, Prisma, and Drizzle output preserve the same non-unique index semantics, and the documentation table exposes the indexed fields.
- **Code bounds**: configured and custom-generated device and user codes are limited to 191 characters, matching the portable MySQL index contract.

> [!IMPORTANT]
> Existing MySQL and SQL Server installations may use unbounded `deviceCode` and `userCode` columns. Before running the migration, convert both columns to strings bounded at 191 characters and resolve longer values. The migration stops with an actionable error instead of creating a non-portable index.

## Review focus

- The device-authorization schema uses the shared table-level index boundary rather than field-local migration logic.
- The migration remains idempotent and does not duplicate unique constraints.
- Documentation generators emit only non-unique explicit indexes; unique fields remain represented by their column constraints.

Closes #10025

### DIFF
--- a/docs/lib/copy-schema/dialects/mysql.ts
+++ b/docs/lib/copy-schema/dialects/mysql.ts
@@ -3,7 +3,12 @@ import type {
 	Resolver,
 	ResolverHandlerContext,
 } from "../types";
-import { filterForeignKeys, getTypeFactory } from "../utils";
+import {
+	filterForeignKeys,
+	filterIndexes,
+	getIndexName,
+	getTypeFactory,
+} from "../utils";
 
 type CustomResolverContext = ResolverHandlerContext & {
 	getType: ReturnType<typeof getTypeFactory>;
@@ -42,6 +47,9 @@ const formatForeignKey = (field: DBFieldAttribute) => {
 	return newLine;
 };
 
+const formatIndex = (field: DBFieldAttribute, tableName: string) =>
+	`CREATE ${field.unique ? "UNIQUE " : ""}INDEX \`${getIndexName(tableName, field)}\` ON \`${tableName}\` (\`${field.fieldName}\`);`;
+
 export const mysqlResolver = {
 	handler: (ctx) => {
 		const getType = getTypeFactory((field) => ({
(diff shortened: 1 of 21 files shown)


GautamBytes:
@ping-maxwell please review it once!

docs/lib/copy-schema/dialects/mysql.ts
```diff
@@ -42,14 +47,31 @@ const formatForeignKey = (field: DBFieldAttribute) => {
...
+	return [
+		`SET @table_exists = (SELECT COUNT(1) FROM information_schema.tables WHERE table_schema = DATABASE() AND table_name = '${escapeSqlString(tableName)}');`,
+		`SET @index_exists = (SELECT COUNT(1) FROM information_schema.statistics WHERE table_schema = DATABASE() AND table_name = '${escapeSqlString(tableName)}' AND index_name = '${escapeSqlString(indexName)}');`,
+		`SET @create_index_sql = IF(@table_exists > 0 AND @index_exists = 0, '${escapeSqlString(createIndex)}', 'SELECT 1');`,
+		"PREPARE create_index_stmt FROM @create_index_sql;",
+		"EXECUTE create_index_stmt;",
+		"DEALLOCATE PREPARE create_index_stmt;",
+	].join("\n");
+};
+
```
ping-maxwell, line 69: we create a `CREATE INDEX` without a prefix length, existing tables store deviceCode/userCode as `TEXT`, which MySQL refuses to index without a key length, so the generated migration SQL may fail on upgraded db

ping-maxwell (changes requested):

ping-maxwell (dismissed):

GautamBytes:

ping-maxwell (commented):
Hey, mind fixing merge conflicts?

gustavovalverde (approved):

PR #10062, opened 2026-06-15
fix(oauth): send Cache-Control: no-store on credential responses

gustavovalverde:
Every OAuth response that returns credentials was missing `Cache-Control: no-store` on the wire, so tokens, client secrets, and device codes could be cached by any proxy, CDN, or browser. The endpoints passed the header to `ctx.json(body, { headers })`, but handlers dispatch with `asResponse: false`, and in that mode `ctx.json`'s second argument is discarded. The headers were built and silently dropped; only `ctx.setHeader` and `ctx.setStatus` reach the response.

The fix routes the header set through a single `setNoStore` primitive in `@better-auth/core/api` instead of inlining `setHeader` at each site. The no-store pair was already hand-copied across endpoints and had drifted: `/device/code` set `Cache-Control` without `Pragma`. One named call makes the correct pair a single definition, so a new credential endpoint reaches for it rather than re-typing two headers.

This covers the token endpoint, dynamic client registration, introspection, and userinfo, plus the device code and device token responses. Registration additionally returns `201 Created` now; that status was passed the same dropped way and never took effect.

A second commit removes seven `ctx.json(x, { status: 200 })` calls in passkey and organization. They passed a dropped argument that already matched the default, so they were no-ops that modeled the same footgun.

---
## Summary by cubic
Fixes missing no-store headers on OAuth/device credential responses and ensures dynamic client registration returns 201 Created. Uses a single `setNoStore` helper so proxies/CDNs/browsers won’t cache tokens, secrets, or PII.

- **Bug Fixes**
  - Send `Cache-Control: no-store` and `Pragma: no-cache` on credential-bearing responses using `setNoStore`.
    - Covers: token (`/oauth2/token`), registration (`/oauth2/register`), introspection (`/oauth2/introspect`), userinfo (`/oauth2/userinfo`), device code (`/device/code`), and device token (`/device/token`).
  - Registration now returns `201 Created` per RFC 7591 (previous status/header args to `ctx.json` were dropped by dispatch).

- **Refactors**
  - Introduced `setNoStore` in `@better-auth/core/api` and re-exported via `better-auth/api`; replaced inlined headers to avoid drift.
  - Removed no-op `ctx.json(..., { status: 200 })` calls in passkey and organization routes.

<sup>Written for commit 4bf11d74fd1daefb0c548a3990c99a3f7b041235. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10062?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/device-authorization-no-store.md
+++ b/.changeset/device-authorization-no-store.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Device authorization responses that carry credentials are now sent with `Cache-Control: no-store` and `Pragma: no-cache`, so intermediaries no longer cache them. This covers the device code response (`/device/code`) and the device token response (`/device/token`).
--- a/.changeset/oauth-provider-no-store.md
+++ b/.changeset/oauth-provider-no-store.md
@@ -1,0 +1,7 @@
+---
+"@better-auth/oauth-provider": patch
+---
+
+OAuth responses that carry credentials are now sent with `Cache-Control: no-store`, so proxies, CDNs, and browsers no longer cache them. This covers the token endpoint (access, refresh, and ID tokens), dynamic client registration (the client secret), token introspection, and userinfo.
+
+Dynamic client registration now returns `201 Created` on success, as required by RFC 7591, instead of `200`.
--- a/packages/better-auth/src/api/index.ts
+++ b/packages/better-auth/src/api/index.ts
@@ -406,6 +406,7 @@ export {
 	createAuthEndpoint,
 	createAuthMiddleware,
 	optionsMiddleware,
+	setNoStore,
 } from "@better-auth/core/api";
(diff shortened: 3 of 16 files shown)


gustavovalverde:
Superseded by #10065, which targets `next` with a declarative `metadata: { noStore: true }` endpoint flag instead of the `main` helper here. next's oauth-provider had diverged (token still dropping the header, registration fixed a third way), so the fix belongs there.

PR #10065, opened 2026-06-15
feat(oauth): enforce no-store on credential responses via a declarative flag

gustavovalverde:
Credential-bearing OAuth and device responses must not be cached, but `next` applied no-store inconsistently. The token and device endpoints passed the header to `ctx.json`'s second argument, which is discarded on the dispatched (`asResponse: false`) path, so it never reached the wire. Introspection and userinfo set nothing. Only registration worked, through a hand-rolled `setHeader` loop.

Rather than patch each site, an endpoint now declares `metadata: { noStore: true }` and `createAuthEndpoint` emits the header set at the boundary. This covers the success response and every thrown error uniformly: the error path reuses the existing `kAPIErrorHeaderSymbol` propagation, so an `unsupported_grant_type` or `invalid_client` response is no-store too. The header set moves to `@better-auth/core` as `NO_STORE_HEADERS`, retiring the oauth-provider-local `OAUTH_NO_STORE_HEADERS`, so device-authorization and any plugin share one definition instead of each plugin keeping its own.

Flagged endpoints: token, introspection, userinfo, dynamic and admin client registration, and device code and token. Schema-validation errors (a malformed body, which carries no credentials) are intentionally left uncovered, since covering them would apply no-store to every oauth endpoint's validation path.

Supersedes #10062, which fixed a subset of this on `main`.

---
## Summary by cubic
Adds a simple `metadata: { noStore: true }` flag to enforce no-store caching on all credential-bearing OAuth and device responses, including client secret rotation. Unifies the header set as `NO_STORE_HEADERS` in `@better-auth/core`.

- New Features
  - `createAuthEndpoint` honors `metadata: { noStore: true }` to add `Cache-Control: no-store` and `Pragma: no-cache` to handler success and error responses.
  - Export `NO_STORE_HEADERS` from `@better-auth/core`; remove provider-local `OAUTH_NO_STORE_HEADERS`.
  - Apply to token, introspection, userinfo, dynamic/admin client registration, client secret rotation, and device code/token endpoints.

- Bug Fixes
  - No-store headers now reach the wire for token and device endpoints (previously dropped by `ctx.json` on the dispatched path).
  - Introspection and userinfo now send no-store.
  - Error responses (e.g., `unsupported_grant_type`, `invalid_client`) also send no-store.
  - Client secret rotation responses are no-store (were cacheable).

<sup>Written for commit 01b7159131aaf07efd10a21b9eafdc0f616587a1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10065?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/oauth-no-store-headers.md
+++ b/.changeset/oauth-no-store-headers.md
@@ -4,6 +4,6 @@
 "better-auth": patch
 ---
 
-OAuth and device-authorization responses that carry credentials now consistently send `Cache-Control: no-store` and `Pragma: no-cache`, so proxies, CDNs, and browsers never cache them. This covers the token, introspection, and userinfo endpoints, dynamic and admin client registration, and the device code and device token responses, including the error responses from those endpoints.
+OAuth and device-authorization responses that carry credentials now consistently send `Cache-Control: no-store` and `Pragma: no-cache`, so proxies, CDNs, and browsers never cache them. This covers the token, introspection, and userinfo endpoints, dynamic and admin client registration, client secret rotation, and the device code and device token responses, including the error responses from those endpoints.
 
 Endpoints declare this with `metadata: { noStore: true }`, and the header set is exported from `@better-auth/core` as `NO_STORE_HEADERS` for responses built by hand.
--- a/.changeset/oauth-no-store-headers.md
+++ b/.changeset/oauth-no-store-headers.md
@@ -1,0 +1,9 @@
+---
+"@better-auth/core": minor
+"@better-auth/oauth-provider": patch
+"better-auth": patch
+---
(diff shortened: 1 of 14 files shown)


gustavovalverde:
@cubic-dev-ai review

PR #10069, opened 2026-06-15
fix(expo): avoid native dynamic imports

### LINKED ISSUE #10028: @better-auth/expo dynamic imports crash Expo SDK 56 native development builds

### Reproduction

1. Create or upgrade an Expo development build to SDK 56 using:
   - `expo@56.0.11`
   - `react-native@0.85.3`
   - `react@19.2.3`
   - `@better-auth/expo@1.6.16`
   - `better-auth@1.6.16`
2. Install the optional Expo peer dependencies used by the client:
   - `expo-network@56.0.5`
   - `expo-web-browser@56.0.5`
3. Configure a native Better Auth client:

   ```ts
   import { expoClient } from "@better-auth/expo/client";
   import Constants from "expo-constants";

     baseURL: "https://example.com",
     plugins: [
       expoClient({
         scheme: Constants.expoConfig?.scheme as string,
         storagePrefix: Constants.expoConfig?.scheme as string,
         storage: SecureStore,
       }),
     ],
   });
   ```

4. Run the app in an iOS development build with Hermes.
5. Importing `@better-auth/expo/client` calls `setupExpoOnlineManager()`. Its `import("expo-network")` creates a lazy Metro module and fails at runtime.

The same Expo SDK 56 failure can also be triggered later by the `await import("expo-web-browser")` path during an OAuth redirect.

Current source locations:

- [`packages/expo/src/online-manager.ts`](https://github.com/better-auth/better-auth/blob/main/packages/expo/src/online-manager.ts)
- [`packages/expo/src/client.ts`](https://github.com/better-auth/better-auth/blob/main/packages/expo/src/client.ts#L465-L480)

Current behavior on iOS development builds:

```text
ERROR [Error: Requiring unknown module "2986". If you are sure the module exists, try restarting Metro. You may also want to run `yarn` or `npm install`.]
```

The numeric module ID changes between bundles. Clearing Metro, rebuilding, and reinstalling dependencies do not resolve it.

Expo is tracking the underlying SDK 56 dynamic-import regression in:

https://github.com/expo/expo/issues/46806

Expected behavior: importing `@better-auth/expo/client` should initialize the online manager without crashing the native app, and OAuth redirects should be able to open `expo-web-browser`.

A local pnpm patch that replaces the runtime dynamic imports with static namespace imports resolves the Better Auth trigger while remaining on SDK 56:

```ts
import * as Network from "expo-network";
import * as WebBrowser from "expo-web-browser";
```

Then:

```ts
const subscription = Network.addNetworkStateListener(...);
const result = await WebBrowser.openAuthSessionAsync(...);
```

Could `@better-auth/expo` use static imports when these peers are installed, expose a native entry point with static imports, or allow the Expo modules to be injected? Any of those would avoid SDK 56's broken lazy-module path without requiring users to downgrade Expo.

`better-auth@1.6.16` and `@better-auth/expo@1.6.16`

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.5.0: Mon Apr 27 20:39:09 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6020",
    "release": "25.5.0",
    "cpuCount": 10,
    "cpuModel": "Apple M2 Pro",
    "totalMemory": "16.00 GB",
    "freeMemory": "0.17 GB"
  },
  "node": {
    "version": "v26.3.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.16.0"
  },
  "frameworks": [
    {
      "name": "react",
      "version": "19.2.3"
    },
    {
      "name": "hono",
      "version": "catalog:"
    }
  ],
  "databases": null,
  "betterAuth": {
    "version": "catalog:",
    "config": null
  }
}
```

This is the unmodified output of the template's requested `npx auth info --json` command. The project uses a pnpm catalog, so the CLI reports `catalog:` for Better Auth. The resolved versions are `better-auth@1.6.16` and `@better-auth/expo@1.6.16`.

- Client
- Package

```typescript
import { expoClient } from "@better-auth/expo/client";
import Constants from "expo-constants";

  baseURL: "https://example.com",
  plugins: [
    expoClient({
      scheme: Constants.expoConfig?.scheme as string,
      storagePrefix: Constants.expoConfig?.scheme as string,
      storage: SecureStore,
    }),
  ],
});
```

### Additional context

- Reproduced locally on iOS with Hermes and an Expo development client.
- Expo SDK 56.0.10 introduced the upstream regression; it remains present in Expo 56.0.11.
- The `expo-network` dynamic import runs immediately when the Better Auth Expo client module initializes, so applications can crash even when they are not using social OAuth.
- `expo-network` and `expo-web-browser` are both installed in the affected application.
- A clean iOS export succeeds after applying the static-import patch.
- Searching Better Auth issues for `expo-network`, `expo-web-browser`, and `Requiring unknown module` did not find an existing report for this Expo SDK 56 failure.

gustavovalverde:
Expo SDK 56 native development builds can fail when Metro evaluates Better Auth's Expo client dynamic imports.

This keeps the existing `@better-auth/expo/client` entry point and uses static Expo module references for the native modules the client already expects. Metro can include those modules in the native bundle instead of relying on lazy module IDs for `expo-network` or `expo-web-browser`.

The package-level optional peer metadata stays unchanged because the root `@better-auth/expo` server plugin does not load these native modules. The docs now list the modules required by the client plugin.

Fixes #10028.

---
## Summary by cubic
Fixes dev build crashes on Expo SDK 56 by replacing dynamic native imports in the `@better-auth/expo` client with static imports so Metro bundles required modules. Restores social sign-in and network tracking. Fixes #10028.

- **Bug Fixes**
  - Use static imports for `expo-web-browser` and `expo-network`; remove async import/require paths in the client and online manager.
  - Update tests to assert synchronous subscription to `Network.addNetworkStateListener`.

- **Migration**
  - Install these client dependencies in your app: `expo-network expo-linking expo-web-browser expo-constants` (docs now clearly list these).

<sup>Written for commit 04b023e0614e660e9cf9521a90a58b0e70ed677b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10069?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/fix-expo-sdk56-native-client.md
+++ b/.changeset/fix-expo-sdk56-native-client.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/expo": patch
+---
+
+Expo native clients no longer crash in SDK 56 development builds when Better Auth starts network tracking or opens a social sign-in browser session.
--- a/docs/content/docs/integrations/expo.mdx
+++ b/docs/content/docs/integrations/expo.mdx
@@ -52,18 +52,13 @@ Expo is a popular framework for building cross-platform apps with React Native. 
     better-auth @better-auth/expo
     ```
 
-    * And you need to install `expo-network` for network state detection.
+    * Install the Expo modules used by the client plugin.
 
     ```package-install
-    expo-network
+    expo-network expo-linking expo-web-browser expo-constants
     ```
 
-    * <small className="text-xs">(Optional)</small> If you're using the default Expo template, these dependencies are already included, so you can skip this step. Otherwise, if you plan to use our social providers (e.g. Google, Apple), your Expo app requires a few additional dependencies.
-
-    ```package-install
-    expo-linking expo-web-browser expo-constants
-
-    ```
+    If you're using the default Expo template, some of these dependencies may already be included.
   </Step>
 
   <Step>
--- a/packages/expo/src/client.ts
+++ b/packages/expo/src/client.ts
@@ -48,7 +48,7 @@ interface ExpoClientOptions {
 	/**
 	 * Options to customize the Expo web browser behavior when opening authentication
 	 * sessions. These are passed directly to `expo-web-browser`'s
-	 * `Browser.openBrowserAsync`.
+	 * `Browser.openAuthSessionAsync`.
 	 *
 	 * For example, on iOS you can use `{ preferEphemeralSession: true }` to prevent
 	 * the authentication session from sharing cookies with the user's default
--- a/packages/expo/src/client.ts
+++ b/packages/expo/src/client.ts
@@ -12,6 +12,7 @@ import {
 } from "better-auth/cookies";
 import Constants from "expo-constants";
 import * as Linking from "expo-linking";
+import * as Browser from "expo-web-browser";
 import { Platform } from "react-native";
 import { setupExpoFocusManager } from "./focus-manager";
 import { setupExpoOnlineManager } from "./online-manager";
@@ -63,7 +64,7 @@ interface ExpoClientOptions {
 	 * });
 	 * ```
 	 */
-	webBrowserOptions?: import("expo-web-browser").AuthSessionOpenOptions;
+	webBrowserOptions?: Browser.AuthSessionOpenOptions;
 }
 
 interface StoredCookie {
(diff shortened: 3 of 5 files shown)


gustavovalverde:
@cubic-dev-ai review

bytaesu:
@gustavovalverde Let me test with an actual Expo app

PR #10070, opened 2026-06-15
fix(core): reuse active transactions

### LINKED ISSUE #9869: magic-link verify deadlocks: runWithTransaction nests adapter.transaction (1.6.11 consumeVerificationValue) instead of reusing the active transaction

### Reproduction

1. better-auth 1.6.11 with the `magicLink` plugin.
2. A database adapter whose `transaction` is a real interactive DB transaction that holds a dedicated connection for its duration (e.g. the Kysely adapter, or `@zenstackhq/better-auth`, backed by `pg`).
3. Limit the client connection pool to one connection: `new Pool({ max: 1 })` -- standard when using Cloudflare Hyperdrive, which pools server-side so the worker keeps a tiny client pool.
4. `await auth.api.signInMagicLink({ body: { email } })` -> works (single insert).
5. `await auth.api.magicLinkVerify({ query: { token } })` (or `GET /api/auth/magic-link/verify?token=...`).

Minimal trigger: any setup where `adapter.transaction` acquires a connection and the pool cannot hand out two connections at once.

**Current:** the verify call hangs forever and never returns. Postgres shows the backend stuck `idle in transaction` (`wait_event = ClientRead`) right after the verification `SELECT`; the connection is never released, so every attempt leaks one idle-in-transaction connection.

Trace of the cause:
- `internalAdapter.consumeVerificationValue` (added in 1.6.11) runs inside `runWithTransaction` -> opens transaction tx1 (`BEGIN`, takes the only pooled connection).
- Consuming the row (`consumeOne`) triggers another `runWithTransaction` -> calls `adapter.transaction` again -> transaction tx2, which tries to acquire a second connection.
- `runWithTransaction` in `@better-auth/core` (`context/transaction.mjs`) always opens a fresh `adapter.transaction`; it does NOT check `adapterAsyncStorage.getStore()` to see whether a transaction is already active in the current async context. So tx2 is nested inside tx1 and blocks forever waiting for the connection tx1 holds -> deadlock.

Note the CRUD path already reuses the active transaction: `getCurrentAdapter(adapter)` returns the `trx` adapter from the ALS store. `runWithTransaction` is inconsistent with this -- it doesn't reuse.

**Expected:** `runWithTransaction` should be reentrant -- when an adapter/transaction is already active in the async-context store, run `fn` within it instead of opening a nested `adapter.transaction` (mirroring `getCurrentAdapter`). Then verify needs only one connection and works on any pool size. (Alternatively/additionally, document that adapters MUST support reentrant nested transactions -- currently adapters like `@zenstackhq/better-auth` open a brand-new transaction on the base client for each `adapter.transaction` call, which is a reasonable contract but incompatible with the current nesting.)

This is security-relevant: the 1.6.11 change that introduced this nesting was the magic-link "mint multiple sessions from one token" race fix, so the fix effectively breaks magic-link verification on connection-limited adapters/pools.

1.6.11

### System info

```bash
> The `auth info` CLI can't introspect this project headlessly because the auth instance is built dynamically via a factory (no static `auth.ts` export), so the values below are assembled manually

{
  "better-auth": { "version": "1.6.11", "@better-auth/core": "1.6.11" },
  "adapter": { "name": "@zenstackhq/better-auth", "version": "3.6.4", "orm": "@zenstackhq/orm 3.6.4 (Kysely 0.28.17)" },
  "database": { "driver": "pg 8.20.0", "pool": { "max": 1 }, "fronted_by": "Cloudflare Hyperdrive" },
  "runtime": { "platform": "Cloudflare Workers (workerd)", "wrangler": "4.95.0", "node_for_tooling": "v22.22.1" },
  "plugins": ["magicLink"],
  "os": "macOS (Darwin 25.5.0)"
}
```

Backend

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

- Bisected away unrelated suspects: the hang reproduces on older framework code and is independent of DI scope changes -- it's purely `1.6.10 -> 1.6.11`.
- Reproduced deterministically in a real `workerd` integration test (request magic link -> capture token -> verify) that times out before the change.
- Relevant code: `@better-auth/core` `context/transaction.mjs` `runWithTransaction` (opens `adapter.transaction` unconditionally) vs `getCurrentAdapter` (reuses the active `trx`); `internalAdapter.consumeVerificationValue` -> `consumeOneWithHooks` -> adapter `consumeOne`.
- Worked around downstream by making the adapter's transaction reentrant (reuse the in-flight transaction via `AsyncLocalStorage`), which is also the correct semantics (nested transaction = one atomic unit) -- but it seems like `runWithTransaction` itself should reuse the active transaction.

gustavovalverde:
Database-backed single-use verification could hang under connection-limited adapters when a consume path entered a transaction and then hit a transaction-based adapter fallback.

This makes Better Auth's transaction context reentrant, so nested transaction scopes join the active transaction adapter instead of asking the database adapter for another transaction. The consume and counter fallbacks now enter that same context, which keeps compatibility for adapters that still rely on the fallback while preserving one atomic unit.

The regression tests cover nested transaction reuse, plain adapter contexts still opening a transaction, deferred hook flushing from nested calls, and the adapter-factory fallback path that previously requested a second transaction.

Fixes #9869

---
## Summary by cubic
Fixes hanging in single-use verification by making the transaction context reentrant and reusing the active transaction in adapter fallbacks. Restores magic-link and token checks on one-connection database pools while keeping operations atomic.

- **Bug Fixes**
  - Nested transaction scopes now join the active adapter instead of opening a new transaction.
  - `consumeOne` and counter fallbacks run within Better Auth’s transaction context using `runWithTransaction` and `getCurrentAdapter`.
  - Deferred hooks from nested calls flush after the outer transaction completes.
  - Tests cover nested reuse, plain adapter contexts, hook flushing, and the factory fallback path.

Patch releases: `better-auth`, `@better-auth/core`.

<sup>Written for commit 4644a58a943fe73a5889475e097d534118db0968. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10070?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/reentrant-transaction-verification.md
+++ b/.changeset/reentrant-transaction-verification.md
@@ -1,0 +1,6 @@
+---
+"better-auth": patch
+"@better-auth/core": patch
+---
+
+Single-use verification flows no longer hang on database adapters that use a one-connection pool. This fixes magic-link verification and similar token checks in connection-limited serverless database setups.
--- a/packages/core/src/context/transaction.test.ts
+++ b/packages/core/src/context/transaction.test.ts
@@ -1,0 +1,84 @@
+import { describe, expect, it } from "vitest";
+import type { DBAdapter, DBTransactionAdapter } from "../db/adapter";
+import {
+	getCurrentAdapter,
+	queueAfterTransactionHook,
+	runWithAdapter,
+	runWithTransaction,
+} from "./transaction";
+
+function createTransactionHarness() {
+	let transactionCalls = 0;
+	const transactionAdapter = {} as DBTransactionAdapter;
+	const adapter = {
+		transaction: async <R>(
+			callback: (trx: DBTransactionAdapter) => Promise<R>,
+		) => {
+			transactionCalls += 1;
+			return callback(transactionAdapter);
+		},
+	} as DBAdapter;
+
+	return {
+		adapter,
+		transactionAdapter,
+		getTransactionCalls: () => transactionCalls,
+	};
+}
+
+describe("runWithTransaction", () => {
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/9869
+	 */
+	it("reuses the active transaction for nested calls", async () => {
+		const { adapter, transactionAdapter, getTransactionCalls } =
+			createTransactionHarness();
+		const adapters: DBTransactionAdapter[] = [];
+
+		await runWithTransaction(adapter, async () => {
+			adapters.push(await getCurrentAdapter(adapter));
+
+			await runWithTransaction(adapter, async () => {
+				adapters.push(await getCurrentAdapter(adapter));
+			});
+		});
+
+		expect(getTransactionCalls()).toBe(1);
+		expect(adapters).toEqual([transactionAdapter, transactionAdapter]);
+	});
+
+	it("still opens a transaction from a plain adapter context", async () => {
+		const { adapter, transactionAdapter, getTransactionCalls } =
+			createTransactionHarness();
+		let activeAdapter: DBTransactionAdapter | null = null;
+
+		await runWithAdapter(adapter, () =>
+			runWithTransaction(adapter, async () => {
+				activeAdapter = await getCurrentAdapter(adapter);
+			}),
+		);
+
+		expect(getTransactionCalls()).toBe(1);
+		expect(activeAdapter).toBe(transactionAdapter);
+	});
+
+	it("runs hooks queued by nested calls after the outer transaction finishes", async () => {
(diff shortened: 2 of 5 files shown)


gustavovalverde:
@cubic-dev-ai review

PR #10084, opened 2026-06-15
chore(deps): bump esbuild override to >=0.28.1

Bekacru:
## Summary

Bumps the `esbuild` override so all transitive resolutions land on a current release (`0.28.1`).

- **Root workspace** (`pnpm-workspace.yaml`) — replaced the stale `esbuild@<=0.24.2: ^0.25.4` override (which still resolved 0.25–0.28.0) with `esbuild@<0.28.1: '>=0.28.1 <0.29'`. The lockfile now resolves a single `esbuild@0.28.1`.
- **`demo/electron`** and **`demo/oidc-client`** — these are standalone pnpm projects with their own `pnpm-workspace.yaml`; added the same override and regenerated their lockfiles.

`esbuild` is a transitive build-time dependency only — no published package's code or declared dependencies change. Lockfiles regenerated with `pnpm install --lockfile-only`; root `pnpm install --frozen-lockfile` passes.

## Test plan

- `pnpm install --frozen-lockfile` passes at root
- Verified every `esbuild@` resolution in all three lockfiles is `0.28.1`

### DIFF
--- a/demo/electron/pnpm-lock.yaml
+++ b/demo/electron/pnpm-lock.yaml
@@ -9,6 +9,7 @@ overrides:
   brace-expansion@<2: 1.1.13
   brace-expansion@>=2 <4: 2.0.3
   brace-expansion@>=4: 5.0.6
+  esbuild@<0.28.1: '>=0.28.1 <0.29'
   fast-uri@<=3.1.1: 3.1.2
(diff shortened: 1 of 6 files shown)


bytaesu:
will fix this in the https://github.com/better-auth/better-auth/pull/10091

PR #10085, opened 2026-06-15
fix(oauth-provider): use updateMany for refresh token CAS

### LINKED ISSUE #10082: oauth-provider: refresh-token rotation uses a non-portable compare-and-swap (fails on Prisma)
Refresh-token rotation in the `refresh_token` grant can fail on adapters where a multi-clause `update` does not return the updated row. On Prisma, the rotation either throws or treats the legitimate winner of a concurrent rotation as the loser, so a valid refresh request is rejected with `invalid_grant`. It passes today only because the default test adapter is in-memory.

### Where

[`packages/oauth-provider/src/token.ts#L609-L625`](https://github.com/better-auth/better-auth/blob/3ef3cfecfa48e5bae6f2153be2339f619920ad2b/packages/oauth-provider/src/token.ts#L609-L625)

```ts
const won = await ctx.context.adapter.update<{ id: string }>({
	model: "oauthRefreshToken",
	where: [
		{ field: "id", value: originalRefresh.id },
		{ field: "revoked", operator: "eq", value: null },
	],
	update: { revoked: new Date(iat * 1000) },
});
if (!won) {
	throw new APIError("BAD_REQUEST", { error: "invalid_grant", ... });
}
```

### Why it breaks

The code reads the truthy return of `adapter.update` as a compare-and-swap result: revoke the row only if it was still unrevoked, otherwise this caller lost the race. That guarantee does not hold for `update`. The adapter contract states `update` may not return the updated data when multiple `where` clauses are provided, and the method is typed `Promise<T | null>`. On Prisma, a compound `where` on the non-unique `revoked` field routes to a unique-input call that rejects the extra clause, so the call throws instead of returning `null`. On any adapter that returns `null` for a multi-clause update even on success, the winner is misclassified as the loser and a valid rotation is rejected.

This is the same compound-where-as-CAS pattern that is non-portable elsewhere: it passes on the in-memory adapter (which returns the first row regardless of clause count) and on Drizzle and Mongo, but not on Prisma.

### Fix

Use a documented race-safe primitive instead of trusting `update`'s return:

- `updateMany({ where: [{ id }, { revoked: null }], update })`, then check the affected count is exactly `1`, or
- `consumeOne` if the row should be claimed once.

### Out of scope

The `FIXME(strict-family-invalidation)` comment just above this block is a separate concern. RFC 9700 §4.14 family invalidation needs a real transaction around the whole mint chain. This issue is only the portability of the single CAS step.

### Reproducing

Exercise refresh-token rotation against the Prisma adapter (`docker compose up -d`) rather than the default in-memory adapter, including two concurrent rotations of the same parent token.

Tushar-Khandelwal-2004:
## Summary

Fixes #10082.

Refresh-token rotation in the OAuth provider was using `adapter.update()` as a compare-and-swap with multiple `where` clauses. That is not portable across adapters: Prisma can reject that compound update shape, and other adapters may return `null` even when the update succeeds.

This PR switches the refresh-token CAS path to `updateMany()` and checks the affected row count instead.

## Changes

- Use `adapter.updateMany()` for refresh-token rotation.
- Use `adapter.updateMany()` for refresh-token revocation as well, since it uses the same CAS pattern.
- Keep the existing “exactly one winner” behavior for concurrent refresh/revoke races.
- Add a regression test that fails if refresh-token rotation relies on compound `adapter.update()`.

---
## Summary by cubic
Fix refresh-token rotation and revocation in `@better-auth/oauth-provider` by using `updateMany` for compare-and-swap. This makes rotation work across adapters (including `Prisma`) while keeping the single-winner behavior under concurrency.

- **Bug Fixes**
  - Use `adapter.updateMany()` for refresh-token rotation and revocation and check the affected row count.
  - Add a regression test to prevent using compound `adapter.update()` for refresh tokens.

<sup>Written for commit 0708dc955ddaa27bb5e0222e6de3c7dfa2b6455c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10085?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/oauth-refresh-token-update-many.md
+++ b/.changeset/oauth-refresh-token-update-many.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/oauth-provider": patch
+---
+
+Use `updateMany` for refresh-token rotation and revocation compare-and-swap checks so adapters that cannot return rows from compound `update` calls, including Prisma, can rotate refresh tokens correctly.
--- a/packages/oauth-provider/src/revoke.ts
+++ b/packages/oauth-provider/src/revoke.ts
@@ -176,7 +176,7 @@ async function revokeRefreshToken(
 	// Atomic compare-and-swap. If a concurrent rotation already revoked
 	// (and re-minted) this row, fail closed and tear down the whole family
 	// so the rotation's offspring cannot be used either.
-	const won = await ctx.context.adapter.update<{ id: string }>({
+	const updated = await ctx.context.adapter.updateMany({
 		model: "oauthRefreshToken",
 		where: [
 			{
@@ -193,7 +193,7 @@ async function revokeRefreshToken(
 			revoked: new Date(iat * 1000),
 		},
 	});
-	if (!won) {
+	if (updated !== 1) {
 		await invalidateRefreshFamily(ctx, clientId, refreshToken.userId);
 		throw new APIError("BAD_REQUEST", {
 			error_description: "refresh token revoked",
--- a/packages/oauth-provider/src/token.test.ts
+++ b/packages/oauth-provider/src/token.test.ts
@@ -11,7 +11,7 @@ import {
 import { jwt } from "better-auth/plugins/jwt";
 import { getTestInstance } from "better-auth/test";
 import { createLocalJWKSet, decodeJwt, jwtVerify } from "jose";
-import { beforeAll, describe, expect, it } from "vitest";
+import { beforeAll, describe, expect, it, vi } from "vitest";
 import { oauthProviderClient } from "./client";
 import { oauthProvider } from "./oauth";
 import type { OAuthOptions, Scope, VerificationValue } from "./types";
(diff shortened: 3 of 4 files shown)


Tushar-Khandelwal-2004:
Closing this in favor of #10086.

That PR fixes the same OAuth provider refresh-token CAS issue, but handles it more broadly by replacing guarded multi-clause `adapter.update` calls with `incrementOne`, which is the portable single-row guarded transition primitive.

It also covers the other affected flows and adds a structural regression test to prevent this pattern from coming back. So #10086 is the more complete fix for #10082.

PR #10099, opened 2026-06-16
feat: support Prisma Next 

### LINKED ISSUE #9636: Support Prisma ORM 8

Prisma is working on a full TS rewrite called prisma next. It is not a minor update, the whole API surface is new, so the current Prisma adapter will not work with it.

There is a new ORM client `db.orm` with composable select, where, include and take calls, plus a separate SQL plan builder `db.sql`.
 I think a separate adapter for prisma next makes more sense than trying to evolve the existing one, given the completly new api surface.
First pre 1.0 versions are already on npm, so it is possible to start trying things out. Repo: https://github.com/prisma/prisma-next
Would the you be open to a prisma next adapter? Happy to help test once something lands.

-

_No response_

### Additional context

_No response_

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/9636


bytaesu (changes requested):
@ping-maxwell Would it be difficult to do this within the existing Prisma adapter package by adding a separate export instead of creating a new package?

bytaesu:
> [!NOTE]
> https://github.com/prisma/prisma-next
> 
> Prisma Next is currently in [Early Access](https://pris.ly/pn-ea) and we're building it in the open with the community. APIs will still evolve as your feedback shapes them, so we don't recommend it for production workloads yet.

**Prisma Next** is still in early access. We plan to include it officially once it's more stable.

Until then, please use the `pkg-pr-new` version.

Shashank-Salian:
> Note
> 
> https://github.com/prisma/prisma-next
> 
> Prisma Next is currently in [Early Access](https://pris.ly/pn-ea) and we're building it in the open with the community. APIs will still evolve as your feedback shapes them, so we don't recommend it for production workloads yet.
> 
> **Prisma Next** is still in early access. We plan to include it officially once it's more stable.
> 
> Until then, please use the `pkg-pr-new` version.

It has now reached RC-2, would you please reconsider this ?

kinglouie:
@ping-maxwell I am following the prisma-next dev very closely and saw that the prisma team also started to implement a better auth extension but never released it:

https://github.com/prisma/prisma-next/pull/968

i believe this was more of a proof of concept for their new schema namespaces. Maybe there is some useful stuff for you in that PR.

bytaesu:
I'll close this PR and use it as a reference while working on an implementation based on the stable version.

PR #10106, opened 2026-06-16
fix: handle async auth init rejections

### LINKED ISSUE #10105: Wrong config setup (for example using default secret) results in a unhandled rejection

### Reproduction

createBetterAuth stores authContext = initFn(options) as an async Promise but never catches it.

```
const createBetterAuth = (options, initFn) => {
    const authContext = initFn(options);   // ← async fn, returns Promise — never .catch()'d
    const { api } = getEndpoints(authContext, options);
    return {
        handler: async (request) => {
            const ctx = await authContext; // ← only awaited here, per-request
            ...
        },
        $context: authContext,            // ← exposed but still no catch
    }
}
```
initFn is the async init() from init.mjs. The returned Promise is stored in authContext and never guarded. It's only ever await'd inside handler, which is a per-request callback. If init rejects at startup (before the first request arrives), Node.js sees a Promise with no .catch() and fires unhandledRejection. 

The first error was:
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:

After adding catch and logging, the cause was logged as:
You are using the default secret. Please set `BETTER_AUTH_SECRET` in your environment variables or pass `secret` in your auth config. 

**Current behavior:**
Unhandled rejection error

**Expected behavior:**
Error caught and re-thrown.

**Workaround:**
export let betterAuthConfig = betterAuth(config)

betterAuthConfig.$context.catch((err) => logger.error('Better-Auth init failed:', err))

Fixed incorrect config after error logged.

1.6.19

### System info

```bash
Node 24.16
Express 5.2.1
```

Backend

```typescript
Remove BETTER_AUTH_SECRET from config.
```

### Additional context

_No response_

Tushar-Khandelwal-2004:
## Summary

Fixes #10105.

This PR prevents Better Auth startup/configuration errors from being reported as unhandled promise rejections before the auth context is actually awaited.

`createBetterAuth` initializes the auth context lazily and exposes the original promise through `$context`. If initialization failed early, for example due to an invalid production config, Node could emit an `unhandledRejection` before any request or `$context` consumer had a chance to await it.

The fix marks the initialization promise as handled immediately with a no-op catch, while preserving the original promise behavior. That means the real initialization error is still thrown when `$context` or `handler` is awaited.

## Changes

- Mark the auth initialization promise as handled inside `createBetterAuth`
- Keep `$context` pointing to the original promise so errors still propagate normally
- Add regression coverage for async initialization rejection handling
- Add a patch changeset

---
## Summary by cubic
Prevents unhandled promise rejections during `better-auth` async startup. Errors still surface when `$context` or `handler` is awaited.

- **Bug Fixes**
  - Mark the `createBetterAuth` init promise as handled with a no-op `catch`.
  - Keep `$context` as the original promise so startup errors propagate on await.
  - Add a regression test to ensure no `unhandledRejection` is emitted.

<sup>Written for commit 50cf0878fdc6a969e506d9134d616d2b0033d87e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10106?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/handled-auth-init-rejection.md
+++ b/.changeset/handled-auth-init-rejection.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Handle Better Auth initialization promise rejections immediately so startup configuration errors do not emit unhandled rejection warnings before `$context` or `handler` is awaited.
--- a/packages/better-auth/src/auth/base.ts
+++ b/packages/better-auth/src/auth/base.ts
@@ -16,6 +16,10 @@ export const createBetterAuth = <Options extends BetterAuthOptions>(
 	initFn: (options: Options) => Promise<AuthContext>,
 ): Auth<Options> => {
 	const authContext = initFn(options);
+	// Auth context initialization is awaited lazily by handlers and $context.
+	// Mark the promise as handled immediately so startup config errors don't
+	// emit unhandled rejections before consumers await the original promise.
+	void authContext.catch(() => undefined);
 	const { api } = getEndpoints(authContext, options);
 	const errorCodes = options.plugins?.reduce((acc, plugin) => {
 		if (plugin.$ERROR_CODES) {
--- a/packages/better-auth/src/auth/full.test.ts
+++ b/packages/better-auth/src/auth/full.test.ts
@@ -4,10 +4,11 @@ import {
 	createAuthMiddleware,
 } from "@better-auth/core/api";
 import type { router } from "better-auth/api";
-import { describe, expect, expectTypeOf, test } from "vitest";
+import { describe, expect, expectTypeOf, test, vi } from "vitest";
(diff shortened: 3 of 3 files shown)


ping-maxwell (approved):

Tushar-Khandelwal-2004:
hey @ping-maxwell, could you authorize the Vercel deployments so we can re-add this to the merge queue?

PR #10109, opened 2026-06-16
feat: custom expiresIn magic-links

### LINKED ISSUE #4284: Allow overriding magic-link expiration

Right now I'm using magic-link to create "user invitations" to add people to teams. I'd like that when I send an invitation (from the backend only) the expiry time for the token be longer (eg 7 days instead of 5min)

would there be a way to be able to change the expiry time for a token just when doing calls from the backend (so that people in the frontend can't try to override that)?

Allow overriding the token expiry time

Reimplement my own invitations system, but this would mean a lot of duplicate code

### Additional context

_No response_

ping-maxwell:
closes https://github.com/better-auth/better-auth/issues/4284

Adds server-side `expiresIn` option to creating magic-links

### DIFF
--- a/.changeset/magic-link-expires-in.md
+++ b/.changeset/magic-link-expires-in.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+feat: allow per-request `expiresIn` override for magic-link tokens on server-side `auth.api` calls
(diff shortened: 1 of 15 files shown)


ping-maxwell:
Recreating as a same-repo PR to bypass fork workflow approval requirements.

Issue #6128, opened 2025-11-20
lastLoginMethod not set when logging in with Passkey


matteobadini:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Login with Passkey
2. Check Last Login Method
3. Result is an empty string

### Current vs. Expected behavior

better-auth.last_used_login_method cookie is not set -> should be "passkey"

### What version of Better Auth are you using?

1.3.34

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.1.0: Mon Oct 20 19:32:41 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6000",
    "release": "25.1.0",
    "cpuCount": 8,
    "cpuModel": "Apple M1 Pro",
    "totalMemory": "16.00 GB",
    "freeMemory": "0.71 GB"
  },
  "node": {
    "version": "v22.20.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.9.3"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.0.3"
    },
    {
      "name": "react",
      "version": "19.2.0"
    },
    {
      "name": "hono",
      "version": "^4.10.6"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.16.0"
    },
    {
      "name": "drizzle",
      "version": "^0.44.7"
    },
    {
      "name": "@neondatabase/serverless",
      "version": "^1.0.2"
    }
  ],
  "betterAuth": {
    "version": "^1.3.34",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Backend, Client

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  emailAndPassword: {  
    enabled: true
  },
  lastLoginMethod()
});
```

### Additional context

Same when `storeInDatabase: true` option is set

I'm willing to create a PR if needed :)

GautamBytes:
working on it!

PR #10112, opened 2026-06-16
fix(last-login-method): record provider for OAuth sign-ins via oAuthProxy

oscar-b:
## Description

When the `oAuthProxy` plugin is enabled, OAuth sign-ins complete at `/oauth-proxy-callback` (where the proxy calls `setSessionCookie`). The `last-login-method` resolver doesn't recognize that path, so the `better-auth.last_used_login_method` cookie — and the optional `lastLoginMethod` DB field — are never written for proxied OAuth logins. Non-proxied methods work because their completion paths are recognized (same class of bug as #6128 for passkey and #6022 for SIWE).

The provider id is only available inside the **encrypted** `profile` payload at the proxy callback, and the resolver runs synchronously, so it can't decrypt it. This PR has the proxy surface the (non-sensitive) provider id as a `provider` query param on the callback redirect, which the resolver then reads.

## Testing

Added a regression test that runs the full proxy flow with `lastLoginMethod()` enabled and asserts the cookie is set to the provider id. It fails on `main` (`expected null to be 'google'`) and passes with this change.

---
## Summary by cubic
Record the last login method for OAuth sign-ins that go through `oAuthProxy`. The proxy now passes a verified `provider` to `/oauth-proxy-callback`, so the cookie and optional DB field are set safely.

- **Bug Fixes**
  - `oAuthProxy`: include a `provider` param on redirects to `/oauth-proxy-callback` and verify it against the decrypted profile before creating the user/session; reject mismatches with `provider_mismatch`.
  - `last-login-method`: recognize `/oauth-proxy-callback`, read `provider`, and persist it; only accept provider-id-shaped tokens.
  - Tests: full proxy flow asserts the cookie is set; tampered `provider` is rejected before any user/session is created (no session, no cookie).

<sup>Written for commit e1717d68410cdc2f4de5e5102b32d811f7efa4be. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10112?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/oauth-proxy-last-login-method.md
+++ b/.changeset/oauth-proxy-last-login-method.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+The last-login-method plugin now records the provider for OAuth sign-ins that complete through the oAuthProxy plugin. The proxy finishes the session at `/oauth-proxy-callback`, a path the plugin did not recognize, so the `better-auth.last_used_login_method` cookie (and the optional database field) were never set for proxied OAuth logins. The proxy now carries the provider id as a `provider` query param on the callback and verifies it against the decrypted, authenticated profile before creating the session, so the value recorded by the plugin cannot be tampered with.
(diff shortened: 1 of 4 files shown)


oscar-b:
@gustavovalverde Hello! Do you need anything from me to move this forward?

bytaesu:
Closing in favor of #10988

PR #10118, opened 2026-06-17
fix(access): aggregate permissions across multiple roles in hasPermission

### LINKED ISSUE #3011: `hasPermission` fails when called with multiple permissions if those permissions are granted by different roles

### To Reproduce

See https://stackblitz.com/edit/stackblitz-starters-wsvsumaj?file=index.js (`node ./index.js` to start)
1. Use Organization plugin with two custom roles, each with a single custom permission (say, role AUser has `a: use-a` and role BUser has `b: use-b`)
2. Assign both roles to one user
3. Call `hasPermission` with each permission separately - both succeed
4. Call `hasPermission` on both permissions together - fails

Expected: if user has the roles that grant permission A and permission B, then `hasPermission` on the union of those permissions (A and B) should also succeed
Actual: `hasPermission` returns `success === false`

1.2.9

```bash
- OS: Windows 11 Build 26100
- WSL2: Linux 6.6.87.1-microsoft-standard-WSL2 
- Debian trixie
```

Backend, Client, Package

```typescript
See Stackblitz repro
```

### Additional context

I think the reason is because the permissions for the roles are checked independently, and no single role has all of the requested permissions.

The way I expected this to work is that the permissions for all the roles are gathered first, and then those permissions are checked against the requested permissions.

adityachaudhary99:
## Problem

`hasPermission` checked each role independently, so a user whose required permissions were split across two roles (role A grants X, role B grants Y) was denied when requesting `{X, Y}`.

## Fix

Merge the statements of all the user's roles into one set before authorizing, so a permission is granted if any role grants it. Applied to both the organization and admin plugins. AND/OR connector semantics and the single-role case are preserved.

## Tests

Added unit coverage in `access.test.ts` (two roles, different permissions) and an integration test in `organization.test.ts` via `checkRolePermission`. Includes a changeset (patch).

Fixes #3011

---
## Summary by cubic
Fix permission checks to aggregate permissions across all roles a user holds. Users with permissions split across roles are now correctly authorized in both admin and organization flows.

- **Bug Fixes**
  - Merge statements from all roles, then authorize once; preserves AND/OR semantics and single-role behavior.
  - Applied in `admin/has-permission.ts` and `organization/permission.ts`.
  - Added unit tests incl. a direct check of exported admin `hasPermission` with comma-separated roles, plus an org integration test.
  - Released as a patch via changeset for `better-auth`.
  - Fixes #3011.

<sup>Written for commit 713fac3fa25e6eb027792d128f01d631c9cacc9f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10118?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/aggregate-roles-has-permission.md
+++ b/.changeset/aggregate-roles-has-permission.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+fix(access): aggregate permissions across multiple roles in hasPermission
--- a/packages/better-auth/src/plugins/access/access.test.ts
+++ b/packages/better-auth/src/plugins/access/access.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, expectTypeOf, it } from "vitest";
 import { createAccessControl, role } from "./access";
 import type { Role } from "./types";
+import { hasPermission } from "../admin/has-permission";
 
 describe("access", () => {
 	const statements = {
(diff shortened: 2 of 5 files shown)


adityachaudhary99:
Good call @greptile-apps — added a direct test of the exported admin `hasPermission` with two roles each granting a different permission (plus deny and single-role cases), so the production merge path is now covered rather than just the local helper. The organization path is exercised by the integration test in organization.test.ts.

PR #10119, opened 2026-06-17
fix(organization): update session context when setting the active organization

### LINKED ISSUE #4708: Organization set-active endpoint is missing session context

### To Reproduce

1. Use the auth config from below.
2. Call `setActive` from the organization plugin

I expect to somewhere find the current userId in the request data. However, it is nowhere to be found. I expect it to be in `ctx.context?.session`, but it is null

1.3.11

### System info

```bash
There is no output, when executing the command.

I'm using better-auth v1.3.11 with Svelte 5.
```

Package

```typescript
  hooks: {
        if (ctx.path === '/organization/set-active') {
            console.log(ctx.context?.session);
        }
    }
  }
});
```

### Additional context

Use Case:
I want to persist the last active organization in the database. Therefore, I need the userId setting the organization active. I wasn't able to find it in the request context.

adityachaudhary99:
## Problem

`setActiveOrganization` refreshed the session cookie (and `newSession`) but left `ctx.context.session` stale, so after-hooks and subsequent middleware still saw the previous `activeOrganizationId`.

## Fix

Re-sync `ctx.context.session` with the updated session after the cookie is set, matching the pattern used in the session/mcp/oidc routes. Guarded for the null adapter return. Response shape unchanged.

## Tests

Added an after-hook test in `organization.test.ts` asserting the session context reflects the new active org after `setActiveOrganization`. Includes a changeset (patch).

Fixes #4708

---
## Summary by cubic
Ensure `setActiveOrganization` keeps the in-request session context in sync so after-hooks and middleware see the correct `activeOrganizationId`. Fixes #4708.

- **Bug Fixes**
  - Re-sync `ctx.context.session` with the updated session after `setSessionCookie` for both set and clear (`organizationId: null`) paths; response shape unchanged.
  - Added after-hook tests to confirm `session` and `newSession` reflect the new active org and the cleared state.

<sup>Written for commit a167c5bd7a9f60276136a734592f0338b97ec2c8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10119?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/fix-organization-set-active-session-context.md
+++ b/.changeset/fix-organization-set-active-session-context.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+fix(organization): update session context when setting the active organization
--- a/packages/better-auth/src/plugins/organization/organization.test.ts
+++ b/packages/better-auth/src/plugins/organization/organization.test.ts
@@ -4183,4 +4183,29 @@ describe("organization set-active updates session context", async () => {
 		expect(observed.session).toBe(organizationId);
 		expect(observed.newSession).toBe(organizationId);
 	});
+
+	it("reflects a cleared active organization (organizationId: null)", async () => {
+		// The clear branch (organizationId === null) received the same fix, so the
+		// after-hook must observe the cleared value rather than the stale id.
+		const { headers } = await signInWithTestUser();
+
+		const org = await auth.api.createOrganization({
+			headers,
+			body: { name: "Clear Probe Org", slug: "clear-probe-org" },
+		});
+		const organizationId = org?.id as string;
+		expect(organizationId).toBeTruthy();
+
+		await auth.api.setActiveOrganization({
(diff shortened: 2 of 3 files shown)


adityachaudhary99:
Added @greptile-apps — a test that sets then clears the active organization (`organizationId: null`) and asserts the after-hook observes the cleared value on `ctx.context.session`, covering the clear branch that received the same fix.

widavies:
Anything I can do to help get this merged?

PR #10120, opened 2026-06-17
fix(oauth): allow sign-in for providers that return no email

### LINKED ISSUE #9124: OAuth sign-in hard-requires email, blocking providers where email is absent (e.g. Discord phone-only accounts)
## Reproduction

1. Configure Better Auth with Discord as a social provider
2. A user with a phone-only Discord account (no email associated) attempts to sign in via Discord OAuth
3. Discord returns the user profile with `email: undefined` (requesting the `email` scope does not guarantee an email exists)
4. Better Auth redirects with `?error=email_not_found`

The user is completely blocked from signing in. The `username` plugin does not help because it only operates on the credential sign-in/sign-up layer and does not hook into the OAuth callback flow.

**Current:** OAuth sign-in fails with `email_not_found` for any provider user that lacks an email address. Three independent runtime guards reject the flow:

| Guard | File | Line |
|-------|------|------|
| `if (!userInfo.email)` | `packages/better-auth/src/api/routes/callback.ts` | 244 |
| `if (!userInfo.email)` | `packages/better-auth/src/api/routes/sign-in.ts` | ~290 (idToken path) |
| `if (!userInfo.email)` | `packages/better-auth/src/api/routes/account.ts` | ~271 (account linking) |

Additionally, `handleOAuthUserInfo` in `packages/better-auth/src/oauth2/link-account.ts` has six `.toLowerCase()` calls on `userInfo.email` (lines 24, 91, 132, 144, 146, 175) that would throw `TypeError` if email ever bypassed the guards.

The `OAuth2UserInfo` type in `packages/core/src/oauth2/oauth-provider.ts` already correctly types email as optional (`email?: (string | null) | undefined`), but the runtime rejects what the type contract allows.

The `DiscordProfile` type (`packages/core/src/social-providers/discord.ts:44`) incorrectly types `email` as `string` instead of `string | undefined`.

**Expected:** OAuth sign-in should succeed for users without an email address. Email should be optional in the user schema. Account linking should fall back to `accountId + providerId` lookup when email is absent (which `findOAuthUser` already does as its primary lookup strategy).

## Affected layers

The fix requires changes across both `packages/core` and `packages/better-auth`:

- **User schema (Zod):** `email: z.string()` in `packages/core/src/db/schema/user.ts:10` needs to become optional/nullable
- **User schema (DB):** `required: true, unique: true` in `packages/core/src/db/get-tables.ts:163-169` needs to allow nullable email (NULLs must not violate uniqueness)
- **Three runtime guards:** change from hard-block to pass-through when email is absent
- **`handleOAuthUserInfo`:** guard all `.toLowerCase()` calls; skip email-based account linking when absent
- **`findOAuthUser`:** make email parameter optional; skip email fallback when absent
- **`createOAuthUser`:** allow creating users without email
- **`DiscordProfile` type:** fix `email: string` to `email?: string`
- **Email verification flow:** skip when email is absent (`link-account.ts:184-199`)

## Areas affected

- Backend
- Types

## Additional context

This affects any OAuth provider where email is not guaranteed: Discord (phone-only accounts), potentially others. The `username` plugin was designed for credential-based auth and does not address this; it does not modify the OAuth callback flow or relax the email requirement in the user schema.

This is a breaking change for the DB schema (existing deployments would need a migration to make the email column nullable).

adityachaudhary99:
## Problem

OAuth sign-in hard-required an email, blocking providers that legitimately omit one, such as phone-only Discord accounts.

## Fix

Added the opt-in provider option `allowSignUpWithoutEmail` (default `false`). When enabled:

- missing and whitespace-only emails are never used as account-match keys
- users receive a deterministic collision-resistant SHA-256 placeholder on `better-auth.invalid`
- existing email and `emailVerified` values are preserved when a provider later omits email
- account lookup falls back to `account.userId` when a joined user is unavailable
- the opt-in propagates through built-in OAuth, Generic OAuth, direct ID-token sign-in, and OAuth proxy passthrough

Default missing-email rejection remains unchanged without the opt-in. `link-social` is intentionally not relaxed.

## Verification

- Generic OAuth: 78 passed
- OAuth account linking: 29 passed
- Internal adapter: 52 passed
- New proxy, direct ID-token, collision, whitespace, and join-fallback regressions passed
- `pnpm typecheck`
- Commit hooks: Biome and spell checks passed

The full OAuth proxy file additionally requires PostgreSQL on `localhost:5432`; its other 28 tests passed locally.

Fixes #9124

### DIFF
--- a/.changeset/oauth-allow-sign-up-without-email.md
+++ b/.changeset/oauth-allow-sign-up-without-email.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+fix(oauth): allow sign-in for providers that return no email via `allowSignUpWithoutEmail`
--- a/packages/better-auth/src/api/routes/callback.ts
+++ b/packages/better-auth/src/api/routes/callback.ts
@@ -239,7 +239,7 @@ export const callbackOAuth = createAuthEndpoint(
 			throw c.redirect(toRedirectTo);
 		}
 
-		if (!userInfo.email) {
+		if (!userInfo.email && !provider.options?.allowSignUpWithoutEmail) {
 			c.context.logger.error(missingEmailLogMessage(provider.id));
 			redirectOnError(c, resolvedErrorURL, "email_not_found");
 		}
--- a/packages/better-auth/src/api/routes/callback.ts
+++ b/packages/better-auth/src/api/routes/callback.ts
@@ -239,7 +239,7 @@ export const callbackOAuth = createAuthEndpoint(
 			throw c.redirect(toRedirectTo);
 		}
 
-		if (!userInfo.email && !provider.options?.allowSignUpWithoutEmail) {
+		if (!userInfo.email?.trim() && !provider.options?.allowSignUpWithoutEmail) {
 			c.context.logger.error(missingEmailLogMessage(provider.id));
 			redirectOnError(c, resolvedErrorURL, "email_not_found");
 		}
@@ -263,6 +263,7 @@ export const callbackOAuth = createAuthEndpoint(
 				disableSignUp:
 					(provider.disableImplicitSignUp && !requestSignUp) ||
 					provider.options?.disableSignUp,
+				allowSignUpWithoutEmail: provider.options?.allowSignUpWithoutEmail,
 				overrideUserInfo: provider.options?.overrideUserInfoOnSignIn,
 			});
 		} catch (e) {
--- a/packages/better-auth/src/api/routes/sign-in.ts
+++ b/packages/better-auth/src/api/routes/sign-in.ts
@@ -288,7 +288,7 @@ export const signInSocial = <O extends BetterAuthOptions>() =>
 						BASE_ERROR_CODES.FAILED_TO_GET_USER_INFO,
 					);
 				}
-				if (!userInfo.user.email) {
+				if (!userInfo.user.email && !provider.options?.allowSignUpWithoutEmail) {
 					c.context.logger.error(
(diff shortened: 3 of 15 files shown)


adityachaudhary99:
Thanks @greptile-apps — both addressed in the latest push:

- **Unique-constraint on `""`**: email-less users are now stored with a per-account synthetic placeholder (`no-email+<provider>-<accountId>@better-auth.invalid`) instead of a literal `""`, so multiple email-less users no longer collide on the unique index. Added a multi-user regression test. The verification-mail send is gated on the real provider email so we don't mail the synthetic address.
- **`emailVerified` downgrade**: on re-sign-in, when the provider returns no email we now preserve the existing `emailVerified` instead of overwriting it with the provider default.

A partial unique index where email is non-null (the v2 TODO in core get-tables.ts) remains the eventual clean fix.

adityachaudhary99:
Addressed the review findings in 8973d64f9: whitespace emails are normalized as missing; overrides preserve stored email/verification; account lookup falls back through account.userId; synthetic addresses use a collision-resistant SHA-256 identity digest; and the opt-in now propagates through built-in, Generic OAuth, direct ID-token, and OAuth proxy paths. Added regressions for each path and updated the PR verification section.

bytaesu:
> [!NOTE]
> https://better-auth.com/docs/concepts/oauth#handling-providers-without-email

Hi @adityachaudhary99,

Better Auth currently requires each user to have a unique email address, enforced also at the database level. This structure is difficult to change in a patch or minor release, so we'll revisit it alongside the broader structural improvements being discussed for v2.

PR #10139, opened 2026-06-18
docs: handle TanStack Start getSession backend failures

### LINKED ISSUE #10132: TanStack Start: auth.api.getSession throwing on backend failure crashes SSR with a RangeError (APIError.status is a non-numeric string)

### Reproduction

The documented TanStack Start `getSession` server function (https://better-auth.com/docs/integrations/tanstack) has **no error handling** around `auth.api.getSession`:

```ts
// straight from the docs
import { createServerFn } from "@tanstack/react-start";
import { getRequestHeaders } from "@tanstack/react-start/server";

export const getSession = createServerFn({ method: "GET" }).handler(async () => {
    const headers = getRequestHeaders();
    const session = await auth.api.getSession({ headers });
    return session;
});
```

`auth.api.getSession` returns `null` for "no session", but it **throws** an `APIError` when the *backend itself* fails (e.g. the database is briefly unreachable — a Postgres restart for a minor-version upgrade, a failover, a connection-pool timeout). The docs' `ensureSession` variant only guards the `null` case (`if (!session) throw ...`); neither sample anticipates a throw. So when `getSession` is called from a route `beforeLoad` during **SSR**, the raw `APIError` propagates into TanStack Start's SSR error responder, which crashes with a misleading `RangeError` that completely masks the real error.

Steps:

1. Set up Better Auth with the documented TanStack Start integration, including the `getSession` server function above, and call it from a route `beforeLoad` (the documented pattern).
2. Make the auth backend fail so `getSession` **throws** instead of returning `null`. Easiest realistic trigger: restart / stop the database while the app is serving, then hard-load (SSR) a page whose `beforeLoad` calls `getSession`. (Isolated trigger without a DB: point the adapter at an unreachable connection string, or temporarily `throw new APIError("INTERNAL_SERVER_ERROR")` from inside the server function — same outcome, proving the crash is the APIError→SSR interaction, not anything DB-specific.)
3. Observe the server crash below instead of a clean 500 / error boundary.

### What actually gets thrown (sanitized server logs)

```
APIError: Failed to get session
     status: "INTERNAL_SERVER_ERROR",   // <-- a STRING, not a number
       body: { message: "Failed to get session", code: "FAILED_TO_GET_SESSION" }
  unhandled: true
      at prepareResponse (.../_ssr/ssr.mjs:518:37)
      at toResponse (.../_ssr/ssr.mjs:487:19)

// ...which the SSR responder turns into:
error: The status provided (-9223372036854775808) must be 101 or in the range of [200, 599]
      at Response (unknown:1:1)
      at errorResponse (.../_ssr/ssr.mjs:612:13)
      at toResponse (.../_ssr/ssr.mjs:538:4)
RangeError: The status provided (-9223372036854775808) must be 101 or in the range of [200, 599]
```

(The triggering query — a `getSession` lookup against the `session` table — fails with `Connection terminated due to connection timeout` / `ECONNREFUSED ...:5432`, i.e. the DB is down. Session token and DB host redacted.)

### Root cause

`better-call`'s `APIError` (v1.3.5) stores the status **name** on `.status` and the numeric code on `.statusCode`:

```js
// better-call/dist/error.mjs
constructor(status = "INTERNAL_SERVER_ERROR", body, headers,
            statusCode = typeof status === "number" ? status : statusCodes[status]) {
    this.status = status;        // => "INTERNAL_SERVER_ERROR"  (string)
    this.statusCode = statusCode; // => 500                      (number)
}
```

TanStack Start's SSR error responder treats a thrown error's `.status` as the **HTTP status code**. So it reads the string `"INTERNAL_SERVER_ERROR"`, coerces it to a number (`Number("INTERNAL_SERVER_ERROR")` → `NaN`), and passes `NaN` to the WHATWG `Response` constructor while building the error response → the `Response` constructor rejects the out-of-range status with a `RangeError`. The real error (DB unreachable) is lost, and on our runtime the request dies mid-render with no usable response.

### Is it Bun-specific?

The **specific magic number** `-9223372036854775808` is Bun/JavaScriptCore specific: Bun's native `Response` binding casts the `NaN` status to `int64`, yielding `INT64_MIN`, then range-checks it. The **underlying bug is NOT runtime-specific** — any `Response` constructor throws a `RangeError` for a `NaN`/out-of-range status. On Node.js (undici) the same `.status`-string → `NaN` → `new Response({ status: NaN })` path throws a `RangeError` too (the message/coerced value differs, e.g. status `0` and "must be in the range of 200 to 599"). The trigger — `APIError.status` being a non-numeric string that a framework reads as an HTTP status — is identical on both runtimes.

**Current:** A transient auth-backend failure during SSR (DB restart, pool timeout) makes the documented `getSession` server function throw a raw `APIError`. Because `APIError.status` is the string `"INTERNAL_SERVER_ERROR"`, TanStack Start's SSR responder coerces it to `NaN`, and the `Response` constructor throws `RangeError: The status provided (-9223372036854775808) must be 101 or in the range of [200, 599]`. The actual cause is masked, the request crashes instead of rendering an error boundary, and (on Bun) the noise/instability extends the outage well past the DB actually coming back.

**Expected:** Following the documented integration, a backend failure should surface as a controlled error — a clean 500 / the router's error component — that recovers as soon as the backend is healthy again, never a cryptic `RangeError` from the `Response` constructor. Concretely, one or more of:

1. **Docs fix (primary ask):** the TanStack Start integration guide should show error handling around `auth.api.getSession`, since it can throw on backend failure (not just return `null`). E.g.:

   ```ts
   export const getSession = createServerFn({ method: "GET" }).handler(async () => {
       const headers = getRequestHeaders();
       try {
           return await auth.api.getSession({ headers });
       } catch (error) {
           // getSession throws (not returns null) when the backend fails — e.g. the DB is briefly unreachable. Re-throw as a plain Error so the framework's SSR responder produces a valid 500 instead of reading APIError.status ("INTERNAL_SERVER_ERROR") as an HTTP status → NaN → Response RangeError.
           throw new Error("Session temporarily unavailable", { cause: error });
       }
   });
   ```

2. **API-shape consideration (secondary):** `APIError.status` holding a non-numeric **string** while `.statusCode` holds the number is a footgun for any framework that (reasonably) treats `error.status` as a numeric HTTP status — TanStack Start is one such consumer. At minimum this is worth documenting; ideally the thrown error wouldn't expose a `.status` that looks like, but isn't, an HTTP status code.

1.6.11  (better-call 1.3.5)

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.5.0: Mon Apr 27 20:40:51 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T8112",
    "release": "25.5.0",
    "cpuCount": 8,
    "cpuModel": "Apple M2",
    "totalMemory": "24.00 GB",
    "freeMemory": "0.12 GB"
  },
  "node": {
    "version": "v24.14.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.9.0"
  },
  "frameworks": [
    {
      "name": "react",
      "version": "^19.2.5"
    }
  ],
  "databases": [
    {
      "name": "pg",
      "version": "^8.20.0"
    },
    {
      "name": "drizzle",
      "version": "^0.45.2"
    }
  ],
  "betterAuth": {
    "version": "Unknown",
    "config": null,
    "error": "Couldn't read your auth config. Make sure to default export your auth instance or to export as a variable named auth."
  }
}
```

Documentation, Backend

```typescript
// Minimal shape — any adapter that can fail/timeout reproduces it.

    database: /* any DB adapter; make it unreachable to trigger the throw */,
    // ...plugins irrelevant to the repro
});
```

### Additional context

- Happens in production during routine DB maintenance (minor-version Postgres upgrade / failover), so it's a recurring availability papercut rather than an edge case.
- It affects **every** session lookup on an SSR path, not just one route — any `beforeLoad`/server function that calls `auth.api.getSession` during SSR.
- Workaround we shipped: a `getSessionSafe()` wrapper that try/catches `auth.api.getSession` and rethrows a plain `Error` (no numeric `.status`), so the SSR responder falls back to a valid 500. Works, but it's exactly the guard the docs sample omits — hence this report.

GautamBytes:
Fixes #10132.

## Summary
- update the TanStack Start integration docs to catch `auth.api.getSession` backend failures
- rethrow a plain `Error` so SSR renders a valid 500/error response

### DIFF
--- a/docs/content/docs/integrations/tanstack.mdx
+++ b/docs/content/docs/integrations/tanstack.mdx
@@ -91,23 +91,31 @@ const signIn = async () => {
 
 To protect resources that require authentication, use `beforeLoad` with a server function. This ensures authentication is checked on every navigation, including client-side navigation via `<Link>` components.
 
-First, create server-side helpers to check the session:
+First, create server-side helpers to check the session. `auth.api.getSession` can throw if the auth backend is unavailable. Catch that error and re-throw a plain `Error` so TanStack Start can render a framework-safe 500 response.
 
 ```ts title="src/lib/auth.functions.ts"
 import { createServerFn } from "@tanstack/react-start";
 import { getRequestHeaders } from "@tanstack/react-start/server";
 import { auth } from "@/lib/auth";
 
-export const getSession = createServerFn({ method: "GET" }).handler(async () => {
+async function getSessionFromRequest() {
     const headers = getRequestHeaders();
-    const session = await auth.api.getSession({ headers });
 
-    return session;
+    try {
+        return await auth.api.getSession({ headers });
+    } catch (error) {
+        // auth.api.getSession can throw when the backend is unavailable.
+        // Re-throw a plain Error so TanStack Start handles it as a 500.
+        throw new Error("Session temporarily unavailable", { cause: error });
+    }
+}
+
+export const getSession = createServerFn({ method: "GET" }).handler(async () => {
+    return getSessionFromRequest();
 });
 
 export const ensureSession = createServerFn({ method: "GET" }).handler(async () => {
-    const headers = getRequestHeaders();
-    const session = await auth.api.getSession({ headers });
+    const session = await getSessionFromRequest();
 
     if (!session) {
         throw new Error("Unauthorized");


bytaesu (changes requested):
No test code is needed to enforce the documentation content.

bytaesu:
Hi @GautamBytes,

https://better-auth.com/docs/concepts/api#error-handling

Since this isn't specific to TanStack and we already have separate guidance for error handling, I'm closing this PR.

Thanks for checking this! 🫡

PR #10148, opened 2026-06-19
feat(totp): add secretEncoding option for RFC 6238 migration compatibility

### LINKED ISSUE #9944: twoFactor: TOTP secret derivation diverges from RFC 6238 convention, blocks migration from compliant libraries

## Summary                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                               
  The `twoFactor` plugin derives the TOTP HMAC key from the stored secret string via `TextEncoder.encode(secret)`. This works correctly for fresh enrollments — base32 round-trips through the QR — but it diverges from the convention used by RFC 6238 reference implementations (Lucia,     
  Auth.js, oslo, otplib, speakeasy, ...), which treat the secret as **raw bytes** and use base32 only as transport for the `otpauth://` URI.                                                                                                                                                   
                                                                                                                                                                                                                                                                                               
  The consequence is that secrets generated by RFC-style libraries cannot be ported to Better Auth: the same `otpauth://...?secret=...` URI yields different HMAC keys on the two sides, so codes never match.
                                                                       
  ## RFC 6238 convention (as implemented by most libraries)        
                                                                                                                                                                                                                                                                                               
  ```                                                                  
  1. K = randomBytes(20)            # raw bytes — the "secret material"                                                                                                                                                                                                                        
  2. Store K (raw, or base32(K))                    
  3. QR URI:   otpauth://totp/...?secret=base32(K)                                                                                                                                                                                                                                             
  4. Verify:   hmac_key = base32_decode(qr_secret) === K                                                                                                                                                                                                                                       
  ```                                                           
                                                                                                                                                                                                                                                                                               
  Authenticator apps (Google Authenticator, Authy, 1Password, Microsoft Authenticator, ...) implement step 4 verbatim.                                                                                                                                                                         
                                                                       
  ## Current Better Auth behavior                                                                                                                                                                                                                                                              
                                                                   
  The twoFactor plugin generates a random string `S`, stores it, and computes:
                                                                
  ```ts                                                                                                                                                                                                                                                                                        
  const hmacKey = new TextEncoder().encode(S); // ASCII bytes of the string
  ```                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                               
  The QR URI exposes `S` via `base32(TextEncoder.encode(S))`, so authenticator apps base32-decode it back to the ASCII bytes of `S` and use those as the HMAC key — matching BA's verify path. End-to-end this works, but the **secret material is `S` as a string**, not raw bytes.           
                                                                                                                                                                                                                                                                                               
  ## Impact                                                            
                                                                                                                                                                                                                                                                                               
  - **Migration is blocked**: any project moving from Lucia, Auth.js, oslo, otplib, speakeasy, etc. to Better Auth cannot port existing TOTP enrollments. The only path is forced re-enrollment of all users — a significant UX disruption for production migrations.
  - **Cross-library interop**: secrets exported from BA also cannot be re-imported into RFC-style libraries (same root cause, reversed direction).                                                                                                                                             
  - **Surprise factor**: the divergence is not documented and is invisible until you try to interoperate.

 ## Proposed resolution                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                               
  Add an opt-in flag on the plugin:                                                                                                                                                                                                                                                            
                                                                   
  ```ts                                                                                                                                                                                                                                                                                        
  twoFactor({                                                                                                                                                                                                                                                                                  
    issuer: "...",                            
    secretEncoding: "raw" | "string", // default "string" — current behavior                                                                                                                                                                                                                   
  })                                                                                                                                                                                                                                                                                           
  ```                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                               
  - `"string"` (default): preserves current behavior, no breaking change.                                                                                                                                                                                                                      
  - `"raw"`: HMAC key derivation does `base32.decode(secret)` before computing the digest, matching RFC 6238 convention. New enrollments would generate `randomBytes(20)` and store `base32.encode(K)`.
                                           
  This is the smallest surface change that unblocks migrations without affecting existing users.
                                                                       
  Alternative shapes considered:                                   
  - **Adapter hook** (`twoFactor({ deriveHmacKey: (secret) => Uint8Array })`) — more flexible but larger API surface.
  - **Auto-detection** based on whether the stored string is valid base32 — fragile, rejected.   

_No response_

### Additional context

## Minimal reproduction                                              
                                                                                                                                                                                                                                                                                               
  ```ts                                                                
  import { base32 } from "@oslojs/encoding";                                                                                                                                                                                                                                                   
  import { randomBytes } from "crypto";                            
                                                                                                                                                                                                                                                                                               
  // Step 1 — simulate Lucia/oslo-style enrollment  
  const K = randomBytes(20);                                                                                                                                                                                                                                                                   
  const qrSecret = base32.encode(K);           // e.g. "JBSWY3DPEHPK3PXP..."
  // User scans otpauth://...?secret=<qrSecret>, app stores K (20 raw bytes).                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                               
  // Step 2 — migrate qrSecret into BA's twoFactor.secret                                                                                                                                                                                                                                      
  await db.update(twoFactor).set({ secret: qrSecret }).where(...);                                                                                                                                                                                                                             
                                                                       
  // Step 3 — user submits 6-digit code from the same authenticator
  await auth.api.verifyTOTP({ body: { code }, headers });       
  // → "Invalid password"                                                                                                                                                                                                                                                                      
  // BA HMACs over TextEncoder.encode(qrSecret) (32 ASCII bytes),  
  // but the authenticator HMACs over base32.decode(qrSecret) (20 raw bytes).                                                                                                                                                                                                                  
  ```

  ## Environment                                                                                                                                                                                                                                                                               
                                                                   
  - `better-auth`: 1.6.11                                                                                                                                                                                                                                                                      
  - Project context: production migration of a SvelteKit + Drizzle (PostgreSQL) app from Lucia 2.7.7 to Better Auth, ~thousands of users with active TOTP enrollments.
                                                                                                                                                                                                                                                                                               
  ## Willingness to contribute                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                               
  Happy to open a PR implementing option 1 if the maintainers agree on the API shape. Let me know your preference on `secretEncoding` vs an alternative name, default value, and whether `"raw"` mode should also change the generation path (`randomBytes(20)` instead of a random string).   
                                                                       
  Thanks for the great library — this is the only blocker we hit on an otherwise smooth migration.

holistis:
## Problem
The `twoFactor` TOTP plugin derives the HMAC key via `TextEncoder.encode(secret)`, which diverges from RFC 6238 convention. Libraries like Lucia, Auth.js, oslo, otplib, speakeasy, and all major authenticator apps (Google Authenticator, Authy, 1Password) use `base32_decode(secret)` as the HMAC key. This makes it impossible to migrate existing TOTP enrollments to Better Auth without forcing users to re-enroll.

Closes #9944

## What does this PR do?
Adds a `secretEncoding` option to `TOTPOptions`:

- `"string"` (default): existing behavior — no breaking change for current users
- `"raw"`: decodes the stored base32 secret to raw bytes before HMAC key derivation, matching RFC 6238 convention

## Migration example
```ts
twoFactor({
  totpOptions: {
    secretEncoding: "raw", // enable for migrated users
  },
})


---
## Summary by cubic
Add a secretEncoding option to the TOTP 2FA plugin to support RFC 6238-compliant secrets and enable migration from libraries like `otplib`, `speakeasy`, `Lucia`, `Auth.js`, and `oslo` without forcing re-enrollment. Default behavior stays the same.

- New Features
  - `TOTPOptions.secretEncoding`: `"string"` (default) keeps current behavior; `"raw"` base32‑decodes the stored secret before HMAC to match authenticator apps (Google Authenticator, Authy, 1Password).
  - Applied in both setup and verification; tests cover migration flow.

- Migration
  - If your stored TOTP secrets are base32 strings from RFC 6238 libraries (e.g., `otplib`, `speakeasy`, `oslo`, `Lucia`, `Auth.js`), set `twoFactor({ totpOptions: { secretEncoding: "raw" } })`. No changes needed otherwise.

<sup>Written for commit 66c6e6863b5e6dbd93a424e37b6e44a398facda7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10148?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>

### DIFF
--- a/.changeset/totp-rfc6238-secret-encoding.md
+++ b/.changeset/totp-rfc6238-secret-encoding.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Add `secretEncoding` option to the TOTP plugin to support RFC 6238-compliant secret handling. When set to `"raw"`, the stored base32 secret is decoded to raw bytes before HMAC key derivation, matching the convention used by Google Authenticator, Authy, Lucia, Auth.js, oslo, otplib, and speakeasy. This unblocks migration of existing TOTP enrollments from RFC-compliant libraries to Better Auth without forced re-enrollment. Default is `"string"` (existing behavior, no breaking change).
--- a/packages/better-auth/src/plugins/two-factor/totp/index.ts
+++ b/packages/better-auth/src/plugins/two-factor/totp/index.ts
@@ -1,6 +1,7 @@
 import { createAuthEndpoint } from "@better-auth/core/api";
 import { APIError, BASE_ERROR_CODES } from "@better-auth/core/error";
 import { createOTP } from "@better-auth/utils/otp";
+import { base32 } from "@better-auth/utils/base32";
 import * as z from "zod";
 import { sessionMiddleware } from "../../../api";
 import { setSessionCookie } from "../../../cookies";
@@ -47,6 +48,23 @@ export type TOTPOptions = {
 	 * Disable totp
 	 */
 	disable?: boolean | undefined;
+	/**
+	 * How the stored secret is interpreted as an HMAC key.
+	 *
+	 * - `"string"` (default): current behavior — the secret string is encoded
+	 *   with TextEncoder before being used as the HMAC key. Works correctly
+	 *   for fresh Better Auth enrollments.
+	 *
+	 * - `"raw"`: RFC 6238 compliant — the stored secret is treated as a
+	 *   base32-encoded string and decoded to raw bytes before use as the HMAC
+	 *   key. This matches the convention used by Google Authenticator, Authy,
+	 *   Lucia, Auth.js, oslo, otplib, speakeasy, and most other libraries.
+	 *   Use this when migrating existing TOTP enrollments from RFC-compliant
+	 *   libraries to Better Auth.
+	 *
+	 * @default "string"
+	 */
+	secretEncoding?: "string" | "raw" | undefined;
 };
 
 const generateTOTPBodySchema = z.object({
@@ -72,6 +90,23 @@ const verifyTOTPBodySchema = z.object({
 		})
 		.optional(),
 });
+
+/**
+ * Resolves a stored TOTP secret to the value passed to createOTP.
+ *
+ * In "raw" mode the stored value is a base32 string (RFC 6238 convention).
+ * We decode it to raw bytes and convert to a latin-1 string so that
+ * TextEncoder.encode() inside createHMAC produces the same byte sequence
+ * that authenticator apps derive from base32_decode(secret).
+ */
+function resolveSecret(
+	secret: string,
+	encoding: "string" | "raw" | undefined,
+): string {
+	if (encoding !== "raw") return secret;
+	const raw = base32.decode(secret);
+	return String.fromCharCode(...raw);
+}
 
 export const totp2fa = (options?: TOTPOptions | undefined) => {
 	const opts = {
@@ -209,7 +244,7 @@ export const totp2fa = (options?: TOTPOptions | undefined) => {
 				}
 				await ctx.context.password.checkPassword(user.id, ctx);
 			}
-			const totpURI = createOTP(secret, {
+			const totpURI = createOTP(resolveSecret(secret, options?.secretEncoding), {
 				digits: opts.digits,
 				period: opts.period,
 			}).url(options?.issuer || ctx.context.appName, user.email);
@@ -285,7 +320,7 @@ export const totp2fa = (options?: TOTPOptions | undefined) => {
 				key: ctx.context.secretConfig,
 				data: twoFactor.secret,
 			});
-			const status = await createOTP(decrypted, {
+			const status = await createOTP(resolveSecret(decrypted, options?.secretEncoding), {
 				period: opts.period,
 				digits: opts.digits,
 			}).verify(ctx.body.code);
--- a/packages/better-auth/src/plugins/two-factor/two-factor.test.ts
+++ b/packages/better-auth/src/plugins/two-factor/two-factor.test.ts
@@ -2536,3 +2536,81 @@ describe("backup codes storage configurations", () => {
 		});
 	}
 });
+
+describe("two factor - secretEncoding: raw (RFC 6238 migration)", async () => {
+	const { testUser, customFetchImpl, sessionSetter, db, auth } =
+		await getTestInstance({
+			secret: DEFAULT_SECRET,
+			plugins: [
+				twoFactor({
+					totpOptions: {
+						secretEncoding: "raw",
+					},
+				}),
+			],
+		});
+
+	const headers = new Headers();
+	const client = createAuthClient({
+		plugins: [twoFactorClient()],
+		fetchOptions: {
+			customFetchImpl,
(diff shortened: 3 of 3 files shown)


holistis:
Thanks for the thorough review — the analysis is correct, and digging in surfaced a deeper issue with this approach that's worth raising before I push a fix.

Root cause: `createOTP(secret: string)` only accepts a string and runs it through `TextEncoder` (UTF-8) internally before using it as the HMAC key. A base32-decoded secret is arbitrary binary, so any byte >= 0x80 expands to a 2-byte UTF-8 sequence — the `String.fromCharCode(...)` (latin-1) round-trip in `resolveSecret` can't survive that. Since roughly half of a random key's bytes are >= 0x80, this breaks essentially every real RFC-library migration, not just an edge case. So it can't be fixed with a small patch to `resolveSecret`.

Two clean ways forward:

1. Add raw-key support to `createOTP` in `@better-auth/utils` (accept `Uint8Array | CryptoKey` for the secret, bypassing the `TextEncoder` path). Keeps the OTP logic in one place and lets `secretEncoding: "raw"` here just pass the decoded bytes through. Since the same org owns `@better-auth/utils`, that feels like the natural home for the fix.
2. Reimplement the raw-key HOTP/TOTP path locally in this plugin via `createHMAC(...).importKey(rawBytes)` (CryptoKey path). Self-contained, but duplicates the OTP loop that already lives in utils.

I'd lean toward (1) as the cleaner design, though it spans two repos. Which fits your roadmap better? Happy to open the `@better-auth/utils` PR and wire this plugin to it, or do the local implementation here.

Either way I'll add a migration test with a high-byte secret (e.g. `0x80, 0x9F, 0xFF`) so this path is actually covered, and wrap the base32 decode so a malformed stored secret returns `INVALID_CODE` instead of a raw 500. Moving this to draft until the direction is settled.

holistis:
Quick update: I went with the first option, adding raw-key support to the better-auth/utils package itself, so the OTP logic stays in one place and anything else built on it benefits too.

PR is here: https://github.com/better-auth/utils/pull/27

It lets createOTP take the secret as raw bytes (or a CryptoKey), so this PR can base32-decode once and pass the bytes straight through instead of running them through TextEncoder (which mangles any byte over 0x7f). Once it's merged and released I'll bump the dependency here, switch this over to it, and drop the local resolveSecret workaround. Happy to adjust the util-side API if you'd prefer a different shape.

holistis:
Hi, just following up here. As discussed, the UTF-8 encoding issue that was flagged would be fixed in better-auth/utils#27, I just bumped that one too. Let me know if you would rather I take a different approach.

PR #10179, opened 2026-06-21
feat(phone-number): add disablePhoneNumberInput option

Bekacru:
## Summary

Adds an opt-in `disablePhoneNumberInput` option to the phone number plugin. When enabled, the user `phoneNumber` field is treated as non-writable input (`input: false`), so a number can only be attached to an account through a flow that verifies ownership of it first, rather than being accepted as plain sign-up or `update-user` input.

By default `phoneNumber` is a writable user field that can be set directly at sign-up before it is verified. This option lets applications require that a phone number is only ever attached after the OTP for it has been verified.

## Behavior

With `disablePhoneNumberInput: true`:

- Passing `phoneNumber` in the sign-up / `update-user` body is rejected with a `400`.
- A number can still be attached through a verified flow:
  - `signUpOnVerification` — account created after the OTP is verified.
  - a session-bound `/phone-number/verify` with `updatePhoneNumber: true`.

The existing `update-user` before-hook (which already blocks phone-number changes) and the internal verified-attach paths are unaffected, since those bypass input parsing.

The option is **default `false`**, so existing behavior is unchanged unless an app opts in.

```ts
phoneNumber({
  sendOTP,
  disablePhoneNumberInput: true,
})
```

## Notes

- The plugin schema is now built fresh per instance instead of mutating the shared module-level `schema` const.
- Note for `schema` option users: `input` can't be set via the plugin `schema` option (it only renames columns), which is why this is a dedicated option.

## Tests

- New tests covering rejection of `phoneNumber` as plain sign-up input and successful attach via the verified flow.
- Full `phone-number.test.ts` suite passes (36).
- `tsc --build` clean.

## Docs

- Documented the new option in `docs/content/docs/plugins/phone-number.mdx`.

### DIFF
--- a/.changeset/phone-number-disable-input.md
+++ b/.changeset/phone-number-disable-input.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Add a `disablePhoneNumberInput` option to the phone number plugin. When enabled, `phoneNumber` is treated as a non-writable user field (`input: false`), so it can only be attached to an account through a verified flow — `signUpOnVerification`, or a session-bound `/phone-number/verify` with `updatePhoneNumber: true` — rather than being accepted as plain sign-up or `update-user` input.
--- a/docs/content/docs/plugins/phone-number.mdx
+++ b/docs/content/docs/plugins/phone-number.mdx
@@ -354,6 +354,24 @@ An object with the following properties:
 
 When enabled, users cannot sign in with their phone number until it has been verified. If an unverified user attempts to sign in, the server will respond with a 401 error (PHONE\_NUMBER\_NOT\_VERIFIED) and automatically trigger an OTP send to start the verification process.
 
+### `disablePhoneNumberInput`
+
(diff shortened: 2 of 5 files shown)


Bekacru:
Addressed the review feedback in d1703e6c0:

- **Schema mutation (index.ts):** every field is now cloned (both branches, plus `phoneNumberVerified`), so `mergeSchema`'s in-place `fieldName` write can no longer leak across plugin instances when a caller passes `schema` field renames. Dropped the inaccurate comment.
- **Docs / JSDoc wording:** reworded to describe the sign-up input behavior only. The previous text implied `update-user` accepts `phoneNumber` by default — it doesn't; the plugin's existing `/update-user` before-hook already rejects non-null phone changes independently of this option.
- **Update-user test suggestion:** intentionally not added. Enabling this option does not change `/update-user` behavior (the before-hook rejects it first, before input parsing), and that path is already covered by the existing "updateUser phone number update prevention" test. The new option's only behavioral effect is at sign-up, which the two new tests cover.

ping-maxwell (approved):
LGTM

PR #10193, opened 2026-06-23
feat(oauth2): extend user field-input rules to OAuth/OIDC provisioning

Paola3stefania:
## Summary
Extends the user model's `input: false` rules to OAuth/OIDC provisioning. Fields marked `input: false` coming from a provider profile (including via `mapProfileToUser`) are now ignored when creating or updating a user on sign-up and account linking, keeping server-owned fields governed by your application schema — consistent with email/password sign-up.

### DIFF
--- a/packages/better-auth/src/db/schema.ts
+++ b/packages/better-auth/src/db/schema.ts
@@ -222,6 +222,23 @@ export function parseUserInput(
 	return parseInputData(user, { fields: schema, action });
 }
 
+/**
+ * Returns a shallow copy of `user` with any field the user model marks as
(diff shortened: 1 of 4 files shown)


packages/better-auth/src/db/schema.ts
bytaesu, line 155: I think it's better to centralize this logic instead of spreading the responsibility across multiple places 🧐 Let me check

Paola3stefania: Yeah I thought of this like but it's a real refactor and touches core behavior...

PR #10198, opened 2026-06-23
fix: honor disableMigration on plugin schema tables

### LINKED ISSUE #8347: `disableMigration` not working on plugins schemas

### To Reproduce

1. Create a custom BetterAuth plugin with `disableMigration: true` in its schema:
     ```ts
     export const testPlugin = () => ({
       id: 'test',
       schema: {
         test: {
           fields: { name: { type: 'string', required: true } },
           disableMigration: true,
         },
       },
     } satisfies BetterAuthPlugin);
  2. Add the plugin to your betterAuth() config: plugins: [testPlugin()]
  3. Run better-auth generate --config auth.ts --output auth.schema.ts
  4. Observe the generated output file

The `test` table is generated in the output schema file despite `disableMigration:
  true` being set on the plugin schema entry. 

**Expected:** when `disableMigration: true` is
  set, the schema generator should skip that model and not emit a table definition for it.


1.5.2

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#1 SMP PREEMPT_DYNAMIC Debian 6.12.73-1 (2026-02-17)",
    "release": "6.12.73+deb13-amd64",
    "cpuCount": 8,
    "cpuModel": "11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz",
    "totalMemory": "19.27 GB",
    "freeMemory": "6.02 GB"
  },
  "node": {
    "version": "v20.10.0",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.9"
  },
  "frameworks": null,
  "databases": [
    {
      "name": "mysql2",
      "version": "^3.15.3"
    },
    {
      "name": "drizzle",
      "version": "^0.45.1"
    }
  ],
  "betterAuth": {
    "version": "~1.5.2",
    "config": null
  }
}
```

Backend

```typescript
import { testPlugin } from "./test-plugin";

  plugins: [testPlugin()],
});
```

### Additional context

The flag disableMigration works correctly at runtime (better-auth does not try to run
  migrations for the table), but the generate CLI command ignores it and still emits the
  table definition in the output file. This means external/custom tables that opt out of
  migration management still pollute the generated schema file.

rachit367:
## Summary

`disableMigration: true` on a plugin schema entry had **no effect**. A table flagged with it was still emitted by `better-auth generate` (issue #8347) — and, despite the issue assuming runtime worked, also still created by the runtime migrator.

Root cause: `getAuthTables` (`packages/core/src/db/get-tables.ts`) drops the flag while assembling plugin tables — it only copies `fields` and `modelName`. Since the flag never made it onto the table object, nothing downstream could read it. Confirmed it is never read anywhere: the only reference in source was its type declaration in `core/src/db/plugin.ts`.

Closes #8347.

## Fix

Propagate the flag and honor it everywhere a table is materialized:

- `get-tables.ts` — carry `disableMigration` from the plugin schema onto the table as `disableMigrations` (the name already declared on `BetterAuthDBSchema`).
- `get-schema.ts` — carry it onto the schema entries that the migrator consumes.
- `get-migration.ts` — skip flagged tables when building the create/alter plan (fixes runtime migration and the Kysely SQL generator, which both go through `getMigrations`).
- `generators/drizzle.ts` and `generators/prisma.ts` — skip flagged tables when emitting the schema file.

## Tests

Added to `packages/cli/test/generate.test.ts`: a plugin exposing one normal table and one with `disableMigration: true`, asserting the normal table is emitted and the flagged one is omitted, for both Drizzle and Prisma. Both pass. `pnpm typecheck` is clean.

Note: 7 pre-existing Prisma snapshot tests fail in my environment with whitespace-only diffs (CRLF) — they fail identically on a clean checkout with my changes stashed, so they're unrelated to this change. The Drizzle snapshots and all non-snapshot tests pass.

---
## Summary by cubic
Honor `disableMigration` on plugin schema tables so flagged tables are skipped by `better-auth generate` (Drizzle/Prisma) and by the runtime migrator, including during schema merges. Closes #8347.

- **Bug Fixes**
  - Carry and preserve the flag through `getAuthTables` and `getSchema` (including when plugins merge on the same table key or `modelName`).
  - Skip flagged tables when building migrations in `getMigrations`.
  - Skip flagged tables in Drizzle and Prisma generators; tests ensure only non-flagged tables are emitted.

<sup>Written for commit 9e68277d791ba0858fa1e7cd62ac4d2a76a5eb20. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10198?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/skip-disabled-migration-tables.md
+++ b/.changeset/skip-disabled-migration-tables.md
@@ -1,0 +1,7 @@
+---
+"better-auth": patch
+"@better-auth/core": patch
+"auth": patch
+---
+
+Honor `disableMigration` on plugin schema tables. Tables flagged with `disableMigration: true` are now skipped by `better-auth generate` (Drizzle and Prisma output) and by the runtime migrator, instead of being emitted and created anyway. The flag was previously dropped while assembling the table list, so it had no effect.
--- a/packages/better-auth/src/db/get-migration.ts
+++ b/packages/better-auth/src/db/get-migration.ts
@@ -224,6 +224,9 @@ export async function getMigrations(config: BetterAuthOptions) {
 	}[] = [];
 
 	for (const [key, value] of Object.entries(betterAuthSchema)) {
+		if (value.disableMigrations) {
+			continue;
+		}
 		const table = tableMetadata.find((t) => t.name === key);
 		if (!table) {
 			const tIndex = toBeCreated.findIndex((t) => t.table === key);
--- a/packages/better-auth/src/db/get-schema.ts
+++ b/packages/better-auth/src/db/get-schema.ts
@@ -34,6 +34,9 @@ export function getSchema(config: BetterAuthOptions) {
 				...schema[table.modelName]!.fields,
 				...actualFields,
 			};
+			if (table.disableMigrations) {
+				schema[table.modelName]!.disableMigrations = true;
+			}
 			continue;
 		}
 		schema[table.modelName] = {
--- a/packages/better-auth/src/db/get-schema.ts
+++ b/packages/better-auth/src/db/get-schema.ts
@@ -9,6 +9,7 @@ export function getSchema(config: BetterAuthOptions) {
 		{
 			fields: Record<string, DBFieldAttribute>;
 			order: number;
+			disableMigrations?: boolean | undefined;
 		}
 	> = {};
 	for (const key in tables) {
@@ -38,6 +39,7 @@ export function getSchema(config: BetterAuthOptions) {
 		schema[table.modelName] = {
 			fields: actualFields,
 			order: table.order || Infinity,
+			disableMigrations: table.disableMigrations,
 		};
 	}
 	return schema;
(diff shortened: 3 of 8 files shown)


rachit367:
Good catch @greptile-apps — fixed in 9e68277. The merge path in `getSchema` (shared `modelName` across plugins) now carries `disableMigrations` through instead of dropping it, so a flagged table can't lose the flag by colliding with an unflagged one. Used an OR semantic (if either entry opts out, the merged table stays opted out). Added accumulation coverage to `get-tables.test.ts` as well.

ping-maxwell (approved):

PR #10203, opened 2026-06-24
fix: harden forwarded client IP resolution

bytaesu:
> [!NOTE]
> Any security-related use of X-Forwarded-For (such as for rate limiting or IP-based access control) must only use IP addresses added by a trusted proxy. Using untrustworthy values can result in rate-limiter avoidance, access-control bypass, memory exhaustion, or other negative security or availability consequences.
> 
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For

Systems that have already configured ipAddressHeaders with trusted values are not affected.

Systems behind multiple proxy hops that have not configured ipAddressHeaders will fall back to the default X-Forwarded-For handling and may break as a result of this hardening.

However, that is the intended behavior of this PR.

### DIFF
--- a/docs/content/docs/concepts/rate-limit.mdx
+++ b/docs/content/docs/concepts/rate-limit.mdx
@@ -74,6 +74,41 @@ export const auth = betterAuth({
 })
 ```
 
+By default Better Auth does not trust comma-separated forwarded IP chains, since
+the leftmost `X-Forwarded-For` token is client-controlled behind an appending
+proxy. Behind a reverse proxy or load balancer, do one of the following.
+
+Point `ipAddressHeaders` at a single trusted header your proxy sets (for example
+`x-real-ip` or `cf-connecting-ip`):
+
+```ts title="auth.ts"
+export const auth = betterAuth({
+    advanced: {
+        ipAddress: {
+            ipAddressHeaders: ["x-real-ip"],
+        },
+    },
+})
+```
+
+Or list your proxies in `trustedProxies`. Better Auth walks the chain right to
+left, skips the trusted hops, and takes the first untrusted address as the client:
+
+```ts title="auth.ts"
+export const auth = betterAuth({
+    advanced: {
+        ipAddress: {
+            // your proxies' addresses, not a broad private range that also covers clients
+            trustedProxies: ["192.0.2.10", "10.0.0.0/24"],
+        },
+    },
+})
+```
+
+For the forwarded chain to be reliable, requests should reach Better Auth only
+through these proxies, so keep your origin reachable through the proxy rather than
+directly.
+
 #### IPv6 Address Support
 
 Better Auth automatically normalizes IPv6 addresses to prevent bypass attacks where attackers use different representations of the same IPv6 address (e.g., `2001:db8::1` vs `2001:0db8:0000:0000:0000:0000:0000:0001`). This ensures that all representations of the same IPv6 address are treated as the same for rate limiting purposes.
--- a/docs/content/docs/concepts/rate-limit.mdx
+++ b/docs/content/docs/concepts/rate-limit.mdx
@@ -105,9 +105,12 @@ export const auth = betterAuth({
 })
 ```
 
-For the forwarded chain to be reliable, requests should reach Better Auth only
-through these proxies, so keep your origin reachable through the proxy rather than
-directly.
+<Callout type="warning">
+  `trustedProxies` only interprets the forwarded header chain and cannot verify the
+  direct sender. Keep your origin reachable only through these proxies, and make each
+  proxy overwrite or sanitize the forwarded IP headers. Otherwise a client can set
+  `X-Forwarded-For` directly and spoof its address.
+</Callout>
 
 #### IPv6 Address Support
 
--- a/packages/api-key/src/utils.ts
+++ b/packages/api-key/src/utils.ts
@@ -19,6 +19,8 @@ import { getIPFromHeader } from "@better-auth/core/utils/ip";
 
 const LOCALHOST_IP = "127.0.0.1";
 
+// TODO: this duplicates `better-auth` request-IP resolver. Centralize it
+// behind a shared export so the security-sensitive logic lives in one place.
 export function getIp(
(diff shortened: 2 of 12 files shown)


bytaesu:
@greptileai review

packages/better-auth/src/utils/get-request-ip.ts
bytaesu, line 18: > [!IMPORTANT]
> Systems that have already configured `ipAddressHeaders` with trusted values are not affected.
> 
> Systems behind multiple proxy hops that have not configured `ipAddressHeaders` will fall back to the default `X-Forwarded-For` handling and may break as a result of this hardening.
> 
> However, that is the intended behavior of this PR.

ping-maxwell (dismissed):
should this instead be minor because of the changes to x-forwarded-for? or do you think it's fine as it is?

packages/core/src/utils/ip.ts
```diff
@@ -195,6 +195,146 @@ export function normalizeIP(
...
+ * Trusted-proxy entries that are not a valid IP address or CIDR range.
+ */
+export function findInvalidTrustedProxies(entries: string[]): string[] {
+	return entries.filter((entry) => parseCIDR(entry) === null);
+}
+
+/**
+ * Resolves the client IP from a forwarded header. The leftmost token is spoofable,
+ * so with `trustedProxies` the chain is stripped from the right to the first
+ * untrusted hop. Otherwise only a single-value header is trusted. `null` = shared bucket.
```
gustavovalverde: This helper is now used by more than the rate limiter (`session.ipAddress` and captcha also go through `getIp`), so `null = shared bucket` leaks one caller's behavior into a cross-cutting utility. Please make this comment domain-neutral, e.g. `null = no trustworthy client IP could be resolved`, and keep the shared-bucket explanation in the rate-limiter code/tests.

packages/core/src/types/init-options.ts
```diff
@@ -240,6 +240,16 @@ export type BetterAuthAdvancedOptions = {
...
+				/**
+				 * Trusted reverse-proxy IPs or CIDR ranges. When set, a forwarded IP
+				 * chain is walked right to left, trusted hops are skipped, and the
+				 * first untrusted address is the client IP. Unset trusts only
+				 * single-value IP headers. Use the actual address or subnet of your
+				 * proxies, not a broad private range that also covers clients.
+				 *
+				 * @example ["192.0.2.10", "10.0.0.0/24"]
+				 */
+				trustedProxies?: string[];
```
gustavovalverde, line 257: Since this is a public option, the JSDoc should carry the same security boundary as the docs: this only interprets a forwarded header chain, it does not prove the direct sender was one of these proxies. Please add one sentence that this is only safe when the app origin is reachable only through those proxies and clients cannot send forwarded headers directly. We can revisit the exact option name in `next`, but the released contract should be explicit now.

docs/content/docs/concepts/rate-limit.mdx
```diff
@@ -74,6 +74,41 @@ export const auth = betterAuth({
...
+    advanced: {
+        ipAddress: {
+            // your proxies' addresses, not a broad private range that also covers clients
+            trustedProxies: ["192.0.2.10", "10.0.0.0/24"],
+        },
+    },
+})
+```
+
+For the forwarded chain to be reliable, requests should reach Better Auth only
```
gustavovalverde: This is the key safety condition for the new option, so I would make it harder to miss. Consider a warning/callout stating that the origin must not be directly reachable and that the proxy must overwrite/sanitize forwarded IP headers. Otherwise `trustedProxies` can look like it fully authenticates the chain, but Better Auth only sees `Request`/`Headers`, not the socket peer.

packages/better-auth/src/api/rate-limiter/rate-limiter.test.ts
```diff
@@ -460,6 +460,119 @@ describe("missing client IP warning", () => {
...
+			});
+
+			if (i >= 3) {
+				expect(response.error?.status).toBe(429);
+			} else {
+				expect(response.error).toBeNull();
+			}
+		}
+
+		expect(store.has("no-trusted-ip|/sign-in/email")).toBe(true);
```
gustavovalverde, line 511: This is the scenario where users will need the new remediation path. The existing runtime warning only points to `ipAddressHeaders`, but with this PR an unresolved multi-hop `X-Forwarded-For` chain is also fixed by configuring `advanced.ipAddress.trustedProxies`. Please update the warning and test expectation so users hit by this hardening can discover the option added here.

packages/api-key/src/utils.ts
```diff
@@ -38,11 +37,12 @@ export function getIp(
 	for (const key of ipHeaders) {
 		const value = "get" in headers ? headers.get(key) : headers[key];
 		if (typeof value === "string") {
-			const ip = value.split(",")[0]!.trim();
-			if (isValidIP(ip)) {
-				return normalizeIP(ip, {
-					ipv6Subnet: options.advanced?.ipAddress?.ipv6Subnet,
-				});
+			const ip = getIPFromHeader(value, {
```
gustavovalverde, line 42: This duplicates the security-sensitive request-IP resolver from `better-auth`. For the patch, I am fine not refactoring the package boundary if we want minimal risk, but please either centralize this through the existing `better-auth/api` export or add a #TODO comment here for `next`

gustavovalverde (dismissed):
I agree with treating this as a patch release because the behavior change is a security hardening, but a few contract edges should be tightened here so the patch does not teach users or future agents the wrong model.

The main things I would address before merge are: keep utility comments domain-neutral, make the exported option docs explicit about the proxy-only origin requirement, point the runtime warning/tests at the new `trustedProxies` remediation, and leave a concrete issue-linked reminder for deduplicating the API-key copy in `next` if we do not centralize it in this patch.

bytaesu:
> should this instead be minor because of the changes to x-forwarded-for? or do you think it's fine as it is?

Yeah, I think `main` is fine 🤔

Issue #7787, opened 2026-02-04
[generic-oauth] Microsoft entra id plugin not handling personal microsoft account name


EjsenKirie:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Create a multi-tenant app registration with audience set to `Accounts in any Microsoft Entra directory and personal Microsoft accounts` enabled. Add the app registration to better-auth as a provider utilizing the `microsoftEntraId` provider helper. Proceed to login with a personal microsoft account.

### Current vs. Expected behavior

Currently, the `getUserInfo` helper for entra id are unable to map the user `name`, resulting in a `name_is_missing` error.

### What version of Better Auth are you using?

1.4.18

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025",
    "release": "6.6.87.2-microsoft-standard-WSL2",
    "cpuCount": 8,
    "cpuModel": "Intel(R) Core(TM) Ultra 7 268V",
    "totalMemory": "15.40 GB",
    "freeMemory": "5.18 GB"
  },
  "node": {
    "version": "v20.20.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "10.8.2"
  },
  "frameworks": [
    {
      "name": "svelte",
      "version": "^5.45.2"
    },
    {
      "name": "@sveltejs/kit",
      "version": "^2.49.5"
    }
  ],
  "databases": [
    {
      "name": "@prisma/client",
      "version": "^7.3.0"
    }
  ],
  "betterAuth": {
    "version": "^1.4.18",
    "config": null
  }
}
```

### Which area(s) are affected? (Select all that apply)

Package

### Auth config (if applicable)

```typescript

```

### Additional context

The issue seems to be that the `betterFetch<MicrosoftEntraIdProfile>` response differs between personal and work accounts:

https://github.com/better-auth/better-auth/blob/9f40c35ee428b0313c90089d857b219b911a51ca/packages/better-auth/src/plugins/generic-oauth/providers/microsoft-entra-id.ts#L59-L66

Personal accounts are returning `profile.givenname` and `profile.familyname` while work accounts returns `profile.given_name` and `profile.family_name`, causing the mapping to fail for personal accounts:
[...]

PR #10204, opened 2026-06-24
fix(microsoft)!: use oid as account id

### LINKED ISSUE #10194: Microsoft provider uses pairwise `sub` as account identifier instead of stable `oid`
### Describe the bug

The Microsoft Entra ID provider uses `user.sub` (from the decoded ID token) as the `id` returned from `getUserInfo`, which becomes `accounts.account_id` in the database.

https://github.com/better-auth/better-auth/blob/main/packages/core/src/social-providers/microsoft-entra-id.ts — in `getUserInfo`:

```js
return {
    user: {
        id: user.sub,  // ← pairwise, app-specific
        ...
    },
    data: user
};
```

Microsoft's `sub` claim is a **pairwise identifier** — it is unique per user **per application registration**. If you change or recreate your Azure AD app registration, all existing Microsoft users' `account_id` values will no longer match, effectively orphaning their accounts.

Microsoft's own documentation explicitly recommends using `oid` instead:

> **sub** — The principal about which the token asserts information. This value is immutable and cannot be reassigned or reused. **However, the subject is a pairwise identifier — it is unique to a particular application ID.**
>
> **oid** — The immutable identifier for the object in the Microsoft identity platform. **Use this claim instead of sub for uniquely identifying users** — even across applications.

Source: [Microsoft identity platform ID token claims reference](https://learn.microsoft.com/en-us/entra/identity-platform/id-token-claims-reference)

**Current:** `accounts.account_id` stores `sub` (pairwise, app-specific).

**Expected:** `accounts.account_id` should store `oid` (stable across app registrations within the same tenant). For multi-tenant scenarios, `{tid}:{oid}` would be the fully qualified stable identifier.

### Impact

- Changing or recreating an Azure AD app registration orphans all existing Microsoft user accounts
- Cross-app user correlation is impossible (relevant for organizations running multiple Better Auth instances)
- Contradicts Microsoft's explicit guidance for user identification

### Suggested fix

In `getUserInfo`, use `oid` instead of `sub`:

```js
return {
    user: {
        id: user.oid,  // stable across app registrations
        ...
    },
    data: user
};
```

Or make it configurable with `oid` as the default. A breaking change migration path would be needed for existing users (update `accounts.account_id` from `sub` to `oid` using the ID token claims).

1.6.16

Social Providers (Microsoft Entra ID)

GautamBytes:
Microsoft Entra `sub` values are pairwise to an application registration, so using them as account identifiers can orphan existing links when an app registration changes. This switches both Microsoft provider paths to the stable, tenant-local `oid` claim while retaining the provider issuer as the tenant boundary.

## What changes

- The built-in `microsoft` provider and Generic OAuth `microsoftEntraId` helper now use `oid` as the provider account identifier.
- Tokens without a valid `oid` are rejected. `mapProfileToUser` continues to map local user attributes but cannot redefine provider identity.
- The Generic OAuth helper refuses initialization unless discovery provides the issuer and JWKS metadata required to verify ID-token claims.
- Microsoft Graph profile data is accepted only when its `sub` matches the ID token. Profile photo requests are skipped when disabled or when no access token is available.
- Personal-account name variants remain normalized, and `email_verified` continues to default to `false`.

> [!IMPORTANT]
> Existing Microsoft account rows created with `sub` must be migrated to `oid` before this change receives production traffic. If stored ID tokens are unavailable, pause Microsoft sign-in and account linking during the cutover or migrate from an external Entra export to prevent duplicate accounts.

Closes #10194


packages/core/src/social-providers/microsoft-entra-id.ts
```diff
@@ -277,34 +277,70 @@ export const microsoft = (options: MicrosoftOptions) => {
...
-							user.picture = `data:image/jpeg;base64, ${pictureBase64}`;
-						} catch (e) {
-							logger.error(
-								e && typeof e === "object" && "name" in e
-									? (e.name as string)
-									: "",
-								e,
-							);
-						}
+			if (oidAccountId) {
```
gustavovalverde: This still performs the Graph photo request whenever an `oid` exists, even when `disableProfilePhoto` is true or when `token.accessToken` is absent, such as an ID-token sign-in path. That creates avoidable network I/O and can send `Authorization: Bearer undefined`. Guard the fetch itself with `!options.disableProfilePhoto && token.accessToken`, then keep the later `mapProfileToUser` invocation independent of the photo fetch so custom account-id mapping still runs.

packages/better-auth/src/plugins/generic-oauth/providers/microsoft-entra-id.ts
```diff
@@ -66,23 +105,34 @@ export function microsoftEntraId(
 		);
 
 		if (error || !profile) {
-			return null;
+			return tokenUserInfo;
 		}
 
-		return {
-			id: profile.sub,
+		const profileWithClaims = {
```
gustavovalverde, line 140: Before merging Graph UserInfo fields into the ID-token profile, this should check that `profile.sub` matches `tokenProfile.sub` when both are present. OIDC Core requires the UserInfo `sub` to exactly match the ID Token subject; otherwise the UserInfo response must not be used. Without that guard, a bad or misrouted Graph response can supply name/email/image fields for a different subject while the stored account id still comes from the token `oid`.

packages/better-auth/src/plugins/generic-oauth/providers/microsoft-entra-id.ts
```diff
@@ -55,7 +62,39 @@ export function microsoftEntraId(
...
+		const accountId =
+			typeof tokenProfile.oid === "string" && tokenProfile.oid.length > 0
+				? tokenProfile.oid
+				: undefined;
+		const tokenUserInfo = {
+			...tokenProfile,
+			...(accountId !== undefined ? { id: accountId } : {}),
+			name:
+				tokenProfile.name ??
+				(`${tokenProfile.given_name ?? ""} ${
```
gustavovalverde: This Microsoft rewrite should absorb the personal-account name normalization from #7787/#7840. The helper currently only falls back through `given_name`/`family_name`, but the reported personal Microsoft account shape used `givenname`/`familyname` without underscores. If we are defining the future Microsoft contract here, add those fields and cover the fallback in the new helper tests so personal accounts do not keep hitting `name_is_missing`.

docs/content/docs/authentication/microsoft.mdx
```diff
@@ -84,6 +84,20 @@ Enabling OAuth with Microsoft Azure Entra ID (formerly Active Directory) allows
...
+
+Better Auth stores the Microsoft Entra `oid` claim as the account identifier for the built-in `microsoft` provider and the Generic OAuth `microsoftEntraId` helper. Microsoft documents `sub` as a pairwise, app-specific identifier, so changing or recreating an app registration can change `sub` for the same user.
+
+If you already have Microsoft accounts created with an older Better Auth version, migrate their account rows before upgrading production traffic:
+
+* Built-in provider rows use `providerId: "microsoft"`.
+* Generic OAuth helper rows use `providerId: "microsoft-entra-id"`.
+* Update each row's account ID from the old `sub` value to the user's `oid` value.
+
+When stored Microsoft ID tokens are available, decode each row's `idToken` and copy the token's `oid` claim into that row's account ID. If you override the account ID with `mapProfileToUser`, for example `${profile.tid}:${profile.oid}` for tenant-qualified identities, migrate existing rows to the exact same format.
```
gustavovalverde: The migration guidance should call out the hard case where stored Microsoft ID tokens are not available. In that case the old account row only has `sub`, so users cannot derive `oid` offline from Better Auth data alone. Please add an explicit warning about duplicate-account risk during rollout and recommend a controlled migration window, such as pausing Microsoft sign-in/linking or migrating from an external Entra export before accepting production traffic on the new identifier.

.changeset/microsoft-oid-account-id.md
```diff
@@ -0,0 +1,6 @@
+---
+"better-auth": major
+"@better-auth/core": major
```
gustavovalverde: Do not use major here, as this will bump the `next` branch to `2.0`

gustavovalverde (changes requested):
A few comments focused on the long-term Microsoft identity contract and the migration surface

GautamBytes:
@gustavovalverde can you review it once ?

GautamBytes:
@gustavovalverde auto-merge was disabled, can you review and merge it whenever feasible!

packages/better-auth/src/plugins/generic-oauth/providers/microsoft-entra-id.ts
```diff
@@ -68,12 +83,39 @@ export function microsoftEntraId(
...
+		const tokenUserInfo = {
+			...tokenProfile,
+			name: getMicrosoftProfileName(tokenProfile),
+			email: tokenProfile.email ?? tokenProfile.preferred_username ?? undefined,
+			image: tokenProfile.picture,
+			emailVerified: tokenProfile.email_verified ?? false,
+		};
+		if (!oid || !tokens.accessToken) {
+			return tokenUserInfo;
+		}
```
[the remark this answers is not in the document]
gustavovalverde: Addressed in ea79d5790f. The central account-key resolver already rejected an empty subject before persistence, but the helper now fails at its own identity boundary and returns null for missing, non-string, or blank oid values. The parameterized regression test also verifies that Graph is never called for those invalid tokens.

packages/better-auth/src/plugins/generic-oauth/providers/microsoft-entra-id.ts
```diff
@@ -84,29 +126,43 @@ export function microsoftEntraId(
...
 	return {
 		providerId: "microsoft-entra-id",
-		accountSubject: ({ profile }) => profile.sub ?? "",
+		accountSubject: ({ profile }) =>
+			typeof profile.oid === "string" ? profile.oid : "",
 		accountIssuer: tenantIssuer,
+		discoveryUrl,
 		authorizationUrl,
 		tokenUrl,
 		userInfoUrl,
```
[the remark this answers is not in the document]
gustavovalverde: Addressed in d8d4c66a4b. Generic OAuth now has an explicit requireIdTokenVerification policy, and the Microsoft helper enables it. Provider initialization fails unless discovery yields the issuer and JWKS configuration needed by the central ID-token verifier, covering both unavailable discovery and incomplete metadata instead of relying only on accountIssuer fallback behavior.

gustavovalverde (approved):
Re-reviewed the Microsoft identity boundary and migration surface. Invalid oid values now fail before profile mapping, and the Generic OAuth helper requires usable discovery verification metadata before trusting ID-token claims.

PR #10231, opened 2026-06-25
feat: add transaction-scoped afterTransaction hook for model creation

### LINKED ISSUE #10202: Add a transaction-scoped database hook after model creation
## Is this suited for github?

Yes, this is suited for github.

## Is your feature request related to a problem? Please describe.

`databaseHooks.<model>.create.before` runs before the row exists, so it cannot create related rows that depend on the new row with a real foreign key.

`databaseHooks.<model>.create.after` runs after the transaction commits. That is good for external side effects, but it means related app-owned database writes cannot be committed atomically with the model creation.

The missing lifecycle point is:

1. insert the model row;
2. run app-owned database writes in the same transaction;
3. commit everything together;
4. if any app-owned write fails, roll back the model creation too.

## Describe the solution you'd like

Add an optional hook that runs after the model row is inserted, but before the transaction commits.

Example API:

```ts
databaseHooks: {
  user: {
    create: {
      afterCreateInTransaction: async (user, ctx) => {
        await ctx.transaction.insert(...);
      },
    },
  },
}
```

This could be available only for adapters that support transactions. For adapters that do not support transactions, Better Auth could omit the transaction context, throw a clear configuration error, or simply not call this hook.

The API does not have to expose a raw ORM transaction if that does not fit Better Auth's adapter model. A storage-agnostic transaction adapter would also work. A raw transaction escape hatch could still be useful for app-owned tables.

## Describe alternatives you've considered

- Use `before`: this can abort creation, but the model row does not exist yet.
- Use `after`: this sees the committed row, but failures cannot roll back model creation.
- Use normal endpoint `hooks.after`: this works for post-commit work, but has the same atomicity problem.
- Add repair/idempotency logic elsewhere in the app: this works, but adds complexity and can leave temporary inconsistent states.
- Replace the built-in signup flow with custom code: this gives full transaction control, but loses the value of Better Auth's built-in flow.

## Additional context

Related: #7260 discussed the same core lifecycle problem. That issue was closed after `after` hooks were moved post-transaction, which is a good default for external side effects.

This request is for a separate, explicitly transaction-scoped hook for database writes that should be atomic with model creation.

Eshan-Jogwar:
Closes #10202

This PR introduces an afterTransaction hook for the user creation flow. This addresses the atomicity gap described in #10202, where developers need to perform related database writes that must be committed atomically with the model creation, or rolled back if the write fails.

Key Changes:

- Implemented afterTransaction hook execution within the createWithHooks transaction block.

- The hook is now awaited inside the runWithTransaction callback; if it throws, the transaction rolls back, ensuring data consistency.

- Added a unit test in src/db/with-hooks.test.ts to verify the control flow and guarantee that hook failures trigger a rollback.

Implementation Details & Limitations:

- Current Scope: Per the request in #10202, this is currently implemented specifically for the user model.

- Extensibility: The architecture is designed for easy expansion. I have structured the internal logic so that this can be enabled for other models by simply updating the hook mapping in createWithHooks. I limited it to the user model for this initial implementation to keep the scope precise, but it can be applied to other collections with minimal effort.

Testing:

- Verified the implementation with a new isolated unit test that mocks the database adapter to simulate both successful commits and rollbacks upon hook failure.

- Note: I observed that 3 existing tests in the suite (db.test.ts and oauth-proxy.test.ts) are failing on main due to local database connection/authentication errors. I have verified these failures persist independently of my changes.

---
## Summary by cubic
Adds a transaction-scoped `afterTransaction` hook for user creation, running all collected hooks in one DB transaction; failures roll back and success commits. Also wraps each hook in `withSpan` for traceable execution, addressing #10202.

- **New Features**
  - Runs all `user.create.afterTransaction` hooks inside `createWithHooks` via `runWithTransaction`; orders multiple hooks and ensures rollback on error, commit on success.
  - Wraps each hook with `withSpan` for tracing (`create.afterTransaction` span with model and source).
  - Extends `BetterAuthOptions` with `user.create.afterTransaction`; limited to the `user` model.
  - Adds tests verifying rollback on failure and commit on success in `packages/better-auth/src/db/with-hooks.test.ts`.

<sup>Written for commit d9e955ca7fe5ee93b69abd2b2c5daf0ba4ee7e7d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10231?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Eshan-Jogwar:
Hi! Just checking in on this PR. Is there anything blocking the review, or any changes you'd like to see? Happy to help with other issues if this is currently on hold.

PR #10234, opened 2026-06-26
fix(siwe): issue addressless nonces

### LINKED ISSUE #8116: SIWE is incompatible with one-step wallet authentication

The current SIWE implementation [associates](https://github.com/better-auth/better-auth/blob/0ce6c7ed7aaefa392fb0c9e649cb0ce2c04e29aa/packages/better-auth/src/plugins/siwe/index.ts#L64) nonces with a specific wallet address & chain ID. This works for wallet connection scenarios where the first step is connecting the wallet, and the second step is signing a SIWE payload for a particular address.

However, with [WalletConnect One-click Auth](https://docs.walletconnect.network/wallet-sdk/web/one-click-auth), we simplify this 2-step interaction to 1-step: connect and sign SIWE in the same step. This improves UX greatly, but unfortunately is incompatible with Better Auth since its nonce generation requires the wallet address and chain ID as parameters, which are unknown to the app prior to connection starting.

The solution should be simple: removing the associated address and chain ID. Only a unique ID per-session necessary to prevent replay attacks, and doesn't need to be associated with a session or wallet address. Furthermore, there's nothing in the [SIWE spec](https://eips.ethereum.org/EIPS/eip-4361) that says that nonces must be unique per-account. The current constraint simply seems unnecessary for the security that the nonce provides.

- Disabling one-click auth for WalletConnect connections -> would worsen UX
- Forking Better Auth to modify this behavior -> would introduce a fork unnecessarily

### Additional context

_No response_

gustavovalverde:
SIWE nonce generation required wallet identity before issuing a nonce, which blocked one-step wallet auth flows where address and chain are only known after signing.

This makes the signed ERC-4361 message the source of truth for wallet identity. Nonce endpoints now issue a single-use challenge without wallet fields, and verification consumes the matching nonce before validating the signed domain, address, Chain ID, and time bounds.

The request schemas reject the old wallet fields instead of stripping them, so client types and runtime behavior agree. Docs and the last-login SIWE integration tests now follow the nonce-first flow.

Fixes #8116.

### DIFF
--- a/.changeset/siwe-addressless-nonces.md
+++ b/.changeset/siwe-addressless-nonces.md
@@ -1,0 +1,5 @@
+---
+"better-auth": minor
+---
+
+The SIWE plugin now issues nonces before the wallet address or Chain ID is known. `authClient.siwe.nonce()` and `authClient.siwe.getNonce()` no longer accept wallet fields, `getNonce` must return an ERC-4361 nonce (8-250 alphanumeric characters), and SIWE verification now reads the wallet address and Chain ID from the signed ERC-4361 message.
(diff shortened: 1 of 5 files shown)


gustavovalverde:
@greptile-apps review

PR #10268, opened 2026-06-28
refactor: export checkPasswordCompromise function

Ionys320:
This allows an external function to check if the password has been pwned without relying on an API call.

This if useful when we define flows outside of the API routes system.

---
## Summary by cubic
Exported `checkPasswordCompromise` from the HaveIBeenPwned plugin so you can run password breach checks in custom flows without calling API routes. No behavior change for existing API usage.

<sup>Written for commit 7f978612536a4dae8f35a4e8351a4325aa2e1612. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10268?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/better-auth/src/plugins/haveibeenpwned/index.ts
+++ b/packages/better-auth/src/plugins/haveibeenpwned/index.ts
@@ -20,7 +20,7 @@ const ERROR_CODES = defineErrorCodes({
 		"The password you entered has been compromised. Please choose a different password.",
 });
 
-async function checkPasswordCompromise(
+export async function checkPasswordCompromise(
 	password: string,
 	customMessage?: string | undefined,
 ) {


packages/better-auth/src/plugins/haveibeenpwned/index.ts
bytaesu, line 46: Hi @Ionys320,

From what I can see, you're exporting this helper to use the logic outside of the Better Auth API. However, it currently throws APIError. We might want to reshape it a bit before exporting it?

Ionys320: Hi @bytaesu! It can be an idea, but I'm also fine by keeping that as it is. It would not be the first package I use that throw a custom object as an error. As long as it's catachable, it seems fine for me.

bytaesu:
Hi! I opened a new PR with a clearer API and a few additional fixes. I'll handle it there.

PR #10270, opened 2026-06-28
fix(cli): escape string default values in generated Prisma schema

mahirhir:
#10259 fixed `better-auth generate` emitting an invalid Drizzle schema when a string default contained a quote or backslash. The Prisma generator has the same bug.

In `packages/cli/src/generators/prisma.ts` the string scalar branch interpolates the value straight into the attribute:

```ts
fieldBuilder.attribute(`default("${attr.defaultValue}")`);
```

So a field with `defaultValue: 'say "hi"'` emits `@default("say "hi"")`, which Prisma can't parse. The JSON object/array branches a few lines up already guard against this by running the value through `.replace(/\\/g, "\\\\").replace(/"/g, '\\"')`; the plain string branch just missed it. This applies the same escaping there.

Normal defaults are unchanged, since the replaces are no-ops unless the value contains `"` or `\`.

Added a regression test next to the Drizzle one from #10259 (it fails on the old code and passes with the fix), plus a changeset.

---
## Summary by cubic
Fixes `better-auth generate` emitting an invalid Prisma schema when a field’s string default contains quotes or backslashes. The Prisma generator now escapes these characters so generated schemas parse correctly.

- **Bug Fixes**
  - Escape quotes and backslashes in string defaults when emitting `@default("...")` in the Prisma generator; other defaults are unchanged.
  - Added a regression test for quoted/backslashed defaults and a changeset.

<sup>Written for commit 406a9f9588d822fc76d2d0a95cda221bf0c57576. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10270?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/prisma-generate-escape-string-default.md
+++ b/.changeset/prisma-generate-escape-string-default.md
@@ -1,0 +1,5 @@
+---
+"auth": patch
+---
+
+Fix `better-auth generate` emitting an invalid Prisma schema when a field's string default value contains a quote or backslash.
--- a/packages/cli/src/generators/prisma.ts
+++ b/packages/cli/src/generators/prisma.ts
@@ -351,7 +351,11 @@ export const generatePrismaSchema: SchemaGenerator = async ({
 						typeof attr.defaultValue === "string" &&
 						provider !== "mysql"
 					) {
-						fieldBuilder.attribute(`default("${attr.defaultValue}")`);
+						// Escape quotes and backslashes so the emitted Prisma default stays a
+						// valid string literal instead of a broken `@default("say "hi"")`.
+						fieldBuilder.attribute(
+							`default("${attr.defaultValue.replace(/\\/g, "\\\\").replace(/"/g, '\\"')}")`,
+						);
 					} else if (
 						typeof attr.defaultValue === "boolean" ||
 						typeof attr.defaultValue === "number"
--- a/packages/cli/test/generate.test.ts
+++ b/packages/cli/test/generate.test.ts
@@ -463,6 +463,42 @@ describe("generate", async () => {
 		expect(schema.code).toContain(String.raw`.default('say "hi"\\done')`);
 	});
 
+	/**
+	 * @see https://github.com/better-auth/better-auth/pull/10259
+	 */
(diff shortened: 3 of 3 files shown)


mahirhir:
552 commits behind, so worth checking whether it still applies. It does, and the codebase already contains the fix one branch away.

On `main` today, the scalar-string branch:

```ts
// packages/cli/src/generators/prisma.ts:561
} else if (typeof attr.defaultValue === "string" && provider !== "mysql") {
  fieldBuilder.attribute(`default("${attr.defaultValue}")`);
}
```

and the JSON-array branch, 53 lines above it:

```ts
// :508
`default("${JSON.stringify(attr.defaultValue).replace(/\\/g, "\\\\").replace(/"/g, '\\"')}")`
```

Same escaping expression this PR adds. So the need is already understood in this file; it is applied to the array path and not the scalar path.

What the two produce, run side by side:

```
  input "say \"hi\""
    :561 scalar  -> default("say "hi"")
    :508 sibling -> default("say \"hi\"")
  input "C:\\path\\to"
    :561 scalar  -> default("C:\path\to")
    :508 sibling -> default("C:\\path\\to")
  input "plain"
    :561 scalar  -> default("plain")           <- unchanged
    :508 sibling -> default("plain")
```

`@default("say "hi"")` is not a valid Prisma string literal, so `better-auth generate` emits a schema that will not parse. The backslash case is quieter and worse: `default("C:\path\to")` parses, and `\p` and `\t` are then whatever Prisma makes of them, so the default silently differs from what the user configured.

The `plain` row is the one that matters for review: ordinary defaults are byte-identical. This only changes output that was already broken.

CI is green and the branch is `MERGEABLE` despite the drift. Happy to rebase onto current `main`, or close this and apply the same `.replace` chain at `:561` directly. Either is fine, no attribution needed.

PR #10296, opened 2026-07-02
fix(auth): use case-insensitive email lookup for manually inserted users

### LINKED ISSUE #10276: Email sign-in does not normalize stored email to lowercase for users

### Reproduction

When manually inserting database entries for users and accounts tables, I was unable to login for a given user when the captialization is not consistent with the database entry.

Steps to reproduce:
1. Insert a user (`Test@eaxmple.com`) into your database with capital letters
2. Insert an account into your database tied to the previously inserted user
3. Attempt to sign in (via `signIn.email` for example) the account you created, but with a different capitialization (`test@example.com`)

I expected the login to succeed due to normalization of the email when fetched from the database, but the sign in failed and logged that the user could not be found.

Since this behavior is out of the norm for creating new users, it might be unreasonable to expect the defensive programming behavior here.

1.6.11

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "#35~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May 26 19:30:42 UTC 2",
    "release": "6.17.0-35-generic",
    "cpuCount": 20,
    "cpuModel": "12th Gen Intel(R) Core(TM) i7-12700",
    "totalMemory": "15.34 GB",
    "freeMemory": "3.52 GB"
  },
  "node": {
    "version": "v24.11.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.1"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "1.6.11",
    "config": null
  }
}
```

Backend

```typescript
// Just need emailAndPassword enabled
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

pranavmamatha:
## Summary

Fixes #10276

Email sign-in fails when users are manually inserted into the database with mixed-case emails (e.g., `Test@Example.COM`). The `findUserByEmail` method lowercases the query input but relies on the database's default case-sensitive comparison, so it cannot match stored emails with different casing.

### `packages/better-auth/src/db/internal-adapter.ts`
- Added `mode: "insensitive"` to the email `where` clause in `findUserByEmail` and both email fallback lookups in `findOAuthUser`
- The adapter's `Where` type already supported this option — it just wasn't being used for email lookups

### `packages/better-auth/src/api/routes/sign-in.test.ts`
- Added a regression test that directly inserts a user with mixed-case email via `db.create` (bypassing `createUser` normalization) and verifies sign-in with lowercase email succeeds

## Verification
- All 18 tests in `sign-in.test.ts` pass
- All 51 tests in `internal-adapter.test.ts` pass
- `pnpm typecheck` passes

---
## Summary by cubic
Make email lookups case-insensitive and handle ambiguous matches in `better-auth`, fixing sign-in and profile updates for users inserted with mixed‑case emails. We resolve a single user by case‑insensitive email, update by `id`, and load accounts separately when requested.

- **Bug Fixes**
  - Use case-insensitive search (`mode: "insensitive"`) with `findMany(limit: 2)` in `findUserByEmail` and both fallbacks in `findOAuthUser`; return `null` on duplicates. Add a regression test for mixed‑case manual insert sign‑in.
  - In `updateUserByEmail`, resolve by case-insensitive email; if exactly one match, update by `id`; otherwise return `null`.
  - Avoid `kysely` limit/join conflicts by fetching users without joins and loading accounts in a separate query when needed.

<sup>Written for commit 0b0c82a623dd2c8e34fbe5c6d3efdd65be794159. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10296?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/petite-lemons-stand.md
+++ b/.changeset/petite-lemons-stand.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Email lookup now uses case-insensitive comparison, fixing sign-in failures for users with mixed-case emails inserted directly into the database
--- a/packages/better-auth/src/api/routes/sign-in.test.ts
+++ b/packages/better-auth/src/api/routes/sign-in.test.ts
@@ -543,4 +543,54 @@ describe("email case insensitivity", async () => {
 			}),
 		).rejects.toThrow();
 	});
+
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/10276
+	 */
+	it("should sign in when the stored email has mixed case from manual DB insertion", async () => {
+		const { auth, db } = await getTestInstance(undefined, {
+			disableTestUser: true,
+		});
+
+		const mixedCaseEmail = "Test@Example.COM";
+		const password = "password123";
+		const hash = await auth.$context.then((ctx) => ctx.password.hash(password));
+
+		// Directly insert user and account into the DB with mixed-case email,
+		// bypassing createUser which would normalize to lowercase.
+		const userId = "manual-user-id";
+		await db.create({
+			model: "user",
+			data: {
+				id: userId,
+				email: mixedCaseEmail, // Intentionally NOT lowercased
+				name: "Manual User",
+				emailVerified: false,
+				createdAt: new Date(),
+				updatedAt: new Date(),
+			},
+			forceAllowId: true,
+		});
+		await db.create({
+			model: "account",
+			data: {
+				userId,
+				providerId: "credential",
+				accountId: userId,
+				password: hash,
+				createdAt: new Date(),
+				updatedAt: new Date(),
+			},
+		});
+
+		// Sign in with lowercase email should find the mixed-case stored email
+		const result = await auth.api.signInEmail({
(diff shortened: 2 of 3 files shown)


ping-maxwell:
I don't think it makes sense to introduce this. The current behavior is intentional and internally consistent. Ideally all forms of inserts, including the manual one, should have emails entirely lower case as a standard.

We're trading a clean, indexable, uniform invariant for a partial, index-defeating, slower version to accommodate data that the framework never created and explicitly normalizes against.

PR #10301, opened 2026-07-02
fix(adapter): coerce numeric-string and number timestamps to Date on read (#9963)

### LINKED ISSUE #9963: Date fields are not guaranteed to be Date objects at call sites (systemic adapter date-coercion gap)
## Summary

Throughout the codebase, code that reads a `date`-typed field (`session.createdAt`, `verification.expiresAt`, `key.expiresAt`, and so on) and then constructs a `Date` from it, calls `.getTime()`, or compares it against the current time assumes the adapter returned a real `Date`. That is not guaranteed. Depending on the adapter, the driver, and the storage path, the same field can arrive as a `Date`, an ISO string, a **numeric string** (`"1774295570569"`), or a **number**.

When the value is a numeric string, the `Date` constructor produces `Invalid Date` and downstream arithmetic becomes `NaN`. When it is a plain string, relational comparisons (`field < new Date()`) silently evaluate to `false`. Either way the failure is silent.

This was surfaced by one call site in the `max_age` work (#9936), but it is a data-layer gap, not local to that plugin. There is already a trail of piecemeal fixes for the same root cause (#7937, #3704, #8104, #8105) and open issues touching adjacent facets (#8619, #8882).

## Root cause

The only central coercion is the adapter factory's read transform, and it leaks three ways:

1. It runs only when `supportsDates === false`, but the default is `true` ([factory.ts#L67](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/core/src/db/adapter/factory.ts#L67)), so Prisma, MongoDB, Memory, and Kysely/postgres|mysql trust the driver's return type with no coercion at all.
2. It matches only `typeof value === "string"` and uses a bare `Date` constructor ([factory.ts#L361-L365](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/core/src/db/adapter/factory.ts#L361-L365)), so a `number` is never coerced and a numeric string becomes `Invalid Date`.
3. Secondary storage and the cookie session cache serialize to JSON and never run the adapter transform on read, so dates come back as strings or numbers.

There is no canonical date-normalization in `@better-auth/core`. Instead there are several private reinventions: `normalizeTimestampValue` and `resolveSessionAuthTime` in oauth-provider ([utils/index.ts#L71](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/oauth-provider/src/utils/index.ts#L71), [#L111](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/oauth-provider/src/utils/index.ts#L111)), `deserializeApiKey` in api-key, `toDate` in cimd, and scattered `new Date(...)` wrapping.

This also corrects the premise of #8619, which assumes the runtime "works correctly" by wrapping values in `new Date()`. It does not, for numeric strings.

## Reachability (calibrated)

Default setups are mostly safe: Prisma/Postgres/Mongo return `Date`; the sqlite path coerces ISO strings; `findSession` re-hydrates session expiry; verification entries self-evict via a store TTL. The gap bites with:

- Adapters or drivers that return dates as numeric strings or numbers (better-sqlite3 INTEGER columns, mysql2 with `dateStrings`, custom adapters).
- A few secondary-storage read paths that do not re-hydrate every field (`session.createdAt`/`updatedAt`).

So this is primarily a robustness and adapter-compatibility gap, with a few security-adjacent edges that only activate under those adapters. A maintainer should decide whether any of those edges warrants separate handling; on its own it does not look advisory-grade, because the worst path (verification-token expiry) is mitigated by the store TTL described below.

## Affected call sites

Security-adjacent (wrong security semantics when the date is a string or number):

- OIDC `max_age` and `auth_time`: [oidc-provider/authorize.ts#L308](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/plugins/oidc-provider/authorize.ts#L308), [#L331](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/plugins/oidc-provider/authorize.ts#L331), [oidc-provider/index.ts#L1063](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/plugins/oidc-provider/index.ts#L1063)
- MCP `auth_time`: [mcp/authorize.ts#L208](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/plugins/mcp/authorize.ts#L208), [mcp/index.ts#L759](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/plugins/mcp/index.ts#L759)
- Fresh-session gate for privileged operations (delete account, change email): [session.ts#L643](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/api/routes/session.ts#L643)
- Session expiry comparison: [session.ts#L370](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/api/routes/session.ts#L370) (and the multi-session and OIDC token-expiry paths)

Correctness and crashes:

- JWKS key handling calls `.getTime()` directly and would throw for a string: [jwt/adapter.ts#L25](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/plugins/jwt/adapter.ts#L25), [jwt/index.ts#L192](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/plugins/jwt/index.ts#L192)
- Session field reshaping wraps with a bare `Date`: [internal-adapter.ts#L465](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/db/internal-adapter.ts#L465), [session.ts#L258](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/api/routes/session.ts#L258)

For contrast, the correct pattern already exists in one place: `consumeVerificationValue` hydrates via `hydrateCachedVerification` ([internal-adapter.ts#L1265](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/db/internal-adapter.ts#L1265)), while its sibling `findVerificationValue` returns the raw JSON ([internal-adapter.ts#L1147](https://github.com/better-auth/better-auth/blob/af572166a26c4c84574ce70f6e850a21a6608d70/packages/better-auth/src/db/internal-adapter.ts#L1147)). The expiry checks on the `findVerificationValue` path (2FA OTP, password reset) would fail open for a string expiry, but the secondary-storage entry self-evicts at its store TTL, so this is not exploitable by default; it only bites a primary DB adapter that returns numeric-string dates.

The oauth-provider plugin already defends its own `max_age`/`auth_time` via `normalizeTimestampValue`. The older oidc-provider and MCP sites above do not.

## Proposed fix

Make the data-layer boundary the single source of truth so call sites can trust a `Date`:

1. Promote `normalizeTimestampValue` into `@better-auth/core` on a stable path (e.g. `@better-auth/core/utils/date`). It already handles `Date`, number, numeric string, and ISO string, and returns `undefined` for genuinely unparseable input.
2. In the factory read transform, coerce **every** `date`-typed field through it, regardless of `supportsDates` (replace the gated bare `new Date(...)`).
3. Re-hydrate dates on the secondary-storage and cookie read paths through the same helper, and update the drizzle `customTransformOutput` similarly.
4. Retire the private reinventions and the scattered `new Date()` wrapping.

This changes the adapter contract to always return a `Date` for `date` fields, which is breaking for any third-party adapter that relied on string pass-through, so it targets `next` with a migration note. It also makes the typing question in #8619 answerable: date fields are `Date`, period.

## Related

- #8619 (types don't reflect `supportsDates`; assumes the runtime is already correct)
- #8882 (drizzle + sqlite INSERT flags, the write path)
- #8618, #3704, #7937, #8104, #8105 (earlier piecemeal coercion fixes)

momomuchu:
## Summary

Fixes #9963 (scoped). On read, better-auth coerces `date`-typed fields back into `Date` objects, but two coercion sites used a bare `new Date(value)`, which returns `Invalid Date` for a raw epoch-millisecond **string** like `"1774295570569"` (the `Date` string constructor only parses ISO 8601, not numeric strings) and skipped raw `number` values entirely. This silently produced `Invalid Date` at call sites (e.g. the OIDC `max_age` check loops on a session whose `createdAt` came back as a numeric string from a Drizzle + libSQL text column).

Closes #9963

## Root cause

Two sites already attempt date coercion on read, both with `new Date(value)`:
1. `packages/core/src/db/adapter/factory.ts` `transformOutput`, gated on `supportsDates === false`, and only checked `typeof === "string"`, so raw `number` was never coerced.
2. `packages/drizzle-adapter/src/drizzle-adapter.ts` `customTransformOutput`, runs unconditionally and is the exact site of the reporter's repro.

`new Date("1774295570569")` → `Invalid Date`.

## Fix

Add `parseDateValue()` in `packages/core/src/db/adapter/utils.ts` (exported via `@better-auth/core/db/adapter`) that correctly parses ISO strings, numeric-millisecond strings, and raw numbers into a `Date`. Wire it into both existing coercion sites.

**Deliberately scoped.** This does NOT change gating logic, adapter contracts, or defaults, it only fixes the two spots that already attempted coercion. The issue's full proposal (coerce every adapter regardless of `supportsDates`, re-hydrate secondary-storage/cookie-cache reads, retire `oauth-provider`'s private timestamp helpers) is an admitted breaking adapter-contract change targeting `next`; that's a maintainer scope decision and is intentionally left out. Every adapter that currently trusts its driver to return `Date` (Prisma, Mongo, Kysely pg/mysql, Memory) is provably untouched.

## Test evidence

- RED confirmed pre-fix in `factory.test.ts` (numeric-string → `Invalid Date`; number → not coerced) and in `drizzle-adapter.test.ts` (reproduces the reported symptom).
- New `parseDateValue` unit tests in core.
- GREEN: `packages/core` 372/372, `packages/drizzle-adapter` 25/25, `packages/kysely-adapter` 21/21 (exercises the `supportsDates:false` branch), `packages/memory-adapter` 16/16, no regressions.
- `tsc` clean for both touched packages; Biome clean.

---
## Summary by cubic
Coerces numeric-string and number timestamps into real Date objects when reading date fields in `@better-auth/core`’s adapter factory and `@better-auth/drizzle-adapter`. Fixes #9963 and stops OIDC `max_age` loops caused by `Invalid Date`.

- Bug Fixes
  - Added `parseDateValue` in `@better-auth/core/db/adapter` to handle `Date`, number, ISO string, and epoch-millisecond numeric-string inputs; only 9+ digit numeric strings are treated as epoch ms.
  - Applied in core `transformOutput` when `supportsDates: false` and in `@better-auth/drizzle-adapter` `customTransformOutput`; `Date`, `null`, and `undefined` pass through unchanged; no adapter contract changes.

<sup>Written for commit 070cf8b343ec85900f46e885a4747defcbc2d701. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10301?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/adapter-date-coercion-9963.md
+++ b/.changeset/adapter-date-coercion-9963.md
@@ -1,0 +1,6 @@
+---
+"@better-auth/core": patch
+"@better-auth/drizzle-adapter": patch
+---
+
+Fix date fields coming back as `Invalid Date` when a driver returns them as a raw epoch-millisecond string or number (e.g. a Drizzle + libSQL text column). The adapter factory's `supportsDates: false` output transform and the Drizzle adapter's `customTransformOutput` now parse numeric-millisecond strings and raw numbers into `Date` in addition to ISO strings.
--- a/packages/core/src/db/adapter/factory.test.ts
+++ b/packages/core/src/db/adapter/factory.test.ts
@@ -231,3 +231,94 @@ describe("createAdapterFactory consumeOne fallback", () => {
 		).rejects.toThrowError(/non-numeric value from deleteMany/);
 	});
 });
+
+describe("createAdapterFactory date field output coercion", () => {
+	function createDateTestAdapter({
+		findOne,
+		supportsDates,
+	}: {
+		findOne: CustomAdapter["findOne"];
+		supportsDates?: boolean;
+	}) {
+		return createAdapterFactory<BetterAuthOptions>({
+			config: {
+				adapterId: "date-test-adapter",
+				adapterName: "Date Test Adapter",
+				supportsDates,
+			},
+			adapter: () => createCustomAdapter({ findOne }),
+		})({});
+	}
+
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/9963
+	 */
+	it("coerces a numeric-millisecond-string date value into a valid Date when supportsDates is false", async () => {
+		const adapter = createDateTestAdapter({
+			supportsDates: false,
+			findOne: async <T>() =>
+				({
+					id: "verification-id",
+					identifier: "token",
+					value: "value",
+					expiresAt: "1774295570569",
+				}) as T,
+		});
+
+		const result = await adapter.findOne<{ expiresAt: unknown }>({
+			model: "verification",
+			where: [{ field: "id", value: "verification-id" }],
+		});
+
+		expect(result?.expiresAt).toBeInstanceOf(Date);
+		expect(Number.isNaN((result?.expiresAt as Date).getTime())).toBe(false);
+		expect((result?.expiresAt as Date).getTime()).toBe(1774295570569);
+	});
+
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/9963
+	 */
+	it("coerces a numeric date value into a valid Date when supportsDates is false", async () => {
+		const adapter = createDateTestAdapter({
+			supportsDates: false,
+			findOne: async <T>() =>
+				({
+					id: "verification-id",
+					identifier: "token",
+					value: "value",
+					expiresAt: 1774295570569,
+				}) as T,
+		});
+
+		const result = await adapter.findOne<{ expiresAt: unknown }>({
+			model: "verification",
+			where: [{ field: "id", value: "verification-id" }],
+		});
+
+		expect(result?.expiresAt).toBeInstanceOf(Date);
+		expect(Number.isNaN((result?.expiresAt as Date).getTime())).toBe(false);
+		expect((result?.expiresAt as Date).getTime()).toBe(1774295570569);
+	});
+
+	it("still coerces a plain ISO date string into a valid Date when supportsDates is false", async () => {
+		const iso = "2026-01-01T00:00:00.000Z";
+		const adapter = createDateTestAdapter({
+			supportsDates: false,
+			findOne: async <T>() =>
+				({
+					id: "verification-id",
+					identifier: "token",
+					value: "value",
+					expiresAt: iso,
+				}) as T,
+		});
+
+		const result = await adapter.findOne<{ expiresAt: unknown }>({
+			model: "verification",
+			where: [{ field: "id", value: "verification-id" }],
+		});
+
+		expect(result?.expiresAt).toBeInstanceOf(Date);
+		expect((result?.expiresAt as Date).toISOString()).toBe(iso);
+	});
+});
--- a/packages/core/src/db/adapter/factory.ts
+++ b/packages/core/src/db/adapter/factory.ts
@@ -31,7 +31,7 @@ import type {
 	AdapterFactoryOptions,
 	AdapterTestDebugLogs,
 } from "./types";
-import { withApplyDefault } from "./utils";
+import { parseDateValue, withApplyDefault } from "./utils";
 
 export {
 	initGetDefaultModelName,
@@ -368,10 +368,17 @@ export const createAdapterFactory =
 							newValue = safeJSONParse(newValue);
 						} else if (
 							config.supportsDates === false &&
-							typeof newValue === "string" &&
+							(typeof newValue === "string" || typeof newValue === "number") &&
 							field.type === "date"
 						) {
-							newValue = new Date(newValue);
+							// `supportsDates: false` means the adapter/driver doesn't
+							// guarantee a `Date` instance back for a `date` field — it may
+							// return an ISO string, epoch milliseconds as a number, or a
+							// numeric-millisecond string. `parseDateValue` handles all
+							// three; a bare `new Date(...)` mis-parses the numeric-string
+							// case into `Invalid Date`.
+							// @see https://github.com/better-auth/better-auth/issues/9963
+							newValue = parseDateValue(newValue);
 						} else if (
 							config.supportsBooleans === false &&
 							typeof newValue === "number" &&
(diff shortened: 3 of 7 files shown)


ping-maxwell (changes requested):
missing changeset.
`bunx changeset`

momomuchu:
Done. Added the changeset in `070cf8b`: `@better-auth/core` and `@better-auth/drizzle-adapter`, both `patch`. Let me know if you'd scope it differently.

PR #10305, opened 2026-07-02
feat(username): handle null email on sign-in + add resend-verification endpoint

### LINKED ISSUE #10163: allow sending verification email via username

when a user signs in via username, we send an "email verification" email if that's configured. If the email failed to send for some reason, there isn't a way to re-send it if we do not know the user's email.

We should allow sending that email verification email based on username

-

### Additional context

This issue is opened for tracking purposes.

AnandSundar:
## Summary

Fixes #10163.

The username plugin had two related defects around email verification:

1. `/sign-in/username` called `createEmailVerificationToken` unconditionally when `emailVerification.sendOnSignIn` was set. Users whose `user.email` was null (typical for OAuth-only or username-pluggable flows) crashed with a `TypeError` instead of returning the documented `EMAIL_NOT_VERIFIED` shape. Gate the token creation on `user.email`.

2. There was no way to re-trigger a verification email from the username-only path. After hitting the documented `EMAIL_NOT_VERIFIED` on sign-in, the only resend option was `/send-verification-email` which requires the email itself — the very thing the user may not remember or have access to. Add `POST /username/send-verification-email`.

## Endpoint

```
POST /username/send-verification-email
{
  "username": "...",
  "password": "...",
  "callbackURL": "..."   // optional
}
```

Returns `{ success: true }` on success and `401 INVALID_USERNAME_OR_PASSWORD` for both missing username and wrong password (no enumeration). Idempotent: already-verified or no-email users get `success` without a send. Falls back to `403 EMAIL_NOT_VERIFIED` only when no `sendVerificationEmail` handler is configured (same shape as the sign-in branch — never a generic 500).

Exposed on the client as `client.username.sendVerificationEmail(...)`.

## Why a new endpoint instead of a query param

Path conflict: `/sign-in/username/send-verification-email` collides with `/sign-in/username` at `client.signIn.username`. `/username/send-verification-email` keeps the type-level `PathToObject<>` mapping clean. Endpoints and method names are independent of sign-in vs. sign-up semantics; the body still mirrors sign-in's shape.

## Security

- Wrong password returns the same code (no enumeration).
- Missing user runs a dummy password hash to keep timing constant with the found-user path.
- Verifies the password before sending — the endpoint is a trigger flow that must authenticate the caller; otherwise any anonymous visitor could spam verification emails to any registered username.

## Tests

`packages/better-auth/src/plugins/username/username.test.ts`:
- New: returns `EMAIL_NOT_VERIFIED` for unverified user with null email (R1).
- New `describe("username send-verification-email ...")` block: 7 cases covering happy path, callbackURL echo, no-email silent-success, missing `sendVerificationEmail` handler (R4), missing username, wrong password (no enumeration), and already-verified idempotency.

All 52 username tests pass; adjacent `email-verification` (24) and `sign-in` (17) suites remain green.

## Out of scope (follow-up)

Obfuscating the email in the `EMAIL_NOT_VERIFIED` response body so the front-end can show a hint without revealing the address. Will arrive as a separate PR.

---
## Summary by cubic
Fixes a crash on username sign-in when a user has no email and adds a username-based resend verification endpoint. Also hardens the resend flow to prevent username/account enumeration and resolves an internal endpoint name collision.

- **New Features**
  - Adds `POST /username/send-verification-email` to resend by username; body `{ username, password, callbackURL? }`, returns `{ success: true }` or `401 INVALID_USERNAME_OR_PASSWORD`; exposes `client.username.sendVerificationEmail(...)`.
  - Authenticates with password, is idempotent (already verified or no email → success), echoes `callbackURL`, and returns `403 EMAIL_NOT_VERIFIED` when no `sendVerificationEmail` handler is configured.

- **Bug Fixes**
  - In `/sign-in/username`, only create a verification token if `user.email` exists when `emailVerification.sendOnSignIn` is enabled; prevents a crash and correctly returns `EMAIL_NOT_VERIFIED`.
  - In `/username/send-verification-email`, validate with `getUsernameToValidate`, normalize once for lookup, and dummy-hash on `!user`, `!account`, and `!currentPassword` branches to keep timing uniform and avoid enumeration.
  - Renames internal endpoint key to `sendUsernameVerificationEmail` to avoid colliding with the top-level `/send-verification-email`; path and client API remain unchanged.

<sup>Written for commit f1cd910a158b21ab25bc3370ea802bd1b3d19b4a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10305?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


AnandSundar:
Addressed all three bot findings on the resend-verification endpoint in bf2bbda8e:

- **Validation-order inversion** (Cubic P2): the new handler now calls the canonical `getUsernameToValidate` helper. Pre-normalization (default) validates the raw input as documented.
- **Double normalization** (Cubic P2): DB lookup normalizes `ctx.body.username` exactly once, so a custom non-idempotent `usernameNormalization` still resolves the user correctly.
- **Timing gap** (Greptile + Cubic P2): `!account` and `!currentPassword` branches now run the same dummy `password.hash` the `!user` branch already runs. No more side channel distinguishing "username resolves to a credential account" from "username resolves to a user with no credential account."
- **Wrong-password test** (Greptile): `emailSent` flipped from `boolean` to `number` mid-block — strict-mode TS error. Aligned with the rest of the `describe` by using a numeric counter.

Test status: all 52 username tests pass; adjacent `email-verification` (24) and `sign-in` (17) suites remain green. biome + spell pre-commit hooks pass.

### Out of scope (will follow up separately)

The same `validationOrder.username` inversion exists in the pre-existing `/sign-in/username` handler (it predates this PR and was not flagged in this review). Happy to fix it as a follow-up PR so it gets its own review surface — leaving it untouched here to keep this commit scoped to the diff the bots actually reviewed.

AnandSundar:
@vlucas @zain @joshnuss @linyiru Could someone please review my PR? Thank you.

packages/better-auth/src/plugins/username/index.ts
```diff
@@ -553,6 +574,188 @@ export const username = (options?: UsernameOptions | undefined) => {
 					});
 				},
 			),
+			sendVerificationEmail: createAuthEndpoint(
```
ping-maxwell: This can't be `sendVerificationEmail`, it's already used.
Change to `sendVerificationEmailFromUsername` or something shorter if you can think of one

ping-maxwell (changes requested):

AnandSundar:
Renamed the endpoint JS key to `sendUsernameVerificationEmail` in f1cd910. Picked this over `sendVerificationEmailFromUsername` because it preserves the verb-object pattern of the outer handler and reads more naturally as 'send the username's verification email'.

Only the JS key changed — endpoint path stays `POST /username/send-verification-email`, and `client.username.sendVerificationEmail(...)` is path-derived via `PathToObject<...>` so it continues to work without any test-side renames. Side benefit: `baseEndpoints.sendVerificationEmail` is no longer shadowed in the `getEndpoints` spread, so top-level `client.sendVerificationEmail(...)` is preserved when the username plugin is installed.

Also included two pre-existing typecheck fixes in `username.test.ts` (lines 835 and 859) that this PR's CI would have surfaced on first run — `as unknown as { username: string }` cast on the captured user assignment and `captured!` non-null assert on the assertion line. These don't affect runtime behavior; tests still pass 47/47 in `username.test.ts`, plus the adjacent email-verification suite stays 24/24. CI typecheck is now green locally.

@ping-maxwell could you take another look?

PR #10318, opened 2026-07-05
fix(core): resolve AsyncLocalStorage from globalThis synchronously on edge runtimes

### LINKED ISSUE #10315: Lazily-cached init promises hang forever on Cloudflare Workers when the initializing request is aborted

### Reproduction

The bug is a runtime-environment interaction (workerd's "hanging promise" semantics), so the reproduction is behavioral rather than a repo:

1. Deploy a Worker (`nodejs_compat`) that caches the auth instance per isolate (`cachedAuth ??= betterAuth(...)`) and serves an authenticated JSON endpoint via `auth.api.getSession({ headers })`.
2. From a browser, issue authenticated fetches that are frequently superseded and aborted mid-flight (`AbortController.abort()` on the in-flight request when a newer one starts) — e.g. a page that refetches on every user action.
3. Keep enough traffic that Cloudflare spins up fresh isolates while aborts are happening. Eventually an abort lands on the request that is lazily driving better-auth's first-touch initialization in a fresh isolate.
4. From that point, every better-auth API call routed to that isolate hangs forever, before any DB access. Because a client's connection tends to stick to one isolate, that user sees a full outage while other users are unaffected.

Mechanism: on Workers, a promise created while handling a request belongs to that request's IoContext; when the client disconnects first, workerd never settles it. better-auth caches such promises across requests:

1. `@better-auth/core/src/async_hooks/index.ts` — module-level `AsyncLocalStoragePromise = import("node:async_hooks")...`, awaited by `ensureAsyncStorage` in `endpoint-context.ts`, `request-state.ts`, and `transaction.ts` until the storage instance is created. Every endpoint dispatch goes through `runWithEndpointContext`, so a poisoned promise here hangs **all** API calls.
2. `createBetterAuth` — `const authContext = initFn(options)` cached on the instance (`$context`), awaited by every call.

Diagnosis evidence from our production incident:

- A `chrome://net-export` capture during an episode showed the HTTP/2 connection was healthy (h2 PINGs ACKed in milliseconds throughout; edge-served static assets on the same connection responded normally) — only Worker-bound streams got no response, and invocations ended with outcome `canceled` once the client gave up.
- Phase markers (`console.info` around each await; canceled invocations flush logs) showed hung requests completed `await auth.$context` but `auth.api.getSession({ headers })` never reached its first D1 query, while a raw `env.DB.prepare('SELECT 1')` probe in the same isolate succeeded. So D1 and the isolate event loop were fine; the hang sat in the dispatch layer before DB access, consistent with `ensureAsyncStorage` awaiting a permanently pending cached import promise.
- For us the trigger became deterministic: two quick comment posts on a viewer page (firing supersede-aborts of in-flight authenticated fetches) reproduced the poisoned-isolate state almost every time.

Current: after a client abort lands on the request driving lazy initialization in a fresh isolate, every later better-auth API call in that isolate awaits a permanently pending cached promise and hangs forever (no error, no timeout). The isolate never recovers until it is recycled.

Expected: initialization is abort-safe — a client abort during first-touch initialization must not poison the cached state for subsequent requests. Options that would achieve this:

- Resolve `AsyncLocalStorage` via a static import in the workerd/edge bundle (`nodejs_compat` provides `node:async_hooks`), or create the storages synchronously when `globalThis.AsyncLocalStorage` exists.
- Document (or expose an API for) eagerly initializing `$context` and the ALS storages outside a request, so integrators can anchor them with `ctx.waitUntil`.
- Or avoid caching pending promises: cache only resolved values and retry initialization if a previous attempt never settled within a bound.

Workarounds we shipped (verified: the deterministic reproduction no longer hangs):

- Pre-seed the three AsyncLocalStorage instances from a static `node:async_hooks` import at global scope (keys `requestStateAsyncStorage`, `endpointContextAsyncStorage`, `adapterAsyncStorage` on `globalThis[Symbol.for("better-auth:global")].context`), so the dynamic-import path is never taken.
- Anchor `auth.$context` to `ctx.waitUntil` once per isolate.
- Wrap `auth.api.getSession` in a `Promise.race` hang detector that rebuilds the instance once and fails null.

1.6.20

### System info

```bash
# npx auth info is not practical inside the Workers runtime; environment summary:
better-auth: 1.6.20
@better-auth/core: 1.6.20
runtime: Cloudflare Workers (workerd), compatibility_flags: ["nodejs_compat"]
database: Cloudflare D1 via kysely-d1 0.4.0 (Kysely 0.29.2), passed as { database: { dialect, type: "sqlite" } }
framework: React Router 8 (framework mode) on the same Worker
plugins: jwt, oauth-provider, device-authorization, email-otp, last-login-method
```

Backend

```typescript
import { D1Dialect } from "kysely-d1"

const dialect = new D1Dialect({ database: env.DB })

let cachedAuth: ReturnType<typeof buildAuth> | undefined
export function createAuth() {
  return (cachedAuth ??= buildAuth())
}

function buildAuth() {
  return betterAuth({
    baseURL: env.BETTER_AUTH_URL,
    secret: env.BETTER_AUTH_SECRET,
    database: { dialect, type: "sqlite" },
    plugins: [jwt(), oauthProvider({ /* ... */ }), deviceAuthorization({ /* ... */ }), emailOTP({ /* ... */ }), lastLoginMethod()],
    socialProviders: { google: { /* ... */ }, microsoft: { /* ... */ } },
  })
}
```

### Additional context

- Only reproducible under real client-abort traffic on Workers; local dev and non-aborting load tests never hit it.
- #10300 reports a different failure mode (a check-await-assign concurrency race) in the same `ensureAsyncStorage` lazy initialization. Both bugs share the same root — `AsyncLocalStorage` resolved through an awaited, lazily-cached dynamic import driven by request traffic — and eager initialization from a static import would fix both.
- Related past reports of the dynamic-import fragility in other environments: #5539 (`node:async_hooks` dynamic import failing to resolve), #8428 (`bun build --compile`).

Thanks for better-auth — happy to provide more detail or test a patch.

tsushanth:
## Problem

`AsyncLocalStoragePromise` is a module-level singleton initialised by a dynamic `import("node:async_hooks")`. On Cloudflare Workers, dynamic imports are bound to the request's IoContext — if the request driving lazy first-touch initialisation is aborted by the client before the import settles, workerd never resolves the promise.

Because the promise is cached at module scope, every subsequent request in the same isolate awaits the same permanently-pending promise and hangs forever with no error and no timeout. The isolate must be recycled to recover, and a client whose connection sticks to that isolate sees a complete outage while other isolates are unaffected.

Reported with a production incident trace in #10315.

## Root cause

```ts
// module scope — created once, shared across all requests in the isolate
const AsyncLocalStoragePromise = import("node:async_hooks")  // request-bound on Workers
  .then(mod => mod.AsyncLocalStorage)
  .catch(...);
```

## Fix

Workers with `nodejs_compat` expose `AsyncLocalStorage` synchronously on `globalThis`. Check for it before launching the dynamic import and resolve immediately with `Promise.resolve()` — a microtask-settled promise with no IoContext attachment that cannot be poisoned by an abort.

```ts
const _globalALS = (globalThis as any).AsyncLocalStorage;

const AsyncLocalStoragePromise = _globalALS
  ? Promise.resolve(_globalALS)        // Workers: synchronous, abort-safe
  : import("node:async_hooks")...;     // Node.js / Bun: unchanged
```

Node.js and Bun do not set `globalThis.AsyncLocalStorage`, so they fall through to the existing dynamic import path with no behaviour change.

## What this doesn't fix

The `$context` caching hang (a second hang vector mentioned in #10315 — if the request driving `initFn(options)` is aborted before it resolves, the cached pending promise poisons the instance) is a separate issue and not addressed here.

Fixes #10315

---
## Summary by cubic
Resolve `AsyncLocalStorage` from `globalThis` on edge runtimes to prevent dynamic import hangs on Cloudflare Workers. Fixes #10315 by making the module-level resolution abort-safe while keeping Node/Bun behavior unchanged.

- **Bug Fixes**
  - Use `globalThis.AsyncLocalStorage` when available and wrap with `Promise.resolve(...)` (abort-safe on Workers).
  - Fallback to `import('node:async_hooks')` on Node.js/Bun; browser environments return `null`, otherwise warnings are logged when unavailable.

<sup>Written for commit 4c4af01bb30426e8416d74fb8b57782afcd71365. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10318?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/core/src/async_hooks/index.ts
+++ b/packages/core/src/async_hooks/index.ts
@@ -2,31 +2,39 @@ import type { AsyncLocalStorage } from "node:async_hooks";
 
 export type { AsyncLocalStorage };
 
+// On Cloudflare Workers (nodejs_compat) and some edge runtimes, AsyncLocalStorage
+// is available synchronously on globalThis. Resolve it without a dynamic import so
+// we never create a request-scoped promise: if the request that drives the lazy
+// import is aborted, workerd never settles the promise and every subsequent request
+// in the same isolate hangs forever. The synchronous path is immune to that.
+const _globalALS = (globalThis as any).AsyncLocalStorage as
+	| typeof AsyncLocalStorage
+	| undefined;
+
 const AsyncLocalStoragePromise: Promise<typeof AsyncLocalStorage | null> =
-	import(
-		/* @vite-ignore */
-		/* webpackIgnore: true */
-		"node:async_hooks"
-	)
-		.then((mod) => mod.AsyncLocalStorage)
-		.catch((err) => {
-			if ("AsyncLocalStorage" in globalThis) {
-				return (globalThis as any).AsyncLocalStorage;
-			}
-			if (typeof window !== "undefined") {
-				return null;
-			}
-			console.warn(
-				"[better-auth] Warning: AsyncLocalStorage is not available in this environment. Some features may not work as expected.",
-			);
-			console.warn(
-				"[better-auth] Please read more about this warning at https://better-auth.com/docs/installation#mount-handler",
-			);
-			console.warn(
-				"[better-auth] If you are using Cloudflare Workers, please see: https://developers.cloudflare.com/workers/configuration/compatibility-flags/#nodejs-compatibility-flag",
-			);
-			throw err;
-		});
+	_globalALS
+		? Promise.resolve(_globalALS)
+		: import(
+				/* @vite-ignore */
+				/* webpackIgnore: true */
+				"node:async_hooks"
+			)
+				.then((mod) => mod.AsyncLocalStorage)
+				.catch((err) => {
+					if (typeof window !== "undefined") {
+						return null;
+					}
+					console.warn(
+						"[better-auth] Warning: AsyncLocalStorage is not available in this environment. Some features may not work as expected.",
+					);
+					console.warn(
+						"[better-auth] Please read more about this warning at https://better-auth.com/docs/installation#mount-handler",
+					);
+					console.warn(
+						"[better-auth] If you are using Cloudflare Workers, please see: https://developers.cloudflare.com/workers/configuration/compatibility-flags/#nodejs-compatibility-flag",
+					);
+					throw err;
+				});
 
 export async function getAsyncLocalStorage(): Promise<
 	typeof AsyncLocalStorage


tsushanth:
Rebased cleanly onto current `main` — this had drifted very far (my branch was based on a point before v1.6.23 and the 1.7 RC work). Isolated the actual fix to a single cherry-picked commit; `tsc --noEmit` passes clean on `packages/core`.

cosbgn:
Any updates on this? It would fix a lot of issues on cloudflare workers which are really hard to find. Everything works locally and then on workers it just hungs forever. The fix is a one line change, so it would be great to get it merged.

PR #10334, opened 2026-07-08
fix(oauth-provider): persist null revoked refresh tokens

### LINKED ISSUE #10309: oauth-provider: refresh rotation CAS guards on `revoked eq null` but rows are created without the field — 400 invalid_grant on adapters that distinguish null from missing
# oauth-provider: refresh rotation guard matches `revoked eq null`, but rows are created without the `revoked` field — breaks on adapters that distinguish null from missing

## Versions

- `@better-auth/oauth-provider` **1.6.22** (same code in 1.6.23 and on `main` @ [`7f510ce`](https://github.com/better-auth/better-auth/blob/7f510cef533d432cc2ea8c23126a8db7617ede2a/packages/oauth-provider/src/token.ts#L343-L400))
- Observed with `@convex-dev/better-auth` 0.12.4/0.12.5; reproduced end-to-end 2026-07-03

## Summary

`createRefreshToken` inserts `oauthRefreshToken` rows **without** a `revoked` key:

https://github.com/better-auth/better-auth/blob/7f510cef533d432cc2ea8c23126a8db7617ede2a/packages/oauth-provider/src/token.ts#L343-L353

but the rotation compare-and-swap guards on `revoked eq null`:

https://github.com/better-auth/better-auth/blob/7f510cef533d432cc2ea8c23126a8db7617ede2a/packages/oauth-provider/src/token.ts#L385-L395

```ts
where: [
  { field: "id", value: originalRefresh.id },
  { field: "revoked", operator: "eq", value: null },
],
```

This implicitly assumes every adapter treats a **never-written optional field** as equal to `null`. That holds for the first-party adapters (SQL: unset optional column IS NULL; Mongo: `{field: null}` matches missing) but not for third-party document-store adapters where "missing" and "null" are distinct values. On the Convex community adapter (`@convex-dev/better-auth`), whose `eq` filter is strict `value === w.value`, the `revoked eq null` leg never matches — core's `incrementOne` fallback (`findMany` + `updateMany`) finds no row, `incrementOne` returns null, and the plugin throws.

**Net effect there: every `refresh_token` grant — including the very first, with no concurrency — fails with 400 `invalid_grant` / `"invalid refresh token"`.** Adapter-side issue with full analysis and e2e repro: get-convex/better-auth#399.

The same "create without the field, later match `eq null`" pattern exists elsewhere, e.g. device-authorization's `userId` CAS ([`routes.mjs` guard](https://github.com/better-auth/better-auth/blob/7f510cef533d432cc2ea8c23126a8db7617ede2a/packages/device-authorization/src/routes.ts)), so any adapter with strict null semantics breaks those too.

## Repro (against any adapter that distinguishes missing from null)

1. authorize + PKCE → `/oauth2/token` code exchange with `offline_access` → refresh token issued ✅
2. `grant_type=refresh_token` → 400 `invalid_grant` "invalid refresh token" on first refresh ❌ (expected 200 with rotation)

Minimal adapter-level version: `create({ model: "oauthRefreshToken", data: { /* no revoked */ } })` then `findMany` with `where: [{ field: "revoked", operator: "eq", value: null }]` → 0 rows.

## Suggested fix

Cheap and fully backward-compatible: **write the field explicitly at creation** so the CAS guard matches by construction on every adapter —

```ts
const newRow = {
  ...,
  revoked: null,
};
```

(or give `revoked` a `defaultValue: null` in the schema). Alternatively/additionally, document the intended `Where { value: null }` contract for adapter authors — today it's only implied by first-party adapter behavior.

I've filed the adapter-side fix suggestion (null/undefined-coalescing `eq`, plus a native atomic `incrementOne`) at get-convex/better-auth#399; fixing either side unbreaks the grant, fixing both makes the pattern robust. Happy to PR the one-liner here if you'd take it.

justjavac:
## Summary
- Persist newly issued OAuth refresh token rows with an explicit `revoked: null` value.
- Add a regression test covering adapters that distinguish missing fields from null during refresh token rotation.
- Add a patch changeset for `@better-auth/oauth-provider`.

Fixes #10309

## Tests
- `pnpm exec vitest --config vitest.config.ts packages/oauth-provider/src/token.test.ts -t "persists unrevoked refresh token rows with explicit null revoked value"`
- `pnpm exec vitest --config vitest.config.ts packages/oauth-provider/src/token.test.ts -t "oauth token - refresh_token"`
- `pnpm typecheck`

---
## Summary by cubic
Persist `revoked: null` for newly issued OAuth refresh tokens to make rotation compare-and-swap checks consistent across adapters. Fixes #10309.

- **Bug Fixes**
  - Set `revoked: null` on insert and update `OAuthRefreshToken` to `revoked?: Date | null` to align behavior across adapters.
  - Add a regression test and a patch changeset for `@better-auth/oauth-provider`.

<sup>Written for commit dc781f2fc20aa47facea581b65b959bac25513ec. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10334?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/fix-oauth-provider-refresh-revoked-null.md
+++ b/.changeset/fix-oauth-provider-refresh-revoked-null.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/oauth-provider": patch
+---
+
+Persist new OAuth refresh token rows with an explicit `revoked: null` value so rotation compare-and-swap guards work with adapters that distinguish missing fields from null.
--- a/packages/oauth-provider/src/token.test.ts
+++ b/packages/oauth-provider/src/token.test.ts
@@ -663,6 +663,44 @@ describe("oauth token - refresh_token", async () => {
 		expect(tokens?.refresh_token).not.toEqual(newTokens.data?.refresh_token);
 	});
 
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/10309
+	 */
+	it("persists unrevoked refresh token rows with explicit null revoked value", async () => {
+		if (!oauthClient?.client_id || !oauthClient?.client_secret) {
+			throw Error("beforeAll not run properly");
+		}
+
+		const context = await authorizationServer.$context;
+		const existingRefreshRows = await context.adapter.findMany<{
+			id: string;
+		}>({
+			model: "oauthRefreshToken",
+			where: [{ field: "clientId", value: oauthClient.client_id }],
+		});
+		const existingRefreshIds = new Set(
+			existingRefreshRows.map((row) => row.id),
+		);
+
+		const scopes = ["openid", "profile", "offline_access"];
+		const tokens = await authorizeForRefreshToken(scopes);
+		expect(tokens?.refresh_token).toBeDefined();
+
+		const refreshRows = await context.adapter.findMany<{
+			id: string;
+			revoked: Date | null;
+		}>({
+			model: "oauthRefreshToken",
+			where: [{ field: "clientId", value: oauthClient.client_id }],
+		});
+		const newRefreshRows = refreshRows.filter(
+			(row) => !existingRefreshIds.has(row.id),
+		);
+
+		expect(newRefreshRows).toHaveLength(1);
+		expect(newRefreshRows[0]?.revoked).toBeNull();
+	});
+
 	it("should preserve auth_time in id_token after refresh (OIDC Core 1.0 Section 12.2)", async ({
 		expect,
 	}) => {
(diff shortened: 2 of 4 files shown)


Bekacru:
cc @gustavovalverde

PR #10350, opened 2026-07-09
feat(device-authorization): support RFC 8707 resource indicators and RFC 9068 JWT access tokens

### LINKED ISSUE #10345: feat(device-authorization): support RFC 8707 resource indicators to issue RFC 9068 JWT access tokens

The `device-authorization` plugin can only issue an **opaque** access token. On `POST /device/token` it creates a normal Better Auth session and returns `session.token`; neither `/device/code` nor `/device/token` accepts a `resource`/`audience` parameter. There is therefore no way to obtain a **JWT access token audience-restricted to a downstream API** from the device flow (nor to exchange the opaque token for one).

This is a gap relative to `@better-auth/oauth-provider`, which already supports RFC 8707 resource indicators → JWT access tokens (`checkResource` / `createJwtAccessToken`), and relative to other providers (e.g. Auth0's device flow, which issues a JWT when a registered `audience`/`resource` is requested).

Add optional **RFC 8707 `resource`** support to the device flow:

- Accept `resource` (string or repeated) at **`/device/code`** (bound to the device code, available for consent) **and** at **`/device/token`**, applying the RFC 8707 §2.2 subset rule when supplied at both.
- Validate each resource against a new `allowedResources` allow-list (absolute-URI, no fragment); reject unknown/invalid with **`invalid_target`**.
- When a valid resource resolves, issue an **RFC 9068 JWT** access token (`typ: at+jwt`, `aud` = resource, `iss`/`sub`/`client_id`/`azp`/`scope`/`iat`/`exp`/`jti`), signed via the existing `jwt` plugin's JWKS. With no `resource`, behavior is **unchanged** (opaque session token) — fully backward compatible.
- New options: `allowedResources`, `customAccessTokenClaims`. The `jwt` plugin is a prerequisite for JWT issuance (resource servers verify via `/jwks`).

- **Token-endpoint only** or **device-code-request only** (Auth0's model). Supporting **both** (chosen) follows RFC 8707 §2.1 + §2.2 and preserves user consent to the audience.
- A separate **RFC 8693 token-exchange** endpoint to convert an existing opaque token into a JWT — deferred as out of scope.

### Additional context

- Standards: RFC 8628 (device grant), RFC 8707 (resource indicators), RFC 9068 (JWT access token profile).
- Related: #8298 (RFC 8707 for oauth-provider), #4268 (OIDC JWT access token).
- Backward compatible; issued JWT is stateless (documented trade-off: not revocable before expiry — keep `expirationTime` short).
- I have a design + implementation plan ready and am happy to submit a PR (targeting `next` per the contribution policy).

sicarius97:
The device authorization plugin could only issue opaque session tokens, which prevented device clients from requesting access tokens scoped to downstream APIs. This adds an RFC 8707 resource boundary and an RFC 9068 JWT path while preserving the existing opaque-token behavior for requests without `resource`.

## What changes

- **Authorization request**: `/device/code` accepts one or more `resource` values in JSON or form-encoded requests, preserves repeated form parameters, and binds the resolved audience to the device grant.
- **User approval**: `GET /device` returns the requesting `client_id`, `scope`, and `resource` only to the authenticated user who owns the request, so approval pages can present the complete grant before accepting it.
- **Token issuance**: `/device/token` can use the full bound resource set or request a subset. A valid resource produces an RFC 9068 JWT whose verification keys come from the JWT plugin's configured JWKS setup; the no-resource path continues creating a Better Auth session and returning its opaque token.
- **Policy controls**: `allowedResources` defines the global resource allow-list, while `onDeviceAuthRequest` can enforce client-specific resource and scope policy. Invalid targets fail before consuming an approved code, and `customAccessTokenClaims` cannot override authorization-server-owned claims.
- **Storage**: device grants persist their approved audience in the nullable `deviceCode.resource` field.

> [!IMPORTANT]
> After upgrading, run the Better Auth schema migration or generation workflow so the `deviceCode` table includes the nullable `resource` field.

## Review focus

- **Grant boundary**: the token endpoint rejects resources that were not approved at `/device/code`, and a token-time resource must be a subset of the bound grant.
- **Issuance failure**: the JWT plugin requirement and resource policy are checked before the approved code is consumed, so a configuration or request error does not force the user through another device flow.
- **Multi-audience tokens**: deployments must ensure that every requested scope has the same meaning for every resource in a multi-resource audience.

Closes #10345.

### DIFF
--- a/.changeset/device-authorization-resource-indicators.md
+++ b/.changeset/device-authorization-resource-indicators.md
@@ -2,4 +2,4 @@
 "better-auth": minor
 ---
 
-Add RFC 8707 resource indicator support to the device authorization plugin. Passing a `resource` to `/device/code` and/or `/device/token` now returns an RFC 9068 JWT access token audience-restricted to that resource, validated against a new `allowedResources` option, with an optional `customAccessTokenClaims` hook. Without a `resource`, behavior is unchanged (opaque token).
+Add RFC 8707 resource indicator support to the device authorization plugin. Bind one or more allowed resources at `/device/code`, then omit `resource` at `/device/token` to use the full grant or request a subset. The token endpoint returns an RFC 9068 JWT access token restricted to the resolved audience, and `GET /device` exposes the bound client, scope, and resource to the authenticated user for approval. The existing opaque-token behavior remains unchanged when the device request omits `resource`.
--- a/.changeset/device-authorization-resource-indicators.md
+++ b/.changeset/device-authorization-resource-indicators.md
@@ -2,4 +2,4 @@
 "better-auth": minor
 ---
 
-Add RFC 8707 resource indicator support to the device authorization plugin. Bind one or more allowed resources at `/device/code`, then omit `resource` at `/device/token` to use the full grant or request a subset. The token endpoint returns an RFC 9068 JWT access token restricted to the resolved audience, and `GET /device` exposes the bound client, scope, and resource to the authenticated user for approval. The existing opaque-token behavior remains unchanged when the device request omits `resource`.
+Device authorization can now bind one or more RFC 8707 `resource` values at `/device/code` and issue an RFC 9068 JWT access token for the approved audience, or a requested subset, at `/device/token`. Enable this path with the JWT plugin and `allowedResources`; requests without `resource` continue returning opaque tokens. Authenticated approval pages can read the requested `client_id`, `scope`, and `resource` from `GET /device`, and `customAccessTokenClaims` can add non-reserved claims to issued JWTs. Re-run Better Auth schema migration or generation to add the nullable `deviceCode.resource` field.
--- a/.changeset/device-authorization-resource-indicators.md
+++ b/.changeset/device-authorization-resource-indicators.md
@@ -1,0 +1,5 @@
+---
+"better-auth": minor
+---
+
+Add RFC 8707 resource indicator support to the device authorization plugin. Passing a `resource` to `/device/code` and/or `/device/token` now returns an RFC 9068 JWT access token audience-restricted to that resource, validated against a new `allowedResources` option, with an optional `customAccessTokenClaims` hook. Without a `resource`, behavior is unchanged (opaque token).
--- a/docs/content/docs/plugins/device-authorization.mdx
+++ b/docs/content/docs/plugins/device-authorization.mdx
@@ -247,8 +247,10 @@ The user authorization flow requires two steps:
 1. **Code Verification**: Validate the user code via `GET /device`. The verification request claims the pending device code for the calling session.
 2. **Authorization**: The session that claimed the code can approve or deny it.
 
+The verification response includes the requesting `client_id`, `scope`, and `resource` for the authenticated user who owns the code. Display these values before offering approval so the user can review which client is requesting which permissions and audiences.
+
 <Callout type="warn">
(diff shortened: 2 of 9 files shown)


sicarius97:
Thanks for the reviews — addressed below. I also **rebased the branch onto `next`** (it had been branched off a fork branch that carried unrelated commits); the PR now contains only the device-authorization work, and it's mergeable.

### Review findings

- **RFC 8707 §2.2 — token-time-only resource** (Greptile P1): **Fixed.** A `resource` supplied at `/device/token` that was not bound at the `/device/code` request is now rejected with `invalid_target`; the token's audience must be authorized at the device-code request (Auth0-style binding). One clarification for the record: RFC 8707 §2.2 is actually *permissive* here (*"the authorization server may limit the acceptable resources to those originally granted"*), so this was a policy choice rather than a spec violation — but binding to the authorization request is the safer default, so I've adopted it.

- **Device code consumed before resource validation** (Greptile P1): **Fixed.** The requested `resource` is now validated *before* `consumeOne`, so an invalid/disallowed resource (or an `allowedResources` change after issuance) returns `invalid_target` without burning the approved code.

- **`typ: at+jwt` not forwarded to custom signers** (Greptile P2): **Resolved by the rebase.** The access token now sets `at+jwt` via `signJWT`'s `header` option (added on `next`), which already forwards the header to custom `jwt.sign` implementations.

- **`customAccessTokenClaims` could override reserved claims** (Copilot): **Fixed.** Registered/protocol claims (`iss`, `sub`, `aud`, `exp`, `nbf`, `iat`, `jti`, `client_id`, `azp`, `scope`, `typ`, `cnf`) are stripped from the hook's output before minting.

- **`allowedResources` not validated at parse time** (Copilot): **Fixed.** Each entry is validated as an absolute URI without a fragment at config-parse time, so misconfiguration fails at `betterAuth()` init rather than at request time.

- Copilot's comments on `package.json` (prepare script), `packages/cli/package.json` (Node engine), and `docs/app/api/og-blog-vercel/route.tsx` (OG date) were on commits that were never part of this feature — they came from the earlier incorrect base branch and are gone after the rebase onto `next`.

### Tests
`packages/better-auth`: device-authorization 71/71, jwt 135/135, `pnpm --filter better-auth typecheck` clean.

sicarius97:
Follow-up: addressed the two remaining device-authorization findings from the latest review pass (f160fc6a6):

- **jwt-plugin check now runs before `consumeOne`** (Greptile): on a misconfigured server (`allowedResources` set without `jwt()` registered), the token exchange returns `server_error` *without* consuming the approved device code, so the user's approval survives. Covered by a test asserting the code row is still `approved`.
- **Bare trailing `#` rejected** (cubic): `URL.hash` is `""` for a value like `https://api.example.com/#`, so validation now also inspects the raw string for `#` (RFC 8707 §2 / RFC 3986 — an empty fragment is still a fragment). Enforced both at config-parse time and at request time, with tests.

Note: cubic/Copilot inline comments on `.changeset/*`, `package.json`, `docs/app/api/og-blog-vercel`, and `packages/cli/*` are on commits that aren't part of this PR — those bots appear to diff against `main`, but this PR targets `next` (its actual diff is the 9 device-authorization files). `pnpm --filter better-auth typecheck` is clean; device-authorization 71/71, jwt 135/135.

sicarius97:
@Bekacru @gustavovalverde Any changes that need to be considered here? I'm still waiting for a review and an authorized person to run the testing

sicarius97:
@gustavovalverde Did this end up getting baked into a different PR?

PR #10408, opened 2026-07-15
feat(sso)!: resolve provisioned users before authentication

gustavovalverde:
> [!NOTE]
> Stacked on #10406. Review the Identity and provider-account boundary first.

Provisioning systems can create a Better Auth User before that person authenticates. SSO previously had no explicit way to bind a verified provider subject to that existing User, which encouraged email matching, duplicate users, or application-specific account writes outside the authentication transaction.

## What changes

- Adds a transaction-bound `resolveUser` callback for both OIDC and SAML.
- Supports three explicit decisions: use the default flow, link an exact User, or reject authentication.
- Requires linked users to choose whether local profile fields are preserved or updated from verified provider data.
- Separates provider profile attributes from the canonical provider Identity key.
- Keeps custom mapped profile fields available with safe `unknown` defaults and opt-in precise generics.
- Carries the same provider-user contract through OAuth Proxy callback packages.

> [!IMPORTANT]
> This changes the internal OAuth authentication contract from `userInfo` to `providerUser` plus a separate `identity`. OAuth Proxy applications and proxies must run the same Better Auth version. Configuring `resolveUser` requires native interactive transactions.

<details>
<summary>Authentication invariants</summary>

- The resolver runs after provider verification and locking, before Identity, Account, User, or Session writes.
- Resolver database work and the complete authentication graph share one transaction.
- Returned authentication failures roll back resolver writes while preserving existing redirect behavior.
- Provider subjects never become Better Auth User IDs.
- Exact linking never falls back to email and rejects conflicts with an Identity already owned by another User.
- `profile: "preserve"` keeps provisioning-owned fields; `profile: "provider"` applies the verified profile for that login.
- Resolver exceptions and malformed decisions return a stable public error without exposing private details.

</details>

---
## Summary by cubic
Adds a transaction-bound SSO user resolution step for OIDC and SAML so apps can link a verified provider subject to an existing User, use the default flow, or reject before any writes. Separates provider identity from profile and replaces `userInfo` with `providerUser` across core and `oauth-proxy`.

- **New Features**
  - `resolveUser` in `@better-auth/sso` runs after provider verification/locking inside the same transaction and returns: `default`, `link { userId, profile: "preserve" | "provider" }`, or `reject { code, message? }`. The callback receives `providerInstanceId`, `identity`, normalized `providerUser`, raw `providerClaims`, and a transactional `database`.
[...]

PR #10409, opened 2026-07-15
feat(scim): gate authentication on active provisioned users

gustavovalverde:
> [!NOTE]
> This PR is stacked on #10408. Review the SSO user-resolution contract first.

## Why

SCIM can provision a Better Auth User before that person signs in, but authentication still needs an exact, lifecycle-aware bridge from the verified provider identity to that existing User. This change makes SCIM activation state authoritative at sign-in without coupling the SSO plugin to SCIM.

## What changes

- Adds `acquireActiveSCIMUserLink` as a small, typed SCIM composition API.
- Resolves only an exact, case-sensitive `connectionId` and `externalId` pair.
- Rejects inactive, deleted, orphaned, inconsistent, and decommissioned sources.
- Fences sign-in against concurrent lifecycle changes while allowing concurrent valid sign-ins.
- Documents the SCIM and SSO composition with immutable provider-subject guidance.
- Adds a black-box HTTP workflow from SCIM provisioning through OIDC callback and session creation.

<details>
<summary>Lifecycle contract</summary>

- Settled unavailable state returns `null`, so the application can reject authentication explicitly.
- A lifecycle change after acquisition begins aborts the shared transaction instead of authenticating stale state.
- Deleted-resource tombstones preserve reprovisioning identity but never authorize sign-in.
- Reprovisioning the same external subject restores the original Better Auth User.
- Identical external IDs in different SCIM connections remain isolated.

</details>

---
## Summary by cubic
Adds `acquireActiveSCIMUserLink` to `@better-auth/scim` to gate SSO on active, SCIM‑provisioned users only. Sign-in links by exact `connectionId` + `externalId` (case‑sensitive) and rejects inactive, deleted, tombstoned, orphaned sources, and decommissioned or domain‑mismatched connections.

- **New Features**
  - `acquireActiveSCIMUserLink`: transaction‑safe exact lookup that returns `{ userId, scimUserId }` or `null`; no email/`userName` fallback.
  - Lifecycle and concurrency safety: fences deactivation, deletion, and connection decommissioning; concurrent valid sign-ins are allowed.
  - Docs and tests: SCIM docs add a full SSO `resolveUser` example with subject guidance; end‑to‑end HTTP test covers SCIM provisioning through OIDC callback using `@better-auth/sso`.

- **Migration**
  - In SSO `resolveUser`, call `acquireActiveSCIMUserLink({ connectionId, externalId: identity.providerAccountId }, context)` and, if found, return `{ action: "link", userId: link.userId, profile: "preserve" }`; otherwise reject.
  - Configure the directory to send the provider’s immutable, verified subject as the SCIM User `externalId` (case‑sensitive). Run the helper inside the transaction provided by `resolveUser`.

<sup>Written for commit 6b1b57b710dd404fea0fdef9573da79cc0106e17. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10409?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/scim-active-user-link.md
+++ b/.changeset/scim-active-user-link.md
@@ -1,0 +1,7 @@
+---
+"@better-auth/scim": minor
+---
+
+Add `acquireActiveSCIMUserLink` for transaction-safe authentication of provisioned users. The helper maps an exact SCIM connection ID and `externalId` to an active Better Auth User while fencing concurrent deactivation, deletion, and connection decommissioning.
+
+Compose the helper with SSO `resolveUser` to link the provisioned User without matching by email or `userName`.
--- a/docs/content/docs/plugins/scim/index.mdx
+++ b/docs/content/docs/plugins/scim/index.mdx
@@ -200,6 +200,79 @@ Choose the profile behavior when you link a user:
 
 Only one source can manage a Better Auth User's profile. Separate connections may link preserved sources to the same User, but one connection cannot create two SCIM resources for the same linked User.
 
+## Authenticate provisioned users with SSO
+
+Use `acquireActiveSCIMUserLink` inside the SSO plugin's `resolveUser` callback when SCIM must control who can sign in. The helper finds an active SCIM User by its exact connection ID and `externalId`, returning `{ scimUserId, userId }` for the active link. It returns `null` for missing, inactive, or deleted SCIM Users and decommissioned connections.
+
+```ts title="auth.ts"
+import { acquireActiveSCIMUserLink, scim } from "@better-auth/scim";
+import { sso } from "@better-auth/sso";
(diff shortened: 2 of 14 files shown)


gustavovalverde:
Superseded by #10474, which exposes active SCIM-provisioned user links on the schema-preserving stack. Closing this stacked implementation in favor of the merged replacement.

PR #10390, opened 2026-07-14
feat(scim)!: decouple provisioning from the organization plugin

gustavovalverde:
SCIM provisioning was coupled to the organization plugin, so directory-owned resources inherited an application-specific tenancy and role model. This rebuild makes canonical SCIM resources connection-owned and moves user linking, lifecycle, and application access into explicit identity and projection boundaries.

[scim-demo-workflow.webm](https://github.com/user-attachments/assets/498b7aaf-ad89-4202-827c-d65fa5e5280b)

## What changes

- **Connection ownership**: `scim({ connections })` defines bearer-authenticated, isolated resource owners, while provisioning domains let several connections contribute to one application boundary.
- **Identity boundary**: `identity.resolveUser` makes create-or-link decisions explicit, and `identity.reconcileUser` applies combined lifecycle state without guessing by email or creating a sign-in method.
- **Access boundary**: projections translate Group sources into validated, source-aware role grants and reconcile the complete desired state without overwriting access owned elsewhere.
- **Protocol surface**: Users, Groups, discovery, filtering, pagination, response attributes, and ordered atomic PATCH behavior are rebuilt around the new model. No-op PATCH requests now succeed.
- **Transaction semantics**: canonical writes and application callbacks commit atomically; cache and session side effects wait until commit and are discarded on rollback.

> [!IMPORTANT]
> This is a breaking SCIM cutover. The old configuration, client APIs, management surface, and organization-backed schema are removed, and existing provisioning state cannot migrate in place. Deployments must pause provisioning, replace the legacy schema, enable native interactive transactions, configure stable connections, and fully reprovision before resuming traffic. Cloudflare D1 cannot provide the required transaction behavior.
>
> Provisioned users also need a separate sign-in method. The plugin never links existing Better Auth Users by email; retaining them requires an explicit `identity.resolveUser` mapping before reprovisioning.

<details>
<summary>Review map and implementation detail</summary>

- **Ownership and lifecycle**: connection binding, identity resolution and tombstones, multi-source active state, and irreversible connection decommissioning.
- **Projection and authorization**: role allowlisting, source-aware grant ownership, complete-state reconciliation, and replay when mappings change.
- **Protocol interoperability**: public discovery, SCIM error and media envelopes, equality filters, response attributes, and Microsoft Entra ID and Okta membership PATCH shapes.
- **End-to-end workflow**: shared adapter-backed HTTP coverage and the Next.js directory workspace exercise the same provisioning path through real requests, storage, authorization checks, and UI state.
[...]

PR #10420, opened 2026-07-15
fix(scim): support externalId filter on GET /scim/v2/Users

### LINKED ISSUE #10419: Bug: `@better-auth/scim` `externalId` filter on `GET /scim/v2/Users` returns 400

### Reproduction

Steps to reproduce - issue a filter query using `externalId`, the standard pre-provision existence check all major SCIM providers perform:

```bash
curl -H "Authorization: Bearer <scim-token>" \
  "https://your-app.example.com/auth/scim/v2/Users?filter=externalId+eq+%22user@example.com%22"
```

```json
{
  "schemas": ["urn:ietf:params:scim:api:messages:2.0:Error"],
  "status": "400",
  "detail": "The attribute \"externalId\" is not supported"
}
```

**Current:** `GET /scim/v2/Users?filter=externalId eq "..."` returns HTTP 400 with `"The attribute \"externalId\" is not supported"`. This is because `externalId` is missing from the `SCIMUserAttributes` map in `packages/scim/src/scim-filters.ts`, and `parseSCIMUserFilter` throws for any attribute not found in that map.

**Expected:** `GET /scim/v2/Users?filter=externalId eq "<value>"` should return a `ListResponse` filtered to users whose linked account has `accountId` matching the given value (or an empty result set if none match). All major SCIM providers use this query as the first step before `POST` to check for existing accounts.

**Real-world consequence:** When a SCIM provider (Rippling, Okta, Azure AD) attempts to provision a user, it first checks `?filter=externalId eq "..."`. Receiving 400 causes the provider to either abort the provisioning flow with an error or fall back to unconditional POST, potentially creating duplicate accounts on re-runs.

## Root cause

`externalId` missing from `SCIMUserAttributes` map, in `packages/scim/src/scim-filters.ts`:

```ts
const SCIMUserAttributes: Record<string, string | undefined> = {
  userName: "email",
  // externalId is not here
};
```

`parseSCIMUserFilter` does:

```ts
const targetAttribute = SCIMUserAttributes[attribute];
const resourceAttribute = SCIMUserResourceSchema.attributes.find(
  (attr) => attr.name === attribute,
);

if (!targetAttribute || !resourceAttribute) {
  throw new SCIMParseError(`The attribute "${attribute}" is not supported`);
}
```

Any filter on an attribute not in `SCIMUserAttributes` throws immediately. Unlike the other filterable attributes, `externalId` lives on the `account` model (as `accountId`), not on the `user` model that `listSCIMUsers` filters directly - so it can't be fixed by simply adding it to the map unchanged; it needs to be resolved against the account records before the user query runs.

`1.6.11` (with `@better-auth/scim@1.6.19`)

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.5.0"
  },
  "node": {
    "version": "v24.6.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.5.1"
  },
  "frameworks": [
    { "name": "hono", "version": "^4.12.25" }
  ],
  "databases": [
    { "name": "drizzle", "version": "^0.45.2" }
  ],
  "betterAuth": {
    "version": "^1.6.11"
  }
}
```

Backend, Package

```typescript
import { scim } from "@better-auth/scim";

  plugins: [
    scim({}),
  ],
});
```

### Additional context

**RFC 7644 §3.4.2 (Querying Resources):**
> "Each expression MUST contain an attribute name followed by an attribute operator and optional value." - there is no restriction prohibiting `externalId` as a filter attribute. All major provider integration guides (Okta SCIM, Azure AD SCIM, Rippling) document `externalId` filtering as the mandatory pre-provision existence check.

**RFC 7643 §3 (`externalId`):**
> "A String that is an identifier for the resource as defined by the provisioning client ... The service provider MUST always interpret the externalId as scoped to the provisioning domain ... a mutability of 'readWrite'."

**Suggested fix:**

`externalId` lives on the `account` model (as `accountId`) rather than directly on the `user` model. The cleanest fix is a two-step lookup in `listSCIMUsers`:

1. When `filter` starts with `externalId`, resolve it against the `account` records already fetched for this provider (`accounts.filter(a => a.accountId === value)`) to get a set of matching `userId`s.
2. Narrow the subsequent `user` `findMany` call to those `userId`s (or return an empty `ListResponse` if the account filter yields no matches).

Note: `POST /scim/v2/Users` with `active: false` on initial provisioning is already handled correctly as of [PR #10242](https://github.com/better-auth/better-auth/pull/10242) - that part of this report is out of scope; this issue is limited to the `externalId` filter.

I tested my reproduction against the latest release (`1.6.11`, main branch commit `502a262`).

dipakparmar:
Fixes #10419

GET /scim/v2/Users?filter=externalId eq "..." always returned 400 because externalId is missing from the SCIMUserAttributes map used to resolve list filters onto database fields. Every major SCIM provider (Okta, Azure AD, Rippling) issues this filter to check whether a user already exists before provisioning; receiving 400 causes providers to abort provisioning or fall back to unconditional POST, risking duplicate accounts on retries.

externalId lives on the account model (as accountId), not on the user model that the other list filters target, so it can't go through the generic SCIMUserAttributes map unchanged. It's intercepted before the generic filter parsing and resolved in-memory against the accounts array listSCIMUsers already fetches for this provider.

Non-breaking - this is a previously-broken code path (always 400), so no existing correct integration can rely on the old behavior. Includes tests for a match, a no-match empty result, and case-sensitivity (externalId is caseExact, unlike userName).

---
## Summary by cubic
Adds support for externalId filtering on GET /scim/v2/Users so providers can run pre-provision checks without 400s. Handles case-insensitive attribute names and returns the correct account when users have multiple SCIM accounts.

- **Bug Fixes**
  - Intercepts externalId (case-insensitive) and matches against `accountId` to scope users before other filters.
  - Exposes `parseSCIMFilter` and updates parsing to return `{ externalId, apiFilters }` for `listSCIMUsers`.
  - Uses the scoped account when building Resources so the matched externalId is rendered for multi-account users.
  - Adds tests for a match, no match, caseExact value behavior, attribute name case-insensitivity, and multi-account rendering.

<sup>Written for commit ace79efc13356ddec6782c0a3db4b9a98631c6c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10420?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/fix-scim-externalid-case-insensitive.md
+++ b/.changeset/fix-scim-externalid-case-insensitive.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/scim": patch
+---
+
+Match the `externalId` filter attribute name case-insensitively (e.g. `ExternalId eq "..."`, `EXTERNALID eq "..."`), per RFC 7644 §3.4.2.2 which specifies that SCIM attribute names are case-insensitive. Previously only the exact-case `externalId` was recognized before falling through to the generic filter path and returning `400`.
--- a/.changeset/fix-scim-externalid-multi-account.md
+++ b/.changeset/fix-scim-externalid-multi-account.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/scim": patch
+---
+
+Fix `GET /scim/v2/Users?filter=externalId eq "..."` returning the wrong account's `externalId` when a user has more than one SCIM account for the same provider (possible via `linkExistingUsers`). The user-ID scoping was already correct; the resource-building step looked up the account from the unfiltered account list instead of the `externalId`-matched one.
--- a/.changeset/fix-scim-user-lifecycle.md
+++ b/.changeset/fix-scim-user-lifecycle.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/scim": patch
+---
+
+Support `GET /scim/v2/Users?filter=externalId eq "..."`, the standard pre-provision existence check all major SCIM providers (Okta, Azure AD, Rippling) perform before creating a user. Previously this returned `400 The attribute "externalId" is not supported` because `externalId` lives on the `account` model (as `accountId`) rather than on the `user` model filtered by the other list filters.
--- a/packages/scim/src/routes.ts
+++ b/packages/scim/src/routes.ts
@@ -20,7 +20,11 @@ import type { AuthMiddleware } from "./middlewares";
 import { buildUserPatch } from "./patch-operations";
 import { SCIMAPIError, SCIMErrorOpenAPISchemas } from "./scim-error";
 import type { DBFilter } from "./scim-filters";
-import { parseSCIMUserFilter, SCIMParseError } from "./scim-filters";
+import {
+	parseSCIMFilter,
+	parseSCIMUserFilter,
+	SCIMParseError,
+} from "./scim-filters";
 import {
 	ResourceTypeOpenAPISchema,
 	SCIMSchemaOpenAPISchema,
@@ -1043,7 +1047,9 @@ export const listSCIMUsers = (authMiddleware: AuthMiddleware) =>
 				Resources: [],
 			} as const;
 
-			const apiFilters: DBFilter[] = parseSCIMAPIUserFilter(ctx.query?.filter);
+			const { externalId, apiFilters } = parseSCIMAPIUserFilter(
+				ctx.query?.filter,
+			);
 
 			const providerId = ctx.context.scimProvider.providerId;
 			const accounts = await ctx.context.adapter.findMany<Account>({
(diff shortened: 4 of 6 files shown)


gustavovalverde:
Superseded by #10390. externalId filtering is implemented against a connection-scoped canonical SCIM key rather than the legacy Account lookup.

PR #10422, opened 2026-07-15
fix(scim)!: reject PATCH remove ops and return 200 with ?attributes

### LINKED ISSUE #10421: Bug: `@better-auth/scim` PATCH silently skips `remove` ops and always returns 204

### Reproduction

Steps to reproduce (`remove` op silently ignored):

```bash
# remove op - silently ignored, returns 204 as if it succeeded
curl -X PATCH -H "Authorization: Bearer <scim-token>" \
  -H "Content-Type: application/scim+json" \
  -d '{
    "schemas": ["urn:ietf:params:scim:api:messages:2.0:PatchOp"],
    "Operations": [
      { "op": "remove", "path": "name.formatted" }
    ]
  }' \
  https://your-app.example.com/auth/scim/v2/Users/<id>
```

Response: `204 No Content` - the operation appears to have succeeded but nothing was changed.

Steps to reproduce (`attributes` query param ignored):

```bash
# attributes param - always 204, never 200 with body
curl -X PATCH -H "Authorization: Bearer <scim-token>" \
  -H "Content-Type: application/scim+json" \
  -d '{
    "schemas": ["urn:ietf:params:scim:api:messages:2.0:PatchOp"],
    "Operations": [
      { "op": "replace", "path": "active", "value": false }
    ]
  }' \
  "https://your-app.example.com/auth/scim/v2/Users/<id>?attributes=active,userName"
```

Response: `204 No Content` - should be `200 OK` with the resource body.

**Current:**
- A PATCH body containing `"op": "remove"` is silently skipped. `buildUserPatch` in `packages/scim/src/patch-operations.ts` iterates operations and executes `continue` for any op that is not `"add"` or `"replace"`. The handler then returns `204 No Content` as if all operations were applied successfully.
- A PATCH request with `?attributes=active,userName` always returns `204 No Content`. The `attributes` query parameter is not declared in the PATCH endpoint's query schema in `packages/scim/src/routes.ts`, so it is never read. The handler unconditionally returns 204 regardless of whether the client requested a response body.

**Expected:**
- A `remove` operation with `path` unspecified (a path-less remove) must fail with `400 Bad Request` and `scimType: "noTarget"` per RFC 7644 §3.5.2.2.
- A `remove` operation with a valid `path` should either be applied or return `400 Bad Request` with a descriptive `detail` if the specific attribute removal is not supported. It must not silently return 204.
- A PATCH request with the `attributes` query parameter present must return `200 OK` with the full or projected resource body per RFC 7644 §3.5.2.

**Real-world consequence:** A SCIM provider that sends a path-less `remove` to clear all values for a multi-valued attribute receives 204 and believes the operation succeeded. The attribute is unchanged. The provider's next reconciliation cycle will either re-send the same remove (looping forever) or consider its state authoritative and stop trying, leaving stale data. A provider that uses `PATCH ?attributes=...` to get an efficient confirmation of the updated fields will always receive 204 with no body, forcing it to issue a follow-up GET to confirm the change - doubling the round trips for every PATCH.

## Root cause

**1. `remove` op skipped via `continue` in `buildUserPatch`**

In `packages/scim/src/patch-operations.ts`:

```ts
for (const operation of operations) {
  if (operation.op !== "add" && operation.op !== "replace") {
    continue;   // <-- silently skips "remove", no error, no effect
  }
  // ... handle add/replace ...
}
```

After the loop, the PATCH handler returns 204 unconditionally. There is no check that all operations were handled - a request composed entirely of `remove` ops returns 204 with zero changes applied.

**2. PATCH endpoint returns 204 unconditionally**

In `packages/scim/src/routes.ts`, `patchSCIMUser`:

```ts
ctx.setStatus(204);
return;
```

There is no branch that checks for an `attributes` query parameter and returns 200 with a body instead.

**3. `attributes` not declared in the PATCH query schema**

The PATCH endpoint has no `query` schema at all, so `ctx.query?.attributes` is always `undefined` regardless of what the client sends.

`1.6.11` (with `@better-auth/scim@1.6.19`)

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.5.0"
  },
  "node": {
    "version": "v24.6.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.5.1"
  },
  "frameworks": [
    { "name": "hono", "version": "^4.12.25" }
  ],
  "databases": [
    { "name": "drizzle", "version": "^0.45.2" }
  ],
  "betterAuth": {
    "version": "^1.6.11"
  }
}
```

Backend, Package

```typescript
import { scim } from "@better-auth/scim";

  plugins: [
    scim({}),
  ],
});
```

### Additional context

**RFC 7644 §3.5.2 (Modifying with PATCH - general):**
> "A PATCH request ... SHALL be treated as atomic. If any one of the values returns a failure condition, the operation SHALL fail."

**RFC 7644 §3.5.2 (response status):**
> "The server MUST return a 200 OK if the 'attributes' parameter is specified in the request."

**RFC 7644 §3.5.2.2 (remove operation):**
> "If 'path' is unspecified, the operation fails with HTTP status code 400 and a 'scimType' error code of 'noTarget'."

**Suggested fix - `remove` op:**

In `buildUserPatch`, replace the silent `continue` with explicit handling:

```ts
for (const operation of operations) {
  if (operation.op === "remove") {
    if (!operation.path) {
      throw new SCIMAPIError("BAD_REQUEST", {
        detail: "remove operation requires a path",
        scimType: "noTarget",
      });
    }
    throw new SCIMAPIError("BAD_REQUEST", {
      detail: `remove is not supported for path "${operation.path}"`,
      scimType: "noTarget",
    });
  }
  // ... existing add/replace handling ...
}
```

Throwing rather than continuing preserves RFC 7644's atomicity requirement - the PATCH either fully succeeds or fails, it never partially applies. Note: this is a breaking change for any client currently sending `remove` ops that received a (spurious) 204.

**Suggested fix - `attributes` query param and 200 response:**

1. Add `attributes` to the PATCH endpoint's query schema.
2. After applying the PATCH operations, check for the param and return the appropriate status - 200 with the resource body when present, 204 otherwise.

I tested my reproduction against the latest release (`1.6.11`, main branch commit `502a262`).

dipakparmar:
Fixes #10421

BREAKING (partial): buildUserPatch silently skipped remove operations via continue, and the endpoint always returned 204, so a client sending a path-less remove believed it succeeded when nothing changed. RFC 7644 §3.5.2.2 requires 400 noTarget when path is unspecified, and §3.5.2 requires the whole PATCH to fail atomically. A remove with a path now also returns 400, since attribute removal isn't implemented for any path - silently no-op'ing is worse than an explicit error a client can act on.

Non-breaking: the attributes query param was absent from the PATCH query schema, so ?attributes=... was silently dropped and the endpoint always returned 204 even though RFC 7644 §3.5.2 requires 200 with the resource body when attributes is specified.

Deployment note for reviewers/adopters: any client currently sending remove ops and getting a (spurious) 204 will start receiving 400 after this lands. Recommend auditing IdP outbound SCIM logs for PATCH requests containing remove before upgrading.

No existing test previously exercised remove-op or ?attributes behavior, so nothing else changes. New tests cover: path-less remove -> 400 noTarget, remove-with-path -> 400, and ?attributes -> 200 with the updated resource body.

---
## Summary by cubic
Make SCIM PATCH compliant with RFC 7644 by rejecting remove operations and returning 200 with a response body when the `attributes` query is used. Stops silent no-ops and aligns client expectations.

- **Bug Fixes**
  - `PATCH /scim/v2/Users/:id`: reject `remove` ops. Path-less → `400` with `scimType: "noTarget"`; with `path` → `400` without `scimType` (removal not implemented). PATCH is atomic; the whole request fails.
  - Return `200 OK` with the updated resource when `?attributes=...` is provided; previously always returned `204` and ignored the param.

- **Migration**
  - Audit clients/IdPs sending SCIM PATCH `remove` ops; these now receive `400` instead of a misleading `204`. Path-bearing `remove` errors no longer include `scimType: "noTarget"`.

<sup>Written for commit 4d9aa0ace75678515e2715202a1f462570597130. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10422?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ping-maxwell (changes requested):
hey there shouldn't be all these commits 😅
please remove them

dipakparmar:
Closing. Looks like #10390 already addresses both of these fixes, and more thoroughly:

- The path-less `remove` -> `400 noTarget` behavior is now in `user-patch.ts`, with proper per-attribute remove semantics (e.g. `userName` explicitly rejects removal) rather than this PR's blanket "not supported" for any path.
- The `?attributes` -> `200` behavior is now in `user-provisioning.ts`'s `patchSCIMUser`, via `requestsProjectedMutationResponse`/`projectSCIMResourceAttributes`, which actually projects down to the requested attributes, closing the gap this PR's own reviewers flagged (we always returned the full resource).

No need to carry this forward separately. Thanks for the reviews!

PR #10423, opened 2026-07-15
fix(two-factor): reject reused TOTP codes on step-up verification

### LINKED ISSUE #10387: fix(two-factor): TOTP codes can be reused after successful validation (RFC 6238 §5.2 requires one-time use)
### Summary

A TOTP code that has already been successfully validated is still accepted on subsequent attempts. RFC 6238 §5.2 requires the opposite:

> Note that a prover may send the same OTP inside a given time-step window multiple times to a verifier. **The verifier MUST NOT accept the second attempt of the OTP after the successful validation has been issued for the first OTP**, which ensures one-time only use of an OTP.

OWASP ASVS 5.0 §6.5.1 (L2) states the same requirement ("time-based one-time passwords (TOTPs) are only successfully usable once").

Verified empirically on `better-auth@1.7.0-rc.1` (memory adapter, `twoFactor()` plugin).

### Reproduction

```ts
const auth = betterAuth({
  database: memoryAdapter({ user: [], session: [], account: [], verification: [], twoFactor: [], rateLimit: [] }),
  secret: "…",
  plugins: [twoFactor()],
});

// sign up, enableTwoFactor(), take the secret from totpURI (base32-decoded), complete enrollment.
// Then, with the (new) authenticated session:
const code = await createOTP(secret, { period: 30, digits: 6 }).totp();
await auth.api.verifyTOTP({ body: { code }, headers, asResponse: true }); // 200
await auth.api.verifyTOTP({ body: { code }, headers, asResponse: true }); // 200  <-- same code
await auth.api.verifyTOTP({ body: { code }, headers, asResponse: true }); // 200  <-- same code

// And across two separate sign-in challenges:
const a = await auth.api.signInEmail({ body: { email, password }, asResponse: true });
await auth.api.verifyTOTP({ body: { code: X }, headers: cookiesOf(a), asResponse: true }); // 200
const b = await auth.api.signInEmail({ body: { email, password }, asResponse: true });
await auth.api.verifyTOTP({ body: { code: X }, headers: cookiesOf(b), asResponse: true }); // 200 <-- same code X reused
```

Observed:

- **Step-up path (active session)**: the same code is accepted 3/3 times (200, 200, 200).
- **Sign-in path**: a code accepted for challenge A is accepted again for a *new* challenge B (200).

### Root cause

- The `twoFactor` schema stores only `secret`, `backupCodes`, `userId`, `verified`, `failedVerificationCount`, `lockedUntil` — there is **no record of consumed OTP values / time steps**, so replay cannot be detected.
- Verification is a stateless computation (`createOTP(secret, { period, digits }).verify(code)`), and the default acceptance window is ±1 time step, so a single code stays valid for roughly 90 seconds.
- The per-sign-in 2FA challenge (verification row + cookie) *is* consumed on success, which prevents replaying a code **against the same challenge** — but a new challenge (a fresh password sign-in) accepts the same code again, and the step-up path (`verifyTOTP` with an active session) has no challenge to consume at all.

### Impact

A TOTP code that becomes known to an attacker **after** the legitimate user already used it (shoulder-surfing, a phishing relay, screen capture, a log, malware) remains usable for the remainder of its acceptance window (~90 s). RFC 6238 mandates that this be impossible, and applications that must claim ASVS L2 conformance cannot do so today.

### Proposal

Persist the last successfully-consumed time step per `twoFactor` row (e.g. `lastUsedStep: number`) and reject a verification whose candidate step is `<= lastUsedStep`. This enforces one-time use (and monotonicity, which also rejects delayed replays of older steps within the window) with a single additional column and no change to the public API.

Happy to open the PR.

Environment: `better-auth@1.7.0-rc.1` (`two-factor` plugin), verified with the memory adapter.

gaurav0107:
## Summary

TOTP codes could be verified more than once on the **step-up (re-verification)
path** — i.e. calling `verifyTOTP` while a session is already active (for
sensitive-action re-authentication). Because the `twoFactor` row recorded no
consumed time step, and the step-up path has no per-challenge verification row
to consume, the *same* code was accepted again and again.

This violates the one-time-use requirement of **RFC 6238 §5.2** ("The verifier
MUST NOT accept the second attempt of the OTP after the successful validation
has been issued for the first OTP") and **OWASP ASVS 5.0 §6.5.1**.

Closes #10387

## Changes

- **schema**: add a nullable `lastUsedStep` field to the `twoFactor` model,
  following the existing `failedVerificationCount` / `lockedUntil` non-breaking
  pattern (`required: false`, `input: false`, `returned: false`).
- **`totp/verifyTOTP`**: walk the same ±1 acceptance window the OTP util's
  `verify()` uses, but capture *which* time step matched. On the step-up path
  (`!isSignIn`), reject a code whose matched step is `<= lastUsedStep`, and
  record the consumed step after a successful verification.
- **types**: add `lastUsedStep?: number | null` to `TwoFactorTable`.
- **changeset**: patch.

Legacy-safe: `lastUsedStep` is `null` on pre-migration rows (and document stores
predating the column), so the first verification always succeeds. The monotonic
`<=` also rejects delayed replays of an older step still inside the window.

## Scope: step-up path only (by design)

The fix is intentionally scoped to the **step-up** path, which is the
unambiguous vector the issue leads with (the reported repro accepts the same
code 3/3 times on an active session) and which has **no** existing protection.

The issue also notes a **sign-in** vector (a code used in challenge A being
accepted in a new challenge B). I did **not** change the sign-in path here,
because:

- The sign-in path already consumes its per-challenge verification row
  atomically on success, which blocks replay against that challenge.
- Extending strict single-use across *separate* sign-in challenges changes
  established behavior that the repo's own recently-added 2FA security tests
  rely on — e.g. `two-factor.account-lockout.test.ts` reuses a correct code
  across multiple challenges, and one test performs two back-to-back successful
  sign-in verifications with the same code. That is a broader contract decision
  I'd rather leave to maintainers than fold into this fix.

Happy to extend enforcement to the sign-in path (and update those tests) in a
follow-up if you'd like it covered here.

## Testing

Added two regression tests to `two-factor.security.test.ts` (each in its own
instance so `lastUsedStep` cannot leak between them), both proven failing-first
on the current code and passing with the fix:

- a reused step-up code is rejected on the second attempt;
- a step-up code succeeds exactly once across three repeated attempts (the
  reported 3/3 repro).

```
pnpm exec vitest run src/plugins/two-factor/   # 85 passed (4 files)
pnpm run typecheck                             # clean
pnpm exec biome check <changed files>          # clean
```

All existing two-factor tests continue to pass unchanged.

---
## Summary by cubic
Prevents TOTP code reuse during step-up (re-verification). We now record the matched time step and reject any code at or below the last used step, enforcing single-use.

- **Bug Fixes**
  - Schema: add nullable lastUsedStep to twoFactor (not exposed in input/response).
  - verifyTOTP: capture the matched step across the ±1 window, reject replays on step-up, and persist lastUsedStep after success. Sign-in flow is unchanged.
  - Types: add lastUsedStep to TwoFactorTable.
  - Tests: add regression tests ensuring a step-up code succeeds once and replays are rejected; backward-compatible for existing rows.

<sup>Written for commit fb11d55578dd94854163cd2d3aea2070a6c8fc14. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10423?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/better-auth/src/plugins/two-factor/schema.ts
+++ b/packages/better-auth/src/plugins/two-factor/schema.ts
@@ -55,6 +55,18 @@ export const schema = {
 				input: false,
 				returned: false,
 			},
+			// The most recent TOTP time-step consumed on the step-up
+			// (re-verification) path. Step-up verification rejects any candidate
+			// step <= this value, enforcing RFC 6238 §5.2 one-time use where the
+			// active session leaves no per-challenge row to consume. Nullable so
+			// pre-migration rows (and document stores predating the column) are
+			// treated as "no step consumed yet" and the first use always succeeds.
+			lastUsedStep: {
+				type: "number",
+				required: false,
+				input: false,
+				returned: false,
+			},
 		},
 	},
 } satisfies BetterAuthPluginDBSchema;
--- a/.changeset/fix-totp-single-use-replay.md
+++ b/.changeset/fix-totp-single-use-replay.md
@@ -1,0 +1,15 @@
+---
+"better-auth": patch
+---
+
+fix(two-factor): reject reused TOTP codes on the step-up verification path
+
+When `verifyTOTP` ran with an already-active session (step-up / re-verification),
+a code that had already been successfully validated was accepted again and again,
+because the `twoFactor` row recorded no consumed time step and — unlike the
+sign-in path — there is no per-challenge verification row to consume. Step-up
+verification now records the last consumed time step (`lastUsedStep`) and rejects
+any candidate step at or below it, enforcing the one-time use required by
+RFC 6238 §5.2 and OWASP ASVS 5.0 §6.5.1. The new field is nullable, so existing
+rows are treated as having no consumed step and the first verification always
+succeeds.
--- a/packages/better-auth/src/plugins/two-factor/totp/index.ts
+++ b/packages/better-auth/src/plugins/two-factor/totp/index.ts
@@ -297,19 +297,56 @@ export const totp2fa = (options?: TOTPOptions | undefined) => {
 				? await beginAttempt(DEFAULT_TWO_FACTOR_ALLOWED_ATTEMPTS)
 				: null;
 			let status: boolean;
+			let matchedStep: number | null = null;
 			try {
 				const decrypted = await symmetricDecrypt({
 					key: ctx.context.secretConfig,
 					data: twoFactor.secret,
 				});
-				status = await createOTP(decrypted, {
+				const otp = createOTP(decrypted, {
 					period: opts.period,
 					digits: opts.digits,
-				}).verify(ctx.body.code);
+				});
+				// Walk the same ±1 acceptance window the util's verify() uses, but
+				// capture WHICH time step matched so one-time use can be enforced.
+				// currentStep is read once so a Date.now() tick across a period
+				// boundary cannot shift the window between deciding validity and
+				// recording the consumed step.
+				const window = 1;
+				const currentStep = Math.floor(Date.now() / (opts.period * 1000));
+				for (let i = -window; i <= window; i++) {
+					const candidateStep = currentStep + i;
+					const candidate = await otp.hotp(candidateStep);
+					if (candidate === ctx.body.code) {
+						matchedStep = candidateStep;
+						break;
+					}
+				}
+				status = matchedStep !== null;
 			} catch (error) {
 				// A server error before the code is checked must not spend the slot.
 				await attempt?.restore();
 				throw error;
+			}
+			// Enforce RFC 6238 §5.2 one-time use on the step-up (re-verification)
+			// path. Here the session is already active, so there is no per-challenge
+			// verification row to consume — nothing stops the exact same code from
+			// being accepted again and again. Reject a code whose step was already
+			// consumed, or any earlier step still inside the acceptance window. The
+			// sign-in path is intentionally left to its existing single-use guard:
+			// its 2FA challenge row is consumed atomically on success, which already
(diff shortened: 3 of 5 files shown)


packages/better-auth/src/plugins/two-factor/schema.ts
```diff
@@ -55,6 +55,18 @@ export const schema = {
...
+			// step <= this value, enforcing RFC 6238 §5.2 one-time use where the
+			// active session leaves no per-challenge row to consume. Nullable so
+			// pre-migration rows (and document stores predating the column) are
+			// treated as "no step consumed yet" and the first use always succeeds.
+			lastUsedStep: {
+				type: "number",
+				required: false,
+				input: false,
+				returned: false,
+			},
```
bytaesu, line 69: This PR introduces a schema change. Since we're trying to avoid schema changes in the v1.x line unless clearly necessary, we should discuss and align on the direction first.

I'll close this PR for now.

Feel free to continue the discussion in the linked issue if you think this is needed.

bytaesu: And if it's necessary, we should target this implementation for the next minor release on the `next` branch.

PR #10439, opened 2026-07-17
feat(social): add Threads OAuth provider

### LINKED ISSUE #10307: feat: Threads OAuth provider

Better Auth ships 35+ built-in social providers (including Twitter/X, TikTok, and Facebook) but has no first-class provider for Threads (Meta's Threads platform, threads.net / graph.threads.net). To add Threads sign-in today you have to drop down to the `genericOAuth` plugin and hand-roll the whole flow: the authorization/token endpoints, the `threads_basic` scope, a fully custom `getUserInfo` against the Threads Graph API, and long-lived-token exchange/refresh (Threads issues short-lived tokens that must be swapped for long-lived ones). That's a meaningful amount of boilerplate for a mainstream provider, it's easy to get the token-exchange step wrong, and it's inconsistent with how sibling Meta/social platforms (Facebook, TikTok, X) are supported out of the box.

Add a native threads social provider under `packages/core/src/social-providers/threads.ts`, modeled on the existing `twitter.ts` / `tiktok.ts` providers, exposing `createAuthorizationURL`, `validateAuthorizationCode`, `refreshAccessToken`, and `getUserInfo`, and registered in the provider index so it's usable via:

```ts
    socialProviders: {
      threads: { clientId: "...", clientSecret: "..." }
    }
```

Endpoints/details it would encapsulate:

- Authorize: https://threads.net/oauth/authorize
- Token: https://graph.threads.net/oauth/access_token
- User info: https://graph.threads.net/v1.0/me
- Default scope: `threads_basic`
- Built-in short-lived to long-lived token exchange + refresh handling

- The `genericOAuth` plugin, which is what I'm using in production today. It works, but requires a custom `getUserInfo` and manual token-exchange logic that a built-in provider should own. (Related pain around the generic path is captured in #9040.)
- Keeping it as a userland helper/snippet, but a native provider matches how comparable platforms are already supported and benefits everyone.

### Additional context

- No existing `threads.ts` provider and no open issue/PR for it I could see.
- I already have a working production implementation on top of `genericOAuth` and am happy to open the PR (fork, feature/threads-provider, Vitest tests, docs) once a maintainer signals they'd accept it.
- One design question for maintainers: Threads' `threads_basic` scope does not return an email, so the provider's user mapping would have a null/absent email. Want to confirm the preferred handling (leave email null vs. require a linking flow) before implementation.

coopbri:
## What

Adds [Threads](https://threads.com) (Meta) as a first-class social provider, so it can be configured like any other:

```ts
  socialProviders: {
    threads: {
      clientId: process.env.THREADS_CLIENT_ID!,
      clientSecret: process.env.THREADS_CLIENT_SECRET!,
    },
  },
});
```

Closes #10307.

Today the only way to use Threads is to hand-roll it on top of `genericOAuth`: the split authorize/Graph hosts, the `threads_basic` scope, a custom `getUserInfo`, and (the easy-to-get-wrong part) the short-lived to long-lived token exchange. This encapsulates all of that behind the standard provider config, matching how the other Meta/social platforms (Facebook, TikTok, Twitter) are already supported.

## Used in production

This provider is a direct distillation of Threads sign-in already running in production on two products, https://topcrush.app and https://thrivestream.live, both currently on the `genericOAuth` path this replaces. The design choices below (id-keyed email, the inline long-lived exchange, self-refresh) are the ones those integrations converged on after hitting the sharp edges in production, not guesses.

## The two Threads-specific things

**1. Token model.** Threads does not do vanilla OAuth refresh. The code exchange returns a short-lived (~1h) token with no `expires_in` and no `refresh_token`. That token must be swapped for a long-lived (60-day) token, which later refreshes itself by presenting itself to `th_refresh_access_token`. The provider:
- exchanges short to long-lived inline inside `validateAuthorizationCode`, recording a real `accessTokenExpiresAt` (the short-lived response carries none, so Better Auth would otherwise store no expiry at all), and
- stores the long-lived token as the `refreshToken` (Threads has no separate refresh credential; the access token is its own), so Better Auth's existing refresh machinery drives the self-refresh via `refreshAccessToken`.

**2. Email.** `threads_basic` returns no email. The provider synthesizes `` `${profile.id}@threads.invalid` ``:
- keyed on the **immutable Threads user id**, never the handle. A handle-keyed placeholder (what a naive mapping, e.g. `<username>@threads.net`, produces) goes stale the moment the user renames, and two accounts can then mint the same address, colliding on the unique `email` constraint and breaking sign-in. This was a real production lesson.
- `.invalid` is the reserved TLD (RFC 2606), so the address can never route real mail or be registered.

`mapProfileToUser` is supported for callers who want to supply a real email or store the Threads id on their user, and `getUserInfo` can be overridden entirely.

The issue explicitly asked for maintainer input on email handling (null vs. placeholder). I went with a stable, id-keyed placeholder so sign-in works out of the box (matching TikTok's "just works" fallback, but without the rename/collision hazard). Happy to switch to `null` + linking if that is the preferred direction.

## Changes

- `packages/core/src/social-providers/threads.ts` - the provider
- `packages/core/src/social-providers/index.ts` - registration
- `packages/better-auth/src/social.test.ts` - 7 tests (config, authorize URL + comma-joined scopes, full sign-in with long-lived upgrade + id-keyed email, name fallback, additional scopes, `mapProfileToUser`, self-refresh)
- `docs/content/docs/authentication/threads.mdx` - docs page
- `docs/components/sidebar-content.tsx`, `docs/components/builder/social-provider.tsx` - nav + builder entries

## Tests

`pnpm vitest run src/social.test.ts` - 30 passed (7 new), Biome clean.

---
## Summary by cubic
Adds Threads as a built-in social provider in `better-auth`, so it can be configured via `socialProviders.threads` like any other provider. The provider handles Threads' short→long‑lived token exchange, stores the 60-day token as the refresh credential for self-refresh, and synthesizes a stable id-keyed placeholder email since `threads_basic` returns none.

**New Features**
- Configure via `socialProviders.threads: { clientId, clientSecret }`.
- Upgrades short-lived tokens to 60-day tokens, stores them as the refresh credential, and refreshes via `th_refresh_access_token`.
- Default scope `threads_basic`; supports extra scopes, `mapProfileToUser`, and overriding `getUserInfo`.
- Stable synthetic email `${profile.id}@threads.invalid` avoids handle-rename collisions.
- Docs at `/docs/authentication/threads` with sidebar entry; tests included.

<sup>Written for commit 57eaee2a5a0a4837e22fb1c5951dda92da39aed6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10439?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bytaesu (commented):
Hi @coopbri,

We've changed how we structure our docs app a bit. Could you switch the base branch to `main` and resolve the conflicts?

bytaesu:
+) I've recently reviewed many provider PRs and found that their types often don't match the actual API contract. I'll also verify this on my side, but could you double-check it as well?

coopbri:
> This PR was automatically retargeted from `main` to `next` because it contains a **minor** changeset. The `main` branch only accepts `patch` (bug fix) changes. Features and breaking changes go through `next` for beta testing before promotion to stable.

@bytaesu thanks for the review. Could you please help me with this? The bot seems to keep overwriting my retarget to main back to next. Should I just remove the changeset for now?

PR #10467, opened 2026-07-21
perf(next-js): reuse the next/headers import promise in production

### LINKED ISSUE #10466: perf: nextCookies re-imports next/headers on every request, expensive under ESM loader hooks
### Summary

```js
const counts = new Map();
});
```

### Suggested fix

```js
let nextHeadersPromise;
```

jlucaso1:
Closes #10466

`nextCookies()` called `await import("next/headers.js")` on every `/get-session` request and on every response that sets cookies. Each `import()` call re-runs ESM resolution, and with loader hooks registered (Sentry, OpenTelemetry) every resolution is a synchronous round trip to the loader hooks thread, which shows up as CPU on the hot path.

This caches the import promise at module scope, gated to production so dev servers and test mocks that reload modules keep the per-call behavior.

Verified against the built package with a `node:diagnostics_channel` subscriber counting `next/headers` imports across 5 hook invocations:

| | before | after |
|---|---|---|
| `NODE_ENV=production` | 5 | 1 |
| `NODE_ENV=development` | 5 | 5 |

Existing `next-js.test.ts` suite passes (6/6) and `pnpm typecheck` is clean.

---
## Summary by cubic
Cache the dynamic import of `next/headers.js` for `nextCookies()` across all environments to avoid repeated ESM resolution and cut CPU on `/get-session` and responses that set cookies.

- **Performance**
  - Cache the `import("next/headers.js")` promise at module scope in all envs; on error the cache resets so later calls retry.
  - Only ESM resolution is cached (not `headers()`/`cookies()`); tests verify reuse between requests and retry after a failed import.

<sup>Written for commit dec7312188f0e950e1d0750df05885f7ca2d4a5c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10467?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bytaesu:
Hi @jlucaso1,

I removed the environment-specific branching and slightly refactored the tests.

PR #10481, opened 2026-07-22
perf: memoize nextCookies dynamic import

### LINKED ISSUE #10466: perf: nextCookies re-imports next/headers on every request, expensive under ESM loader hooks

MaNaN1803:
Fixes #10466

---
## Summary by cubic
Memoized the dynamic import of `next/headers.js` in `nextCookies` to avoid repeated module loads in production. This reduces overhead and speeds up header/cookie access.

- **Performance**
  - Added `loadNextHeaders()` with a cached Promise in production; dev still imports fresh for HMR.
  - Switched `headers()` and `cookies()` calls to use `loadNextHeaders()` instead of direct dynamic imports.

<sup>Written for commit 66c3143291ecd329e755915b93622e59dfa1c92e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10481?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/better-auth/src/integrations/next-js.ts
+++ b/packages/better-auth/src/integrations/next-js.ts
@@ -24,6 +24,14 @@ export function toNextJsHandler(
 	};
 }
 
+let nextHeadersPromise: Promise<typeof import("next/headers.js")> | undefined;
+const loadNextHeaders = () => {
+	if (process.env.NODE_ENV === "production") {
+		return (nextHeadersPromise ??= import("next/headers.js"));
+	}
+	return import("next/headers.js");
+};
+
 export const nextCookies = () => {
 	let hasWarned = false;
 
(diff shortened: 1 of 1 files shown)


bytaesu:
Since the same code has been submitted in multiple PRs targeting different issues, I'm closing this as it doesn't appear to have been properly reviewed.

PR #10484, opened 2026-07-22
perf(nextjs): memoize next/headers dynamic import to fix ESM hook CPU usage

### LINKED ISSUE #10466: perf: nextCookies re-imports next/headers on every request, expensive under ESM loader hooks

MaNaN1803:
Fixes #10466. The \
extCookies()\ plugin was previously re-importing \
ext/headers.js\ on every request, which caused significant CPU usage when ESM loader hooks (like Sentry or OpenTelemetry) were active. This PR memoizes the import promise at module scope in production, bypassing the expensive dynamic import on subsequent requests.

---
## Summary by cubic
Memoize the `next/headers.js` dynamic import in the Next.js integration to stop re-importing on every request and reduce CPU usage when ESM loader hooks (e.g., Sentry, OpenTelemetry) are active. In production we cache and reuse the import promise; development behavior is unchanged.

<sup>Written for commit 61c1d8d9194cbf80d5a6bb62f9deb6ceed1765ee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10484?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/better-auth/src/integrations/next-js.ts
+++ b/packages/better-auth/src/integrations/next-js.ts
@@ -5,6 +5,14 @@ import { parseSetCookieHeader, toCookieOptions } from "../cookies";
 import { PACKAGE_VERSION } from "../version";
 import { warnIfCookiePluginNotLast } from "./cookie-plugin-guard";
 
+let nextHeadersPromise: Promise<typeof import("next/headers.js")> | undefined;
+const loadNextHeaders = () => {
+	if (process.env.NODE_ENV !== "production") {
+		return import("next/headers.js");
+	}
+	return (nextHeadersPromise ??= import("next/headers.js"));
+};
+
 export function toNextJsHandler(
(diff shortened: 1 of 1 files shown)


bytaesu:
Since the same code has been submitted in multiple PRs targeting different issues, I'm closing this as it doesn't appear to have been properly reviewed.

Issue #6662, opened 2025-12-10
Organization Plugin: Make the "slug" parameter optional


joshsedl:
### Is this suited for github?

- [x] Yes, this is suited for github

### Is your feature request related to a problem? Please describe.

Currently, when using the organization plugin, the "slug" parameter is mandatory, when creating a new organization.

I don't quite understand, why this is the case. The organization can be easily identified through the organization ID. I don't see a reason why this parameter should be required?

### Describe the solution you'd like

Make the "slug" optional

### Describe alternatives you've considered

Add an option to make the slug optional.

JPustkuchen:
Would be nice to understand why it is forced to be required at all? Is there any specific reasons? The example above shows that it doesn't seem to make sense for cases where the "company profile" is not public?

EthanKim88:
I agree with this request. Not all applications need URL-friendly slugs.

For internal B2B systems or admin dashboards, organizations are typically accessed via ID, not slug-based URLs.
Making slug mandatory forces unnecessary input from users.

When I first evaluated Better Auth, the organization plugin was a key feature for my use case. However, I hesitated
to adopt it because slug was mandatory. My original data model didn't have a slug field for organizations - I had to
add it specifically to satisfy Better Auth's requirement. It would have been much easier if slug was optional from
the start.

There's already an open PR #5955 addressing this. Any update on when it might be merged?

ovitorhilario:
Still relevant

ping-maxwell:
Hey, this will be part of our organization rewrite in https://github.com/better-auth/better-auth/pull/7886

JPustkuchen:
Definitely still highly relevant and already planned by @ping-maxwell !

joshsedl:
Yes, still relevant

PR #10502, opened 2026-07-23
feat(organization): slug optional

dvanmali:
Allows explicit optional "slug" parameter for organization creation via `additionalFieldsSchema` marking `slug` as optional.

```ts
plugins: [
  organization({
    schema: {
      organization: {
        additionalFields: {
          slug: {
            type: "string",
            // Required schema changes
            required: false,
            unique: false,
            sortable: false,
            // Optional - forces `slug` to be returned as null in responses
            transform: {
              output: () => null,
            },
          },
        }
      }
    },
  })
]
```

**Closes:**
#6662

### DIFF
--- a/.changeset/organization-slug-optional.md
+++ b/.changeset/organization-slug-optional.md
@@ -14,7 +14,6 @@ plugins: [
             type: "string",
             // Required schema changes
             required: false,
-            // Optional schema changes
             unique: false,
             sortable: false,
             // Optional - forces `slug` to be returned as null in responses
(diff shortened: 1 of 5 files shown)


bytaesu:
Hi @dvanmali 😁

This would break the contract, so we don’t have plans to change the existing org plugin for now.

Max is planning an org plugin rewrite, though. @ping-maxwell could you confirm?

dvanmali:
@bytaesu Updated for a non-schema change approach. Uses additionalFields to allow slug bypass.

PR #10505, opened 2026-07-23
fix(types): expand workspace and consumer type checking

gustavovalverde:
The root build-mode check reached only projects referenced by the root solution and could reuse incremental state, so standalone declaration consumers and demos could fail after the root command passed. This change keeps coverage within existing TypeScript, Turbo, and package-local boundaries, allowing each consumer to own its compiler options and generated prerequisites.

## What changes

- **Project graph**: completes the root references for package sources and semantic test projects, and forces build-mode checking so stale incremental state cannot produce a green result.
- **Workspace consumers**: adds a Turbo `typecheck:consumer` task that builds dependencies before checking docs and standalone declaration consumers with their own compiler configurations.
- **Generated applications**: cleans and regenerates framework types in package-owned scripts, while the demo matrix installs frozen dependencies and runs every demo's local typecheck.
- **Surfaced contracts**: repairs client-plugin inference, strict optional-property call sites, and demo or test integrations that had drifted from current dependency APIs.

## Review boundaries

The consumer task uses `^build` because these fixtures verify the declarations that applications consume, not only compatibility with workspace source. Runtime-only Solid/Vinxi, smoke orchestration, and IPv6 fixtures remain owned by their existing e2e jobs.

> [!NOTE]
> This diff does not establish direct `dash()` compatibility with the published `@better-auth/infra`; the Next.js demo's runtime ID guard cannot prove nested hook compatibility.


docs/package.json
```diff
@@ -16,7 +16,8 @@
     "sync-beta": "node scripts/sync-beta-content.ts",
     "sync-typesense": "node scripts/sync-typesense.ts",
     "format:md": "remark . --ext md,mdx --output --quiet",
-    "format:check": "remark . --ext md,mdx --quiet --frail"
+    "format:check": "remark . --ext md,mdx --quiet --frail",
+    "typecheck:consumer": "git clean -xdf .next .source && next typegen && fumadocs-mdx && tsc --noEmit --incremental false"
```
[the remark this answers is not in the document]
bytaesu: cc @gustavovalverde

packages/core/src/error/codes.ts
```diff
@@ -1,3 +1,4 @@
+import type {} from "@better-auth/core";
```
bytaesu, line 1: what issue does this addition address? 🤔

bytaesu:
nit: as a type-checking improvement PR, targeting `main` could be fine. We'll promote next -> main soon anyway, so it's fine

packages/core/src/error/codes.ts
gustavovalverde: This type-only self-import forces TypeScript to resolve `@better-auth/core` before augmenting it, without emitting runtime code. Without it, package-local `tsc` reports `TS2664`

gustavovalverde:
> We'll promote next -> main soon anyway, so it's fine

At the same time, we'll be promoting `next` to `main` soon, so I'll keep it in `next` for an RC, just in case, to confirm in a real deployment.

packages/core/src/error/codes.ts
bytaesu: @gustavovalverde  I can't reproduce this. Does it only occur under specific conditions?

bytaesu:
> https://github.com/better-auth/better-auth/pull/10505#discussion_r3658855277
> https://github.com/better-auth/better-auth/pull/10505#discussion_r3657188813

I think these are not resolved yet @gustavovalverde

PR #10522, opened 2026-07-25
docs: add plugin marketplace

ping-maxwell:


bytaesu (commented):
If this is just for showcasing plugins, I think it'd be better to improve the existing plugin docs UI rather than introduce a separate "Marketplace" 🤔

bytaesu:
> If this is just for showcasing plugins, I think it'd be better to improve the existing plugin docs UI rather than introduce a separate "Marketplace" 🤔

Closing this PR for now. Feel free to reopen it, use it as a reference, or build on top of it.

PR #10554, opened 2026-07-27
fix(mcp): export MCPOptions so consumer declaration emit can name it

### LINKED ISSUE #10532: mcp: MCPOptions is never exported (no ./plugins/mcp subpath either) — TS4023 under declaration emit for consumers
### Reproduction

Any package compiled with `declaration: true` (composite builds, or projects that lint exported-API portability) that exports a value typed by the mcp plugin:

```ts
// auth.ts
import { mcp } from "better-auth/plugins";

  plugins: [mcp({ loginPage: "/login" })],
});
```

```
error TS4023: Exported variable 'auth' has or is using name 'MCPOptions' from external module ".../node_modules/better-auth/dist/plugins/mcp/index" but cannot be named.
```

The same failure surfaces on any exported value whose inferred type transitively references the auth instance — e.g. tRPC procedures built over a context that embeds `auth`.

**Current:** `MCPOptions` is declared but never exported anywhere reachable:

- `packages/better-auth/src/plugins/mcp/index.ts` (~line 49) declares `interface MCPOptions {` without `export`, and the module's export statement omits it. In the published dist this is `dist/plugins/mcp/index.d.mts`, which ends with `export { getMCPProtectedResourceMetadata, getMCPProviderMetadata, mcp, oAuthDiscoveryMetadata, oAuthProtectedResourceMetadata, withMcpAuth };` — the type is missing from the list.
- The `better-auth/plugins` barrel re-exports every sibling plugin's options type (`AdminOptions`, `OIDCOptions`, `MagicLinkOptions`, `DeviceAuthorizationOptions`, …) but not `MCPOptions`.
- The package `exports` map has `./plugins/mcp/client` and `./plugins/mcp/client/adapters` but no `./plugins/mcp`, so the declaring module has no resolvable specifier at all.

TypeScript's declaration emitter therefore has no accessible path to name the type, and there is no consumer-side workaround short of patching the package (a type cannot be re-exported by consumers if it was never exported).

**Expected:** `MCPOptions` is importable like every other plugin's options type, so declaration emit can synthesize a reference to it (`import("better-auth/plugins/mcp").MCPOptions` or via the `plugins` barrel).

### Suggested fix

Two additive, type-only changes:

1. Export the interface in `packages/better-auth/src/plugins/mcp/index.ts` (`export interface MCPOptions`).
2. Make it reachable: re-export it from the `plugins` barrel like the sibling `*Options` types, and/or add a `./plugins/mcp` entry to the package `exports` map mirroring the existing `./plugins/mcp/client` shape.

We currently run a patched build with exactly this change (type export + `./plugins/mcp` subpath) and it resolves the error with no runtime impact — nothing needs to import the new subpath at runtime; it exists purely so declaration emit can name the type. Happy to send a PR.

Reproduced on 1.6.11; still present on 1.6.25 (latest at time of filing — dist declarations and exports map unchanged in this respect).

### System info

TypeScript 5.9.3, `declaration: true` (with `noEmit` — declaration diagnostics only). Not OS-specific (observed on macOS and Linux CI).

MCP plugin, TypeScript types.

### Additional context

Note that only exporting the type without making its module resolvable would shift the error to TS2742 ("likely not portable") rather than fixing it — both halves of the fix are needed unless the barrel re-export is used.

guillaume-flambard:
## Problem

`MCPOptions` is the only plugin options interface that is not exported. Any consumer compiled with `declaration: true` that exports a value typed by the auth instance fails:

```
error TS4023: Exported variable 'auth' has or is using name 'MCPOptions' from external module ".../better-auth/dist/plugins/mcp/index" but cannot be named.
```

The declaration emitter has no accessible path to name a type that was never exported, and there is no consumer-side workaround. Reported in #10532 (reproduced on 1.6.11, still present on 1.6.25).

## Fix

Add `export` to the `interface MCPOptions` declaration, exactly as its siblings do (`BearerOptions`, etc.). Because `packages/better-auth/src/plugins/index.ts` already does `export * from "./mcp"`, and `./plugins` is in the package `exports` map, the interface immediately becomes reachable as `import("better-auth/plugins").MCPOptions` — a portable path declaration emit can synthesize. Type-only, one word, no runtime impact.

## Verification

Minimal reproduction of the mechanism (an unexported interface used by an exported value, `declaration: true`):

```ts
// before: interface Opts {...}          -> TS4023 on the exported consumer
// after:  export interface Opts {...}    -> clean
```

`tsc` reports TS4023 before the export and nothing after. The barrel re-export gives the same portable name in this package.

## A note on the exports map

The issue also suggests adding a `./plugins/mcp` subpath to the `exports` map (mirroring the existing `./plugins/mcp/client`). I left that out on purpose: the barrel path above already resolves the error without it, so the one-line export is the minimal fix. If you'd prefer the symmetry with `./plugins/mcp/client` — a direct `import("better-auth/plugins/mcp").MCPOptions` — I'm happy to add the subpath entry in this PR; it's a design call on the package surface that's yours to make.

## Ticket

Fixes #10532

---
## Summary by cubic
Exported `MCPOptions` from the MCP plugin to fix TS4023 for consumers compiling with `declaration: true` by making the type reachable via `better-auth/plugins`; type-only change with no runtime impact. Fixes #10532.

<sup>Written for commit 13379bad797124a50a252543f1ed428ceab44aa2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10554?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/export-mcp-options.md
+++ b/.changeset/export-mcp-options.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Export the `MCPOptions` interface from the MCP plugin. It was the only plugin options type left unexported, so any consumer compiling with `declaration: true` that exports a value typed by the auth instance failed with TS4023 ("has or is using name 'MCPOptions' … but cannot be named"). The interface is now exported like its siblings (e.g. `BearerOptions`) and reachable through the `better-auth/plugins` barrel, giving declaration emit a portable path to name it. Type-only change, no runtime impact.
--- a/packages/better-auth/src/plugins/mcp/index.ts
+++ b/packages/better-auth/src/plugins/mcp/index.ts
@@ -46,7 +46,7 @@ declare module "@better-auth/core" {
 	}
 }
 
-interface MCPOptions {
+export interface MCPOptions {
 	loginPage: string;
 	resource?: string | undefined;
 	oidcConfig?: OIDCOptions | undefined;


gustavovalverde:
We won't make further fixes to MCP in 1.6; please refer to the MCP plugin in 1.7, and if the fix is necessary there, target this to the `next` branch. Thank you.

PR #10564, opened 2026-07-27
docs: add agent feedback endpoint

Bekacru:
## Summary

Adds a public, unauthenticated endpoint at `POST /api/feedback` that AI agents can use to report incorrect, missing, or unclear documentation, plus bugs and feature requests.

## Why this shape

There is no cross-vendor standard for agent feedback submission. `/.well-known/` names are IANA-registered under RFC 8615 and there is no `feedback` entry, so a `/.well-known/feedback` route would squat the registry namespace while no agent probes for it. Discovery instead goes through the channels agents actually read.

## What's included

**Endpoint** — `docs/app/api/feedback/route.ts`

- `POST` is zod-validated with an Upstash sliding-window rate limit of 10/h per IP, mirroring the existing enterprise-contact route.
- `GET` on the same path returns a self-describing JSON schema, so an agent holding only the URL can learn the payload shape without reading prose docs.
- `422` responses include per-field `issues` alongside the schema, making a failed submission self-correcting.
- CORS is open with an `OPTIONS` preflight, since agents may call from any origin.

**Storage** — `docs/lib/feedback.ts`

- Writes to Upstash: `LPUSH` onto `feedback:inbox`, `LTRIM` to 5000, plus a per-type counter.
- No IP is retained — only user-agent and referer. The rate limiter already uses the IP.
- `notifySlack()` is wired but no-ops unless `SLACK_FEEDBACK_WEBHOOK_URL` is set, so enabling Slack delivery is an env-var change rather than a code change. A notification failure never fails the submission.

**Discovery**

- A `## Feedback` block near the top of `/llms.txt`, in the spec's details slot above the long table of contents.
- A footer appended to every per-page markdown response under `/llms.txt/**.md`. This channel is agent-only by construction: humans get the rendered HTML and never see it.
- Both strings are shared constants, so they cannot drift from the schema.

**Docs** — new `docs/content/docs/ai-resources/feedback.mdx` page.

## Deployment note

`SLACK_FEEDBACK_WEBHOOK_URL` needs to be set in Vercel before Slack delivery starts. Without it, submissions are stored silently.

## Not included

The documentation MCP server at `mcp.better-auth.com` is deployed from a separate repository, so the `submit_feedback` tool is not part of this PR. It is a thin wrapper: an input schema matching `feedbackSchema`, and a handler that POSTs to this endpoint.

## Test plan

The docs app has no test infrastructure (no vitest config, no test script, no test files), so this was verified end to end against a local dev server:

- `GET /api/feedback` → `200` with the descriptor
- Valid `POST` → `201` with a submission id
- Invalid `POST` (short message, bad enum) → `422` with per-field issues
- Malformed JSON → `400`
- `OPTIONS` → `204` with CORS headers
- `/llms.txt` renders the Feedback section; `/llms.txt/docs/introduction.md` renders the footer
- `/docs/ai-resources/feedback` renders `200`

`tsc --noEmit` on the docs app is clean apart from one pre-existing unrelated error in `docs/components/blog/blog-toc.tsx`.

### DIFF
--- a/.cspell/tech-terms.txt
+++ b/.cspell/tech-terms.txt
@@ -73,3 +73,5 @@ pjpeg
 styl
 vbscript
 unvalidated
+lpush
+ltrim
--- a/docs/app/api/feedback/route.ts
+++ b/docs/app/api/feedback/route.ts
@@ -7,6 +7,8 @@ import {
 	FEEDBACK_DESCRIPTOR,
 	feedbackSchema,
 	notifySlack,
+	sanitizeReferer,
+	sanitizeUserAgent,
 	storeFeedback,
 } from "@/lib/feedback";
 
@@ -36,9 +38,19 @@ function getRatelimit(): Ratelimit {
 /**
  * Self-describing discovery response, so an agent that finds the URL can learn
  * the payload shape without reading the docs.
+ *
+ * `endpoint` is derived from the request rather than taken from the constant,
+ * so preview and self-hosted deployments describe the host that actually
+ * served the response instead of pointing back at production.
  */
-export function GET() {
-	return NextResponse.json(FEEDBACK_DESCRIPTOR, { headers: CORS_HEADERS });
+export function GET(request: Request) {
+	return NextResponse.json(
+		{
+			...FEEDBACK_DESCRIPTOR,
+			endpoint: new URL("/api/feedback", request.url).toString(),
+		},
+		{ headers: CORS_HEADERS },
+	);
 }
 
 export function OPTIONS() {
@@ -86,8 +98,8 @@ export async function POST(request: Request) {
 			...parsed.data,
 			id: crypto.randomUUID(),
 			receivedAt: new Date().toISOString(),
-			userAgent: request.headers.get("user-agent"),
-			referer: request.headers.get("referer"),
+			userAgent: sanitizeUserAgent(request.headers.get("user-agent")),
+			referer: sanitizeReferer(request.headers.get("referer")),
 		};
 
 		const stored = await storeFeedback(record);
--- a/docs/app/api/feedback/route.ts
+++ b/docs/app/api/feedback/route.ts
@@ -1,0 +1,123 @@
+import { Ratelimit } from "@upstash/ratelimit";
+import { Redis } from "@upstash/redis";
+import { NextResponse } from "next/server";
+import { getClientIP } from "@/lib/ai-chat/rate-limit";
+import type { FeedbackRecord } from "@/lib/feedback";
+import {
+	FEEDBACK_DESCRIPTOR,
+	feedbackSchema,
+	notifySlack,
+	storeFeedback,
+} from "@/lib/feedback";
+
+// Agents may call this from any origin, including a browser context.
+const CORS_HEADERS = {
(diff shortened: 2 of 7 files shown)


Bekacru:
Addressed all review feedback in 5c9501a54. Deduping across Greptile, Copilot, and cubic, there were five distinct issues — all valid.

**1. Slack webhook rejections passed silently** (Greptile P1, cubic P2)

Fixed. `notifySlack` now throws on a non-2xx response. The caller still catches it, so a failed notification never fails the submission — it just reaches the error log instead of vanishing.

Verified against a local webhook returning `403`: the request still returns `201`, and the server logs `Slack feedback notification failed Error: Slack webhook returned 403 Forbidden`.

**2. Untrusted content interpolated into Slack mrkdwn** (Greptile P2, cubic P2)

Fixed. Every field on a submission is attacker-controlled since the endpoint is public and unauthenticated.

- The message body now renders as a `plain_text` block, which processes no markup at all. That's what actually neutralizes the vector, rather than trying to escape mrkdwn correctly.
- The remaining mrkdwn — the fallback `text` field and the metadata line — is entity-escaped for `&`, `<`, `>`, so `<https://evil|looks-official>` and `<!channel>` cannot be injected.
- Backticks in `context` are replaced, so it cannot close its code fence early and let the remainder render as active formatting.

Verified by posting `{"message": "<https://evil.example|Click here to reset your password> & <!channel> injection attempt", "context": "line1\n```\n*not bold* <https://evil.example|x>", "page": "/docs/<script>"}` at a local webhook receiver and inspecting the delivered payload: the fallback text arrives as `&lt;https://evil.example|...&gt; &amp; &lt;!channel&gt;`, the fence content as ` ```line1\n'''\n*not bold* &lt;...&gt;``` `, and the page field as `/docs/&lt;script&gt;`.

**3. Invalid JSON in the `/llms.txt` example and the per-page footer** (Copilot ×2, cubic)

Fixed, and this was the most consequential one — both embedded a union expression inside a JSON string, so an agent copying either got a `400` before it could learn the schema. Each now shows one valid enum value, with the allowed values listed alongside rather than inlined into the example.

Verified by piping the `/llms.txt` example through `json.load` (parses, `type = docs-missing`) and by POSTing the footer example verbatim (`201`).

**4. Hard-coded `endpoint` in the discovery response** (Copilot)

Fixed. `GET /api/feedback` now derives `endpoint` from the request URL, so preview and self-hosted deployments describe the host that actually served the response. Verified: a request to the dev server returns `http://localhost:3459/api/feedback` rather than the production URL. The static strings in `/llms.txt` and the page footer deliberately stay canonical, since those point at the published docs.

**5. Referer may carry tokens, and neither header was bounded** (Copilot)

Fixed. Query string and fragment are stripped from the stored referer, and both referer and user-agent are truncated. Verified: a request with `Referer: https://example.com/docs/page?access_token=SECRET123#frag` stores `https://example.com/docs/page`.

`tsc --noEmit` on the docs app is clean.

bytaesu (commented):
I don't think this is really necessary.

If something feels off during development, AI can scan the code and easily open a GitHub issue. I don't see much added value in providing a prompt with extra instructions.

Closing for now. Feel fee to reopen this.

PR #10570, opened 2026-07-28
fix(db): always set a ttl on secondary-storage verification reservations

### LINKED ISSUE #10569: reserveVerificationValue stores secondary-storage reservations permanently when the ttl floors to zero

### Reproduction

`reserveVerificationValue` writes its reservation to secondary storage with a TTL computed by `getTTLSeconds`, which floors to whole seconds and clamps at `0`:

```ts
function getTTLSeconds(expiresAt: Date | number, now = Date.now()): number {
	const expiresMs = typeof expiresAt === "number" ? expiresAt : expiresAt.getTime();
	return Math.max(Math.floor((expiresMs - now) / 1000), 0);
}
```

In `packages/better-auth/src/db/internal-adapter.ts` the secondary-storage-only branch passes that value straight through, with no `ttl > 0` guard:

```ts
await secondaryStorage.set(
	cacheKey,
	JSON.stringify({ id: reservationId, identifier: storedIdentifier, value: data.value, expiresAt: data.expiresAt }),
	getTTLSeconds(data.expiresAt), // 0 when under a second remains
);
```

Secondary storage implementations read a zero TTL as "no expiry". `@better-auth/redis-storage` is explicit about it:

```ts
async set(key, value, ttl) {
	if (ttl !== undefined && ttl > 0) {
		await client.setex(prefixedKey, ttl, value);
	} else {
		await client.set(prefixedKey, value); // no expiry
	}
}
```

So a reservation with under one second of remaining lifetime is stored **permanently**.

Steps to reproduce:

1. Configure Better Auth with a secondary storage and leave `verification.storeInDatabase` unset (the default), so reservations take the secondary-storage-only path.
2. Call `reserveVerificationValue` with an `expiresAt` under one second away:
   ```ts
   await auth.$context.internalAdapter.reserveVerificationValue({
     identifier: "reserve:demo",
     value: "demo",
     expiresAt: new Date(Date.now() + 100),
   });
   ```
3. Inspect the key in Redis: `TTL better-auth:verification:reserve:demo` returns `-1` (no expiry) instead of a positive value.

**Current behavior**

The reservation key is written with no expiry and is never reclaimed. Every such reservation accumulates permanently in the store.

This is reachable in production through the SAML replay guard, which is the only non-test caller. In `packages/sso/src/routes/saml-pipeline.ts` the expiry is derived from IdP-supplied input:

```ts
const expiresAt = conditions?.notOnOrAfter
	? new Date(conditions.notOnOrAfter).getTime() + clockSkew
	: Date.now() + constants.DEFAULT_ASSERTION_TTL_MS;
```

and `packages/sso/src/saml/timestamp.ts` accepts an assertion when:

```ts
if (now > notOnOrAfterTime + clockSkew) { /* reject */ }
```

The accepted bound and the reservation expiry are the same value, so a valid assertion can have arbitrarily little time left. Any assertion arriving in the final second of its (skew-extended) validity floors to a TTL of `0` and writes a permanent replay tombstone. With the default 5 minute clock skew, an assertion whose `NotOnOrAfter` is already ~5 minutes stale still passes validation and lands right at that boundary.

**Expected behavior**

The reservation should always carry a positive TTL so it expires and is reclaimed, matching the other verification writes in the same file, both of which already guard with `if (ttl > 0)`.

1.6.25

### System info

```bash
Better Auth: better-auth 1.6.25 (packages/better-auth)
Affected: packages/better-auth/src/db/internal-adapter.ts
Caller: packages/sso/src/routes/saml-pipeline.ts
Storage: @better-auth/redis-storage (any SecondaryStorage treating ttl<=0 as no-expiry)
Runtime: Node.js (platform independent)
```

Backend, Package

```typescript
import { Redis } from "ioredis";
import { redisStorage } from "@better-auth/redis-storage";

  secondaryStorage: redisStorage({ client: new Redis() }),
  // verification.storeInDatabase left unset -> secondary-storage-only reservations
  plugins: [sso()],
});
```

### Additional context

**Scope, and an open question for maintainers.**

While tracing this I audited every `secondaryStorage.set` call in `internal-adapter.ts`. The same flooring issue exists in three session paths, but they are **not** included in the linked PR because an existing test pins the current behaviour:

- `refreshUserSessions` passes `Math.floor(sessionTTL)` with no guard.
- The `active-sessions` list writes in the session update and delete paths guard on the raw millisecond expiry (`furthestSessionExp > now`) while passing a whole-second TTL, so a list with under a second left still yields `0`.

The test `should calculate TTL correctly with Math.floor for secondary storage` asserts `expect(capturedTTLs.at(-1)).toBe(0)` for a 500ms session, explicitly documenting "less than 1 second should round to 0". That assertion is about the rounding arithmetic rather than the storage consequence, but changing the session paths would mean changing a behaviour the suite currently pins, so it seemed wrong to do unilaterally in the same PR.

Worth noting these session entries are a storage leak only, not an auth bypass: the read path in `api/routes/session.ts` re-validates `session.session.expiresAt < new Date()` and rejects expired sessions regardless of the stored TTL.

Happy to extend the PR to cover the session paths (and update that test) if you agree they should be fixed too.

jashkarangiya:
## Summary

`reserveVerificationValue` wrote its secondary-storage reservation with `getTTLSeconds(data.expiresAt)`, which floors to `0` once the remaining lifetime drops below one second. Secondary storage implementations treat a zero TTL as "no expiry", so the reservation was stored **permanently** and never reclaimed.

Closes #10569

## The mismatch

`getTTLSeconds` clamps at zero:

```ts
return Math.max(Math.floor((expiresMs - now) / 1000), 0);
```

and `@better-auth/redis-storage` reads a zero TTL as "store forever":

```ts
if (ttl !== undefined && ttl > 0) {
	await client.setex(prefixedKey, ttl, value);
} else {
	await client.set(prefixedKey, value); // no expiry
}
```

The caller means "no time left"; the store hears "keep this forever".

## Why it is reachable

The only non-test caller is the SAML replay guard, and its expiry comes from IdP-supplied input:

```ts
// packages/sso/src/routes/saml-pipeline.ts
const expiresAt = conditions?.notOnOrAfter
	? new Date(conditions.notOnOrAfter).getTime() + clockSkew
	: Date.now() + constants.DEFAULT_ASSERTION_TTL_MS;
```

```ts
// packages/sso/src/saml/timestamp.ts
if (now > notOnOrAfterTime + clockSkew) { /* reject as expired */ }
```

The bound the validator accepts and the reservation expiry are the same value, so an assertion arriving in the final second of its skew-extended validity is accepted and floors to a TTL of `0`. With the default 5 minute clock skew, an assertion whose `NotOnOrAfter` is already ~5 minutes stale still passes and lands exactly on that boundary.

## The fix

```diff
-					getTTLSeconds(data.expiresAt),
+					// A sub-second remaining lifetime floors to 0, which secondary
+					// storage implementations read as "no expiry" and would persist
+					// the reservation forever. Keep the tombstone alive for the rest
+					// of its (sub-second) window instead of leaking it.
+					Math.max(getTTLSeconds(data.expiresAt), 1),
```

The two sibling verification writes in the same file already guard with `if (ttl > 0)`; this path was the odd one out. I used a one second floor rather than skipping the write, because skipping would silently drop replay protection for that assertion, while the floor both preserves the tombstone for its remaining window and guarantees it expires.

## Tests

Added `always sets a ttl on the secondary-storage reservation`, using the mock-storage style already in this file. It reserves with a 100ms expiry and asserts the stored key carries a positive TTL.

Verified it fails on the unpatched code with `received "undefined"` for the TTL while the key itself is present, which is the permanent-key signature, and passes after the fix.

`internal-adapter.test.ts` 52/52, the full `packages/sso` suite 441/441, and `pnpm typecheck` exits 0.

## Scope note

I audited every `secondaryStorage.set` call in `internal-adapter.ts` and the same flooring issue exists in three session paths (`refreshUserSessions`, and the `active-sessions` list writes on session update and delete, which guard on raw millisecond expiry while passing a whole-second TTL).

Those are deliberately **not** in this PR: the existing test `should calculate TTL correctly with Math.floor for secondary storage` asserts `expect(capturedTTLs.at(-1)).toBe(0)` for a 500ms session, so changing them means changing behaviour the suite currently pins. Those entries are also a storage leak only, not an auth bypass, since the read path re-validates `expiresAt` and rejects expired sessions regardless of the stored TTL.

Happy to extend this PR to cover them and update that test if you would like.

---
## Summary by cubic
Ensure verification reservations in secondary storage always expire by clamping TTL to at least 1s, preventing permanent tombstones when less than a second remains. This covers the SAML replay guard path and stops storage leaks while keeping replay protection.

- **Bug Fixes**
  - Clamp TTL with Math.max(getTTLSeconds(expiresAt), 1) in reserveVerificationValue to avoid zero-TTL “store forever” behavior.
  - Add a test that asserts a positive TTL is set for sub-second expiries.

<sup>Written for commit a369c907b576675d2089054444274f468c025e16. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10570?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/verification-reservation-ttl.md
+++ b/.changeset/verification-reservation-ttl.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Always set a positive TTL when reserving a verification value in secondary storage. A reservation with under a second of remaining lifetime floored to a TTL of `0`, which secondary storage implementations read as "no expiry", leaving the entry stored permanently.
--- a/packages/better-auth/src/db/internal-adapter.test.ts
+++ b/packages/better-auth/src/db/internal-adapter.test.ts
@@ -1962,5 +1962,45 @@ describe("internal adapter test", async () => {
 			expect(first).toBe(true);
 			expect(second).toBe(true);
 		});
+
+		it("always sets a ttl on the secondary-storage reservation", async () => {
+			// A reservation whose remaining lifetime is under one second floors to
+			// a ttl of 0. Secondary storage implementations treat a zero ttl as
+			// "no expiry" (e.g. redis SETEX is skipped in favour of SET), so the
+			// tombstone would be written permanently and leak. Callers can hit
+			// this with an externally supplied expiry: the SAML replay guard
+			// derives `expiresAt` from the assertion's NotOnOrAfter plus the
+			// clock skew, which is exactly the bound the timestamp validator
+			// allows, so an assertion arriving in its final second lands here.
+			const dataMap = new Map<string, string>();
+			const ttlMap = new Map<string, number>();
+			const adapter = await makeAdapter({
+				secondaryStorage: {
+					set(key: string, value: string, ttl?: number) {
+						dataMap.set(key, value);
+						if (ttl) ttlMap.set(key, ttl);
+					},
+					get(key: string) {
+						return dataMap.get(key) || null;
+					},
+					delete(key: string) {
+						dataMap.delete(key);
+						ttlMap.delete(key);
+					},
+				},
+			});
+
+			const reserved = await adapter.reserveVerificationValue({
+				identifier: "reserve:expiring-now",
+				value: "jti-expiring",
+				expiresAt: new Date(Date.now() + 100),
+			});
+
+			expect(reserved).toBe(true);
+			const key = "verification:reserve:expiring-now";
+			expect(dataMap.has(key)).toBe(true);
+			// Without a ttl the key would never expire.
+			expect(ttlMap.get(key)).toBeGreaterThan(0);
+		});
 	});
 });
--- a/packages/better-auth/src/db/internal-adapter.ts
+++ b/packages/better-auth/src/db/internal-adapter.ts
@@ -1437,7 +1437,11 @@ export const createInternalAdapter = (
 						value: data.value,
 						expiresAt: data.expiresAt,
 					}),
-					getTTLSeconds(data.expiresAt),
+					// A sub-second remaining lifetime floors to 0, which secondary
+					// storage implementations read as "no expiry" and would persist
+					// the reservation forever. Keep the tombstone alive for the rest
+					// of its (sub-second) window instead of leaking it.
+					Math.max(getTTLSeconds(data.expiresAt), 1),
 				);
 				return true;
 			}


jashkarangiya:
Hey @Bekacru, could you take a look at this one when you get a chance? It clamps the verification reservation ttl to at least 1 second so it is not stored permanently when under a second remains.

jashkarangiya:
Closing as obsolete. #10000 replaced the secondary-storage reservation branch with a throw, so the path this patched no longer exists and the diff resolves to empty against main.

PR #10572, opened 2026-07-28
fix(anonymous): link accounts after oauth proxy callback

### LINKED ISSUE #10562: `oAuthProxy` is not behaviourally transparent: completing auth off-pipeline silently disables every path-matched plugin hook

### Reproduction

Steps to reproduce:

1. Configure `betterAuth` with a social provider, the `anonymous` plugin (with an `onLinkAccount` callback) and `oAuthProxy` — see the auth config below.
2. From an origin that is **not** `productionURL` (a preview deployment, a dev tunnel, or localhost), call `authClient.signIn.anonymous()`.
3. Call `authClient.signIn.social({ provider: "google" })` and complete the provider flow.
4. Observe that `onLinkAccount` never fires.

Setting `productionURL` to the current origin (so the proxy no-ops) makes the same flow work.

**Expected:** enabling `oAuthProxy` should not change what authentication *does*. It exists because Google and Apple require exact-match `redirect_uri`s, so preview deployments and dev tunnels can never be registered — it's a transport detour, not a behavioural one. A flow completing through the proxy should be indistinguishable, to plugins, from one that doesn't.

**Actual:** when the request origin differs from `productionURL`, sign-in completes on `/oauth-proxy-callback`, which finalises auth itself — decrypting the payload and calling `handleOAuthUserInfo` and `setSessionCookie` directly, then redirecting. It does not re-enter the canonical `/callback/:id` route. Any plugin whose hooks are matched against the canonical completion paths therefore never fires.

There is no second chance on the production leg either: `oAuthProxy` registers its `/callback/:id` hook as a **before** hook that throws a redirect, and a throwing before-hook propagates out of `runBeforeHooks` before the endpoint or `runAfterHooks` are reached (`api/to-auth-endpoints.ts`). So off-production, `/oauth-proxy-callback` is the *only* place a session is established — and it is the one place plugins aren't looking.

**The concrete symptom here** is the `anonymous` plugin: `onLinkAccount` never fires under the proxy, so everything an anonymous user accumulated pre-signup is silently dropped. Its `hooks.after` matcher (`plugins/anonymous/index.ts`, unchanged on `main` at time of writing) lists `/sign-in`, `/sign-up`, `/callback`, `/oauth2/callback`, `/magic-link/verify`, `/email-otp/verify-email`, `/one-tap/callback`, `/passkey/verify-authentication`, `/phone-number/verify`, `/verify-email`. `/oauth-proxy-callback` matches none of them.

Instrumenting a `matcher: () => true` after-hook shows that both inputs the plugin needs are present on that path — it is simply never invoked:

```json
{"path":"/oauth-proxy-callback",
 "requestSession":{"id":"144a326c…","isAnonymous":true},
 "newSession":{"id":"bf13b53d…"}}
```

**This is not specific to the `anonymous` plugin.** The same root cause has been reported and patched one plugin at a time:

- #6022 — SIWE
- #6128 — passkey
- #10112 — `last-login-method` (open; its description explicitly calls this "the same class of bug as #6128 for passkey and #6022 for SIWE")
- #9390 / #9391 / #9392 — `linkSocial` (the mirror image: the proxy fails to *engage*, because its before-hook matches only `/sign-in/social` and `/sign-in/oauth2`)

The failure is invisible in production, where the proxy no-ops, and present in every preview, tunnel and local environment.

1.6.11

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "release": "25.4.0",
    "cpuCount": 16,
    "cpuModel": "Apple M3 Max"
  },
  "node": { "version": "v25.9.0", "env": "development" },
  "packageManager": { "name": "npm", "version": "11.12.1" },
  "frameworks": [
    { "name": "next", "version": "^16.3.0-canary.28" },
  ],
  "betterAuth": { "version": "1.6.11" }
}
```

Backend

```typescript
import { anonymous, oAuthProxy } from "better-auth/plugins"

  socialProviders: {
    google: { clientId: "...", clientSecret: "..." },
  },
  plugins: [
    anonymous({
        console.log("onLinkAccount", anonymousUser.user.id, "->", newUser.user.id)
      },
    }),
    oAuthProxy({ productionURL: "https://example.com" }),
  ],
})
```

### Additional context

Reproducible locally and on preview deployments — anywhere the request origin differs from `productionURL`. Not reproducible in production, where the proxy no-ops.

**A request about the shape of the fix:** please don't resolve this by adding `/oauth-proxy-callback` to the `anonymous` plugin's matcher. That would close this report while leaving the cause in place for the next plugin — and for third-party plugins, which can't be patched centrally at all. Given this is now the fifth instance, I'd rather report the shared root cause than one more symptom.

Some directions, though you'll know the codebase far better than I do:

- Have the proxy's return leg re-enter the canonical callback pipeline rather than reimplementing its tail, so existing matchers keep working untouched.
- Or introduce a path-independent lifecycle event for "a session was established", so plugins stop enumerating routes. Worth noting if you go this way: `setSessionCookie` is also called by `/multi-session/set-active`, which is a session *switch* rather than an establishment — a naive `() => true` hook would conflate the two.
- Or give plugins a declarative way to declare "I care about auth completion" that the proxy participates in.

bytaesu:
Closes #10562

### DIFF
--- a/packages/better-auth/src/plugins/anonymous/index.ts
+++ b/packages/better-auth/src/plugins/anonymous/index.ts
@@ -256,6 +256,7 @@ export const anonymous = (options?: AnonymousOptions | undefined) => {
 							ctx.path?.startsWith("/sign-up") ||
 							ctx.path?.startsWith("/callback") ||
 							ctx.path?.startsWith("/oauth2/callback") ||
+							ctx.path?.startsWith("/oauth-proxy-callback") ||
 							ctx.path?.startsWith("/magic-link/verify") ||
 							ctx.path?.startsWith("/email-otp/verify-email") ||
 							ctx.path?.startsWith("/one-tap/callback") ||
--- a/.changeset/link-anonymous-oauth-proxy.md
+++ b/.changeset/link-anonymous-oauth-proxy.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Ensure the anonymous plugin's `onLinkAccount` callback runs after signing in through the OAuth proxy.
--- a/packages/better-auth/src/plugins/oauth-proxy/oauth-proxy.test.ts
+++ b/packages/better-auth/src/plugins/oauth-proxy/oauth-proxy.test.ts
@@ -14,6 +14,8 @@ import { parseJSON } from "../../client/parser";
 import { signJWT, symmetricDecrypt, symmetricEncrypt } from "../../crypto";
 import { getTestInstance } from "../../test-utils/test-instance";
 import { DEFAULT_SECRET } from "../../utils/constants";
+import { anonymous } from "../anonymous";
+import { anonymousClient } from "../anonymous/client";
 import { oAuthProxy } from ".";
 
 let testIdToken: string;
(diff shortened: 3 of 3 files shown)


packages/better-auth/src/plugins/anonymous/index.ts
```diff
@@ -256,6 +256,7 @@ export const anonymous = (options?: AnonymousOptions | undefined) => {
 							ctx.path?.startsWith("/sign-up") ||
 							ctx.path?.startsWith("/callback") ||
 							ctx.path?.startsWith("/oauth2/callback") ||
+							ctx.path?.startsWith("/oauth-proxy-callback") ||
```
bytaesu, line 259: I think relying on an allowlist of request paths to identify authentication flows that may replace an anonymous session is structurally fragile. As this PR demonstrates, new authentication flows can easily be missed unless their paths are added explicitly.

bytaesu:
Closing in favor of #10988

PR #10573, opened 2026-07-28
fix(sso): read SAML login InResponseTo from extract.response

### LINKED ISSUE #10504: [SSO] SAML login InResponseTo extraction reads extract.inResponseTo (always undefined) — validation no-ops, and allowIdpInitiated:false rejects all SP-initiated logins

### Reproduction

1. Configure `sso({ saml: { allowIdpInitiated: false } })` (InResponseTo validation left at its default ON).
2. Register a SAML provider and perform a normal **SP-initiated** login against any IdP (reproduced against Keycloak 23.0.4; the SAMLResponse carries a correct `InResponseTo` on the `Response` element).
3. The ACS rejects the login with `unsolicited_response` / "IdP-initiated SSO not allowed", logging `SAML InResponseTo validation failed: ... InResponseTo missing`.

**Current:** in the login SAML pipeline the request id is read as `extract.inResponseTo`:

```js
const inResponseTo = extract.inResponseTo;
```

samlify's `loginResponseFields` extractor places the Response element's attributes under the `response` key (`extractor.ts`: `{ key: 'response', localPath: ['Response'], attributes: ['ID', 'IssueInstant', 'Destination', 'InResponseTo'] }`), so `extract.inResponseTo` is always `undefined`. Consequences:

- With default options (`allowIdpInitiated` true), every SP-initiated login silently falls into the unsolicited branch — **InResponseTo replay validation never actually runs**, even though it is "on".
- With `allowIdpInitiated: false`, **every SP-initiated login is rejected** as unsolicited.

The logout pipeline in the same file already uses the corrected form (`extract?.response?.inResponseTo`).

**Expected:** the login path reads the id where samlify puts it, e.g.

```js
const inResponseTo = extract.inResponseTo ?? extract.response?.inResponseTo;
```

With that one-line change (verified via patch-package): SP-initiated logins pass with `allowIdpInitiated: false`, and replaying the identical IdP-signed SAMLResponse is refused with "Unknown or expired request ID" — i.e. the one-shot AuthnRequest correlation works end-to-end.

`@better-auth/sso` 1.6.23 (`dist/index.mjs` — login pipeline reads `extract.inResponseTo`; logout pipeline reads `extract?.response?.inResponseTo`)

Backend, Package

### Additional context

This is the #8607 bug re-manifesting in the login pipeline: #8607 was fixed by #9055 (shipped v1.7.0-beta.1), and a comment on #8607 (2026-07-14) already notes the regression in `saml-pipeline.ts` while `sso.ts` stays fixed. Filing separately since #8607 is closed and the practical impact differs by configuration: silent no-op validation by default, or a hard SP-initiated outage with `allowIdpInitiated: false`. Happy to send a PR mirroring the logout path's read.

nexio-jurgen:
Fixes #10504.

## The bug

`samlify`'s `loginResponseFields` extractor places the `<samlp:Response>` element's attributes under the `response` key:

```ts
{ key: 'response', localPath: ['Response'], attributes: ['ID', 'IssueInstant', 'Destination', 'InResponseTo'] }
```

The login pipeline reads it from the top level instead:

```ts
// packages/sso/src/routes/saml-pipeline.ts
const inResponseTo = (extract as SAMLAssertionExtract).inResponseTo;  // always undefined
```

so the value is always `undefined`. The logout pipeline in `sso.ts` already reads the correct path (`extract?.response?.inResponseTo`), so this aligns login with logout.

## Impact

It fails differently depending on one option, and the default is the quieter of the two:

| `allowIdpInitiated` | Behaviour |
|---|---|
| `true` (default) | Every SP-initiated login takes the unsolicited branch, so **InResponseTo replay validation never runs** — while appearing configured. |
| `false` | **Every SP-initiated login is rejected** with `unsolicited_response` / "IdP-initiated SSO not allowed". |

The combination means InResponseTo binding cannot currently be enforced: leaving the default silently skips the check, and disabling IdP-initiated SSO blocks all logins.

## The change

- `saml-pipeline.ts` — read `extract.response?.inResponseTo`.
- `types.ts` — `SAMLAssertionExtract` gains the nested `response` shape samlify actually produces. The top-level `inResponseTo` is marked `@deprecated` rather than removed, to avoid a breaking change to the exported type; it was read in exactly one place, the line this PR fixes.

## Verification

Reproduced against a SAML fixture that mints genuine signed assertions and drives the real ACS:

- the assertion verifies with no signature or certificate error;
- `InResponseTo` is present on both `<samlp:Response>` and `<saml:SubjectConfirmationData>`;
- the AuthnRequest record is stored under a matching `saml-authn-request:<id>` identifier;
- the only failure is `SAML InResponseTo validation failed: ... InResponseTo missing`.

With the accessor corrected, the same assertion completes the login and establishes a session, and a replay of that assertion is then rejected — i.e. the replay reservation starts working as intended.

I have not added a regression test in this PR; `packages/sso` has no existing coverage referencing `inResponseTo`, and I did not want to guess at the fixture conventions you would prefer. Happy to add one in whatever shape you would like — a unit test asserting the extraction path, or an end-to-end assertion test.

Per #10504 this appears to be a regression: PR #9055 fixed this previously, and the fix looks to have been lost when SAML processing moved into `saml-pipeline.ts`.

---
## Summary by cubic
Fix SAML login by reading `InResponseTo` from `extract.response.inResponseTo`, matching `samlify` output and the logout path. Restores replay validation for SP-initiated logins and avoids unsolicited rejections when `allowIdpInitiated` is false.

- **Bug Fixes**
  - `packages/sso/src/routes/saml-pipeline.ts`: read `extract.response?.inResponseTo` in the login pipeline.
  - `packages/sso/src/types.ts`: add `response` attributes to `SAMLAssertionExtract`; deprecate top-level `inResponseTo`.

<sup>Written for commit 2da833fa88c09923599d2004567afd9b35f34c58. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10573?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/sso/src/routes/saml-pipeline.ts
+++ b/packages/sso/src/routes/saml-pipeline.ts
@@ -431,9 +431,11 @@ export async function processSAMLResponse(
 	}
 
 	// 12. InResponseTo validation
-	const inResponseTo = (extract as SAMLAssertionExtract).inResponseTo as
-		| string
-		| undefined;
+	// samlify nests the Response element's attributes under `response`, so the
+	// value must be read from `response.inResponseTo`. This matches the logout
+	// pipeline in `sso.ts`, which already reads the correct path.
+	const inResponseTo = (extract as SAMLAssertionExtract).response
+		?.inResponseTo as string | undefined;
 	const shouldValidateInResponseTo =
 		options?.saml?.enableInResponseToValidation !== false;
 
--- a/packages/sso/src/types.ts
+++ b/packages/sso/src/types.ts
@@ -116,7 +116,24 @@ export interface SAMLSessionRecord {
 export interface SAMLAssertionExtract {
 	nameID?: string;
 	sessionIndex?: string;
+	/**
+	 * @deprecated Never populated. samlify's `loginResponseFields` extractor
+	 * places the Response element's attributes under the `response` key, so this
+	 * top-level field is always `undefined`. Read `response.inResponseTo`
+	 * instead. Retained only to avoid a breaking change to the exported type.
+	 */
 	inResponseTo?: string;
+	/**
+	 * Attributes of the `<samlp:Response>` element, as extracted by samlify
+	 * (`{ key: 'response', localPath: ['Response'], attributes: ['ID',
+	 * 'IssueInstant', 'Destination', 'InResponseTo'] }`).
+	 */
+	response?: {
+		id?: string;
+		issueInstant?: string;
+		destination?: string;
+		inResponseTo?: string;
+	};
 	conditions?: {
 		notBefore?: string;
 		notOnOrAfter?: string;


nexio-jurgen:
Correcting something I got wrong in the description above, and adding a finding that materially affects this fix.

## Correction: this is not a regression

I repeated #10504's claim that PR #9055 fixed this and the `saml-pipeline.ts` refactor lost it. I checked the published artifacts and that is not what happened:

| version | `extract.inResponseTo` (login) | `extract?.response?.inResponseTo` |
|---|---|---|
| 1.4.0 | – | – (InResponseTo validation not present yet) |
| **1.4.8** | **present** | absent |
| 1.6.2 | present | present (**logout only**) |
| 1.6.4 | present | present |
| 1.6.25 | present | present |
| **1.7.0-rc.2** | **absent** | **present (login + logout)** |

The login accessor has been wrong since **1.4.8**, the release that introduced InResponseTo validation. The `saml-pipeline.ts` refactor landed in 1.6.3/1.6.4, roughly four months later, so it cannot have dropped a fix that was never there. The correct form appears only in the logout path from 1.5.6 onward — so #9055 most plausibly fixed **logout**, and that got generalised to login in the issue thread. Apologies for propagating it.

**1.7.0-rc.2 already contains this fix** (and additionally defaults `allowIdpInitiated` to `false`). If a 1.7.0 stable is close, this PR may be redundant — though 1.6.25 shipped a day after rc.2 without the fix, so the 1.6 line currently has no correct release.

## A further issue this fix does not address

`extract.response` is the one login field with no `shortcut: assertion`, so it is read from the raw document rather than the verified assertion node. `libsaml.verifySignature` accepts either a message-level **or** an assertion-level signature, and Microsoft Entra's default Signing Option is *Sign SAML assertion* — under which `<samlp:Response InResponseTo="…">` is covered by **no signature at all** and can be rewritten by anyone holding an otherwise valid assertion.

So with this patch applied, InResponseTo binding still rests on an unsigned attribute in the common Entra configuration.

The signature-covered copy lives at:

```
Assertion/Subject/SubjectConfirmation[@Method="urn:oasis:names:tc:SAML:2.0:cm:bearer"]
  /SubjectConfirmationData/@InResponseTo
```

which no version consults — `validateSAMLResponseBinding` checks Audience, bearer Recipient and Destination, but not SubjectConfirmationData's InResponseTo.

I have deliberately left that out of this PR to keep it to the one-line accessor alignment. Happy to extend it to prefer the SubjectConfirmationData value and fall back to the Response attribute, or to raise it separately — whichever you prefer.

nexio-jurgen:
Closing this — **1.7.0-rc.2 already contains the same fix** (`ctx.extract.response?.inResponseTo`, plus it defaults `allowIdpInitiated` to `false`), so this only duplicates it on the 1.6 line. Better left to the maintainers to decide whether the 1.6 line warrants a backport; I don't want to nudge that call from outside.

Leaving the two findings above in case they're useful:

1. **Not a regression.** The login accessor has been wrong since 1.4.8, the release that introduced InResponseTo validation — four months before the `saml-pipeline.ts` refactor. #9055 most plausibly fixed the logout path, which is where the correct form first appears (1.5.6). #10504's description says otherwise; that part is worth correcting there.

2. **`extract.response` is not signature-covered.** It has no `shortcut: assertion`, so it's read from the raw document, and `verifySignature` accepts either a message-level or assertion-level signature. Under Entra's default *Sign SAML assertion*, `<samlp:Response InResponseTo>` is unsigned and rewritable by anyone holding an otherwise valid assertion — so InResponseTo binding still rests on an unsigned attribute even with the accessor corrected, including in rc.2. The signature-covered copy is at `Assertion/Subject/SubjectConfirmation[@Method=…cm:bearer]/SubjectConfirmationData/@InResponseTo`, which no version currently consults.

The second point seems the more consequential of the two, and is independent of which release the accessor fix lands in.

Thanks for the library — the SAML pipeline is otherwise solid, and the assertion-ID reservation in particular is a nicely done piece of work.

PR #10580, opened 2026-07-29
fix(db): skip null-parsed session token in findSessions instead of returning early

Emmaccen:
skip null-parsed session token in findSessions instead of returning early

In `findSessions`, after fetching a session token from secondary storage and parsing it, there was a guard:

```ts
if (!s) return [];
```

`return []` exits the entire function immediately, throwing away any sessions already accumulated in the loop. The correct behavior is `continue`, skip the bad token and carry on, which is exactly what the `catch` block right below does for JSON syntax errors.

The scenario that triggers this: `secondaryStorage.get(token)` returns the string `"null"` (truthy, so it passes the outer `if (sessionStringified)` check), but `JSON.parse("null")` evaluates to `null`. That makes `!s` true, and the old code would silently wipe out every valid session processed so far.

In practice this means a user with multiple active sessions could lose all of them if any one session's secondary storage entry happened to contain `"null"`, whether from a misbehaving adapter, external tooling, or a cache eviction policy.

The fix is a one-character change. A regression test covering the `JSON.parse("null")` path is included alongside the existing syntax-error test.

---
## Summary by cubic
Fixes `findSessions` to skip tokens that parse to null instead of returning early, so valid sessions aren’t dropped. Users with multiple sessions now get correct results even if one token’s secondary storage value is `"null"` or malformed JSON.

- **Bug Fixes**
  - Replaces `return []` with `continue` when parsed value is `null` in `findSessions`.
  - Expands regression tests to cover both malformed JSON and `"null"` entries, ensuring other valid sessions are still returned.

<sup>Written for commit 94860d3d8727c740f5301c1c64bbd32f57f48c2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10580?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/skip-invalid-secondary-sessions.md
+++ b/.changeset/skip-invalid-secondary-sessions.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Skip invalid secondary-storage session entries without discarding other valid sessions.
--- a/packages/better-auth/src/db/internal-adapter.test.ts
+++ b/packages/better-auth/src/db/internal-adapter.test.ts
@@ -987,6 +987,54 @@ describe("internal adapter test", async () => {
 		);
 	});
 
+	it("findSessions should skip a null-parse token without discarding other sessions", async () => {
+		const testMap = new Map<string, string>();
+
+		const testOpts = {
+			database: new DatabaseSync(":memory:"),
+			secondaryStorage: {
+				set(key: string, value: string, ttl?: number) {
+					testMap.set(key, value);
+				},
+				get(key: string) {
+					return testMap.get(key) || null;
+				},
+				delete(key: string) {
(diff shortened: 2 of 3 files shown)


bytaesu:
It seems regression from #7166

bytaesu (approved):
LGTM

@Emmaccen Thanks for submitting many useful PRs lately 😁

PR #10591, opened 2026-07-30
fix(custom-session): forward session cookie cleanup when session is null

### LINKED ISSUE #10567: `customSession` drops the inner `getSession`'s Set-Cookie headers when the session is null — expired/invalid session cookies are never cleaned up
## Is this suited for github?

## Reproduction

1. Configure better-auth with the `customSession` plugin (any pass-through `fn`).
2. Sign in, then make the session invalid server-side: delete the session row (or let it expire) while the client keeps its `session_token` cookie.
3. Call `GET /api/auth/get-session` (or `auth.api.getSession({ headers })`) and inspect the response's `Set-Cookie` headers.

For comparison, repeat step 3 **without** the `customSession` plugin.

**Current:** The response contains **no** `Set-Cookie` headers — the stale `session_token` cookie stays in the browser indefinitely, and every subsequent request performs a full DB session lookup for a token that no longer exists.

Without the plugin, the core endpoint's not-found/expired branch calls `deleteSessionCookie(ctx)` and the response expires `session_token` / `session_data` as expected.

**Cause:** In the plugin's endpoint, the null-response early return sits *before* the loop that forwards the inner call's cookies:

https://github.com/better-auth/better-auth/blob/e2c73fbec87f5e19f6a2b5ac371bc5bba9bd49ff/packages/better-auth/src/plugins/custom-session/index.ts#L110-L121

```ts
if (!session?.response) {
    return ctx.json(null);       // <- inner Set-Cookie headers discarded here
}
const fnResult = await fn(session.response as any, ctx);

for (const cookieStr of session.headers.getSetCookie()) {
    // deleteSessionCookie's headers never reach this loop
```

The inner `getSession()` invocation accumulates its cookies on its own `returnHeaders` (plugin after-hooks such as `nextCookies`/`sveltekitCookies` only run on the outer dispatched endpoint), so discarding `session.headers` discards the cleanup entirely.

**Expected:** Cookie headers from the inner call should be forwarded regardless of whether the response is `null` — i.e., move the `getSetCookie()` copy loop (or a dedicated copy) above the early return. #9231 and #8127 fixed bugs inside this same loop, which suggests forwarding the inner call's cookies faithfully is the intended behavior.

1.6.23 (behavior unchanged on current `main`, e2c73fb)

## System info

```json
{
  "system": { "platform": "darwin", "arch": "arm64", "release": "25.4.0" },
  "node": { "version": "v24.15.0", "env": "development" },
  "frameworks": [
    { "name": "svelte", "version": "5.46.4" },
    { "name": "@sveltejs/kit", "version": "2.49.5" }
  ],
  "databases": [
    { "name": "drizzle", "version": "0.45.2" },
    { "name": "@planetscale/database", "version": "1.19.0" }
  ],
  "betterAuth": { "version": "1.6.23", "config": null }
}
```

(Note: the bug is environment-independent — it's visible from the plugin source.)

Plugins (`customSession`), Cookies

## Additional context

Related but distinct: #10566 covers the same endpoint converting internal lookup errors into `null` via `.catch((e) => null)`.

kdelay:
### What

`customSession()`'s `/get-session` endpoint returned early when the inner `getSession()` call produced a `null` response, which discarded the `Set-Cookie` headers that call had accumulated on its own `returnHeaders`.

The core endpoint calls `deleteSessionCookie(ctx)` on the not-found/expired branch, so without the plugin an invalid session cookie is expired on the response. With the plugin the response carried no `Set-Cookie` at all, so the stale `session_token` / `session_data` cookies stayed in the browser indefinitely and every later request performed a full session lookup for a token that no longer exists.

This moves the cookie/header forwarding above the null check, so the inner call's cleanup reaches the response regardless of whether a session was found. The forwarding logic itself is unchanged, and `fn()` is still only invoked when there is a session.

### Verification

Added a regression test in `packages/better-auth/src/plugins/custom-session/custom-session.test.ts` that signs in, deletes the backing session row while the client keeps its cookies, then calls `get-session` through the plugin and asserts the response expires `better-auth.session_token` (`value: ""`, `max-age: 0`).

```
npx vitest run packages/better-auth/src/plugins/custom-session/custom-session.test.ts
```

- Before the fix: the new test fails (no `Set-Cookie` for `better-auth.session_token`).
- After the fix: 13 passed.
- `npx vitest run packages/better-auth/src/plugins` -> 1220 passed; the only failure is `oauth-proxy` "database mode + UUID", which needs the Postgres container (`ECONNREFUSED 127.0.0.1:5432`) and is unrelated to this change.
- `pnpm typecheck` passes.

Fixes #10567

---
## Summary by cubic
Fix `customSession()` so expired or revoked sessions clear cookies by forwarding `Set-Cookie` and headers from the inner `getSession()` before the null check. This prevents stale `better-auth.session_token`/`session_data` cookies and avoids extra session lookups.

- **Bug Fixes**
  - Forward cookies/headers before the null check (only when present) so `deleteSessionCookie` reaches the response; behavior with valid sessions is unchanged.
  - Added a regression test that deletes the DB session and asserts `better-auth.session_token` is expired (`value: ""`, `max-age: 0`).

<sup>Written for commit a0c8cf7329101e68397537817cf9e9e421b80985. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10591?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/custom-session-expired-cookie-cleanup.md
+++ b/.changeset/custom-session-expired-cookie-cleanup.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Fix `customSession()` leaving stale session cookies in the browser. When the session is expired or no longer exists, `get-session` now expires the `session_token` / `session_data` cookies as it does without the plugin, instead of returning `null` with no `Set-Cookie` headers.
--- a/packages/better-auth/src/plugins/custom-session/custom-session.test.ts
+++ b/packages/better-auth/src/plugins/custom-session/custom-session.test.ts
@@ -312,6 +312,63 @@ describe("Custom Session Plugin Tests", async () => {
 		});
 	});
 
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/10567
+	 */
+	it("should expire the session cookies when the session is no longer valid", async () => {
+		const staleOptions = {
+			plugins: [
+				customSession(async ({ user, session }) => {
+					return { user, session };
+				}),
+			],
+		} satisfies BetterAuthOptions;
+
+		const {
+			auth: staleAuth,
+			client: staleClient,
+			signInWithTestUser: signInWithStaleUser,
+			db,
+		} = await getTestInstance(staleOptions, {
+			clientOptions: {
+				plugins: [customSessionClient<{ options: typeof staleOptions }>()],
+			},
+		});
+
+		const { headers } = await signInWithStaleUser();
+		const activeSession = await staleAuth.api.getSession({ headers });
+		const sessionToken = activeSession?.session.token;
+		if (!sessionToken) throw new Error("expected an active session");
+
+		// Revoke the backing session row while the client keeps its cookies.
+		await db.delete({
+			model: "session",
+			where: [{ field: "token", value: sessionToken }],
+		});
+
+		let setCookies: string[] = [];
+		const session = await staleClient.getSession({
+			fetchOptions: {
+				headers,
+				onResponse(context) {
+					setCookies = context.response.headers.getSetCookie();
+				},
+			},
+		});
+
+		expect(session.data).toBeNull();
+
+		const parsedCookies = new Map(
+			setCookies.flatMap((cookieString) =>
+				Array.from(parseSetCookieHeader(cookieString).entries()),
+			),
+		);
+		const expiredSessionToken = parsedCookies.get("better-auth.session_token");
+		expect(expiredSessionToken).toBeDefined();
+		expect(expiredSessionToken?.value).toBe("");
+		expect(expiredSessionToken?.["max-age"]).toBe(0);
+	});
+
 	it.skipIf(globalThis.gc == null)(
 		"should not create memory leaks with multiple plugin instances",
 		async () => {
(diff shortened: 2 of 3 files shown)


kdelay:
Rebased onto current `main` (the branch had drifted 25 commits behind) and re-ran the gates on today's tree:

- `vitest run src/plugins/custom-session/custom-session.test.ts` — 13/13 pass
- `vitest run src/plugins` — 1225 pass, 2 skipped, 1 failure: `oauth-proxy > passthrough mode > should work with database mode + UUID`, which fails with `password authentication failed for user "user"` because the compose Postgres service is not up locally
- `pnpm typecheck` clean, `biome check` clean on the three touched files
- Reverting only `custom-session/index.ts` back to `main` and keeping the test still fails exactly one test (`should expire the session cookies when the session is no longer valid`), so the guard is still tied to this change on the current base

One scope note for merge ordering: #10671 edits the same block in `custom-session/index.ts` (it removes the `.catch` around the inner `getSession()` call). The two are independent — applying its diff onto this branch with `git apply -3` merges cleanly, and with only its source change on top of `main` the test added here still fails, so that PR does not cover this path. Either merge order works.

PR #10594, opened 2026-07-30
Fix checkOrgaizationSlug return handling

Schmell:
This function is throwing when a match is found which is unintuitive a it would return {status:true} without a match. Not sure if the try catch is necessary?

---
## Summary by cubic
Fixes checkOrganizationSlug to return a boolean status instead of throwing when a slug already exists, making the API consistent and easier for clients to consume.

- **Bug Fixes**
  - Return `{ status: true }` when the slug is available; `{ status: false }` when it’s taken.
  - Add try/catch around `findOrganizationBySlug`; throw a `BAD_REQUEST` with a generic message only on unexpected adapter errors.

<sup>Written for commit f66d93aeb1629b3a9f504994fc7c8f59b480977f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/better-auth/src/plugins/organization/routes/crud-org.ts
+++ b/packages/better-auth/src/plugins/organization/routes/crud-org.ts
@@ -320,16 +320,27 @@ export const checkOrganizationSlug = <O extends OrganizationOptions>(
 		},
 		async (ctx) => {
 			const orgAdapter = getOrgAdapter<O>(ctx.context, options);
-			const org = await orgAdapter.findOrganizationBySlug(ctx.body.slug);
-			if (!org) {
-				return ctx.json({
-					status: true,
-				});
-			}
+			try{
+				const org = await orgAdapter.findOrganizationBySlug(ctx.body.slug);
+				if (!org) {
+					return ctx.json({
+						status: true,
+					});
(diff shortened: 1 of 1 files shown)


bytaesu (changes requested):
This changes the existing API contract, so it would be difficult to ship in a patch release. Currently, when a slug is already taken, `authClient.organization.checkSlug` reports it through the `error` field of `{ data, error }`. With this change, the same case would instead be returned through `data` as `{ status: false }`. Server-side callers would also change from receiving an APIError to receiving a successful response.

bytaesu:
Please check:
- https://better-auth.com/docs/concepts/api#error-handling
- https://better-auth.com/docs/concepts/client#handling-errors

PR #10599, opened 2026-07-30
fix(oauth-proxy): preserve Apple user data

### LINKED ISSUE #10598: oAuthProxy drops Apple’s one-time user/name payload during form_post callback
## Is this suited for GitHub?

Yes. This appears to be a bug in the `oAuthProxy` callback path and can permanently discard profile data that Apple only returns once.

## Describe the bug

When Sign in with Apple is completed through `oAuthProxy`, authentication succeeds but the user’s name is lost.

Apple returns the user’s name in the `user` form field during the first authorization only. The name is not included in the identity token.

The regular Better Auth OAuth callback handles this correctly by parsing the posted `user` value and passing it to the provider in [`packages/better-auth/src/api/routes/callback.ts`](https://github.com/better-auth/better-auth/blob/v1.6.25/packages/better-auth/src/api/routes/callback.ts#L131-L150):

```ts
const parsedUserData = userData ? safeJSONParse(userData) : null

const userInfo = await provider.getUserInfo({
  ...tokens,
  user: parsedUserData ?? undefined,
})
```

The `oAuthProxy` callback path merges `ctx.query` and `ctx.body`, but its callback schema in [`packages/better-auth/src/plugins/oauth-proxy/index.ts`](https://github.com/better-auth/better-auth/blob/v1.6.25/packages/better-auth/src/plugins/oauth-proxy/index.ts#L135-L138) only retains `code` and `error`:

```ts
const oauthCallbackQuerySchema = z.object({
  code: z.string().optional(),
  error: z.string().optional(),
})
```

It subsequently calls `getUserInfo` without the posted Apple user data in [`packages/better-auth/src/plugins/oauth-proxy/index.ts`](https://github.com/better-auth/better-auth/blob/v1.6.25/packages/better-auth/src/plugins/oauth-proxy/index.ts#L473-L475):

```ts
const userInfoResult = await provider.getUserInfo(tokens)
```

Therefore, Apple’s posted `user` value is discarded before the Apple provider can read it.

The Apple provider then falls back to `profile.name || ""` in [`packages/core/src/social-providers/apple.ts`](https://github.com/better-auth/better-auth/blob/v1.6.25/packages/core/src/social-providers/apple.ts#L196-L206). Because the identity token does not contain the name, new users are created with an empty name.

## Current behavior

1. A new user authorizes Sign in with Apple for the first time.
2. Apple sends a `form_post` callback containing:

```text
code
state
id_token
user
```

3. The `user` field contains:

```json
{
  "name": {
    "firstName": "Jane",
    "lastName": "Doe"
  },
  "email": "..."
}
```

4. `oAuthProxy` successfully exchanges the authorization code.
5. `provider.getUserInfo(tokens)` is called without the posted `user` data.
6. `mapProfileToUser` receives an Apple profile whose `name` is empty.
7. Better Auth creates the user with `name: ""`.

The authentication itself succeeds, so the loss of profile data is silent.

## Expected behavior

The `oAuthProxy` path should preserve Apple’s posted `user` field and pass it to the provider in the same way as the regular OAuth callback path.

For a first-time Apple authorization, the created user should contain the name supplied by Apple.

## Reproduction

Configure Apple with `oAuthProxy`:

```ts
betterAuth({
  plugins: [
    oAuthProxy({
      productionURL: "https://auth.example.com",
    }),
  ],
  socialProviders: {
    apple: {
      clientId: process.env.APPLE_CLIENT_ID!,
      clientSecret: process.env.APPLE_CLIENT_SECRET!,
      scope: ["email", "name"],
      mapProfileToUser(profile) {
        console.log("Apple profile name:", profile.name)

        return {}
      },
    },
  },
})
```

Initiate authentication from an origin different from `productionURL` so the OAuth proxy flow is used.

Authorize the application with an Apple account that has not previously authorized it.

Inspect the callback:

```http
POST /api/auth/callback/apple
Content-Type: application/x-www-form-urlencoded

code=...&state=...&id_token=...&user={"name":{"firstName":"Jane","lastName":"Doe"},"email":"..."}
```

Authentication completes, but the logged profile name and persisted user name are empty.

Because Apple only returns `user` during the first authorization, repeating the login will not reproduce the original payload unless access to the application is revoked from the Apple account first.

## Suggested fix

Update [`packages/better-auth/src/plugins/oauth-proxy/index.ts`](https://github.com/better-auth/better-auth/blob/v1.6.25/packages/better-auth/src/plugins/oauth-proxy/index.ts#L135-L138) to include `user` in the proxy callback schema:

```ts
const oauthCallbackQuerySchema = z.object({
  code: z.string().optional(),
  error: z.string().optional(),
  user: z.string().optional(),
})
```

Then, in [`packages/better-auth/src/plugins/oauth-proxy/index.ts`](https://github.com/better-auth/better-auth/blob/v1.6.25/packages/better-auth/src/plugins/oauth-proxy/index.ts#L391-L474), parse and pass it to `getUserInfo`, matching the regular callback implementation:

```ts
const { code, error, user: userData } = query.data
const parsedUserData = userData ? safeJSONParse(userData) : null

const userInfoResult = await provider.getUserInfo({
  ...tokens,
  user: parsedUserData ?? undefined,
})
```

A regression test should simulate an Apple-style `form_post` containing a `user` field and verify that the resulting proxied profile retains the name.

## Related issue

This appears to be the remaining portion of:

- #8889
- #8895

Issue #8889 noted that it might be necessary to preserve Apple’s posted `user` payload. PR #8895 fixed reading `code` from the POST body, but its test did not include `user`, and the payload is still not forwarded to the provider.

## Better Auth versions

Confirmed in:

- `better-auth@1.6.10`
- `better-auth@1.6.25`
- `better-auth@1.7.0-rc.2`

## Additional context

Apple documents that:

- The full name is passed directly in the authorization response.
- It is not included in the identity token.
- The `user` object is returned only the first time the user authorizes the application.

Consequently, silently discarding this field cannot be repaired on a subsequent login.

bytaesu:
Closes #10598


bytaesu:
I reverted https://github.com/better-auth/better-auth/pull/10599/commits/f40be81ddbf2e3c2ce298451af40003faf2ae718

PR #10605, opened 2026-07-31
fix(email-otp): verify OTP before revealing whether the email exists

### LINKED ISSUE #10603: "email-otp/check-verification-otp" leaks account existence before the OTP is verified

### Reproduction

```ts
  emailAndPassword: {
    enabled: true,
  },
  plugins: [
    emailOTP({
      disableSignUp: true,
      overrideDefaultEmailVerification: true,
      async sendVerificationOTP({ email, otp, type }) { /* ... */ },
    }),
  ],
});
```

1. Sign up `known@example.com` (no verification needed, just needs to exist).
2. Unauthenticated, POST `/email-otp/check-verification-otp` with `{ email: "known@example.com", type: "email-verification", otp: "000000" }` → `400 INVALID_OTP`.
3. Same request with `{ email: "unknown@example.com", ... }` → `400 USER_NOT_FOUND`.

Two different error codes, no OTP required, no session required. That's an enumeration oracle.

The handler looks the user up and throws before it ever touches the OTP ([[routes.ts#L373-L376](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/email-otp/routes.ts#L373-L376)](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/email-otp/routes.ts#L373-L376)):

```ts
const user = await ctx.context.internalAdapter.findUserByEmail(email);
if (!user) {
  throw APIError.from("BAD_REQUEST", BASE_ERROR_CODES.USER_NOT_FOUND);
}
const identifier = toOTPIdentifier(ctx.body.type, email);
const verificationValue = await ctx.context.internalAdapter.findVerificationValue(identifier);
```

The other three OTP endpoints do it the other way round — verify first, then look up the user — and two of them say why in a comment:

- `verify-email` ([[L489-L503](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/email-otp/routes.ts#L489-L503)](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/email-otp/routes.ts#L489-L503)): *"safe to leak the existence of a user, given the user has already the OTP from the email"*
- `reset-password` ([[L946-L958](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/email-otp/routes.ts#L946-L958)](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/email-otp/routes.ts#L946-L958))
- `change-email` ([[~L1220-L1230](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/email-otp/routes.ts#L1220-L1230)](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/plugins/email-otp/routes.ts#L1220-L1230)): *"safe to leak the existence of a user as a valid OTP has been provided"*

So the rule is already established: leaking is fine *once possession of the OTP is proven*. `check-verification-otp` leaks with nothing proven. It looks like an ordering oversight rather than an intentional exception — it's the only one of the four with no comment.

I expected it to return `INVALID_OTP` for an unknown email, same as it does for a known email with a wrong code.

Why this matters in this particular config: with `requireEmailVerification: true`, everything else is already closed. `/sign-up/email` returns a uniform 200 with a dummy password hash, `/sign-in/email` returns `INVALID_EMAIL_OR_PASSWORD` on all failure branches, and `/email-otp/send-verification-otp` creates and deletes the OTP row so unknown emails still get `{ success: true }` (there's even a test for it, "should prevent user enumeration when disableSignUp is enabled"). `check-verification-otp` is the one path that undoes all of that.

Rate limiting doesn't blunt it either. `getDefaultSpecialRules()` ([[rate-limiter/index.ts#L513-L541](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/api/rate-limiter/index.ts#L513-L541)](https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/api/rate-limiter/index.ts#L513-L541)) covers `/sign-in*`, `/sign-up*`, `/change-*` at 3/10s and the send-OTP paths at 3/60s. `/email-otp/check-verification-otp` matches neither, so it falls back to the global limit — enough to probe a list in bulk.

Related, same class, both fixed: #7972 (`sign-up/email` leaked via 422) and #8096 (`/change-email`).

**Suggested fix:** move the `findUserByEmail` / `USER_NOT_FOUND` block below the OTP check, so it matches the three sibling endpoints. Happy to send a PR if you'd like.

**Workaround** for anyone hitting this before it's patched — a `before` hook that normalises the response:

```ts
hooks: {
    if (ctx.path !== "/email-otp/check-verification-otp") return;
    const email = ctx.body?.email;
    if (typeof email !== "string") return; // let the endpoint's own validation run
    if (await ctx.context.internalAdapter.findUserByEmail(email)) return;
    throw new APIError("BAD_REQUEST", { code: "INVALID_OTP", message: "Invalid OTP" });
  }),
}
```

1.6.25

### System info

```bash
platform: linux
arch: x64

Package Manager:
  name: bun
  version: 1.3.14

Frameworks:
  name: react
  version: ^19.2.8

DB: drizzle-adapter
```

Backend

```typescript
import { emailOTP } from "better-auth/plugins/email-otp";

  emailAndPassword: {
    enabled: true,
  },
  plugins: [
    emailOTP({
      disableSignUp: true,
      overrideDefaultEmailVerification: true,
        // send mail
      },
    }),
  ],
});
```

### Additional context

Reproduced against 1.6.25 and confirmed the ordering is still the same on `main`. The docs' [[Email Enumeration Protection](https://www.better-auth.com/docs/reference/security#email-enumeration-protection)](https://www.better-auth.com/docs/reference/security#email-enumeration-protection) section only claims coverage for sign-up and change-email, so I don't think the OTP plugin was ever audited for this — flagging it rather than assuming it's intended.

XXMOHAMED012:
Fixes #10603.

 /email-otp/check-verification-otp called findUserByEmail and threw USER_NOT_FOUND before any OTP validation. Unauthenticated, with any junk code, a known email returned 400 INVALID_OTP and an unknown one 400 USER_NOT_FOUND — enough to tell them apart.
The other three OTP endpoints already do it the other way round. verify-email, reset-password and change-email all verify the OTP first and only look the user up afterward, two of them with a comment saying that's the point: "safe to leak the existence of a user, given the user has already the OTP from the email". This moves check-verification-otp onto the same ordering and reuses that comment.

 No behaviour change for legitimate callers: a valid OTP for a missing user still returns USER_NOT_FOUND, and the wrong-code path still increments the attempt counter before failing. The only difference is that the account-existence signal now comes after possession of the code is proven.

Nothing in the docs specifies error codes for this endpoint — the checkVerificationOtp block documents the request shape only — so no docs change.

Test

Added to the existing email-otp-verify describe block, next to "should prevent user enumeration when disableSignUp is enabled". It asserts an unregistered email returns the same status and code as a registered one with a bad OTP. Fails on main with expected 'USER_NOT_FOUND' to be 'INVALID_OTP', passes with the fix.

vitest run packages/better-auth/src/plugins/email-otp/   # 88 passed
pnpm typecheck                                            # clean

---
## Summary by cubic
Verify OTP before revealing whether an email exists in `/email-otp/check-verification-otp` to prevent user enumeration and match other email OTP endpoints. Fixes #10603.

- **Bug Fixes**
  - Moved user lookup after OTP validation; unknown emails now return 400 INVALID_OTP for bad codes.
  - Kept behavior for valid OTP with missing user (still 400 USER_NOT_FOUND); wrong codes still increment attempts.
  - Added a test asserting parity: both registered and unregistered emails return 400 INVALID_OTP with a bad OTP (with disableSignUp enabled).

- **Dependencies**
  - Added changeset for `better-auth` patch release.

<sup>Written for commit 414f80c50498f4bde9fe0dff0ad20279e7816983. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10605?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bytaesu:
**_Email OTP verification endpoints_**

```
check-verification-otp : USER lookup -> OTP verify (exception)

verify-email           : OTP verify -> USER lookup
sign-in/email-otp      : OTP verify -> USER lookup
reset-password         : OTP verify -> USER lookup
change-email           : OTP verify -> USER lookup
```

bytaesu (dismissed):
LGTM

Thanks @XXMOHAMED012

bytaesu (approved):

PR #10694, opened 2026-08-05
fix(oauth): set account cookie after implicit account linking

### LINKED ISSUE #10690: storeAccountCookie: account_data cookie is not written when a new provider account is linked to an existing user.

## Describe the bug

In stateless mode (`account.storeAccountCookie: true`, no database), `handleOAuthUserInfo` has three outcomes, and only two of them write the `account_data` cookie:

1. **New user** → `createOAuthUser(...)` → `setAccountCookie(c, createdAccount)` ✅
2. **Existing user, already-linked account** → `setAccountCookie(c, { ...linkedAccount, ...freshTokens })` ✅
3. **Existing user (matched by email), NEW provider** → `internalAdapter.linkAccount(...)` only — `setAccountCookie` is never called ❌

https://github.com/better-auth/better-auth/blob/main/packages/better-auth/src/oauth2/link-account.ts (the `linkAccount` call inside the `!linkedAccount` branch)

After a path-3 sign-in the browser has a valid session but **no `account_data` cookie**. Every subsequent `getAccessToken({ providerId })` then fails: the account-cookie lookup misses, the `internalAdapter.findAccounts` fallback is empty in DB-less deployments (memory adapter, other replica/restart), and the call throws `ACCOUNT_NOT_FOUND` — a permanently degraded session that looks logged-in but can never get a provider token.

## How we hit it

Multi-tenant deployment: one stateless better-auth instance serves many brand domains, with one `genericOAuth` provider per domain (same IdP, different client ids). A user logs into domain A (provider A, path 1). They then log into domain B (provider B): the shared in-process memory adapter still holds the user from the domain-A login, `findOAuthUser` matches it **by email**, and the sign-in takes path 3. Domain B ends up with a session cookie but no account cookie → `ACCOUNT_NOT_FOUND` on every token fetch. It reproduces flakily in production because it depends on which replica's memory served the callback.

## To reproduce

1. `betterAuth({ secret, account: { storeAccountCookie: true }, plugins: [genericOAuth({ config: [providerA, providerB] })] })` — no database. Both providers return the same email.
2. Complete a sign-in with provider A (fresh cookies).
3. Complete a sign-in with provider B **on the same server instance** (no cookies needed — the memory adapter carries the user).
4. Inspect the callback response of step 3: `session_token`/`session_data` are set, `account_data` is not.
5. `POST /get-access-token { providerId: providerB }` with step-3 cookies → `ACCOUNT_NOT_FOUND`.

## Proposed solution

Write the account cookie in the link branch, exactly as the other two branches do — `linkAccount` already returns the created account record:

```ts
const newLinkedAccount = await c.context.internalAdapter.linkAccount({
	providerId: account.providerId,
	accountId: userInfo.id.toString(),
	userId: dbUser.user.id,
	accessToken: await setTokenUtil(account.accessToken, c.context),
	refreshToken: await setTokenUtil(account.refreshToken, c.context),
	idToken: account.idToken,
	accessTokenExpiresAt: account.accessTokenExpiresAt,
	refreshTokenExpiresAt: account.refreshTokenExpiresAt,
	scope: account.scope,
});
if (c.context.options.account?.storeAccountCookie) {
	await setAccountCookie(c, newLinkedAccount);
}
```

We are running this as a local patch on 1.6.25 and it resolves the issue; a full sign-in → callback → `getAccessToken` regression test over two same-email providers goes red without the change and green with it.

## Related history

This is the last of the three `handleOAuthUserInfo` outcomes to get the cookie write — the siblings were each fixed separately: #7217 (re-login with `updateAccountOnSignIn: false`), #9375 / #9470 (first-time login). Path 3 was missed both times.

**Version:** better-auth 1.6.25 (also present on current `main`)

Current: No account cookie exists after login

Expected: Account cookie to exist with fresh tokens

1.6.25

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:40 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6041",
    "release": "24.6.0",
    "cpuCount": 14,
    "cpuModel": "Apple M4 Pro",
    "totalMemory": "48.00 GB",
    "freeMemory": "0.15 GB"
  },
  "node": {
    "version": "v24.14.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.1"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null
  }
}
```

Package

```typescript
  emailAndPassword: {  
    enabled: true
  },
});
```

### Additional context

_No response_

Vishesh-Verma-07:
## Summary

This PR fixes a bug where `storeAccountCookie` was not honored after implicit OAuth account linking.

When an existing user signs in with a new OAuth provider using the same email address, Better Auth correctly links the provider to the existing account. However, unlike the new-user and already-linked flows, the implicit linking path did not call `setAccountCookie()`.

As a result, `getAccessToken()` could not resolve the newly linked provider in stateless (`storeAccountCookie`) deployments.

This PR:
- Stores the `linkAccount()` result.
- Sets the account cookie after successful implicit account linking when `storeAccountCookie` is enabled.
- Adds a regression test covering this flow.

## Fixes

Closes #10690

## Testing

- Added a regression test covering:
  - First sign-in with Provider A.
  - Implicit account linking with Provider B using the same email.
  - Account cookie creation.
  - Successful `getAccessToken()` for the newly linked provider.

## Breaking Changes

None.

---
## Summary by cubic
Fixes missing account cookie after implicit OAuth account linking when `storeAccountCookie` is enabled. Ensures `getAccessToken()` works for newly linked providers in stateless setups in `better-auth`.

- **Bug Fixes**
  - Save `linkAccount()` result and call `setAccountCookie()` after implicit linking when `storeAccountCookie` is true.
  - Add regression test covering provider A sign-in, implicit linking with provider B (same email), cookie creation, and successful `getAccessToken()` for provider B.

<sup>Written for commit af1e7f75fe94e313ff9201bd7c3bdd1285b50f52. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10694?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/honest-readers-shake.md
+++ b/.changeset/honest-readers-shake.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Fix account cookie creation after implicit OAuth account linking when storeAccountCookie is enabled.
--- a/packages/better-auth/src/oauth2/link-account.ts
+++ b/packages/better-auth/src/oauth2/link-account.ts
@@ -38,6 +38,7 @@ export async function handleOAuthUserInfo(
 		});
 	let user = dbUser?.user;
 	const isRegister = !user;
+	let newLinkedAccount: Account;
 
 	if (dbUser) {
 		const linkedAccount =
@@ -73,7 +74,7 @@ export async function handleOAuthUserInfo(
 				};
 			}
 			try {
-				await c.context.internalAdapter.linkAccount({
+				newLinkedAccount = await c.context.internalAdapter.linkAccount({
 					providerId: account.providerId,
 					accountId: userInfo.id.toString(),
 					userId: dbUser.user.id,
@@ -90,6 +91,9 @@ export async function handleOAuthUserInfo(
 					error: "unable to link account",
 					data: null,
 				};
+			}
+			if (c.context.options.account?.storeAccountCookie) {
+				await setAccountCookie(c, newLinkedAccount);
 			}
 
 			// Reachable only when `requireLocalEmailVerified: false` lets the link
--- a/packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
+++ b/packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts
@@ -318,6 +318,124 @@ describe("oauth2", async () => {
 		expect(accessTokenRes.data?.accessToken).toBeTruthy();
 	});
 
+	it("should resolve getAccessToken after implicit account linking (storeAccountCookie)", async () => {
+		const { customFetchImpl, auth } = await getTestInstance({
+			account: {
+				storeAccountCookie: true,
+			},
+			plugins: [
+				genericOAuth({
+					config: [
+						{
+							providerId: "test-implicit-link-a",
+							discoveryUrl: `http://localhost:${port}/.well-known/openid-configuration`,
+							clientId,
+							clientSecret,
+							pkce: true,
+						},
+						{
+							providerId: "test-implicit-link-b",
+							discoveryUrl: `http://localhost:${port}/.well-known/openid-configuration`,
+							clientId,
+							clientSecret,
+							pkce: true,
+						},
+					],
+				}),
+			],
+		});
+
+		const ctx = await auth.$context;
+		const accountDataCookieName = ctx.authCookies.accountData.name;
+
+		const newAuthClient = createAuthClient({
+			plugins: [genericOAuthClient()],
+			baseURL: "http://localhost:3000",
+			fetchOptions: {
+				customFetchImpl,
+			},
+		});
+
+		const sharedEmail = "implicit-link-cookie@test.com";
+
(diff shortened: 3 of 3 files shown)


bytaesu:
This issue also appears to have already been fixed in the latest version, so I'm closing this PR.

PR #10730, opened 2026-08-10
fix(cimd): answer the socket's all-addresses lookup in the Node transport

### LINKED ISSUE #10810: @better-auth/cimd/node: fetchClientMetadataResource fails on Node >= 20 (ERR_INVALID_IP_ADDRESS)
`fetchClientMetadataResource` pins the resolved address by passing `https.request` a custom `lookup` that calls back with a **single** address:

```js
lookup: (_hostname, _options, callback) => {
  callback(null, pinnedAddress.address, pinnedAddress.family);
}
```

Node 20 enabled `autoSelectFamily` by default, so `net.connect` requests `all: true` from the lookup and expects an **array**. It receives a string, indexes `[0]`, and reads `.address` off a single character — `undefined`.

**Every CIMD metadata fetch therefore fails on Node ≥ 20 with default flags.** Not IPv6-specific and not dual-stack-specific: any host, any address family. Callers see `invalid_client` / `"Failed to fetch metadata document (network error or redirect blocked)"`, which points the operator at their document hosting rather than at the transport — it cost us a while to find.

## Repro

The mechanism, isolated (Node v20.19.5) — note `all: true` in the options Node passes:

```js
require("node:https").request("https://example.com/", {
  lookup: (h, o, cb) => { console.log("opts:", JSON.stringify(o)); cb(null, "93.184.215.14", 4); },
}, r => console.log("OK", r.statusCode))
  .once("error", e => console.log("FAILED:", e.message, e.code))
  .end();

// opts: {"hints":32,"all":true}
// FAILED: Invalid IP address: undefined ERR_INVALID_IP_ADDRESS
```

Through the package (`@better-auth/cimd@1.7.0-rc.5`), against a real public metadata document:

```
$ node -e '... await fetchClientMetadataResource(url, { method: "GET" }) ...'
FAILED: TypeError Invalid IP address: undefined ERR_INVALID_IP_ADDRESS

$ node --no-network-family-autoselection -e '... same call ...'
OK 200 application/json; charset=utf-8
```

## Suggested fix

Any of:

- set `autoSelectFamily: false` on the `https.request` options, or
- set `family: pinnedAddress.family` so Node doesn't ask for a list, or
- honour `options.all` in the lookup callback and return `[{ address, family }]`.

## Why this may not have been reported yet

`fetchClientMetadataResource` is a required option with no default, so anyone who passed a plain `fetch` instead has a working flow — the only users currently succeeding at CIMD are the ones who skipped the SSRF protection this transport exists to provide. Our own test suite missed it too: our rejection cases fail on URL validation before a socket is opened, so only a real registration against a real document surfaces it.

## Workaround

Run the server with `NODE_OPTIONS=--no-network-family-autoselection`, at the cost of IPv4/IPv6 fallback on all outbound connections.

## Environment

- `@better-auth/cimd@1.7.0-rc.5`, `@better-auth/oauth-provider@1.7.0-rc.5`, `better-auth@1.7.0-rc.5`
- Node v20.19.5 (Linux). `autoSelectFamily` defaults to true on Node 20 and 22.

erikpr1994:
Closes #10810

## The bug

`fetchClientMetadataResource` from `@better-auth/cimd/node` fails on every call on Node >= 20. The transport pins the resolved DNS answer through a custom `lookup` override on `https.request`, and the override always answers in the legacy three-argument form:

```ts
lookup: (_hostname, _options, callback) => {
  callback(null, pinnedAddress.address, pinnedAddress.family);
},
```

Since Node 20, `net.autoSelectFamily` defaults to `true`, so the socket calls the override with `{ all: true }` and expects an address **list** back. The legacy answer fails the connection with `ERR_INVALID_IP_ADDRESS: Invalid IP address: undefined` before it opens. The CIMD plugin catches the transport error and rewrites it, so every authorization against a server wired with this transport answers:

```
{"error":"invalid_client","error_description":"Failed to fetch metadata document (network error or redirect blocked)"}
```

No CIMD client can ever register — observed in production on Vercel (Node 22) with Claude as the client, and reproducible with ten lines: any `https.request` with a three-argument custom `lookup` on stock Node >= 20 fails the same way, and succeeds with `autoSelectFamily: false`.

## The fix

The override answers the `all: true` question with the pinned address as a one-entry list, and keeps the legacy answer for the legacy question. The pin, Host header, TLS SNI, `agent: false`, and no-redirect behavior are unchanged.

The new test drives the captured override with `{ all: true }` and asserts the list answer — it fails on the old transport with the exact production symptom. All 8 transport tests pass.

---
## Summary by cubic
Fixes DNS lookup handling in the Node transport for `@better-auth/cimd/node` so `fetchClientMetadataResource` works on Node 20+ (where `net.autoSelectFamily` defaults to true), preventing connection failures and restoring client registration.

- **Bug Fixes**
  - The `lookup` override returns a one-item address list when called with `{ all: true }` and keeps the legacy three-arg response otherwise; pinning, `Host`, SNI, `agent: false`, and no-redirect behavior are unchanged.
  - Adds a test that reproduces the failure and updates the test mocks to exercise the all-addresses callback path.

<sup>Written for commit 8547bf8994d96f086e509454e8b664a969a58195. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10730?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/fix-cimd-node-lookup.md
+++ b/.changeset/fix-cimd-node-lookup.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/cimd": patch
+---
+
+Fix CIMD client metadata discovery failing with `ERR_INVALID_IP_ADDRESS` on supported Node.js versions.
--- a/packages/cimd/src/node.test.ts
+++ b/packages/cimd/src/node.test.ts
@@ -150,6 +150,38 @@ describe("Node CIMD metadata transport", () => {
 		expect(pinnedAddress).toBe("93.184.216.34");
 	});
 
+	it("answers an all-addresses lookup with the pinned address list", async () => {
+		// Since Node 20, `net.autoSelectFamily` defaults to true and the socket
+		// asks the lookup override with `all: true`, expecting an address list.
+		// The legacy three-argument answer fails every connection with
+		// ERR_INVALID_IP_ADDRESS before it opens.
+		mocks.lookup.mockResolvedValue([
+			{ address: "93.184.216.34", family: 4 },
+			{ address: "2606:2800:220:1:248:1893:25c8:1946", family: 6 },
+		]);
+		mockHttpsResponse({ body: "ok" });
+
+		await fetchClientMetadataResource(
+			"https://client.example.com/client.json",
+		);
+
+		const options = mocks.request.mock.calls[0]?.[1] as {
+			lookup: (
+				hostname: string,
+				options: { all?: boolean },
+				callback: (
+					error: Error | null,
+					addresses: { address: string; family: number }[],
+				) => void,
+			) => void;
+		};
+		const answered = vi.fn();
+		options.lookup("client.example.com", { all: true }, answered);
+		expect(answered).toHaveBeenCalledWith(null, [
+			{ address: "93.184.216.34", family: 4 },
+		]);
+	});
+
 	it("returns redirect responses without following them", async () => {
 		mocks.lookup.mockResolvedValue([{ address: "93.184.216.34", family: 4 }]);
 		mockHttpsResponse({
--- a/packages/cimd/src/node.test.ts
+++ b/packages/cimd/src/node.test.ts
@@ -1,20 +1,26 @@
 import { EventEmitter } from "node:events";
 import { Readable } from "node:stream";
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 
 const mocks = vi.hoisted(() => ({
 	lookup: vi.fn(),
 	request: vi.fn(),
 }));
 
-vi.mock("node:dns/promises", () => ({ lookup: mocks.lookup }));
-vi.mock("node:https", () => ({ request: mocks.request }));
+vi.mock(import("node:dns/promises"), () => ({ lookup: mocks.lookup }));
+vi.mock(import("node:https"), () => ({ request: mocks.request }));
 
 import { fetchClientMetadataResource } from "./node";
+
+interface MockLookupAddress {
+	address: string;
+	family: number;
+}
 
 interface MockResponseOptions {
 	body?: string;
 	headers?: Record<string, string>;
+	lookupOptions?: { all?: boolean };
 	status?: number;
 }
 
@@ -25,11 +31,11 @@ function mockHttpsResponse(options: MockResponseOptions = {}) {
 			requestOptions: {
 				lookup: (
 					hostname: string,
-					options: object,
+					lookupOptions: { all?: boolean },
 					callback: (
 						error: Error | null,
-						address: string,
-						family: number,
+						result: string | MockLookupAddress[],
+						family?: number,
 					) => void,
 				) => void;
 			},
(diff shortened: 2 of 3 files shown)


onmax:
this works in my project. thank :)

jodok:
Independent confirmation of the bug this fixes, plus one residual the current shape leaves behind.

**Confirmation.** We hit this in production on Node **v26.8.1** with `@better-auth/cimd@1.7.1`: every CIMD client registration failed, for every `client_id` host. `fetchClientMetadataResource` threw `TypeError: Invalid IP address: undefined` against `example.com`, `claude.ai` and our own domain alike, while a plain `fetch()` to the same URL from the same container returned 200 — which is what makes this so hard to diagnose (more on that below).

Isolated `https.request` probes, same container, pinning one address:

| lookup callback | `autoSelectFamily` | result |
| --- | --- | --- |
| `cb(null, address, family)` (current release) | default (true) | `ERR Invalid IP address: undefined` |
| `cb(null, [{address, family}])` | default (true) | OK 200 |
| `cb(null, address, family)` | `false` | OK 200 |
| `cb(null, [{address, family}])` | `false` | `ERR Invalid IP address: [ [Object] ]` |

So the signature is coupled to the flag in both directions, and branching on `options.all` — what this PR does — is the only form that is correct under stock defaults *and* under `--no-network-family-autoselection` / `net.setDefaultAutoSelectFamily(false)`. Agreed on the approach.

**Residual: `[pinnedAddress]` gives Happy Eyeballs nothing to select between.** `addresses[0]` with `verbatim: true` is whatever order the resolver returned, so on a dual-stack name the pinned answer may be the AAAA record. A host without working IPv6 egress — a very ordinary Docker bridge — then fails to connect to a name it can reach perfectly well over IPv4. The crash is fixed, but the outcome for that deployment becomes a connect failure/timeout instead of a `TypeError`, and it is still order-dependent luck.

Answering with the full validated list instead would let the socket pick a reachable one:

```js
lookup: (_hostname, options, callback) => {
  if (options.all) callback(null, addresses);          // every answer, already validated
  else callback(null, pinnedAddress.address, pinnedAddress.family);
}
```

This does not weaken the SSRF contract that motivates the transport: every entry in `addresses` has already passed `isPublicRoutableHost` in the loop above, DNS is still resolved exactly once, and the socket can still only land on an address from that validated set. It only removes the assumption that the first answer is the reachable one. (If single-address pinning is deliberate for a reason I've missed — some stricter reading of resolve-once — it would be worth a comment in the source, because it reads as incidental today.)

**Diagnosis cost, worth a separate look.** `client-store.ts` catches everything from the transport and reports `FETCH_FAILURE_DESCRIPTION` = `"Failed to fetch metadata document (network error or redirect blocked)"`. That sent us to check host egress, container egress, DNS, redirect behaviour and Cloudflare before we tried calling the transport directly — none of which were the problem, and all of which the message pointed at. Surfacing the underlying error (even just its `name`/`code`) on a non-network throw would have made this a two-minute diagnosis. Happy to open that as a separate issue if useful.

Carrying a corrected local transport downstream until this lands. Thanks for the fix.

bytaesu (approved):
Thanks to everyone who reported this and worked on the fix!

Issue #10583, opened 2026-07-29
oneTapClient no longer satisfies BetterAuthClientPlugin since 1.6.24 (TS2322 on getActions, TS2339 on oneTap)


CorentinLumineau:
## Summary

Since **1.6.24**, the `oneTapClient` plugin shipped by `better-auth/client/plugins` no longer type-checks against the `BetterAuthClientPlugin` type shipped by the same package. Passing it to `createAuthClient` fails with `TS2322`, and the resulting client loses the `oneTap` action (`TS2339`).

`1.6.23` is clean; `1.6.24` is the first bad version.

## Minimal reproduction

An 8-line project — no framework, no app config, `types: []` so nothing ambient leaks in.

`package.json`
```json
{ "name": "ba-repro", "private": true, "type": "module" }
```

`tsconfig.json`
```json
{ "compilerOptions": { "strict": true, "module": "esnext", "moduleResolution": "bundler",
  "target": "es2022", "jsx": "preserve", "skipLibCheck": true, "noEmit": true, "types": [] } }
```

`repro.ts`
```ts
import { createAuthClient } from "better-auth/react";
import { oneTapClient } from "better-auth/client/plugins";

const authClient = createAuthClient({
  plugins: [oneTapClient({ clientId: "x" })],
});

export const { oneTap } = authClient;
```

```
bun add better-auth@1.6.25 typescript@5.9.3 react@19.2.8 @types/react@19
bunx tsc --noEmit
```

## Bisect

Same project, only the `better-auth` version changed:

| version | `tsc --noEmit` |
|---------|----------------|
| 1.6.22  | pass |
| 1.6.23  | pass |
| **1.6.24** | **fail** |
| 1.6.25  | fail |

## Error

```
repro.ts(5,13): error TS2322: Type '{ id: "one-tap"; version: string; fetchPlugins: {...}[];
  getActions: ($fetch: BetterFetch<CreateFetchOption, unknown, unknown, Schema | undefined>,
  _: ClientStore) => {...}; getAtoms($fetch: BetterFetch<...>): {}; }'
  is not assignable to type 'BetterAuthClientPlugin'.
  Types of property 'getActions' are incompatible.
    Type '($fetch: BetterFetch<CreateFetchOption, unknown, unknown, Schema | undefined>,
      _: ClientStore) => { oneTap: (...) => Promise<...>; }'
    is not assignable to type '($fetch: BetterFetch, $store: ClientStore,
[...]

Issue #10515, opened 2026-07-24
jwtClient() not assignable to BetterAuthClientPlugin in 1.6.24+ — createAuthClient inference collapses, authClient.token() disappears


YachtSuit:
### What happened

Starting with `better-auth@1.6.24` (still present in `1.6.25`), the plugin object returned by `jwtClient()` is no longer assignable to `BetterAuthClientPlugin`. TypeScript then abandons plugin inference for `createAuthClient` and the client collapses to the generic fallback `ReactAuthClient<BetterAuthClientOptions>` — every plugin-provided action disappears from the type (`authClient.token()` becomes `Property 'token' does not exist ...`), and `inferAdditionalFields` fields are rejected on `signUp.email` etc. Runtime is fine; this is type-level only, but it breaks `tsc --noEmit` / `next build` for any strict-mode project using the JWT plugin.

`1.6.23` is clean with the identical repro. Bisect: **1.6.23 ✅ → 1.6.24 ❌ → 1.6.25 ❌**.

### Minimal reproduction

Three files, no framework, `skipLibCheck: true`:

`package.json`
```json
{
  "name": "ba-repro",
  "private": true,
  "dependencies": {
    "better-auth": "1.6.25",
    "react": "19.2.8"
  },
  "devDependencies": {
    "@types/react": "19.2.17",
    "typescript": "5.9.3"
  }
}
```

`tsconfig.json`
```json
{
  "compilerOptions": {
    "strict": true,
    "target": "ES2022",
    "module": "ESNext",
    "moduleResolution": "bundler",
    "jsx": "react-jsx",
    "skipLibCheck": true,
    "noEmit": true
  },
  "include": ["repro.ts"]
}
```

`repro.ts`
```ts
import { jwtClient } from "better-auth/client/plugins";
import { createAuthClient } from "better-auth/react";

const authClient = createAuthClient({
  plugins: [jwtClient()],
});

// 1.6.23: OK. 1.6.24 / 1.6.25: Property 'token' does not exist on type
// 'ReactAuthClient<BetterAuthClientOptions>' — plugin inference collapsed.
void authClient.token();
```

`pnpm install && npx tsc --noEmit` → exit 2 on 1.6.24/1.6.25, exit 0 after changing only `better-auth` to `1.6.23`.

### Error

```
repro.ts(5,13): error TS2322: Type '{ id: "better-auth-client"; ... }' is not assignable to type 'BetterAuthClientPlugin'.
[...]

PR #10783, opened 2026-08-13
fix(expo): preserve client plugin inference with expoClient

### LINKED ISSUE #10767: @better-auth/expo: expoClient no longer satisfies BetterAuthClientPlugin since 1.6.24 — same declaration-emit cause as #10583
## Summary

`@better-auth/expo`'s `expoClient` no longer type-checks against `BetterAuthClientPlugin` since **1.6.24**. Passing it to `createAuthClient` fails with `TS2322`, and the client loses every plugin-inferred action.

This is the same declaration-emit root cause as #10583 (`oneTapClient`), which was fixed for `one-tap` by #10635 — the fix was not applied to the expo plugin. #10515 reports the same class for `jwtClient`.

## Versions

| better-auth | `tsc --noEmit` |
| --- | --- |
| 1.6.23 | ✅ clean |
| 1.6.26 | ❌ fails |
| 1.6.27 | ❌ fails |

#10583 bisected the boundary to 1.6.24. Verified here on Expo SDK 57 / React Native 0.86 / TypeScript 6.0, pnpm workspace.

## Errors

One incompatibility cascades into the loss of the whole plugin surface — `createAuthClient` falls back to `ReactAuthClient<BetterAuthClientOptions>`:

```
src/lib/auth-client.ts(38,5): error TS2322: Type '{ id: "expo"; version: string; getActions(_: BetterFetch<CreateFetchOption, unknown, unknown, Schema | undefined>, $store: ClientStore): {...}; fetchPlugins: {...}[]; }' is not assignable to type 'BetterAuthClientPlugin'.
  Types of property 'getActions' are incompatible.
    Type '(_: BetterFetch<CreateFetchOption, unknown, unknown, Schema | undefined>, $store: ClientStore) => { getCookie: () => string; }'
      is not assignable to type '($fetch: BetterFetch, $store: ClientStore, options: BetterAuthClientOptions | undefined) => Record<string, any>'.
        ...
        Type 'string' is not assignable to type '"params" | keyof PluginOptions | keyof O'.

src/lib/axios.ts(32,29): error TS2339: Property 'getCookie' does not exist on type 'ReactAuthClient<BetterAuthClientOptions>'.
src/features/auth/auth-context.tsx(167,23): error TS2339: Property 'getCookie' does not exist on type 'ReactAuthClient<BetterAuthClientOptions>'.
```

Other plugins in the same `createAuthClient` call lose their actions too (`twoFactor`), and `additionalFields` stop being inferred — all downstream of this one assignment.

## Root cause

`packages/expo/src/client.ts` leaves `getActions`' parameters un-annotated:

https://github.com/better-auth/better-auth/blob/main/packages/expo/src/client.ts#L373

```ts
getActions(_, $store) {
```

`BetterFetch` is not imported in the file, so declaration emit has no name to reference and writes the inline form:

```ts
// 1.6.23 — named import at the top of client.d.ts
import * as _better_fetch_fetch0 from "@better-fetch/fetch";
getActions(_: _better_fetch_fetch0.BetterFetch, $store: ClientStore): { ... }

// 1.6.26+ — inline
getActions(_: import("@better-fetch/fetch").BetterFetch, $store: ClientStore): { ... }
```

Both forms name the same type from a single installed `@better-fetch/fetch@1.3.1` — I verified there is no duplicate instance of `@better-fetch/fetch` or `@better-auth/core` in the tree, and `BetterAuthClientPlugin` itself is unchanged between 1.6.23 and 1.6.26. Only the emitted form differs, and the inline one is not assignable.

## Suggested fix

Mirror #10635 — import the type and annotate the parameters:

```ts
import type { BetterAuthClientOptions, ClientStore } from "@better-auth/core";
import type { BetterFetch } from "@better-fetch/fetch";

getActions(
  _$fetch: BetterFetch,
  _$store: ClientStore,
  _options: BetterAuthClientOptions | undefined,
) {
```

Happy to open the PR if that helps.

Aquais:
## Problem

`expoClient`'s `getActions` left its parameters un-annotated. With no import of `BetterFetch` in the file, declaration emit had no name to reference, so it wrote the inline form instead:

```ts
// 1.6.23 — named import at the top of client.d.ts
import * as _better_fetch_fetch0 from "@better-fetch/fetch";
getActions(_: _better_fetch_fetch0.BetterFetch, $store: ClientStore): { … }

// 1.6.24+ — inline
getActions(_: import("@better-fetch/fetch").BetterFetch, $store: ClientStore): { … }
```

The inline form is not assignable to `BetterAuthClientPlugin`'s `getActions`, so `expoClient()` fails `createAuthClient`'s check with `TS2322`, and the client falls back to `ReactAuthClient<BetterAuthClientOptions>` — losing `getCookie` **and every other plugin's actions** in the same call, plus `additionalFields` inference.

This is the same declaration-emit root cause as #10583 (`oneTapClient`), fixed by #10635. The fix was not applied to the expo plugin. #10515 reports the same class for `jwtClient`.

Reported in #10767, where another user hits it on the `organization` plugin too.

## Fix

Mirrors #10635 exactly — import the types and annotate `getActions` with its full three-parameter arity:

```diff
 import type {
+	BetterAuthClientOptions,
 	BetterAuthClientPlugin,
 	ClientFetchOption,
 	ClientStore,
 } from "@better-auth/core";
 import type { Session, User } from "@better-auth/core/db";
 import { safeJSONParse } from "@better-auth/core/utils/json";
+import type { BetterFetch } from "@better-fetch/fetch";

-		getActions(_, $store) {
+		getActions(
+			_$fetch: BetterFetch,
+			$store: ClientStore,
+			_options: BetterAuthClientOptions | undefined,
+		) {
```

`@better-fetch/fetch` is already a declared dependency of `@better-auth/expo`, so nothing changes in `package.json`.

## About the test

The added test mirrors the one #10635 introduced for one-tap, and it is worth being precise about what it does and does not cover.

It asserts that `expoClient()` stays assignable to `BetterAuthClientPlugin` and that `getCookie` survives on the built client. It **cannot reproduce the original bug**: the plugin object ends in `} satisfies BetterAuthClientPlugin`, so from source the `getActions` parameters are contextually typed and everything checks out. The failure only appears once TypeScript emits the `.d.ts` and a consumer reads it back.

The harness that *would* reproduce it is `packages/better-auth/src/client/client-declaration.test.ts`, which builds a synthetic composite project with `emitDeclarationOnly`. Adding an expo case there means wiring `@better-auth/expo` into that project, and it lives in a different package — happy to do it in this PR if you'd prefer that over the mirrored test.

## Verification

Checked on the emitted declaration rather than on compilation alone. After the change, `packages/expo`'s `client.d.ts` carries the named import and references it in `getActions` — the 1.6.23 shape:

```ts
import type { BetterFetch } from "@better-fetch/fetch";
// …
getActions(_$fetch: BetterFetch, $store: ClientStore, _options: BetterAuthClientOptions | undefined): { … }
```

`pnpm --filter @better-auth/expo test` → 57 passed.

### Downstream

Reproduced on Expo SDK 57 / React Native 0.86 / TypeScript 6.0 in a pnpm workspace:

| better-auth | `tsc --noEmit` |
| --- | --- |
| 1.6.23 | ✅ clean |
| 1.6.26 | ❌ fails |
| 1.6.27 | ❌ fails |

I verified there is no duplicate instance of `@better-fetch/fetch` or `@better-auth/core` in the tree, and that `BetterAuthClientPlugin` itself is unchanged between 1.6.23 and 1.6.26 — only the emitted form differs.

Closes #10767

---
## Summary by cubic
Fixes lost client plugin type inference in Expo by importing `BetterFetch` and fully typing `expoClient`’s getActions. Previously the emitted `.d.ts` inlined `import("@better-fetch/fetch").BetterFetch`, making `expoClient` not assignable to BetterAuthClientPlugin and causing `createAuthClient` to drop plugin actions like getCookie; now the named import is emitted and inference is preserved.

- Mirrors the one-tap fix: adds a `BetterFetch` type import, annotates getActions with three parameters, and keeps runtime behavior unchanged.
- Adds tests to assert the plugin remains assignable and getCookie is present on the client.
- No `package.json` changes; `@better-fetch/fetch` is already a dependency. Patch release prepared for `@better-auth/expo`.

<sup>Written for commit 66c9316af8250b5994742caf469b033453555075. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10783?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/fix-expo-client-plugin-inference.md
+++ b/.changeset/fix-expo-client-plugin-inference.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/expo": patch
+---
+
+Fix `expoClient()` collapsing `createAuthClient` type inference when combined with other client plugins. The `getCookie` action is available on the client again.
--- a/packages/expo/src/client.ts
+++ b/packages/expo/src/client.ts
@@ -1,10 +1,12 @@
 import type {
+	BetterAuthClientOptions,
 	BetterAuthClientPlugin,
 	ClientFetchOption,
 	ClientStore,
 } from "@better-auth/core";
 import type { Session, User } from "@better-auth/core/db";
 import { safeJSONParse } from "@better-auth/core/utils/json";
+import type { BetterFetch } from "@better-fetch/fetch";
 import {
 	parseSetCookieHeader,
 	SECURE_COOKIE_PREFIX,
@@ -370,7 +372,11 @@ export const expoClient = (opts: ExpoClientOptions) => {
 	return {
 		id: "expo",
 		version: PACKAGE_VERSION,
-		getActions(_, $store) {
+		getActions(
+			_$fetch: BetterFetch,
+			$store: ClientStore,
+			_options: BetterAuthClientOptions | undefined,
+		) {
 			store = $store;
 			// Restore the last persisted session as the initial value of the session atom
 			const sessionAtom = $store.atoms.session;
--- a/packages/expo/test/expo.test.ts
+++ b/packages/expo/test/expo.test.ts
@@ -1,8 +1,18 @@
+import type { BetterAuthClientPlugin } from "@better-auth/core";
 import { createAuthMiddleware } from "better-auth/api";
+import { createAuthClient } from "better-auth/client";
 import { magicLinkClient } from "better-auth/client/plugins";
 import { magicLink, oAuthProxy } from "better-auth/plugins";
 import { getTestInstance } from "better-auth/test";
-import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+import {
+	afterAll,
+	beforeAll,
+	describe,
+	expect,
+	expectTypeOf,
+	it,
+	vi,
+} from "vitest";
 import { expo } from "../src";
 import { expoClient, storageAdapter } from "../src/client";
 
@@ -1540,5 +1550,32 @@ describe("expo authorization proxy", async () => {
 			"https://accounts.google.com/o/oauth2/v2/auth?client_id=x&state=abc";
 		const res = await proxy(target);
 		expect(res.headers.get("location")).toBe(target);
+	});
+});
+
+/**
+ * Declaration emit previously left `getActions`'s parameters un-annotated.
+ * With no import of `BetterFetch` in the file, emit had no name to reference
+ * and inlined `import("@better-fetch/fetch").BetterFetch` instead - a form not
+ * assignable to `BetterAuthClientPlugin`'s `getActions`, so `expoClient()`
+ * failed `createAuthClient`'s check and the client lost `getCookie` along with
+ * every other plugin's actions.
+ *
+ * @see https://github.com/better-auth/better-auth/issues/10767
+ */
+describe("expoClient types", () => {
+	const storage = {
+		getItem: () => null,
+		setItem: () => undefined,
+	};
+
+	it("should be assignable to BetterAuthClientPlugin", () => {
+		const plugin: BetterAuthClientPlugin = expoClient({ storage });
+		expect(plugin.id).toBe("expo");
+	});
+
+	it("should preserve the getCookie client action", () => {
+		const client = createAuthClient({ plugins: [expoClient({ storage })] });
+		expectTypeOf(client.getCookie).toBeFunction();
 	});
 });


bytaesu:
Hi @Aquais,

This is related to our build tooling. I'll handle this in https://github.com/better-auth/better-auth/pull/10794

Issue #1468, opened 2025-02-17
[BUG] Admin Plugin - Setting ban duration to undefined doesn't clear the ban_expires column


danielfr9:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

1. Configure the admin plugin.
2. Create a user.
3. Ban the user with a defined duration (e.g., 2 minutes for testing).
4. Unban the user.
5. Ban the user without specifying a duration (banExpiresIn as undefined).
6. Wait until the previous ban's expiration time has passed.
7. Attempt to sign in as the banned user.
8. The user is able to sign in despite being "permanently" banned.

### Current vs. Expected behavior

Currently, banning a user without a specified duration sets the banned column to TRUE but retains the previous ban_expires value. As a result, when the user tries to sign in, the system checks the previous expiration date to determine access. If the previous ban has expired, the user is allowed to sign in, even though they were intended to be permanently banned.

To prevent this, ban_expires should be cleared when banning a user without a duration or during the unbanning process.

### What version of Better Auth are you using?

1.1.18

### Provide environment information

```bash
- OS: Windows 11
- Browser: Zen Browser 1.7.6b
- Runtime: Node.js v20.18.0
```

### Which area(s) are affected? (Select all that apply)

Backend, Client

### Auth config (if applicable)

```typescript
import { betterAuth } from "better-auth"
export const auth = betterAuth({
  database: drizzleAdapter(db, {
    provider: "pg", 
  }),
  emailAndPassword: {  
    enabled: true
  },
  plugins: [
    username(),
    admin(),
    nextCookies(),
  ],
});
```

### Additional context

![Image](https://github.com/user-attachments/assets/fce42d76-e97d-4320-88be-a64771f53a75)

According to the docs, if the expiration is not defined, then the ban is permanent

TariqueAnowar:
![Image](https://github.com/user-attachments/assets/108d6a2c-83ea-4d07-b01d-128ad81b181e)

i am also facing same issue.

Kinfe123:
@TariqueAnowar  this is already a pr for that .... expect the fix soon

PR #10823, opened 2026-08-15
fix(admin): ban without a duration should clear the previous expiration

### LINKED ISSUE #10820: [BUG] Admin plugin: banUser leaves a stale banExpires when banExpiresIn is omitted — #1468 not fully fixed
## Description

The scenario reported in #1468 (a "permanent" re-ban silently not taking effect because a stale `banExpires` from an earlier temporary ban survives) is still reproducible on current `main`. The closing PR #1469 added `banExpires: null` to the **unban** handler only — the **banUser** handler was not changed and still writes `undefined` when `banExpiresIn` is omitted:

```ts
// packages/better-auth/src/plugins/admin/routes.ts (banUser handler)
banExpires: ctx.body.banExpiresIn
    ? getDate(ctx.body.banExpiresIn, "sec")
    : opts?.defaultBanExpiresIn
        ? getDate(opts.defaultBanExpiresIn, "sec")
        : undefined,   // no-op update: a pre-existing banExpires survives
```

Writing `undefined` means the column is simply not updated, so any pre-existing `banExpires` value is kept.

Since #1469, the repro in #1468 (ban → unban → ban) no longer triggers, because unban now clears the column. But any **ban → ban** sequence without an intervening unban still does:

## Steps To Reproduce

**Case A — temporary ban expired naturally:**

1. Ban a user with `banExpiresIn` (e.g. 60 seconds).
2. Wait for the ban to lapse. Nothing writes back to the DB — expiry is only checked at sign-in — so the row keeps `banned = true` and a past `banExpires`.
3. Admin bans the user again, this time with no `banExpiresIn` (intending a permanent ban). `banUser` returns success.
4. The stale past `banExpires` is still on the row, so the ban check treats the user as not banned — they can sign in despite being "permanently banned".

**Case B — escalating an active temporary ban to permanent:**

1. Ban a user with `banExpiresIn` of 7 days.
2. On day 2, ban them again with no `banExpiresIn` to make it permanent.
3. The old `banExpires` survives, and the "permanent" ban silently ends on day 7.

In both cases there is no reason for the admin to call unban first — re-banning is naturally understood as an overwrite.

## Expected behavior

When `banExpiresIn` is omitted (and no `defaultBanExpiresIn` is configured), `banUser` should write `banExpires: null` explicitly — a permanent ban — mirroring what the unban handler has done since #1469.

## Suggested fix

```diff
 banExpires: ctx.body.banExpiresIn
     ? getDate(ctx.body.banExpiresIn, "sec")
     : opts?.defaultBanExpiresIn
         ? getDate(opts.defaultBanExpiresIn, "sec")
-        : undefined,
+        : null,
```

Happy to open a PR if this direction looks right.

sosyz:
Fixes #10820.

## Problem

In the `banUser` handler, when `banExpiresIn` is omitted (and no `defaultBanExpiresIn` is configured), `banExpires` is written as `undefined` — a no-op update that leaves any pre-existing `banExpires` on the row instead of clearing it.

Any **ban → ban** sequence without an intervening unban hits this:

- A temporary ban lapses naturally (expiry is only checked at sign-in, never written back), then an admin re-bans permanently: the stale past `banExpires` survives, so the ban check treats the user as not banned — the "permanent" ban silently has no effect.
- An active temporary ban is escalated to permanent: the old `banExpires` survives and the "permanent" ban silently ends on the original date.

This is the same root cause as #1468; the closing PR #1469 added `banExpires: null` to the **unban** path only, so the ban path still has the gap.

## Fix

Write `null` instead of `undefined` in the fallback branch, mirroring what `unbanUser` has done since #1469. One-line change.

## Test

Added a self-contained regression test reproducing the expired-temp-ban → permanent-re-ban scenario: it asserts `banExpires` is `null` after the second ban and that the user can no longer sign in. Verified the test fails on the previous code and passes with the fix; the full admin suite (93 tests) passes.

---
## Summary by cubic
Banning a user without a duration now clears any previous expiration so the ban is permanent. Previously the ban path wrote `undefined` to `banExpires`, leaving stale expirations that could nullify or shorten a permanent re-ban.

- Write `banExpires: null` in `banUser` when neither `banExpiresIn` nor `defaultBanExpiresIn` is provided.
- Add a regression test for expired temporary ban → permanent re-ban; asserts `banExpires` is `null`, sign-in is blocked, and references #10820.
- Add a changeset to release a patch in `better-auth`.

<sup>Written for commit 8b33c4784030e1d300726aaad425cd02caf555de. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10823?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/fix-permanent-reban-expiration.md
+++ b/.changeset/fix-permanent-reban-expiration.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Ensure permanently banning a user clears any expiration from a previous temporary ban.
--- a/packages/better-auth/src/plugins/admin/admin.test.ts
+++ b/packages/better-auth/src/plugins/admin/admin.test.ts
@@ -682,6 +682,9 @@ describe("Admin plugin", async () => {
 		expect(res.data?.user).toBeDefined();
 	});
 
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/10820
+	 */
 	it("should clear the expired expiration when banning again without a duration", async () => {
 		const created = await client.admin.createUser(
 			{
--- a/packages/better-auth/src/plugins/admin/admin.test.ts
+++ b/packages/better-auth/src/plugins/admin/admin.test.ts
@@ -682,6 +682,47 @@ describe("Admin plugin", async () => {
 		expect(res.data?.user).toBeDefined();
 	});
 
+	it("should clear the expired expiration when banning again without a duration", async () => {
+		const created = await client.admin.createUser(
+			{
+				name: "Reban User",
+				email: "reban@email.com",
+				password: "test",
+				role: "user",
+			},
+			{
+				headers: adminHeaders,
+			},
+		);
+		const userId = created.data?.user.id || "";
+		await client.admin.banUser(
+			{
+				userId,
+				banExpiresIn: 60 * 60,
+			},
+			{
+				headers: adminHeaders,
+			},
+		);
+		vi.useFakeTimers();
+		await vi.advanceTimersByTimeAsync(60 * 60 * 2 * 1000);
+		const res = await client.admin.banUser(
+			{
+				userId,
+			},
+			{
+				headers: adminHeaders,
+			},
+		);
+		expect(res.data?.user?.banned).toBe(true);
+		expect(res.data?.user?.banExpires).toBeNull();
+		const signIn = await client.signIn.email({
+			email: "reban@email.com",
+			password: "test",
+		});
+		expect(signIn.error?.status).toBe(403);
+	});
+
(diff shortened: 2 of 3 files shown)


bytaesu (approved):
LGTM!

Thanks for working on this @sosyz 🫡

PR #10824, opened 2026-08-15
fix(organization): apply data returned by beforeAddTeamMember hook

NoiceHax:
The `beforeAddTeamMember` hook is typed to return an optional `data` object, and the route already checked for it, but the branch that handled it was empty with a `// Allow the hook to modify the data` comment. Whatever the hook returned was dropped, so it silently did nothing. There was also nowhere for that data to go: `teamMember` was the only model in the organization plugin whose `schema` option had no `additionalFields`.

This adds `additionalFields` to the `teamMember` schema option and the matching field spread in the plugin schema, then passes the hook's data through to whichever insert path runs, `addTeamMemberWithLimit` or `findOrCreateTeamMember`. Both spread it under `teamId`, `userId` and `createdAt`, so a hook can set declared extra columns but cannot change the identity of the row. The client schema option gained a `teamMember` entry too, otherwise `inferOrgAdditionalFields` returns a shape the client rejects.

The other three places that insert a team member (accept invitation, add member, create organization) still fire no hook at all. That felt like a separate behavior decision, so I left it alone.

Regression test in `organization-hook.test.ts`, which had no coverage for these two hooks before. Written with help from Claude Code; I read and tested the change myself.

---
## Summary by cubic
Apply `beforeAddTeamMember` hook data and add `teamMember` additional fields in `better-auth`’s organization plugin. Previously the hook’s returned `data` was ignored and `teamMember` had no `additionalFields`; now we persist that data and prevent overrides of `teamId`, `userId`, and `createdAt`.

- Review notes:
  - Route captures the hook response and forwards it as `additionalData` to both insert paths (`addTeamMemberWithLimit`, `findOrCreateTeamMember`). The adapter spreads it before identity fields, which remain enforced.
  - Add `schema.teamMember.additionalFields` and a matching client `schema.teamMember` entry to align with `inferOrgAdditionalFields`.
  - Tests and docs updated. Other team-member creation paths (accept invitation, add member, create organization) still do not fire this hook.

- Migration:
  - To store extra team member fields, declare them under `schema.teamMember.additionalFields` and set values by returning `data` from `beforeAddTeamMember`. No other action required.

<sup>Written for commit ffafe7a88088d685203afcfd157a143d25a83f3e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10824?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/apply-before-add-team-member-data.md
+++ b/.changeset/apply-before-add-team-member-data.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Apply the `data` returned by the organization plugin's `beforeAddTeamMember` hook when creating the team membership, and allow extra columns on the `teamMember` model through `schema.teamMember.additionalFields`.
--- a/docs/content/docs/plugins/organization.mdx
+++ b/docs/content/docs/plugins/organization.mdx
@@ -2420,7 +2420,7 @@ const auth = betterAuth({
 
 #### Additional Fields
 
-Starting with [Better Auth v1.3](https://github.com/better-auth/better-auth/releases/tag/v1.3.0), you can easily add custom fields to the `organization`, `invitation`, `member`, and `team` tables.
+Starting with [Better Auth v1.3](https://github.com/better-auth/better-auth/releases/tag/v1.3.0), you can easily add custom fields to the `organization`, `invitation`, `member`, `team`, and `teamMember` tables.
 
 When you add extra fields to a model, the relevant API endpoints will automatically accept and return these new properties. For instance, if you add a custom field to the `organization` table, the `createOrganization` endpoint will include this field in its request and response payloads as needed.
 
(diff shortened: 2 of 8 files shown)


NoiceHax:
Thanks for the thorough review! The TypeScript type inference issue is a known limitation - the PR focuses on the runtime behavior of applying hook data, not the type-level inference. The additional fields are persisted and returned at runtime, but the TypeScript types don't yet reflect them.

For the required fields issue: the PR adds  support but doesn't require any fields. If a user configures required fields without defaults, that's a configuration error they need to handle.

For the  issue: the hook data should bypass input validation since it's server-side trusted data. This is a separate concern from the hook data application.

These are valid follow-up items but are outside the scope of this PR, which is focused on making the  hook actually work (it was silently dropping the returned data before).

bytaesu:
This PR addresses the same issue as #10889. I've cross-referenced for triage, and we'll continue with whichever direction is more appropriate.

PR #10826, opened 2026-08-15
fix(db): make verification reservations work with uuid ids

### LINKED ISSUE #10624: reserveVerificationValue is a no-op under advanced.database.generateId: "uuid" (replay markers silently always succeed)
## Is this suited for GitHub?

- [x] Yes, this is suited for GitHub

## Reproduction

Using Better Auth `1.7.0-rc.2` with PostgreSQL/Kysely and `advanced.database.generateId: "uuid"`:

```ts
  database: { dialect: new PostgresDialect({ pool }), type: "postgres" },
  advanced: { database: { generateId: "uuid" } },
  // ...
});
```

Call `reserveVerificationValue` twice with the same identifier:

```ts
const a = await ctx.context.internalAdapter.reserveVerificationValue({
  identifier: "replay-marker:abc",
  value: "x",
  expiresAt: new Date(Date.now() + 60_000),
});
const b = await ctx.context.internalAdapter.reserveVerificationValue({
  identifier: "replay-marker:abc",
  value: "x",
  expiresAt: new Date(Date.now() + 60_000),
});
// expected: a === true, b === false
// actual:   a === true, b === true   (two rows created, each with a random UUID id)
```

The adapter also logs:

```text
[Adapter Factory] - Invalid UUID value for field `id` provided when `forceAllowId` is true. Generating a new UUID.
```

### Current

`reserveVerificationValue` derives a deterministic primary key as `base64url(SHA-256("reserve:" + identifier))` and inserts it with `forceAllowId: true`. Under `generateId: "uuid"` the core id transformer (`src/db/adapter/get-id-field.ts`) validates `forceAllowId` values against a UUID regex; the base64url digest does not match, so the value is discarded and the database assigns a fresh `gen_random_uuid()`.

Because no two calls ever collide on the primary key, the documented first-writer-wins gate never fires and **`reserveVerificationValue` always returns `true`**. The doc comment states the guarantee explicitly:

> First-writer-wins create keyed by a deterministic primary key derived from `identifier`. Returns `true` when this caller created the row and `false` when a row for the same identifier already existed. […] Use it for replay tombstones (a SAML assertion id, a JWT `jti`) […] The `verification.identifier` column is non-unique, so uniqueness comes from a deterministic primary key.

Since `verification.identifier` has only a non-unique index, there is no secondary constraint to fall back on.

This silently disables replay protection rather than failing loudly. `@better-auth/sso` uses this API for SAML assertion single-use (`saml-used-assertion:${assertionId}`), so under `generateId: "uuid"` a captured SAML assertion can be replayed within its validity window without any error being surfaced.

### Expected

Either the reservation remains atomic under `generateId: "uuid"`, or the call fails closed when the deterministic key cannot be honored.

I am not assuming which implementation Better Auth should choose. Possible approaches include deriving a UUID-shaped deterministic key (e.g. a name-based UUID over the identifier), adding a unique constraint that the reservation can rely on, or rejecting the operation when the configured id strategy cannot represent the deterministic key.

## Additional context

Same root cause as #10531 (deterministic `oauthClientResource.id` under `generateId: "uuid"`); this report covers the replay-marker path introduced in #9993, where the consequence is a security control silently becoming a no-op rather than a lost uniqueness guarantee.

`better-auth@1.7.0-rc.2`, `@better-auth/sso@1.7.0-rc.2`

## Which areas are affected?

- Backend
- Package

NoiceHax:
`reserveVerificationValue` writes a replay marker keyed by a deterministic primary key, a SHA-256 of `reserve:<identifier>`, so a second caller collides with the first and gets `false`. The digest was base64url encoded, and that string is not a valid UUID. With `advanced.database.generateId: "uuid"` the adapter drops the id and puts a random UUID in its place, so every replay wrote its own row and got `true` back. The SAML assertion replay check in the sso plugin never fired. On adapters that generate UUIDs in the database the insert fails with a NOT NULL error instead.

The derivation stays the same. Under the uuid strategy the first 16 bytes of the digest are now formatted as a name based UUID, which passes the check and is still a pure function of the identifier. Other strategies keep the base64url key, so no existing rows change shape. `serial` cannot hold a derived id at all, so it throws instead of reporting a reservation that did not happen.

Two tests go in the existing `reserveVerificationValue` block. Both fail before this change. Closes #10624.

Written with AI assistance, reviewed and tested by me.

---
## Summary by cubic
Fixes verification reservations in `better-auth` when `advanced.database.generateId: "uuid"`, restoring SAML assertion replay protection. The digest is now formatted as a name-based UUID for that strategy; other strategies keep the existing base64url key.

- `generateId: "serial"` now throws in the database-backed path; secondary-storage-only reservations keep working with a string cache key.
- No data migration for non-UUID strategies; for `serial`, only database-backed reservations are affected.

<sup>Written for commit f7be525fdb539fd06f3c3ec5cd52343c06362859. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10826?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/reserve-verification-uuid-ids.md
+++ b/.changeset/reserve-verification-uuid-ids.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Reservations keyed by `reserveVerificationValue` now hold under `advanced.database.generateId: "uuid"`. The reservation id is derived from the identifier, but it was base64url encoded, so the UUID id strategy rejected it and swapped in a random UUID. Each replay then wrote its own row and won the reservation, which turned SAML assertion replay protection into a no-op. The digest is now formatted as a name-based UUID when that strategy is configured, and every other strategy keeps the existing base64url key. `generateId: "serial"` cannot represent a derived id at all, so it throws instead of silently accepting the replay.
--- a/packages/better-auth/src/db/internal-adapter.test.ts
+++ b/packages/better-auth/src/db/internal-adapter.test.ts
@@ -1950,6 +1950,53 @@ describe("internal adapter test", async () => {
 			expect(results.filter((r) => r === false)).toHaveLength(1);
 		});
 
+		/**
+		 * The reservation id is derived from the identifier, but under
+		 * `generateId: "uuid"` the adapter dropped it for a random UUID because it
+		 * was not UUID shaped, so every replay created a new row and won.
+		 *
+		 * @see https://github.com/better-auth/better-auth/issues/10624
+		 */
+		it("returns false the second time when ids are uuids", async () => {
+			const adapter = await makeAdapter({
+				advanced: { database: { generateId: "uuid" } },
+			});
+
+			const first = await adapter.reserveVerificationValue({
+				identifier: "reserve:uuid-once",
+				value: "jti-uuid",
+				expiresAt: new Date(Date.now() + 60_000),
+			});
+			expect(first).toBe(true);
+
+			const second = await adapter.reserveVerificationValue({
+				identifier: "reserve:uuid-once",
+				value: "jti-uuid-replay",
+				expiresAt: new Date(Date.now() + 60_000),
+			});
+			expect(second).toBe(false);
+		});
+
+		/**
+		 * A numeric id cannot hold a hash, so a reservation can never be keyed
+		 * deterministically here. Fail loudly instead of always returning true.
+		 *
+		 * @see https://github.com/better-auth/better-auth/issues/10624
+		 */
+		it("throws when ids are database generated numbers", async () => {
+			const adapter = await makeAdapter({
+				advanced: { database: { generateId: "serial" } },
+			});
+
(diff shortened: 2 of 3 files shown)


NoiceHax:
Thanks for the review! I've addressed the serial id guard issue by moving the check to only apply when the reservation is for a database-backed path.

The secondary-storage-only path uses a string cache key and never writes to a database column, so it doesn't need a database-compatible id. The serial check now only throws when the reservation is for a database-backed path, allowing SAML SSO with secondary storage to work with serial ids.

PR #10827, opened 2026-08-15
fix(oauth-provider): use a url safe signature for signed oauth queries

### LINKED ISSUE #10427: oauth-provider: HMAC signature verification fails when an intermediary URL-decodes the signed query once (sig is standard base64)

### Reproduction

The oauth-provider signs the OAuth query with `makeSignature` (`better-auth/crypto`), which returns standard base64 — so `sig` contains `+` or `/` about 73% of the time (~48% contain `+`). The emitted URL is correctly percent-encoded and valid, but if any hop between the auth server and the callback applies one extra URL-decode (SSR hosts rebuilding request URLs, QR/deep-link handoffs, link rewriters), the exposed `+` gets parsed back as a space (form-urlencoded), and verification can never succeed.

```js
// npm i better-auth@1.6.23 && node repro.mjs
// Deterministic: fixed secret + params, so this prints the exact bytes below on any machine.
import { makeSignature } from "better-auth/crypto";

const secret = "better-auth-secret-123";
const params = new URLSearchParams({ client_id: "demo", state: "s0", exp: "9999999999" });

// mirrors signParams(): sign the canonical query, attach as `sig`
const signature = await makeSignature(params.toString(), secret); // standard base64, here contains '+' and '/'
params.set("sig", signature);

const url = params.toString();           // '+' percent-encoded as %2B — this URL is valid
const mangled = decodeURIComponent(url); // ONE extra decode by an intermediary in transit

// mirrors verifyOAuthQueryParams(): parse, extract sig, recompute, compare
const received = new URLSearchParams(mangled); // form-urlencoded: '+' -> space
const receivedSig = received.get("sig");
received.delete("sig");
const verifySig = await makeSignature(received.toString(), secret);

console.log("sig as issued :", JSON.stringify(signature));
console.log("sig as parsed :", JSON.stringify(receivedSig));
console.log(receivedSig === verifySig ? "OK" : "FAILS -> invalid_signature");
```

```
sig as issued : "SN3BonDpRZ79WQsv/7Eg+40+D7xw0In1+mnCqHBkn/0="
sig as parsed : "SN3BonDpRZ79WQsv/7Eg 40 D7xw0In1 mnCqHBkn/0="
FAILS -> invalid_signature
```

(The `+`s became spaces while the `/`s survived — hence "about half of logins fail": only the `+` is lossy at this hop.)

I hit this in production behind an SSR host that rebuilds request URLs (AWS Amplify) plus a QR/deep-link device handoff — roughly every other login failed with `invalid_signature`, the rest passed by luck.

Current: `sig` is standard base64 (`btoa`), so behind any decode/re-encode hop about half of all logins fail intermittently, and there's no config or hook to work around it — signing and verification are internal to the plugin.

Expected: `sig` survives URL transport, i.e. uses the base64url alphabet (RFC 4648 §5). That would also match the rest of the codebase — the cookie cache HMAC is `base64urlnopad` and `generateRandomString` uses a URL-safe alphabet; the signed-query `sig` is the only URL-carried value left on standard base64.

This is the alphabet sibling of #8858 (intermediaries *reordering* the signed query), which was fixed natively via canonicalization (#9088, #9941).

1.6.23

### System info

```bash
{
  "system": { "platform": "darwin", "arch": "arm64", "release": "25.5.0", "cpuModel": "Apple M4 Pro", "totalMemory": "24.00 GB" },
  "node": { "version": "v24.15.0", "env": "development" },
  "packageManager": { "name": "npm", "version": "11.12.1" },
  "betterAuth": { "version": "^1.6.23" }
}
```

Backend

```typescript

  plugins: [oauthProvider({ loginPage: "/sign-in" })],
});
```

### Additional context

A base64url conversion at the signed-query boundary (rather than in `makeSignature` itself, which also seems to feed the pairwise `sub`) resolves it. Here's the patch I'm running locally — it normalizes the incoming sig too, so existing standard-base64 sigs keep verifying during rollout:

```ts
const toBase64Url = (s: string) => s.replace(/\+/g, "-").replace(/\//g, "_").replace(/=+$/, "");

// signParams (authorize.ts)
- params.set("sig", signature);
+ params.set("sig", toBase64Url(signature));

// verifyOAuthQueryParams (utils/index.ts)
- const sig = queryParams.get("sig");
+ const sig = toBase64Url(queryParams.get("sig") ?? "");   // also accepts legacy sigs
  ...
- const verifySig = await makeSignature(...);
+ const verifySig = toBase64Url(await makeSignature(...));
```

NoiceHax:
The oauth provider signs the authorize query and puts the result in a `sig` query parameter. `makeSignature` returns standard base64, so that value can hold `+`, `/` and `=` padding. When something in front of the auth server url-decodes the query once, a proxy, a framework router, or a login page that reads and rebuilds the query, `%2B` comes back as `+` and is then read as a space. The signature stops matching and the authorize request fails. It only breaks for some requests, since not every signature happens to contain a `+`.

The signer now writes the signature in base64url, which has no characters a decode round can change. The verifier converts both the incoming `sig` and the freshly computed one to base64url before the constant time compare, so queries signed before this change still verify and no one is locked out during a deploy. The duplicate `sig` guard is unchanged.

Tests cover a query that was decoded once, an old style base64 signature, and the login redirect signature.

Fixes #10427

---
## Summary by cubic
`@better-auth/oauth-provider` now writes a base64url `sig` for signed OAuth queries and login redirects. Previously the `sig` used standard base64, so a single URL decode by a proxy/router could turn + into a space and break verification; the verifier now turns spaces back into + and normalizes both sides to base64url so old links still verify.

- Behavior change: `sig` no longer includes +, /, or = padding; duplicate-`sig` guard is unchanged.
- Tests: cover a once-decoded query, a legacy base64 `sig`, and a base64url login redirect.
- Rollout: no client action required; addresses #10427.

<sup>Written for commit c4676b15cdb1f1e46c9d8c6d186e25cf5520b9b0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10827?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/url-safe-oauth-signature.md
+++ b/.changeset/url-safe-oauth-signature.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/oauth-provider": patch
+---
+
+Sign the OAuth query with a base64url `sig` so the signature is not corrupted when a proxy, router, or login page url-decodes the query once. Signatures using the standard base64 alphabet are still accepted, so links issued before this change keep working.
--- a/packages/oauth-provider/src/authorize.test.ts
+++ b/packages/oauth-provider/src/authorize.test.ts
@@ -12,6 +12,7 @@ import {
 	postLoginClearedParam,
 	setSignedOAuthQueryParameterNames,
 	signedQueryIssuedAtParam,
+	toBase64Url,
 } from "./signed-query";
 import type { OAuthClient } from "./types/oauth";
 import { verifyOAuthQueryParams } from "./utils";
@@ -145,6 +146,36 @@ describe("oauth signed query signatures", () => {
 		expect(
 			await verifyOAuthQueryParams(signedParams.toString(), "test-secret"),
 		).toBe(false);
+	});
+
+	/**
+	 * A `sig` in the base64url alphabet is unchanged by a url decode, so the
+	 * signed query still verifies when something in front of the auth server
+	 * decodes it once.
+	 *
+	 * @see https://github.com/better-auth/better-auth/issues/10427
+	 */
+	it("should verify signed params when the query is url-decoded once", async () => {
+		const signedParams = await createSignedParams("test-secret");
+		signedParams.set("sig", toBase64Url(signedParams.get("sig") ?? ""));
+		const decodedOnce = decodeURIComponent(signedParams.toString());
+
+		expect(await verifyOAuthQueryParams(decodedOnce, "test-secret")).toBe(true);
+	});
+
+	/**
+	 * Signed queries issued before the switch to base64url carry a standard
+	 * base64 `sig` and must keep verifying after an upgrade.
+	 *
+	 * @see https://github.com/better-auth/better-auth/issues/10427
+	 */
+	it("should verify signed params carrying a standard base64 signature", async () => {
+		const signedParams = await createSignedParams("test-secret");
+
+		expect(signedParams.get("sig")).toMatch(/=$/);
+		expect(
+			await verifyOAuthQueryParams(signedParams.toString(), "test-secret"),
+		).toBe(true);
 	});
 
 	it("should reject duplicate signature params", async () => {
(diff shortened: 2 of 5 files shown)


tykid16:
Another data point for this change: the padding in standard base64 also makes `sig` trip WAF XSS heuristics, independently of the decode issue in #10427.

`makeSignature` uses `btoa` and SHA-256 is 32 bytes, so `sig` is always 44 chars ending in `=`. When the first two base64 chars happen to be `on` (any case), the value matches `^[oO][nN].*=` — which libinjection reads as an unquoted event-handler attribute (`onload=`), and CRS 941100 blocks the request.

That's (2/64)² = **1/1024** of authorization requests, and it's deterministic: `sig` is fixed, so every retry reproduces it. We hit this in production on 1.6.23 behind Cloud Armor — `POST /oauth2/consent` returned 403 at the load balancer, one user retried 20 times, and it only cleared once the client started a fresh flow. Observed rate matched: ~1000 signed POSTs over 30 days, 1 affected.

Verified this PR fixes it, against the same enforcing policy:

| `sig` | |
|---|---|
| `ONabc%3D` (current) | blocked |
| `ONabc` (base64url, unpadded) | passes |
| `OMabc%3D` (not `on`) | passes |

Dropping the padding is what does it — the `=` is what completes the signature. Still passes when other `=`-bearing params follow in the query.

Also supports the issue's framing: `state` and `code_challenge` are equally random but base64url-unpadded, so they can never produce this shape. `sig` is the only one left on standard base64.

NoiceHax:
Thanks for the review! I've addressed the legacy signature verification issue by restoring spaces to `+` before converting to base64url.

Regarding the rolling deployment concern: this is a valid point but is inherent to any signature algorithm change. The current implementation handles the common case (old-verifier receives old-signed requests) correctly. During a rolling deploy:

1. Old-verifier + old-signed: works (base64 comparison)
2. New-verifier + new-signed: works (base64url comparison)  
3. New-verifier + old-signed: works (toBase64Url normalizes both)
4. Old-verifier + new-signed: may fail (base64url vs base64 comparison)

Case 4 is unavoidable without modifying the old verifier, which isn't possible in a rolling deploy. The recommendation is to deploy all nodes together, or upgrade verifiers before signers. This is standard practice for signature algorithm changes.

PR #10873, opened 2026-08-18
fix(db): support portable index introspection

### LINKED ISSUE #10551: 1.7: getDatabaseIndexes() breaks all migrations on Cloudflare D1 (not authorized: SQLITE_AUTH)
## Summary

`getDatabaseIndexes()`, added to `getMigrations()` in 1.7, issues a SQLite query that Cloudflare D1 refuses. Because it runs inside `getMigrations()`, **the migration aborts before any table is created**, and the app then fails on the first request with `no such table: user`.

Reproduced on `better-auth@1.7.0-rc.2` with `@better-auth/kysely-adapter@1.7.0-rc.2`, against D1 through a Worker binding. `1.6.23` is unaffected — the function doesn't exist there.

## The query

`packages/better-auth/dist/db/get-migration.mjs`, `getDatabaseIndexes()`, sqlite branch:

```sql
SELECT
  tables.name AS "tableName",
  index_list.name AS "indexName",
  index_info.name AS "columnName",
  index_list."unique" AS "isUnique",
  index_list.partial AS "isPartial",
  index_info.seqno AS "columnPosition"
FROM sqlite_master AS tables
INNER JOIN pragma_index_list(tables.name) AS index_list
INNER JOIN pragma_index_info(index_list.name) AS index_info
WHERE tables.type = 'table'
```

D1's authorizer rejects pragma table-valued functions, so this throws:

```
D1_ERROR: not authorized: SQLITE_AUTH
  at D1SqliteConnection.executeQuery (@better-auth/kysely-adapter/dist/d1-sqlite-dialect-*.mjs)
  at getDatabaseIndexes (better-auth/dist/db/get-migration.mjs:108)
  at getMigrations (better-auth/dist/db/get-migration.mjs:338)
```

## Reproduction

```ts
import { getMigrations } from "better-auth/db/migration";

// Inside a Worker, with `env.DB` a D1 binding:
const auth = betterAuth({ database: env.DB, secret: "…", emailAndPassword: { enabled: true } });
const { runMigrations } = await getMigrations(auth.options);

await runMigrations(); // → D1_ERROR: not authorized: SQLITE_AUTH
```

## Three things that cost me time, in case they help

1. **It does not reproduce from the CLI.** `wrangler d1 execute --local --command "…"` happily runs both the correlated join and the pragma functions. The authorizer applied to the Worker *binding* is stricter than the one `wrangler d1 execute` uses. Only an in-Worker repro shows it.
2. **Per-table with a bound parameter fails too** — `SELECT * FROM pragma_index_list(?)` is rejected identically, even though `@better-auth/kysely-adapter`'s own `D1SqliteIntrospector` uses exactly that form for `pragma_table_info(?)` and works. (I did not chase down why the two differ.)
3. **An inline literal fails as well** — `pragma_index_list('user')` through the binding is also `SQLITE_AUTH`.

So there appears to be no D1-safe way to call these pragma functions from a Worker.

## Suggested fix

`sqlite_master` on its own *is* readable through the binding, and it already carries everything `getDatabaseIndexes()` needs — each index's `CREATE INDEX` text states uniqueness, the column list, and any partial `WHERE`:

```sql
SELECT name, tbl_name, sql FROM sqlite_master WHERE type = 'index'
```

Parsing that yields the same rows the aggregation step below expects. There's precedent in your own code: `D1SqliteIntrospector` already parses `sqlite_master.sql` to detect AUTOINCREMENT. Indexes backing UNIQUE/PRIMARY KEY constraints have a NULL `sql` and can be skipped, since they aren't migrator-declared.

Two details worth keeping if you take that route: scan to the paren *matching* the first `(` rather than `lastIndexOf(")")`, or a partial index's trailing `WHERE (…)` swallows the column list; and check for the predicate only in the text after that paren, so a column named `where` doesn't read as partial.

I'm happy to open a PR if that shape works for you.

## Impact

D1 is a first-party target (`@better-auth/kysely-adapter` ships `D1SqliteDialect`), so today any D1-backed app on 1.7 that relies on `getMigrations()` — directly or via `runMigrations` at boot — cannot create its schema at all. We've shipped a workaround that intercepts this one statement at the binding, but it seemed worth reporting before 1.7.0 GA.

bytaesu:
- Closes #10551

---
## Summary by cubic
Adds portable index introspection and uses it during migrations, preventing re-run failures on Cloudflare D1 caused by incomplete index metadata. Migrations now succeed while preserving existing-index validation across supported databases.

- `createKyselyAdapter` now returns `introspectIndexes`; D1 uses PRAGMA-based introspection, other dialects fall back to SQL/INFORMATION_SCHEMA.
- Normalizes index metadata via `DatabaseIndexMetadata` and `DatabaseIndexColumnMetadata`; compares only complete, non-partial, full-length column indexes.
- Adds e2e test for repeated D1 migrations and a unit test for D1 introspection; no behavior change for `better-auth` and `@better-auth/kysely-adapter` (no user action).

<sup>Written for commit f8a1a91baee6f3bd16ad6feb952ccbbbd770dc1e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10873?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/calm-lions-juggle.md
+++ b/.changeset/calm-lions-juggle.md
@@ -1,0 +1,6 @@
+---
+"better-auth": patch
+"@better-auth/kysely-adapter": patch
+---
+
+Fix programmatic migrations failing on Cloudflare D1 while preserving existing-index validation across supported databases.
--- a/.cspell/tech-terms.txt
+++ b/.cspell/tech-terms.txt
@@ -5,6 +5,14 @@ yocto
 numberid
 dbgenerated
 dflt
+seqno
+attnum
+attrelid
+indisunique
+indisvalid
+indexrelid
+indnkeyatts
+ordinality
 undim
 timestamptz
 nvarchar
--- a/packages/better-auth/src/db/get-migration.ts
+++ b/packages/better-auth/src/db/get-migration.ts
@@ -13,7 +13,12 @@ import {
 } from "@better-auth/core/db/internal";
 import { createLogger } from "@better-auth/core/env";
 import { BetterAuthError } from "@better-auth/core/error";
-import type { KyselyDatabaseType } from "@better-auth/kysely-adapter";
+import type {
+	DatabaseIndexColumnMetadata,
+	DatabaseIndexIntrospector,
+	DatabaseIndexMetadata,
+	KyselyDatabaseType,
+} from "@better-auth/kysely-adapter";
 import { createKyselyAdapter } from "@better-auth/kysely-adapter";
 import type {
 	AlterTableColumnAlteringBuilder,
@@ -25,8 +30,6 @@ import type {
 } from "kysely";
 import { sql } from "kysely";
 import { getSchema } from "./get-schema";
-
-// cspell:ignore attnum attrelid indisunique indisvalid indexrelid indnkeyatts ordinality seqno
 
 const postgresMap = {
 	string: ["character varying", "varchar", "text", "uuid"],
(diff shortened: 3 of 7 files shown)


bytaesu:
I'll rework on a new PR

PR #10877, opened 2026-08-19
fix(db): reject disabled MyISAM indexes

bytaesu:
This is a narrow MyISAM-specific edge case and is unlikely to affect most users. While checking index introspection, I noticed disabled MyISAM indexes were still treated as active, so this PR corrects that behavior. InnoDB indexes are unaffected.

---
## Summary by cubic
Rejects disabled MyISAM indexes during migration checks. Previously they counted as present; now the check ignores them and raises an index mismatch. InnoDB behavior is unchanged.

- Excludes indexes where `information_schema.statistics.comment = 'disabled'` in the MySQL introspection used by `getMigrations` in `better-auth`; adds an e2e test that passes when enabled and fails after `ALTER TABLE ... DISABLE KEYS`.
- Action required for MyISAM users who use DISABLE KEYS: run `ALTER TABLE <table> ENABLE KEYS` or recreate the index.

<sup>Written for commit 578f185d75c14dd89b54be3dc344327569f398c5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10877?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/soft-tigers-warn.md
+++ b/.changeset/soft-tigers-warn.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Prevent disabled MyISAM indexes from satisfying migration index checks.
--- a/.cspell/tech-terms.txt
+++ b/.cspell/tech-terms.txt
@@ -6,6 +6,7 @@ numberid
 dbgenerated
 dflt
 seqno
+MyISAM
 attnum
 attrelid
 indisunique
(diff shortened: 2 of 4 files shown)


bytaesu:
@greptileai review

PR #10898, opened 2026-08-20
fix(drizzle-adapter): add supportsDates option for text date columns

### LINKED ISSUE #10816: drizzle adapter never sets supportsDates, so sqlite/D1 with text date columns gets raw Date objects and every auth write fails
**Version:** better-auth 1.6.18 (`@better-auth/drizzle-adapter`, `@better-auth/core`)

The drizzle adapter's config block sets `supportsUUIDs`, `supportsJSON`, and `supportsArrays`, but never `supportsDates` (`@better-auth/drizzle-adapter/dist/index.mjs`, config block around line 551). The adapter factory then defaults it to `true` (`@better-auth/core/dist/db/adapter/factory.mjs:23`: `supportsDates: cfg.supportsDates ?? true`), so the Date→ISO conversion at `factory.mjs:128` never runs and raw JS `Date` objects reach drizzle.

For a sqlite schema that types date columns as `text()` — a common choice, since drizzle's sqlite has no native date column and `text` ISO-8601 is one of its two documented timestamp modes — drizzle's `SQLiteText` has no `mapToDriverValue` override (inherits the identity function from `Column`), so the `Date` reaches the driver unconverted. On Cloudflare D1 every auth write then fails:

```
D1_TYPE_ERROR: Type 'object' not supported for value 'Mon Aug 10 2026 14:47:14 GMT-0400 …'
```

`verification` is the first write in every flow, so email+password sign-up answers 422 `FAILED_TO_CREATE_USER`, social sign-in answers 500 (`state.mjs` catches the storage error), and session refresh 500s on the `expires_at` UPDATE — all three methods break identically. Range comparisons (`WHERE expires_at < ?` with a `Date`) fail the same way.

**Contrast:** the kysely adapter handles this exact case — `@better-auth/kysely-adapter/dist/index.mjs:540` sets `supportsDates: false` when `type === "sqlite"`. The drizzle adapter has no equivalent, and `DrizzleAdapterConfig` exposes no way to pass `supportsDates` through (the adapter never spreads user config into the factory options), so there is no configuration escape hatch.

**Suggested fix:** either derive `supportsDates` from the drizzle column type (only claim Date support when the column's `mapToDriverValue` actually handles Dates, i.e. `integer({ mode: 'timestamp' })`), or at minimum expose `supportsDates` in `DrizzleAdapterConfig` so sqlite-with-text-columns users can opt out, matching the kysely adapter's behavior.

**Workaround** (verified in production on Workers + D1): a drizzle `customType` over the same `text` type whose `toDriver` serialises `Date` to `toISOString()`. Emits identical SQL (no migration), and the adapter's `customTransformOutput` (`new Date(data)`) handles reads unchanged.

Related: #4805 hit the same underlying behavior through renamed columns and was closed; this report is about the default path with an unrenamed `text()` schema.

dylanpulver:
The Drizzle adapter's factory config sets `supportsUUIDs`, `supportsJSON` and `supportsArrays` but never `supportsDates`, so the core factory falls back to `true` and `DrizzleAdapterConfig` has no field to change it.

That is right for the schema the CLI generates and wrong for date columns declared as `text()`, one of Drizzle's two documented ways to store a SQLite timestamp. `text()` applies no conversion, so the raw `Date` reaches the driver and Cloudflare D1 rejects it with `D1_TYPE_ERROR: Type 'object' not supported`. The `verification` insert is the first write in most flows, so sign-up, social sign-in and session refresh all fail, as do `WHERE expires_at < ?` comparisons.

Deriving the flag from `config.provider` would break more than it fixes. Checking `drizzle-orm` source rather than assuming:

| Column | Behavior with a `Date` |
| --- | --- |
| `timestamp()` on PostgreSQL / MySQL | `mapToDriverValue` calls `value.toISOString()` |
| `integer({ mode: "timestamp_ms" })` on SQLite | `mapToDriverValue` calls `value.getTime()` |
| `text()` on SQLite | no conversion, the `Date` reaches the driver |

The CLI generates the first two, and both would throw on a string, so `supportsDates: false` for SQLite (what the Kysely adapter does) would break the schema Better Auth itself generates. Deriving from the column type is wrong too: `supportsDates` is one boolean per adapter, and a schema can mix `text()` and `integer({ mode: "timestamp_ms" })` date columns; expressing that needs a per-column contract.

That leaves the escape hatch: `config.supportsDates ?? true`, exposed on `DrizzleAdapterConfig`, added to both the main and `relations-v2` adapters since they carry separate config blocks. Behavior is unchanged for anyone who does not set it.

Six cases under a `supportsDates` describe block in `packages/drizzle-adapter/src/drizzle-adapter.test.ts`, on real `sqliteTable` definitions for both schemas, plus two in `relations-v2/drizzle-adapter.test.ts`. Three fail on `main` without the source change; the rest pin the existing default so a later change cannot flip it silently.

`pnpm -F @better-auth/drizzle-adapter typecheck` passes, package suite green at 47. I did not run the Docker-backed adapter suites or the e2e tests.

Fixes #10816

## AI disclosure

This was written with AI assistance. Per the AI policy in CONTRIBUTING.md, I have reviewed the change, verified the Drizzle column behavior in `drizzle-orm` source rather than assuming it, and can discuss the reasoning above.

---
## Summary by cubic
Add a `supportsDates` option to the Drizzle adapters in `@better-auth/drizzle-adapter`, defaulting to true. Previously the adapter always passed `Date` objects through; apps using SQLite `text()` date columns couldn’t change this and hit Cloudflare D1 type errors. Now you can disable passthrough so Better Auth serializes to ISO strings on writes and where operands, and parses on reads.

- Migration: If your SQLite schema stores timestamps as `text()`, set `supportsDates: false` in the Drizzle adapter config to avoid driver rejections.
- Caution: Do not set `supportsDates: false` when using Drizzle `timestamp` columns (PostgreSQL/MySQL) or SQLite `integer({ mode: "timestamp_ms" })`; those expect a `Date` and will fail.

<sup>Written for commit b375e39a4d2c0e62eab3d59c357f85cdba297745. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10898?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/drizzle-supports-dates.md
+++ b/.changeset/drizzle-supports-dates.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/drizzle-adapter": patch
+---
+
+Add a `supportsDates` option to the Drizzle adapter. Set it to `false` when your date columns are declared as `text()`, so Better Auth writes an ISO 8601 string instead of a raw `Date` that strict drivers such as Cloudflare D1 reject.
--- a/docs/content/docs/adapters/drizzle.mdx
+++ b/docs/content/docs/adapters/drizzle.mdx
@@ -211,6 +211,37 @@ export const auth = betterAuth({
 });
 ```
 
+## Text Date Columns
+
+Drizzle converts `Date` values itself before they reach the driver, so Better
+Auth hands it a `Date` and lets it do the work. That is what the generated
+schema relies on: `timestamp` on PostgreSQL and MySQL, and
+`integer({ mode: "timestamp_ms" })` on SQLite.
+
+A `text()` column has no such conversion. If you declared your date columns as
+`text()` instead, the raw `Date` reaches the driver, and a strict driver such as
+Cloudflare D1 rejects it:
+
+```
+D1_TYPE_ERROR: Type 'object' not supported for value 'Mon Aug 10 2026 14:47:14 GMT-0400'
+```
+
+Set `supportsDates` to `false` so Better Auth writes an ISO 8601 string and
+parses it back into a `Date` on read:
+
+```ts
+export const auth = betterAuth({
+  database: drizzleAdapter(db, {
+    provider: "sqlite",
+    supportsDates: false, // [!code highlight]
+  }),
+});
+```
+
+Leave it alone if your date columns came from the Better Auth CLI. Setting it to
+`false` against a `timestamp` or `integer({ mode: "timestamp_ms" })` column sends
+a string to a column that expects a `Date`, which fails.
+
 ## Custom Schema namespace
 
 If you're using PostgreSQL and you want to generate the schema with a custom schema namespace,
--- a/packages/drizzle-adapter/src/drizzle-adapter.test.ts
+++ b/packages/drizzle-adapter/src/drizzle-adapter.test.ts
@@ -6,6 +6,12 @@ import {
 	text,
 	timestamp,
 } from "drizzle-orm/pg-core";
+import type { SQLiteTable } from "drizzle-orm/sqlite-core";
+import {
+	integer as sqliteInteger,
+	sqliteTable,
+	text as sqliteText,
+} from "drizzle-orm/sqlite-core";
 import { describe, expect, it, vi } from "vitest";
 import { drizzleAdapter } from "./drizzle-adapter";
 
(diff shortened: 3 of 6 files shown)


dylanpulver:
Checked this — the code reading is right, but the path is not reachable, so I'd rather not widen the change.

The conversion is gated on `value instanceof Date`, so an array of `Date` would indeed pass through untouched. But `Where.value` cannot hold one (`packages/core/src/db/adapter/index.ts:329`):

```ts
value: string | number | boolean | string[] | number[] | Date | null;
```

There is no `Date[]` member, so `in: [date1, date2]` is not expressible against the public type.

Checking the call sites rather than trusting the type: every `in` / `not_in` usage in `packages/` targets a string or number field — `id`, `userId`, `scimUserId`, `connectionId`, `groupId`, `token`, `organizationId`, `role`, `clientId`, `resourceId`, `refreshId`, `provisioningDomainId`, `identifier`, `grantKey`. None is a `date` field.

The only path where the operator is caller-supplied is `admin/list-users`, and its `filterValue` is validated to `string | number | boolean | string[] | number[]` (`packages/better-auth/src/plugins/admin/routes.ts`) — plus it arrives as JSON, which has no `Date` literal.

Scalar date comparisons are the real usage and they are covered: e.g. `packages/api-key/src/routes/index.ts` does `{ field: "expiresAt", operator: "lt", value: new Date() }`, and the adapter test suite compares `createdAt` with `lt`/`lte`.

If `Date[]` is ever added to the `Where.value` union, the conversion should be extended at the same time. Until then this would be dead code.

Verification on this branch: `drizzle-adapter` 47/47 and `core` `factory.test.ts` 7/7 pass; upstream `main` is also green (`drizzle-adapter` 60/60), so no failure-set change. `biome check` over `packages/drizzle-adapter` and `packages/core` reports no findings across 174 files, and `tsc --noEmit` is clean.

The two failing Vercel checks are `Authorization required to deploy.` — a fork deploy gate that needs a maintainer to authorize, not something the branch can fix.

PR #10801, opened 2026-08-13
fix(types): make plugin declarations portable

bytaesu:
Projects can now use either or both of these TypeScript options when exporting Better Auth instances:

```json
{
  "compilerOptions": {
    "declaration": true,
    "composite": true
  }
}
```

This enables monorepos to build and share a dedicated `auth` package with portable TypeScript declarations, and benefits libraries and plugins built on top of Better Auth.

For example:

```text
packages/
  auth/
apps/
  web/
  api/
  admin/
```

Related real-world workarounds:

- https://github.com/stagewise-io/stagewise/commit/1966c66168082383d4804278094ebe4fb5d2c97b
- https://github.com/trycompai/comp/commit/7118c6e7db83752e82f2e28321bbbcbb1a4f510d

**_Before_**

before

**_After_**

after

---

- Closes #10532
- Closes #10556
- Closes #10710
- Closes #10560
- Supersedes #10636
- Supersedes #10689

---
## Summary by cubic
Standardizes plugin type declarations by introducing named public plugin interfaces across server and client. This keeps emitted `.d.ts` portable under `declaration` and `composite` and replaces `ReturnType<typeof plugin>` inference with stable interfaces; runtime behavior is unchanged.

- Core and external server plugins now use an internal `createXPlugin` factory and export named interfaces (for example `AdminPlugin`, `JwtPlugin`, `OAuthProviderPlugin`, `OAuthPopupPlugin`, `ApiKeyPlugin`, `PasskeyPlugin`, `SCIMPlugin`, `I18nPlugin`, `ExpoPlugin`, `ElectronPlugin`, `OAuthProviderResourceClientPlugin`, `SSOPlugin`, `SSODomainVerificationPlugin`). Client plugins’ `$InferServerPlugin` now reference these public interfaces and re-export them for consumer typing across `better-auth` and `@better-auth/*`.
- E2E typecheck adds composite and declaration fixtures and asserts emitted `.d.ts` import only public plugin types (including `@better-auth/oauth-provider/resource-client`) and do not reference build internals. Tests cover a wide set of server and client plugins.
- Docs clarify the factory + named-interface pattern and support for projects enabling `declaration`/`composite`. Patch releases across affected packages.

- Migration: no changes required for app code. For published custom plugins and their clients, export a named plugin interface and use it in `$InferServerPlugin` to keep declarations portable.

<sup>Written for commit fedeb4cd882df767c6194df0b9ba9c62a6c5e3fd. Summary will update on new commits.</sup>

[...]

PR #10907, opened 2026-08-20
fix(client): a client with more plugins fits a narrower client type again

### LINKED ISSUE #10897: 1.7: client with more plugins no longer assignable to a client type with fewer plugins

### Reproduction

```ts
import { adminClient, emailOTPClient } from "better-auth/client/plugins";
import { createAuthClient, type ReactAuthClient } from "better-auth/react";

// Shared package: a helper that needs only the email-OTP surface.
type EmailOtpClient = ReactAuthClient<{ plugins: [ReturnType<typeof emailOTPClient>] }>;
declare function useSignInOtp(client: EmailOtpClient): void;

// App: the real client has email-OTP plus one more plugin.
const client = createAuthClient({ plugins: [emailOTPClient(), adminClient()] });

useSignInOtp(client); // 1.6.28: ok — 1.7.1: TS2345
```

We hit this while upgrading a monorepo from 1.6.28. A shared package types its auth hooks against "a client with plugin X"; each app passes its own client, which has X plus app-specific plugins. On 1.6 that worked. On 1.7.1 every such annotation fails with TS2345, ending in:

> Types of property 'hydrateSession' are incompatible. [...] Property 'banned' is missing in type 'StripEmptyObjects<{ ... }>' but required in type 'StripEmptyObjects<{ ... } & { banned: ... }>'.

`hydrateSession` is new in 1.7. Its parameter is the session type inferred from the plugin list, so any plugin that widens the session breaks the match — in the repro it's the admin plugin adding `banned` to the user. The bigger client's `hydrateSession` only accepts sessions that include `banned`, while the narrow type permits calls without it. TypeScript flags that as unsafe and rejects the whole client.

Expected: a client that has at least the plugins named in the type still matches, as in 1.6.

Our workaround is `Pick`-ing the members the helpers use, e.g. `Pick<ReactAuthClient<...>, "emailOtp" | "signIn" | "$ERROR_CODES">`. It works, but it took a while to figure out what actually broke.

### Possible fixes

- Type `hydrateSession`'s parameter as the base session instead of the plugin-widened one. A setter can accept less-detailed input, so this removes the conflict without loosening any other check.
- Alternatively, declare `hydrateSession` with method syntax instead of an arrow-function property. TypeScript checks method parameters loosely (React's own types make the same choice), which restores the 1.6 behavior with no runtime change. Verified: switching the declaration style in a minimal model resolves the repro.

1.7.1 (worked on 1.6.28)

### System info

```
TypeScript 5.x, strict; monorepo with a shared package typing hooks against subsets of the app's client plugins
plugins in the failing app client: emailOTP, deviceAuthorization, admin
```

Types

```typescript
createAuthClient({ plugins: [emailOTPClient(), adminClient()] })
```

### Additional context

None.

heliohm:
Closes #10897. Option 1 from the issue, as @bytaesu suggested there.

hydrateSession is now declared as a method. TypeScript checks method parameters loosely, so a client with more plugins matches a narrower client type again, like on 1.6. No runtime change.

Added a type test for this. It fails on main and passes here.

I think the portable types plan in #10801 is the longer-term answer to the shared-package problem from the issue. This PR only restores the 1.6 behavior on the 1.7 track.

---
## Summary by cubic
Restores 1.6 assignability: clients created with more plugins match narrower client types again by declaring hydrateSession as a method. In 1.7 it was an arrow-function property, which made the session parameter contravariant and broke assignments; this restores the 1.6 behavior with no runtime change.

- Converts hydrateSession to a method in all `better-auth` clients: react, vue, svelte, solid, lynx, and vanilla.
- Adds a type test to ensure a multi-plugin client is assignable to a client type with fewer plugins.
- Patch release to `better-auth`; no migration required.

<sup>Written for commit 8d3d71d4c7f24bc4830543e0ff60b266a017299f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10907?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/hydrate-session-variance.md
+++ b/.changeset/hydrate-session-variance.md
@@ -2,4 +2,4 @@
 "better-auth": patch
 ---
 
-A client created with more plugins is again assignable to a client type declaring fewer plugins, as in 1.6. `hydrateSession` is now declared with method syntax, so its session parameter no longer blocks the assignment.
+A client created with more plugins is again assignable to a client type declaring fewer plugins, as in 1.6.
--- a/.changeset/hydrate-session-variance.md
+++ b/.changeset/hydrate-session-variance.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+A client created with more plugins is again assignable to a client type declaring fewer plugins, as in 1.6. `hydrateSession` is now declared with method syntax, so its session parameter no longer blocks the assignment.
--- a/packages/better-auth/src/client/client.test.ts
+++ b/packages/better-auth/src/client/client.test.ts
@@ -420,10 +420,10 @@ describe("run time proxy", async () => {
 });
 
 describe("type", () => {
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/10897
+	 */
 	it("keeps a client with more plugins assignable to a client type with fewer plugins", () => {
-		// Regression for #10897: hydrateSession as an arrow-function property
-		// made this check contravariant, so any plugin that widens the session
-		// (admin adds user.banned) broke the assignment.
 		type EmailOtpClient = ReactAuthClient<{
 			plugins: [ReturnType<typeof emailOTPClient>];
 		}>;
--- a/packages/better-auth/src/client/client.test.ts
+++ b/packages/better-auth/src/client/client.test.ts
@@ -17,6 +17,7 @@ import {
 	organizationClient,
 	twoFactorClient,
 } from "./plugins";
+import type { ReactAuthClient } from "./react";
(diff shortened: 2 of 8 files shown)


bytaesu (approved):
Thanks for working on this @heliohm 

LGTM!

Issue #9490, opened 2026-05-07
[Expo] expo-origin → Origin translation in onRequest doesn't reach origin-check middleware


narajaon:
### Setup

- `better-auth@1.4.18`
- `@better-auth/expo@1.4.18`
- React Native client (Expo) calling `authClient.signIn.social({ provider: "google", idToken: { token } })` against a server that has `expo()` registered in `betterAuth({ plugins: [...] })`.

### Symptom

The mobile request lands at `/api/auth/sign-in/social` with `expo-origin: hark-mobile://` (no `Origin` header — RN's fetch doesn't set one for non-http schemes). The server returns:

```json
{ "code": "MISSING_OR_NULL_ORIGIN", "message": "Missing or null Origin" }
```

with HTTP 403, even though `hark-mobile://` is in `trustedOrigins`.

### Expected

The Expo plugin's `onRequest` hook is documented as translating `expo-origin` → `Origin` server-side so the origin-check middleware passes:

```ts
// @better-auth/expo/dist/index.mjs
async onRequest(request, ctx) {
  if (options?.disableOriginOverride || request.headers.get("origin")) return;
  const expoOrigin = request.headers.get("expo-origin");
  if (!expoOrigin) return;
  const req = request.clone();
  req.headers.set("origin", expoOrigin);
  return { request: req };
}
```

### Actual

The cloned `req` with the new `Origin` header doesn't appear to flow into the origin-check middleware (`packages/better-auth/src/api/middlewares/origin-check.ts:validateOrigin`). The middleware reads `ctx.request.headers.get("origin")` and gets `null` because the cookie header is set (`useCookies = true`), triggering `MISSING_OR_NULL_ORIGIN`.

### Reproduction

1. Server: standard `betterAuth({ plugins: [expo(), ...], trustedOrigins: ["hark-mobile://"] })`.
2. Client (RN): `createAuthClient({ baseURL, plugins: [expoClient({ scheme: "hark-mobile", ... })] })` and call `signIn.social({ provider: "google", idToken: { token } })` with a real Google ID token.
3. Inspect outbound request: contains `expo-origin: hark-mobile://`, no `Origin`.
4. Server responds 403 `MISSING_OR_NULL_ORIGIN`.

### Workaround

Set `Origin` directly on the auth client's `fetchOptions.headers`:

```ts
createAuthClient({
  baseURL,
  fetchOptions: { headers: { Origin: "hark-mobile://" } },
[...]

Issue #7014, opened 2025-12-27
better-auth expo plugin expo-origin header bypass takes no effect because of setting readonly request headers


techzealot:
### Is this suited for github?

- [ ] Yes, this is suited for github

### To Reproduce

1. use regular better-auth expo config and use emailAndPassword plugin
2. config your trustedOrigins such as "my-app://"
3. call signIn or any other api
4. MISSING_OR_NULL_ORIGIN error occurred

### Current vs. Expected behavior

current:
expo-origin header override failed to  add  origin, thus cause MISSING_OR_NULL_ORIGIN error

expected:
the expo plugin add correct origin to avoid origin check error

### What version of Better Auth are you using?

1.4.7

### System info

```bash
1. macos
2. node v23.10.0
3. better-auth 1.4.6
4. expo 53.0.24
```

### Which area(s) are affected? (Select all that apply)

Package

### Auth config (if applicable)

```typescript
const options = {
  experimental: {
    joins: true
  },
  database: drizzleAdapter(db, {
    provider: "pg"
  }),
  
  baseURL: CONFIG.BETTER_AUTH_URL,

  emailAndPassword: {
    enabled: true,
    requireEmailVerification: false,
  },
  plugins: [
    expo(),
    admin(),
 ],

  trustedOrigins: [
    "http://localhost:3000",
    "http://localhost:8081", // Expo dev server
    "my-app://", // Expo app scheme
    "exp://", // Expo Go
  ]
} satisfies BetterAuthOptions;
```

### Additional context

I think the header bypass logic caused the problem:

https://github.com/better-auth/better-auth/blob/9c54c108db80c5e51e824242416b3f2554331b10/packages/expo/src/index.ts#L26-L43

the request.headers is readonly ,set takes no effect

I write a fix plugin to test the correct solution, it works:
```
export const expoOriginFix = (): BetterAuthPlugin => {
  return {
    id: "expo-origin-fix",
    onRequest: async (request) => {
     
      const expoOrigin = request.headers.get("expo-origin");

      if (!expoOrigin) {
       
        return;
      }

      // create new header because request.headers is readonly 
      const newHeaders = new Headers(request.headers);
      newHeaders.set("origin", expoOrigin)

      const newRequest = new Request(request, {
        headers: newHeaders
      });

      return {
        request: newRequest,
      };
    },
  };
};

```

bytaesu:
[...]

Issue #5573, opened 2025-10-25
Email & Password Sign Up/Logout throws 403 MISSING_OR_NULL_ORIGIN


royjosephargumido:
### Is this suited for github?

- [x] Yes, this is suited for github

### To Reproduce

Sign Out:
1. Upgrade to better-auth 1.3.31
3. Using Postman 11.68.5, clear all Cookies
4. Successfully sign in using Email and Password
5. After signing in, sign out.
6. Throws 403, `MISSING_OR_NULL_ORIGIN`.

After the above process, try signing in using Email and Password: 
1. Throws 403, `MISSING_OR_NULL_ORIGIN`.

### Current vs. Expected behavior

**Current Behavior:**
When attempting to sign in using Email and Password with better-auth version 1.3.31, the request fails with a `403 Forbidden` error. The response returns:

```
{
	"code": "MISSING_OR_NULL_ORIGIN",
	"message": "Missing or null Origin"
}
```

To temporarily resolve, I cleared all Postman cookies and retries the sign in. However, after successful sign in, the sign out throws the same error. Further, after successful sign in, attempting to sign in again using Email and Password throws the same error.

**Expected Behavior:**
Signing in using a valid and correct Email and Password should return a successful response, indicating a complete sign in process. Likewise, users should be able to sign out successfully. Furthermore, signing in again even if already signed in, should also result in a successful email sign in.

### What version of Better Auth are you using?

1.3.31

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Home",
    "release": "10.0.26200",
    "cpuCount": 8,
    "cpuModel": "11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz",
    "totalMemory": "23.73 GB",
    "freeMemory": "12.15 GB"
  },
  "node": {
    "version": "v24.9.0",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.6.0"
  },
  "frameworks": null,
  "databases": [
    {
      "name": "pg",
      "version": "^8.16.3"
    },
    {
      "name": "drizzle",
      "version": "^0.44.7"
    }
  ],
  "betterAuth": {
    "version": "^1.3.31",
    "config": {
      "appName": "redacted",
      "emailAndPassword": {
        "enabled": true,
        "requireEmailVerification": true
      },
      "plugins": [
        {
          "name": "open-api",
          "config": {
            "id": "open-api",
[...]

PR #5545, opened 2025-10-24
fix(expo): origin check failing due to null origin in expo

Bekacru:
closes #5536

---
## Summary by cubic
Fixes Expo origin check failures by overriding a null Origin header with a new expo-origin header, ensuring auth requests succeed in Expo environments.

- **Bug Fixes**
  - Set expo-origin in the Expo client and use it server-side to populate Origin when missing.
  - Added disableOriginOverride option to opt out of the override behavior.
  - Updated router to handle rate-limit middleware responses that return a request object.
  - Added test to confirm origin override works during email sign-in.

- **Dependencies**
  - Bumped Expo and related demo app packages; updated pnpm lockfile.

PR #10944, opened 2026-08-22
fix(origin-check): accept a same-origin form post whose Origin the browser nulled

himself65:
## The bug

`validateOrigin` rejects `Origin: null` outright:

```ts
if (!originHeader || originHeader === "null") {
    throw APIError.from("FORBIDDEN", BASE_ERROR_CODES.MISSING_OR_NULL_ORIGIN);
}
```

That treats `null` as always-foreign. It isn't. Per [Fetch — "Append a request `Origin` header"](https://fetch.spec.whatwg.org/#append-a-request-origin-header) step 3.1, the **initiating document's referrer policy** serializes the origin of a non-CORS non-`GET` request as `null` when that policy is `no-referrer`.

And `Referrer-Policy: no-referrer` is the standard recommendation for exactly the pages whose URL carries a secret — password reset, OAuth consent. So a first-party page, hardened per that recommendation, submitting an HTML form to its own better-auth backend arrives with `Origin: null` and gets a 403.

Two things make it hurt more than it looks:

1. **A request carrying cookies never reaches the Fetch Metadata branch at all.** In `validateFormCsrf`, cookies short-circuit straight into `validateOrigin`:

   ```ts
   const hasAnyCookies = headers.has("cookie");
   if (hasAnyCookies) {
       return await validateOrigin(ctx);
   }
   ```

   So the signed-in form post — consent, re-auth, account settings — is decided purely by the Origin value, which in this case cannot carry the information.

2. **`fetch()` never reproduces it.** Mode `"cors"` always carries a real Origin regardless of referrer policy, so only the plain-HTML-form path breaks — the progressive-enhancement path this middleware exists to support, and the one with no client-side JS to debug it with. The failure lands on the last click of the flow.

## The fix

```ts
if (originHeader === "null" && headers.get("sec-fetch-site") === "same-origin") {
    return;
}
```

`Sec-Fetch-Site` settles it without loosening anything else. It is a **forbidden header name**, so no page can set it, and `same-origin` is the browser's own statement that the initiator's origin *is* this request's origin — precisely the question `trustedOrigins` is asking. An opaque origin (sandboxed iframe, `data:`/`blob:` document) is never same-origin with the target, so it cannot reach this branch.

Deliberately kept as narrow as possible — nothing that passes or fails today changes:

| Case | Before | After |
|---|---|---|
| `Origin: null` + `Sec-Fetch-Site: same-origin` | 403 | **allowed** |
| `Origin: null` + `cross-site` | 403 | 403 |
| `Origin: null` + `same-site` | 403 | 403 |
| `Origin: null`, no Fetch Metadata | 403 | 403 |
| No `Origin` at all (even with `same-origin`) | 403 | 403 |

The last row is intentional: this PR only rescues the case the spec proves a browser produces, and leaves the absent-Origin path exactly as it was.

## Tests

Five tests in `origin-check.test.ts` under `Fetch Metadata CSRF Protection`, one per row above. Verified that removing the fix fails **only** the accepting test — the four rejection tests pass with and without it, so they pin behavior rather than just following the change.

```
Tests  43 passed (43)          # src/api/middlewares/origin-check.test.ts
Tests  427 passed (427)        # src/api
```

`pnpm typecheck` passes. Full `packages/better-auth` suite is 2589 passed with 2 pre-existing failures (`db.test.ts` and `oauth-proxy.test.ts` postgres cases, which need the Docker containers) — confirmed identical on a clean tree with my changes stashed.

## Context

Reported downstream: an app's OAuth consent screen sets `no-referrer` to keep `code_challenge`/`state` out of Referer, and its own CSRF gate — written to the same shape as this one — 403'd every approval for a week. Every MCP client at once, and the `fetch()`-based tests were all green.

Worth noting that `MISSING_OR_NULL_ORIGIN` has generated a steady stream of reports — #5536, #5545, #5573, #5765, #7014, #9490 — but every one of those is Expo, Postman, or cURL. The browser `no-referrer` path doesn't seem to have been reported before, which is probably why `null` reads as "always foreign" in the current code.

---
## Summary by cubic
Accepts same-origin HTML form posts whose browser sent Origin: null under Referrer-Policy: no-referrer. Previously these requests returned 403 (MISSING_OR_NULL_ORIGIN); now they pass only when Sec-Fetch-Site is same-origin, with all other null-or-absent Origin cases unchanged.

**Review notes**
- Updates `validateOrigin` in `packages/better-auth/src/api/middlewares/origin-check.ts` to allow Origin: null only when `Sec-Fetch-Site: same-origin`; no other logic paths change.
- Security posture unchanged: `Sec-Fetch-Site` is a forbidden header and cannot be spoofed; absent Origin still rejects.
- Affects HTML form navigation flows; `fetch()` callers are unaffected.
- Adds 5 tests in `packages/better-auth/src/api/middlewares/origin-check.test.ts` covering the accepted case and four rejections.
- No configuration or migration changes required in `better-auth`.

<sup>Written for commit 4250099d4364faab8cac9f5e178c906cec2c1426. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10944?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/null-origin-same-origin-form-post.md
+++ b/.changeset/null-origin-same-origin-form-post.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Accept a same-origin form submission whose `Origin` the browser nulled. A page served `Referrer-Policy: no-referrer` — the recommended policy for screens whose URL carries a secret, such as password reset and OAuth consent — sends the literal `Origin: null` on an HTML form it navigates with, and `validateOrigin` rejected it as `MISSING_OR_NULL_ORIGIN`. `Sec-Fetch-Site: same-origin` now vouches for that case. Every other `null` Origin still fails: a sandboxed iframe, a `data:` document, or a request with no Fetch Metadata at all. `fetch()` callers are unaffected either way.
--- a/packages/better-auth/src/api/middlewares/origin-check.test.ts
+++ b/packages/better-auth/src/api/middlewares/origin-check.test.ts
@@ -553,6 +553,79 @@ describe("Fetch Metadata CSRF Protection", async () => {
 			);
 		}
 	});
+
+	/**
+	 * `Referrer-Policy: no-referrer` is the recommended policy for any page whose
+	 * URL carries a secret — password reset, OAuth consent. A page serving it
+	 * sends the LITERAL `Origin: null` on the HTML form it navigates with, so the
+	 * Origin value alone cannot distinguish a first-party form post from a
+	 * sandboxed iframe. `Sec-Fetch-Site` is what separates them.
+	 *
+	 * @see https://fetch.spec.whatwg.org/#append-a-request-origin-header
+	 */
+	describe("null Origin from a page served Referrer-Policy: no-referrer", () => {
+		const formPost = (headers: Record<string, string>) =>
+			new Request("http://localhost:3000/api/auth/sign-in/email", {
+				method: "POST",
+				headers: {
+					"content-type": "application/json",
+					// Signed-in surfaces are where `no-referrer` pages live, and a
+					// request carrying cookies is the path that reaches `validateOrigin`
+					// directly rather than the Fetch Metadata branch.
+					cookie: "better-auth.session_token=whatever",
+					"Sec-Fetch-Mode": "navigate",
+					"Sec-Fetch-Dest": "document",
+					...headers,
+				},
+				body: JSON.stringify({
+					email: testUser.email,
+					password: testUser.password,
+				}),
+			});
+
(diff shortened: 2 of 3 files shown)


bytaesu:
Thanks, Alex! I just opened a new PR with you as a co-author.
Closing this PR in favor of https://github.com/better-auth/better-auth/pull/10959.

PR #10956, opened 2026-08-23
fix(oauth-provider): accept host-bearing custom-scheme native redirects

### LINKED ISSUE #10946: OAuth provider rejects Cursor's host-bearing cursor:// redirect URI

### Reproduction

Configure `@better-auth/mcp` with unauthenticated DCR enabled, then connect Cursor to the MCP resource. Cursor dynamically registers approximately this metadata:

```json
{
  "application_type": "web",
  "redirect_uris": ["cursor://anysphere.cursor-mcp/oauth/callback"],
  "token_endpoint_auth_method": "none",
  "grant_types": ["authorization_code", "refresh_token"],
  "response_types": ["code"]
}
```

Steps:

1. Enable the OAuth provider / MCP plugin with dynamic client registration.
2. Register a native public client whose `redirect_uris` include `cursor://anysphere.cursor-mcp/oauth/callback` (Cursor itself currently sends `application_type: web`; coerce or register as `native` first — this issue is the native validator, not web→native coercion).
3. Observe DCR return `invalid_redirect_uri`.

Cursor's client-side `application_type: web` bug is tracked separately: https://forum.cursor.com/t/cursor-does-not-send-application-type-native-when-registering-mcp-oauth-clients/136907

Following the steps above, I expected a native public client with a non-reserved custom-scheme redirect (including Cursor's host-bearing `cursor://` callback) to register successfully. Dangerous and network schemes (`javascript:`, `data:`, `http:`, `https:`, `ws:`, `wss:`, `file:`, …) must remain rejected, and the authorization request must still exactly match the registered URI.

What I observed: after treating the client as `native`, Better Auth 1.7.1 still rejects `cursor://anysphere.cursor-mcp/oauth/callback` because the native private-use validator requires a reverse-domain scheme and requires the naming authority/host to be omitted (`com.example.app:/callback`).

1.7.1 (`better-auth`, `@better-auth/mcp`, `@better-auth/oauth-provider`)

### System info

```bash
# Library-level DCR validator bug; not an app `betterAuth({...})` config issue.
# `npx auth info --json` is not applicable (no consumer app auth config).
Node.js: v24.19.0 (darwin arm64)
OS: macOS 15.5 (darwin 25.5.0)
better-auth: 1.7.1
@better-auth/mcp: 1.7.1
@better-auth/oauth-provider: shipped with 1.7.1
Cursor callback: cursor://anysphere.cursor-mcp/oauth/callback
```

- Backend
- Package

```ts
import { mcp } from "@better-auth/mcp";

  plugins: [
    mcp({
      loginPage: "/sign-in",
      // unauthenticated DCR enabled for MCP clients
    }),
  ],
});
```

### Additional context

RFC 8252 §7.1 *recommends* the authority-free reverse-domain form; it does not forbid host-bearing private-use URIs. Native MCP clients such as Cursor register the latter.

A scratch MCP authorization server interoperates with Cursor only by (1) coercing the DCR client to `application_type: native` in gateway middleware, and (2) accepting host-bearing custom-scheme redirects while still rejecting reserved schemes. With that, Cursor completes discovery → DCR → PKCE → consent → token → MCP tool call.

Fix: https://github.com/better-auth/better-auth/pull/10956

disbell:
## Summary

Native dynamic client registration currently requires private-use redirect URIs to be an **authority-free reverse-domain** URI (`com.example.app:/callback`). RFC 8252 §7.1 *recommends* that shape; it does not forbid a host.

Native MCP clients such as Cursor register URIs like `cursor://anysphere.cursor-mcp/oauth/callback`. Those fail `isReverseDomainPrivateUseRedirectUri` on three counts (host present, `//` form, scheme is not reverse-domain), so DCR returns `invalid_redirect_uri`.

This PR adds a **sibling** predicate, `isHostBearingPrivateUseRedirectUri`, and accepts either form for **native** clients. It does **not**:

- change `isReverseDomainPrivateUseRedirectUri`
- coerce `application_type: web` → `native` (Cursor sending `web` with a custom scheme remains a client bug)
- relax the reserved-scheme denylist (`file:`, `javascript:`, `data:`, `mailto:`, …)

CIMD documents that omit `application_type` already validate as native, so they pick up the host-bearing form automatically.

Closes #10946

## Test plan

- [x] `packages/core` unit tests for both predicates
- [x] `packages/oauth-provider` DCR: `cursor://…` and `com.example.app://host/callback` return 201 for `application_type: native`
- [x] RFC form `com.example.app:/callback` still 201
- [x] `myapp:/callback` (no host, not reverse-domain) still 400
- [x] reserved schemes still 400
- [x] same `cursor://` URI still 400 for `application_type: web` (no coercion)

## Notes for reviewers

**Correction (thanks @0-uddeshya-0):** current Cursor builds (observed on 3.13.25, macOS) **omit** `application_type` rather than send `web` — my original `web` observation was from an earlier build and not a controlled capture. An omitted value defaults to `web` per OIDC Registration §2, so the validator fix alone does not unblock a client that omits the field.

This PR therefore also adds `clientRegistrationDefaultApplicationType` (`"web"` default | `"native"` | `"infer"`): server-owned policy for the **absent** field only. `"infer"` classifies a dynamic registration as `native` only when every `redirect_uri` uses a non-http(s) scheme. A sent `application_type` is never overridden (no `web`→`native` coercion), managed clients and CIMD are unaffected, and the strict `"web"` default keeps stock behavior unchanged. Docs note that accepting period-free schemes like `cursor:` deviates from the RFC 8252 §7.1 SHOULD, mitigated by mandatory public-client PKCE.

Cursor's client-side half (sending a proper `application_type`) is tracked on their forum: https://forum.cursor.com/t/cursor-does-not-send-application-type-native-when-registering-mcp-oauth-clients/136907

Made with [Cursor](https://cursor.com)

---
## Summary by cubic
Accepts host-bearing custom-scheme redirect URIs for native OAuth clients and CIMD documents, and adds `clientRegistrationDefaultApplicationType` so current Cursor builds (which omit `application_type`) can register `cursor://` callbacks. Previously only authority-free reverse-domain URIs were accepted and omitted `application_type` always defaulted to `web`.

- `@better-auth/core`: adds `isHostBearingPrivateUseRedirectUri` and `isNativePrivateUseRedirectUri`, and treats `ws:`/`wss:` as non-private-use schemes.
- `@better-auth/oauth-provider`: validates native redirects with the union, rejects `ws:`/`wss:`, and keeps `web` clients with custom schemes returning `invalid_redirect_uri`.
- `@better-auth/oauth-provider`: adds `clientRegistrationDefaultApplicationType` (`"web"` default, `"native"`, or `"infer"`) for registrations that omit `application_type`; a sent `application_type` is never overridden and CIMD/managed clients are unaffected.
- `@better-auth/cimd`: validates redirects with the native union and keeps host-bearing `post_logout_redirect_uris` origin-bound.
- Docs and tests updated; reverse-domain authority-free remains the recommended private-use shape.

<sup>Written for commit 236e353c5020d802bf94ee3f0ac949d1f29eb9b4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10956?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


disbell:
Greptile's CIMD note is from the first commit only (`cb077163`). The second commit (`0fbe9bc2`) already switches CIMD metadata validation to `isNativePrivateUseRedirectUri` and adds Cursor-style `cursor://` coverage in `packages/cimd`.

Follow-up just pushed: reject `ws://` / `wss://` as well (Cubic P2). Please re-review head.

disbell:
Follow-up at `438103788`:

- CIMD origin-binding now skips only for authorization `redirect_uris`. Host-bearing `post_logout_redirect_uris` stay origin-bound (Greptile’s remaining note).
- Dedicated Cursor DCR test now asserts `application_type: native` and the registered URI; dropped the duplicate accept-list entry.
- Documented that host-bearing private-use URIs require a path (`cursor://host` / query-only forms stay rejected).

Local `pnpm typecheck` is green on Node 24. Package tests for `core` / `oauth-provider` / `cimd` / `mcp` pass. Full `pnpm test` still has unrelated local failures (`redis-storage` missing vitest project `docs`, and `@better-auth/sso` SAML fixture generation returning undefined) — not caused by this change.

0-uddeshya-0:
Independent confirmation from a third deployment, plus one data point that may affect the "Notes for reviewers" section.

**Setup:** an MCP server on `better-auth` / `@better-auth/mcp` / `@better-auth/oauth-provider` 1.7.2, running on Cloudflare Workers with Neon Postgres. Same symptom as reported: Cursor could not register, and the client surfaced it only as a generic connection failure.

**The data point: current Cursor appears to OMIT `application_type` rather than send `web`.**

The issue body and this PR both state that Cursor registers with `application_type: "web"`. On Cursor 3.13.25 (macOS) I could not reproduce that. I tested with a local patch whose *only* effect on classification is the value passed as the **default** to `applyOAuthClientRegistrationDefaults` — it never overrides a value the client actually sent, since that helper resolves `client.application_type ?? default`.

Two registrations against that server, everything else identical:

```
# application_type omitted
POST /oauth2/register {"redirect_uris":["cursor://anysphere.cursor-mcp/oauth/callback"],
                       "token_endpoint_auth_method":"none", ...}
-> 201, application_type: "native"

# application_type explicitly "web"
POST /oauth2/register {"application_type":"web",
                       "redirect_uris":["cursor://anysphere.cursor-mcp/oauth/callback"], ...}
-> 400 invalid_redirect_uri
   "web clients require https redirect URIs on non-loopback hosts: cursor://anysphere.cursor-mcp/oauth/callback"
```

The real Cursor client then registered successfully against that same server and was stored as `application_type: native`, `token_endpoint_auth_method: none`, with `cursor://anysphere.cursor-mcp/oauth/callback` among its redirect URIs. Since an explicitly-sent `"web"` is provably still rejected there, Cursor cannot have sent `"web"` — it must have omitted the field.

Before the change, that same client failed with the **web-branch** error above, which is what an omitted `application_type` produces via the OIDC DCR default.

**Why this matters here:** if current Cursor omits the field, then this PR's validator fix alone will not unblock it — the omitted value defaults to `web`, and the https-only rule rejects the URI before `isHostBearingPrivateUseRedirectUri` is ever consulted. That would make the "and either Cursor sending `native` or the AS rewriting the registration" caveat the binding constraint for every operator on a current Cursor build, rather than an edge case.

It might be worth a `defaultApplicationType` option (`"web"` default, plus an opt-in `"infer"` that classifies a registration as native only when *every* `redirect_uri` is a non-http(s) scheme), so operators can close the remaining half without patching `node_modules`. That would keep this PR's scope intact and leave strict behaviour as the default.

Two caveats on my side: Cursor may simply have changed between builds, so both observations can be correct for different versions — worth pinning the version in the issue either way. And I have not yet run this against your branch; happy to test `cursor://…` end to end (discovery → DCR → PKCE → consent → token → `tools/call`) against it and report back if that is useful.

For reference, RFC 8252 §7.1 also says authorization servers "SHOULD" reject private-use schemes containing no period, which `cursor:` does not have. So accepting it is a deliberate deviation from that recommendation rather than a pure bug fix — defensible, in my view, precisely because this provider already refuses any public-client authorize request without `code_challenge` (`pkce is required for public clients`), which is the control that makes an intercepted code unusable. Might be worth stating explicitly in the docs for the option.

disbell:
@0-uddeshya-0 Thank you — this is an excellent catch, and your explicit-`web` A/B is more rigorous than my original observation. Confirmed against this branch: dynamic DCR defaults an omitted `application_type` to `web` (`applyOAuthClientRegistrationDefaults(..., "web")` in `packages/oauth-provider/src/register.ts`), so the web-branch https rule fires before the new predicate is ever consulted; only CIMD registrations preserved the absence. On a Cursor build that omits the field, the validator fix alone did not unblock plain DCR — you're right that the caveat was the binding constraint, not an edge case.

**Your suggestion is now implemented on this branch** (`236e353c5`):

- `clientRegistrationDefaultApplicationType?: "web" | "native" | "infer"` on the provider options (flows through `@better-auth/mcp` automatically since `McpOptions extends OAuthOptions`).
- `"web"` remains the default — stock behavior is byte-for-byte unchanged. `"infer"` classifies a dynamic registration as `native` only when **every** `redirect_uri` uses a non-http(s) scheme, exactly as you proposed.
- A sent `application_type` is never overridden (it only fills the absent field via `client.application_type ?? default`), and it applies to dynamic registration only — managed clients and CIMD are untouched.
- Tests cover: omitted + `cursor://` → 201 `native` under `infer`; mixed custom+https set stays `web` (and 400s); explicit `web` + `cursor://` still 400s under `infer`; unset option keeps the strict default.
- Docs now state the RFC 8252 §7.1 deviation explicitly: accepting period-free schemes like `cursor:` departs from the SHOULD, mitigated by this provider's mandatory PKCE for public-client authorization requests (`isPKCERequired` returns required unconditionally for `token_endpoint_auth_method: "none"`).

I've also corrected the PR description: the unqualified "Cursor sends `application_type: web`" claim is gone, replaced with your pinned observation (omitted on 3.13.25) — my original was from an earlier build and not a controlled capture, so both may have been true across versions, as you said.

Your end-to-end offer would be very useful — especially discovery → DCR → PKCE → consent → token → `tools/call` on 3.13.25 against this branch with `clientRegistrationDefaultApplicationType: "infer"` set, since that's the configuration that proves the full path without any `node_modules` patching. I'd be glad to cite the result here.

PR #10973, opened 2026-08-24
fix(cli): emit defineRelationsPart for drizzle-orm 1.0 in generate

### LINKED ISSUE #10924: generate: emitted schema uses relations(), which drizzle 1.0 removed

### Reproduction

Install the CLI (the `auth` package, 1.7.1, which ships the `better-auth` bin) together with `better-auth@1.7.1` and `drizzle-orm@1.0.0-rc.4` (the `rc` tag).

```ts
// auth.ts

  database: drizzleAdapter({}, { provider: "sqlite" }),
});
```

```
npx better-auth generate --config auth.ts --output schema.ts
npx tsc --noEmit schema.ts
```

The generator emits the same output on any drizzle version; the failure appears when the file is compiled against drizzle 1.0. The generated schema ends with the old relations API:

```ts
import { relations } from "drizzle-orm";

export const userRelations = relations(user, ({ many }) => ({
  sessions: many(session),
}));
```

Drizzle 1.0 removed `relations` and replaced it with `defineRelations`, so the generated file does not compile:

> '"drizzle-orm"' has no exported member named 'relations'. Did you mean 'Relation'?

The tables themselves generate fine. Expected: when drizzle 1.0 is installed, emit the new `defineRelations` form, or skip the relations blocks. The same relations in the 1.0 form, verified to compile against rc.4:

```ts
import { defineRelations } from "drizzle-orm";

export const authRelations = defineRelations({ user, session }, (r) => ({
  user: {
    sessions: r.many.session({ from: r.user.id, to: r.session.userId }),
  },
  session: {
    user: r.one.user({ from: r.session.userId, to: r.user.id }),
  },
}));
```

Better Auth 1.7 supports drizzle 1.0 at runtime (the peer range allows `>=1.0.0-rc.1`), but generation was not updated with it.

auth CLI 1.7.1, better-auth 1.7.1, drizzle-orm 1.0.0-rc.4

### System info

```
drizzle sqlite adapter (D1)
```

CLI

```typescript
betterAuth({ database: drizzleAdapter(db, { provider: "sqlite" }), plugins: [organization(), admin(), emailOTP(), deviceAuthorization(), bearer()] })
```

### Additional context

None.

houssaineamzil:
## Summary

Fixes #10924.

`generate` for the Drizzle adapter always emitted the old `relations()` API. Drizzle 1.0 removed `relations()` in favor of `defineRelations`/`defineRelationsPart`, so the generated `auth-schema.ts` fails to compile once a project installs drizzle-orm 1.0 (`'"drizzle-orm"' has no exported member named 'relations'`), even though better-auth's peer range already allows `>=1.0.0-rc.1`.

- Added `getDrizzleVersion(cwd)` to read the project's declared `drizzle-orm` version.
- Threaded `cwd` through the schema generator so it can detect the installed version.
- When drizzle-orm `>=1.0.0` is detected, `generateDrizzleSchema` now emits a single `defineRelationsPart({...}, (r) => ({...}))` block (the same shape already used by `@better-auth/drizzle-adapter/relations-v2`'s generator) instead of per-table `relations()` calls. Below 1.0, or when no `cwd` is available, output is unchanged.

## Test plan

- [x] New tests in `packages/cli/test/drizzle-relations-v2.test.ts`: drizzle-orm 1.0 detection, <1.0 fallback, no-`cwd` fallback, duplicate-relation disambiguation without `relationName`.
- [x] Existing `generate`/`generate-all-db` suites pass unchanged (no snapshot diffs).
- [x] Full `cli` package test suite passes (364 passed, 1 pre-existing skip).
- [x] `tsc` typecheck of the `cli` package passes.
- [x] Manually reproduced the issue: ran the built CLI against a project with `drizzle-orm@1.0.0-rc.4` installed (matching the issue) and confirmed the generated `auth-schema.ts` now type-checks cleanly with real `tsc` against that install.

---
## Summary by cubic
Generates Drizzle relations that compile with `drizzle-orm` 1.0+. Previously the CLI always emitted per-table `relations()`; for 1.0+ it now emits a single `defineRelationsPart({...}, (r) => ({...}))`. Output for `<1.0` or when `cwd` is unknown is unchanged.

- Detects the installed `drizzle-orm` major via `getDrizzleVersion(cwd)`; prefers the version in `node_modules/drizzle-orm/package.json` and falls back to the declared range. Threads `cwd` through `generate` and `init`.
- Switches imports between `relations` and `defineRelationsPart` and renders v2 relations with explicit from/to on both sides. Duplicate FKs to the same model are disambiguated by distinct keys (no `relationName`).
- Migration (projects on `drizzle-orm` 1.0+): if you import per-table `*Relations`, change to the single `authRelations` export.

<sup>Written for commit 339782de2942e156b653ec0cad6192df5ee91fd4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10973?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/drizzle-1.0-relations-generate.md
+++ b/.changeset/drizzle-1.0-relations-generate.md
@@ -1,0 +1,5 @@
+---
+"auth": patch
+---
+
+Fix `generate` emitting the removed Drizzle `relations()` API when drizzle-orm 1.0 is installed. The generator now detects the project's drizzle-orm version and emits `defineRelationsPart` instead so the generated schema compiles against drizzle-orm 1.0+.
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -197,6 +197,7 @@ async function generateAction(opts: any) {
 		adapter,
 		file: resolvedOutputPath ?? options.output,
 		options: config,
+		cwd,
 	});
 
 	spinner.stop();
--- a/packages/cli/src/commands/init/index.ts
+++ b/packages/cli/src/commands/init/index.ts
@@ -1169,6 +1169,7 @@ export const auth = betterAuth({
 					adapter,
 					options: config,
 					file: outputPath,
+					cwd,
 				});
 			} else if (isPrisma) {
 				// For Prisma, output to prisma/schema.prisma
--- a/packages/cli/src/generators/drizzle.ts
+++ b/packages/cli/src/generators/drizzle.ts
@@ -10,6 +10,7 @@ import type { BetterAuthDBSchema, DBFieldAttribute } from "better-auth/db";
 import { getAuthTables } from "better-auth/db";
 import type { BetterAuthOptions } from "better-auth/types";
 import prettier from "prettier";
+import { getDrizzleVersion } from "../utils/get-package-info";
 import type { SchemaGenerator } from "./types";
 
 function convertToSnakeCase(str: string, camelCase?: boolean) {
@@ -37,6 +38,7 @@ export const generateDrizzleSchema: SchemaGenerator = async ({
 	options,
 	file,
 	adapter,
+	cwd,
 }) => {
 	const tables = getAuthTables(options);
 	const filePath = file || "./auth-schema.ts";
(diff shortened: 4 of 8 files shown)


houssaineamzil:
Issue 1 (declared range) fixed. Issue 2 (init generates before installing deps) is real but not live today — `drizzle-orm`'s `latest` tag is still 0.x, and fixing it means reordering `init`'s install/generate flow, which is a bigger UX change I'd rather leave as a follow-up than bundle into this fix.

PR #10978, opened 2026-08-24
fix(generic-oauth): skip providers whose discovery fails

### LINKED ISSUE #10961: One generic OAuth provider's failed discovery permanently 500s the entire auth API, including email/password sign-up

### Reproduction

Deterministic. No database or configuration beyond the memory adapter.

```bash
mkdir /tmp/ba-blast-radius && cd /tmp/ba-blast-radius
npm init -y && npm install better-auth@1.7.1
# save the script below as repro.mjs
node repro.mjs
```

```js
import { genericOAuth } from "better-auth/plugins";

const store = { user: [], session: [], account: [], verification: [] };

const auth = betterAuth({
  secret: "reproduction-only-secret-reproduction-only-secret",
  database: memoryAdapter(store),
  plugins: [
    genericOAuth({
      config: [
        {
          providerId: "broken-idp",
          clientId: "client",
          clientSecret: "secret",
          // Unreachable on purpose (port 1 refuses connections instantly).
          discoveryUrl: "http://127.0.0.1:1/.well-known/openid-configuration",
        },
      ],
    }),
  ],
});

const signUp = (email) =>
  auth.api
    .signUpEmail({ body: { email, password: "correct-horse-battery-staple", name: "Repro" } })
    .then(() => "OK")
    .catch((err) => `ERROR: ${err.message}`);

console.log("attempt 1 (email/password sign-up, unrelated to the OAuth provider):", await signUp("a@example.test"));
console.log("attempt 2 (retry, ~1s later):                                       ", await signUp("b@example.test"));
console.log("attempt 3 (retry, ~2s later):                                       ", await signUp("c@example.test"));
console.log("\nusers actually created in the store:", store.user.length);
```

Actual output on `better-auth@1.7.1`:

```
ERROR [Better Auth]: Discovery fetch failed for "broken-idp": TypeError: fetch failed
attempt 1 (email/password sign-up, unrelated to the OAuth provider): ERROR: Provider "broken-idp": discovery returned no valid data. Provider initialization stopped to keep its account issuer stable.
attempt 2 (retry, ~1s later):                                        ERROR: Provider "broken-idp": discovery returned no valid data. Provider initialization stopped to keep its account issuer stable.
attempt 3 (retry, ~2s later):                                        ERROR: Provider "broken-idp": discovery returned no valid data. Provider initialization stopped to keep its account issuer stable.

users actually created in the store: 0
```

Only `broken-idp`'s `discoveryUrl` was ever unreachable. `signUpEmail` never touches that provider, yet it fails every single time — including the third attempt, seconds after the first, by which point a real network blip would typically have cleared.

**Current.** `betterAuth(options)` calls its `init` function exactly once and keeps the resulting `Promise<AuthContext>` in one closure variable, reused by every request for the lifetime of the process:

https://github.com/better-auth/better-auth/blob/v1.7.1/packages/better-auth/src/auth/base.ts#L18-L30

```ts
const authContext = initFn(options);
const { api } = getEndpoints(authContext, options);
...
const handler = async (request: Request) => {
    const ctx = await authContext;
    ...
```

That single promise resolves by running every plugin's `init` hook in a sequential loop, with no isolation between plugins:

https://github.com/better-auth/better-auth/blob/v1.7.1/packages/better-auth/src/context/helpers.ts#L23-L37

```ts
export async function runPluginInit(context: AuthContext) {
    ...
    for (const plugin of plugins) {
        if (plugin.init) {
            const initPromise = plugin.init(context);
            ...
```

The `generic-oauth` plugin's `init` hook throws when a provider's discovery fetch fails and no `accountIssuer` fallback is configured:

https://github.com/better-auth/better-auth/blob/v1.7.1/packages/better-auth/src/plugins/generic-oauth/index.ts#L254-L257

```ts
} else if (!c.accountIssuer) {
    throw new Error(
        `Provider "${c.providerId}": discovery returned no valid data. Provider initialization stopped to keep its account issuer stable.`,
    );
}
```

Because a JS promise only ever settles once, that throw permanently rejects the **one** `authContext` promise every request awaits. From that point on — for the rest of the process's life, regardless of how many requests come in or how long is waited between them — `await authContext` re-throws the same rejection immediately, without ever re-attempting discovery. Every endpoint fails: not just this provider's own sign-in, but email/password sign-up, sign-in, session lookup, everything the `better-auth` instance serves. There is no retry, backoff, or recovery path short of restarting the process at a moment when the IdP happens to be reachable.

**Expected.** A provider that cannot be initialized should be unavailable — for OAuth flows that target that provider, specifically. It should not be able to affect unrelated endpoints (email/password, other providers, session/admin/organization plugin routes, etc.), and a transient IdP outage shouldn't require restarting the whole auth server to recover once it's over.

Notably, the very PR that introduced this rewrite originally described the intended behavior as a log, not a hard failure ("Startup diagnostics: error-level log when discovery fails without fallback endpoints" — PR #9069). The unconditional `throw` was tightened in a later PR (#10403, "scope accounts by issuer") whose purpose was fixing an unrelated account-linking vulnerability (#10214: an OAuth account could be linked to two different users). That change is a reasonable fix for the vulnerability it targets, but as a side effect it turned "this provider can't establish a stable identity" into "the whole server is down," because nothing in `runPluginInit`/`createBetterAuth` isolates one plugin's `init` failure from the rest.

1.7.1

### System info

```json
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "version": "[redacted]",
    "release": "[redacted]",
    "cpuCount": "[redacted]",
    "cpuModel": "[redacted]",
    "totalMemory": "[redacted]",
    "freeMemory": "[redacted]"
  },
  "node": {
    "version": "v24.16.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "11.21.0"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null
  }
}
```

Backend

```typescript
import { genericOAuth } from "better-auth/plugins"

  emailAndPassword: {
    enabled: true
  },
  plugins: [
    genericOAuth({
      config: [
        {
          providerId: "my-idp",
          clientId: process.env.OIDC_CLIENT_ID,
          clientSecret: process.env.OIDC_CLIENT_SECRET,
          discoveryUrl: process.env.OIDC_DISCOVERY_URL,
          // No accountIssuer set — this is the state most configs are in
          // unless someone has specifically read the discovery-failure
          // behavior documented on the generic-oauth plugin page and gone
          // out of their way to add one.
        }
      ]
    })
  ]
})
```

Any deployment where the discovery URL is momentarily unreachable when the process (re)starts — a DNS hiccup, the IdP restarting at the same time, a firewall rule not yet applied — hits this. It doesn't require misconfiguration, only bad timing.

### Additional context

I could not find an existing issue describing this specific failure mode (searched by provider/discovery/init-failure keywords, and read through the PRs that shaped this code path — #9069, #10403, and #10403's source issue #10214 — without finding this blast-radius consequence discussed anywhere).

The `generic-oauth` plugin's own reference docs do mention that "Provider initialization fails if discovery cannot establish a valid configuration," and document `accountIssuer` as the way to give a provider a stable identity independent of live discovery. Setting it does prevent this specific throw. But nowhere — not the plugin docs, not the 1.7 upgrade guide — does it say that this failure isn't scoped to the provider: it takes the entire `better-auth` instance down for every caller, for as long as the process runs, with no self-healing. That's the part that made this take a while to diagnose in production: the visible symptom (email/password sign-up 500ing) has nothing to do with the actual cause (an unrelated OAuth provider we don't even use in that deployment).

Two independent things seem worth fixing, and either would resolve the practical impact:

1. **Isolate plugin `init` failures.** `runPluginInit` could catch a failing plugin's `init` and either drop that plugin (with the error surfaced through logging/diagnostics, matching the "error-level log" language originally intended in #9069) or otherwise avoid poisoning the shared `authContext` promise for everyone else.
2. **Don't let a permanently-rejected promise stay permanently rejected.** Even independent of (1), retrying `initFn(options)` on the next request after a failure (rather than caching the rejection forever) would let a deployment recover once the IdP comes back, instead of needing a process restart.

BetterAndBetterII:
## Summary
- Failed generic-oauth discovery used to throw during plugin `init`, permanently rejecting the shared `authContext` promise so unrelated endpoints (including email/password sign-up) returned errors for the rest of the process.
- Skip the provider that cannot establish a stable issuer instead of taking down the whole auth API. Sibling providers and email/password keep working.
- Docs now describe the skip rather than a process-wide init failure.

Fixes #10961

## Test plan
- [x] Regression: one unreachable discovery URL plus a healthy sibling; `signUpEmail` succeeds and the broken provider is not registered.
- [x] Microsoft Entra helper skips when discovery returns 503.
- [x] `packages/better-auth/src/plugins/generic-oauth/generic-oauth.test.ts` (121 tests).

---
## Summary by cubic
Skip registering a `generic-oauth` provider when OIDC discovery fails or returns invalid metadata, so a bad provider no longer breaks the entire auth API. Previously, discovery errors threw during init and rejected the shared auth context; now we log and skip the failing provider while other providers and email/password continue to work.

- Providers with failed or invalid discovery are omitted from `socialProviders`; clients won't see or use them until discovery is fixed.
- With `requireIdTokenVerification`, a `discoveryUrl` that fails now skips the provider; without one, missing verification metadata still fails init.
- Docs: updated `discoveryUrl` and `requireIdTokenVerification` sections to describe skip behavior.
- Tests: add regression for mixed healthy/broken providers and update `microsoftEntraId` to skip on 503.
- Release: patch changeset for `better-auth`. No migration required; monitor logs for skipped providers.

<sup>Written for commit a994a62f80f9d0f9a5cddede03a64ba898dc4e0b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10978?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bytaesu (approved):
Thanks for working on this!

PR #10981, opened 2026-08-25
refactor: centralize placeholder email generation

bytaesu:
Requiring unique email is a limitation of the current architecture, and we plan to make this more flexible in v2. Until then, if we need to generate placeholder emails, the behavior should at least be consistent.

---
## Summary by cubic
Unifies how we create placeholder emails when providers don’t return one. Previously each integration used ad‑hoc values; now all use the namespaced, non‑routable format `{identifier}@{namespace}.placeholder.invalid`, and placeholders are never marked verified.

**Refactors**
- Adds `createPlaceholderEmail` in `@better-auth/core` and applies it across providers and plugins.
- `anonymous` default email is now `{id}@anonymous.placeholder.invalid`; `emailDomainName` still overrides.
- `siwe` default email is now `{wallet}@siwe.placeholder.invalid`; `emailDomainName` still overrides.
- Microsoft Entra ID now derives placeholders from `oid` and sets `emailVerified` only when a real email exists.
- Providers updated to use stable IDs for placeholders: Reddit, Roblox, TikTok, Twitter, WeChat.
- Docs and logs updated to say “placeholder email” and use `.invalid` examples.

**Migration**
- Update tests or code that asserted old formats (e.g., `.local`, `temp@{id}.com`, username-based values).
- If you relied on SIWE/anonymous default domains, adjust expectations or set `emailDomainName`.
- Expect `emailVerified: false` when placeholders are used (notably for Microsoft Entra ID).
- No data migration; existing stored emails remain unchanged.

<sup>Written for commit 01b6f04d22e6b574e1e1496c37948813f24e89f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/10981?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bytaesu:
let me open a new pr..

Issue #10933, opened 2026-08-22
Define model naming and pluralization contract


bytaesu:
## Problem

The interaction between schema keys, custom `modelName` values, and `usePlural` is not consistently defined.

This causes already-plural names such as `jwks` and `accounts` to become `jwkss` and `accountss`.

Although undesirable, this is observable behavior that may already be reflected in deployed schemas. Changing it can rename existing tables or relation keys and break existing applications.

## Goal

Define the model-naming and pluralization contract for schema keys, physical model names, relation keys, and `usePlural` across core, adapters, plugins, and schema generators.

Audit current behavior and determine backward-compatibility and migration requirements before changing the implementation.

### Pluralization behavior

- #10930
- #10931
- #7773
- #7786
- #6198
- #3069
- #3084
- #6132
- #7814

### Naming contract and configuration

- #10616
- #10631
- #10695
- #7323

PR #10994, opened 2026-08-25
fix(core): do not double-pluralize model names ending in s

### LINKED ISSUE #10930: When `usePlural=true`, `jwks` table become `jwkss`

### Reproduction

1. Set `database.usePlural = true`, add `jwt` plugin
2. Run database migration

~~jwkß imo~~

Rename `jwks` to `jwt`, then `usePlural` can work correctly.

```diff
-export const jwkss = pgTable('jwkss', {
+export const jwks = pgTable('jwks', {
   id: text('id').primaryKey(),
   publicKey: text('public_key').notNull(),
   privateKey: text('private_key').notNull(),
   createdAt: timestamp('created_at').notNull(),
   expiresAt: timestamp('expires_at'),
   alg: text('alg'),
   crv: text('crv'),
 });
```

1.7.1

### System info

```bash
{
  "system": {
    "platform": "win32",
    "arch": "x64",
    "version": "Windows 11 Home China",
    "release": "10.0.26200",
    "cpuCount": 14,
    "cpuModel": "Intel(R) Core(TM) Ultra 5 225H",     
    "totalMemory": "31.50 GB",
    "freeMemory": "6.67 GB"
  },
  "node": {
    "version": "v24.18.0",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.4.0"
  },
  "frameworks": [
    {
      "name": "nuxt",
      "version": "^4.5.2"
    }
  ],
  "databases": [
    {
      "name": "postgres",
      "version": "^3.4.9"
    },
    {
      "name": "drizzle",
      "version": "1.0.0-rc.4"
    }
  ],
  "betterAuth": {
    "version": "Unknown",
    "config": null,
    "error": "Couldn't read your auth config. Make sure to default export your auth instance or to export as a variable named auth."
  }
}
```

Backend

```typescript

```

### Additional context

_No response_

jashkarangiya:
## Summary

With `usePlural` enabled, `getModelName` appended an `s` unconditionally, so the already-plural `jwks` table was generated with an extra trailing `s`. The same happened to any custom `modelName` ending in `s`.

Closes #10930

## The fix

```diff
+/** `jwks` is already plural, so appending another `s` gives `jwkss`. */
+const pluralize = (name: string) => (name.endsWith("s") ? name : `${name}s`);
```

Applied to both branches of `getModelName`, so it covers the built-in `jwks` model and user-supplied `modelName` overrides that are already plural.

Fixed at `getModelName` rather than special-casing `jwks` at the call sites, since every adapter and the schema generators route through it.

## Tests

New `get-model-name.test.ts`: an already-plural model stays unchanged, a singular model still pluralizes, and `usePlural: false` leaves both alone.

The first case fails on the unpatched code with `expected the doubled name to be `jwks``.

`packages/core` and `packages/better-auth/src/db` are 688 passing, and `pnpm typecheck` reports no new errors. The single failure (`preserve a forced UUID on postgres`) needs Docker and fails identically on a clean `main`.

### DIFF
--- a/.changeset/jwks-plural.md
+++ b/.changeset/jwks-plural.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/core": patch
+---
+
+Stop double-pluralizing model names that already end in `s` when `usePlural` is enabled. The `jwks` table was generated with an extra trailing `s`.
--- a/packages/core/src/db/adapter/get-model-name.test.ts
+++ b/packages/core/src/db/adapter/get-model-name.test.ts
@@ -1,0 +1,28 @@
+import { describe, expect, it } from "vitest";
+import { initGetModelName } from "./get-model-name";
+
+const schema = {
+	user: { modelName: "user", fields: {} },
+	jwks: { modelName: "jwks", fields: {} },
+} as any;
+
+describe("getModelName usePlural", () => {
+	/**
+	 * @see https://github.com/better-auth/better-auth/issues/10930
+	 */
+	it("does not double-pluralize a model that already ends in s", () => {
+		const getModelName = initGetModelName({ usePlural: true, schema });
+		expect(getModelName("jwks")).toBe("jwks");
+	});
+
+	it("still pluralizes singular models", () => {
+		const getModelName = initGetModelName({ usePlural: true, schema });
+		expect(getModelName("user")).toBe("users");
+	});
+
+	it("leaves names alone when usePlural is off", () => {
+		const getModelName = initGetModelName({ usePlural: false, schema });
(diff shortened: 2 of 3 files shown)


bytaesu:
Hi @jashkarangiya,

We're tracking this here https://github.com/better-auth/better-auth/issues/10933

Even if it's grammatically correct, changing it now would break our backward compatibility contract.

jashkarangiya:
Thanks, closing in favour of #10933. Agreed the rename is a backward-compatibility break, not a bugfix: anyone already running usePlural has the current names in their deployed schema, so this needs the contract and migration path defined first.

PR #11056, opened 2026-08-29
fix(oauth): revoke tokens when consent is deleted

### LINKED ISSUE #10386: feat(oauth-provider): revoke issued tokens when a user removes a consent (no user-facing revocation today)
### Summary

A user currently has **no way to revoke the tokens of a connected client**. `/oauth2/delete-consent` removes the consent row only, `/oauth2/revoke` (RFC 7009) is client-authenticated, and session deletion deliberately keeps `offline_access` refresh tokens. Since refresh tokens are only issued when `offline_access` is granted, in practice **no user-initiated action revokes a refresh token**. This makes OWASP ASVS 5.0 **L2** (V10 §10.4.9) unattainable.

### Current behaviour (`@better-auth/oauth-provider@1.7.0-rc.1`)

- `/oauth2/delete-consent` deletes the `oauthConsent` row; no token is touched.
- `/oauth2/revoke` requires client credentials, so it is a client-facing endpoint (RFC 7009), not a user-facing one — a user does not hold the token value.
- The plugin registers `databaseHooks.session.delete.before` → `revokeAndPlanBackchannelLogout`, which revokes session-bound access tokens and refresh tokens **except** those carrying `offline_access`. The doc comment states this explicitly:

  > `offline_access` refresh tokens survive so long-lived API access can outlive the browser session.

- Token issuance only creates a refresh token when `offline_access` is among the granted scopes.

So for the common case (an API/MCP client holding an `offline_access` refresh token), logging out, revoking the session, and deleting the consent all leave the refresh token usable until the sliding inactivity window lapses.

### Why this matters

OWASP ASVS 5.0, V10 (OAuth and OIDC):

> **10.4.9** (L2) — Verify that refresh tokens and reference access tokens can be revoked by an authorized user using the authorization server user interface, to mitigate the risk of malicious clients or stolen tokens.

A user who removes an application from a "connected apps" screen reasonably expects that application to lose access. Today it keeps refreshing.

### Proposal

When a user deletes a consent via `/oauth2/delete-consent` (and, when scopes are narrowed, on `/oauth2/update-consent`), revoke the `oauthAccessToken` / `oauthRefreshToken` rows for that `(userId, clientId)` pair — either as the default behaviour, or behind an option (e.g. `revokeTokensOnConsentDelete?: boolean`, defaulting to `true`).

This keeps the deliberate "offline_access survives logout" semantics intact (logout ≠ consent removal) while giving the user an actual revocation path.

Happy to open the PR if the direction is agreed.

Garv978:
## What does this PR do?

Revokes OAuth access and refresh tokens when a user deletes their OAuth consent.

This ensures that removing a connected application also invalidates tokens issued to that client.

## Changes

- Revoke `oauthAccessToken` records associated with the deleted consent.
- Revoke `oauthRefreshToken` records associated with the deleted consent.
- Keep the existing logout/session deletion behavior unchanged.
- Add test coverage verifying that associated tokens are revoked.

## Testing

- `pnpm vitest --run src/oauthConsent/endpoints.test.ts`
- Full test suite: 923 tests passed
- `git diff --check`

Closes #10386

---
## Summary by cubic
Revokes OAuth access and refresh tokens when a user deletes their OAuth consent, scoped to the consent's `referenceId` so only that removed app's credentials are invalidated. Previously, deleting a consent left its tokens active; now standalone access tokens without a linked refresh token are revoked too, while tokens for other consented apps stay valid. Logout and session deletion behavior is unchanged.

**Also in this PR**
- Prevents slow parsing of malformed custom-scheme callback URLs in trusted-origin checks.
- Aligns generated OpenAPI required fields with Zod 4.5, including passkey registration responses.
- Optimizes request IP validation and organization role permission checks.
- Upgrades `zod` to 4.5.4 through the workspace catalog.
- Consolidates dashboard plugin docs into `dashboard.mdx` with redirects, and adds the social provider contribution policy and community plugin listings.
- Updates the SAML test mock IdP to use redirect binding.

<sup>Written for commit 74e76bc96f7a405b375ddf4d371365aaca5ec1bb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11056?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Garv978:
@gustavovalverde I’m a bit confused about the latest review comment and would really appreciate some guidance on how I should proceed. 🙏

Issue #9814, opened 2026-05-30
expo: large chunked account cookie dropped after sign-in (cookie jar stored under a single storage key)


gustavovalverde:
### Impact

In an Expo app using `storeAccountCookie` without a database, the account cookie is lost after signing in with a provider that issues large tokens (Keycloak and other enterprise IdPs). Server-side `auth.api.accountInfo` and `auth.api.getAccessToken` then return `ACCOUNT_NOT_FOUND`, and any `customSession` enrichment that depends on them fails, leaving `useSession()` null. This was reported as a core account-resolution bug in #9151, but it is a `@better-auth/expo` client-storage defect: the better-auth server reads chunked account cookies correctly when they are actually replayed.

### Root cause

The Expo client serializes the entire cookie jar (session cookies plus every chunked `account_data.N`) into a single JSON value and writes it under one storage key.

- `getSetCookie` collects all Set-Cookie entries into one object and `JSON.stringify`s it: [packages/expo/src/client.ts#L74-L99](https://github.com/better-auth/better-auth/blob/d3919dc1a560625d8f09161d64701e257452940f/packages/expo/src/client.ts#L74-L99)
- that single string is stored under one `cookieName` key: [packages/expo/src/client.ts#L347-L359](https://github.com/better-auth/better-auth/blob/d3919dc1a560625d8f09161d64701e257452940f/packages/expo/src/client.ts#L347-L359)

Large IdP tokens push the account cookie past the 4096-byte cookie limit, so the server chunks it into `better-auth.account_data.0`, `.1`, and so on. The Expo client merges all of those, plus the session cookies, into one value. The documented storage backend, `expo-secure-store`, warns that values larger than 2048 bytes may not be stored reliably, so the oversized write is dropped or rejected silently. On the next request the rebuilt `Cookie` header is missing the account cookie, it never reaches the server, and with no database the `findAccounts` fallback is empty, producing `ACCOUNT_NOT_FOUND`.

This matches the 1.4.x to 1.6.x regression in #9151: account cookies began carrying full provider tokens (chunked) in 1.6.x, which is exactly what overflows a single-key store. A small session-only jar fits; a jar that also holds chunked Keycloak tokens does not.

### Reproduction

[...]

PR #11057, opened 2026-08-29
fix(expo): serialize async secure-storage access so concurrent requests never read a torn cookie jar

denis110000:
## Summary

On native, `@better-auth/expo` stores the whole cookie jar under one secure-storage key and, since #10438 / v1.7, reads and writes it with the async SecureStore API, chunking values over 1800 chars (#9814). A chunked write is several native round-trips in a fixed order: clear the base key → write `<key>.0..N` → write the `ba-chunks:N` marker.

Nothing serializes those round-trips against other adapter calls. On app start several requests fire together (`useSession`, any `getSession()` the app makes, …) and, with `session.cookieCache` enabled, nearly every response re-sets `session_data` and rewrites the jar in `onSuccess`. A request whose `init` runs while another response is mid-write reads the cleared base key, gets `""`, and:

- `safeJSONParse("")` throws → `ERROR [Better Auth]: Error parsing JSON {"error":{}}` in the console (the 1.6 client swallowed this silently, 1.7 logs it), and
- `getCookie()` returns `""`, so that request goes out **without a `Cookie` header** and is handled as anonymous.

Two concurrent chunked writes can also interleave their chunk writes.

Any app whose jar exceeds 1800 chars hits this — with `cookieCache` on, a bare user with no extra fields is already at ~1750, so real accounts always take the chunked path.

## Fix

`storageAdapter` now runs `getItemAsync` / `setItemAsync` through a per-key FIFO queue, so every read of a key happens fully before or fully after a write to it and writes never interleave; the cookie jar and the session cache are independent keys and never wait on each other. The write order (clear → chunks → marker) is unchanged, so the existing fail-closed behaviour on an interrupted write is preserved. A rejected operation does not stall the queue; the rejection still reaches its caller. The sync `getItem` / `setItem` are untouched (the plugin itself no longer uses them).

## Tests

- read issued while a chunked write is in flight returns the new value, never `""` or a torn value (fails on `main`: returns `""`)
- two concurrent chunked writes leave the last value, not a mix
- a chunked write to one key does not delay a read of another
- a rejected operation doesn't block subsequent ones

## Changeset

`@better-auth/expo`: patch

### DIFF
--- a/.changeset/expo-storage-queue.md
+++ b/.changeset/expo-storage-queue.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/expo": patch
+---
+
+Serialize async secure-storage access in the Expo client so a request started while another response is rewriting the chunked cookie jar no longer reads an empty or torn value, logs `Error parsing JSON`, and goes out without a `Cookie` header.
--- a/packages/expo/src/client.ts
+++ b/packages/expo/src/client.ts
@@ -298,7 +298,30 @@ function getStorageWrites(key: string, value: string): [string, string][] {
 	return writes;
 }
 
+/**
+ * Runs async storage operations one at a time, in call order. A chunked write
+ * is several native round-trips (clear the base key, write each chunk, set the
+ * marker), and any read that lands between them sees an empty or half-written
+ * jar. That is exactly what happens on app start, where several requests fire
+ * together and every response rewrites the cookie jar: one request's `init`
+ * reads while another's `onSuccess` is mid-write, gets `""`, logs
+ * "Error parsing JSON" and goes out without a Cookie header. Queueing the
+ * operations keeps every read either fully before or fully after a write, and
+ * keeps two concurrent writes from interleaving their chunks.
+ */
+function createQueue() {
+	let tail: Promise<unknown> = Promise.resolve();
+	return <T>(task: () => Promise<T>): Promise<T> => {
+		const run = tail.then(task, task);
+		// Swallow here only so a rejected task doesn't stall the queue; callers
+		// still get the original rejection through `run`.
+		tail = run.catch(() => {});
+		return run;
+	};
+}
+
 export function storageAdapter(storage: ExpoClientStorage) {
+	const enqueue = createQueue();
 	return {
 		/**
 		 * Reads a value, reassembling it if it was split across chunk keys. A value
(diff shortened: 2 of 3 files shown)


denis110000:
Fair point — the jar and the session cache are separate keys and do not need to wait on each other. Switched to a per-key queue in the follow-up commit (`createKeyedQueue`, entries dropped once their tail settles) and added a test that a chunked write to `_session_data` does not delay a read of `_cookie`.

denis110000:
Not introduced by this PR, and queueing cannot fix it either way. The chunk keys `<key>.0..N` share the physical namespace with plain keys since chunking landed (#9814): a caller writing `foo.0` while `foo` is chunked clobbers a chunk of `foo` on `main` too, serialized or not — the collision is in the storage layout, not in the ordering. Serializing `foo.0` behind `foo` would only change *which* write wins, not stop the corruption.

The plugin itself only ever uses the two fixed keys `<prefix>_cookie` and `<prefix>_session_data`, which cannot alias each other's chunks. Reserving a collision-free chunk namespace means changing the on-device layout (and reading the old one back) — a separate change from this race fix; happy to open it as a follow-up if maintainers want it.

fierysolid:
Filed [#11082](https://github.com/horsehq-org/horsehq-web/issues/11082) for the same race before finding this — happy to close mine as a duplicate.

Production confirmation in case it helps prioritize: on @better-auth/expo@1.6.30, jar ~1900 chars (2 chunks), we logged 39 empty-jar reads across 15 users in 24h, iOS and Android, always in a cold start's parallel request burst. Reads land between the [key, ""] write and the marker write, get "", send no cookie, take a 401.

One follow-up thought: the queue removes interleaving, but a single write still isn't atomic — getStorageWrites clears the base key first and commits with the marker last, and setItem catches failures rather than throwing, so an interrupted write (process killed, app suspended, a native write rejecting) leaves the base key "" and every later read gets an empty jar. That's a code-reading argument, not something I've measured.

We're patching that with a slot flip on the same chunking: base key holds <marker><count>.<slot>, the writer fills the inactive slot and flips the base key in one write, base key never cleared. Readers always see a complete jar, and it survives a half-finished write. ~55 lines, old marker format still read so jars migrate. Complementary to this PR rather than competing — happy to send it as a follow-up on top of this branch if you want it.

denis110000:
Thanks — this is exactly the failure pattern the PR describes, and production numbers at that scale are the strongest signal the maintainers could ask for. Linking #11082 here so both threads carry it.

Your reading of the interrupted-write case is correct: the clear-first / marker-last order plus the swallowed `setItemAsync` errors means a write cut short by process death (or a native rejection) leaves the base key `""` for good, and since requests then go out anonymous, nothing ever rewrites the jar — the session is lost until re-login. This PR deliberately keeps that behaviour: it's a pure concurrency fix with the on-disk format untouched, so it stays small and easy to verify.

A slot flip is the right shape for the durability half — single-write commit, base key never cleared, and it composes cleanly with the queue here. I'd suggest sending it as a separate PR once this one lands rather than stacking on this branch (a stack on an unmerged fork makes both harder to review), with a regression test that kills a write between the chunk writes and the flip. Happy to review it.

bytaesu (commented):
Hi @denis110000, thanks for working on this.

This will be superseded by https://github.com/better-auth/better-auth/pull/11099

PR #11079, opened 2026-08-31
docs: give the OAuth issuer value in the Auth.js migration guide

michelecanini:
The guide gives the exact issuer value for credential accounts (`local:credential`) but not for OAuth accounts, so a migration that moves existing OAuth rows has no value to write for the new `issuer` column.

The obvious guess is `createOAuthAccountIssuer()` for every provider. That is right for the providers that declare no issuer of their own, which is most of them, and wrong for the OpenID Connect ones that do: among the built-ins, Google, Apple, Facebook and LINE, plus Cognito and Entra ID whose value depends on configuration or on the token.

The failure is silent. Sign-in resolves an account by the `(issuer, accountId)` pair with no fallback on `providerId`, so a row with the wrong issuer is never found and nothing is logged. Linking by email would normally cover it, but `accountLinking.requireLocalEmailVerified` defaults to `true` and Auth.js leaves `emailVerified` null for most accounts created through OAuth, so those users get `account not linked` and cannot sign in at all.

This adds a callout beside the existing credentials one. The values were read from `accountIssuer` on the built-in providers in 1.7.2.

The same value is documented in the 1.7 upgrade guide and in the Clerk, Auth0 and Supabase migration guides. This is the guide that does not have it.

Docs only, no behaviour change.

---
## Summary by cubic
Adds the missing OAuth `issuer` value to the Auth.js migration guide so migrations can fill the new `issuer` column correctly.

- The correct value differs by provider: `createOAuthAccountIssuer()` for providers without their own issuer, provider-declared values for OpenID Connect ones (Google, Apple, Facebook, LINE, and Cognito/Entra ID depending on configuration or token claims), and Slack depends on integration (built-in uses local form, `genericOAuth` uses `https://slack.com`).
- Writing the wrong value fails silently because sign-in matches on `(issuer, accountId)` with no fallback on `providerId`, and email linking is usually refused since `requireLocalEmailVerified` defaults to `true`.
- Docs only, no behavior change.

<sup>Written for commit f2efbd4678ae89ca32fa7a49ffb9b812b4ddb1cd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11079?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


michelecanini:
The review is right, and the failure it describes is the one this callout exists to prevent, so thanks for catching it.

Slack was the wrong example because it is the one provider where the answer depends on how you configured it: `packages/core/src/social-providers/slack.ts` declares no `accountIssuer`, so the built-in social provider does take the local form, while `packages/better-auth/src/plugins/generic-oauth/providers/slack.ts` sets `accountIssuer: "https://slack.com"`.

Rather than drop it, I moved Slack into the sentence that already covers the providers whose value depends on your own setup, next to Cognito and Microsoft Entra ID, and spelled out both cases.

I also checked the other names in the callout so the same ambiguity is not left elsewhere. Among the generic-oauth providers only `slack` and `line` declare an `accountIssuer`; LINE was already listed with `https://access.line.me`, which matches. GitHub and Discord have no generic-oauth helper at all, so they stay unambiguous examples.

bytaesu:
Closing this PR in favor of #11153

PR #11087, opened 2026-08-31
fix(oauth-provider): detect adapter-wrapped unique violations when seeding resources

### LINKED ISSUE #11034: oauth-provider: seedResources concurrent-insert guard misses Drizzle-wrapped unique violations, aborting plugin init
### Summary

`seedResources` has a guard for the concurrent-insert race, but it is defeated when the Drizzle adapter is used with PostgreSQL: it tests `err.message`, while the Postgres SQLSTATE lives on `err.cause`. The guard never fires, the error is rethrown, and the plugin's `init` aborts **before returning its `databaseHooks`** — so the losing instance silently ends up without the back-channel-logout session hooks registered.

Verified present in **1.7.1 and 1.7.2** (`packages/oauth-provider`, `dist/introspect-*.mjs`).

### The guard

```js
try {
  await ctx.adapter.create({ model: modelName, data: buildSeedRow(input, now) });
} catch (err) {
  const message = err instanceof Error ? err.message : String(err);
  if (/unique|duplicate|UNIQUE/i.test(message)) {
    logger.debug(`oauth-provider: resource ${input.identifier} already inserted by a concurrent process — skipping.`);
    continue;
  }
  ...
  throw err;
}
```

The intent is right. The problem is only *where* it looks.

### Why it misses

`drizzle-orm` wraps driver errors. The thrown error's `message` is the failed SQL:

```
Failed query: insert into "oauthResource" ("id", "identifier", "name", "accessTokenTtl", ...)
values ($1, $2, ...) returning "id", ...
params: yeumgpUL3jtoCH8o23IwPw3RjuFl4OL5,http://localhost:3000/api/mcp,...
```

That string contains neither `unique` nor `duplicate`. The actual Postgres error is on `err.cause`:

```
error: duplicate key value violates unique constraint "oauthResource_identifier_unique"
  code: '23505'
  detail: 'Key (identifier)=(http://localhost:3000/api/mcp) already exists.'
  constraint: 'oauthResource_identifier_unique'
```

So `/unique|duplicate/i.test(err.message)` is `false`, and `throw err` runs.

### Consequence

The failing call is `await seedResources(ctx, opts)` at the top of the plugin's `init`. Everything after it is skipped, including the return value:

```js
init: async (ctx) => {
  if (ctx.options.secondaryStorage && ...) throw new BetterAuthError(...);
  await seedResources(ctx, opts);                  // throws
  logEnforcePerClientResourcesResolution(opts);    // skipped
  if (!opts.disableJwtPlugin) { /* issuer validation */ }   // skipped
  return { options: { databaseHooks: { session: { delete: { ... } } } } };  // never returned
}
```

That return value is what registers the back-channel logout hooks on session deletion. So this is not merely a noisy log: the instance that loses the race completes startup without those hooks. It also surfaces as an `unhandledRejection`, which under Node's default `--unhandled-rejections=throw` can terminate the process rather than just log.

Practically this hits **the first boot after the `oauthResource` row is absent** — a fresh database, a reset dev schema, or the first deploy of a multi-replica service, where N replicas call `init` concurrently and N-1 can fail.

### Reproduction

1. Postgres + `drizzleAdapter`, `oauthProvider`/`mcp` configured with a `resource`.
2. Ensure `oauthResource` is empty (fresh DB or `delete from "oauthResource"`).
3. Start two instances concurrently (e.g. Next.js dev, which initializes more than once).

One insert wins; the other logs the above and never returns from `init`. Observed with `better-auth` 1.7.1, `@better-auth/mcp` 1.7.1, `@better-auth/oauth-provider` 1.7.1, `drizzle-orm` + `pg`.

### Suggested fix

Match the SQLSTATE rather than the message text, walking the `cause` chain so wrapped adapter errors are recognised:

```js
function isUniqueViolation(err) {
  for (let e = err; e != null; e = e.cause) {
    if (e.code === "23505") return true;                       // Postgres
    if (/unique|duplicate|UNIQUE/i.test(String(e.message ?? ""))) return true;  // other adapters
  }
  return false;
}
```

Keeping the message regex as a fallback preserves current behaviour for adapters that do not expose a code (SQLite/MySQL wording differs). The same `cause`-walking issue applies to `MISSING_TABLE_PATTERN` just below it, which will also miss a wrapped error.

An adapter-level upsert for the seed would remove the need to classify the error at all, if that is available across adapters.

I am happy to open a PR for the `isUniqueViolation` version if you would like.

zoom-dustin-moriarty:
Closes #11034

## Root cause

`seedResources` detects a lost insert-race by testing `err.message` against a unique-constraint regex. The Drizzle adapter wraps driver errors: the thrown error's `message` is the failed SQL text, and the Postgres SQLSTATE lives on `err.cause`. The regex never matches, so `throw err` runs and the plugin's `init` aborts before its final `return` — which is what registers the back-channel-logout `databaseHooks` on session deletion.

## Fix

Added one internal module, `isUniqueConstraintError`/`isMissingTableError` in `packages/oauth-provider/src/utils/db-errors.ts`, that walks the error's `cause` chain checking driver identifiers (`code`/`errno`/`errcode`/`number` against `23505`, `1062`, `11000`, `2067`, `2601`, `2627`, `ER_DUP_ENTRY`, `P2002`, `SQLITE_CONSTRAINT_UNIQUE`) before falling back to the existing message regex, so adapters exposing no code keep working unchanged. Every inline `err.message` check in the package (`resources.ts`, `oauthResource/endpoints.ts`, `register.ts`) now goes through these two functions, and `register.ts`'s duplicated private helper is removed.

`packages/better-auth/src/plugins/device-authorization/routes.ts` has the same latent gap but lives in a different package — left out of scope here, happy to follow up separately.

## Behavior notes

- No public API change; the module is internal to `oauth-provider`.
- The seed loop keeps `continue` on a detected unique violation in every `resourceSeedMode` (`insertOnly`, `merge`, `overwrite`). This is safe by construction, not just for the default mode: a race can only happen when `findOne` found no row, meaning the winner just inserted a row built by `buildSeedRow` from the same config. `buildSeedUpdate` in `"overwrite"` mode is `buildSeedRow` minus `identifier`/`createdAt`, and `"merge"` writes a subset of those same values — so the update the loser skips would only rewrite what the winner already wrote.

## Tests

- `packages/oauth-provider/src/utils/db-errors.test.ts` (new): both predicates against bare driver errors, Drizzle-wrapped errors, message-only adapters, cyclic `cause` chains, a depth cap, and non-object throwables.
- `packages/oauth-provider/src/resources.test.ts`: a regression test with a Drizzle-shaped wrapped error that fails on `main` today (rejects instead of resolving), plus a negative case confirming an unrelated wrapped failure still rethrows, and a wrapped missing-table case.
- Also in `resources.test.ts`: `oauthProvider`'s `init` is called directly against a stub `AuthContext` (table exists, `create` throws the wrapped error) to prove the actual consequence — `init` still returns `options.databaseHooks.session.delete` after losing the race. This can't be shown through the existing `getTestInstance` harness, since it runs `runMigrations()` after plugin init and seeding always defers there before reaching `create`.

## End-to-end reproduction (Postgres + Drizzle)

The repo has no Drizzle test infrastructure, so here's a standalone reproduction instead of an in-tree e2e test:

```ts
import { drizzle } from "drizzle-orm/node-postgres";
import { Pool } from "pg";

const pool = new Pool({ connectionString: process.env.DATABASE_URL });
const db = drizzle(pool);

await db.execute(`delete from "oauthResource"`);

function makeAuth() {
  return betterAuth({
    plugins: [oauthProvider({ resources: ["http://localhost:3000/api/mcp"] })],
  });
}

const [a, b] = [makeAuth(), makeAuth()];
const results = await Promise.allSettled([a.$context, b.$context]);

console.log(results.map((r) => r.status));
// On main:  [ 'fulfilled', 'rejected' ]  — the loser's init throws
// Fixed:    [ 'fulfilled', 'fulfilled' ] — both settle, both get their hooks
```

## Interim workaround

For anyone hitting this before a release with the fix: pre-insert the `oauthResource` row (`insert ... on conflict do nothing`) in a migration before replicas start, so `findOne` finds it and `create` is never attempted. There's no configuration escape hatch — the adapter API has no `upsert`, no option disables seeding, and with the `mcp` plugin `resources` is never empty (it unconditionally injects the MCP resource), so the seed always runs.

---
## Summary by cubic
Fixes `oauth-provider`'s seed logic so a wrapped unique-constraint error no longer aborts plugin init. The Drizzle adapter stores the SQLSTATE on `error.cause` instead of the message, so the old regex never matched and `init` threw before registering the back-channel-logout session hooks.

- Adds internal `isUniqueConstraintError`/`isMissingTableError` helpers that walk the `cause` chain and check driver codes (`23505`, `ER_DUP_ENTRY`, etc.) before falling back to message text.
- The message fallback matches "duplicate", "unique constraint", and "unique key" phrases rather than a bare "unique" word, so unrelated errors like "could not uniquely identify row" still rethrow.
- Replaces every inline error-message check in the package with these helpers and removes a duplicated private predicate.
- Adds regression tests proving `init` still returns the session delete hook after losing an insert race, and that unrelated wrapped failures still rethrow.

<sup>Written for commit 962a7b0527e57f88a14282d6cee3ec83c582385d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11087?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/oauth-provider-wrapped-unique-violation.md
+++ b/.changeset/oauth-provider-wrapped-unique-violation.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/oauth-provider": patch
+---
+
+Recognise unique-constraint and missing-table errors that adapters wrap in `error.cause`. The Drizzle adapter reports the failed SQL as the error message and puts the Postgres SQLSTATE on `cause`, so `seedResources` did not detect a lost insert race and rethrew — aborting the plugin's `init` before it returned the back-channel-logout session hooks.
--- a/packages/oauth-provider/src/oauthResource/endpoints.ts
+++ b/packages/oauth-provider/src/oauthResource/endpoints.ts
@@ -8,6 +8,7 @@ import type {
 	OAuthResourceInput,
 	Scope,
 } from "../types";
+import { isUniqueConstraintError } from "../utils/db-errors";
 
 /**
  * Action types passed to {@link OAuthOptions.resourcePrivileges}. Mirrors
@@ -145,8 +146,7 @@ export async function createResourceEndpoint(
 			data: buildResourceRow(input, now),
 		});
 	} catch (err) {
-		const message = err instanceof Error ? err.message : String(err);
-		if (/unique|duplicate|UNIQUE/i.test(message)) {
+		if (isUniqueConstraintError(err)) {
 			throw new APIError("BAD_REQUEST", {
 				error: "invalid_request",
 				error_description: `resource ${input.identifier} already exists`,
@@ -347,8 +347,7 @@ export async function linkClientResourceEndpoint(
 			} as never,
 		});
 	} catch (err) {
-		const message = err instanceof Error ? err.message : String(err);
-		if (/unique|duplicate|UNIQUE/i.test(message)) {
+		if (isUniqueConstraintError(err)) {
 			return ctx.json({ linked: true, alreadyLinked: true });
 		}
 		throw err;
--- a/packages/oauth-provider/src/register.ts
+++ b/packages/oauth-provider/src/register.ts
@@ -33,6 +33,7 @@ import type {
 import type { GrantType, OAuthClient } from "./types/oauth";
 import { parseClientMetadata, storeClientSecret } from "./utils";
 import { isPrivateHostname } from "./utils/client-assertion";
+import { isUniqueConstraintError } from "./utils/db-errors";
 import { authorizeInitialAccessToken } from "./utils/initial-access-token";
 
 const DEFAULT_REGISTRATION_GRANT_TYPES = [
@@ -1127,36 +1128,6 @@ export async function registerClientMetadataDocument(
 			throw retryError;
 		}
 	}
-}
-
-function isUniqueConstraintError(error: unknown): boolean {
-	if (!error || typeof error !== "object") return false;
-	const candidate = error as {
-		code?: unknown;
-		errno?: unknown;
-		name?: unknown;
-		message?: unknown;
-	};
-	const code = String(candidate.code ?? candidate.errno ?? "");
-	if (
-		[
-			"1062",
-			"11000",
-			"23505",
-			"ER_DUP_ENTRY",
-			"P2002",
-			"SQLITE_CONSTRAINT_UNIQUE",
-		].includes(code)
-	) {
-		return true;
-	}
-	if (candidate.name === "MongoServerError" && candidate.code === 11000) {
-		return true;
-	}
-	return (
-		typeof candidate.message === "string" &&
-		/(?:duplicate|unique constraint|unique key)/i.test(candidate.message)
-	);
 }
 
 export function createOAuthClientEndpoint(
--- a/packages/oauth-provider/src/resources.test.ts
+++ b/packages/oauth-provider/src/resources.test.ts
@@ -260,6 +260,97 @@ describe("seedResources (integration via plugin init)", () => {
 			allowedScopes: null, // cleared
 		});
 	});
+
+	/**
+	 * `drizzle-orm` throws an `Error` whose `message` is the failed SQL and
+	 * whose `cause` carries the Postgres SQLSTATE. Reported as
+	 * better-auth#11034.
+	 */
+	const drizzleUniqueViolation = () =>
+		Object.assign(
+			new Error(
+				'Failed query: insert into "oauthResource" ("id", "identifier") values ($1, $2) returning "id"',
+			),
+			{
+				cause: Object.assign(
+					new Error(
+						'duplicate key value violates unique constraint "oauthResource_identifier_unique"',
+					),
+					{ code: "23505", constraint: "oauthResource_identifier_unique" },
+				),
+			},
+		);
+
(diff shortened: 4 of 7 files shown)


zoom-dustin-moriarty:
Good catch — fixed in 962a7b052. `register.ts`'s original fallback was the narrower `/(?:duplicate|unique constraint|unique key)/i`; the shared module's bare `/unique|duplicate/i` widened matching for its call sites and could misclassify an unrelated failure (e.g. a message that just happens to contain "unique") as a lost insert race. Restored the narrower phrase match, which still covers every documented adapter wording in the fallback path (SQLite's "UNIQUE constraint failed", MySQL/MongoDB's "Duplicate entry"/"duplicate key"), and added a regression test for the false-positive case.

PR #11099, opened 2026-09-01
fix(expo): prevent torn cookie storage writes

### LINKED ISSUE #11082: [BUG] [Expo] Chunked cookie-jar write is non-atomic — concurrent reads send no cookie (401s)
## Summary

`storageAdapter.setItem` in `@better-auth/expo` writes an oversized cookie jar by
**clearing the base key first**, then writing each chunk, then writing the marker.
Every one of those is an awaited native write, so the clear and the commit are
separated by several event-loop turns. A concurrent read during that window gets
`""`, and `getCookie` turns that into an **empty cookie header rather than an
error** — so the request is sent unauthenticated and the server answers 401.

Present in 1.6.30 and unchanged in 1.7.2 (the writes were refactored into
`getStorageWrites`, which still emits `[key, ""]` first, and the cookie hooks
still use the awaited `setItemAsync` / `getItemAsync` pair).

## The code

```js
function getStorageWrites(key, value) {
  if (value.length <= STORAGE_VALUE_LIMIT) return [[key, value]];
  const count = Math.ceil(value.length / STORAGE_VALUE_LIMIT);
  const writes = [[key, ""]];                       // <-- window opens here
  for (let i = 0; i < count; i++) writes.push([`${key}.${i}`, value.slice(...)]);
  writes.push([key, `${CHUNK_MARKER}${count}`]);    // <-- commit
  return writes;
}
```

Read side, during that window: `storage.getItem(key)` returns `""`, which does not
start with `CHUNK_MARKER`, so it is returned as-is. Then:

```js
function getCookie(cookie) {
  const parsed = safeJSONParse(cookie) ?? {};   // "" -> null -> {}
  return Object.entries(parsed).reduce(..., "");// -> ""  (no cookie sent)
}
```

The failure is silent by construction: the jar "fails closed" for a torn write, but
the caller cannot distinguish that from "no session", so it sends the request
anyway, without credentials.

## Why it fires in practice

Cold start issues a burst of parallel `/api/auth/*` requests (`get-session`,
`organization/list`, `get-full-organization`, `get-active-member`, ...). Each
response carries `Set-Cookie`, so each one triggers `onSuccess` -> a full jar
rewrite, while other in-flight requests are calling `init` -> a jar read. There is
no mutex or write queue anywhere in the package.

## Evidence from production

Instrumented client, ~24h, one app:

- Signed-in jar: **1922 characters, 2 chunks** (limit is 1800, so the chunked path
  is taken). Cookie cache enabled (`strategy: 'jwe'`) is what pushes it over.
- **39 empty-cookie events across 15 users**, on **both iOS and Android**, each one
  on the first request of a cold start — i.e. coincident with the jar rewrite.
- Server side, the same query 401s and then succeeds a second later with no change
  in session state.

Not iOS/Keychain specific.

## Suggested fix

Make the commit atomic instead of clearing first. Write chunks to a slot the
current marker does not point at, then flip the base key in a single write:

- marker encodes count **and** slot, e.g. `<count>.<slot>`
- chunks live at `<key>.<slot>.<i>`
- writer picks the inactive slot, writes all chunks, then writes the marker once
- reader resolves the slot from the marker; keep reading the old marker format so
  existing jars survive the upgrade

The base key is then never cleared, so a concurrent reader always sees a complete
jar — the previous value or the new one, never a torn or empty read. Stale chunks
in the other slot are overwritten on the next flip.

I have this running as a patch against 1.6.30 and can open a PR if the approach
looks right. Verified by interleaving a read after every write step: all reads
return a complete jar, none empty; shrinking back to an unchunked value
round-trips; legacy-format jars still read.

## Environment

- `@better-auth/expo` 1.6.30 (same code path in 1.7.2)
- `better-auth` 1.6.30
- Expo SDK 55, React Native 0.83.6, Hermes
- `expo-secure-store`, iOS + Android

bytaesu:
- Closes #11082
- Supersedes #11057

Following the concurrency analysis in [#11057](https://github.com/better-auth/better-auth/pull/11057), this PR serializes asynchronous writes per storage key. Reads remain lock-free because clear-first chunk writes are replaced with recoverable copy-on-write storage.

---
## Summary by cubic
Fixes a race in the Expo storage adapter that sent unauthenticated requests while chunked cookie updates were in progress (closes #11082). Updates now use copy-on-write and per-key write serialization so reads always see a complete value, and interrupted updates recover the previous cookie value.

- Writes chunked cookies to an inactive slot first, then switches the base marker, instead of clearing the old value up front.
- Async writes are serialized per key across adapters sharing the same storage object; independent storage objects keep separate queues.
- Cookie updates run as atomic read-modify-write operations inside the queue.
- If the active slot is incomplete, reads fall back to the previous slot's chunks.
- Legacy chunk markers (no slot info) still read back correctly.
- Values requiring more than 100 chunks are rejected.

<sup>Written for commit 85e7886e6d8b8b9e3c0d5f5a73ee18fc061c7b1f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11099?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/calm-taxis-store-cookies.md
+++ b/.changeset/calm-taxis-store-cookies.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/expo": patch
+---
+
+Prevent Expo apps from sending unauthenticated requests while large stored cookies are being updated, and recover the previous cookie value when an update is interrupted or incomplete.
--- a/docs/content/docs/integrations/expo.mdx
+++ b/docs/content/docs/integrations/expo.mdx
@@ -485,7 +485,7 @@ export function TRPCProvider(props: { children: React.ReactNode }) {
 
 ### Expo Client
 
-**storage**: the SecureStore-compatible storage used to cache session data and cookies.
+**storage**: the SecureStore-compatible storage used to cache session data and cookies. Write coordination is scoped to the provided object, so reuse it when configuring multiple clients that access the same stored data.
 
 ```ts title="lib/auth-client.ts"
 import { createAuthClient } from "better-auth/react";
--- a/packages/expo/src/client.ts
+++ b/packages/expo/src/client.ts
@@ -272,6 +272,7 @@ export function normalizeCookieName(name: string) {
  * @see https://github.com/better-auth/better-auth/issues/9151
  */
 const STORAGE_VALUE_LIMIT = 1800;
+const MAX_STORAGE_CHUNKS = 100;
 
 /**
  * Marks a base key whose value is split across multiple storage keys. Legacy
@@ -295,7 +296,10 @@ function parseChunkCount(value: string | undefined): number | null {
 		return null;
 	}
 	const count = Number(value);
-	return Number.isInteger(count) && count > 0 ? count : null;
+	if (!Number.isInteger(count) || count < 1 || count > MAX_STORAGE_CHUNKS) {
+		return null;
+	}
+	return count;
 }
 
 function parseChunkMarker(baseValue: string): ChunkMarker | null {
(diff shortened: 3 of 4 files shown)


bytaesu:
Tested a minimal reproduction locally in Expo Go and checked the login flow works correctly. Merging.

PR #11108, opened 2026-09-02
fix(email-otp): surface sendVerificationOTP errors when not deferred

### LINKED ISSUE #11107: sendVerificationOTP errors still silently swallowed on 1.7.2 (real prod repro, #9183 wasn't actually fixed)
## Summary

`sendVerificationOTP`/`sendVerificationEmail` errors are still silently swallowed on 1.7.2, confirmed by reading the installed source, not just observing symptoms. This is the same root cause already described in #9183 ("tracking: email/OTP pipeline (error surfacing, Prisma v7 regressions)"), which was closed by the stale bot after 7 days of no follow-up — not because it was fixed. Its own "Resolution criteria" ("Errors thrown in sendVerificationOTP/sendVerificationEmail surface to the caller with the original status and payload") is still unmet. That issue is now locked, so filing fresh with independent reproduction evidence rather than commenting there.

## Root cause (confirmed against installed source)

`context/create-context.mjs`'s `runInBackgroundOrAwait`:

```js
async runInBackgroundOrAwait(promise) {
  try {
    if (options.advanced?.backgroundTasks?.handler) {
      if (promise instanceof Promise) options.advanced.backgroundTasks.handler(promise.catch((e) => {
        logger.error("Failed to run background task:", e);
      }));
    } else await promise;
  } catch (e) {
    logger.error("Failed to run background task:", e);
  }
},
```

Without a custom `backgroundTasks.handler` configured (the common case), the `else await promise` branch does run — this isn't a fire-and-forget issue in the way one might first assume. But the whole thing is wrapped in one outer `try/catch` that only calls `logger.error(...)` on failure and never rethrows. `plugins/email-otp/routes.mjs`'s `send-verification-otp` route calls this and then unconditionally returns `ctx.json({ success: true })` right after, regardless of whether the awaited promise resolved or the catch branch ran.

## Real-world impact (reproduced against a real production deploy, not a synthetic test)

Our own `sendVerificationOTP` implementation threw because two required env vars were briefly missing from the running container (an ops-side deploy gap on our end, since fixed — not a bug in our own send logic). The endpoint returned `200 {"success":true}` regardless — no email was ever sent, and the only trace of the failure was a `logger.error` line in server-side console output: never visible to the HTTP caller, and not correlatable to the specific request that triggered it. A user retrying "resend code" would get the same false-success response indefinitely, with zero signal anything was wrong on either the client or server-observable side.

## Suggested resolution

Same as #9183's own stated criteria: let an error thrown inside `sendVerificationOTP`/`sendVerificationEmail` propagate to the route handler with its original status/payload (or at minimum, let the response reflect send failure) instead of being unconditionally swallowed into a same-process log line.

## Environment

- `better-auth`: 1.7.2
- Plugin: `emailOTP` (`better-auth/plugins/email-otp`)
- Node.js 24, Express 5, `drizzleAdapter` (Postgres)

## Related

- #9183 (closed by stale bot, not by fix — now locked)

vyrnsynx:
Fixes #11107.

runInBackgroundOrAwait logged and swallowed awaited promise rejections whenever advanced.backgroundTasks.handler was not configured. sendVerificationOTP then returned success even when the send callback failed.

This change rethrows the original error after logging when no background handler is set, so callers see the original status and payload. Fire-and-forget behavior is unchanged when a handler is configured. Regression tests cover both paths.

---
## Summary by cubic
Fixes #11107: OTP send errors were logged and swallowed when no background handler was configured, so `sendVerificationOTP` returned success on failure. `runInBackgroundOrAwait` now rethrows rejections when no handler is set, so callers see the original error status and payload.

**Behavior**

- Password-reset, forget-password, change-email, and phone password-reset sends are fire-and-forget via `runInBackground`, so registered vs unknown requests aren't distinguishable by latency and still return generic success.
- Send callbacks are deferred through `Promise.resolve` so synchronous throws are logged the same way as rejections.
- Fire-and-forget behavior is unchanged when a handler is configured; failures are logged and don't block the request.
- Adds regression tests for email and phone routes and documents the propagation behavior in the hooks guide and `AuthContext` type comments.

<sup>Written for commit a1dc65b4c786a56cbc090ac22e748b45c44164c7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11108?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/content/docs/concepts/hooks.mdx
+++ b/docs/content/docs/concepts/hooks.mdx
@@ -303,6 +303,8 @@ export const auth = betterAuth({
 
 Defers the task when a handler is configured, otherwise awaits it. Use for operations that must complete (e.g. sending emails) but benefit from not blocking when a handler exists. Configure the handler in [advanced.backgroundTasks](/docs/reference/options#backgroundtasks).
 
+Without a handler, a rejected promise is logged and then rethrown so the caller (and HTTP response) can surface the original error. With a handler, failures are logged on the background promise and do not block the request.
+
 ```ts title="auth.ts"
 import { betterAuth } from "better-auth";
 import { createAuthMiddleware } from "better-auth/api";
--- a/packages/better-auth/src/api/routes/password.ts
+++ b/packages/better-auth/src/api/routes/password.ts
@@ -132,20 +132,18 @@ export const requestPasswordReset = createAuthEndpoint(
 		});
 		const callbackURL = redirectTo ? encodeURIComponent(redirectTo) : "";
 		const url = `${ctx.context.baseURL}/reset-password/${verificationToken}?callbackURL=${callbackURL}`;
-		try {
-			await ctx.context.runInBackgroundOrAwait(
-				Promise.resolve().then(() =>
-					ctx.context.options.emailAndPassword.sendResetPassword(
-						{
-							user: user.user,
-							url,
-							token: verificationToken,
-						},
-						ctx.request,
-					),
+		ctx.context.runInBackground(
+			Promise.resolve().then(() =>
+				ctx.context.options.emailAndPassword.sendResetPassword?.(
+					{
+						user: user.user,
+						url,
+						token: verificationToken,
+					},
+					ctx.request,
 				),
-			);
-		} catch {}
+			),
+		);
 		return ctx.json({
 			status: true,
 			message:
--- a/packages/better-auth/src/api/routes/password.ts
+++ b/packages/better-auth/src/api/routes/password.ts
@@ -132,16 +132,21 @@ export const requestPasswordReset = createAuthEndpoint(
 		});
 		const callbackURL = redirectTo ? encodeURIComponent(redirectTo) : "";
 		const url = `${ctx.context.baseURL}/reset-password/${verificationToken}?callbackURL=${callbackURL}`;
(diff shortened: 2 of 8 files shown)


vyrnsynx:
Addressed the review notes: password-reset and forget-password send failures now keep the generic success response, so they cannot be used to enumerate accounts. OTP verification error surfacing is unchanged.

vyrnsynx:
Follow-up for the latest review: reset SMS/email sends now go through Promise.resolve().then so synchronous throws are logged by runInBackgroundOrAwait. The phone reset test now asserts the registered path actually invoked sendPasswordResetOTP.

vyrnsynx:
Reset, forget-password, change-email, and phone reset sends now use runInBackground so we do not await the provider on registered identities. OTP verification errors still surface via runInBackgroundOrAwait.

vyrnsynx:
Closing this for now. I'll come back with a smaller, more focused patch if this is still needed.

PR #11113, opened 2026-09-02
fix(scim): restore inactive SCIM user on Okta reprovision POST

### LINKED ISSUE #11111: SCIM: Okta cannot reprovision an inactive User because POST conflicts on externalId
- [x] This issue is suited for GitHub.

### Reproduction

This requires an Okta SCIM application connected to a Better Auth app using `@better-auth/scim`.

1. Assign a user to the Okta application. Okta sends `POST /scim/v2/Users`, and the user is provisioned successfully.
2. Deactivate the Okta account. Okta sends a SCIM update setting `active` to `false`. Better Auth correctly keeps the SCIM User resource and marks it inactive.
3. Reactivate the Okta account. Okta has removed the user's former app assignments as part of the account-deactivation lifecycle.
4. Assign the user to the application again. Okta sends a new `POST /scim/v2/Users` with the same stable Okta `externalId` used by the inactive SCIM User.
5. Better Auth responds with `409 Conflict` and the detail `SCIM User externalId already exists`.

Okta then reports:

```text
Push new user to external application FAILURE: Conflict.
Errors reported by remote server: SCIM User externalId already exists
```

Its subsequent “Sync user in external application” operation also fails, so provisioning is not restored.

This is specifically the `active: false` deprovision/reprovision lifecycle. A hard SCIM `DELETE` follows Better Auth's separate tombstone-recovery path and does not reproduce the same retained-row conflict.

**Current:** `POST /scim/v2/Users` checks uniqueness before `identity.resolveUser`. The retained inactive `scimUser` row matches the incoming connection-scoped `externalId`, so `assertSCIMUserKeysAvailable` returns a SCIM uniqueness conflict. The application therefore cannot recover this lifecycle in its identity callbacks.

**Expected:** A `POST` for the same connection and stable `externalId` should support restoring the existing inactive SCIM resource (preserving its SCIM ID), refresh its submitted profile, mark it active, and rerun the normal projection/group reconciliation. An active-resource collision, or an identity belonging to another connection/provisioning domain, should continue to return a conflict.

If implicitly restoring an inactive resource on `POST` is not desirable as the default, the plugin needs a supported hook/recovery result that runs before the uniqueness rejection so applications can implement Okta-compatible reprovisioning safely.

### Better Auth version

`better-auth@1.7.2` and `@better-auth/scim@1.7.2`

### System information

Generated with `npx auth@latest info --json --config ./src/auth.ts`:

```json
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 24.6.0",
    "release": "24.6.0",
    "cpuCount": 16,
    "cpuModel": "Apple M4 Max",
    "totalMemory": "128.00 GB"
  },
  "node": {
    "version": "v24.13.0",
    "env": "development"
  },
  "packageManager": {
    "name": "pnpm",
    "version": "10.20.0"
  },
  "frameworks": [
    {
      "name": "next",
      "version": "16.2.11"
    },
    {
      "name": "react",
      "version": "19.2.7"
    }
  ]
}
```

The CLI could not import the application config because a transitive server-only module was unavailable in the CLI loader. The package versions above are taken from the lockfile and installed package manifests.

### Affected areas

Backend, Package, Enterprise / SCIM

### Example configuration

Sanitized shape of the relevant configuration:

```ts
import { scim } from '@better-auth/scim';

  plugins: [
    scim({
      connections: [],
      managedConnections: {
        credentialHashSecret: process.env.SCIM_CREDENTIAL_HASH_SECRET!,
      },
      identity: {
        resolveUser: async () => ({ action: 'create' }),
      },
    }),
  ],
});
```

### Additional context

The initial provisioning, `active: false` deactivation, profile updates, and role/group projection all work. The failure is limited to restoring a user after Okta's full account deactivation/reactivation cycle removes and later recreates the app assignment.

Inspection of the published `@better-auth/scim@1.7.2` build shows `createSCIMUser` calling `assertSCIMUserKeysAvailable` both before identity resolution and again inside the create transaction. That check does not distinguish an inactive matching resource from an active duplicate. Tombstone resolution cannot help here because the `active: false` path retains the SCIM User instead of deleting it.

breken-ai:
> Built by breken, your AI support engineer - breken.ai - this one's on us.

**Fixes #11111.**

## The problem in plain words

When someone leaves a company, Okta deactivates their account, and the SCIM user in Better Auth is kept but marked inactive. When that person comes back, Okta sends a brand-new "create user" request with the same externalId it always used. Better Auth answers "409: that externalId is taken" - taken by the person's own old, inactive record. Okta can't reprovision them, and every retry fails the same way, so the user stays locked out of every app behind that connection.

## Root cause

`createSCIMUser` in `packages/scim/src/user-provisioning.ts` calls `assertSCIMUserKeysAvailable` before identity resolution. That check matches the retained inactive `scimUser` row on `externalIdKey` and throws a uniqueness conflict. It never distinguishes "this externalId belongs to an active user" (a real conflict) from "this externalId belongs to an inactive user the IdP is trying to bring back" (a restore).

## The fix

When an incoming POST matches an **inactive** SCIM user on the same connection + externalId, restore that row in place instead of rejecting it:

- the SCIM id is preserved, so Okta keeps its mapping;
- the submitted profile is refreshed (same field update as `replaceSCIMUser` / PUT);
- `active` reflects the request (Okta sends `active: true` on reprovision);
- the linked Better Auth user's email/name is updated when this SCIM user is the profile source;
- projection and identity reconciliation rerun, exactly like a PUT;
- the endpoint returns 200 with the resource, since no new resource was created.

An **active** row with the same externalId still returns 409, and a userName now owned by a different user still returns 409 - both enforced inside the restore transaction, so two racing reprovision POSTs degrade to the same conflict as before.

## Behavior change worth knowing about

`POST /scim/v2/Users` with a connection-scoped externalId that belongs to an inactive user now succeeds (200) and revives that user instead of failing with 409. Everything else about the endpoint is unchanged.

## Reproduction and evidence

New test file `packages/scim/src/scim-okta-reprovision.test.ts` replays Okta's lifecycle over real HTTP against a sqlite-backed instance:

1. POST create with an externalId -> 201
2. PATCH `active: false` (deprovision) -> 200
3. POST again with the same externalId (reprovision)
   - before this change: 409 `SCIM User externalId already exists` (the reported bug)
   - after: 200, same SCIM id, `active: true`, refreshed profile
4. Two guard tests: the same externalId on an ACTIVE user still 409s, and an inactive row whose userName is now owned by another user still 409s.

## Verified vs. not verified

- ✅ The restore test fails on master and passes with this change (before/after confirmed).
- ✅ Full `@better-auth/scim` suite passes locally: 334 tests across 33 files, including create-uniqueness, okta-conformance, user-patch, and user-concurrency.
- ⚠️ 3 suite files (scim-http-e2e, scim-sso-http-e2e, scim-sso-registration) could not run in my environment because `@better-auth/test-utils` does not build there; they fail at import resolution, unrelated to this change.
- ⚠️ Not exercised against a live Okta tenant; the lifecycle is modeled from Okta's documented deactivation/reassignment behavior and the reporter's trace.
- ⚠️ Deliberate choice: restore returns 200 rather than 201 because no new resource is created - happy to switch if you prefer create semantics.

Happy to iterate.

---
## Summary by cubic
Fixes Okta reprovisioning (issue #11111) by restoring inactive SCIM users on POST instead of returning a 409 conflict.

**Behavior change**
- `POST /scim/v2/Users` now returns 200 and revives the inactive user when the externalId matches an inactive row on the same connection.
- The SCIM id is preserved and the profile is refreshed like a PUT.
- Active rows with the same externalId still return 409, and userName collisions are still enforced.
- Returns 200 instead of 201 because no new resource is created.

<sup>Written for commit d35a855fb7334ec113a96041e071bae5b1e5aa43. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11113?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/scim/src/scim-okta-reprovision.test.ts
+++ b/packages/scim/src/scim-okta-reprovision.test.ts
@@ -1,0 +1,260 @@
+import { DatabaseSync } from "node:sqlite";
+import { NodeSqliteDialect } from "@better-auth/kysely-adapter/node-sqlite-dialect";
+import { getMigrations } from "better-auth/db/migration";
+import { getHttpTestInstance } from "better-auth/test";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { scim } from ".";
+
+const SCIM_USER_SCHEMA = "urn:ietf:params:scim:schemas:core:2.0:User";
+const SCIM_PATCH_SCHEMA = "urn:ietf:params:scim:api:messages:2.0:PatchOp";
+const SCIM_ERROR_SCHEMA = "urn:ietf:params:scim:api:messages:2.0:Error";
+const SCIM_MEDIA_TYPE = "application/scim+json";
+const SCIM_TOKEN = "okta-reprovision-token";
+
+interface SCIMUserResponse {
+	schemas: string[];
+	id: string;
+	userName: string;
+	externalId?: string;
+	displayName?: string;
+	active: boolean;
+}
+
+interface SCIMErrorResponse {
+	schemas: string[];
+	status: string;
+	detail?: string;
+	scimType?: string;
+}
+
+async function readJSON<T>(response: Response): Promise<T> {
+	return (await response.json()) as T;
+}
+
+function createSCIMHeaders(
+	authorization = `Bearer ${SCIM_TOKEN}`,
+): HeadersInit {
+	return {
+		accept: SCIM_MEDIA_TYPE,
+		authorization,
+		"content-type": SCIM_MEDIA_TYPE,
+	};
+}
+
+async function createTestInstance() {
+	const sqlite = new DatabaseSync(":memory:");
+	const instance = await getHttpTestInstance(
+		{
+			database: {
+				dialect: new NodeSqliteDialect({ database: sqlite }),
+				type: "sqlite",
+				transaction: true,
+			},
+			plugins: [
+				scim({
+					connections: [
+						{
+							id: "okta-reprovision",
+							credentials: [
+								{
+									type: "bearer",
+									id: "okta-reprovision-token",
+									token: SCIM_TOKEN,
+								},
+							],
+						},
+					],
+				}),
+			],
+		},
+		{ disableTestUser: true, testWith: "sqlite" },
+	);
+	await (await getMigrations(instance.auth.options)).runMigrations();
+	return { instance, sqlite };
+}
+
+/**
+ * Okta's account deactivation/reactivation lifecycle removes the user's app
+ * assignment on deactivate and sends a fresh `POST /scim/v2/Users` with the
+ * same stable externalId when the account is re-assigned. The retained
+ * inactive SCIM User must be restored (same SCIM id) instead of rejected
+ * with a uniqueness conflict.
+ *
+ * @see https://github.com/better-auth/better-auth/issues/11111
+ */
+describe("Okta deprovision/reprovision lifecycle", () => {
+	let instance: Awaited<ReturnType<typeof createTestInstance>>["instance"];
+	let sqlite: Awaited<ReturnType<typeof createTestInstance>>["sqlite"];
+	let usersURL: string;
+
+	const externalId = "okta-external-id-11111";
+	const userName = "reprovision.case@example.com";
+
+	beforeAll(async () => {
+		const created = await createTestInstance();
+		instance = created.instance;
+		sqlite = created.sqlite;
+		usersURL = `${instance.baseURL}/api/auth/scim/v2/Users`;
+	});
+
+	afterAll(async () => {
+		await instance.server.close();
+		sqlite.close();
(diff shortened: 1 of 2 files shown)


bytaesu (changes requested):
Check the comments, follow up where needed, and resolve them if there are no issues.

PR #11123, opened 2026-09-02
docs: recommend the Nuxt module setup

onmax:
Adds the Nuxt module to the Resources list.

### DIFF
--- a/docs/content/docs/integrations/nuxt.mdx
+++ b/docs/content/docs/integrations/nuxt.mdx
@@ -5,9 +5,27 @@ title: Nuxt Integration
 description: Integrate Better Auth with Nuxt.
 ---
 
-Before you start, make sure you have a Better Auth instance configured. If you haven't done that yet, check out the [installation](/docs/installation).
+## Install the Nuxt module
 
-## Mount the handler
+The [`@nuxtjs/better-auth`](https://better-auth.nuxt.dev) module is the recommended way to use Better Auth with Nuxt. It mounts the auth handler and adds Nuxt composables for accessing the current session during SSR and on the client.
+
+<Callout type="info">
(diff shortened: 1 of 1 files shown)


bytaesu (changes requested):
I discussed this with a Nuxt team member. The project looks cool, but ideally neither side should need too much magic for the integration.

Hono is a good example: both projects are based on Web Standards and work cleanly together with almost no special setup. That's the direction I'd like us to explore long-term.

docs/content/docs/integrations/nuxt.mdx
```diff
@@ -151,12 +171,15 @@ export default defineNuxtConfig({
...
+  </Accordion>
+</Accordions>
+
 ## Resources & examples
 
 {/* cspell:disable */}
 
 * [`nuxt-modules/better-auth`](https://github.com/nuxt-modules/better-auth)
-* [`atinux/nuxthub-better-auth`](https://github.com/atinux/nuxthub-better-auth)
+* [NuxtHub database integration](https://better-auth.nuxt.dev/integrations/nuxthub)
```
bytaesu: For now, I think it's better to keep the community project in the Resources section rather than officially recommend it. I'd suggest narrowing this PR to simply updating the URL to the latest GitHub repository.

onmax: Thanks for clarifying. I ahve narrowed the PR to a single Resources entry. I do think the module provides a better DX and its nine months of active maintenance are a good start, but I understand the concern and we can revisit this later.

I will work with Sebastian (atinux) and update to latests his template since it is 2 years old. Therefore this PR just adds one line.

docs/content/docs/integrations/nuxt.mdx
```diff
@@ -155,6 +155,7 @@ export default defineNuxtConfig({
 
 {/* cspell:disable */}
 
+* [`@nuxtjs/better-auth`](https://better-auth.nuxt.dev)
```
bytaesu, line 158: These two seem a bit redundant.

I think we don't need a change for now, so let's close this PR and revisit a better integration later!

PR #11154, opened 2026-09-04
fix(auth): identify accounts by providerId and accountId again

Bekacru:
Accounts are identified by `(providerId, accountId)` again, as in 1.6. This removes the `issuer` column added in 1.7.0 and everything built on it.

The column was required, so a populated 1.6 database could not take the 1.7 schema without a per-row backfill, which is why the guided migration existed. In the default configuration the stored value was derived from `providerId`, so the new key carried no extra information. Dropping it removes the schema change, the strategy option, the compatibility warning, and the migration tool.

## Changes

- Remove `account.issuer`, its unique index, and the `local:*` issuer helpers. `AccountKey` is `{ providerId, accountId }`.
- Remove `account.identityStrategy` (#10969), the `accountIssuer` provider option (#10403), and the guided `auth migrate plan` / `auth migrate apply` workflow (#10575). `auth migrate` is a single command again.
- Revert #10986, the demo e2e for the guided migration.
- Generic OAuth no longer skips a provider when discovery returns no issuer. It still skips on an invalid `jwks_uri` or an unsatisfiable `requireIdTokenVerification`.
- Docs, snapshots, e2e fixtures, and the lockfile updated.

Account APIs still select by row `id` or `useAccountCookie`, providers still resolve identity through `accountSubject`, and the SSO provider-id collision guard and domain-verified linking are unchanged.

## Verification

`pnpm typecheck` passes. better-auth, core, and CLI test projects pass except two Postgres-backed tests that need the Docker container.


bytaesu:
Closing this in favor of #11153

PR #11156, opened 2026-09-04
fix(core): make consumeOne and incrementOne optional on custom adapters

Bekacru:
## Problem

1.7.0 (#10000) made `consumeOne` and `incrementOne` required on `CustomAdapter` and removed the factory fallback that 1.6.17 (#9993) shipped with. A custom adapter written for 1.6 now fails `tsc` after upgrading, and at runtime it throws `must implement consumeOne` on the first OTP, magic link, password reset, device code, or two-factor attempt. Nothing checks this at startup, so sign-in works and the failure shows up later in production.

The stated reason for removing the fallback was that it "could not provide the same one-winner guarantee across processes". That holds for read-then-write without a guard. It does not hold for a single-row `DELETE ... WHERE id = ?` or a single `UPDATE ... WHERE id = ? AND counter = ?` with an affected-count check, which every supported store executes atomically.

## Change

- `consumeOne` and `incrementOne` are optional on `CustomAdapter` again.
- Without a native `consumeOne`, the factory runs `findOne`, then `deleteMany` guarded on the row id, and returns the row only when exactly one row was deleted.
- Without a native `incrementOne`, the factory runs a bounded compare-and-swap loop: `findOne`, then `updateMany` guarded on the id and the counter values it just read. An affected count of 0 while the selector still matches means a concurrent writer won, so it retries. A selector that no longer matches returns `null`.
- Non-numeric affected counts from the fallback still fail loud.
- Native implementations remain the fast path and are unchanged for every built-in adapter.

## Tests

`packages/core/src/db/adapter/factory.test.ts` replaces the two "throws when missing" cases with five fallback cases: consume with id guard, consume losing the delete race, CAS increment with the guard shape, CAS retry after a lost race, and guard matching no row.

## Docs

Updated the create-a-db-adapter guide callouts and the 1.7 upgrade guide section to describe the methods as optional with the fallback behavior.

### DIFF
--- a/.changeset/optional-atomic-adapter-methods.md
+++ b/.changeset/optional-atomic-adapter-methods.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/core": patch
+---
+
+Make `consumeOne` and `incrementOne` optional on custom adapters again. When an adapter omits them, the factory falls back to `findOne` plus an id-guarded `deleteMany`, or a compare-and-swap `updateMany` loop, instead of throwing at request time.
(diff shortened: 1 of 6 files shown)


bytaesu:
Replaced by #11189

PR #11164, opened 2026-09-04
docs: add Wraps to the email provider list

stewartjarod:
Adds Wraps to the Bring Your Own Email Provider list on the Email concepts page.

- Guide: https://wraps.dev/docs/guides/better-auth
- npm: https://www.npmjs.com/package/@wraps.dev/better-auth

Wraps sends Better Auth's transactional email — verification, password reset, magic link, OTP, and organization invites — through the app's own AWS SES account.

This PR originally added an entry to the community plugins list. Per review, it now follows the same pattern as the other providers in that section instead.

https://claude.ai/code/session_01UGX6MF5cGx2Eg3iTMhnNbY


bytaesu (commented):
This feels more like adding a preset through options than a plugin that actually extends the auth flow.

bytaesu:
What about adding a Better Auth integration guide to the WRAPS docs and listing it here? (instead of the community plugin approach)

https://better-auth.com/docs/concepts/email#bring-your-own-email-provider

stewartjarod:
You're right, it's presets through `init()`, no endpoints or schema of its own.

The guide already exists: https://wraps.dev/docs/guides/better-auth

Swapped the PR over: dropped the community plugins entry, added Wraps to the Bring Your Own Email Provider list instead. Down to one line.

---

For context on why the plugin route looked right: @sequenzy/better-auth is already in the community list doing the same signup-to-contacts sync.

Fair difference though, Sequenzy hooks the request with createAuthMiddleware, ours only fills in options through init() plus a databaseHooks handler. So your read holds for our case.

The provider listing works for us.

bytaesu (approved):
better-auth grew quickly, so some community projects were listed early on without clear rules. We don't want to remove them, but we've since introduced some review criteria.

Thanks for working on this!

stewartjarod:
Thank you!

PR #11168, opened 2026-09-04
feat(db): validate the database schema against the configuration

Bekacru:
Stacked on #11154.

Better Auth now checks that the database can hold what it writes and fails with the full list of problems instead of failing on the first insert. The rollback in #11154 leaves 1.7.x databases with a required `issuer` column nothing writes; without this, the first sign-up after upgrading fails with a bare constraint error.

## What is checked

- Every table and column Better Auth writes exists.
- A column Better Auth does not write is nullable or has a default. A required column without one rejects every insert into that table.

Each finding comes with its fix, and the `issuer` case points at the index that has to go first.

## Where it runs

- **Drizzle and Prisma**: against the schema definition when `betterAuth()` is called, so a stale schema file fails the build.
- **Built-in database**: one introspection query before the first database call. `betterAuth()` initializes eagerly, often during a build step, so the database is only contacted once a request needs it. The expected schema is captured at construction and a failed check retries on the next call.
- **`auth migrate`** stops before running anything when it finds a required column it cannot fix. **`auth generate`** prints the same findings.

Opt out with `advanced.database.validateSchema: false`.

## Also fixed

- `mergeSchema` mutated the module-level schema constant plugins pass in, so two instances with different `twoFactorTable` names in one process leaked into each other. It returns copies now.
- The internal-adapter test never awaited `runMigrations`.

## Not included

Type and length drift, and missing unique indexes. Both fit behind the same `diffSchema` later.

## Verification

`pnpm typecheck` passes. better-auth, core, CLI, Drizzle, and Prisma suites pass except the two Postgres tests that need Docker.


bytaesu:
Closing this PR in favor of #11178

PR #11171, opened 2026-09-04
feat(cookies): support the __Host- cookie prefix

### LINKED ISSUE #10806: feat(cookies): support the `__Host-` cookie prefix
## Summary

`createCookieGetter` can emit `__Secure-` but never `__Host-`. `HOST_COOKIE_PREFIX` is already declared in `cookie-utils.ts` and `stripSecureCookiePrefix` already handles it on the read side — only emission is missing.

## Why

`__Secure-` stops a plaintext-HTTP write on the same host from overwriting a session cookie. It does not stop a **related-domain** write.

In a multi-tenant deployment where tenants are subdomains of one registrable domain (`acme.example.com`, `globex.example.com`), any sibling can set `Domain=example.com` — a compromised tenant, an XSS'd one, or a dangling DNS record in the zone. The browser then sends that cookie to every other sibling. Cookies carry no record of which host set them, so when two arrive under the same name the server cannot tell them apart, and the collision becomes cross-tenant session fixation.

`__Host-` closes this structurally: a cookie carrying a `Domain` attribute is not allowed to use a `__Host-` name, so an injected cookie can never collide with the real one. It is the standard mitigation for this topology, and tenant-per-subdomain is common enough to be worth supporting.

## Current state

- `cookie-utils.ts` — `HOST_COOKIE_PREFIX = "__Host-"` is declared, and `stripSecureCookiePrefix` strips either prefix.
- `cookies/index.ts:71` — the only emit path, hardcoded to the other constant:
  ```ts
  const secureCookiePrefix = secure ? SECURE_COOKIE_PREFIX : "";
  ```
- `cookies/index.ts` imports `HOST_COOKIE_PREFIX` solely to re-export it.

Related: #4908 / #5172. The plugin docs used to say `createAuthCookie` "implements things like `__secure` prefix and `__host` prefix"; the fix was to delete the `__host` claim. So the gap is known — this issue asks to close it in the other direction.

## What the defaults already give you

With `crossSubDomainCookies` disabled, `createCookie` already emits `Secure`, `Path=/`, and no `Domain` — exactly the three conditions `__Host-` requires. The prefix is the only missing piece.

## Proposed shape

Happy to follow whatever you prefer:

- `advanced.useHostCookiePrefix?: boolean`, or extend `useSecureCookies` to accept `"host"`.
- Throw when combined with `advanced.crossSubDomainCookies.enabled` — mutually exclusive by definition, in the style of the existing `baseURL is required when crossSubdomainCookies are enabled` check.
- Reject a `path` override in `advanced.cookies.*.attributes` / `defaultCookieAttributes`, since `__Host-` requires `Path=/`.

## Two things that make this more than a one-liner

1. **Existing sessions.** Changing the prefix renames the cookie. `getSessionCookie` (`cookies/index.ts:484`) falls back `__Secure-${name}` → `${name}`, so a deployment that switches to `__Host-` finds no existing cookie and signs every user out. Either document it as breaking and opt-in, or widen the chain to `__Host-` → `__Secure-` → bare — though that fallback reintroduces exactly the collision the prefix exists to prevent, so the former seems right.
2. **Other packages reconstruct the prefix.** `packages/electron/src/cookies.ts` and `packages/expo/src/client.ts` both build `__Secure-` names directly and would need the same branch.

## Offer

Happy to open the PR if you'd like it. Tell me which option shape you want and how you'd prefer (1) to be handled, and I'll implement it with tests.

### LINKED ISSUE #11061: Session cookie name cannot use the __Host- prefix (cookiePrefix becomes a suffix of __Secure-)
### Summary

The session cookie name cannot be configured to use the [`__Host-` prefix](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#cookie_prefixes). Any configured `advanced.cookiePrefix` (or an explicit `advanced.cookies.session_token.name`) becomes a *suffix* of the hardcoded `__Secure-` prefix.

### Detail

In `src/cookies/index.ts`, `secureCookiePrefix` is the constant `__Secure-` whenever secure cookies are on, and every cookie name is built as `` `${secureCookiePrefix}${name}` ``.

With:

```ts
advanced: {
  cookiePrefix: '__Host-myapp',
  useSecureCookies: true,
}
```

the emitted header is:

```
Set-Cookie: __Secure-__Host-myapp.session_token=...; Path=/; HttpOnly; Secure; SameSite=Lax
```

That is a `__Secure-` cookie with the characters `__Host-` inert in the middle. The read path agrees with the write path (`getCookie` looks up `__Secure-${name}`), so behaviour is self-consistent — just not what was configured.

### Why it matters

`__Secure-` and `__Host-` are not equivalent:

| Prefix | Browser enforces |
|---|---|
| `__Secure-` | `Secure` only — the cookie may carry `Domain` |
| `__Host-` | additionally no `Domain`, and `Path=/` |

`__Host-` is what prevents a sibling or sub-domain from setting a session cookie for the parent host (cookie fixation). Applications that set `Path=/` and omit `Domain` already meet `__Host-`'s requirements in substance, but without the prefix the browser does not enforce them, so the protection is voluntary rather than guaranteed.

### Version

`better-auth@1.7.2`.

### Reproduction / sentinel test

```ts
// Emitted name starts with __Secure-, never __Host-, regardless of cookiePrefix.
const res = await auth.api.signInEmail({ body: { email, password }, returnHeaders: true })
const cookie = res.headers.getSetCookie().find((c) => c.includes('session_token'))!

expect(cookie).toMatch(/Path=\//)
expect(cookie).not.toMatch(/Domain=/)
expect(cookie).toMatch(/Secure/)

expect(cookie.startsWith('__Secure-')).toBe(true)
expect(cookie.startsWith('__Host-')).toBe(false)   // <- the limitation
```

### Possible resolution

Allow the secure prefix to be chosen — e.g. `advanced.securePrefix: '__Secure-' | '__Host-'`, or honour an explicit `advanced.cookies.*.name` verbatim without prepending — with `__Host-` validated against its own requirements (`Secure`, `Path=/`, no `Domain`) so it cannot be selected in a configuration where the browser would reject the cookie.

Happy to open a PR if you'd like a particular shape.

Bekacru:
## Summary

Adds `advanced.useHostCookiePrefix` so Better Auth can name its cookies with the `__Host-` prefix instead of `__Secure-`, plus an opt-in `migrateSecureCookies` for a seamless cutover of existing `__Secure-` sessions.

## Root cause

`createCookieGetter` only ever emitted `""` or `__Secure-` as the cookie prefix. `HOST_COOKIE_PREFIX` was declared in `cookie-utils.ts` and stripped on the read side, but nothing on the emit path produced it. Putting `__Host-` into `advanced.cookiePrefix` therefore produced `__Secure-__Host-...`, which browsers reject because the name does not start with a valid prefix. `getSessionCookie`, `getCookieCache`, the multi-session plugin, and the Electron/Expo packages also only understood `__Secure-` names.

## Fix

- `advanced.useHostCookiePrefix?: boolean | { enabled, migrateSecureCookies? }` (`@better-auth/core`). When enabled, cookies are emitted as `__Host-<name>`, `Secure` is forced, and the getter throws at init on configurations that browsers would reject: `crossSubDomainCookies` (needs a `Domain`), a `path` other than `/`, a per-cookie `domain`, and an explicit `useSecureCookies: false`.
- Enabling it alone renames the cookies, so existing `__Secure-` sessions are no longer read and signed-in users sign in again once. This is the default: accepting `__Secure-` cookies is exactly the collision the prefix exists to prevent.
- `migrateSecureCookies: true` (default `false`) opts into a seamless cutover: an internal plugin mirrors `__Secure-*` request cookies to their `__Host-*` name for the request when no `__Host-` twin exists, then on the response re-issues them under the `__Host-` name and expires the `__Secure-` copies. Sessions migrate on their first request after the upgrade. The docs tell operators to remove it after `session.expiresIn`, since a visitor without a `__Host-` cookie is exposed to a planted `__Secure-` cookie while it is on.
- `getSessionCookie` and `getCookieCache` prefer a `__Host-` cookie (including chunked `session_data`) before the `__Secure-`/bare names.
- Multi-session restores `__Host-` casing when expiring per-session cookies; Electron uses `stripSecureCookiePrefix` instead of slicing `__Secure-` by hand; Expo checks the `__Host-` OAuth state cookie name.
- Docs: `concepts/cookies.mdx`, `reference/options.mdx`, `concepts/plugins.mdx`.

## Tests

`packages/better-auth/src/cookies/cookies.test.ts` (`useHostCookiePrefix` block):
- sign-in emits `__Host-better-auth.session_token` with `Secure`, `Path=/`, no `Domain`, and no `__Secure-` cookie; the session reads back
- `crossSubDomainCookies`, a non-`/` path, and `useSecureCookies: false` are rejected
- `getSessionCookie` prefers `__Host-` over `__Secure-` and bare names
- `getCookieCache` reads the `__Host-` `session_data` cookie
- by default, a session signed in under `__Secure-` is not read after switching
- with `migrateSecureCookies: true`, that session stays valid, is re-issued as `__Host-`, the `__Secure-` copy is expired, and the migrated cookies work on their own

Ran `pnpm vitest run src/cookies` (119 passed), `src/plugins/multi-session` (10 passed), `@better-auth/electron` (88 passed), `@better-auth/expo` (74 passed), and `pnpm typecheck` (passes).

Closes #11061
Closes #10806

### DIFF
--- a/.changeset/cookies-host-prefix.md
+++ b/.changeset/cookies-host-prefix.md
@@ -1,0 +1,8 @@
+---
+"better-auth": patch
+"@better-auth/core": patch
+"@better-auth/electron": patch
+"@better-auth/expo": patch
+---
+
+Add `advanced.useHostCookiePrefix` to name cookies with the `__Host-` prefix, reject configurations that are incompatible with it (`crossSubDomainCookies`, a custom cookie `path`, `useSecureCookies: false`), and offer an opt-in `migrateSecureCookies` that keeps existing `__Secure-` sessions alive by re-issuing them as `__Host-` cookies on their next request.
--- a/docs/content/docs/concepts/cookies.mdx
+++ b/docs/content/docs/concepts/cookies.mdx
@@ -105,6 +105,35 @@ export const auth = betterAuth({
 })
 ```
 
+### `__Host-` Prefix
+
+Secure cookies are named with the `__Secure-` prefix. Set `useHostCookiePrefix` to use the [`__Host-` prefix](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#cookie_prefixes) instead. Browsers only accept a `__Host-` cookie when it is `Secure`, has `Path=/`, and carries no `Domain` attribute, so a cookie set by a sibling subdomain can never collide with your session cookie. Use it when tenants live on subdomains of one registrable domain.
+
+`useHostCookiePrefix` forces secure cookies and cannot be combined with `useSecureCookies: false`, `crossSubDomainCookies`, or a custom cookie `path`.
+
+```ts title="auth.ts"
+import { betterAuth } from "better-auth"
+
+export const auth = betterAuth({
+    advanced: {
+        useHostCookiePrefix: true
+    }
+})
+```
+
+Switching renames the cookies, so existing `__Secure-` sessions are no longer read and signed-in users sign in again once. For a seamless cutover, set `migrateSecureCookies: true`: existing `__Secure-` cookies are still accepted and are re-issued as `__Host-` cookies on their next request, with the old copies expired. While that migration is on, a visitor who has no `__Host-` cookie yet is still exposed to a `__Secure-` cookie planted by a sibling subdomain, so turn it off once existing sessions have rolled over (after `session.expiresIn`):
+
+```ts title="auth.ts"
+export const auth = betterAuth({
+    advanced: {
+        useHostCookiePrefix: {
+            enabled: true,
+            migrateSecureCookies: true // remove after session.expiresIn
+        }
+    }
+})
+```
+
 ## Safari, ITP, and Cross-Domain Setups
 
 Safari includes a privacy feature called Intelligent Tracking Prevention (ITP) that blocks third-party cookies.
--- a/docs/content/docs/concepts/plugins.mdx
+++ b/docs/content/docs/concepts/plugins.mdx
@@ -129,7 +129,7 @@ Create Auth endpoints wraps around `createEndpoint` from Better Call. Inside the
 * `db`: The Kysely instance used by Better Auth to interact with the database.
 * `adapter`: This is the same as db but it give you `orm` like functions to interact with the database. (we recommend using this over `db` unless you need raw sql queries or for performance reasons)
 * `internalAdapter`: These are internal db calls that are used by Better Auth. For example, you can use these calls to create a session instead of using `adapter` directly. `internalAdapter.createSession(userId)`
-* `createAuthCookie`: This is a helper function that lets you get a cookie `name` and `options` for either to `set` or `get` cookies. It implements things like `__Secure-` prefix for cookies based on whether the connection is secure (HTTPS) or the application is running in production mode.
+* `createAuthCookie`: This is a helper function that lets you get a cookie `name` and `options` for either to `set` or `get` cookies. It implements things like the `__Secure-` prefix for cookies based on whether the connection is secure (HTTPS) or the application is running in production mode, or the `__Host-` prefix when `advanced.useHostCookiePrefix` is enabled.
 * `trustedOrigins`: This is the list of trusted origins that you specified via `options.trustedOrigins`.
 * `isTrustedOrigin`: This is a helper function that allows you to quickly check whether a given url or path is trusted based on the trusted origins configuration.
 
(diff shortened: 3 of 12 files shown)


bytaesu:
I'll handle this in the PR https://github.com/better-auth/better-auth/pull/11315

PR #11175, opened 2026-09-04
fix(oauth-provider): drop unsupported grant types from dynamic registration instead of rejecting it

### LINKED ISSUE #11081: oauth-provider: DCR rejects Claude when an unsupported grant accompanies supported grants
## Is this suited for GitHub?

- [x] Yes, this is suited for GitHub

## Reproduction

<details>
<summary>Minimal stock-1.7.2 reproduction</summary>

Install the released packages and Vitest:

```sh
npm i better-auth@1.7.2 @better-auth/oauth-provider@1.7.2 vitest@4
```

Save this as `repro.test.mjs` and run `npx vitest run repro.test.mjs`:

```js
import { oauthProvider } from '@better-auth/oauth-provider'
import { getTestInstance } from 'better-auth/test'
import { jwt } from 'better-auth/plugins/jwt'
import { expect, test } from 'vitest'

const { customFetchImpl } = await getTestInstance({
  baseURL: 'http://localhost:3000',
  plugins: [
    jwt(),
    oauthProvider({
      loginPage: '/login',
      consentPage: '/consent',
      allowUnauthenticatedClientRegistration: true,
    }),
  ],
})

test('registers a client when at least one grant type is supported', async () => {
  const response = await customFetchImpl('http://localhost:3000/api/auth/oauth2/register', {
    method: 'POST',
    headers: { 'content-type': 'application/json' },
    body: JSON.stringify({
      redirect_uris: ['https://claude.ai/api/mcp/auth_callback'],
      token_endpoint_auth_method: 'none',
      grant_types: [
        'authorization_code',
        'refresh_token',
        'urn:ietf:params:oauth:grant-type:jwt-bearer',
      ],
      response_types: ['code'],
    }),
  })

  const body = await response.text()
  expect(response.status, body).toBe(201)
})
```

The test fails with:

```text
{"error":"invalid_client_metadata","error_description":"unsupported grant_type urn:ietf:params:oauth:grant-type:jwt-bearer"}: expected 400 to be 201
```

</details>

## Current and expected behavior

Claude sends this grant union when it uses OAuth dynamic client registration for an MCP connector. `authorization_code` and `refresh_token` are supported, but the additional enterprise `jwt-bearer` grant makes Better Auth reject the whole registration.

PR #10731, released on the 1.7 line through #11010, fixed the same interoperability problem for Client ID Metadata Documents. It deliberately left dynamic registration strict.

For DCR, RFC 7591 section 2 permits the authorization server to replace requested metadata with suitable registered values. Better Auth could narrow `grant_types` to the supported intersection, return those values in the registration response, and reject only when the intersection is empty. The token endpoint would still reject grants for which the server has no handler.

The same Claude client could then use either MCP registration mode without a client-specific `jwt-bearer` exception.

## Better Auth version

`better-auth@1.7.2` and `@better-auth/oauth-provider@1.7.2`

## System

```text
Node 24.19.0
```

## Which areas are affected?

- Backend
- OAuth Provider plugin

## References

- RFC 7591 section 2: https://datatracker.ietf.org/doc/html/rfc7591#section-2
- CIMD fix: https://github.com/better-auth/better-auth/pull/10731
- v1.7 backport: https://github.com/better-auth/better-auth/pull/11010

Bekacru:
## Summary

Dynamic client registration now narrows the requested `grant_types` to the grants the server supports and returns the effective list in the registration response, instead of rejecting the whole registration because one unsupported grant accompanies supported ones. A registration that shares no grant with the server is still rejected with `invalid_client_metadata`.

## Root cause

`checkOAuthClient` in `packages/oauth-provider/src/register.ts` required every requested grant to be supported for non-CIMD registrations. Claude's MCP client registers via DCR with `["authorization_code", "refresh_token", "urn:ietf:params:oauth:grant-type:jwt-bearer"]`, so the `jwt-bearer` entry made the whole request fail and the connection never completed. #10731 relaxed the same check for Client ID Metadata Documents but deliberately left DCR strict.

## Fix

- Dynamic registration and CIMD share one check: at least one requested grant must be supported. Dynamic registration then replaces the requested set with the supported subset (RFC 7591 §3.2.1); CIMD keeps its union as before and the token endpoint refuses unsupported grants.
- The supported-grant check moved ahead of the redirect URI and `authorization_code`/`code` correlation checks so they run against the narrowed set. A `code` response type left over after `authorization_code` is dropped is still rejected.
- `checkOAuthClient` returns the effective `grant_types`; `persistOAuthClientRegistration` stores them so the registration response reflects what was registered.
- Managed (administrator) registration and client updates remain strict about unsupported grants.
- Docs: noted the narrowing behavior in the Dynamic Client Registration section.

## Tests

`packages/oauth-provider/src/register.test.ts`:
- registering with a supported grant next to an unsupported one succeeds, and both the response and the stored client list only the supported grants
- a dynamic registration with only unsupported grants is rejected
- `response_types` is validated against the narrowed grants
- managed registration still rejects an unsupported grant
- the CIMD intersection test now asserts the returned union

`pnpm vitest run` in `packages/oauth-provider`: 40 files, 930 tests passed. `pnpm typecheck` passes.

Closes #11081

### DIFF
--- a/.changeset/oauth-provider-dcr-grants.md
+++ b/.changeset/oauth-provider-dcr-grants.md
@@ -1,0 +1,5 @@
+---
+"@better-auth/oauth-provider": patch
+---
+
+Narrow dynamic client registration `grant_types` to the grants the server supports instead of rejecting the whole registration when an unsupported grant accompanies supported ones.
--- a/docs/content/docs/plugins/oauth-provider.mdx
+++ b/docs/content/docs/plugins/oauth-provider.mdx
@@ -1318,6 +1318,8 @@ oauthProvider({
 
 Dynamic registration allows for authorized registration of both public and confidential clients.
 
+Requested `grant_types` are narrowed to the grants this server supports and the effective list is returned in the registration response ([RFC 7591 §3.2.1](https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.1)). A registration sharing no grant with the server is rejected with `invalid_client_metadata`.
+
 ```ts title="auth.ts"
 oauthProvider({
   allowDynamicClientRegistration: true, // [!code highlight]
--- a/packages/oauth-provider/src/register.test.ts
+++ b/packages/oauth-provider/src/register.test.ts
@@ -537,7 +537,13 @@ describe("oauth register", async () => {
 					},
 					{ registrationSource: "clientMetadataDocument" },
 				),
-			).resolves.toBeUndefined();
+			).resolves.toEqual({
+				grant_types: [
+					"authorization_code",
+					"refresh_token",
+					"urn:ietf:params:oauth:grant-type:jwt-bearer",
+				],
+			});
 		});
 
 		it("rejects a client metadata document sharing no grant with the server", async () => {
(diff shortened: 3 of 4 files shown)


onmax:
Related: https://github.com/better-auth/better-auth/pull/10987 which fixes a separate OAuth interoperability issue affecting ChatGPT’s `private_key_jwt` authentication with UUID-backed databases

PR #11177, opened 2026-09-04
fix(react): return the pending session as the server snapshot during hydration

### LINKED ISSUE #10972: BUG: `useSession()` causes a React hydration mismatch: `useStore` passes `getSnapshot` as `getServerSnapshot`
By the way, Opus 5 helped me debug this error I was having; I'll leave all the information below.

---

## Environment

| | |
|---|---|
| `better-auth` | 1.7.1 |
| `nanostores` | 1.5.1 (transitive) |
| React | 19.2.8 |
| Next.js | 16.3.1 (App Router, Turbopack, dev mode) |
| Client | `createAuthClient()` from `better-auth/react` |

## Summary

Any client component that branches on `authClient.useSession()` and is server-rendered can hydrate against different markup than the server produced. React logs:

```
A tree hydrated but some attributes of the server rendered HTML didn't match the client properties.
This won't be patched up.
```

The mismatch is **intermittent by design of the race**, not random: it depends on whether the session store resolves before that particular subtree hydrates. In our app one page mismatched on every single load while another never did.

## Reproduction

A client component whose markup depends on `isPending`:

```tsx
"use client"
import { authClient } from "@/lib/auth/client"

export function SaveButton() {
  const { isPending } = authClient.useSession()
  return <button aria-disabled={isPending}>Save</button>
}
```

Server-render it deep in a page that also renders something subscribing to the session higher up (a header with a user menu, for example).

Observed on a property detail page — the server sent `aria-disabled="true"`, the client hydrated with `false`:

```
  <button
    aria-busy={false}
+   aria-disabled={false}
-   aria-disabled="true"
    aria-label="Save property"
  >
```

The same component on a different route never mismatched, because there its markup was not part of the SSR payload.

## Root cause

`dist/client/react/react-store.mjs` (vendored from `@nanostores/react`):

```js
function useStore(store, options = {}) {
	const snapshotRef = useRef(store.get());
	// ...
	const get = () => snapshotRef.current;
}
```

The third argument to `useSyncExternalStore` is `getServerSnapshot`. React calls it **both** during server rendering **and** during client hydration, precisely so the two renders agree. Here it is the same function as `getSnapshot`, so during hydration it returns the *live* client store value.

That value can already be resolved by then. From `dist/client/query.mjs`, the session atom starts pending:

```js
const value = atom({ data: null, error: null, isPending: true, isRefetching: false, refetch })
```

and flips to `isPending: false` in `onSuccess`. The fetch is started from `onMount`, i.e. when the atom gets its first subscriber. Because React hydration is not atomic — subtrees commit at different times, and Suspense boundaries widen the gap — a subscriber near the top of the tree can trigger the fetch and have it resolve before a subtree lower down hydrates. That lower subtree then reads `isPending: false` from `getServerSnapshot`, while the server had rendered `isPending: true`.

This also explains the asymmetry we saw: components rendered only on the client (behind a client-side data fetch) never mismatch, because there is no server markup to compare against.

## Expected behavior

`getServerSnapshot` should return the same value on the server and during hydration, so the first client render reproduces the server's markup. The real value should land on the next render, after the store subscription is live.

## Suggested fix

`useStore` cannot know a generic store's initial value, but `better-auth` does know it for the atoms it creates in `useAuthQuery`: the pending state the atom is constructed with. Two options:

1. Give `useStore` an explicit server snapshot option and pass the atom's initial value from `createAuthClient`:

```js
return useSyncExternalStore(subscribe, get, getServerSnapshot ?? get)
```

2. Keep a reference to the initial state object inside `useAuthQuery` and expose it on the atom, so the React adapter can return it as the server snapshot.

Either way the server snapshot must be a **stable reference** — returning a freshly built object on each call makes React loop.

## Note on `hydrateSession`

`createAuthClient` exposes `hydrateSession`, but it does not address this. `hydrateSessionAtom` returns early on the server:

```js
function hydrateSessionAtom(sessionAtom, session) {
	if (typeof window === "undefined") return;
	// ...
	sessionAtom.set({ ...currentSession, data: session, error: null, isPending: false });
}
```

So it can only change what the *client* renders, never what the server rendered. If anything it widens the gap. (We have not tested it in our app — this is read from the source.)

## Workaround

Pin the first client render to the pending state so both sides agree, and let the real value arrive one render later:

```tsx
"use client"

import { useEffect, useState } from "react"
import { authClient } from "@/lib/auth/client"

export function useSession() {
  const { data, isPending } = authClient.useSession()
  const [isHydrated, setIsHydrated] = useState(false)

  useEffect(() => {
    setIsHydrated(true)
  }, [])

  return {
    isPending: isPending || !isHydrated,
    session: isHydrated ? data : null,
  }
}
```

This removed the mismatch on every route we could reproduce it on. The cost is one extra render, and any pending placeholder now always shows for a frame.

## Scope of what was verified

- The source above is from the installed `better-auth@1.7.1` build, quoted verbatim.
- The mismatch and the workaround were both verified in our application (reproduced on every load of one route, gone after the change).
- We did **not** build a standalone minimal reproduction, and the precise hydration ordering that wins the race is inferred from React's behaviour, not instrumented.

Bekacru:
## Summary

Client components that server-render on `authClient.useSession()` (or any other React atom hook) could hydrate against different markup than the server produced, logging a React hydration mismatch. The failure was intermittent because it depended on whether the session store resolved before that particular subtree hydrated.

## Root cause

`useStore` in `packages/better-auth/src/client/react/react-store.ts` passed its live `getSnapshot` as the third argument to `useSyncExternalStore`, which is `getServerSnapshot`. React calls `getServerSnapshot` on the server and again during client hydration and expects both to return the same value. The server rendered the atom's initial pending state, but by the time a lower subtree hydrated, a subscriber higher in the tree (or `hydrateSession`) could already have resolved the store, so the live getter returned `isPending: false` against server markup rendered with `isPending: true`.

## Fix

- `useStore` accepts an optional `getServerSnapshot` and passes it to `useSyncExternalStore`, falling back to the live getter as before.
- `createAuthClient` in `packages/better-auth/src/client/react/index.ts` passes the atom's `init` (nanostores' stable initial-value reference) for every generated hook. Nothing mounts an atom during SSR, so `init` is exactly what the server rendered. Computed stores have no `init` and keep the previous behavior.

Hydration now renders the initial state and React re-renders with the live value immediately after commit. Client-only rendering is unaffected since React never calls `getServerSnapshot` outside SSR and hydration. `init` is read instead of `get()` because `get()` on an unmounted nanostores atom mounts it and would start the session fetch at client creation.

## Tests

- `packages/better-auth/src/client/react/react-store.test.ts`: renders `useSession` with `renderToString`, resolves the store via `hydrateSession`, hydrates with `hydrateRoot`, and asserts no recoverable error or console error, that the hydration render used the pending state, and that the DOM then reflects the live value. Fails on `main` with the exact `aria-disabled` and text mismatch from the issue.
- `pnpm vitest run src/client` in `packages/better-auth`: 10 files, 119 tests passed.
- `pnpm typecheck` passes.

Closes #10972

### DIFF
--- a/.changeset/react-server-snapshot.md
+++ b/.changeset/react-server-snapshot.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Return the pending session as the React server snapshot during hydration so `useSession` no longer causes hydration mismatches when the session resolves before a subtree hydrates.
--- a/packages/better-auth/src/client/react/index.ts
+++ b/packages/better-auth/src/client/react/index.ts
@@ -92,7 +92,14 @@ export function createAuthClient<Option extends BetterAuthClientOptions>(
 	} = getClientConfig(options);
 	const resolvedHooks: Record<string, any> = {};
 	for (const [key, value] of Object.entries(pluginsAtoms)) {
-		resolvedHooks[getAtomKey(key)] = () => useStore(value);
+		// Nothing mounts an atom during SSR, so the server rendered its initial
+		// value. Hand React that same value while hydrating; the live value may
+		// already have resolved. Computed stores have no `init` and keep the
+		// live value.
+		const getServerSnapshot =
+			value.init === undefined ? undefined : () => value.init;
+		resolvedHooks[getAtomKey(key)] = () =>
+			useStore(value, { getServerSnapshot });
 	}
 
 	const routes = {
--- a/packages/better-auth/src/client/react/react-store.test.ts
+++ b/packages/better-auth/src/client/react/react-store.test.ts
@@ -1,0 +1,83 @@
+// @vitest-environment happy-dom
+import { act, createElement } from "react";
+import type { Root } from "react-dom/client";
+import { hydrateRoot } from "react-dom/client";
+import { renderToString } from "react-dom/server";
+import { afterEach, expect, it, vi } from "vitest";
+import { createAuthClient } from "./index";
+
+afterEach(() => {
+	vi.unstubAllGlobals();
+	vi.restoreAllMocks();
+});
+
+/**
+ * @see https://github.com/better-auth/better-auth/issues/10972
+ */
+it("hydrates useSession with the pending state the server rendered", async () => {
+	vi.stubGlobal("IS_REACT_ACT_ENVIRONMENT", true);
+	const client = createAuthClient({
+		baseURL: "http://localhost:3000",
+		fetchOptions: {
+			customFetchImpl: () => new Promise<Response>(() => {}),
+		},
+	});
+	const rendered: boolean[] = [];
+	const SaveButton = () => {
+		const { isPending } = client.useSession();
+		rendered.push(isPending);
+		return createElement(
+			"button",
+			{ "aria-disabled": isPending },
+			isPending ? "loading" : "save",
+		);
+	};
+
+	const html = renderToString(createElement(SaveButton));
+	expect(html).toContain('aria-disabled="true"');
+
+	// A subscriber higher in the tree already resolved the session before this
+	// subtree hydrates.
+	const now = new Date();
+	client.hydrateSession({
+		user: {
+			id: "user-id",
+			name: "Test User",
+			email: "test@example.com",
+			emailVerified: true,
+			createdAt: now,
+			updatedAt: now,
+		},
+		session: {
+			id: "session-id",
+			userId: "user-id",
+			token: "token",
+			expiresAt: now,
+			createdAt: now,
+			updatedAt: now,
+		},
+	});
+	expect(client.$store.atoms.session!.get().isPending).toBe(false);
+
+	const container = document.createElement("div");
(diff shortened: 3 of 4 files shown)


Smidge:
Opened #11316 as an alternative follow-up, with credit to this PR's diagnosis and stable server snapshot approach.

The new implementation limits the snapshot override to auth query atoms. Nano Stores computed stores also retain an `init` property, so checking its presence would change computed SSR output. It also preserves ordinary plugin atoms' existing SSR values, including atoms initialized with `undefined`.

The new tests reproduce four hydration failures on a fresh build of `main` and pass with the fix. All 123 client tests and the package typecheck pass locally. The follow-up includes documentation and a patch changeset. Happy for the implementation and tests to be folded into this PR if that is easier to review.

bytaesu:
Closing this PR in favor of #11316

PR #11184, opened 2026-09-05
docs: make header dropdown menus keyboard accessible

remcostoeten:
As a keybord  user I noticed the header isn't accessible through keybord controls, so i fixed it.

The `products` and `resources` header tabs were hover-only `div`s, so keyboard users skipped both dropdowns.

Triggers are now `<button>`s with `aria-expanded`. The menu opens on focus, closes on Escape or when focus leaves the tab, and Tab moves through the links inside. 

https://github.com/user-attachments/assets/62fee45d-d62b-4c32-84d8-5ccd30898224


bytaesu (approved):
I missed this since I hadn't tried opening the nav with the keyboard. Hope this makes it a bit easier to use 😁

PR #11185, opened 2026-09-05
fix(oauth-provider): distinguish post-login continuation from initial selection

### LINKED ISSUE #11093: oauth-provider: postLogin selection page loops forever for multi-org users — post_login redirect never carries the ba_pl marker
### Reproduction

Configure the oauth-provider plugin exactly as shown in the docs section [“Post Login Screen”](https://www.better-auth.com/docs/plugins/oauth-provider) (organization selection before consent):

```ts
oauthProvider({
  scopes: ["openid", "profile", "email", "read:organization"],
  postLogin: {
    page: "/select-organization",
    shouldRedirect: async ({ session, scopes, headers }) => {
      const userOnlyScopes = ["openid", "profile", "email", "offline_access"];
      if (scopes.every((sc) => userOnlyScopes.includes(sc))) return false;
      const organizations = await auth.api.listOrganizations({ headers });
      return organizations.length > 1 || !(
        organizations.length === 1 && organizations.at(0)?.id === session.activeOrganizationId
      );
    },
    consentReferenceId: ({ session }) => session?.activeOrganizationId ?? undefined,
  },
})
```

Steps:

1. Sign in as a user who belongs to **two or more** organizations.
2. Start an authorization request that includes an org scope: `GET /oauth2/authorize?...&scope=openid+read:organization`.
3. You are redirected to `/select-organization?<signed query>` as expected.
4. On that page, follow the docs snippet:
   ```ts
   await authClient.organization.setActive({ organizationId });
   await client.oauth2.oauth2Continue({ postLogin: true });
   ```
5. The `oauth2Continue` response `url` is `/select-organization?<signed query>` again — not the consent page. Repeating steps 4–5 loops forever.

Observed on `@better-auth/oauth-provider` 1.7.1 and 1.7.2 (dist), and the relevant code is unchanged on current `main` (permalinks below). Users with exactly **one** organization are not affected, which makes the loop easy to miss in manual testing.

**Expected:** after the user selects an organization, `oauth2Continue({ postLogin: true })` advances the flow to the consent page (or straight to the redirect with a code when consent is already on file).

**Current:** the flow can never leave the post-login page for multi-org users:

1. The `ba_pl` (`postLoginClearedParam`) marker is only minted when redirecting to the **consent** page — never for `post_login` redirects: [authorize.ts#L977-L978](https://github.com/better-auth/better-auth/blob/3e4343921890eee1c7f0f5f1b9aa66c1321b64f8/packages/oauth-provider/src/authorize.ts#L977-L978). So the signed query on `postLogin.page` never contains `ba_pl`.
2. `POST /oauth2/continue { postLogin: true }` derives `settings.postLogin` exclusively from that marker: [continue.ts#L95-L97](https://github.com/better-auth/better-auth/blob/3e4343921890eee1c7f0f5f1b9aa66c1321b64f8/packages/oauth-provider/src/continue.ts#L95-L97). With no marker it is always `false`, so authorize re-runs `shouldRedirect` against the live session.
3. For a multi-org user the documented `shouldRedirect` returns `organizations.length > 1 || …`, which is still `true` after `organization.setActive` — so authorize redirects back to `postLogin.page`: [authorize.ts#L795](https://github.com/better-auth/better-auth/blob/3e4343921890eee1c7f0f5f1b9aa66c1321b64f8/packages/oauth-provider/src/authorize.ts#L795). Infinite loop.

In other words: the comment in `continue.ts` says the flow should be released by `shouldRedirect` returning `false` once selection completed, but the docs example never becomes `false` for multi-org users, and any “ask which org to authorize on every grant” UX (arguably the natural behavior for an OAuth consent flow) is impossible to express — there is no flow-scoped signal an app could consult, because that is exactly what `ba_pl` is and it is never minted for this page.

The existing tests only cover rejection of *forged* markers ([authorize.test.ts](https://github.com/better-auth/better-auth/blob/3e4343921890eee1c7f0f5f1b9aa66c1321b64f8/packages/oauth-provider/src/authorize.test.ts)); there is no happy-path test for the multi-org postLogin → continue transition.

### Suggested fix

Mint the session-bound marker for `post_login` redirects too — the session is already authenticated at that call site, and `continue` already verifies the marker against the live session id, so a stale or copied marker still fails. This is a two-line change:

```diff
-			return redirectWithPromptCode(ctx, opts, "post_login");
+			return redirectWithPromptCode(ctx, opts, "post_login", { sessionId: session.session.id });
```

```diff
 	const queryParams = await signParams(ctx, opts, {
 		postLoginClearedForSession:
-			type === "consent" && opts.postLogin ? options?.sessionId : undefined,
+			(type === "consent" || type === "post_login") && opts.postLogin ? options?.sessionId : undefined,
 	});
```

With this, the page is shown exactly once per authorization flow, `oauth2Continue({ postLogin: true })` proceeds to consent, and `consentReferenceId` still validates the selected organization. We are running this as a patch in production and verified the full flow end to end (multi-org OTP sign-in → select page → continue → consent → code → token → refresh), including that consent-accept still works and forged markers are still rejected.

If you prefer keeping the marker consent-only, the docs example needs to change instead (e.g. `shouldRedirect` returning `session.activeOrganizationId == null` for org scopes), plus a note that `shouldRedirect` must become `false` after selection — but that silently binds returning users to whatever organization happens to be active, so the marker fix seems strictly better.

1.7.1 / 1.7.2 (`better-auth` and `@better-auth/oauth-provider`); code unchanged on `main` @ 3e43439.

### System info

```bash
runtime: Bun 1.3.14 (also reproduces under Node)
framework: Next.js 16 (App Router)
database: Postgres via Kysely adapter
plugins: oauthProvider, organization, emailOTP, jwt
```

`@better-auth/oauth-provider` (postLogin / `oauth2/continue`)

johanneskares:
Fixes #11093.

The documented organization selector returns `true` for users with several organizations, even after they select one. `/oauth2/continue` runs that callback again, so these users return to the selection page instead of reaching consent.

Pass `isContinue` to `postLogin.shouldRedirect`: it is `true` for post-login continuation and `false` for the initial authorization request. Update the example to check the active organization against the user's memberships on continuation. Each new authorization request still asks users with several organizations to select one. Existing callbacks keep their current behavior; affected apps can use the updated example to stop the loop.

The issue suggested issuing `ba_pl` on the selection redirect. This PR keeps that marker limited to consent. Issuing it before selection would break the existing test that prevents clients from claiming completion without passing the server's selection check. The new flag identifies the request stage; it does not bypass that check.

Add tests with two real organizations for selection → continuation → consent → authorization code, the saved consent reference, selection on a new grant, and rejection when no organization is selected. Include a patch changeset and correct the client method in the example.

Validation:

- The new success test fails on unchanged `main`: expected `/consent`, received `/select-organization`. It passes with this fix.
- `pnpm vitest packages/oauth-provider/src --run`: 943 tests passed across 42 files, including the existing selection security tests.
- `pnpm typecheck`: passed.
- Commit hooks: Biome, spelling, and documentation formatting passed.

Browser verification (September 5, 2026):

- Used Chromium 152 with separate browser sessions and local in-memory SQLite databases. The test app used real Better Auth HTTP endpoints and the OAuth Provider browser client. No API responses were mocked.
- Both builds used the same app callback from the updated example. The original plugin source came from `4d09d502`; the fixed source came from `36ee4d32`.
- Before: signed in through the UI, started authorization, selected the second of two organizations, and clicked continue twice. Both attempts returned to the selection page, although the selected organization was saved.
- After: the same steps reached consent. Accepting consent returned an authorization code with the expected state and saved consent for the selected organization.
- After: continuing without a selected organization stayed on the selection page. Starting a new authorization after success asked for an organization again.

This browser check covers a local reproduction through authorization-code issuance. It does not cover a production app or token exchange.

Browser screenshots (original captures from the test above):

<details>
<summary>Before: selection returns to the same page twice</summary>

Original code: Second Organization is active, but both continuation attempts return to selection

</details>

<details>
<summary>After: selection reaches consent</summary>

PR code: the browser reaches consent for Second Organization

</details>

<details>
<summary>After: authorization succeeds for the selected organization</summary>

PR code: authorization code received, state verified, and consent saved for Second Organization

</details>

---
## Summary by cubic
Fixes #11093: the post-login organization selector loops back to the selection page for multi-organization users even after they select one. `postLogin.shouldRedirect` now receives an `isContinue` flag (`true` on post-login continuation, `false` on the initial authorization) so callbacks can validate the selected organization without repeatedly showing the selector.

- Updated the multi-organization example to check the active organization on continuation; existing callbacks keep current behavior.
- `isContinue` does not bypass the server-side selection check; each new authorization still asks multi-organization users to select.
- Added tests for selection → continuation → consent → authorization code, saved consent, re-selection on a new grant, and rejection when no organization is selected, plus a patch changeset.

<sup>Written for commit a7005b380ac7d4a0c52a93754fac5e8002e0b21f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11185?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


johanneskares:
bump, this is urgent, as it prevents me from upgrading. Anything that can be done from my side to help get it merged?

PR #11217, opened 2026-09-08
feat(test-utils): accept additional session fields in auth helpers

onmax:
`testUtils` auth helpers cannot populate required session fields without defaults, so session creation fails with a database constraint error. This adds optional per-session values to `login`, `getAuthHeaders`, and `getCookies`:

```diff
 const { headers, cookies } = await test.login({
   userId: user.id,
+  session: { providerToken: "test-token" },
 })
```

Values are supplied before persistence and override field defaults. Better Auth continues to generate standard session fields such as the user ID and token.

[Repro](https://github.com/onmax/repros/tree/repro/better-auth-session-fixtures/better-auth-session-fixtures): 1.7.3 ignores the proposed option and fails with `NOT NULL constraint failed: session.providerToken`. The [patched fixture](https://github.com/onmax/repros/tree/repro/better-auth-session-fixtures/better-auth-session-fixtures-fix) runs the same verifier and successfully reads the field through an authenticated session lookup.

### DIFF
--- a/.changeset/test-utils-session-fields.md
+++ b/.changeset/test-utils-session-fields.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Allow `testUtils` auth helpers to accept additional session fields through the `session` option, including required fields without defaults and per-session overrides of configured defaults.
--- a/docs/content/docs/plugins/test-utils.mdx
+++ b/docs/content/docs/plugins/test-utils.mdx
@@ -156,6 +156,17 @@ const member = await test.addMember({
 
 Auth helpers create authenticated sessions for testing protected routes.
 
+All three helpers accept an optional `session` object with additional session fields. Use it to supply required fields without defaults or to override configured defaults for one session:
+
+```ts
+const { headers } = await test.login({
+    userId: user.id,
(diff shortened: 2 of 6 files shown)


bytaesu:
I think we may need a better long-term design than adding fields like `userId` and `session` one by one.

That said, since this PR is scoped to the current issue, I'm fine with keeping it as-is for now.

bytaesu (approved):
@onmax Thanks for working on this. LGTM!

onmax:
> I think we may need a better long-term design than adding fields like `userId` and `session` one by one.
> 
> That said, since this PR is scoped to the current issue, I'm fine with keeping it as-is for now.

I agree, but for my case this is what I need. Redesigning this requires more discussion, which I am open to :)

I might open more test-utils related issues in future

bytaesu:
> > I think we may need a better long-term design than adding fields like `userId` and `session` one by one.
> > That said, since this PR is scoped to the current issue, I'm fine with keeping it as-is for now.
> 
> I agree, but for my case this is what I need. Redesigning this requires more discussion, which I am open to :)
> 
> I might open more test-utils related issues in future

Sounds good 😁

PR #11219, opened 2026-09-08
feat(react)!: suspend initial client queries

timneutkens:
React session and plugin query hooks currently expose their initial pending state through `isPending`. This change makes React 19 hooks suspend on a shared, stable query promise so the closest Suspense boundary supplies the loading UI. Other framework clients keep `isPending`.

During SSR, query hooks leave the Suspense fallback in the HTML before reading or mounting the client atom. When available, React DOM's `browser()` API is passed to `use()` to resume rendering in the browser without reporting a recoverable error. Since this API is currently Canary/Experimental-only, stable React 19 uses a server-error bailout and reports it through React's error callbacks. Ordinary plugin atoms remain server-renderable.

The session mount timer also avoids repeating an initial Suspense request that already completed, including a failed request, while later mounts continue to revalidate. Documentation includes the migration and an SSR example that renders the server-provided session in the Suspense fallback.

## Validation

- 71 focused tests passed on React 19.2.7: React Suspense, SSR, hydration, shared queries, and client behavior.
- 10 React tests passed against the actual React/React DOM `19.3.0-canary-f1f7ed2a-20260904` packages, installed in an isolated temporary directory. They cover browser bailout reporting, SSR without query reads or fetches, hydration with one fetch and no recoverable error, pending UI, and error results.
- The SSR fallback and duplicate failed-request regressions were reproduced with failing tests before their fixes.
- Focused runs used a temporary Vitest configuration to resolve workspace source exports and consistent framework runtimes; repository dependencies and the lockfile are unchanged.
- `pnpm typecheck` and focused code/documentation checks passed.

## Breaking change

React query hooks require React 19 and no longer return `isPending`. Wrap their consumers in Suspense boundaries. Server Components should await `auth.api.getSession` with incoming request headers.

---
## Summary by cubic
Switches the React client's initial session and plugin query reads to React Suspense, so pending UI is handled by Suspense boundaries instead of `isPending`. React query hooks now render only in the browser, leaving the nearest Suspense fallback in server-rendered HTML.

- React hooks require React 19 and suspend while the initial request is in flight; other framework clients keep `isPending`.
- `isPending` is removed from React hook results.
- On stable React 19, query hooks bail out of server rendering with an error reported to React's `onError`/`onRecoverableError`; React DOM Canary's `browser()` path uses `onBrowserBailout` instead.

**Migration**
- Wrap components that call `useSession` or client plugin query hooks in a `Suspense` boundary and use its fallback for loading UI.
- In Server Components, use `auth.api.getSession` instead of React client hooks.

<sup>Written for commit ab0fcf8a6a68889cc0ab155074a9b58c3acf2896. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11219?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


docs/content/docs/concepts/client.mdx
```diff
@@ -203,6 +211,11 @@ In addition to the standard methods, the client provides hooks to easily access
   </Tab>
 </Tabs>
 
+<Callout type="info">
+  Suspense-enabled React hooks require React 19 or newer. In a Server Component,
+  await `auth.api.getSession` instead of using the React client hook.
```
timneutkens, line 216: This should link the relevant API docs.

docs/content/docs/concepts/client.mdx
```diff
@@ -407,14 +420,26 @@ await authClient.signIn.email({
...
-```ts
+```tsx
+import { Suspense } from "react"
 import { useSession } from "@/lib/auth-client"
 
-const { data, error, isPending } = useSession()
-if (error) {
-    //handle error
+function User() {
+    const { data, error } = useSession()
```
timneutkens, line 434: Errors should probably throw and trigger an error boundary.

PR #11246, opened 2026-09-10
feat(organization): email confirmation for deletion & ownership transfer, plus explicit confirmationMode

### LINKED ISSUE #10748: feat: email confirmation for org deletion & ownership transfer, plus opt-in confirmationMode

The `user.deleteUser` flow supports an optional email confirmation step
(`sendDeleteAccountVerification`): the user requests deletion, gets an email
with a confirmation link, and the account is only deleted after they click
it. This protects against accidental or malicious one-click account deletion
(e.g. XSRF-ish misuse, a compromised session, a stray API call).

The organization plugin has no equivalent for its two most destructive
member-facing actions:

- `organization/delete` deletes the organization immediately once the caller
  passes the `organization:delete` permission check — no confirmation step.
- There is no "transfer ownership" operation at all. Reassigning the
  `creatorRole` (owner) today means calling `organization/update-member-role`
  directly, which takes effect immediately with no verification that the
  current owner actually intended to give up control.

Both are effectively irreversible from the acting member's perspective (data
loss for deletion, loss of control for ownership transfer), so they deserve
the same opt-in confirmation pattern `deleteUser` already has.

Mirror the existing `user.deleteUser` contract for these two organization
actions, as two independent, opt-in features under `OrganizationOptions`:

**1. Organization deletion confirmation**

```ts
organization({
  organizationDeletion: {
    sendDeleteOrganizationVerification: async ({ organization, user, url, token }, request) => {
      // send email
    },
    deleteTokenExpiresIn: 60 * 60 * 24, // default 1 day
  },
})
```

- `organization/delete` gains an optional `token` field in its body, same as
  `deleteUser`. Without `sendDeleteOrganizationVerification` configured,
  behavior is unchanged (immediate deletion) — fully backward compatible.
- With it configured, the first call creates a single-use verification token
  scoped to `(organizationId, requesting user)`, emails a callback URL, and
  returns `{ success: true, message: "Verification email sent" }` instead of
  deleting.
- New `GET /organization/delete/callback?token=...` endpoint (mirroring
  `/delete-user/callback`) re-validates the session and permission, consumes
  the token atomically, then runs the existing deletion path (including
  `beforeDeleteOrganization`/`afterDeleteOrganization` hooks).

**2. Ownership transfer confirmation**

A new dedicated endpoint rather than overloading `update-member-role` (which
is a generic, synchronous, multi-role endpoint and shouldn't grow an async
confirmation branch):

```ts
organization({
  ownershipTransfer: {
    sendTransferOwnershipVerification: async ({ organization, currentOwner, newOwner, url, token }, request) => {
      // send email
    },
    transferTokenExpiresIn: 60 * 60 * 24,
  },
})
```

- `POST /organization/transfer-ownership` — body: `organizationId`,
  `newOwnerMemberId`. Callable only by a current owner (or someone with
  `organization:update` on the owner role, consistent with today's
  last-owner rules in `update-member-role`).
- Without `sendTransferOwnershipVerification` configured: performs the same
  atomic demote-current/promote-target role swap that calling
  `update-member-role` with the creator role does today.
- With it configured: emails the *current owner* (not the target — they
  didn't initiate anything) a confirmation link; only on click does the
  atomic role swap happen. Token is single-use, scoped to
  `(organizationId, current owner, target member)` so it can't be replayed
  against a different target after org membership changes.
- Reuses the existing "can't leave an org with zero owners" guard from
  `update-member-role`/`remove-member`/`leave-organization`.

Both features:
- Are fully opt-in (no config change = current behavior, no breaking change).
- Get `beforeDelete`/`afterDelete`-style hooks where useful, consistent with
  existing `organizationHooks`.
- Ship with unit tests under `packages/better-auth/src/plugins/organization/routes/*.test.ts`
  and docs updates in `docs/content/docs/plugins/organization.mdx`.

- **Overloading `update-member-role` with a confirmation branch for the
  creator role** — rejected: that endpoint handles arbitrary multi-role
  updates for any member, not just ownership; bolting an async
  email-confirmation branch onto it changes its contract in a confusing,
  role-specific way and complicates testing/typing for a case that's a small
  minority of its calls.
- **A single generic "confirm sensitive org action" token type reused across
  delete/transfer** — rejected for the initial version: the two actions have
  different payloads (org vs. two members) and different recipients (acting
  member vs. current owner); a shared abstraction can be extracted later if a
  third use case shows up, but building it now would be speculative.

### Additional context

Existing prior art in this codebase to follow for both the API shape and the
security details (atomic single-use token consumption via
`consumeVerificationValue`, session-freshness bypass on the callback via
`disableCookieCache`, `originCheck` middleware on the callback URL):

- `packages/better-auth/src/api/routes/update-user.ts` (`deleteUser`,
  `deleteUserCallback`)
- Options contract: `packages/core/src/types/init-options.ts` (`user.deleteUser`)

Related existing organization primitives this feature builds on:
- `packages/better-auth/src/plugins/organization/routes/crud-org.ts`
  (`deleteOrganization`, `beforeDeleteOrganization`/`afterDeleteOrganization`
  hooks)
- `packages/better-auth/src/plugins/organization/routes/crud-members.ts`
  (`updateMemberRole`, existing last-owner protection logic)

mcorbelli:
Closes #10748

## Context

This issue proposed two independent, opt-in additions to bring the
organization plugin's most destructive actions (deletion, ownership
transfer) in line with the confirmation pattern `user.deleteUser` already
has, plus a follow-up `confirmationMode` generalizing the "the emailed link
is itself destructive" concern across all four confirmation surfaces
(`deleteUser`, `changeEmail`, and the two new organization ones).

**Note on process:** at the time of opening this PR, the issue is still
untriaged on GitHub (no `target:*` label, no maintainer response on the
thread yet). I received explicit go-ahead to open this PR from **@max**
(ping-maxwell) on Discord. Happy to adjust scope, split this up, or close it
if the direction doesn't match what the team has in mind — opening this was
meant to give something concrete to react to rather than pre-empt the
triage/discussion CONTRIBUTING.md asks for.

## What changed

**Organization deletion confirmation**
- `organizationDeletion.sendDeleteOrganizationVerification` gates
  `organization.delete` behind a confirmation email. Token scoped to
  `(organizationId, requesting user)`, single-use, expiry via
  `deleteTokenExpiresIn` (default 1 day).
- New `GET /organization/delete/callback`: re-validates session and the
  `organization:delete` permission (both may have changed since the email
  was sent), consumes the token, then runs the existing deletion path
  (including `beforeDeleteOrganization`/`afterDeleteOrganization`).

**Ownership transfer**
- New `POST /organization/transfer-ownership`, deliberately not folded into
  `update-member-role` (which stays generic and synchronous). Same
  authorization shape as `update-member-role`: current owner, or a member
  with `member:update` permission.
- New `adapter.transferOwnership`, wrapped in a transaction, so a crash
  between promoting the target and demoting the current owner can't leave
  the organization with zero owners or two.
- `ownershipTransfer.sendTransferOwnershipVerification` gates it behind a
  confirmation email sent to the **current owner** (they're the one giving
  up control, not the target). Token scoped to `(organizationId, current
  owner, target member)`.
- New `beforeTransferOwnership`/`afterTransferOwnership` hooks.

**`confirmationMode: "instant" | "explicit"`** (default `"instant"`), on all
four surfaces (`user.deleteUser`, `user.changeEmail`, `organizationDeletion`,
`ownershipTransfer`)
- Today, the emailed link for all of these is a plain `GET` that applies the
  change on visit. Some email clients and security scanners open links
  automatically to scan them (Outlook Safe Links, link-preview proxies),
  which can trigger the change without the user actually clicking anything.
- In `"explicit"` mode, the emailed `url` points at the app's own callback
  URL with the token attached instead of better-auth's endpoint, and a
  `.../preview` (GET, no mutation) + `.../confirm` (POST, applies it) pair
  replaces the single destructive GET.
- Default stays `"instant"` everywhere, so this is fully backward compatible.

Both organization features and `confirmationMode` are opt-in: without
configuring the relevant callback, behavior is unchanged.

## Things worth flagging in review

- **`organization.delete`'s response type is now a union**
  (`Organization | { success, message }`), to accommodate the
  pending-verification acknowledgement — mirrors `deleteUser`'s existing
  contract, but it's a type change for `authClient.organization.delete`'s
  callers even when they don't use the new option.
- **`changeEmail`'s token isn't single-use.** It's a signed JWT, not a
  `verification` DB row, so in `"explicit"` mode a `/change-email/confirm`
  replay before expiry just re-applies the same (idempotent) email change.
  This is pre-existing behavior of the instant-mode link too, not something
  this PR introduces — documented explicitly rather than left implicit.
- **Naming collision with an existing option.** The organization plugin
  already had a top-level `disableOrganizationDeletion: boolean`. The docs
  incorrectly described it as living under an `organizationDeletion` key —
  which this PR now uses for the real, nested confirmation config. Fixed
  the doc text to distinguish the two; happy to rename either option if the
  team would rather avoid the near-collision.

## Testing

- New tests: `crud-org.test.ts` (delete confirmation, instant + explicit +
  permission-revoked-after-send), `transfer-ownership.test.ts` (immediate
  swap, instant + explicit confirmation, self-transfer / already-owner /
  outside-member rejections), `update-user.test.ts` (`confirmationMode`
  explicit mode for both `deleteUser` and `changeEmail`, both
  token-generation sites for the latter).
- `pnpm typecheck` passes.
- Full existing suites re-run clean: `packages/better-auth/src/api` (435
  tests) and `packages/better-auth/src/plugins/organization` (259 tests +
  the new ones above).

## Docs

Updated `docs/content/docs/plugins/organization.mdx` (delete confirmation,
new Transfer Ownership section, Options list) and
`docs/content/docs/concepts/users-accounts.mdx` (`confirmationMode` for
`deleteUser`/`changeEmail`), plus the coarse `reference/options.mdx` sample.

https://claude.ai/code/session_01U3d98r2fYSginqYX37zbhh

---
## Summary by cubic
Adds opt-in email confirmation for organization deletion and a new ownership-transfer endpoint, plus a `confirmationMode: "explicit"` option for `deleteUser`, `changeEmail`, and both new organization actions so emailed links preview instead of apply, preventing auto-opening mail clients and security scanners from triggering destructive changes. Default stays `"instant"` and all changes are backward compatible.

**Breaking Changes**
- `organization.delete` now returns `Organization | { success, message }` instead of just `Organization` when a verification email is configured.

**Notes**
- Removes the unused optional `better-sqlite3` peer dependency to prevent installation conflicts.
- `changeEmail` tokens are signed JWTs, not single-use; a replay before expiry just re-applies the same idempotent change (pre-existing).
- The new nested `organizationDeletion` option is named close to the existing top-level `disableOrganizationDeletion`; docs now distinguish both.

<sup>Written for commit ad90be49a3543d8f3304c68403dd4d3a54773f10. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11246?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mcorbelli:
Quick status update after pushing the fixes:

**Greptile's three findings from the first review are all addressed:**
- "Admins Can Create Owners" → 6c65d4c3f (`fix(organization)!: require the actual owner for ownership transfer`)
- "Unauthorized Requests Burn Tokens" (organization delete) → 6467ec0e3 (`fix(organization): validate before consuming the organization-delete token`)
- "Concurrent Transfers Create Owners" → f617d7391 (`fix(organization): guard the ownership demotion against concurrent transfers`)

The same token-burn-before-auth pattern also existed on the ownership-transfer resolver (not called out by name in the original review, but the same class of bug) — fixed alongside 6c65d4c3f. Each fix has a dedicated regression test that fails without it.

**Separately, I don't think Greptile can currently re-review this PR at all**, unrelated to anything in this diff: this branch is based on `main` (as CONTRIBUTING.md instructs), and got auto-retargeted to `next` for the minor changeset. But `next`'s current tip is 132 commits behind `main` — including the v1.7.0, v1.7.1, and v1.7.4 release commits — so the `next...head` diff GitHub computes includes all of that unrelated backlog on top of this PR's actual ~15 files, which is presumably why the last review comment says "Too many files changed for review (683 files, 100 file limit)." Flagging in case a `main` → `next` sync is overdue independently of this PR.

mcorbelli:
For anyone reviewing manually: the PR's "Files changed" tab is comparing against `next`'s stale tip and pulls in ~670 unrelated files from the `main`→`next` backlog above. The actual diff this PR introduces is 18 files: https://github.com/better-auth/better-auth/compare/main...mcorbelli:better-auth:feat/org-action-confirmation

mcorbelli:
A second automated review (cubic) also ran, on 683 files. 9 of its 10 findings are in files this PR never touches (workflow files, sso/hono docs, mdx-components.tsx) — same `next`/`main` staleness noise as Greptile's file-count error above, verified each one against `git diff upstream/main`. The 10th is real and now fixed in ad90be49a: the changeset understated `organization.delete`'s return-type change as fully backward compatible when it's actually a TypeScript breaking change (return type is now a union) for every caller, not just those opting into the new option. Added an explicit Breaking change note to the changeset.

mcorbelli:
Superseded by #11250.

PR #11247, opened 2026-09-10
fix(one-tap): verify Google ID token nonce in callback

### LINKED ISSUE #10926: one-tap: /one-tap/callback never passes a nonce to verifyGoogleIdToken
### Reproduction

Code-level; no runtime repro needed.

1. Register `oneTapClient({ clientId, ... })` and call `authClient.oneTap({ nonce: 'abc123' })`. The client forwards the nonce to `google.accounts.id.initialize`, so Google mints an ID token containing `nonce: "abc123"`.
2. The client posts only `{ idToken, callbackURL }` to `/one-tap/callback`.
3. Send the same request again with any other `nonce` value, or none, or a token whose nonce claim is a value the server never issued. All are accepted identically.

`verifyGoogleIdToken` supports nonce validation. `packages/core/src/social-providers/google.ts`:

```ts
// :90-101
audience: string | string[];
nonce?: string | undefined;
...
export const verifyGoogleIdToken = async ({ token, audience, nonce }) => {
// :117
  if (nonce && jwtClaims.nonce !== nonce) {
```

The One-Tap endpoint is the one caller that never passes it. `packages/better-auth/src/plugins/one-tap/index.ts`:

```ts
// :119-121
const payload = (await verifyGoogleIdToken({
  token: idToken,
  audience,          // no `nonce`
```

Because the check is guarded by `if (nonce && ...)`, omitting the argument skips validation silently. And there is no way for a caller to supply one: the request schema (`:43-52`) declares only `idToken` and `callbackURL`, so a nonce sent in the body is stripped before the handler sees it.

Net effect: the client option makes nonce protection look supported, while the server never verifies it. A captured ID token stays replayable for its full lifetime (~1h for Google), with no `nonce` or `jti` check bounding it.

**Expected:** the endpoint accepts an expected nonce and forwards it to `verifyGoogleIdToken`, so the existing comparison runs.

This is the same finding as #6521 ("Security: Google provider should verify nonce in verifyIdToken", closed 2025-12-08), which added the validation to the shared verifier. That fix did not reach the One-Tap call site.

1.6.26. Also present on `main` (line numbers above are from `main`) and 1.7.0-rc.5. The One-Tap plugin has never read a nonce at any release.

### System info

Node 24, Next.js 15 (App Router), `better-auth` server + `better-auth/react` client, Google One-Tap via `oneTapClient`. Not environment-specific — it is a missing argument at the call site.

`plugins/one-tap`, `social-providers/google`

### Additional context

Smallest fix is to add an optional `nonce` to `oneTapCallbackBodySchema` and forward it to `verifyGoogleIdToken`. Note that for the check to be meaningful the *expected* value must be anchored server-side (server state, or a cookie signed with a server secret) — an unsigned cookie holding a hash of the nonce is forgeable, since the nonce is a plaintext claim in the token the attacker already holds.

Alternatively, Google ID tokens carry a `jti`, so caching seen `jti`s for the token lifetime would give One-Tap replay protection with no client changes and no anchoring problem.

waleed-mirza:
## Summary
- `/one-tap/callback` now accepts an optional `nonce` in the request body and forwards it to `verifyGoogleIdToken`, which compares it against the ID token's `nonce` claim
- `oneTapClient` now sends `opts.nonce` in the callback POST body (previously it only passed the nonce to `google.accounts.id.initialize`, so the server never verified it — a captured token stayed replayable for its lifetime)
- Documented the `nonce` option in the One Tap plugin docs

Closes #10926

#### Test plan
- [x] `pnpm vitest packages/better-auth/src/plugins/one-tap --run` — 27 tests pass, including 4 new nonce cases (matching nonce accepted; mismatched/missing claim rejected with 400; no nonce still accepted for backward compatibility)
- [x] `pnpm typecheck` passes
- [x] biome/format/spell hooks pass
- [x] Changeset added (`patch`)

Generated with [Devin](https://devin.ai)

---
## Summary by cubic
Verifies the Google One Tap nonce on the server so a captured ID token can no longer be replayed for its lifetime. Closes #10926.

- `/one-tap/callback` now accepts an optional `nonce` in the request body and forwards it to `verifyGoogleIdToken`, which checks it against the token's `nonce` claim.
- `oneTapClient` now sends `opts.nonce` in the callback POST body; previously it only passed the nonce to `google.accounts.id.initialize`, so the server never verified it.
- Tokens with a mismatched or missing nonce claim are rejected with 400 when a nonce is provided; requests without a nonce still work for backward compatibility.
- Documents the `nonce` option in the One Tap plugin docs.

<sup>Written for commit 5395b4aed734392f9f039f345a9c71255ad0245d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11247?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/better-auth/src/plugins/one-tap/client.ts
+++ b/packages/better-auth/src/plugins/one-tap/client.ts
@@ -259,7 +259,11 @@ export const oneTapClient = (options: GoogleOneTapOptions) => {
 						async function callback(idToken: string) {
 							const res = await $fetch("/one-tap/callback", {
 								method: "POST",
-								body: { idToken, callbackURL: opts?.callbackURL },
+								body: {
+									idToken,
+									callbackURL: opts?.callbackURL,
+									nonce: opts?.nonce,
+								},
 								...opts?.fetchOptions,
 								...fetchOptions,
 							});
@@ -314,7 +318,11 @@ export const oneTapClient = (options: GoogleOneTapOptions) => {
 					async function callback(idToken: string) {
 						const res = await $fetch("/one-tap/callback", {
 							method: "POST",
-							body: { idToken, callbackURL: opts?.callbackURL },
+							body: {
+								idToken,
+								callbackURL: opts?.callbackURL,
+								nonce: opts?.nonce,
+							},
 							...opts?.fetchOptions,
 							...fetchOptions,
 						});
--- a/packages/better-auth/src/plugins/one-tap/index.ts
+++ b/packages/better-auth/src/plugins/one-tap/index.ts
@@ -55,6 +55,19 @@ const oneTapCallbackBodySchema = z.object({
 			description: "URL to redirect to after a successful sign-in",
 		})
 		.optional(),
+	/**
+	 * The nonce the client passed to `google.accounts.id.initialize`. Google
+	 * embeds it in the ID token's `nonce` claim; forwarding it lets
+	 * `verifyGoogleIdToken` confirm the token was minted for this sign-in
+	 * attempt instead of replayed.
+	 */
+	nonce: z
+		.string()
+		.meta({
+			description:
+				"Nonce passed to the Google One Tap API, verified against the ID token's nonce claim",
+		})
+		.optional(),
 });
 
 export const oneTap = (options?: OneTapOptions | undefined) =>
@@ -99,7 +112,7 @@ export const oneTap = (options?: OneTapOptions | undefined) =>
 					},
 				},
 				async (ctx) => {
-					const { idToken } = ctx.body;
+					const { idToken, nonce } = ctx.body;
 					const googleProvider =
 						typeof ctx.context.options.socialProviders?.google === "function"
 							? await ctx.context.options.socialProviders?.google()
@@ -119,6 +132,7 @@ export const oneTap = (options?: OneTapOptions | undefined) =>
 					const payload = (await verifyGoogleIdToken({
 						token: idToken,
 						audience,
+						nonce,
 					})) as Partial<GoogleProfile> | null;
 					if (!payload) {
 						throw new APIError("BAD_REQUEST", {
--- a/.changeset/tame-donuts-repeat.md
+++ b/.changeset/tame-donuts-repeat.md
@@ -1,0 +1,5 @@
+---
+"better-auth": patch
+---
+
+Verify the Google One Tap `nonce` option on the server. The `/one-tap/callback` endpoint now accepts an optional `nonce` and forwards it to `verifyGoogleIdToken`, so ID tokens minted with a nonce are bound to the sign-in attempt that requested them instead of remaining replayable.
--- a/docs/content/docs/plugins/one-tap.mdx
+++ b/docs/content/docs/plugins/one-tap.mdx
@@ -158,6 +158,18 @@ await authClient.oneTap({
 });
 ```
 
+#### Binding the ID Token with a Nonce
+
+Pass a unique `nonce` per sign-in attempt. It is forwarded to the Google One Tap API (which embeds it in the ID token) and then verified server-side against the token's `nonce` claim, so a captured token cannot be replayed without it:
+
+```ts
+import { authClient } from "@/lib/auth-client";
+
+await authClient.oneTap({
+  nonce: crypto.randomUUID()
+});
+```
+
 #### Handling Prompt Dismissals with Exponential Backoff
 
 If the user dismisses or skips the prompt, the plugin will retry showing the One Tap prompt using exponential backoff based on your configured promptOptions.
--- a/packages/better-auth/src/plugins/one-tap/one-tap.test.ts
+++ b/packages/better-auth/src/plugins/one-tap/one-tap.test.ts
@@ -820,3 +820,97 @@ describe("oneTapClient types", () => {
 		expectTypeOf(client.oneTap).toBeFunction();
 	});
 });
+
+/**
+ * @see https://github.com/better-auth/better-auth/issues/10926
+ */
+describe("one-tap nonce verification", async () => {
+	afterEach(() => {
+		Object.assign(verifiedPayload, defaultVerifiedPayload);
+		(verifiedPayload as Record<string, unknown>).nonce = undefined;
+	});
+
+	const googleProvider = {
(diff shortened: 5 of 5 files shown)


waleed-mirza:
Good point from the AI reviewers — worth addressing explicitly.

They're correct that a client-supplied `nonce` only verifies self-consistency (token claim == body value), which an attacker holding the token can trivially satisfy, and omitting the field skips the check. Full replay protection requires a **server-issued nonce**: server generates and stores it (cookie or verification row), client passes it to Google, server compares on callback and consumes it — the same pattern used in the OIDC discovery-provider flow.

This PR implements the scope accepted in #10926 ("the endpoint accepts an expected nonce and forwards it to `verifyGoogleIdToken`", labeled `target: patch`). It closes the gap where the client option suggested nonce protection existed while the server never verified it, and it keeps the field optional for backward compatibility.

Happy to take this further if you'd prefer — the server-bound design would need a nonce-issuing endpoint (or piggyback on an existing mechanism) and one-time consumption, which feels like a larger change than patch scope. Let me know whether to expand this PR or land this and follow up separately.

packages/better-auth/src/plugins/one-tap/index.ts
```diff
@@ -55,6 +55,19 @@ const oneTapCallbackBodySchema = z.object({
 			description: "URL to redirect to after a successful sign-in",
 		})
 		.optional(),
+	/**
```
sunejunior, line 58: nit: since this field exists specifically to prevent replay attacks, making it .optional() means callers can skip verification entirely by omitting it. Was this intentional for backward compat, or should it be required going forward (maybe with a deprecation warning if omitted) so the fix can't be silently bypassed?

sunejunior (commented):
Tested this locally, works as expected, nonce verification against the ID token's nonce claim correctly rejects mismatched/replayed tokens. One nit: nonce is optional, which seems to weaken the security guarantee this PR is adding. Worth clarifying if that's intentional (e.g. migration period) or should be required.

packages/better-auth/src/plugins/one-tap/index.ts
waleed-mirza: Agreed -- leaving it optional would have let callers skip the check.

That is no longer the case. `/one-tap/callback` no longer reads a body `nonce`. The expected value comes from a signed HttpOnly cookie plus a single-use verification record; missing/expired/consumed state is a 400 before token verification.

packages/better-auth/src/plugins/one-tap/client.ts
```diff
@@ -239,9 +243,38 @@ export const oneTapClient = (options: GoogleOneTapOptions) => {
...
+						return nonce;
+					};
+
 					// Button mode: render a button instead of showing the prompt
 					if (opts?.button) {
-						await loadGoogleScript();
+						let nonce: string;
+						try {
+							nonce = await getServerNonce();
+							await loadGoogleScript();
```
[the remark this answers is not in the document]
waleed-mirza: Fixed in c71ecb8. Button mode now reissues a server nonce after each credential callback and shortly before TTL expiry, then reinitializes the rendered button with the new nonce.

waleed-mirza:
@sunejunior thanks for testing this — your point about the optional nonce is addressed, and the design changed substantially since your review.

The nonce is no longer client-supplied or optional. `/one-tap/callback` does not read a `nonce` from the body at all. Each attempt is server-issued via `/one-tap/nonce`, which stores a single-use verification record and binds it to the browser through a signed HttpOnly cookie. A callback with no cookie, or with an expired or already-consumed record, is a 400 before the token is verified — so there is no path that skips the check.

Two follow-ups since then, from the AI reviewers:

- A browser previously held exactly one outstanding attempt, so any new `/one-tap/nonce` call invalidated a nonce Google may have already minted a credential against. That rejected legitimate sign-ins when a scheduled nonce refresh landed between a click and its callback, or when a page rendered more than one button. The cookie now carries a bounded set of outstanding states and the callback matches the token's verified nonce claim against that set, consuming only the record it matched. Replay protection is unchanged: the cookie is signed and HttpOnly so the candidate set is not attacker-controlled, and each record stays single-use.
- A failed button nonce refresh now retries with backoff instead of stranding the button on a spent nonce.

Covered by tests for cross-attempt acceptance, replay of a consumed attempt while a sibling is outstanding, concurrent consumption, and the client retry path. Happy to take another look if anything still reads as weak.

waleed-mirza:
Addressed the findings from the latest pass in `9127a778d` — all three were valid.

**Concurrent issuance loses nonces** (Greptile P1). Correct, and my client-side dedupe did not cover it: that guard is per-container, while two buttons are two independent `oneTap()` calls. Issuance appends to the set carried by the cookie, so it is a read-modify-write over browser state — both requests read the pre-request cookie and each write its own successor, and the later `Set-Cookie` wins. `/one-tap/nonce` requests are now serialized per document through a module-level chain, so each request is sent with the previous response's cookie already applied. Separate tabs still race; that is inherent to cookie state and is noted in the code.

**Nonce limit invalidates buttons** (Greptile P1 / cubic P2). Also correct — `.slice(-3)` evicted the oldest state unconditionally, so a fourth rendered button silently invalidated the first. The cap was meant as a bound on cookie growth, not a button limit, but it was doing both. Spent and expired attempts are now pruned on each issuance, and the cap is raised to 10 as a backstop. Because pruning runs first, the cap only binds if a page really has that many buttons with simultaneously live nonces.

**Missing render assertions** (cubic P3). Added `renderButton` / `replaceChildren` counts to the retry test, matching its sibling tests.

Coverage for the first two: a client test asserting nonce requests never overlap and that three concurrent buttons each get a distinct nonce, and a server test that issues four attempts and verifies the oldest still completes. Both fail against the previous commit — I checked by reverting each fix in isolation. 36 tests pass, `tsc --noEmit` clean for one-tap.

PR #11252, opened 2026-09-10
feat: add Link as a social provider

stevekaliski-stripe:
Hello from Stripe/Link! Adding support for Link as a social provider in order to support for integrations like https://github.com/stripe/link-cli.


packages/core/src/social-providers/link.ts
```diff
@@ -0,0 +1,201 @@
...
+				body,
+			});
+			if (error) throw error;
+		},
+		async getUserInfo(token) {
+			if (options.getUserInfo) {
+				return options.getUserInfo(token);
+			}
+
+			const { data: profile, error } = await betterFetch<LinkProfile>(
```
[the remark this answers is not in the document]
stevekaliski-stripe: In comparable implementations I don't see this being used? Am I missing something?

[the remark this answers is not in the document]
bytaesu: Yes, this is consistent.

packages/core/src/social-providers/link.ts
```diff
@@ -0,0 +1,201 @@
...
+export const link = (options: LinkOptions) => {
+	return {
+		id: "link",
+		name: "Link",
+		accountSubject: ({ profile }) => {
+			if (!profile.email) {
+				throw new BetterAuthError("Link profile email is required");
+			}
+			return profile.email;
+		},
```
bytaesu, line 79: Does Link guarantee that email can be treated as an immutable account ID? As far as I know, Link allows users to change their email, so this feels a bit unstable.

bytaesu (commented):
Hi @stevekaliski-stripe, interesting 👀

Would it be helpful to briefly document (or blog post) the intended user flow, including how the resulting access token is used with Link CLI? Unlike a typical social provider such as Google (Sign in with Google), Link OAuth seems authorize a hosted agent to access a user's Link account, so it isn't immediately clear whether `signIn.social()` is intended as a sign-in flow or an account-connection flow.

docs/content/docs/authentication/link.mdx
```diff
@@ -0,0 +1,99 @@
...
+## Usage
+
+### Connect a Link account
+
+Start the Link authorization flow with `signIn.social`:
+
+```ts title="auth-client.ts"
+import { createAuthClient } from "better-auth/client"
+
+const authClient = createAuthClient()
```
bytaesu, line 56: We use this pattern:

```ts
import { authClient } from "@/lib/auth-client"
```

packages/core/src/social-providers/link.ts
```diff
@@ -0,0 +1,201 @@
...
+const userinfoEndpoint = "https://api.link.com/userinfo";
+
+/**
+ * The user profile returned by Link's `/userinfo` endpoint.
+ */
+export interface LinkProfile {
+	/**
+	 * The email address associated with the Link account.
+	 */
+	email?: string;
```
bytaesu, line 28: https://github.com/stripe/link-cli/blob/main/packages/sdk/src/types/index.ts#L208

`email` seems nullish

docs/content/docs/authentication/link.mdx
stevekaliski-stripe: ack!

stevekaliski-stripe:
Hey @bytaesu! Thank you for the feedback! The use case is that developers building on top of the Link Agent Wallet (see Instinct, Grok, Muse, etc) access those capabilities to get Link user info and generate payment credentials via OAuth and interact via API (such as using the SDK https://github.com/stripe/link-cli/tree/main/packages/sdk).

One can imagine a progressive where the user, interacting with chat/assistant application, needs to authenticate with different services to unlock new capabilities, such as Link for payment.

bytaesu:
> Hey @bytaesu! Thank you for the feedback! The use case is that developers building on top of the Link Agent Wallet (see Instinct, Grok, Muse, etc) access those capabilities to get Link user info and generate payment credentials via OAuth and interact via API (such as using the SDK https://github.com/stripe/link-cli/tree/main/packages/sdk).
> 
> One can imagine a progressive where the user, interacting with chat/assistant application, needs to authenticate with different services to unlock new capabilities, such as Link for payment.

Hmm... would a plugin be cleaner here? I'm asking because this doesn't seem to fit the `signIn.social()` flow very well.

like:

```ts
betterAuth({
  plugins: [
    linkAgentWallet({
      clientId,
      clientSecret,
      publishableKey,
    }),
  ],
})
```

```ts
await authClient.link.connect()
```

docs/content/docs/authentication/link.mdx
```diff
@@ -0,0 +1,99 @@
...
+
```
sunejunior, line 57: publishableKey uses STRIPE_PUBLISHABLE_KEY while clientId/clientSecret use LINK_CLIENT_ID/LINK_CLIENT_SECRET, worth using a consistent LINK_ prefix across all three, or explaining why publishableKey is the odd one out (e.g. it's a Stripe-wide key, not Link-specific)

sunejunior (commented):
Tested this locally, works as expected. One nit on naming, env var prefix is inconsistent between LINK_* and STRIPE_* for what's otherwise the same provider config block.

stevekaliski-stripe:
Closing! We're opting for a plugin model instead, living in https://github.com/stripe/link-cli. We'll re-evaluate in the future once we have a stable identifier.

PR #11254, opened 2026-09-11
refactor(cookies): centralize cookie cache version validation

Emmaccen:
getCookieCache repeated the same version check across the JWE, JWT, and compact strategies.

Centralize the comparison in a shared helper while preserving each strategy's validation order and the fast path when versioning is not configured. This is a behavior-preserving refactor.

---
## Summary by cubic
Extracts the repeated cookie cache version validation into a shared `matchesVersion` helper so JWE, JWT, and compact strategies use one implementation. No behavior change.

- Adds a `CookieCacheVersion` type for the string-or-function version config.
- Hoists the `payload` null check so each strategy returns early before validation.

<sup>Written for commit c110a6df3824c46ff3915c468501684b2204f2d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11254?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/better-auth/src/cookies/index.ts
+++ b/packages/better-auth/src/cookies/index.ts
@@ -666,6 +666,18 @@ export const getCookieCache = async <
 	}
 
 	if (sessionData) {
+		const validateVersion = async (payload: CookieCachePayload) => {
+			if (!config?.version) return true;
+			const cookieVersion = payload.version || "1";
+			let expectedVersion = "1";
+			if (typeof config.version === "string") {
+				expectedVersion = config.version;
+			} else if (typeof config.version === "function") {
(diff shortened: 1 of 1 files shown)


bytaesu:
Thanks @Emmaccen,

I refined it a bit more based on your work, also using the shared type helper.

bytaesu (approved):

PR #11281, opened 2026-09-14
▚▚ fix(organization): refresh session cookie cache after accepting an invitation

### LINKED ISSUE #11275: organization.acceptInvitation never refreshes the session cookie after the active-organization write, leaving the cookie cache stale
In the organization plugin's `acceptInvitation` endpoint (`dist/plugins/organization/routes/crud-invites.mjs`; verified in `1.7.2` and byte-identical in shape in `1.7.4` at ~L318/L330), the endpoint updates the accepting user's "active" pointers in two places, and in neither ordering does the signed cookie end up carrying the post-accept `activeOrganizationId`:

1. The active-team branch (single-team invitations):
   ```js
   await setSessionCookie(ctx, {
   	session: await adapter.setActiveTeam(session.session.token, teamId, ctx),
   	user: session.user,
   });
   ```
2. A few lines later, the active-organization write:
   ```js
   await adapter.setActiveOrganization(session.session.token, acceptedI.organizationId, ctx);
   ```
   — bare, with no corresponding `setSessionCookie`. The team branch's `setSessionCookie` runs BEFORE this write, so even a single-team accept signs a cookie still carrying the pre-accept `activeOrganizationId`.

`adapter.setActiveOrganization` updates the session row in the database, but the signed cookie-cache payload is never refreshed after that write.

**Precondition:** `session.cookieCache.enabled: true` (not the default; fallback maxAge 300s). Any caller reading the session via the cookie-cache-permitting path (`GET /get-session` without `disableCookieCache=true`) then sees the PRE-accept `activeOrganizationId` for up to the cache TTL after a successful accept, even though the DB is already updated.

**Repro:** configure `session.cookieCache.enabled: true`; sign in as a user with a pending invitation (sign-in primes the cookie cache); `POST /organization/accept-invitation`; immediately `GET /get-session` without `disableCookieCache`. Expected: `session.activeOrganizationId` = the accepted org. Actual: the pre-accept value (often `null`) until the cache expires.

**Suggested fix (one line):** `adapter.setActiveOrganization` already returns the updated session in exactly the shape `adapter.setActiveTeam` does, so the organization write can be wrapped the same way the team branch already is:

```js
await setSessionCookie(ctx, {
	session: await adapter.setActiveOrganization(
		session.session.token,
		acceptedI.organizationId,
		ctx,
	),
	user: session.user,
});
```

placed after the last active-pointer write so the signed cookie carries the final state on both branches.

We work around it with a `hooks.after` on `/organization/accept-invitation` that re-signs the cookie from the authoritative session, so this isn't blocking us — flagging since the team branch already shows the intended shape. Checked unfixed as of `1.7.4` (current latest); the 1.7.3/1.7.4 changelogs contain no related entries.

breken-ai:
Closes #11275.

With `session.cookieCache.enabled: true`, `POST /organization/accept-invitation` updates the session row in the database but never re-signs the session cookie from that write. The `session_data` cookie keeps the pre-accept snapshot, so `GET /get-session` (without `disableCookieCache`) returns the old `activeOrganizationId` — usually `null` — until the cache TTL expires, even though the row already holds the accepted organization.

## Root cause

In `acceptInvitation`, `adapter.setActiveOrganization(...)` was called bare and its returned session discarded, so `setSessionCookie` never ran for the organization write. The single-team branch did call `setSessionCookie`, but from `setActiveTeam`'s result and before the organization write, so that cookie also carried the stale `activeOrganizationId`. The `/organization/set-active` endpoint already re-signs the cookie from the `setActiveOrganization` result; this endpoint did not.

## Fix

- `packages/better-auth/src/plugins/organization/routes/crud-invites.ts`: sign the session cookie once, after the last active-pointer write, from the session returned by `setActiveOrganization`. The team branch keeps its `setActiveTeam` call but drops its early cookie write, so a single `Set-Cookie` carries both `activeTeamId` and `activeOrganizationId`.
- `.changeset/accept-invitation-cookie-cache.md`: `better-auth` patch.

## Tests

- `crud-invites.test.ts`: new regression (`@see` #11275) enables the cookie cache, signs in the invitee capturing the full cookie set, accepts the invitation while applying the response cookies, then asserts the live session (`disableCookieCache: true`) and the cached session both report the accepted organization. Without the fix the cached read returns `null`.
- Organization suites (crud-invites, team, organization, crud-org): 198 passed. `pnpm typecheck` clean.

## Not changed

- `/organization/create` also calls `setActiveOrganization` / `setActiveTeam` without re-signing the cookie (crud-org.ts:274, 286). Same class of gap, but outside this issue; left for a separate change.

---
▚▚ **Shipped by breken** — self-healing software. This one's on us. [breken.ai](https://breken.ai)

---
## Summary by cubic
Fixes `acceptInvitation` leaving the session cookie cache stale when `session.cookieCache.enabled` is on. The endpoint wrote the accepted organization to the session row but never re-signed the session cookie, so cached `GET /get-session` reads kept returning the old `activeOrganizationId` until the cache TTL expired. The cookie is now signed once after the final active-pointer write, so both `activeTeamId` and `activeOrganizationId` are reflected immediately.

**Bug Fixes**
- Single-team invitations no longer write the cookie early; `setActiveTeam` still runs, but the session returned by `setActiveOrganization` produces one `Set-Cookie` with both values.
- Adds a regression test covering the cookie-cache flow and a `better-auth` patch changeset.
- Leaves `/organization/create` with the same gap for a separate change.

Closes #11275.

<sup>Written for commit 2044b5f4b5276acfae549a71f557ea85c6d7a182. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11281?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bytaesu:
Since there doesn't seem to be any communication around the PR title either, I'm going to close this PR.

PR #11300, opened 2026-09-15
fix(oauth): persist account cookie after implicit linking

bytaesu:
Related to https://github.com/better-auth/better-auth/pull/11298

---
## Summary by cubic
Fixes stateless OAuth sign-in failing with `ACCOUNT_NOT_FOUND` when retrieving an access token after a new provider is implicitly linked to an existing user. The linked account's cookie is now persisted when `account.storeAccountCookie` is enabled, so subsequent `getAccessToken` calls see the correct account.

- Sets the account cookie after implicit linking if `storeAccountCookie` is on, and returns an error if linking fails.
- Adds a test covering this scenario, including a second stateless instance that has no record of the user.

<sup>Written for commit d231ee13479a24a00d28ee8d79ca205fb4c2dfe9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11300?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bytaesu:
@greptileai review

PR #11302, opened 2026-09-16
fix(client): normalize network fetch failures

### LINKED ISSUE #11284: Client rejects with `TypeError: Failed to fetch` on network failure

### Reproduction

Steps to reproduce:

1. Point a client at a host that is not listening, so fetch fails at the network level rather than returning an HTTP error:

```
   const authClient = createAuthClient({ baseURL: "http://localhost:9" });
```

2. Call any method using the documented error-as-value pattern:

```
   const { data, error } = await authClient.signIn.email({
     email: "a@b.com",
     password: "password1234",
   });
```
3. Observe that the await throws instead of resolving, so `error` is never read:
```
   Uncaught (in promise) TypeError: Failed to fetch
       at betterFetch
       at async $fetch
       at async Object.apply
```
   `onError` is not called either. A call made without await/.catch() surfaces as an unhandled rejection and reaches global error handlers.

Current: every authClient.* call rejects with TypeError: Failed to fetch when the API host is unreachable. onError is not invoked and no error value is returned.

Expected: network failures resolve to { data: null, error } and invoke onError, as the client docs describe ("Most client functions return a response object containing data and error properties"), and as the client behaved from 1.1.7.

This was fixed once and has regressed. Commit c8c65404 ("fix: return response error on network failure") wrapped the default fetch:

```
    customFetchImpl: async (input, init) => {
      try {
        return await fetch(input, init);
      } catch (error) {
        return Response.error();
      }
    },
```

In #987 a maintainer confirmed "Response error is returned on network failures starting from 1.1.7", and #657 was closed as completed for the same symptom (signIn.email not calling onError on a fetch failure).

That guard is no longer present. packages/better-auth/src/client/config.ts:85 on main reads:

```
    customFetchImpl: fetch,
    ...restOfFetchOptions,

```
@better-fetch/fetch only converts a thrown network error into an error value when catchAllError is set (see createFetch in its dist/index.js), and catchAllError does not appear anywhere in better-auth. So the bare fetch throw propagates out of $fetch and out of every client method.

Was removing the guard deliberate? If network failures are meant to throw now, the client error-handling docs should say so, since the current wording implies otherwise and the previous behaviour was the documented resolution of #657 and #987. If it was not deliberate, restoring the wrapper in getClientConfig looks like a one-line change.

1.6.22

### System info

```bash
{
  "system": {
    "platform": "linux",
    "arch": "x64",
    "release": "6.18.50-2-lts",
    "cpuCount": 4,
    "totalMemory": "7.40 GB"
  },
  "node": {
    "version": "v24.20.0",
    "env": "development"
  },
  "packageManager": {
    "name": "bun",
    "version": "1.3.5"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "1.6.22",
    "config": null
  }
}
```

Client, Package

```typescript
// Client side. The server config is not involved: the failure happens before a
// request reaches it.

  baseURL: process.env.NEXT_PUBLIC_SERVER_URL,
});
```

### Additional context

My agent tested against 1.6.22 and read main to confirm the line is still there.

the documented pattern is error-as-value, so application code written against the docs does not wrap these calls in try/catch. A brief network blip then turns ordinary UI code into unhandled rejections. In our app a fire-and-forget authClient.signIn.social(...) in a click handler produced 135 unhandled rejections in our error tracker inside a single 4 minute window, in bursts matching a user repeatedly clicking a button that silently did nothing.

vjymisal0:
### Summary
- Restore the client fetch error wrapper for network-level failures.
- Convert rejected `fetch` calls into `Response.error()` responses so better-fetch returns `{ data: null, error }`.
- Ensure the configured `onError` hook runs for unreachable hosts.

### Verification
- Added a regression test covering the error response and `onError` callback.
- `pnpm exec vitest run packages/better-auth/src/client/config.test.ts --no-file-parallelism` (1 test passed)
- Biome and cspell passed for changed files.
- Full package test run was attempted; unrelated workspace package builds are required first and caused existing module-resolution failures.

Fixes #11284

---
## Summary by cubic
Fixes #11284 by normalizing network fetch failures so client requests no longer reject. Previously, unreachable hosts caused `fetch` to reject and skip the configured `onError` hook; now those failures are converted to `Response.error()` responses, so `better-fetch` returns `{ data: null, error }` and `onError` runs with a status-0 error. AbortError rejections still propagate unchanged.

<sup>Written for commit f3097d238791235ed32c316e02746c82a6209552. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11302?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


vjymisal0:
Thanks for the careful review. Network failures are still normalized to status-zero responses, but AbortError is now rethrown so caller cancellations and timeouts preserve their existing behavior. I also added an AbortError regression test and the required @see issue links. Diff checks passed; pushed as f3097d238.

vjymisal0:
Thanks @greptile-apps and @cubic-dev-ai for the review. I confirmed the follow-up is already pushed in f3097d2: AbortError rejections are preserved, ordinary network failures still normalize to status-0 responses, and the regression tests include the required issue links. Verified with the focused config test (2/2) and Biome checks. The remaining Vercel failures require better-auth Team authorization rather than a code change.

vjymisal0:
Thanks for the review! I rechecked the branch at f3097d2: Greptile confirms the AbortError regression is covered and the implementation is safe. The remaining Vercel failures are deployment authorization for the better-auth Team, not code or test failures, so there is no additional code change to push. The targeted client tests pass locally (2/2). A Team member will need to authorize the deployment.

vjymisal0:
Rechecked the latest commit f3097d2: Greptile confirms the AbortError regression is addressed, and the focused client config tests pass (2/2). The only failing checks are the two Vercel deployments, which require authorization by a better-auth Team member; there are no remaining code issues to fix on this branch.

vjymisal0:
The latest review confirms the network-error normalization and abort handling are correct, with regression coverage and changeset present. The only failing checks are Vercel deployment authorization checks; no source fix is indicated. Please rerun or authorize the deployments.

vjymisal0:
The two failing Vercel checks report “Authorization required to deploy,” while the code review reports the implementation and tests are addressed. Could a maintainer authorize the preview deployments or rerun the Vercel checks?

PR #11307, opened 2026-09-16
feat(captcha): allow a custom fetch implementation for siteverify

### LINKED ISSUE #11306: captcha: allow a custom fetch implementation for the siteverify call

The captcha plugin's siteverify call always goes through the global `fetch`.
`siteVerifyURLOverride` can point that call at a different URL, but there is no
way to supply the fetch implementation itself — so verification cannot be
resolved without a network round trip. `betterFetch` already accepts
`customFetchImpl`; the plugin just doesn't forward one.

That matters because the gate is fail-closed (correctly): if the siteverify call
throws, `onRequest` returns 500 `UNKNOWN_ERROR` and the credential request fails.
The *transport* therefore becomes an availability dependency of every sign-in,
even where the verification *answer* is a foregone conclusion.

Concrete case — a Cloudflare Workers app running locally and in CI with
Cloudflare's published always-pass Turnstile test secret
(`1x0000000000000000000000000000000AA`):

1. Verifying against `challenges.cloudflare.com` makes every local sign-in depend
   on external egress. A transient failure there 500s the sign-in.
2. Pointing `siteVerifyURLOverride` at a local mock server removes the egress but
   keeps the transport, which traded one flake for another: workerd closes an
   idle keep-alive connection at exactly 5.000s, and a siteverify issued into the
   ~45ms window where that close is in flight fails with `Network connection
   lost`. Measured across five full e2e runs, every failure landed 5.005–5.048s
   after the previous siteverify, with none among ~200 calls outside that window
   — roughly 2% of sign-ins, surfacing as unrelated-looking test flakes.

There is no third URL to point at that isn't a transport. The missing seam is the
fetch itself.

Add `customFetchImpl?: FetchEsque` to `BaseCaptchaOptions` and forward it through
`handlerParams` into each verify handler's existing `betterFetch` call.

Purely additive: unset, behaviour is unchanged. When set, every other part of the
gate still runs — endpoint matching, the missing-token 400, the fail-closed catch
— only the network call is answered by the host.

```ts
captcha({
  provider: "cloudflare-turnstile",
  secretKey: TEST_SECRET,
  // local/test only
  customFetchImpl: () => Promise.resolve(Response.json({ success: true })),
});
```

- **Retrying the siteverify call.** Not viable in general: Turnstile tokens are
  single-use, so a retry of a call that did reach Cloudflare returns
  `timeout-or-duplicate` and 403s a legitimate user.
- **`siteVerifyURLOverride` at a local mock.** What we do today; it removes the
  external dependency but not the transport, which is where the failures come
  from (above).
- **Reimplementing the gate outside the plugin.** Duplicates endpoint matching
  and the status contract, and drifts on every release.

### Additional context

Also useful for testing the gate without mocking `betterFetch` at module level,
and for runtimes whose outbound transport is not the global fetch.

msnelling:
Closes #11306.

## What

Adds `customFetchImpl?: FetchEsque` to `BaseCaptchaOptions`, forwarded through
`handlerParams` into the `betterFetch` call every verify handler already makes
(turnstile, reCAPTCHA, hCaptcha, CaptchaFox).

Purely additive — unset, behaviour is unchanged and the call still goes to the
provider over the global fetch.

## Why

`siteVerifyURLOverride` can point the siteverify call at a different URL, but the
call always goes through the global fetch, so verification can never be resolved
without a network round trip. The gate is fail-closed (correctly), so the
transport becomes an availability dependency of every credential request even
where the verification answer is a foregone conclusion — e.g. running locally
with Cloudflare's always-pass test secret.

In our case (Cloudflare Workers) that cost ~2% of local sign-ins: pointing
`siteVerifyURLOverride` at a local mock removed the external egress but kept an
HTTP call, and workerd closes an idle keep-alive connection at exactly 5.000s, so
a siteverify issued into the ~45ms window where that close is in flight fails with
`Network connection lost` and 500s the sign-in. Full measurements in the issue.

With an injected fetch the whole gate still runs — endpoint matching, the
missing-token 400, the fail-closed catch — and only the network call is answered
by the host.

## Tests

Two cases in `captcha.test.ts`: the option is forwarded to the siteverify call
when configured, and left `undefined` when it is not.

`pnpm vitest src/plugins/captcha/captcha.test.ts --run` → 42 passed.

---
## Summary by cubic
Adds an optional `customFetchImpl` to `captcha()` so the siteverify call can be resolved in-process instead of over the global fetch. Previously `siteVerifyURLOverride` could only change the URL, never the transport, which made verification an availability dependency even when the answer was predetermined (e.g. Cloudflare's always-pass test secret locally). Unset, behavior is unchanged; all four verify handlers (Turnstile, reCAPTCHA, hCaptcha, CaptchaFox) forward the option, and the fail-closed gate still runs in full. Closes #11306.

<sup>Written for commit da3427b61ed98c2333080e81c2e869ede737fc98. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11307?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bytaesu:
I will revisit this PR after discussing in the issue.

Since the PR is already linked, I will close it for now and communicate in the issue first.

PR #11309, opened 2026-09-16
fix(organization): reject role names containing the list separator

### LINKED ISSUE #11303: organization: `create-role` and `update-role` accept a comma in the role name, persisting a role that can never be assigned

### Reproduction

Minimal repro on a clean `better-auth@1.7.5` project (in-memory adapter, no external DB):

```js
// repro.mjs  —  node repro.mjs
import { organization } from "better-auth/plugins";
import { createAccessControl } from "better-auth/plugins/access";
import { adminAc, defaultStatements, memberAc, ownerAc } from "better-auth/plugins/organization/access";

const statement = { ...defaultStatements, project: ["read", "write"] };
const ac = createAccessControl(statement);
const roles = {
  owner: ac.newRole({ ...ownerAc.statements, project: ["read", "write"] }),
  admin: ac.newRole({ ...adminAc.statements, project: ["read"] }),
  member: ac.newRole({ ...memberAc.statements, project: ["read"] }),
};
const db = { user: [], session: [], account: [], verification: [], organization: [], member: [], invitation: [], organizationRole: [], team: [], teamMember: [] };

const auth = betterAuth({
  secret: "a-very-long-test-secret-value-for-repro-0123456789",
  database: memoryAdapter(db),
  plugins: [organization({ ac, roles, dynamicAccessControl: { enabled: true } })],
});

const signUp = await auth.api.signUpEmail({
  body: { email: "owner@example.com", password: "password12345", name: "Owner" },
  returnHeaders: true,
});
const headers = new Headers({ cookie: (signUp.headers.get("set-cookie") ?? "").split(";")[0] });
const org = await auth.api.createOrganization({ body: { name: "Acme", slug: "acme" }, headers });

// 1. create a role whose NAME contains the list separator
const created = await auth.api.createOrgRole({
  body: { organizationId: org.id, role: "owner,temp", permission: { project: ["read"] } },
  headers,
});
console.log("created:", created.success, created.roleData.role);

// 2. try to assign it
try {
  await auth.api.updateMemberRole({
    body: { organizationId: org.id, memberId: db.member[0].id, role: "owner,temp" },
    headers,
  });
} catch (e) { console.log("assign rejected:", e.body); }

// 3. rename a clean role into a comma name
await auth.api.createOrgRole({ body: { organizationId: org.id, role: "reporter", permission: { project: ["read"] } }, headers });
const renamed = await auth.api.updateOrgRole({
  body: { organizationId: org.id, roleName: "reporter", data: { roleName: "owner,reporter" } },
  headers,
});
console.log("renamed to:", renamed.roleData.role);
console.log("stored roles:", db.organizationRole.map((r) => r.role));
```

Observed output:

```
created: true owner,temp
assign rejected: { code: 'ROLE_NOT_FOUND', message: 'ROLE_NOT_FOUND: temp' }
renamed to: owner,reporter
stored roles: [ 'owner,temp', 'owner,reporter' ]
```

**Current:** `create-role` and `update-role` accept any string as a role name and persist it verbatim. The body schema is `role: z.string()` with no character validation, and `normalizeRoleName` only lowercases. The reserved-name guard compares the whole string, so `["owner","admin",...].includes("owner,temp")` is `false` and the name is stored. Both a fresh create and a rename succeed.

**Expected:** a role name containing `,` is rejected at creation and at rename, because `,` is the list separator: `Member.role` stores roles comma-joined and `hasPermissionFn` resolves them with `role.split(",")`.

The resulting row is inert but incoherent. As the output above shows, the role can never be assigned — `updateMemberRole` splits the value and validates each token, so `owner,temp` fails with `ROLE_NOT_FOUND: temp`. So creation succeeds in producing a role that no endpoint can ever use, which still occupies a row, appears in `list-roles`, and counts against `maximumRolesPerOrganization`. In-use checks that compare split tokens (for example delete-role's scan for members still holding a role) can never match it either.

1.7.5

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "release": "25.6.0",
    "cpuCount": 10,
    "cpuModel": "Apple M5",
    "totalMemory": "24.00 GB"
  },
  "node": { "version": "v24.18.0", "env": "development" },
  "packageManager": { "name": "npm", "version": "11.16.0" },
  "frameworks": null,
  "databases": null,
  "betterAuth": { "version": "1.7.5", "config": null }
}
```

Backend

```typescript
import { organization } from "better-auth/plugins";
import { createAccessControl } from "better-auth/plugins/access";
import { defaultStatements, ownerAc } from "better-auth/plugins/organization/access";

const ac = createAccessControl({ ...defaultStatements, project: ["read", "write"] });

  plugins: [
    organization({
      ac,
      roles: { owner: ac.newRole({ ...ownerAc.statements, project: ["read", "write"] }) },
      dynamicAccessControl: { enabled: true },
    }),
  ],
});
```

### Additional context

**This is not a security report.** The privilege-escalation path through this was fixed by #9962 and I verified that independently against the published tarballs: 1.6.16 affected / 1.6.17 not, and 1.7.0-beta.5 affected / 1.7.0-beta.6 not. On 1.7.5 every web-reachable writer of `Member.role` splits before resolving (`update-member-role` and the invitation paths; `addMember` is server-only), which is why the assign step above is rejected. Filing this as ordinary hardening, not a vulnerability.

**Why it still seems worth fixing.** #9962's changeset already states the principle: *"unknown or malformed role values are rejected instead of being persisted."* That was applied to `updateMemberRole`; the creation endpoints still persist malformed names. #9962 also added a test named `"should not allow a comma-delimited role string"` — that guarantee currently holds only at the assignment chokepoint. `hasPermissionFn` still splits unconditionally, so the invariant "a role name never contains the list separator" depends on every present and future writer of `Member.role` remembering to split first. Validating at creation removes that class of regression instead of defending against it one endpoint at a time.

**Note on the pending rewrite (#7886).** This survives it: the new `packages/organization/src/addons/dynamic-access-control/routes/create-role.ts` still uses `role: z.string().min(1)` plus a whole-string reserved check, and `packages/organization/src/helpers/validate-roles.ts` only validates that roles are *recognized* — it splits on `,` itself. `validate-roles.ts` looks like the natural home for a shared name-validation rule if you'd prefer it centralized.

**Suggested fix.** Reject `,` in role names at `create-role` and at `update-role` (`data.roleName`); cheapest form is a refinement on the existing zod schema.

seifsheikhelarab:
Closes #11303

## Description

`create-role` and `update-role` accepted role names containing a comma. Member roles are stored comma-joined on `member.role` and resolved by splitting on `,` (`hasPermissionFn`, member role updates, delete-role's in-use scan), so a dynamic role named `owner,temp` could be created but never assigned — every web-reachable writer of `Member.role` splits before resolving. The resulting row is inert but incoherent: it occupies a `maximumRolesPerOrganization` slot, appears in `list-roles`, and the in-use checks that compare split tokens can never match it.

#9962 established the principle that "unknown or malformed role values are rejected instead of being persisted" and applied it to `updateMemberRole` (the assignment chokepoint). This PR extends the same guarantee to the creation chokepoints, so the invariant "a role name never contains the list separator" no longer depends on every writer of `Member.role` remembering to split first.

## Changes

- Added `INVALID_ROLE_NAME` ("Role name contains a reserved character") to `ORGANIZATION_ERROR_CODES`, mirroring the existing `INVALID_TEAM_ID` convention for comma-joined values
- `create-role` and `update-role` (`data.roleName`) now reject names containing `,` with a `400` and that error code, via a shared `checkForInvalidRoleName` helper following the file's existing `check*` helper pattern. Runs right after `normalizeRoleName`, before any DB write
- Regression test with `@see` for the issue: create rejects a comma name (nothing persisted), and renaming an existing role into a comma name is rejected
- Docs: warning callout under "Creating a role" and a note under "Updating a role" in the organization plugin docs
- Changeset: `better-auth` patch

## Testing

- The new regression test failed first and reproduced the issue exactly (role `comma,role` was persisted), then passes with the fix
- Full organization plugin suite passes: 254 tests (including the existing #9962 tests)
- `pnpm typecheck` passes

---
## Summary by cubic
Fixes organization role creation and renaming to reject comma-containing names, which previously could be created but never assigned because member roles are stored comma-joined and resolved by splitting on `,`. Now `create-role` and `update-role` return a `400` with the `INVALID_ROLE_NAME` error code, and the rejected name is logged as JSON metadata so control characters can't forge log records, closing #11303.

<sup>Written for commit b2417799b8bb37e1b984d79b30a762ec8cdcec6f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11309?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/better-auth/src/plugins/organization/routes/crud-access-control.test.ts
aleksola: Ijywu

PR #11316, opened 2026-09-17
fix(react): preserve auth query snapshots during hydration

### LINKED ISSUE #10972: BUG: `useSession()` causes a React hydration mismatch: `useStore` passes `getSnapshot` as `getServerSnapshot`

Smidge:
## Why

When an auth query resolves before a streamed React subtree hydrates, `useSyncExternalStore` currently reads the live value for its server snapshot. The subtree then renders resolved content against pending server HTML and React reports a hydration mismatch.

Closes #10972. This is an alternative follow-up to #11177, building on @Bekacru's diagnosis and stable server snapshot approach.

## What changed

Auth query atoms now retain their initial snapshot behind an internal symbol. The React adapter uses that snapshot for server rendering and hydration, then subscribes to the live value as before. This covers both the session atom and plugin queries created by `useAuthQuery`, without adding a public hook option.

The distinction is explicit rather than inferred from Nano Stores' `init`: ordinary atoms can start with `undefined`, and computed stores also retain an `init` property. Applying either an `init !== undefined` or an `"init" in store` rule to arbitrary plugin stores can change their SSR output. Ordinary and computed plugin stores keep their existing behavior here.

Includes a patch changeset and documentation of the pending snapshot and server-session fallback. Client-only rendering still reads the current value immediately.

## Validation

- On a fresh build of `main`, four regression cases fail with React hydration errors: signed-in and signed-out late subscribers, `hydrateSession` before hydration, and a resolved organization query. All pass with this change.
- The fifth test preserves SSR values for an ordinary atom initialized with `undefined`, an ordinary atom with a defined initial value, and a computed plugin store.
- All client tests pass: **123 tests in 10 files**, including subsequent query updates.
- `pnpm exec turbo build --filter=better-auth...` passes.
- `pnpm exec tsc --build packages/better-auth/tsconfig.json --force` passes, including its referenced packages.
- Repository commit hooks pass: Biome, spelling, and documentation formatting.
- Full `pnpm typecheck` was attempted, but the local install was restricted to Better Auth and its dependencies after running out of disk space. Other workspace packages report missing dependencies; full-workspace verification still needs CI.

@gustavovalverde, GitHub has requested your review through the code owners. Could you review the query-only scope and approve the fork workflows? CI, changeset verification, and E2E currently require maintainer approval; the Vercel previews also need team authorization. Happy for this to replace #11177 or for the implementation and tests to be folded into it.


bytaesu (approved):
@Smidge Thank you!

PR #11319, opened 2026-09-17
fix(prisma-adapter): validate capitalized custom model names

### LINKED ISSUE #11310: Prisma validateSchema false-positives on capitalized custom modelNames — all auth requests fail with SCHEMA_MISMATCH

## Reproduction

1. Prisma schema using conventional capitalized model names (PostgreSQL):

```prisma
model Account {
  id            String    @id
  email         String    @unique
  emailVerified Boolean   @default(false)
  name          String
  createdAt     DateTime  @default(now())
  updatedAt     DateTime  @updatedAt
  sessions      Session[]
}
// ... Session / AuthIdentity (account) / Verification models likewise
```

2. Configure auth with matching capitalized `modelName`s:

```ts
  database: prismaAdapter(prisma, { provider: "postgresql" }),
  user: { modelName: "Account" },
  session: { modelName: "Session" },
  account: { modelName: "AuthIdentity" },
  verification: { modelName: "Verification" },
})
```

3. Run migrations, generate the client, then call any auth endpoint (e.g. `POST /sign-up/email` or `auth.api.signUpEmail`).

Every auth request fails with `500` / `SchemaMismatchError`:

```
ERROR [Better Auth]: Prisma schema mismatch

  Missing tables
    Account, Session, AuthIdentity, Verification

  help: Run `npx auth generate` to refresh the Prisma schema, then run `prisma migrate`.
```

although all four tables exist with the correct columns (verified directly in the database). Expected: validation passes, since the schema is correct.

Root cause: one-sided normalization in the comparison.

- `packages/prisma-adapter/src/schema-check.ts` — `introspectPrismaDataModel` keys actual tables by `clientProperty(model)` (first letter lowercased): `Account` → `account`.
- `packages/core/src/db/schema-diff.ts` — `getExpectedSchema` keys expected tables by `table.modelName` verbatim: `Account`.

So `Account` never equals `account`. Note the reverse case also exists, which rules out "just stop lowercasing": plugin tables use lowercase-first `modelName`s (e.g. `oauthClient`) while Prisma models are conventionally capitalized (`OauthClient`), so normalizing only one side can never satisfy both. The fix is to normalize **both** sides to client-property form — which is also how the adapter addresses the client at runtime (`db[model]`):

```ts
export function findPrismaSchemaProblems(
  dataModel: PrismaRuntimeDataModel,
  options: BetterAuthOptions,
  usePlural?: boolean | undefined,
): SchemaFinding[] {
  const expected = getExpectedSchema(options, { usePlural });
  const normalized = Object.fromEntries(
    Object.entries(expected).map(([name, table]) => [clientProperty(name), table]),
  );
  return diffSchema(normalized, introspectPrismaDataModel(dataModel));
}
```

I verified this locally (patched package): previously-failing sign-up succeeds with validation enabled, and a genuinely missing table is still reported.

`better-auth@1.7.5`, `@better-auth/core@1.7.5`, `@better-auth/prisma-adapter@1.7.5` (also reproducible on 1.7.3–1.7.4, i.e. since validation was enabled by default)

## System info

better-auth 1.7.5, @better-auth/core 1.7.5, @better-auth/prisma-adapter 1.7.5, prisma 7.9.1, PostgreSQL 16 (docker), bun 1.4.0, darwin arm64

Backend

## Auth config

(see Reproduction above — minimal config; full config available on request, no sensitive data involved)

## Additional context

- Runtime CRUD is unaffected — only the startup/request schema check is broken, so this bites every Prisma user with capitalized custom `modelName`s (the Prisma naming convention) on 1.7.3+.
- Current workaround: `advanced: { database: { validateSchema: false } }`, which disables a useful guardrail.
- Checked existing issues before filing: #6391 / #1901 / #1414 cover adjacent `modelName` casing problems but not this validation comparison; no open issue tracks this one.
- Happy to open a PR with the fix above plus a regression test (capitalized core `modelName` + oauth-provider plugin tables) if that helps.

bytaesu:
Closes #11310

---
## Summary by cubic
Fixes schema validation in `@better-auth/prisma-adapter` so capitalized custom Prisma model names (e.g., `Account` for user) no longer produce false validation errors. Closes #11310.

- Maps expected schema keys to Prisma client property names before diffing.
- Adds tests for capitalized custom model names and lower-camel plugin model names.

<sup>Written for commit b1b3f37f8f104417901079ee73023aef33f2b283. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11319?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


antonio-ivanovski:
Reverted my commit that removed the custom model names (https://github.com/antonio-ivanovski/spliit-cloud/commit/091554c96a0dd9da02495d4940d39818bdf1e6c5) and tried the fixed version. It is working correctly and no failure could be observed. Thanks for the fix, hope it fixes for other users, for me, i will be sticking to the default names (should have had these in first place).

bytaesu:
> Reverted my commit that removed the custom model names ([antonio-ivanovski/spliit-cloud@091554c](https://github.com/antonio-ivanovski/spliit-cloud/commit/091554c96a0dd9da02495d4940d39818bdf1e6c5)) and tried the fixed version. It is working correctly and no failure could be observed. Thanks for the fix, hope it fixes for other users, for me, i will be sticking to the default names (should have had these in first place).

Thanks for letting me know 🙏

PR #11320, opened 2026-09-17
fix(oauth-provider): defer resource seeding until Drizzle schema is ready

Tushar-Khandelwal-2004:
## Summary

Fixes schema generation for apps that add the MCP plugin before their Drizzle schema includes the OAuth resource tables.

MCP configures an OAuth protected resource, which triggers resource seeding during initialization. Previously, a Drizzle schema missing `oauthResource` caused initialization to fail before `auth generate` could update that schema. Resource seeding now defers in this transitional state and retries on first resource access after the schema is regenerated.

Also clarifies that Drizzle Relations v2 requires Drizzle ORM `1.0.0-rc.1` or later.

---
## Summary by cubic
Fixes OAuth resource seeding so apps that add the MCP plugin before their Drizzle schema includes the OAuth resource tables no longer fail during initialization. Seeding defers in this transitional state and retries on first resource access after `auth generate` regenerates the schema.

- Catches the Drizzle adapter's missing-model error in addition to missing-table errors.
- Documents that Relations v2 requires Drizzle ORM `1.0.0-rc.1` or later.

<sup>Written for commit 985e2664203a1babff4a48f5b6fc44f0bca02d32. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11320?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


docs/content/docs/adapters/drizzle.mdx
```diff
@@ -254,6 +254,12 @@ Install the adapter:
 npm install @better-auth/drizzle-adapter
 ```
 
+<Callout type="warn">
+  Relations v2 requires Drizzle ORM `1.0.0-rc.1` or later. If you use the
+  Relations v2 adapter with an earlier Drizzle release, generated schemas fail
+  at runtime because `defineRelationsPart` is unavailable.
+</Callout>
+
```
bytaesu: This seems pretty self-explanatory. "Relations v2" is a term from Drizzle, so I don't think we need a warning here.

Tushar-Khandelwal-2004: Removed the warning and updated the PR. Thanks for the feedback.

PR #11335, opened 2026-09-19
fix(auth): fail closed when deleting sessions from secondary storage

### LINKED ISSUE #11326: `deleteUserSessions` reports success when the secondary-storage eviction fails (regression in 1.7)

### Reproduction

In 1.6, `internalAdapter.deleteUserSessions` evicted the cached sessions first
and deleted the database rows afterwards, both awaited:

```js
// 1.6.33, db/internal-adapter.mjs
deleteUserSessions: async (userId) => {
  await deleteSecondaryStorageSessions(userId);
  if (databaseStoresSessions && !preservesDatabaseSessions) await deleteDatabaseSessions(userId);
},
```

A `secondaryStorage.delete` that rejects therefore aborted the revocation before
anything was applied: the caller saw the error, the rows were untouched, and a
retry completed the whole operation.

In 1.7 the order is inverted and the eviction is queued behind an
after-transaction hook whose error handler only logs:

```js
// 1.7.4, db/internal-adapter.mjs
async function queueCachedUserSessionDeletion(userId, sessionReferences) {
  if (!secondaryStorage) return;
  const references = sessionReferences ?? await getActiveSessionReferences(userId);
  await queueAfterTransactionHook(() => deleteCachedUserSessions(userId, references), {
    onError(error) {
      logger.error("Failed to delete committed user sessions from secondary storage", error);
    },
  });
}
```

So when the secondary storage is unreachable, `deleteUserSessions` **resolves**,
the session rows are gone, and the cached tokens are still there. Because
`findSession` answers from the cache without re-reading the row, every session
the call was meant to revoke keeps working until its cache entry expires — and
the caller, told it succeeded, never retries.

## Why it matters

This is the primitive behind password reset (`revokeSessionsOnPasswordReset`),
account deactivation and account deletion. In all three, "revoked" has to mean
the session stops working. A cache outage silently turns each of them into a
no-op that reports success, for as long as the cached entries live — in our case
up to a day, and without a TTL cap it would be the session's full lifetime.

It is also invisible: the only trace is one log line from `onError`, on a call
that returned normally.

## Reproduction

`better-auth@1.7.4`, a memory database adapter, and a secondary storage whose
`delete` rejects the way a real client does when its backend is unreachable.

```js
import { memoryAdapter } from 'better-auth/adapters/memory';

const db = { user: [], session: [], account: [], verification: [] };
const cache = new Map();

const auth = betterAuth({
  baseURL: 'http://localhost:3000',
  secret: 'a-test-secret-with-enough-entropy-for-hs256',
  database: memoryAdapter(db),
  secondaryStorage: {
    get: async (key) => cache.get(key) ?? null,
    set: async (key, value) => void cache.set(key, value),
    delete: async () => {
      throw new Error('ECONNREFUSED');
    },
  },
});

const ctx = await auth.$context;
const user = await ctx.internalAdapter.createUser(
  { email: 'a@example.com', name: 'A', emailVerified: true },
  { method: 'email-password' },
);
await ctx.internalAdapter.createSession(user.id);
const token = db.session[0].token;

let outcome = 'resolves';
try {
  await ctx.internalAdapter.deleteUserSessions(user.id);
} catch (err) {
  outcome = `rejects (${err.message})`;
}

console.log('call:          ', outcome);
console.log('rows left:     ', db.session.length);
console.log('token cached:  ', cache.has(token));
console.log('session usable:', !!(await ctx.internalAdapter.findSession(token)));
```

(The second `createUser` argument is required from 1.7 and ignored by 1.6, so the
same script runs on both.)

On 1.6.33:

```
call:           rejects (ECONNREFUSED)
rows left:      1
token cached:   true
session usable: true
```

On 1.7.4 and 1.7.5, identically — and the stack trace is the only sign anything
went wrong:

```
ERROR [Better Auth]: Failed to delete committed user sessions from secondary storage Error: ECONNREFUSED
    at deleteCachedUserSessions (better-auth/dist/db/internal-adapter.mjs:83:75)
    at queueAfterTransactionHook.onError (better-auth/dist/db/internal-adapter.mjs:99:41)
    at async Object.deleteUserSessions (better-auth/dist/db/internal-adapter.mjs:521:37)
call:           resolves
rows left:      0
token cached:   true
session usable: true
```

| after `deleteUserSessions()` with the cache down | 1.6.33 | 1.7.4 | 1.7.5 |
| --- | --- | --- | --- |
| call | rejects | **resolves** | resolves |
| session rows | 1 — retryable | **0 — deleted** | 0 |
| cached token still served | yes | **yes** | yes |

In 1.6 the end state is recoverable, because nothing was applied and the caller
knows. In 1.7 it is not: the rows are gone, so a retry has nothing left to find,
and the caller was told it worked.

1.7.*

### System info

```bash
{
  "system": {
    "platform": "darwin",
    "arch": "arm64",
    "version": "Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:15 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6041",
    "release": "25.5.0",
    "cpuCount": 12,
    "cpuModel": "Apple M4 Pro",
    "totalMemory": "48.00 GB",
    "freeMemory": "0.18 GB"
  },
  "node": {
    "version": "v24.14.1",
    "env": "development"
  },
  "packageManager": {
    "name": "npm",
    "version": "11.11.0"
  },
  "frameworks": null,
  "databases": null,
  "betterAuth": {
    "version": "Unknown",
    "config": null
  }
}
```

Backend

```typescript

```

## What we would like

Some supported way to make a revocation fail closed — today there appears to be
none that does not reach into library internals.

Any one of these would be enough for us:

1. **Await the eviction and let it reject** (optionally behind a flag, e.g.
   `session.revocation: 'fail-closed'`). Closest to 1.6, and the DB delete then
   only happens when the cache eviction actually succeeded.
2. **Evict on both sides of the DB delete**, and let the second one reject —
   keeps the deletion durable while still reporting a cache that could not be
   cleared.
3. **A configurable error policy for `secondaryStorage`**, so an application can
   say that a failed `delete` is fatal while a failed `get`/`set` is not. That
   distinction is the one that matters: reads and writes should degrade to the
   database, deletes should not be allowed to fail quietly.
4. Failing those, **report it in the return value** — anything the caller can
   branch on beats a log line on a resolved promise.

Vishesh-Verma-07:
What changed and why

Regression introduced in the 1.7 release line: deleteUserSessions and deleteUser no longer fail closed when secondaryStorage is
configured (Closes #11326).

In 1.6 the cached session eviction was awaited inline, so any eviction failure rejected the request and left the database rows
untouched - a retry could complete the revocation. In 1.7 that path was moved to queueCachedUserSessionDeletion, which defers the
eviction to an after-transaction hook wrapped in a swallowing logger.error handler. When eviction fails the route now reports
success while the revoked sessions keep working from the cache (e.g. a diff against a cached value that throws leaves the user
logged in indefinitely).

The change evicts the user's cached sessions inline and awaited, before any database row is touched:

- An eviction failure rejects the call before the DB mutation runs, so nothing is partially deleted and a retry completes the
whole revocation from an intact state.
- The existing semantics for storeSessionInDatabase (DB rows still deleted, preserveSessionInDatabase still ends preserved
sessions) are unchanged - only the ordering vs. the cache is fixed.

The queueCachedUserSessionDeletion helper (and its suppressed-error hook) is removed.

Breaking changes

None - patch, no API changes. Drafted as a restoration of the pre-1.7 contract; note that users of 1.7 who relied on the silent-
skip behavior will now see an error on eviction failure instead.

Testing

Reworked/extended integration coverage in packages/better-auth/src/db/secondary-storage.test.ts asserting the fail-closed
ordering for both deleteUserSessions and deleteUser, including the preserved-session and storeSessionInDatabase branches.
- pnpm vitest packages/better-auth/src/db/internal-adapter.test.ts packages/better-auth/src/db/secondary-storage.test.ts - 95
tests passing.
- pnpm typecheck passes.

Changeset

.changeset/delete-session-fail-closed.md - patch (restores the pre-1.7 contract; no new API).

---
## Summary by cubic
Fixes a 1.7 regression where `deleteUserSessions` and `deleteUser` could report success even when cached session eviction from `secondaryStorage` failed, leaving revoked sessions usable (Closes #11326).

- Eviction is now awaited inline before any database row is touched; on failure the call rejects before the mutation runs, so a retry completes from an intact state.
- Removes the deferred eviction helper and its suppressed-error hook; eviction also runs when row deletion is vetoed, so the cache reflects surviving rows.
- Database deletes target only the captured session tokens so a session created during the cache-first await keeps its cache entry and row; an empty snapshot falls back to a user-wide sweep.
- Users on 1.7 who relied on the silent-success path now see an error instead.
- Patch-only change with no API surface changes; adds integration coverage for fail-closed ordering, retries, vetoed deletes, and preserved-session branches.

<sup>Written for commit 8740ffd86bdff3be403c2acc76f274ec4cf3410f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/better-auth/better-auth/pull/11335?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Vishesh-Verma-07:
Thanks for the reviews — here's where each point landed.

**P1 / "Concurrent sessions survive deletion" — fixed in `9f9c0e45f`**

`deleteUser` and `deleteUserSessions` no longer do a user-wide database delete after the cache-first await. The database work is now scoped to the captured revocation set (`deleteManyWithHooks` with `token IN (…)`), so a session created during the await keeps both its cache entry and its row instead of a cached session outliving the row it referred to. When the cache snapshot is empty (e.g. `storeSessionInDatabase` with no active-cache record), the code falls back to the user-wide sweep so database-only sessions are still cleaned up.

**P3 / partial eviction — fixed in `9f9c0e45f`**

Added a test in `packages/better-auth/src/db/secondary-storage.test.ts` with two sessions and a storage whose `delete` fails on the second call. It asserts:
- the call rejects with the secondary-storage failure,
- exactly one token key was already evicted while the active-session index still lists both,
- both database rows remain intact,
- a retry completes the revocation (both keys and both rows gone).

This also addresses the "resumable" half of the P2 partial-eviction point: on retry, `deleteCachedUserSessions` re-reads the active-session index rather than assuming an intact pre-revocation state, so a partial eviction is resumable by construction.

**P2 / rollback trade-off — intentional, previously existing, and documented**

Evicting inline is the pre-existing contract: since v1.6.33 `deleteUserSessions` has called `evictCachedUserSessionsInTransaction` inline, and `deleteUser`'s session delete was inline. This PR deliberately does not restore the removed queued-commit behavior; the upstream issue (#11326) explicitly lists "option 1 — keep the 1.6 behavior" as an acceptable fix and this PR is the closest match. No in-tree caller (`update-user.ts`, `admin/routes.ts`) wraps these methods in `runWithTransaction`, and with `getAndDelete` there is no rollback primitive available to compensate. Please do not treat this as departing from the documented contract — it is exactly the historic contract.

One factual correction: it's stated that the rollback coverage "was deleted". That is inaccurate — the test was rewritten rather than removed. `internal-adapter.test.ts` still asserts the rollback scenario, now as "evicts cached sessions inline even within a transaction", which locks in the current contract instead of the removed one. If you want a compensating that's a separate, larger change and I'd be happy to file an issue for it.

**Vercel checks**

The `Vercel Security Review`/deployment checks are failing with "Authorization required to deploy" — this needs a maintainer to authorize deployment for the fork PR; nothing on the branch side can unblock it. The `Validate PR title` check passes after retitling.

Local verification: `pnpm vitest packages/better-auth/src/db/internal-adapter.test.ts packages/better-auth/src/db/secondary-storage.test.ts` (96 passing) and `pnpm typecheck` are green.

